PR #470, opened 2018-08-24
IIIF Attachment Server integration

robertahunt:
We, at Natural History Museum of Denmark, would like to add the ability to choose between different attachment servers. Specifically we are interested in having a IIIF compliant public server and another private attachment server for private documents. 

We are hoping this will become more general, so other institutions can add their own custom servers using a similar framework, however, we are not yet at that stage. This is an initial attempt at a minimum-viable-working-product for us, which we are hoping we can get feedback on.


specifyweb/settings/specify_settings.py
```diff
@@ -30,6 +30,10 @@
 # Set to true if asset server requires auth token to get files.
 WEB_ATTACHMENT_REQUIRES_KEY_FOR_GET = False
 
+PUBLIC_IMAGE_SERVER_FILEUPLOAD_URL = None
```
benanhalt: Maybe having a setting for the server base URL would be better so that it could be used for both getting and uploading? So like `PUBLIC_IMAGE_SERVER_BASE_URL = `. Then, your code can append `/upload_image` or `${filename}/...`

specifyweb/frontend/js_src/lib/attachmentplugin.js
```diff
@@ -52,7 +54,10 @@ module.exports =  UIPlugin.extend({
                 .append(self.progressBar)
                 .dialog({ modal:true });
 
-            attachments.uploadFile(file, function(progressEvt) {
+            if (isPublicImage) { var attachmentserver = attachmentserverpublic;}
```
benanhalt: It's a bit strange to declare a variable in the separate branches of an if-statement. More idiomatic is `var attachmentserver = isPublicImage ? attachmentserverpublic : attachmentserverprivate;`

specifyweb/frontend/js_src/lib/attachmentserverpublic.js
```diff
@@ -0,0 +1,60 @@
...
+  },
+  openOriginal: function(attachment) {
+      var attachmentlocation = attachment.get('origfilename');
+      var src = `http://localhost:5050/loris/${attachmentlocation}/full/full/0/default.jpg`;
+      window.open(src);
+  }
+};
+
+/*Inherit other functions from base server*/
+attachmentserverpublic.prototype = attachmentserverbase;
```
benanhalt: This won't do what you want. You need some combination of `Object.create()` and `Object.assign()`, probably. 

 Maybe like:
```
module.exports = Object.assign(Object.create(attachmentserverbase), attachmentserverpublic);
```

specifyweb/frontend/js_src/lib/attachmentserverpublic.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+var attachmentserverpublic = {
+  getThumbnail: function(attachment, scale) {
+      scale || (scale = 256);
+      var style = "max-width:" + scale + "px; " + "max-height:" + scale + "px;";
+
+      var attachmentlocation = attachment.get('attachmentlocation');
+      return functionthatjustpipesthings(attachmentlocation).pipe(function(token) {
+          return $('<img>', {src: `http://localhost:5050/loris/${attachmentlocation}/full/${scale},/0/default.jpg`, style: style});
```
benanhalt: If you used a base URL in the settings, then you could use that here:

E.g. `${settings.public_image_server_base_url}/${attachmentlocation}/full/${scale},/0/default.jpg`

specifyweb/frontend/js_src/lib/attachmentserverpublic.js
```diff
@@ -0,0 +1,60 @@
...
+      });
+  },
+  uploadFile: function(file, progressCB) {
+      var formData = new FormData();
+      var attachment;
+
+      formData.append('media', file);
+      var attachmentlocation = file.name;
```
benanhalt: Using the user supplied filename will cause name collision problems when more than one file with same name gets uploaded. Also depending on how smart your image server is, it could open things up to directory traversal attacks and such.

You may already be planning do more here. I'm just noting it to be thorough.

specifyweb/frontend/js_src/lib/attachmentplugin.js
```diff
@@ -34,15 +35,16 @@ module.exports =  UIPlugin.extend({
             return this;
         },
         addAttachment: function() {
-            this.$el.append('<form enctype="multipart/form-data"><input type="file" name="file"></form>');
-            this.$('input').click();
+            this.$el.append('<form enctype="multipart/form-data"><input type="checkbox" id="isPublicImage">  Is this a public image file?<br><input type="file" name="file"></form>');
```
benanhalt: The checkbox should probably only be included if the public image server is configured.

Also, it seems like the UI should force the user to pick public/private before choosing the file, because the file gets uploaded immediately (which is admittedly not ideal either, on my part).

specifyweb/settings/specify_settings.py
robertahunt: Good idea!

specifyweb/frontend/js_src/lib/attachmentplugin.js
robertahunt: Great! I didn't know how to do that in JS.

specifyweb/frontend/js_src/lib/attachmentserverpublic.js
robertahunt: Seems to work, thanks!

specifyweb/frontend/js_src/lib/attachmentserverpublic.js
robertahunt: Good idea!

specifyweb/frontend/js_src/lib/attachmentserverpublic.js
robertahunt: Yeah, this is indeed a problem. We are waiting to decide what the filename convention should be, so I just left it as the user uploaded filename for now.

robertahunt:
Hey Ben,
I also made a fork of the Specify 7 Wiki, with instructions on how to add your own custom attachment server. https://github.com/robertahunt/specify7-wiki/blob/master/Custom-Attachment-Servers.md

PR #783, opened 2021-03-25
Update example Apache configuration

MattBlissett:
Allow Apache access only to the required directories rather than the whole server.  Also update the documented WSGI parameter name.

And use HTTPS rather than Git protocol for clone (more likely to be allowed through a firewall), and update some changed links.

(Issues noticed while helping someone with a Specify installation. I don't run Specify myself, so please treat these changes as suggestions.)


benanhalt:
Thanks Matt.

PR #784, opened 2021-03-25
Add CentOS / Red Hat instructions.

MattBlissett:
This lists the Red Hat 7 / CentOS 7 packages needed to install Specify, equivalent to the Ubuntu ones.

There are also small changes, since the two distributions refer to Apache HTTPD as apache2 (Ubuntu) and httpd (Red Hat / CentOS).


benanhalt:
Thanks for the Red Hat info!

PR #792, opened 2021-04-13
Add wbplanview unit tests and pre-commit hooks

maxxxxxdlp:
Convert `testmain.js` to TypeScript
Add unit tests for functional `wbplanview` components
Add pre-commit hooks from `pre-commit.com`


.pre-commit-config.yaml
```diff
@@ -0,0 +1,146 @@
...
+# had changes.
+#
+# You can manuall trigger the hooks like this:
+# pre-commit run
+#
+# You can tell it to check specific files:
+# pre-commit run --files *.ts
+#
+# Or give it all files:
+# pre-commit run --all-files
```
benanhalt, line 22: Is there a way to run it just for the output without updating any files?

specifyweb/frontend/js_src/lib/tests/testmain.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+(cd specifyweb/frontend/js_src && npm run test)
```
benanhalt, line 2: Does this shell script get used anywhere?

maxxxxxdlp: I defined a local hook here:
https://github.com/specify/specify7/blob/f25006e0b1b70f63ecbc7984699f80a2d66a0027/.pre-commit-config.yaml#L40

All it needs is to run `npm run test` if any js or ts file was modified.

Since `npm` is a global command, rather than a local shell file, I wasn't able to get it to work nicely. So I created this local shell script that get's called by the hook.

Though, maybe there is a better way to do this.

.pre-commit-config.yaml
maxxxxxdlp: I don't think there is such parameter.
One solution would be to run `git stash && git stash apply` to save a copy of the working tree into stash. Then, after running pre-commit you can recover the previous state by running `git checkout stash -- .`

benanhalt:
You can go ahead and merge this if you want. Thanks for adding the tests.

I'd like to wait before enabling the *pre-commit* hooks, however, until it is determined if the linting can be made less nit-picky. The way it is now it is going to pollute minor commits will all sorts of "fixes". If we make a commit to fix an issue that only changes one or two lines, but the linter changes 100s of lines, it becomes impossible to tell what changes were actually related to the fix. If it is possible to adjust the lint settings so that it mostly accepts the existing code base without much complaint, then it can be turned on.

Issue #772, opened 2021-01-25
Web Content Accessibility Compliance


norinespears:
Request from members: 

The University of New England is evaluating the Specify application to replace our existing herbarium data management software and would appreciate if you could advise if version 7 is compliant to WCAG 2.0 or 2.1. 

Thanking you in advance,
Best regards,
Sylvia

beach53:
Please indicate in potential member requests like this which collections and institutions the petitioner represents.  Thanks!

beach53:
test

maxxxxxdlp:
I began going over WCAG 2.1 requirements and implementing them.
It is available on the branch `a11y`.

maxxxxxdlp:
Specify 7 Accessibility Statement - https://docs.google.com/document/d/1KXoNTUFQ1_MmyaJ_VRNTYXeyJc_lhcVCKiiVOOQ4x2o/edit#
An up-to-date list of accessibility issues - https://github.com/specify/specify7/projects/12

maxxxxxdlp:
Need to go over all of these and check that Specify is compliant:
- a11yproject:
   - [ ] https://www.a11yproject.com/checklist/#identify-links-that-open-in-a-new-tab-or-window
   - [ ] https://www.a11yproject.com/checklist/#use-fieldset-and-legend-elements-where-appropriate
   - [ ] https://www.a11yproject.com/checklist/#make-sure-that-error-warning-and-success-states-are-not-visually-communicated-by-just-color
   - [ ] https://www.a11yproject.com/checklist/#check-your-content-in-specialized-browsing-modes
   - [ ] https://www.a11yproject.com/checklist/#increase-text-size-to-200percent
   - [ ] https://www.a11yproject.com/checklist/#double-check-that-good-proximity-between-content-is-maintained
   - [ ] https://www.a11yproject.com/checklist/#make-sure-color-isnt-the-only-way-information-is-conveyed
   - [ ] https://www.a11yproject.com/checklist/#ensure-animations-are-subtle-and-do-not-flash-too-much
   - [ ] https://www.a11yproject.com/checklist/#make-sure-all-animation-obeys-the-prefers-reduced-motion-media-query
   - [ ] https://www.a11yproject.com/checklist/#check-the-contrast-for-all-normal-sized-text
   - [ ] https://www.a11yproject.com/checklist/#check-the-contrast-for-all-large-sized-text
   - [ ] https://www.a11yproject.com/checklist/#check-the-contrast-for-all-icons
[...]

PR #1012, opened 2021-10-11
Make Specify 7 WCAG 2.1 Compliant

maxxxxxdlp:
See #772


maxxxxxdlp:
Full list of issues that have to be tested as part of this Pull Request: https://github.com/specify/specify7/projects/12#column-16149473

specifyweb/context/views.py
```diff
@@ -602,3 +609,24 @@ def api_endpoints(request):
 def api_endpoints_all(request):
     """Returns a JSON description of all endpoints."""
     return JsonResponse(generate_openapi_for_endpoints(True))
+
+@require_GET
+@never_cache
+def set_language(request, language: str):
```
benanhalt, line 615: This should be a POST method and use a request parameter instead of a url path parameter. As a general rule, GET methods shouldn't "change" anything.

maxxxxxdlp: This particular endpoint wasn't used. I have just removed it (ac2037b5)
But I see what you are saying about GET not being supposed to make changes.

specifyweb/context/views.py
```diff
@@ -602,3 +610,19 @@ def api_endpoints(request):
 def api_endpoints_all(request):
     """Returns a JSON description of all endpoints."""
     return JsonResponse(generate_openapi_for_endpoints(True))
+
+@require_http_methods(['GET', 'POST'])
+def language(request):
+    """Get List of available languages OR set current language."""
+    if request.method == 'GET':
+        return JsonResponse({
+            'data': {
```
benanhalt, line 619: Why are you wrapping all of your JSON responses in `{"data": ...}`?

maxxxxxdlp: There is no reason for now, but I have seen some APIs respond like `{error: false, data: ...}` or `{error: 'Out of range': data: ...}`, which seems nicer (and more consistent), rather than delivering error messages in plain text.
Also, I think json:api standard does something similar.

benanhalt: IDK. For one thing, the error message is still plain text! So I don't see how it helps. And the type of the `error` parameter is `Union[str, Literal[False]]` which is funky. We have HTTP status codes for indicating that the request failed.

I think you are anticipating an `data Either = Ok Data | Failure Error` monad encoded in JSON, which is great. (See: https://en.wikipedia.org/wiki/Result_type) But, I wouldn't start including half implementations. If there is no reason for something now, then there is no reason for it. Since now is now, after all.

maxxxxxdlp: Fixed in d1e305f7

PR #1014, opened 2021-10-11
Disallow revoking your own admin status

### LINKED ISSUE #175: Can remove all admins
When logged in as an admin for a database, you can make any other user an admin, and you can remove admin status from any user - including yourself. With this capability, users could accidentally remove all admins and then no longer have the ability to manage users, which would require them to either restore the database from a backup or to have our support team go in and manually return the user's admin status.

maxxxxxdlp:
Fixes #175


grantfitzsimmons (approved):
No matter how hard I tried, I could not revoke my own administrator status.

Issue #1011, opened 2021-10-08
Create User screen should not offer save and add another


tlammer:
Currently, you have to put in the agent name then save the record but save and add another is an option.  But if you do that the agent you just created doesn't get created properly because it doesn't have a password or an agent assigned.

maxxxxxdlp:
Fixed in b2c6e051
The "Add Another" button is present, but the system explicitly asks user to set password and agents before proceeding

Issue #124, opened 2015-06-26
"Save and Add Another" error on next save


garrettgirard:
For permits, loans, and RAs (possibly other fields that have the "save and add another" ability, although there is a different issue with this for Gifts - see #123), selecting "save and add another" and then either attempting to save or save and add another causes an error.

This is probably because it lets you try to save even without changing the data in fields that are required to be unique (permit number, loan number, RA number). The expected action is that these fields are highlighted red with a warning message "value must be unique to division" if you select the "save" or "save and add another" options.

benanhalt:
Still present in accession. The underlying problem is the carry-forward mechanism that replaces the accession number field when it should be unique. Maybe carry forward should ignore unique fields that are not autonumbered.

garrettgirard:
This is also a problem when adding new users. If you create a user and do "Save and Add Another", saving will give you a "duplicate entry" error.

Error:
IntegrityError at /api/specify/specifyuser/
(1062, "Duplicate entry 'usetester' for key 'Name'")

PR #1015, opened 2021-10-11
Fix for "Users can only be added at manager level"

maxxxxxdlp:


grantfitzsimmons (dismissed):
KQAUGII4DL

grantfitzsimmons (changes requested):
Their is no longer a button for the username or notifications elements on this branch.
Instead of appearing like this:
look1
They are appearing like this:
look2

_In the interest of being user friendly, maybe a password, agents, and collection permissions should be required upon each user creation._

The Specify name field is carried over when clicking **Save and Add Another**, but if you attempt to add a duplicate user with the same name, this wall of text appears:
errorwhenidenticalnameisadded
This field should be cleared when adding another user and/or potentially a friendly error should appear.

maxxxxxdlp:
> Their is no longer a button for the username or notifications elements on this branch

Fixed. The issue occurred because this branch was out of date with production

> In the interest of being user friendly, maybe a password, agents, and collection permissions should be required upon each user creation.

That's a valid point and it has been brought up before. It is something we would consider when redesigning the user creation and permissions system

> The Specify name field is carried over when clicking Save and Add Another, but if you attempt to add a duplicate user with the same name, this wall of text appears:

That's a know issue too. There are two bugs open for that: https://github.com/specify/specify7/issues/1011 and https://github.com/specify/specify7/issues/124

maxxxxxdlp:
This is redundant now because the user type plugin has been rewritten

PR #1016, opened 2021-10-11
Show error message if user has no agent

### LINKED ISSUE #349: If a user agent was not found for a collection, the web GUI stalls at 'Specify is loading...'
Specify 7: v7.2.0-1

If you select a collection that your user does not have an user agent for, the web interface just says `Specify is loading...`. All GUI elements are gone and the only option is to hit the browser back button.

Suggestion: Warning message indicating that the user account does not have access to the collection with a link to go back to previous page.

maxxxxxdlp:
Fixes #349


specifyweb/context/views.py
```diff
@@ -258,6 +258,10 @@ def user(request):
     data = obj_to_data(request.specify_user)
     data['isauthenticated'] = request.user.is_authenticated
     data['available_collections'] = users_collections(connection.cursor(), request.specify_user.id)
+
+    if not hasattr(request.specify_user_agent, '_meta'):
```
benanhalt, line 262: I think this should be a simple `request.specify_user_agent is None` check.

specifyweb/context/views.py
```diff
@@ -258,6 +258,10 @@ def user(request):
...
+        return HttpResponseBadRequest('user does not have an agent')
+
     data['agent'] = obj_to_data(request.specify_user_agent)
```
benanhalt, line 265: Better yet just set `data['agent']` to None if `request.specify_user_agent` is None. Then the request can succeed.

specifyweb/frontend/js_src/lib/userinfo.js
```diff
@@ -2,12 +2,36 @@
...
+const commonText = require('./localization/common').default;
+const $ = require('jquery');
 
     var userInfo = {};
     initialContext.load('user.json', function(data) {
         _.extend(userInfo, data, {
-            isReadOnly:  !_(['Manager', 'FullAccess']).contains(data.usertype)
+            isReadOnly: !_(['Manager', 'FullAccess']).contains(data.usertype)
         });
+    }, (error)=>{
```
benanhalt, line 13: If you set the `agent` field to null as described above then you can check for that here instead of just assuming that BadRequest implies there was no agent.

specifyweb/context/views.py
maxxxxxdlp: This appears to be necessary as type(request.specify_user_agent) returns `<class 'django.utils.functional.SimpleLazyObject'>`

specifyweb/context/views.py
maxxxxxdlp: Fixed

specifyweb/context/views.py
```diff
@@ -258,10 +258,7 @@ def user(request):
     data = obj_to_data(request.specify_user)
     data['isauthenticated'] = request.user.is_authenticated
     data['available_collections'] = users_collections(connection.cursor(), request.specify_user.id)
-
-    data['agent'] = obj_to_data(request.specify_user_agent) \
-        if hasattr(request.specify_user_agent, '_meta') \
-        else None
+    data['agent'] = obj_to_data(request.specify_user_agent) if request.specify_user_agent != None else None
```
benanhalt, line 261: request.specify_user_agent is a Django SimpleLazyObject, but applying an equality operator will force it.

specifyweb/frontend/js_src/lib/userinfo.js
```diff
@@ -7,7 +7,7 @@ const $ = require('jquery');
 
     var userInfo = {};
     initialContext.load('user.json', function(data) {
-        if(typeof data.user === 'undefined'){
```
benanhalt: The attribute is data.agent not data.user. Also it will be null, not undefined.

benanhalt (approved):

specifyweb/frontend/js_src/lib/userinfo.js
maxxxxxdlp: Oh, I tested the case when the error was supposed to pop up, but did not test the opposite case 😃

PR #1017, opened 2021-10-11
Don't trip off unloadProtect on Query export to KML

### LINKED ISSUE #687: User is asked to save query changes when no changes where made
Similar to #570

Pressing the `Create KML` button makes it think that I made some changes. After that, generating a new record set or going to any other page would trigger this window

Screen Shot 2020-06-25 at 2 41 57 PM

maxxxxxdlp:
Fixes #687

A better fix as per https://github.com/specify/specify7/issues/687#issuecomment-881052932


grantfitzsimmons (approved):
unloadProtect is not triggered on my end when attempting to export to KML from a query.

PR #1020, opened 2021-10-11
Add double click support in MappingView

### LINKED ISSUE #895: Add ability to double click to add field to mapping
Currently, double clicking on a field in the field list does not add it to the mapping regardless of whether a corresponding column has been selected or not.
Ideally, if a column is selected, double clicking the field would add it to that column and if no column is selected, it would add it as a new column

maxxxxxdlp:
Fixes #895


grantfitzsimmons (approved):
This works as intended on my end. I am able to double click and map to each column successfully once they have been added.

I would like if when you double click without a column selected it would create a new column and map it. Maybe when you double click, it could automatically move to the next column after assigning it as well. @maxxxxxdlp

PR #1021, opened 2021-10-11
Send additional data when opening GeoLocate in WB

### LINKED ISSUE #931: Send additional data when opening GeoLocate in WB
Currently WB sends either Geography information or Locality information
GeoLocate supports sending both at the same time

maxxxxxdlp:
Fixes #931

@specify/ux-testing To test this, create a new data set with the following columns:
Latitude1, Longidute1, Latitude2, Longidute2, Lat Long Type, Lat Long Accuracy, Country Name, State Name, County Name.
Fill some data into all of these fields
Open GeoLocate
Make sure GeoLocate received data from all WB columns

### DIFF
--- a/specifyweb/frontend/js_src/lib/wblocalitydataextractor.ts
+++ b/specifyweb/frontend/js_src/lib/wblocalitydataextractor.ts
@@ -189,10 +189,12 @@ export const getLocalitiesDataFromSpreadsheet = (
         index: customRowNumbers[index] ?? index,
       }))
       .filter(({ locality }) => typeof locality !== 'boolean')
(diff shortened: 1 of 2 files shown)


grantfitzsimmons (commented):
image

The Lat/Long data is not appearing in GeoLocate

grantfitzsimmons (changes requested):
image
GEOLocate is not pulling the data from Latitude2 and Longitude2 on my end.

maxxxxxdlp:
It looks like GeoLocale's API does not support passing latitude2 and longtitude2 values

PR #1022, opened 2021-10-11
WB Front-end performance improvements

maxxxxxdlp:
This PR includes a lot of WB changes, but I tested it thoroughly (reversed rows, change column order and in that state went over all WB features and options) 
I also found some bugs in the process, which were fixed in the process of refactoring

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -26,7 +26,9 @@ server {
 
     # proxy everything else to specify 7
     location / {
-        client_max_body_size 100M;
+        client_max_body_size 400M;
+        client_body_buffer_size 400M;
+        client_body_timeout 120;
         resolver 127.0.0.11 valid=30s;
         set $backend "http://specify7:8000";
         proxy_pass $backend;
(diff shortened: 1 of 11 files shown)


maxxxxxdlp:
To test this:
1. Upload a Data Set in the Workbench
2. Move the data set columns to change their order (click on the column header and drag left/right)
3. Sort multiple columns
4. Now that columns are reordered and Data Set is sorted, do all of these and make sure everything works great
5. Search Cells
6. Replace all found instances
7. Open the Search Config (the gear icon next to the "Replacement Value" input box) and change the Replace mode to "Replace Next"
8. Put a different search query and run Search&Replace on the resulting cells one by one
9. Navigate through the modified cells left and right
10. Change the navigation direction from "Column first" to "Row first" (or vise-versa) in the Search config
11. Check cell navigation again
12. Open the "Tools" menu
12. Run GeoLocate on a few localities (must have locality name and geography country/state/county columns mapped in the data set)
13. Try Coordinate Conversion on all coordinate columns
14. Select some coordinate cells and try Coordinate conversion
15. Plot the localities on the map. Click on the pins on the map and make sure pop-up bubbles show locality information
16. Try to undo all of the actions in steps 6-14 by pressing Ctrl+Z multiple times until there "Modified Cells" counter says 0 again

grantfitzsimmons (approved):
After testing these changes with several data sets, the behavior seems to act exactly as expected.

maxxxxxdlp:
Merged into a11y

PR #1023, opened 2021-10-11
Send Occurrences data to the SpiceDig page

maxxxxxdlp:


maxxxxxdlp:
To test this:
1. Use `freshfish` database on the server
1. Open "KU Fish Voucher" Collection
1. Open the "k" record set
1. Click on the "Specify Network" logo on the form
1. The webpage should show information for this collection object and this species from different aggregators
1. There should also be a map at the bottom of the page with clustered pins.
1. Clicking on a cluster of pins should zoom in the map.
1. Clicking on an individual pin should load information for that occurrence.
1. There should also be a "View Record" button that opens that occurrence in the form view in a new tab.
1. Go back to Specify, find some taxon record of the "species" rank (e.g "Ameiurus melas") and open it in form view.
1. Repeat steps 4-9. The Specify Network page should behave identically, except there won't be a table at the top with occurrence info.

grantfitzsimmons (changes requested):
_I could not find the record set "k" in the freshfish database. This is all done with the "oVert" record set._

In Safari, the Specify Network map is missing most of the functionality in Edge. It does not have the pins and links Chromium based browsers have access to. Edge let me click on individual pins and load information for each occurrence. The view record brought up the new tab no problem.

**Safari 15.1 Screenshot**
image

On Edge and Safari I navigated to the taxon record of "Ameiurus melas", then clicked Specify Network. This is the error I was shown:
screenshot
"Amia calva", another "species" rank taxon presented the same error. 

Returning to that record set and clicking the Specify Network link returned this view:
Screen Shot 2021-12-02 at 3 24 27 PM

maxxxxxdlp:
The "Invalid request URL" occurs when Specify Network link is located in a Taxon subform
Fixed by removing Specify Network link from subforms.

The issue with markers not working in Safari is a Leaflet bug (https://stackoverflow.com/questions/65369083/popup-does-not-open-when-clicking-on-marker-safari/65369228). Leaflet should release a new version in a few weeks with a fix for that.

maxxxxxdlp:
Merged into a11y

PR #1045, opened 2021-10-28
Uploader memory optimization

benanhalt:
This changes the dataset schema and requires a `python manage.py migrate workbench` to work. This schema change cannot be backed out.

The basic ideas is to store the row upload results in a separate table from the dataset. These are inserted as the upload progresses so that the entire result set doesn't need to be stored in memory until the end of the upload.


maxxxxxdlp:
Is this ready to be tested?

benanhalt:
Not yet. I want to look over it again. Also it introduces schema changes in the Specify 7 managed tables (migrations). And I need to figure out how to manage that in terms of tests. For example if you use a database with this branch, the workbench won't work on other branches afterwards because the migrations will have been applied.

maxxxxxdlp:
Another possible optimization:

In the rowresults structure, use column indexes instead of column names. That will reduce the size by a lot:

Screen Shot 2022-07-05 at 16 29 37

PR #1051, opened 2021-11-05
Fix Query Builder using Tree rank names instead of titles

### LINKED ISSUE #1050: Query does not show tree rank titles
Barcelona observed that the query was showing the rank name instead of the title, which they had changed to their own language.
this is what the query looks like:
image
This is what the titles look like in their definitions:
image

maxxxxxdlp:
Closes #1050

### DIFF
--- a/specifyweb/frontend/js_src/lib/queryfield.js
+++ b/specifyweb/frontend/js_src/lib/queryfield.js
@@ -205,7 +205,7 @@ module.exports =  Backbone.View.extend({
             });
         },
         addTreeLevelsToFieldSelect: function(getTreeDef) {
-            var optGroup = $('<optgroup label="Tree Ranks">').appendTo( this.$('.field-select') );
+            var optGroup = $(`<optgroup label="${queryText('treeRanks')}">`).appendTo( this.$('.field-select') );
 
             getTreeDef.pipe(function(treeDef) {
                 return treeDef.rget('treedefitems');
@@ -214,11 +214,11 @@ module.exports =  Backbone.View.extend({
             }).done(function(treeDefItems) {
                 treeDefItems.each(function(item) {
                     $('<option>', {value: 'treerank-' + item.get('name')})
-                        .text(item.get('name'))
+                        .text(item.get('title') ?? item.get('name'))
                         .appendTo(optGroup);
                     if (item.specifyModel.name == 'TaxonTreeDefItem') {
                         $('<option>', {value: 'treerank-' + item.get('name') + ' Author'})
-                            .text(queryText('treeRankAuthor')(item.get('name')))
+                            .text(queryText('treeRankAuthor')(item.get('title') ?? item.get('name')))
                             .appendTo(optGroup);
                     }
                 });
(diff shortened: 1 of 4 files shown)


grantfitzsimmons (dismissed):
On the picklist, the tree rank shows the title. After that is selected, the query shows the name of the rank.

maxxxxxdlp:
@benanhalt the issue @grantfitzsimmons found occurs at this line: 
https://github.com/specify/specify7/blob/233b8931faf3d12bcccc45f4bc35db294706af2f/specifyweb/frontend/js_src/lib/queryfield.js#L161

Field name get's printed, rather than the field title

I am not sure where should I get treeDefItems from in this scope. Do you store that information anywhere or should I fetch it here?

benanhalt:
@maxxxxxdlp I think you will just have to fetch them for now. Some day the tree definitions should be set up to just load at start up. But that is a broader scope fix than this issue.

Also, the tree level names are used in the header for the results, so that probably needs to be changed to use the titles as well.

benanhalt (commented):
See the comment about fetching tree levels.

maxxxxxdlp:
Fixed in 6595f9c2
The code is looking ugly with all that in-place rank fetching.
Would need to do a comprehensive refactor at some point.

specifyweb/frontend/js_src/lib/queryfield.js
```diff
@@ -11,6 +11,10 @@ var QueryFieldSpec    = require('./queryfieldspec.js');
 var QueryFieldInputUI = require('./queryfieldinput.js');
 const queryText = require('./localization/query').default;
 const commonText = require('./localization/common').default;
+const dataModelStorage = require('./wbplanviewmodel').default;
```
benanhalt, line 14: I'd rather you not use Workbench code here. 

Use code similar to https://github.com/specify/specify7/blob/e369d1f81a4234d74998ca5c852f37c66d4869e0/specifyweb/frontend/js_src/lib/queryfield.js#L232-L245

specifyweb/frontend/js_src/lib/queryresultstable.js
```diff
@@ -10,11 +10,29 @@ var ScrollResults = require('./scrollresults.js');
 var QueryResults  = require('./queryresults.js');
 const queryText = require('./localization/query').default;
 const commonText = require('./localization/common').default;
+const dataModelStorage = require('./wbplanviewmodel').default;
```
benanhalt, line 13: Also here. Use `domian.getTreeDef` instead. That way when that gets refactored, this will get fixed.

benanhalt (changes requested):

specifyweb/frontend/js_src/lib/queryfield.js
maxxxxxdlp: > Some day the tree definitions should be set up to just load at start up. But that is a broader scope fix than this issue.

dataModelStorage serves exactly that purpose. It stores the names and titles of all tree ranks

> I'd rather you not use Workbench code here.

wbplanviewmodel does not depend on other workbench files. It might be appropriate to rename it. Also, it is already used outside of the workbench (Leaflet and S^N)

And unlike schema.models, dataModelStorage has type definitions

Do you still want me to revert to fetching tree definitions here?

benanhalt: Yes, please. The whole data model is not needed in this context, just the tree definition.

benanhalt (approved):

PR #1066, opened 2021-11-12
ES Modules

maxxxxxdlp:
Update to babel 7 and core-js 3
Update other development dependencies
Improve webpack config
Improve source maps
Use ES Modules instead of common js
Use import() instead of deprecated require.ensure()


maxxxxxdlp:
@benanhalt Development dependencies have been updated
Production dependencies remain unchanged, so this PR should be more or less stable

benanhalt:
What prompted you to make these changes?

maxxxxxdlp:
> What prompted you to make these changes?

This PR fixes a few dependabot alerts, speeds up webpack, updates to the newest import syntax, makes codebase more consistent and makes it easier for my IDE to infer types of imported modules

Though, it looks like back-end localization is not yet merged into production. Would need to do that first

benanhalt:
Do these changes depend on the backend localization?

maxxxxxdlp:
It didn't initially, but I rebased it of `next-release`, which has backend-localization merged into it

benanhalt:
I can separate the commits from the next-release branch unless it actually depends on them. If they are independent changes I would prefer to do that.

maxxxxxdlp:
Sure, go ahead. They are indepentent.

These are the commits you need to keep:
Screen Shot 2021-11-19 at 4 14 05 PM

Everything before this one can be dropped:
```
2481ea20 Max Patiiuk 4 months ago Replace deprecated code splitting method
```

benanhalt:
It seems like webpack is not generating the `manifest.py` file on this branch.

maxxxxxdlp:
I see the issue
I told webpack to clean previous build artifacts (to prevent disk space leaks). Apparently, it was also deleting manifest.py in the process
Should be fixed now

maxxxxxdlp:
Merged into a11y

PR #1076, opened 2021-11-22
fix field mismatch when report query fields are set to not display.

### LINKED ISSUE #1075: Some labels that generate correctly from the reports dialog fail when invoked with the "Generate Label" button on the form.
**Describe the bug**
Some labels that generate correctly from the reports dialog fail when invoked with the "Generate Label" button on the form. The bug is triggered by the label query having a non-displaying field in some position other than last.

**To Reproduce**
1. Open the query for a working label in the query builder.
2. Add an arbitrary field to the query and move it near the top of the query.
3. Set the new field to not display (by clicking the lightbulb).
4. Save the query.
5. Verify that running the label from the Reports menu still generates the correct results.
6. See that running the label by clicking "Generate Label" on a form either causes an exception or puts the data into the wrong fields in the label.

**Reported By**
Yugra

benanhalt:
fixes #1075

### DIFF
--- a/specifyweb/frontend/js_src/lib/reports.js
+++ b/specifyweb/frontend/js_src/lib/reports.js
@@ -278,6 +278,9 @@ function getReportParams(reportResources) {
 }
 
 function runReportForRecord(reportResources) {
+    // for consistency's sake since we set all fields to isdisplay in QueryParamsDialog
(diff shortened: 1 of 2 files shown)


grantfitzsimmons (approved):
I can no longer replicate this bug on the issue-1075 branch.

PR #1081, opened 2021-11-25
If no fields are visible, run Query in CountOnly mode

### LINKED ISSUE #869: Query pagination does not work if no fields are made visible
If "Show in results" is unset for all the fields in the Query Builder, the code that is responsible for fetching only visible results fails and sends request after request instead until all rows are fetched

maxxxxxdlp:
Fixes #869


grantfitzsimmons (approved):
The output is the same as checking Count. Works!

maxxxxxdlp:
The fix is now out of date since query builder UI is being rewritten
The new fix was made in a13ef1d4

PR #1083, opened 2021-11-25
Don't let title enlarge Dialog window

### LINKED ISSUE #703: Long titles no longer enlarge to full width of dialog window
In some cases we have very long collecting event codes and these are concatenated to the full locality string on the event form modal when accessed from the collecting event form. The modal resizes it's width to fit the full string, sometimes pushing the edge of the modal off the right hand side of the screen. Please fix so that the modal does not resize to the width of the text. Please see image below

image

maxxxxxdlp:
Fixes #703

Since resource view dialog has width: auto, long titles can increase the
width of the dialog. A somewhat ugly workaround is to record the dialog
width before changing the title, and setting title's max-width to
previous dialog width.

### DIFF
--- a/specifyweb/frontend/js_src/lib/querycbx.js
+++ b/specifyweb/frontend/js_src/lib/querycbx.js
@@ -525,7 +525,14 @@ var QueryCbx = Backbone.View.extend({
         this.fillIn();
     },
     changeDialogTitle: function(resource, title) {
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (approved):

maxxxxxdlp:
This is redundant now since the new dialog system does not have this problem

PR #1085, opened 2021-11-25
Fix redirect loop on "Choose Collection" page

### LINKED ISSUE #870: "Choose collection" page redirects to itself
If the user types the password incorrectly several times, and after finally typing the correct value, the "Choose collection" page would have a ?next_page= get parameter that references itself.

This means the user is being asked to "Choose collection" as many times as there were invalid login attempts.

To Test this:
1. Use a database that has multiple collections
1. Type the password incorrectly several times
2. Type the password correctly
3. Chose the collection
4. You should be asked to choose the collection only once, at which point you should see the welcome page

maxxxxxdlp:
Fixes #870

### DIFF
--- a/specifyweb/frontend/templates/login.html
+++ b/specifyweb/frontend/templates/login.html
@@ -35,7 +35,14 @@
         {{ form.password.errors }}
         {{ form.password }}
         <input type="hidden" name="this_is_the_login_form" value="1" />
-        <input type="hidden" name="next" value="/accounts/choose_collection/?next={{ next|urlencode }}" />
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (changes requested):
image
There is now automatically a ?next=/specify/ at the end of the URL. If this is intentional, ignore this request. A /specify/ text appears below the log in. If these are planned behaviors, merge!

maxxxxxdlp:
The "?next_page" should appear in the URL, but only once, even if you entered incorrect password multiple times.
The `/specify/` text should not be appearing there, though that should be fixed now

PR #1086, opened 2021-11-25
Sort list of collections alphabetically

### LINKED ISSUE #741: Choose collection should be ordered
We have quite many collections and personnel are often switching between them. As it is now, the choose collection list is not ordered (or ordered by collectionID). It would be nice if the list was alphabetically sorted instead.

image

maxxxxxdlp:
Fixes #741


grantfitzsimmons (approved):
They're alphabetical. What can I say!

PR #1087, opened 2021-11-25
Add ability to delete a record from Record Set

### LINKED ISSUE #839: Remove records from record sets without deleting in database
Currently, there is no way to remove a record from a record set without deleting it out of the database entirely.

maxxxxxdlp:
Fixes #839

### DIFF
--- a/specifyweb/specify/api.py
+++ b/specifyweb/specify/api.py
@@ -115,21 +115,7 @@ def resource_dispatch(request, model, id):
                             content_type='application/json')
 
     elif request.method == 'DELETE':
-        recordsetid = request.GET.get('recordsetid', None)
-
-        if recordsetid is None:
-            delete_resource(request.specify_user_agent, model, id, version)
-        else:
-            obj = get_object_or_404(model, id=int(id))
-            try:
-                delete_from_record_set(
-                    obj,
-                    version,
-                    recordsetid,
-                )
-            except (FilterError, OrderByError) as e:
-                return HttpResponseBadRequest(e)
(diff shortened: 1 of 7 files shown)


specifyweb/specify/api.py
```diff
@@ -115,7 +115,20 @@ def resource_dispatch(request, model, id):
                             content_type='application/json')
 
     elif request.method == 'DELETE':
-        delete_resource(request.specify_user_agent, model, id, version)
+        recordsetid = request.GET.get('recordsetid', None)
```
benanhalt, line 118: Did you test this? I don't think `request.GET` should contain the request parameters if the  method is DELETE.

In any case, I don't think the REST verb DELETE on the resource should be overloaded for record-set-item deletion. It should be done by issuing DELETE on the record-set-item resource: `DELETE /api/specify/recordsetitem/id/`. That would require knowing the record-set-item id. For that, recommend extending the `record_set_info` metadata that gets added to the main resource api to include the record-set-item id.

benanhalt (changes requested):

specifyweb/specify/api.py
benanhalt: This is actually a bug in the audit log code because RecordSetItem doesn't have a version field. It looks like `recordversion` in the audit log cannot be null, so perhaps it should be record as zero if `obj` doesn't have a version attribute. OTOH, should changes to record sets even be recorded in the audit log?

maxxxxxdlp:
Superseeded by 691cc00f

PR #1088, opened 2021-11-25
Make Forms and Interactions dialogs open faster

### LINKED ISSUE #593: User can open `Interactions` and `Forms` windows at the same time
Screenshot from 2020-01-21 07-13-14

Even though this should not be possible, user is able to open both `Interactions` and `Forms` selection windows at the same time

maxxxxxdlp:
Fixes #593


grantfitzsimmons (approved):
image
I could not recreate the initial bug. I can open plenty of windows if I remove the "ui-widget-overlay" blocking the user but this fix solves any natural issues that could come up.

PR #1091, opened 2021-11-25
Display helpful message when attachments server not available

### LINKED ISSUE #837: Attempting to add an attachment to a record when no attachment server is configured causes a server error.

maxxxxxdlp:
Fixes #837

### DIFF
--- a/specifyweb/frontend/js_src/lib/attachmentplugin.js
+++ b/specifyweb/frontend/js_src/lib/attachmentplugin.js
@@ -18,9 +18,10 @@ module.exports =  UIPlugin.extend({
         },
         render: function() {
             var self = this;
-            if (!attachments) {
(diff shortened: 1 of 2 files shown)


grantfitzsimmons (changes requested):
image
This message appears in every data entry attachment window I create.

You can continue to raise the count of attachments by pressing new without an attachment server available. These do disappear when saving the collection object, but creating new attachments should probably be disabled.

maxxxxxdlp:
@benanhalt I might need your help here to come up with the best solution.
Let's say I add an `isPluginAvailable` method to the Attachment UIPlugin view
How can I figure out that the plugin is not available from inside of `recordselector.js`?

maxxxxxdlp:
Cherry-picked onto a11y in 1f09eb63

PR #1093, opened 2021-11-25
'Division' field name and picklist items missing

### LINKED ISSUE #735: 'Division' field name and picklist items missing
The division field caption and information in accession form doesn't show up
Screenshot (134)

maxxxxxdlp:
Fixes #735

### DIFF
--- a/specifyweb/frontend/js_src/lib/divisionfieldcbx.js
+++ b/specifyweb/frontend/js_src/lib/divisionfieldcbx.js
@@ -1,0 +1,51 @@
+"use strict";
+
+var $        = require('jquery');
+var _        = require('underscore');
+var Backbone = require('./backbone.js');
+
+var schema = require('./schema.js');
+
+module.exports =  Backbone.View.extend({
+        __name__: "DivisionPicklistCBX",
+        events: {
+            change: 'set'
+        },
+        initialize: function(info) {
+            this.resource = info.resource;
+            this.resource.on('change:fieldname change:tablename change:type', this.render, this);
+            const divisionQuery = new schema.models.Division.LazyCollection();
+            this.divisionsPromise = new Promise((resolve)=>
+                divisionQuery.fetch().then(()=>resolve(divisionQuery.models))
+            );
+            this.divisions = [];
+        },
+        getPickListFields: function() {
+            return this.divisions;
+        },
+        render: function() {
+            if(typeof this.divisionsPromise === 'object'){
+                this.divisionsPromise.then((divisions)=>{
+                    this.divisions = divisions.map((division)=>({
+                        value: division.get('id'),
+                        title: division.get('name')
+                    }));
+                    this.divisionsPromise = undefined;
+                    this.render();
+                });
+                return this;
+            }
+
+            var options = this.getPickListFields().map(function(item) {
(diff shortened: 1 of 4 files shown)


grantfitzsimmons (changes requested):
accessionlist
This is the 2021_11_25_20_18 database on the test panel. The accession form still is missing the division name. Tested with Specify7 branch issue-735.

grantfitzsimmons (changes requested):
Bad news. With this branch, you cannot use the Data Entry feature. It does not respond
badbad
I tested this with freshfish and 2021_11_25_20.

On the bright side, the picklist items and Division field name are present.
accession

specifyweb/frontend/js_src/lib/divisionfieldcbx.js
```diff
@@ -0,0 +1,51 @@
...
+            return this.divisions;
+        },
+        render: function() {
+            if(typeof this.divisionsPromise === 'object'){
+                this.divisionsPromise.then((divisions)=>{
+                    this.divisions = divisions.map((division)=>({
+                        value: division.get('id'),
+                        title: division.get('name')
+                    }));
+                    this.divisionsPromise = undefined;
```
benanhalt: Don't do this! One of the main features of promises is that invoking `then` on an already resolved promise will invoke the continuation immediately so you don't have to keep track of the state manually!

specifyweb/frontend/js_src/lib/divisionfieldcbx.js
```diff
@@ -0,0 +1,51 @@
...
+                    this.render();
+                });
+                return this;
+            }
+
+            var options = this.getPickListFields().map(function(item) {
+                return $('<option>').attr('value', item.value).text(item.title)[0];
+            });
+            this.$el.empty().append(options).prop('disabled', options.length < 1);
+            this.$el.val(this.resource.get('fieldname'));
```
benanhalt: `'fieldname'`? How is that supposed to work?

benanhalt:
@grantfitzsimmons, would you please test this with a database that has more than one division? Make sure that the division can be set, saved, reloaded with the correct value appearing, and then changed to a new value, saved and reloaded.

specifyweb/frontend/js_src/lib/divisionfieldcbx.js
maxxxxxdlp: Oh my god, why didn't I know about this sooner!
I always though that this code is invalid:
```js
const a = Promise.resolve();
a.then(console.log);
a.then(console.log);  // Error?
```

benanhalt: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#:~:text=A%20promise%20can%20participate%20in%20more%20than%20one%20nesting

grantfitzsimmons (changes requested):
**The data entry forms no longer work on this branch!**

After testing this with multiple divisions, the division can be set, saved, reloaded multiple times and preserve its original value. This issue appears to be solved.

benanhalt:
The data entry forms not working is a separate issue that has been fixed on another branch.

PR #1094, opened 2021-11-25
Check unloadProtect stack before changing collection

maxxxxxdlp:


grantfitzsimmons (approved):
Works exactly as expected.

PR #1095, opened 2021-11-25
Increase clickable area for query results

### LINKED ISSUE #913: Empty cells in query results are not clickable
If the query result contains an empty cell, that cell is not clickable and thus the record can not be viewed

This is because only the text content of the query result is clickable
As a fix, the whole cell area should be clickable, not just the text content

maxxxxxdlp:
Fixes #913

### DIFF
--- a/specifyweb/frontend/js_src/css/queryresults.css
+++ b/specifyweb/frontend/js_src/css/queryresults.css
@@ -34,11 +34,10 @@
 
 .query-results td {
   border-right: 1px solid #aaa;
-  padding: 0;
+  padding: 5px;
 }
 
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (approved):
The entire query result is clickable as intended. Do users often need to copy data or select text from this fields? If there is a way to make the text selectable while retaining this behavior that would be a good addition.

maxxxxxdlp:
Oh right, I haven't thought about that.
Fixed. It should be easier to copy&paste the text now
You can also right-click on the link and select "Copy" or "Copy link text" from the context menu

PR #1104, opened 2021-12-01
Add way to add another record after record has been saved

### LINKED ISSUE #544: Add Records from DB to an existing Record Set
Requested by MN dragonfly +many
Dragonfly noted that after they have saved a record, the only way to add a new record is to go to data, then select the table from the list.  They asked if there could be a button allowing them to add a new record from the record they are currently on.  In 6, you have a green plus at the top of the window to add a new record.

maxxxxxdlp:
Fixes #544


maxxxxxdlp:
This branch was replaced by the cbda32de commit on a11y

PR #1105, opened 2021-12-02
fix regression introduced by pr #1088.

benanhalt:
You can't use `this.forms` because every time the dialog is invoked it is a new instance... But, you already have all the values in the `getFormsPromise` which is captured in a closure. So just use that!

Please test by making sure entries in both Data Entry and Interactions dialogs can be invoked multiple times in succession.


maxxxxxdlp (approved):
This all makes sense.

Though, it seems like `.isFulfilled()` is not available for native promises. For native we would need to either do it in an async way though `.then()` or wrap the promise in a closure

PR #1107, opened 2021-12-03
Make web asset server work in docker

maxxxxxdlp:
Back-end replaces the origin of the URLs returned from the web server
by the origin taken from ASSET_SERVER_URL

This makes web asset server work both on the front-end and the back-end

For environments outside of docker, if ASSET_SERVER_URL and web asset
server point to the same origin, this change would have no effect


maxxxxxdlp:
This was superseded by https://github.com/specify/specify7/pull/1116

Issue #1025, opened 2021-10-13
Should picklist size even be enforced?


benanhalt:
**Is your feature request related to a problem? Please describe.**
Currently there is a business rule that rejects additions to picklists that would cause them to exceed their size limit parameter.
Because of #108 that just results in an exception.

**Describe the solution you'd like**
Does anyone really care about size limits on picklists? Does Sp6 enforce it even? Maybe this is an antifeature.

**Describe alternatives you've considered**
Just fixing #108 would at least make it clear what is happening.

**Additional context**

maxxxxxdlp:
@benanhalt Sp6 had a use case for size limit.
Sp7 treats size limit differently. Thus, the field should either be set to follow sp6 behavior, or sp7 should ignore the field altogether

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/specify-7-7-3-release-announcement/774/1

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/specify-7-7-4-release-announcement/807/1

Issue #108, opened 2015-06-23
Backend business rules exceptions are not  handled well in front end


timo11:

benanhalt:
See if these are showing in the exception catcher.

maxxxxxdlp:
Currently, the back-end throws a 500 error invalid business rule.

Instead, it should return a 400 error with a payload object that specifies the offending field if applicable.

The payload object should have an error code, which is then resolved to an error message by the front-end for purposes of localization.

grantfitzsimmons:
**For pick lists, ** I propose maintaining the uniqueness constraint, however it should mitigate the error (which loses all progress) by appending a (2) or (3) after the name (or at least prompt the user to enter a new name).

maxxxxxdlp:
Looks like @benanhalt began this in https://github.com/specify/specify7/commit/8f92c915637c581fff5a1e9bc7d600a89f7a4865

Still need to modify the response code to return a 4xx instead of 5xx, as 5xx would return an HTML response instead of JSON if the back-end is running in development mode

benanhalt:
Inventory the exceptions.
Can recoverable ones be identified?
Which ones should be prevented by FE?

maxxxxxdlp:
I was thinking Front-End would just catch the error and display it next to the field that triggered the issue.
The error would disappear on any edits to that field, and user would be able to press "Save" again to check for business rule errors again.

Thus, in the short term, front-end won't be preventing any errors, but it would display them nicely (with localized context aware error messages)

maxxxxxdlp:
@benanhalt does this sound good to you?
If so, just modify the response code for business rule exceptions from 5xx to 4xx and return a JSON response

benanhalt:
I think that is the basic idea, but it will be somewhat more involved. For instance I think there are instances where the exception can't be traced to one particular field.

maxxxxxdlp:
In that cases a dialog message would be displayed. The front-end already does it for save blockers:
All save blockers that can be attributed to an HTML field (not subviews), is going to be shown in that field.
If all remaining save blockers are not associated with any field, a dialog message would be displayed

maxxxxxdlp:
@benanhalt Another reason to modify the response code:
BussinessRule exceptions don't show the error message when in production (since they return a 500 error, and 500 errors are hidden when not in debug mode).
See https://github.com/specify/specify7/issues/2197#issue-1381099164 for an example.

I feel like this is a really terrible situation and should be fixed soon.

[...]

PR #1115, opened 2021-12-13
Schema editor

maxxxxxdlp:
Schema editor is available in the User Tools menu


grantfitzsimmons:
schema_6
schema_7
There are some big schema configurator inconsistences between 6 and 7. I highlighted tables that appear in each version but not in the other. Is this intentional behavior?
_freshfish database_

grantfitzsimmons (commented):
Adding a language to a database, making changes, and saving has no effect. Navigating to Schema Config does not show the added language.

grantfitzsimmons:
The Schema Configurator is missing the ability to create a new table format or table aggregation. These functions are present in Specify 6.

grantfitzsimmons:
No Seperation between user and system picklists
When selecting a Pick List, there is no separation between system and user-defined ones. There is no easy way to create a new Pick List without needing to navigate back through the Schema Configuration tool after saving, nor the ability to import or export one.
picklist6

grantfitzsimmons:
prepType_7
prepType_6

The Pick List editor shows pick lists that are inaccessible in 6 and non-functional in 7. The size limit being -1 is invalid when trying to save and continue configuration, and if you try to add pick list items, this error occurs:

picklistlimit
_Screenshot from a different picklist_

maxxxxxdlp:
@grantfitzsimmons 
1. > There are some big schema configurator inconsistences between 6 and 7

   Removed some system tables (treedef, treedefitem)
   Some of the remaining tables should be hidden in Sp6

2. > Adding a language to a database, making changes, and saving has no effect. Navigating to Schema Config does not show the added language. 

    Fixed Schema Config not fetching all languages properly

3. > The Schema Configurator is missing the ability to create a new table format or table aggregation

   > nor the ability to import or export one.

    Maybe for next release

4. > In 6, the Table Format defaults to the correct format.
  The Field Format pick list is restricted to None.

    Click on that pick list and there would be an option to change it's value to something other than none

5. > The Field Format options allow for the isLoanable field to be used as a web link, pick list, or in another format

    Also, from Andy:

    > I did notice that you can set formatted, weblink or pick list fields for non-text fields like integer fields.

    Fixed. Only text fields can be weblinks. Though, it's debatable whether numeric fields or yes/no files are to be allowed to be set as picklists

6. > When selecting a Pick List, there is no separation between system and user-defined ones
   
    Fixed

7. > There is no easy way to create a new Pick List without needing to navigate back through the Schema Configuration tool after saving

   You can open the link in the new tab. Keep in mind though, that newly added picklists won't appear in the list until the page is refreshed.

8.  > The size limit being -1 is invalid when trying to save and continue configuration,

    That's an issue with form definition. See https://github.com/specify/specify7/issues/1025

9. > ... and if you try to add pick list items, this error occurs:
 
    Known Issue. See https://github.com/specify/specify7/issues/108

maxxxxxdlp:
Merged into a11y

PR #1201, opened 2022-03-22
do self joins in queries backwards.

benanhalt:
This is to work around a possible bug in SQLAlchemy.


benritchie:
@benanhalt  - if it would be helpful, I'd happily test this for you.  could you patch onto RBGE Hosted specify if so?

Thanks

Ben

benritchie:
It has indeed resolved the issue.  
I've tested for a couple of cases:
Geo's - both the parent, and child links.  
Accepted taxon.  
Both look good.  
Thanks!

PR #1219, opened 2022-04-06
add basic loan prep return from query workflow.

benanhalt:


maxxxxxdlp:
Cherry-picked onto a11y and merged.
See #1676

PR #1286, opened 2022-04-19
Implement authentication via request.META variable.

benanhalt:
This PR adds the `AUTH_USING_META_VAR` setting to the Specify 7 server with the following operation:

> Configure Specify to authenticate by matching a Django request META
variable to a Specify user attribute. This is to support
authentication via Apache mod_shib which exposes SP session values
as META variables in requests. This setting should be a pair
consisting of the META variable and the Specify user attribute to
match. For example,

> AUTH_USING_META_VAR = ('REMOTE_USER', 'name')

> will login the the user with the user name given by the REMOTE_USER
META variable. This authentication occurs automatically when the
client GETs the /account/login/ URL which is the default login page
presented to the user unless OpenId Connect is configured. If the
META variable is not set or no matching user exists, the usual login
page will be returned. When using this form of authentication,
permission to modify the Specify user table must be restricted to
prevent users from compromising authentication by changing the
values in matching field of the user table. Set this option to None
to disable this form of authentication.

### DIFF
--- a/specifyweb/accounts/urls.py
+++ b/specifyweb/accounts/urls.py
@@ -9,11 +9,11 @@ urlpatterns = [
     path('login/',
          views.oic_login
          if settings.OAUTH_LOGIN_PROVIDERS
-         else auth_views.LoginView.as_view(template_name='login.html')
+         else views.legacy_login
          ),
 
     # Login with Specify username and password:
-    path('legacy_login/', auth_views.LoginView.as_view(template_name='login.html')),
+    path('legacy_login/', views.legacy_login),
 
     # OpenId Connect callback endpoint:
     path('oic_callback/', views.oic_callback),
--- a/specifyweb/accounts/views.py
+++ b/specifyweb/accounts/views.py
@@ -18,6 +18,7 @@ from django.utils import crypto
 from django.utils.http import is_safe_url, urlencode
 from django.views.decorators.cache import cache_control, never_cache
 from django.views.decorators.http import require_GET, require_http_methods
+from django.contrib.auth import views as auth_views
 from typing import Union, Optional, Dict
 from typing_extensions import TypedDict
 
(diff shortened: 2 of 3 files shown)


benanhalt:
FYI, Django includes OS environment variables in the request.META so it is possible to test this to a certain extent by doing
`REMOTE_USER=testuser python manage.py runserver`
or similar and then accessing /accounts/login/.

mcruz-umich:
Thanks for the update Ben!
I have two questions:

1. I understand Django is being used for the TDD here. Can you confirm though that the end-result will not be Django-dependent?
2. Would it be possible to implement something such that if SSO is being used, but authentication succeeds at the IdP but no existing matching Specify account is found, that instead of directing the user to the internal Specify Login page, they could instead be sent to a 404 page. A 404 vs a not authorized page would potentially be better from a security standpoint to give less info to a hacker. Your thoughts on if this sounds like a good idea?

mcruz-umich:
I am sorry but I do not fully understand this comment. Could you provide more information please?
# When using this form of authentication,
# permission to modify the Specify user table must be restricted to
# prevent users from compromising authentication by changing the
# values in matching field of the user table.

benanhalt:
> Would it be possible to implement something such that if SSO is being used, but authentication succeeds at the IdP but no existing matching Specify account is found, that instead of directing the user to the internal Specify Login page, they could instead be sent to a 404 page. A 404 vs a not authorized page would potentially be better from a security standpoint to give less info to a hacker. Your thoughts on if this sounds like a good idea?

I think returning a 404 in this situation would be a poor trade-off. While it would be potentially misleading to a hypothetical attacker, it is also misleading to legit users and admins trying to troubleshoot e.g. a misconfiguration. The situation can only arise if, as you say, authentication has already happened at the IdP. That means either an attacker would have to either be a member of the organisation greatly limiting the potential, or the IdP is comprised in which case all bets are off anyways.

benanhalt:
My caveat about the user table is just to note that changing values in the user table could potentially affect authentication when using this method. E.g. if the META variable is being matched to users' email addresses and a user has access to change email addresses in the user table, then they can change who gets logged in as who in Specify.

mcruz-umich:
Another use case - Will admins have a way to "impersonate" other users to verify their access and functional levels?
In that case, suppose I use SSO to sign in... I would have to sign in via Umich SSO and then once in, I would want to tell S7 to let me impersonate a user.

For my own development purposes, I would want to log in as my test users at guest, full-access, manager levels I can have the IAM team set up test accounts in our SSO system that I can use to sign in via SSO. But that is for development, not for troubleshooting a legitimate user's account.

Your thoughts?

mcruz-umich:
> My caveat about the user table is just to note that changing values in the user table could potentially affect authentication when using this method. E.g. if the META variable is being matched to users' email addresses and a user has access to change email addresses in the user table, then they can change who gets logged in as who in Specify.

Ok, that makes sense.

benanhalt:
> Another use case - Will admins have a way to "impersonate" other users to verify their access and functional levels? In that case, suppose I use SSO to sign in... I would have to sign in via Umich SSO and then once in, I would want to tell S7 to let me impersonate a user.
> 
> For my own development purposes, I would want to log in as my test users at guest, full-access, manager levels I can have the IAM team set up test accounts in our SSO system that I can use to sign in via SSO. But that is for development, not for troubleshooting a legitimate user's account.
> 
> Your thoughts?

There is not currently any UI mechanism for doing this through the web interface. It is possible to use the "support login" mechanism if you have CLI access to the server. If you turn on the configuration setting at 

https://github.com/specify/specify7/blob/production/specifyweb/settings/specify_settings.py#L54-L56

you can run a command `python manage.py support_login --username whoever` and it will return a URL path that will log your browser session in as the indicated user.

PR #1868, opened 2022-07-20
check length of picklist values when uploading. fixes #1837

### LINKED ISSUE #1837: WB does not handle errors arising from pick list values that exceed field length
Having a value that is over the field length limit produces this not user-friendly error:

image

See also #1365

benanhalt:
fixes #1837


grantfitzsimmons (approved):
Works exactly as intended.

PR #1870, opened 2022-07-20
some improvements to workbench rollback function.

### LINKED ISSUE #1663: Poor performance for roll back
chrome_RhEPcH4BWq

It has been moving along at around 2-3 agents per second

benanhalt:
Improve rollback speed by using raw deletes. fixes #1663

Better reporting of integrity errors. Inculdes ID and type of
offending record.

Skips records that were deleted subsequent to the upload.


grantfitzsimmons:
It was definitely quicker and rolled back 5,000 rows without an issue. Need to test with another data set, but it seems to work well.

grantfitzsimmons (approved):
The object specific warning is a great touch.

PR #1883, opened 2022-07-23
Fix query table sorting not accounting for hidden fields

### LINKED ISSUE #1880: [SP7.7] Sorting in search results, by clicking on column heading, is broken if some columns are set to not display in results
**Describe the bug**
its possible to click on column headings in the search results, which places sort symbols on these.  this should sort the results accordingly.  
However, currently this function is broken if some columns ae hidden.  - it seems that there's an "off-by-1" error - if 1 column is hidden, then that results in the sorts being applied to a column 1 to the left of the one that was requested.  

this is the list if I click on the barcode column.  it claims to be sorting by barcode, but its clearly not.  I think its sorted by taxon instead...:
image

in terms of the underlying bug.  I think its this:  If I scroll up and look at the query - it looks like there's a off by 1 error between the table columns and the search.  Likely triggered by the fact that the first column is set to be non visible.  
query screen - showing the sorts that were set by the clicking on columns I did:
image

columns - note that which columns are filtered on is offset by 1:
image

if there's no hidden columns, everything works as expected.  

**Reported By**
RBGE

maxxxxxdlp:
Fixes #1880

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/queryresultstable.tsx
+++ b/specifyweb/frontend/js_src/lib/components/queryresultstable.tsx
@@ -282,7 +282,7 @@ export function QueryResultsTable({
   readonly sortConfig?: RA<QueryField['sortType']>;
   readonly onSelected?: (resource: RA<number>) => void;
   readonly onSortChange?: (
-    fieldIndex: number,
+    fieldSpec: QueryFieldSpec,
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (approved):
Works great. Ran several queries with hidden fields- as well as sorted on multiple columns- no issues.

PR #1887, opened 2022-07-23
Disallow negating "any" filter

### LINKED ISSUE #1876: NOT can be hidden in query when "Any" is selected
chrome_SwPZmoY4nZ

**Recreate:**
1. Create a query
2. Choose any operator besides "Any" and click the NOT button
3. Switch operator to Any
4. Run query
5. See NOT still in effect

version: 7.7.0-edge

maxxxxxdlp:
Fixes #1876

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/querybuilderfield.tsx
+++ b/specifyweb/frontend/js_src/lib/components/querybuilderfield.tsx
@@ -139,18 +139,21 @@ export function QueryLine({
 
       const newFilters = hasParser
         ? field.filters.map((filter) => {
-            const filterType =
+            const resetFilter =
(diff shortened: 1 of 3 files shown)


grantfitzsimmons (approved):
Seems to be working great across several queries and databases

grantfitzsimmons:
Important note- the negate option is visible for aggregated fields. Is that intentional? It does not cause a problem as far as I can tell.

maxxxxxdlp:
I don't see the negate option. I do see the "+" button for running OR queries.

Since it doesn't make sense to run an OR query for a field that doesn't have any field, I would remove that button in cases like this

Created #1904

PR #1889, opened 2022-07-23
Add a workarround for day.js bug with parsing dd/mm/yyy

### LINKED ISSUE #1875: Can't Save Collection Update Because of Date Format Issue
**Describe the bug**
With the new version, my Specify switched my date format to MM/DD/YYYY and I've been trying to find a way to change it to DD/MM/YYYY without switching my computer's settings. From what I did, the blank date box shows that I can put a date in as DD/MM/YYYY, but it won't let me save the collection object unless I change the date to MM/DD/YYYY format. In other words, when I have 20/01/1970 in my date, I can't save, but when I change it to 01/20/1970 I'm given the option to save.

**To Reproduce**
1. Click username to open settings.
2. Click preferences
3. Unselect 'Use accessible full date picker'
4. Start a new collection object
5. Blank date box should read 'DD/MM/YYYY'
6. Add a collection object as normal with dd/mm/yyyy date format and save button will be red and unusable.

**Expected behavior**
I was hoping that by unselecting 'Use accessible full date picker' I would be able to have the dd/mm/yyyy format for my Specify without changing my computer's settings.

**Screenshots**
I have the date in dd/mm/yyyy format and the save button is red
image
Changed to mm/dd/yyyy format even though the blank box says I should use dd/mm/yyyy format and save button is orange and clickable.
image

 - OS: Windows 11
 - Browser: Chrome
 - Specify 7 Version: 7.7.0 (I believe, but I'm not sure how to check)
 
**Reported By**
Cornell University

Thank you for your help!

maxxxxxdlp:
Fixes #1875

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/partialdateui.tsx
+++ b/specifyweb/frontend/js_src/lib/components/partialdateui.tsx
@@ -32,15 +32,48 @@ const reversePrecision: RR<number, PartialDatePrecision> = {
 };
 export type PartialDatePrecision = keyof typeof precisions;
 
+// TODO: [REFACTOR] migrate from day.js to date-fns
+/**
+ * Several bugs have been discovered in day.js in the process of testing
+ * Specify 7. The bugs are present in the latest stable version as of this
+ * writing (1.11.4).
+ * Day.js 2.0 is currently in beta. Need to wait for 2.0 to get released or
+ * migrate to date-fns.
+ */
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (approved):
Works great with dd-MM-yyyy and yyyy-MM-dd. Tested on CO, CE, and Locality forms.

PR #1891, opened 2022-07-23
Don't include separators for empty formatter parts

### LINKED ISSUE #1873: [Sp7.7] - extraneous commas in table formats
**Describe the bug**
Best shown by a picture.  Note the extra commas.  
7.6:
image
7.7:
image

What's happening is that the table format includes a load of optional fields.  For this (and many other) records, these fields are not populated.  7.6 does the sensible thing, and only prints separators when the field is present.  7.7 seems to always print them.  

this effects a range of fields.  Most effected is Agents, but also seen in locality's.  

Here's the form config in question:
image

**To Reproduce**
feel free to log in to RBGE sandbox.  look at pretty much any record.  e.g. 
https://sandbox-rbge.specifycloud.org/specify/view/collectionobject/298/

**Reported By**
RBGE

maxxxxxdlp:
Fixes #1873

### DIFF
--- a/specifyweb/frontend/js_src/lib/dataobjformatters.ts
+++ b/specifyweb/frontend/js_src/lib/dataobjformatters.ts
@@ -213,7 +213,7 @@ export async function format<SCHEMA extends AnySchema>(
             const field = defined(
               resource.specifyModel.getField(fieldName) as LiteralField
             );
-            return `${separator}${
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (approved):
This looks great! It makes such a big difference.

PR #1895, opened 2022-07-25
update python from 3.6 to 3.8.

benanhalt:
Updates Python used by the back end from 3.6 to 3.8 which is good until October 2024.

I think this should be a pretty safe update. It would be good to test drive it using test.specifysystems.org before merging.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ LABEL maintainer="Specify Collections Consortium <github.com/specify>"
 RUN apt-get update \
  && apt-get -y install --no-install-recommends \
         gettext \
-        python3.6 \
+        python3.8 \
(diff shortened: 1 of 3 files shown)


grantfitzsimmons (approved):
I was able to 
* Successfully create, clone, add another, and delete records from CO, Locality, CE, Taxon, and Agent

* Validate, Upload, Roll back, and modify WB data sets

* Tested several reports and labels- ran correctly

maxxxxxdlp:
Is this a duplicate of https://github.com/specify/specify7/pull/1895?

benanhalt:
This is #1895. I think you mean #1915 which I mistakenly said updates Python. I'll fix that now.

PR #1897, opened 2022-07-26
Hide front-end only fields from the Query Builder

### LINKED ISSUE #1896: Query Builder exposes front-end only fields
Front-end only fields like `Collection Object -> Current Determination` are exposed in the query builder.

Trying to run a query on such field results in an exception. Thus, they should be hidden.

maxxxxxdlp:
Fixes #1896


grantfitzsimmons (approved):
Those front end fields are not visible anymore. It would be nice to have Current Determination as a relationship for queries or reports and labels someday...

Issue #1845, opened 2022-07-18
Table relationships showing up as hidden in query builder when not


tlammer:
Discovered by Willem

Preparation property is showing up as hidden in the query builder but neither the field nor the table are hidden.  I unhid the field in the schema (preparation.preparationproperty), confirmed the table itself is not hidden, then created a new query.  Preparation property is listed under hidden relationships

maxxxxxdlp:
When we were working on the workbench mapper, we found all kinds of relationships that are visible in the default schema configuration, but probably should not be exposed in the workbench mapper.

There were so many of these "exceptions", that I had to develop an entire system that would reclassify some hardcoded list of fields and relationships as hidden, optional, readonly, required, system or even remove them completely (even from the `Reveal Hidden Fields`).

The full list of such overrides is here:
https://github.com/specify/specify7/blob/44fe80fc9a9d8da3112af97afaad8487811fcafd/specifyweb/frontend/js_src/lib/schemaoverrides.ts#L64-L201

I would be strongly in favor of getting rid of most if not all of those overrides. As it currently stands, it is impossible for people to unhide those fields, even if they are marked as unhidden in the schema config.

grantfitzsimmons:
I think launching a fix for this ASAP is super important. I would make the query builder ignore these overrides and update the official version of 7.7.

maxxxxxdlp:
Pushed a fix on branch `1845`

@grantfitzsimmons Please test it, and once tested we can merge it into production

PR #1899, opened 2022-07-26
Make Query Builder ignore "hide field" schema overrides

### LINKED ISSUE #1894: [Sp7.7] isCurrent is showing up as a hidden field in query builder, when it is not hidden
**Describe the bug**
In Sp7.6. we could build queries using isCurrent (in our instance, this is renamed to filingname).  
on upgrade to 7.7, isCurrent has become hidden in the query builder:
(as evidenced by its presence in the hidden list below, and the gray background it gets in the builder)    
image

However, in the schema, it is not hidden:
image

Bug here is that it shouldn't be hidden.  why is it?

Additionally - there's a new- super useful looking field been added to the query builder: "Current Determination".  
image
However, its hidden by default, which will mean that few people find it... - and I cna't see any way to unhide it.  
is there a way?  can we get it un hidden by default if not?

Thanks

**Reported By**
RBGE

maxxxxxdlp:
While this kind of schema override is useful in the WorkBench, it
creates issues in Query Builder

Fixes #1894
See also #1845

### DIFF
--- a/specifyweb/frontend/js_src/lib/wbplanviewnavigator.ts
+++ b/specifyweb/frontend/js_src/lib/wbplanviewnavigator.ts
@@ -402,9 +402,7 @@ export function getMappingLineData({
               (generateFieldData === 'all' &&
                 isFieldVisible(
                   showHiddenFields,
-                  scope === 'queryBuilder'
-                    ? model.idField.isHidden
-                    : model.idField.overrides.isHidden,
+                  model.idField.isHidden,
                   model.idField.name
                 ))
                 ? [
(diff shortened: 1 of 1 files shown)


grantfitzsimmons:
image
image

maxxxxxdlp:
Fixed. Please test again

Btw, next time there is an invalid fix, you can press the "Reject" button in the UI to reject the change.
This will give me an "Ask to review again" button that I can press after I fix the bug again.

grantfitzsimmons (approved):
Works great! Excited to get this pushed out to `edge`

PR #1906, opened 2022-07-27
Make specify look for "invisible=true" in form definition

### LINKED ISSUE #1070: Visible=false not making a field invisible on forms
Reported by CSIRO
they are attempting to hide a field on the form, but not comment it out.  They want the field to autopopulate, which means it must be on the form, but they don't want to see it.  They are using the following code:
                <row>

                    <cell type="label" labelfor="4"/>

                    <cell type="field" id="4" name="cataloger" uitype="querycbx" initialize="name=Agent;title=Catalog Agent" isrequired="true"/>

                    <cell type="field" id="5" uitype="plugin" name="this" initialize="name=PartialDateUI;df=catalogedDate;tp=catalogedDatePrecision;visible=false" uifieldformatter="Date" default="Today"/>

                </row>

It works in 6; they don't see it but they can see the field is filling out through queries and MySQL.

maxxxxxdlp:
Allows to make a field invisible

Fixes #1070

### DIFF
--- a/specifyweb/frontend/js_src/lib/parseformcells.ts
+++ b/specifyweb/frontend/js_src/lib/parseformcells.ts
@@ -276,9 +276,9 @@ export function parseFormCell(
      * this checkbox and button would disappear from forms in Specify 7 when
      * users update to 7.7.0.
      * To mitigate the above issues, Specify 7 form definitions are using
-     * "invisible=true" instead of "visible=false" for makign fields invisible
+     * "invisible=true" instead of "visible=false" for making fields invisible
      */
-    visible: getBooleanAttribute(cellNode, 'invisible') ?? true,
+    visible: getBooleanAttribute(cellNode, 'invisible') !== true,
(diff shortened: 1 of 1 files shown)


grantfitzsimmons:
```
                <row>
                    <cell type="label" labelfor="19"/>
                    <cell type="field" id="19" name="text2" invisible="true" uitype="text" colspan="9"/>
				</row>
```

DOI is text2. I am still seeing it appear...

image

maxxxxxdlp:
This actually was looking for `invisible` inside of `initialize` (i.e `initialize="invisible=false"`), but I think your idea of putting it outside of initialize is better

grantfitzsimmons (changes requested):
https://issue-1070.test.specifysystems.org/specify/view/referencework/new/

Same form defintion as before-

db: `roles_sp77_fish`

grantfitzsimmons (approved):
image

Works great!

PR #1915, opened 2022-07-29
Update deps

benanhalt:
Python was updated from 3.6 to 3.8 by https://github.com/specify/specify7/pull/1895

Updates django from 2.2 to 3.2
Updates mysqlclient from 1.3 to 2.1
Updates requests from 2.26 to 2.28
Updates celery from 5.0 to 5.2
Updates pycryptodome from 3.9 to 3.15

Removes django-jsonfield-backport which is no longer needed with django 3.2

Manual testing should concentrate on basic form operations, logging in, changing passwords, and workbench functions.


requirements.txt
```diff
@@ -1,9 +1,9 @@
 redis==3.5.3
 celery==5.0.3
-Django==2.2.10
-mysqlclient==1.3.13
+Django==3.2.14
```
maxxxxxdlp, line 3: This makes me excited!

grantfitzsimmons:
@melton-jason Can you test this soon? 

@maxxxxxdlp Procedural testing required? Is there anything else we need to be aware of before merging this?

maxxxxxdlp (approved):

grantfitzsimmons:
> Manual testing should concentrate on basic form operations, logging in, changing passwords, and workbench functions.

melton-jason (commented):
Looks good as far as I can tell! 

Here is most of what I tested: 
- Created three types of records successfully (Collection Object, Taxon, Agent) with varying amounts of data in each. 
- Queried and subsequently edited a portion of each record successfully. 
- Created a new user, tested logging in, changed password, and assigned the previous Agent to the user. 
- Created a new CO record as new user.
- Imported a sample dataset into the Taxon tree from the workbench. 
- Created a new CO form with one of the new Taxon records. 
- Tried deleting the new user (expected and got a deletion blocker because they owned a Record Set). 
- Deleted record set and then the new user.
- Tried to delete Agent (expected and got deletion blockers with records they created).
- Save, edit, and delete a Query.

maxxxxxdlp:
Then I think this should be merged.
From this kind of change, if something were to be broken, it would throw obvious errors

grantfitzsimmons:
Tested and successfully:

- Created new CO, CE, Agent, and Geo records with all subforms filled out on two databases (`newmexico` and `coldFish`)
- Edit and save all new records (removing relationships, setting new values, removing values)
- Created two new users, set passwords, change passwords on existing users, test log in and user deletion
- Created new queries, modify new query, save, and delete queries
- Modified existing queries, modify queries, and delete
- Modified WorkBench data set, mapping, validated, and uploaded data sets. Rollback works great!

maxxxxxdlp:
Can't merge this for now as one of the back-end tests is failing after we rebased the code (it worked before the Django update)

grantfitzsimmons:
@acwhite211 

We need to do another round of this soon so that we can resolve the issue with developers manually adding their `requirements.txt`.

From @maxxxxxdlp:
> we could try updating ubuntu used in docker and getting newer python - have to do that anyway, and it might just fix this

maxxxxxdlp:
good idea (as long as you don't have any other burning fires right now)
would suggest updating to the newest version for as many libs and dependencies as possible so that we don't have to do it again for another year

PR #1916, opened 2022-07-30
Fix QueryComboBox not being full width

### LINKED ISSUE #801: Allow users to define grid column widths on data entry forms
Requested by Edinburgh.  Currently, the browser resizes based on content

maxxxxxdlp:
Fixes #801


grantfitzsimmons (approved):
Looks good on RBGE's test database after the fix.

PR #1965, opened 2022-08-15
Fix "Create CSV" button missing for distinct queries

### LINKED ISSUE #1956: Cannot export distinct query results to CSV
In Specify 6, the ability to export the results of a query with "Distinct" checked is available.

In Specify 7, this function is hidden. It seems potentially related to #1464, but as it works in 6, it can be implemented.

**Reported By:** Corinna at CSIRO

maxxxxxdlp:
Fixes #1956


maxxxxxdlp:
This change also displays the "Create KML" button when running a distinct query.
@grantfitzsimmons can you check if it is working correctly

grantfitzsimmons (approved):
It is working! Exported several queries and KML. Works as it does when it is not distinct.

PR #1975, opened 2022-08-16
Fix WB crashing on no data set update permission

### LINKED ISSUE #1932: Unable to open a data set since WB cannot find locality components
image

image

**Max:**

That's where the error happens

image

The error is because workbench can't find GeoMap, GeoLocate or Coordinae Converter Buttons

This might happen because user doesn't have sufficient permissions for those actions (i.e, no access to the Locality table)

**Reported By:** Lucas for the KU Mammals collection

maxxxxxdlp:
Fixes #1932


grantfitzsimmons (approved):
image

PR #1988, opened 2022-08-17
Fetch pick lists on demand

### LINKED ISSUE #1987: Improve pick list fetching performance
In Specify 7.7, all pick lists are fetched every time the app opens.

This can take up to 10 seconds to complete if the database is busy, or there are a lot of pick lists.

RGBE experienced performance issues after the update, which in part can be attributed to this.

Need to fix this by returning back to loading pick lists dynamically.

----

For future reference:
Reasons why Specify 7.7 is loading all pick lists on app load, rather than on demand:
- I did it this way just because it was simpler (I fetch all pick lists together so that places that deal with pick lists don't have to deal with async code)
- The other reason is that not fetching things on demand makes UI feel faster, at a cost of slower initial page load time. I did that trade-off because I assumed specify would be opened once and used for several hours.
- Also, this request is not part of the initial context, so it does not block the UI from rendering.

About the only place that really needs to fetch all pick lists is the schema editor.

maxxxxxdlp:
Fixes #1987

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/picklisttablecombobox.tsx
+++ b/specifyweb/frontend/js_src/lib/components/picklisttablecombobox.tsx
@@ -1,13 +1,10 @@
 import React from 'react';
-
-import { f } from '../functools';
 import { getPickListItems, PickListTypes } from '../picklistmixins';
-import { fetchPickLists, getPickLists } from '../picklists';
 import { resourceOn } from '../resource';
 import type { RA } from '../types';
 import type { DefaultComboBoxProps, PickListItemSimple } from './combobox';
-import { useAsyncState } from './hooks';
 import { PickListComboBox } from './picklist';
+import { pickListTablesPickList } from '../picklists';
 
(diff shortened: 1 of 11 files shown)


grantfitzsimmons:
@maxxxxxdlp How can I test this?

maxxxxxdlp:
This pull request improves load performance of Specify on slow databases.
It is now deployed to RGBE's Specify instance and they will tell us if the performance is improved

In the meanwhile, we need to check that this pull request did not break pick lists:
- Check that pick lists are rendered properly in the form
- Check that pick list items are fetched properly in:
   - Query Builder filters
   - Workbench cells for columns assigned to a pick list
- Pick lists item titles are used in query results (rather than pick list item value)
- List of pick lists is displayed properly in the schema editor

grantfitzsimmons (approved):

grantfitzsimmons:
@maxxxxxdlp There is a conflict before merging

grantfitzsimmons (approved):

PR #1999, opened 2022-08-18
Add backward compatibility with pick lists in Sp6 QB

### LINKED ISSUE #1934: [Sp7.7] - on upgrade, filters for some picklists disappear from queries in the UI.
Scenario:
we setup some queries in Specify 6, or 7.6 (not sure which they were created through).  
these queries were for loans, and included filtering on the LoanAgent field.  
Like this:  
image

However, when I view the same query in Specify 7.7, the Herbarium picklist bit isn't there:
image

I can select Herbarium from the dropdown, but it should be set already, and isn't.  
image

interestingly, Herbarium does seem to be being applied in the search - so this is a query builder UI glitch, not a underlying query engine glitch I think.  

If I then set the AgentType to be herbarium through the 7.7 query builder, and hit save, then it does remember it.  therefore I assume that what is happening is that the way that 7.7, and 6 / 7.6 is storing queries that include picklists is slightly different, and there's an incompatibility there?

thanks.

maxxxxxdlp:
Sp7 QB stores PickListItem.value when selecting a pick list item as a
column filter

Sp6 QB uses PickListItem.title column, which is probably a bug or
an oversight as that column is intended for UI purposes only.

This commit makes Sp7 UI be able to resolve PickListItem even if it
was saved using the title field. The value would still be saved using
the value field, as that is the correct behavior.

Fixes #1934

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/querybuilderfieldfilter.tsx
+++ b/specifyweb/frontend/js_src/lib/components/querybuilderfieldfilter.tsx
@@ -167,7 +167,14 @@ function QueryInputField({
         {...commonProps}
         required={Boolean(validationAttributes.required)}
         multiple={listInput}
-        value={listInput ? value.split(',').map(f.trim) : value}
+        value={
+          listInput
+            ? value
+                .split(',')
+                .map(f.trim)
+                .map((value) => resolveItem(pickListItems, value))
+            : resolveItem(pickListItems, value)
+        }
         size={listInput ? selectMultipleSize : 1}
       >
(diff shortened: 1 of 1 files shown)


maxxxxxdlp:
To test:
Create a query with `Agent -> Agent Type` pick list in Sp6
Open it in Sp7
The pick list item should still be selected
Chose a different pick list item
Open query in sp6
The pick list item is not selected (because sp6 is doing pick lists wrong!)

grantfitzsimmons (approved):

PR #2035, opened 2022-08-25
Allow customizing SubView grid column widths

### LINKED ISSUE #2008: Add a better way to customize column widths
The `colspan` hack for https://github.com/specify/specify7/issues/801 works, but we need a better solution.

Possibly, can add a line like `<coldef os="table">...</coldef>` that would supply the column width specifically for the grid view.

This would also make it easier to create a visual editor for the forms

maxxxxxdlp:
Fixes #2008

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/basic.tsx
+++ b/specifyweb/frontend/js_src/lib/components/basic.tsx
@@ -193,6 +193,18 @@ const dataEntryButton =
       />
     );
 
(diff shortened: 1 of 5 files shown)


maxxxxxdlp:
Documentation: https://github.com/specify/specify7/wiki/Form-System#customizing-form-table-column-widths

grantfitzsimmons (approved):

PR #2065, opened 2022-08-30
Don't check for institutional policies in roles

### LINKED ISSUE #2019: Permission error when Collection Admin tries to create a new user
Screen Shot 2022-08-23 at 12 42 01

maxxxxxdlp:
If user has a "All -> All" policy on the collection level, don't use
that as a justification for an institutional policy, since you can
not give it at a collection level.

Fixes #2019

To Test:
- Create a user that has a "Collection Admin" role and no policies (other than collection access)
- Sign in as that user
- Open the "Create User" form
- Wait for everything to load
- No permission error should happen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix error when adding Address of Record SubView to Borrow form
   ([#2006](https://github.com/specify/specify7/pull/2006)) -
   _Reported by CSIRO_
+- Fix Specify incorrectly checking for `Permissions -> List Admins ->
+  read` permission ([#2019](https://github.com/specify/specify7/issues/2019))
 
 ## [7.7.3](https://github.com/specify/specify7/compare/v7.7.2...v7.7.3) (26 September 2022)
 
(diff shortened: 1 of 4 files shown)


grantfitzsimmons (changes requested):
image

After following these steps, my user `user-man` encounters an institution permission error when trying to create a new user.

`KU_Fish`

grantfitzsimmons:
https://user-images.githubusercontent.com/37256050/191332024-aa42d9c3-9895-49a4-ba77-1b1d9cbbdccd.mov

**Grant:** Should I be able to make a user?

**Max:** Probably, do you get a permission erorr?

**Grant:** Sometimes- It’s odd. The first time I did not get an error, was able to set a password from the dialog that appeared, but it didn’t save

maxxxxxdlp:
FYI: Setting a password from that dialog does not save. You have to manually press the Save button again.
It's that way because it was simpler. If it's a usability issue we can fix it.

maxxxxxdlp:
The user did not have permision to set the password, so the button was missing.

When saving, the dialog prompt to set a password occurred, but it should not have since you did not have permission. It is a bug. Fixed

maxxxxxdlp:
The earlier exception happeded because you checked `Collection Access`, which requires also assigning an agent, but you did not have permission to assign an agent. That bug is fixed now, where the `Collection Access` checkbox is disabled if the user does not already have an agent assigned and you don't have permission to assign an agent

Please test this bug again. Thanks!

grantfitzsimmons:
@maxxxxxdlp I'll test it as soon as it goes live on the test panel. I could tell there was some cascade of issues occurring between my responses. Thanks!

grantfitzsimmons (approved):

PR #2145, opened 2022-09-13
Fix race condition with pick list fetching

### LINKED ISSUE #2144: Race condition when fetching pick list items
Happens in the CSIRO database. Probably can happen in other places too.

To fix:
if pick list is already being fetched, return the same promise, rather than start fetching again.

maxxxxxdlp:
Fixes #2144


grantfitzsimmons:
@maxxxxxdlp How can I test this?

maxxxxxdlp:
I was able to replicate the bug locally and check that the fix fixes the bug

The only thing remaining to test:
Can you please check that places in sp7 that use pick lists still work? (the change affected pick list fetching)

grantfitzsimmons:
image

I'm not running out of resources like before, but the query results are not showing. 

https://anfc-issue-2144.test.specifysystems.org/specify/query/76/

username: `pae008`

Should I request changes or is this another bug?

maxxxxxdlp:
the bug should be reproducible on `production`, but not reproducible on `issue-2144`.

which branch is your screenshot from?

grantfitzsimmons:
> the bug should be reproducible on `production`, but not reproducible on `issue-2144`.
> 
> which branch is your screenshot from?

`issue-2144`

grantfitzsimmons (changes requested):
Try the query

maxxxxxdlp:
Sorry, I misread your message. Yes, this is a bug

maxxxxxdlp:
After investigation, turned out that the database had a circular dependency. Fixing that fixed the issue.

See #2160 for more details.

grantfitzsimmons (approved):
image

PR #2147, opened 2022-09-13
Fix list of tables pick list not working on form load

### LINKED ISSUE #2146: List of tables pick list not working on form load
List of tables pick list not working on form load

It shows `(current, invalid value)` and is disabled

Screen Shot 2022-09-13 at 13 07 58

Changing the picklist type and reverting it back fixes the issue

maxxxxxdlp:
Fixes #2146

To test:
Open any pick list that has type of `Entire table`.
Make sure the list of tables pick list is not disabled and is interactive

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,81 +8,149 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


maxxxxxdlp:
Needed to change pick list type from `User Defined Items` back to `Entrie table`

The bug should happen again now

grantfitzsimmons (approved):

Issue #2006, opened 2022-08-19
Adding addressOfRecord to Borrow breaks the form


grantfitzsimmons:
> I’ve been having some issues with the Borrow form in S7 and I think I have narrowed it down to the addressOfRecord subform.
> 
> The below xml code to add the addressOfRecord subform as a button works fine in the Loans. The Borrow form with this snippet opens in S6, however, in S7 I get an error message (see attached). I also get an error when converting it to a ‘normal’ subform. When I edit out the subform altogether, the Borrow form opens in S7 as well.

`<cell type="subview" id="aor" name="addressOfRecord" viewname="AddressOfRecord" initialize="btn=true"/>`
 

Error message:
```
Cannot read properties of undefined (reading 'model')
 
Stack:
 
TypeError: Cannot read properties of undefined (reading 'model')
    at d (http://anwc.specify-dev.csiro.au/static/js/main.2378b38b2f21141133e8.bundle.js:2:25863)
    at AddressOfRecordDependentCollection.initialize (http://anwc.specify-dev.csiro.au/static/js/main.2378b38b2f21141133e8.bundle.js:2:26218)
    at n.Collection (http://anwc.specify-dev.csiro.au/static/js/main.2378b38b2f21141133e8.bundle.js:2:345732)
    at AddressOfRecordDependentCollection.CollectionBase [as constructor] (eval at c (http://anwc.specify-dev.csiro.au/static/js/main.2378b38b2f21141133e8.bundle.js:2:17494), <anonymous>:6:17)
    at AddressOfRecordDependentCollection.constructor (http://anwc.specify-dev.csiro.au/static/js/main.2378b38b2f21141133e8.bundle.js:2:26116)
    at AddressOfRecordDependentCollection.DependentCollectionBase [as constructor] (eval at c (http://anwc.specify-dev.csiro.au/static/js/main.2378b38b2f21141133e8.bundle.js:2:17494), <anonymous>:6:17)
    at new AddressOfRecordDependentCollection (eval at c (http://anwc.specify-dev.csiro.au/static/js/main.2378b38b2f21141133e8.bundle.js:2:17494), <anonymous>:6:17)
    at http://anwc.specify-dev.csiro.au/static/js/185.fc70b8fffe6ae0eb5b94.bundle.js:1:178701
```
 
 **Reported By:** Corinna at CSIRO

maxxxxxdlp:
@benanhalt I have a question on what would be the best solution to this issue 

[...]

PR #2155, opened 2022-09-15
address of record should not be a dependent relationship.

benanhalt:
for #2006 

Changes all relationships to addressOfRecord to independent.


melton-jason:
## Specify 7
The subview for AdressOfRecord now shows on the Borrow form, but it is not allowing any data to be added (see attached picture below)

sp7_subview

For something of a workaround, I _was_ able to use a Query Combo Box's add feature to create an Address of Record related to the Borrow, but Address of Record is not defined in Specify 6's `/config/backstop/typesearch_def.xml` so no existing Address of Records can be related with this way. 

XML used for Subview:
`<cell type="subview" id="aor" name="addressOfRecord" viewname="AddressOfRecord" colspan="10"/>`

XML used for Query Combo Box:
`<cell type="label" labelfor="aor1"/>`
`<cell type="field" id="aor1" name="addressOfRecord" uitype="querycbx" initialize="name=AdressOfRecord"/>`

### Notes
If both the Query Combo Box and Subview are on the same form, whenever an Address of Record is created from the QueryComboBox, it populates data in the Address of Record Subview as read-only. (See Below)

sp7_combo_and_subview


## Specify 6
The behavior in Specify 6 is quite different.  The  Address of Record subview works as intended, with no issues. When trying to use a Query Combo Box, Specify throws an error (stating that Address of Record is not in `typesearch_def.xml`) and does not show the Borrow form.

melton-jason (commented):
## Specify 7
The subview for AdressOfRecord now shows on the Borrow form, but it is not allowing any data to be added (see attached picture below)

sp7_subview

For something of a workaround, I _was_ able to use a Query Combo Box's add feature to create an Address of Record related to the Borrow, but Address of Record is not defined in Specify 6's `/config/backstop/typesearch_def.xml` so no existing Address of Records can be related with this way. 

XML used for Subview:
`<cell type="subview" id="aor" name="addressOfRecord" viewname="AddressOfRecord" colspan="10"/>`

XML used for Query Combo Box:
`<cell type="label" labelfor="aor1"/>`
`<cell type="field" id="aor1" name="addressOfRecord" uitype="querycbx" initialize="name=AdressOfRecord"/>`

### Notes
If both the Query Combo Box and Subview are on the same form, whenever an Address of Record is created from the QueryComboBox, it populates data in the Address of Record Subview as read-only. (See Below)

sp7_combo_and_subview


## Specify 6
The behavior in Specify 6 is quite different.  The  Address of Record subview works as intended  (Seem to be only able to relate 1 Address of Record ).  When trying to use a Query Combo Box, Specify throws an error (stating that Address of Record is not in `typesearch_def.xml`) and does not show the Borrow form.

grantfitzsimmons (changes requested):
See @melton-jason's breakdown

maxxxxxdlp:
@benanhalt Why does the subview work in Specify 6? How does it handle rendering subviews for independent relationships?

benanhalt:
It's probably treating it as a one-to-one.

maxxxxxdlp:
Looks good to me.

Since the relationship is now independent, it is not supposed to be used as a SubView. Instead, it should be rendered as a query combo box.

For the sp6 error, the user will have to add an entry to the `TypeSearches` app resource.

melton-jason:
@grantfitzsimmons @maxxxxxdlp 
After doing some digging, it seems like a plethora of other files would need to be edited as well to allow AddressOfRecord be used as a query combo box. The files for Specify 6 seemed to have been created with the intention that AddressOfRecord would not be a query combo box. 

Aside from the query combo box being added to the views.xml, here is a list of other files I have found that would (most likely) need to be edited:
* Located in "Specify/config/backstop/"
  * typesearch_def.xml
  * search.views.xml
  * search_config.xml
  * dialog_defs.xml

* Located in "Specify/config/"
  * (possibly) specify_tableid_listing.xml

I have attempted editing, with the goal of adding AddressOfRecord as a query combo box, these files but have had no luck in Specify 6.  If needed, I can provide the changes I have tried.

### Note
The behavior does change when AddressOfRecord is added to typesearch_def.xml. An error message is shown (attached below) and the form loads, but does not load the query combo box. 

Screenshot 2022-09-21 170154

I have narrowed the problem in the source code of Specify 6 to an error being thrown in `src/edu/ku/brc/af/ui/forms/validation/ValComboBoxFromQuery` (line 216) stating a variable in `src/edu/ku/brc/af/core/db/DBTableInfo` (public facing getter for variable is on line 230) is null. 

The only reference I have found of this variable in the xml forms is in `specify_tableid_listing.xml`

maxxxxxdlp:
> It's probably treating it as a one-to-one.

Can we do the same in Specify 7?

Or is sp6 behaving that way just because it has ugly exceptions for arbituary fields?

benanhalt:
If I set it back to a dependent relationship, that makes it a one-to-one and it should be embeddable as a subview. But it seems like that wasn't working for some reason? Or maybe we just thought it should be shared?

maxxxxxdlp:
I have it working now on #2201 (with dependent relationship). Should be compatible with Sp6.

@benanhalt A reverse relationship must exist for subview to be able to display it (Because of collectionApi.js requiring that). See https://github.com/specify/specify7/issues/2006#issuecomment-1245587592 for more detail

I fixed that in that PR by creating the missing relationship on the front-end: https://github.com/specify/specify7/pull/2201/files#diff-456bc7bb48b232db3a3d29dd34e9381c4bbfb0d78054544402c5b110998907e3R332

Do you want to declare it on the back-end instead?

benanhalt:
I think #2201 is the correct way to handle it. It could be added to the backend, but it would just be an override like you did in the frontend. By rights, it should be declared in the Specify 6 Hibernate annotations. Then it would be included in the specify_datamodel.xml and thence the Specify 7 datamodel.

PR #2157, opened 2022-09-16
Support customizing CO formatter in ColRels Plugin

### LINKED ISSUE #1981: Add ability to choose the dataObjFormatter to use for displaying related collection object using the collection relationship plugin.
The current plugin represents the related collection object with its catalog number. Since multiple formatters can be defined for a table, it should be possible to define one for use in the plugin. The collectionRelType table should be augmented with two columns "leftSideFormatter" and "rightSideFormater" that would contain the formatter name to use when presenting the related collection object from the other side of the relationship. If the formatter name is null it would default to using the catalog number for continuity with existing behavior.

maxxxxxdlp:
Fixes #1981

Added a `formatting` attribute - https://github.com/specify/specify7/wiki/Form-System#collectionrelonetomanyplugin

The [ColRelTypePlugin](https://github.com/specify/specify7/wiki/Form-System#colreltypeplugin) plugin supports that attribute too

It supports the name of the formatter only at present. It will be expanded in the future to support providing multiple values

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/collectionrelonetomanyplugin.tsx
+++ b/specifyweb/frontend/js_src/lib/components/collectionrelonetomanyplugin.tsx
@@ -45,7 +45,8 @@ type Data = {
 
 const processRelationships = async (
   relationships: RA<SpecifyResource<CollectionRelationship>>,
-  otherSide: 'left' | 'right'
+  otherSide: 'left' | 'right',
+  formatting: string | undefined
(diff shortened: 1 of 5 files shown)


grantfitzsimmons:
image

This is pretty cool.

grantfitzsimmons (approved):
Nice!

specifysoftware:

https://discourse.specifysoftware.org/t/specify-7-7-3-release-announcement/774/1

Issue #1283, opened 2022-04-19
Collection Relationships should work without matching Cat # formatters


grantfitzsimmons:
**Is your feature request related to a problem? Please describe.**
Specify 6 and 7 have no way to create a collection relationship with collections that use different catalog number formats. This restriction is very frustrating for users.

CSIRO wants to create collection relationships for their Fish Collection, connecting the main collection to tissues and X-rays. They use unique catalog number formatting making this difficult to implement due to the current restrictions.

**Describe the solution you'd like**
CRs should be configurable between all collections of varying Catalog Number formatting.

**Screenshots**

The restriction seems to be arbitrary:

![image](https://user-images.githubusercontent.com/37256050/164062327-b91355d8-ed95-4cd1-a6ad-8200c20c4558.png)
_Screenshot from CSIRO_

**Reported By**
CSIRO

maxxxxxdlp:
I don't know enough about how collection relationships work, but what specific part in specify 7 is preventing collection relationships between collections with different catalog numbers?

Is there some error message?

Or is it just a matter of the search dialog enforcing a certain catalog number formatter?

If the latter is the case, you can open the "Query Builder" dialog, and add a "Contains" or "Like" filter on the "Catalog Number" (those don't have auto-formatting enabled)

benanhalt:
It is my strong suspicion that this restriction stems from the Specify 6 UI determining the formatters for fields based on the logged in collection. Obviously using a collection relationship will involve dealing with catalog numbers of the associated collection. If those are different, then there would have to be some mechanism for handling two different formatters for the "same" field.

The same problem would arise in Specify 7. Resolving it would probably involve adding a mechanism to any UI involved with the collection relationships to keep track of the "foreign" catalog number formatter. That would include being able to access the formatter definition itself. A collection's catalog number formatter is indicated in the collection table, so at least it wouldn't be necessary to access the schema config of the related collection, but it would be necessary to load the "UIFormatters" app resource of the foreign collection. That is relatively straightforward on the back end and could be accomplished by adding a collectionid  parameter to the app resource endpoint.

I'll let @maxxxxxdlp address the front end ramifications.

grantfitzsimmons:
![image](https://user-images.githubusercontent.com/37256050/189924311-bf33a4c4-2a4d-4f4a-9189-6ba1c8d9a755.png)

For the record, I was able to establish a Collection Object Relationship between two collections with different catalogNumber formatters.

This required the following form definition to establish:

```
[...]

PR #2186, opened 2022-09-19
Allow running query with invalid search parameters

### LINKED ISSUE #2185: Invalid query search parameters should not prevent running query
Query builder's search filters may be considered invalid because of a bug, a limitation of the validation system or because user is looking for invalid data. (see #1283 for an example)

Searching with invalid fields does not corrupt data (unlike saving records with invalid data), thus the validation system should not prevent running a query that has invalid filters

maxxxxxdlp:
Fixes #2185

Entering invalid parameters still shows that they are invalid, but does not prevent running the query


grantfitzsimmons (changes requested):
The query ignores the invalid search term:

image

You can try it here:
https://kufish-issue-2185.test.specifysystems.org/specify/query/42/

user: `abentley`

maxxxxxdlp:
Looks like it didn't like invalid values and used send empty strings instead. Fixed.

Though, now the back-end is complaining about invalid values:

Screen Shot 2022-09-20 at 12 55 38

@benanhalt

benanhalt:
I fixed the backend choking on values that don't match the formatter. I'm seeing "format doesn't match" popup though that doesn't scroll with the rest of the page.
image

maxxxxxdlp:
The popup is generated by your browser. Unless the issue get's fixed by Firefox, there isn't much we can do

benanhalt:
This should fix https://github.com/specify/specify7/issues/1283#issuecomment-1256589366

grantfitzsimmons:
image

grantfitzsimmons (approved):

Issue #1959, opened 2022-08-15
Add support for explicitly choosing a formatter in Query Builder


maxpatiiuk:
Query Builder currently uses the default formatter.

If we were to add an ability to explicitly select a formatter, users would be able to create a formatter than includes all the fields they need, with a separator they need. Then, that formatter could be used in reports/labels.

This will essentially add a way to concatenate fields in the Query Builder, which has been requested by many (including CSIRO)

benanhalt:
Pretty sure the backend already supports including a format name for a query field. The name would identify one of the formatters/aggregators defined in dataobjformatters.xml. 

For use in labels, it would be preferable to extend the formatter/aggregator syntax to generate JSON structures that could then be destructured using a scriptlet. Ideally it would support some kind of JSON path syntax.

https://stackoverflow.com/questions/41405510/json-query-and-json-path

maxxxxxdlp:
What is the syntax for specifying Formatter/Aggregator?

The following `SpQueryField` does not work:

```
{
      "tablelist": "1,10,2",
      "stringid": "1,10,2.locality.locality.FormatterName",
      "fieldname": "locality",
      "isrelfld": true,
      "sorttype": 0,
      "position": 0,
      "isdisplay": true,
      "operstart": 8,
      "startvalue": "",
      "isnot": false
    }
```

grantfitzsimmons:
Andy is interested in having his catalogNumber added to the base URL (“https://freshfish-testability.test.specifysystems.org/specify/bycatalog/KUI/” + “catalogNumber”) in the query output for each item.

This could be accomplished using a custom table format, but then it would apply everywhere. Solving this could allow him to specify a specific format for queries/forms and a specific format for exporting data.

benanhalt:
@maxxxxxdlp try something like:
```
 {
      "tablelist": "1,10,2",
      "stringid": "1,10,2.locality.locality",
      "fieldname": "locality",
      "isrelfld": true,
      "sorttype": 0,
      "position": 0,
      "isdisplay": true,
      "operstart": 8,
      "startvalue": "",
      "isnot": false,
      "formatname": "name of the formatter defined in dataobjformatters.xml"
    }
```

maxxxxxdlp:
That works! Awesome. I will add a UI for it

benanhalt:
[...]

PR #2195, opened 2022-09-20
Export improvements

benanhalt:
This PR adds some improvements for the back end for the data export or publishing system. ie Darwin Core Archives and the RSS export feed.

1. It converts the export jobs to Celery tasks so that they execute on the worker process (like workbench uploads) and not as a thread on the main Specify 7 application server.

2. Adds a Celery "beat" definition to check whether the export feeds are due for updates. Previously this relied on cron jobs or Systemd timers to execute a Django management command (https://github.com/specify/specify7/blob/v7.7.5/specifyweb/export/management/commands/update_feed.py)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,7 @@ typecheck:
 	$(MYPY) --follow-imports silent \
 	specifyweb/permissions specifyweb/workbench specifyweb/accounts \
 	specifyweb/specify/schema.py specifyweb/specify/load_datamodel.py \
-	specifyweb/specify/api.py specifyweb/context/user_resources.py
+	specifyweb/specify/api.py specifyweb/context/user_resources.py \
+	specifyweb/export/tasks.py
 
 .FORCE:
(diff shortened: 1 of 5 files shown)


benanhalt:
Future directions:

1. Add support for a _static field_ declaration in query stanzas (https://github.com/specify/specify7/wiki/Darwin-Core-Archive-Publishing#query-stanzas). This would allow exporting property type tables by including a query stanza that filters on each property type and includes the type as a static field and the value as a queried field. Also could be useful for exporting attachments and attachment-like records from different tables. The static fields could be used to fill in columns where needed to get the different stanzas all returning the same columns.

2. Support specifying the field formatters for fields in the queries so they don't just use the default ones for the collection. This would help with formatting dates correctly for aggregators and adding rounding to latlong values, etc. (https://github.com/specify/specify7/issues/1959#issuecomment-1301310540)

PR #2202, opened 2022-09-22
remove picklist size business rule.

### LINKED ISSUE #1025: Should picklist size even be enforced?

benanhalt:
Fixes #1025


maxxxxxdlp (approved):
Was able to add 4 items to a pick list with size limit of 3

grantfitzsimmons:
Excellent!

PR #2203, opened 2022-09-22
Fix autocomplete bugs on backspace

### LINKED ISSUE #2169: Backspacing in the det.taxonname field triggers autocomplete
**Describe the bug**
After selecting a name in the taxon name field of a determination, if I back space in the field, it takes a few backspaces then puts the name back again. For example, from an empty taxon name field, I add 'Gerrhosaurus flav', which gives me two picklist options. I select 'Gerrhosaurus flavigularis flavigularis'. Then if I backspace a few times it removed the characters, but then 'Gerrhosaurus flavigularis flavigularis' is added in again, and the cursor jumps back to the end of the string. This only happens with the backspace key, I don't get the same behaviour if I select part of the name and hit the delete key. 

**To Reproduce**
1. Select a taxon name in the det taxon field, and then hit the backspace key a few times.

**Expected behavior**
After backspacing, the picklist should appear again, and the name that was in the full string that was in the field should not be added back automatically.

**Screenshots**

**Desktop:**
 - OS: Windows 10
 - Browser: Chrome
 - Specify 7 Version: 7.7.2

**Reported By**
Natural Science Collections Facility

**Additional context**
None

maxxxxxdlp:
Fixes #2169


grantfitzsimmons (approved):

grantfitzsimmons:
@maxxxxxdlp This can be merged when you are done with the changelog

chanulee1 (changes requested):
Same behavior

melton-jason:
I will put this here, for any who needs it:

The issue seems to be caused by the search algorithm as defined in Preferences -> Forms -> Query Combo Box -> Search Algorithm (for relationships with tree tables). 
When the algorithm is searching using "Contains", there is no issue. However, the bug is present if the search algorithm is "Starts-with".

maxxxxxdlp:
Fixed again. Please test (after it builds - https://github.com/specify/specify7/actions/runs/3206411389)

chanulee1 (changes requested):
As Jason noted, the bug seems to happen when the search algorithm for relationships with tree tables is "Starts with". (Used taxon qcbx in collection object form).
db: `kubirds`

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/specify-7-7-4-release-announcement/807/1

PR #2219, opened 2022-09-26
Business rule improvements

benanhalt:
In the production version of Specify 7, when business rules are violated a BusinessRuleException is raised which contains only a string type message. This exception is caught by Django and results in an http 500. This PR begins to address this by adding structure to BusinessRuleException instances (at least a type which could be interpreted by the frontend, and a message which is analogous to the old string type message). It also adds Django middleware to catch these exceptions and return the structured data in the response with status of 400 instead so they may be distinguished by the frontend from generic server errors.


grantfitzsimmons:
@maxxxxxdlp Are we looking to revisit implementing these changes after 7.8 releases?

maxxxxxdlp:
Yes. Fixing this would be awesome!

melton-jason:
A version of these changes have been implemented in #2702

PR #2255, opened 2022-09-30
Fix relationship pick lists not working as expected

### LINKED ISSUE #2230: Pick list is being displayed as an API link
image

image

https://emoryherbarium-testability.test.specifysystems.org/specify/view/collectionobject/23110/?recordsetid=878

When saving a preparation with a valid value in the picklist, the following error occurs:

image

[Specify 7 Crash Report - 2022-09-22T21 46 11.929Z.txt](https://github.com/specify/specify7/files/9659301/Specify.7.Crash.Report.-.2022-09-22T21.46.11.929Z.txt)

User: `spadmin`
Database: `emoryherbarium`

maxxxxxdlp:
Fixes #2230

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   _Reported by CSIRO_
 - Fix Specify incorrectly checking for `Permissions -> List Admins ->
   read` permission ([#2019](https://github.com/specify/specify7/issues/2019))
+- Fix PickLists to relationship fields not being displayed correctly
+  ([#2230](https://github.com/specify/specify7/issues/2230))
 
 ## [7.7.3](https://github.com/specify/specify7/compare/v7.7.2...v7.7.3) (26 September 2022)
 
--- a/specifyweb/frontend/js_src/lib/components/picklist.tsx
+++ b/specifyweb/frontend/js_src/lib/components/picklist.tsx
@@ -9,7 +9,7 @@ import { formsText } from '../localization/forms';
 import { queryText } from '../localization/query';
 import { hasToolPermission } from '../permissions';
 import { PickListTypes } from '../picklistmixins';
-import { resourceOn } from '../resource';
+import { getResourceApiUrl, resourceOn } from '../resource';
 import { schema } from '../schema';
 import type { RA } from '../types';
 import { Autocomplete } from './autocomplete';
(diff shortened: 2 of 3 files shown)


grantfitzsimmons:
chrome_zXnSxsVap8

Is this a problem with the fix?
https://emoryherbarium-issue-2230.test.specifysystems.org/specify/view/collectionobject/23110/?recordsetid=878

grantfitzsimmons:
chrome_eY4RJQU4Pz

Seems to be.

On v7.6.1

grantfitzsimmons (changes requested):
I cannot select an item in this list

maxxxxxdlp:
The issue is that you have to pick list items with the same value - https://emoryherbarium-issue-2230.test.specifysystems.org/specify/view/picklist/3/

You will also see the error messages in the console `Duplicate picklist entries found` when you open the form, though since the console has a lot of non-error error messages, it is hard to see the real issue.

The form editor will expose all of the form-related errors in the UI

---

Yes, the Specify 7 behavior for picklists with duplicate items changed between 7.6.1 and 7.7, but since this is an edge case, I don't think we should strive to have consistent edge cases. Just tell the user to fix their pick list or wait for a UI that makes these errros easier to spot.

grantfitzsimmons:
Willem is reporting the same issue.

image

Here it is replicated on `issue-2230`.

>I’m a bit mystified as to why preptypeid 142 (DNA Extraction), as well as preptypeid 129 and 130 (Tissue [vouchered] and Tissue [unvouchered]) are displaying as invalid values in my Tissue Collection. Actually, all preptypes in all collections seem to be displaying as invalid.
>
> As I was investigating this, I created a new preparation record in the Tissue Collection. I noticed that the preptype displayed fine when I clicked the Save button. But when I ran a query and reopened the that newly created record, the preptype displayed as ‘invalid’.
>
> And now Sp 7 is giving me an error message and crash report when I try to create a new preparation record in any collection (or at least the Tissue Collection and National Fish Collection, which are both in the Ichthyology Discipline).
> 
> I’m pretty sure that this is a new issue in 773 because I take my preptypes very seriously, and I remember correcting a few that were incorrect (a few weeks ago) – i.e. the preptypeids were from the wrong collection, so I corrected them to ids from the correct collection.
> 
> In the screenshot below you can see that DNA Extraction (id 142) is displaying as invalid. There should only be three preptypes available in this picklist (Tissue [vouchered], Tissue [unvouchered] and DNA Extraction). All the other preptypes you see belong only in National Fish Collection.
>
> image
>
> image
>
>image

maxxxxxdlp:
https://github.com/specify/specify7/pull/2255#issuecomment-1263688041

@grantfitzsimmons Can you test this again after fixing the issue with their picklist - https://github.com/specify/specify7/pull/2255#issuecomment-1267216359

maxxxxxdlp:
@grantfitzsimmons 

https://github.com/specify/specify7/pull/2255#issuecomment-1267497381

I was able to replicate this in 7.7.3, but it worked correctly in 7.7.0

It is caused by https://github.com/specify/specify7/issues/2285 and will be fixed by https://github.com/specify/specify7/pull/2286

chanulee1 (approved):
Did not see api link as value in picklist and picklist items could be selected when values were made unique. (Tested picklist for "Prep Type" field)
db: `emoryherbarium`
user: `spadmin`

grantfitzsimmons:
image

Still think this should be looked at. Does this PR have the changes from #2286 PR?

grantfitzsimmons (approved):

PR #2310, opened 2022-10-10
Fix an infinite fetch loop in FormTable

### LINKED ISSUE #2309: Forms with FormTable records crash browser
Caused by the fix for https://github.com/specify/specify7/issues/2301
To replicate: open an existing record with FormTable. (i.e, a Collection Object with Preparations)
Open the Dev Tools console
Make sure there are continuously erupting error messages in the console (there shouldn't be more than 20, and not continuously popping up)

maxxxxxdlp:
Fixes #2309


grantfitzsimmons (dismissed):
image

is this behavior intentional or another problem? Still seeing lots of errors.

https://willemcatno-issue-2309.test.specifysystems.org/specify/view/loan/2143588469/

maxxxxxdlp:
Unfortunately, those errors are expected (for now)

As long as there are no hundreds of errors, it's ok

Issue #1056, opened 2021-11-09
Attachments: Faulty assumption that all collections use the same storage folder


FedorSteeman:
For many years now, we've had issues with viewing attachments in Specify7 for our major museum's installation (NHMD). 

After some digging, it now turns out a single commit long ago is apparently the cause of our woes: 
https://github.com/specify/specify7/commit/0050c8fd6a9c931a3712595e717a8190acd6e0b8#commitcomment-59690296 

The way our app server is setup for NHMD is that we use multiple folders for each of our collections. This can be seen in the settings.py file for the web asset server, where each collection identifier string points to different folders for the NHMD collections at least. In the same settings file, we made sure to set WEB_ATTACHMENT_COLLECTION to 'None', because, again, we're using multiple folders and not a single one. 

Instead of Specify passing the collection name string in the URL, depending on what collection one is logged on to, it defaults to the collection with the lowest ordinal, which in our case is “NHMD Vertebrate Paleontology”. So for instance, when I’m logged in to Specify under the NHMD Entomology collection, instead of putting “NHMD+Entomology” in the URL, it puts “NHMD+Vertebrate+Paleontology” in the URL, causing not only the thumbnails to break, but any further linkage. 

For example, Specify7 would produce this link for the thumbnail (which won’t work for you, because you’re not logged in): 

https://specify-attachments.science.ku.dk/fileget?coll=NHMD+Vertebrate+Paleontology&amp;type=T&amp;filename=sp63631302825063967879.att.jpg&amp;token=2c13f25f87bb6887e059866b33e0f009%3A1632475049&amp;scale=1024

But it should be the following, since we’re logged in under NHMD Entomology, and this (manually fixed) URL works: 

https://specify-attachments.science.ku.dk/fileget?coll=NHMD+Entomology&amp;type=T&amp;filename=sp63631302825063967879.att.jpg&amp;token=2c13f25f87bb6887e059866b33e0f009%3A1632475049&amp;scale=1024

It’s a mystery to me why Specify7 is no longer set up to use the current collection for constructing the URL, and instead defaults to the collection with the lowest ordinal in the set. Seeing the source code, it's clear that it's doing it, but we don't understand why this was changed from the original code 8 years ago.

FedorSteeman:
This is apparently revamped in v7.6.1, but since we're still stuck with v7.5.0, I added a quick fix to a fork of that version that will do  for the time being.

FedorSteeman:
[...]

PR #2375, opened 2022-10-31
try using logged in collection for accessing asset server.

benanhalt:
for #1056

### DIFF
--- a/specifyweb/attachment_gw/views.py
+++ b/specifyweb/attachment_gw/views.py
@@ -22,11 +22,41 @@ server_time_delta = None
 class AttachmentError(Exception):
     pass
 
-def get_collection():
-    "Assumes that all collections are stored together."
+def get_collection(request=None):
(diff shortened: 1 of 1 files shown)


grantfitzsimmons:
From Fedor:

> Yes! I installed it and everything works perfectly now. Thanks for the assistance!

https://github.com/specify/specify7/issues/1056

maxxxxxdlp:
I wasn't in the meeting where this was discussed, but is this ready for merging?
The code looks good and it works for Fedor
Or does this code include changes that we don't want to apply for everyone?

grantfitzsimmons:
@maxxxxxdlp 

After discussing this, we decided that it was ready for merging. It does include changes that I would encourage us to look at closer for web portal purposes.

If the URL to the attachments remain the same, there is no problem, but just something to consider.

maxxxxxdlp (approved):

PR #2440, opened 2022-11-11
Add hide mapping editor button to Query Builder

### LINKED ISSUE #2070: Add "Hide Mapping Editor" button to the Query Builder
We have that button in the Workbench Mapper right now.
it is very useful on laptop screens to save real estate

melton-jason:
Fixes #2070

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Components.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Components.tsx
@@ -3,6 +3,7 @@ import { useNavigate } from 'react-router-dom';
 
 import { commonText } from '../../localization/common';
 import { queryText } from '../../localization/query';
+import { wbText } from '../../localization/workbench';
 import { f } from '../../utils/functools';
 import type { RA } from '../../utils/types';
 import { Button } from '../Atoms/Button';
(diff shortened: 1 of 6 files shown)


specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx
```diff
@@ -88,6 +90,17 @@ export function QueryHeader({
           </ProtectedTable>
         </ProtectedAction>
       )}
+      {hasToolPermission(
+        'queryBuilder',
+        queryResource.isNew() ? 'create' : 'update'
+      ) && (
```
maxxxxxdlp, line 96: The button should be visible to all users, regardless of permissions, right? Thus, should remove this check for permissions

specifyweb/frontend/js_src/lib/components/QueryBuilder/Components.tsx
```diff
@@ -108,6 +108,29 @@ export function SaveQueryButtons({
   );
 }
 
+export function ToggleMappingViewButton({
```
maxxxxxdlp, line 111: Code is correct, but, need to reformat it with Prettier. Please configure that plugin in your IDE to have it run automatically on save

specifyweb/frontend/js_src/lib/localization/query.tsx
```diff
@@ -287,6 +287,14 @@ export const queryText = createDictionary({
     'en-us': 'Items have been returned',
     'ru-ru': 'Items have been returned',
   },
+  hideMappingEditor: {
```
maxxxxxdlp, line 290: Do we already have these localization strings in the `workbench.tsx` file? If so, you can have query builder use those strings, rather than declaring them again here. (that is okay, because the mapping editor itself is also coming from the workbench files)

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

PR #2442, opened 2022-11-14
Match arbitrary agent

benanhalt:
This branch modifies the Workbench uploader to choose an arbitrary agent when multiple agents are matched. The modification is just hardwired in with no control flag or anything. This was for experimental purposes.


grantfitzsimmons:
This was extraordinarily helpful for my specific use case. I could not have done the conversion without it.

PR #2444, opened 2022-11-14
Occurrence export

benanhalt:
This experimental branch added an API endpoint for accessing the DwCA data for a single occurrence based on the export definitions for a collection, but the data is returned as JSON and is queried from the database in real time. The idea is that the DwCA export that get ingested by an aggregator represents a snapshot of the entire collection. Using this endpoint and the occurrenceId from an aggregator Specify Network or other API client could fetch any updates to that occurrence since the last export.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,5 @@
+[mypy]
+plugins = mypy_django_plugin.main
+    
+[mypy.plugins.django-stubs]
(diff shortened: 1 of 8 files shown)


benanhalt:
This work is probably too stale to be directly merged, but could be the basis of future work.

Issue #2446, opened 2022-11-15
API call to replace agent records by AgentID


acwhite211:
Feature Request:
Users want to replace old agent records with new ones easily without updating all the foreign keys manually.

Solution:
Check to make sure both the old and new agent IDs exist in the table.  Then get all of the columns in all of the tables of specify the are foreign keys referencing AgentID.  Build and execute a query to update all of the records with foreign keys referencing the AgentID.  Finally dedupe by deleting the agent that is being replaced.

PR #2469, opened 2022-11-16
API call to replace agent records by AgentID

acwhite211:
Replaces all the foreign keys referencing the old AgentID with the new AgentID, and deletes the old agent record.  This is just an API change, front end integration will need to be a later issue.  Further changes to this commit might be to change the api to accept json, and to change the static SQL with SQLalchemy.

### DIFF
--- a/specifyweb/specify/api_tests.py
+++ b/specifyweb/specify/api_tests.py
@@ -560,7 +560,7 @@ class ReplaceRecordTests(ApiTests):
             collectingevent=collecting_event
         )
 
-        # Assert that the api rrequest ran successfully
+        # Assert that the api request ran successfully
         response = c.post(
             f'/api/specify/replace/agent/{agent_1.id}/{agent_2.id}/',
             data=[],
--- a/specifyweb/specify/api_tests.py
+++ b/specifyweb/specify/api_tests.py
@@ -530,3 +530,54 @@ class UserApiTests(ApiTests):
             json.loads(response.content),
             {'AgentInUseException': [self.agent.id]}
         )
+
+class ReplaceRecordTests(ApiTests):
+    def test_replace_agents(self):
+        c = Client()
+        c.force_login(self.specifyuser)
+
+        # Create agents and a collector relationship
+        agent_1 = models.Agent.objects.create(
+            id=7,
+            agenttype=0,
+            firstname="agent",
+            lastname="007",
+            specifyuser=None)
+        agent_2 = models.Agent.objects.create(
+            id=6,
+            agenttype=0,
+            firstname="agent",
+            lastname="006",
+            specifyuser=None)
+        collecting_event = models.Collectingevent.objects.create(
+            discipline=self.discipline
+        )
+        collector = models.Collector.objects.create(
+            id=7,
+            isprimary=True,
+            ordernumber=1,
+            agent=agent_1,
+            collectingevent=collecting_event
+        )
+
+        # Assert that the api rrequest ran successfully
+        response = c.post(
+            f'/api/specify/replace/agent/{agent_1.id}/{agent_2.id}/',
+            data=[],
+            content_type='text/plain')
+        self.assertEqual(response.status_code, 204)
+
+        # Assert that the collector relationship was updated correctly to the new agent
+        collector = models.Collector.objects.get(id=agent_1.id)
+        self.assertEqual(collector.agent.id, 6)
+
+        # Assert that the old agent was deleted
+        with self.assertRaises(models.Agent.DoesNotExist):
+            models.Agent.objects.get(id=7)
+
+        # Assert that a new api request will not find the old agent
+        response = c.post(
+            f'/api/specify/replace/agent/{agent_1.id}/{agent_2.id}/',
+            data=[],
+            content_type='text/plain')
+        self.assertEqual(response.status_code, 404)
--- a/specifyweb/specify/views.py
+++ b/specifyweb/specify/views.py
@@ -5,6 +5,7 @@ A few non-business data resource end points
 import json
 import mimetypes
 from functools import wraps
+from itertools import groupby
 
 from django import http
 from django.conf import settings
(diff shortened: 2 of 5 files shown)


specifyweb/specify/views.py
```diff
@@ -391,3 +392,84 @@ def set_admin_status(request, userid):
...
+        # Build query to update all of the records with foreign keys referencing the AgentID
+        sql_update = ""
+        for table_name, column_names in groupby(foreign_key_cols, lambda x: x[0]):
+            for col in [c[1] for c in column_names]:
+                sql_set_clause = col + " = " + new_agent_id
+                sql_where_clause = col + " = " + old_agent_id
+                sql_update += "UPDATE " + table_name + " SET " + sql_set_clause + " WHERE " + sql_where_clause + ";\n"
+        
+        # Execute update query for agent children
+        cursor.execute(sql_update)
```
maxxxxxdlp, line 470: Could this fail due to business rule errors? @benanhalt do you know how to catch that?
Should return some sort of 4xx error to the front-end in that case

specifyweb/specify/views.py
```diff
@@ -391,3 +392,84 @@ def set_admin_status(request, userid):
...
+                        },
+                        'required': ['old_agent_id', 'new_agent_id'],
+                        'additionalProperties': False
+                    }
+                }
+            }
+        },
+        "responses": {
+            "204": {"description": "Success",},
+            "400": {"description": "The AgentID specified does not exist."}
```
maxxxxxdlp, line 424: Would it make sense to return 404 on "does not exist", instead of 400?

specifyweb/specify/views.py
```diff
@@ -391,3 +392,84 @@ def set_admin_status(request, userid):
...
+        }
+    },
+})
+@login_maybe_required
+@require_POST
+def agent_record_replacement(request: http.HttpRequest, old_agent_id, new_agent_id: int) -> http.HttpResponse:
+    """Replaces all the foreign keys referencing the old AgentID
+    with the new AgentID, and deletes the old agent record.
+    """
+    check_permission_targets(request.specify_collection.id, request.specify_user.id, [])
```
maxxxxxdlp, line 434: Need to integrate this endpoint with the permission system (take a look at other endpoints to see how that works).
Check for Agent Delete permission. On top of that, could probably add a special permission for this endpoint (as this endpoint modifies many tables, it would be considered a bit sensitive)

maxxxxxdlp (changes requested):

specifyweb/specify/views.py
```diff
@@ -391,3 +392,84 @@ def set_admin_status(request, userid):
...
+
+    # Create database connection cursor
+    cursor = connection.cursor()
+
+    with transaction.atomic():
+        # Check to make sure both the old and new agent IDs exist in the table
+        cursor.execute("SELECT * FROM specify.agent WHERE AgentID = %s;", [old_agent_id])
```
benanhalt, line 441: I would make these SELECT FOR UPDATE statements. In the default isolation level another transaction can delete the rows after you have selected them. So in a race condition you could find the agents exist here, but they are gone by the time you try to use them. FOR UPDATE locks the rows until your transaction completes.

I would also consider using the Django ORM for these checks. It could make the intent a little clearer. Something like 
```
if not models.Agent.objects.filter(id=old_agent_id).select_for_update().exists():
   ...
```

I think `select_for_update` and `exists` can be combined in that manner. It's a matter of taste though, and you should use whichever you prefer since the function is already using raw SQL.

specifyweb/specify/views.py
```diff
@@ -391,3 +392,84 @@ def set_admin_status(request, userid):
...
+        cursor.execute("SELECT * FROM specify.agent WHERE AgentID = %s;", [new_agent_id])
+        if len(cursor.fetchall()) == 0:
+            return http.HttpResponseBadRequest("AgentID: " + new_agent_id + " does not exist.")
+
+        # Get all of the columns in all of the tables of specify the are foreign keys referencing AgentID
+        sql_get_cols_ref_agent_id = """
+        SELECT TABLE_NAME, COLUMN_NAME
+        FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE
+        WHERE
+        REFERENCED_TABLE_SCHEMA = 'specify' AND
```
benanhalt, line 453: Here there `REFERENCED_TABLE_SCHEMA` should be the name of the database that Specify 7 is using. It won't be "specify" in general. You could get it from the django.conf.settings DATABASE_NAME variable. Or there might be a preferred way to get it in Django. I would google something like "get current database name in Django".

specifyweb/specify/views.py
benanhalt: Yes. BRs that are enforced by MySQL could occur here. They will probably appear as django.db.IntegrityError or some such. I would approach this by starting to write some tests to see what can happen. You can take a look at https://github.com/specify/specify7/blob/production/specifyweb/specify/api_tests.py#L514-L532 for an example where I was testing an API using the Django "test client" to simulate requests.

One case where I know you can get such an exception to occur is if a CollectingEvent has two different Collectors that get merged to the same agent. You could write a test to set up such a situation and see what exception is produced and what information can be gleaned from it; hopefully at least which table was involved.

benanhalt:
Looking good so far! 

At this point, the database name thing is the only thing that I think would stop it from working. But I think some tests need to be added, and the race condition resolved. Also as Max pointed out, there will need to be permissions added for this action as well, but I think it is reasonable to save that until everything else is sorted.

specifyweb/specify/views.py
```diff
@@ -421,7 +421,7 @@ def set_admin_status(request, userid):
         },
         "responses": {
             "204": {"description": "Success",},
-            "400": {"description": "The AgentID specified does not exist."}
+            "404": {"description": "The AgentID specified does not exist."}
```
maxxxxxdlp, line 424: Thanks!

You also need to change `http.HttpResponseBadRequest` to `http.HttpResponseNotFound` for the endpoint to actually return 404 http codes

specifyweb/specify/api_tests.py
```diff
@@ -530,3 +530,54 @@ def test_set_user_agents_in_use_exception(self):
...
+        )
+        collector = models.Collector.objects.create(
+            id=7,
+            isprimary=True,
+            ordernumber=1,
+            agent=agent_1,
+            collectingevent=collecting_event
+        )
+
+        # Assert that the api rrequest ran successfully
```
maxxxxxdlp, line 563: Typo `rrequest`

specifyweb/specify/views.py
maxxxxxdlp: Looks to be fixed in [d686310](https://github.com/specify/specify7/pull/2469/commits/d686310e857a224a3a62745ca9df836381829a84)

maxxxxxdlp (approved):

maxxxxxdlp:
Fixes #2446

Issue #2428, opened 2022-11-10
Modified By date not displaying on the form


grantfitzsimmons:
![image](https://user-images.githubusercontent.com/37256050/201134960-87cf757a-c777-4609-a3c1-ffd4e4a100c8.png)

left: `testability`
right: `edge`

https://umich-testability.test.specifysystems.org/specify/view/collectionobject/579/
https://umich-edge.test.specifysystems.org/specify/view/collectionobject/579/

grantfitzsimmons:
<img width="2508" alt="image" src="https://user-images.githubusercontent.com/37256050/201966492-b8f070c6-479f-4086-b2a0-448de0e19f20.png">

Still an issue

melton-jason:
I'm pretty sure this is the same underlying issue as #2438. I will close this in favor of #2438 as #2438 covers more use cases.
Reopen if the issue turns out to be separate

Issue #2415, opened 2022-11-06
Form repeadedly believes it is edited after a change


grantfitzsimmons:
https://user-images.githubusercontent.com/37256050/200190150-ccb6b264-08bd-4bb5-9937-b2be114fefa4.mp4

**Steps to Recreate:**

1. Run a query on the CO table
2. Open a record in the form view
3. Make a change to a field
4. Click save
5. See that you can save infinitely

Not reproducible on `edge`. user is `peteniel`

https://gothenberg-testability.test.specifysystems.org/specify/query/1/

browser: chrome on macOS

maxxxxxdlp:
Caused by a bug where the date picker overwrites a read-only field (timestamp modified). It overwrites it because back-end saves date in a different format than the front-end

To test: check that the bug does not happen anymore in this particular case

chanulee1:
Seems to have been fixed.
db: `gothenberg`, `kufishnotests`

PR #2478, opened 2022-11-18
Fix placeholder date format showing for readonly date fields

### LINKED ISSUE #2438: [testability] Date fields do not show date in Read-Only Mode
**Describe the bug**
Date fields do not display their values when a form is in Read-Only Mode.

https://kufishnotests-testability.test.specifysystems.org/specify/view/collectionobject/6/?recordsetid=44
Original:
image

Read-Only Mode:
image

Subview also not displaying date correctly:

CE subview on CO form:
image

CE form:
image

melton-jason:
Fixes #2438 (see also #2428)
Ultimately, Read-only date fields are not being assigned  a value in their value element in the HTML., which is defaulting to showing the placeholder format instead of the actual date. 

I believe the issue arises because where it was normally set, it is returning (see the first commit). 
Another potential fix would be to manually set the value of the value HTML element within the <Input.Generic> (see second commit)

P.S.
In regards to the first commit, that change was implemented within [fcf850766e](https://github.com/specify/specify7/commit/fcf850766e)

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/PartialDateUi.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/PartialDateUi.tsx
@@ -249,7 +249,7 @@ export function PartialDateUi<SCHEMA extends AnySchema>({
   ]);
 
   React.useEffect(() => {
-    if (isReadOnly) return;
+    // if (isReadOnly) return;
     /*
      * If resource changes, a new moment is set, but its value won't get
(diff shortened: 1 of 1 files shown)


maxxxxxdlp:
When testing this fix, please also test https://github.com/specify/specify7/issues/2415

chanulee1 (approved):
Date fields work in Read-Only Mode; #2415 also seems to have been fixed on this branch.
db: `kufishnotests`, `gothenberg`

PR #2674, opened 2022-12-15
Converts the top menu into a side menu

### LINKED ISSUE #1232: [a11y] Cannot distinguish if cells are read-only between views
**Describe the bug**
The background colors and read-only indicators are different between forms and pop-outs.

**To Reproduce**
1. View a CE record from the CO form
2. View the same CE record on its own page

**Screenshots**
Full form view:
image

Subview:
image

The gray background is an editable field in the subview pop-out:
image

The white background is an editable field on the form:
image

maxxxxxdlp:
Fixes #1232

### DIFF
--- a/specifyweb/frontend/js_src/css/base.css
+++ b/specifyweb/frontend/js_src/css/base.css
@@ -13,7 +13,6 @@ body {
     overflow: hidden;
     display: flex;
     row-gap: 1rem;
-    flex-direction: column;
 }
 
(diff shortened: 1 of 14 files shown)


maxxxxxdlp:
This was just for testing

maxxxxxdlp:
For reference, this feature was implemented for real in https://github.com/specify/specify7/pull/2725

PR #2676, opened 2022-12-15
Allow right side of HostTaxon to be from any tree

### LINKED ISSUE #2675: Allow right side of HostTaxon to be from any taxon tree
Currently, HostTaxon request explicitly specifying the Collection Relationship name.

Need to add support for omitting relationship name - which should be interpreted as right side being from any collection

maxpatiiuk:
Fixes #2675


maxxxxxdlp:
Can't merge it yet becase of https://github.com/specify/specify7/issues/2675#issuecomment-1367487823

PR #2687, opened 2022-12-16
Adds a border to the WebLink button & ColRel plugin

grantfitzsimmons:
Makes it consistent with the rest of the form. No buttons or plugins left behind!

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/CollectionRelOneToMany.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/CollectionRelOneToMany.tsx
@@ -68,7 +68,9 @@ export function CollectionOneToManyPlugin({
   const loading = React.useContext(LoadingContext);
   const navigate = useNavigate();
   return data === false ? null : (
-    <div className="w-fit rounded bg-[color:var(--form-background)] p-2">
+    <div className="w-fit rounded bg-[color:var(--form-background)] p-2
+    ring-1 ring-gray-500 dark:ring-0 disabled:ring-gray-500 disabled:dark:ring-neutral-500
+    ">
       <table className="grid-table grid-cols-[repeat(3,auto)] gap-2">
         <thead>
           <tr>
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/CollectionRelOneToMany.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/CollectionRelOneToMany.tsx
@@ -71,7 +71,7 @@ export function CollectionOneToManyPlugin({
     <div
       className={`
         w-fit rounded bg-[color:var(--form-background)] p-2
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/FormPlugins/CollectionRelOneToMany.tsx
```diff
@@ -68,7 +68,9 @@ export function CollectionOneToManyPlugin({
   const loading = React.useContext(LoadingContext);
   const navigate = useNavigate();
   return data === false ? null : (
-    <div className="w-fit rounded bg-[color:var(--form-background)] p-2">
+    <div className="w-fit rounded bg-[color:var(--form-background)] p-2
+    ring-1 ring-gray-500 dark:ring-0 disabled:ring-gray-500 disabled:dark:ring-neutral-500
+    ">
```
maxxxxxdlp, line 73: CSS looks good. As far as code, could you format the className like it is here:
https://github.com/specify/specify7/blob/fc440840baae3b82c3ff09631971079c11bc8476/specifyweb/frontend/js_src/lib/components/Core/SplashScreen.tsx#L12-L15

Notice the use of backticks and `{`, `}` instead of quotes
JavaScript does not natively support using `"` for multi-line strings, thus some tools fail to parse it

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/FormPlugins/CollectionRelOneToMany.tsx
grantfitzsimmons: Fixed in [53d7500](https://github.com/specify/specify7/pull/2687/commits/53d75006a049398467c5c2eaca6b889a8b358f7c)

maxxxxxdlp: Almost there. I had to make just one slight change to it.

For consistency, you have to format multiline class names like this - https://github.com/specify/specify7/pull/2687/commits/3ed96715948ecf39cd439374371a1a8535f7e000

This might look like wasting a lot of space, but wasted space doesn't matter in a code (as text files are small). But more importantly, this formatting allows editing and copying class names easier (i.e, since the quotes are on a different line, you won't accidentally erase/override a quote)

maxxxxxdlp:
It looks like you copied the `ring` styles from the input fields. This includes ring styles for when the element is disabled. While inputs could be disabled, thus still style made sense, you applied these styles to a `<div>` element, which doesn't really have a notion of being "disabled" (div is just a `divison` - i.e a container for elements)

Made a small change:
https://github.com/specify/specify7/pull/2687/commits/11fa623774462a238aa47af8c54ef4d3127c6631

maxxxxxdlp (approved):

grantfitzsimmons:
@maxxxxxdlp Appreciate the help! I was thinking about that when using those classes– didn't know if there was some kind of read-only view of that component if the permissions restricted access in another collection.

PR #2688, opened 2022-12-17
Allow for integration with Localization frameworks

maxxxxxdlp:
[Link to Weblate](https://hosted.weblate.org/projects/specify-7/)

- [x] Migrate to typesafe-i18n library
- [x] Pick a localization UI
- [x] Integrate with it
- [ ] Currently, Weblate is listening for strings on the `l10n-improvements` branch. Need to make it look at `production` instead. (in `test.yml` and `push.yml`)
- [x] Configure a webhook to automatically update weblate on localization changes
- [x] Configure automated translation. Try it out on Spanish and French languages.
- [x] Extend this documentation and adapt it for non-developers: https://github.com/specify/specify7/blob/l10n-improvements/specifyweb/frontend/js_src/lib/localization/README.md#syntax-for-string-parameters-and-plurals

To test:
- Open a data set that was created on a different branch that has validation errors and make sure validation messages are displayed properly
- Trigger any number of the following validation errors and make sure the error message is formatted correctly:
  - on invalid boolean: `value "{value:string}" not resolvable to True or False`
  - on invalid decimal: `value "{value:string}" is not a valid decimal value`
  - on invalid value of read-only pick list:
    ```
    "{value:string}" is not a legal value in this picklist field.

    Click on the arrow to choose among available options.
    ```
  - on invalid agent type: `bad agent type: "{badType:string}". Expected one of {validTypes:string}`
  - on skipped required tree rank: `Missing or unmapped required tree parent rank value for "{names:string}".`
  - on pick list value too long: `value from picklist {pickList:string} longer than the max of {maxLength:number|formatted} for field`
  - on field value too long: `value must not have length greater than {maxLength:number|formatted}`
  - on date with invalid year: `date value must contain four digit year: {value:string}`
  - on date with incorrect format: `bad date value: {value:string}. expected: {format:string}`
  - on coordinate in a wrong format: `bad latitude or longitude value: {value:string}`
  - on latitude out of range: `latitude must be between -90 and 90. actual: {value:string}`
  - on longitude out of range: `longitude must be between -180 and 180. actual: {value:string}`
  - on tree node with empty Name, but non empty other column (i.e, Taxon Author) - `this field must be empty if {column:string} is empty`
  - on missing required field: `field is required by upload plan mapping`
  - Create two nodes in the taxon tree with the name `Uploaded`. Both should be at the root level (have to do this though database). WB data set validation should print: `There are multiple "Uploaded" placeholder values in the tree!`
- Go though all front-end pages and make sure localization strings are rendered like expected. Only need to do this for one language. This is just to check for typos. Things to keep a lookout for includes cases when dynamic localization string is printed with something like `{value|string` (i.e, the value wasn't substituted)
- Modify any localization string in Specify 7 on the `l10n-improvements` branch. Should be able to see it in `Weblate` after less than 10 minutes (link: https://hosted.weblate.org/projects/specify-7/)
- Modify an localization string in Weblate (link: https://hosted.weblate.org/projects/specify-7/). In 3 hours after modification, the string should be modified in Specify 7's source code on the `l10n-improvements` branch. If it requires approval to modify a string, message me and I will give you edit permissions

##### Pros:
- If developers change the English translation in the code, Weblate will mark all other translations of that string as in need of review
- Weblate checks for inconsistent punctuation and a [hundred other things](https://docs.weblate.org/en/latest/user/checks.html#check-same) - really cool
- We are hosted for free, but the platform is open source, so if we are not satisfied with the performance we can always move to self-hosting
- Provides stats and audit log. Has a simple permission system. Translators can subscribe for notifications and receive emails when languages that they are familiar with are in need of translation again.
- If I setup it correctly, anyone from the outside could give us suggestions/comments/propose edits. People whom we explicitly approved could get their changes merged into production automatically!
 -Supports automated translation using Google Translate and 20 other localization services. Though, we need to provide our own API key. Fortunately, we quality under Google Translate's free tier.
- A really fancy feature: you can upload screenshots of the application, it will do OCR, identify all the localization strings used on that component and link them to the screenshot so that translators can see context when localizing! Additionally, I setup it up to tell you which places in the code the string is used in and how many times it is used.
- Provides interesting and in-depth stats. See screenshots

##### Cons:
- The documentation is not great. I had to read the source code to understand how it works
- It supports only some file formats and does not provide easy integration with other formats. I had to do some trickery to make it work (using a fake [weblate-localization](https://github.com/specify/specify7/tree/weblate-localization) branch and two GitHub actions that do a two-way sync and file format translation)
- Documentation mentions support for translations coming from all branches, but they don't describe it in a lot of detail so I am still not sure how that would work. For now, Weblate would work with the `production` branch only. That should be sufficient for the time being as we are moving to a more agile dev strategy.

image

image

image


grantfitzsimmons (changes requested):
See my comment on the error in the WB

maxxxxxdlp:
The bug is because back end is now sending validation errors in a different format. (instead of sending the message as a string, it sends the localization key and an arguments for the string).

I though I made it backwards compatible, but apparently not completely. I can fix this so that data-sets validated before this change still work after this change. However, it won't work the other way: data-sets created after this change won't open properly in older versions of sp7 (unless you run the validation again)

@grantfitzsimmons is this fine? If not, I can revert the back-end validation response format change. The drawback of that would be that workbench validation messages won't be localizable though Weblate

grantfitzsimmons:
I am not aware of any institution using two versions of Specify 7 independently for any reason _yet_. I think that data sets validated in the current version not being backwards compatible is fine, but past ones need to be visible in the new version.

maxxxxxdlp:
Fixed. Should be compatible with data sets validated before this change

Screenshot 2022-12-29 at 10 16 22

grantfitzsimmons:
image

`Metadata` is showing as `Remarks`

grantfitzsimmons:
image

left is `edge`, right is `l10n-improvements`

* The name of the import file is missing
* Metadata is two words
* Other misc differences that are not good/bad in my opinion

grantfitzsimmons:
This PR:

image

`edge`:
image

grantfitzsimmons:
This PR and `edge`:
image

Maybe this should be something like "invalid integer" instead...

maxxxxxdlp:
> image
> 
> `Metadata` is showing as `Remarks`

I changed the front-end to use the label from the `Workbench.Remarks` field rather than have a separate `Metadata` localization string. Can revert this. My reasoning was that eventually we would also have nice localization solution for the schema field labels, so it's best not to duplicate field labels whenever possible.

Another unpleasant side effect of this change is that if schema language and UI language differ, button labels would be in different languages - though I am not sure how common that would be, especially once we make it easier to localize the UI and schema

maxxxxxdlp:
> > image
> > `Metadata` is showing as `Remarks`
> 
> I changed the front-end to use the label from the `Workbench.Remarks` field rather than have a separate `Metadata` localization string. Can revert this. My reasoning was that eventually we would also have nice localization solution for the schema field labels, so it's best not to duplicate field labels whenever possible.
> 
> Another unpleasant side effect of this change is that if schema language and UI language differ, button labels would be in different languages - though I am not sure how common that would be, especially once we make it easier to localize the UI and schema

Nevermind half of what I said. The bug is that I am using the label for the `Remarks` field instead of `Meta Data` - fixed. The other things I mentioned still apply

maxxxxxdlp:
> image
> 
> left is `edge`, right is `l10n-improvements`
> 
> * The name of the import file is missing
> * Metadata is two words
> * Other misc differences that are not good/bad in my opinion

Fixed missing values for the `Imported` and `File Name` labels.
Other misc differents are caused by the transition from using UI localization strings to schema localization whenever possible. (`Meta Data` is two words because apparently it is that way in the schema localization)

maxxxxxdlp:
> This PR:
> 
> image
> 
> "actual" should be capitalized
> 
> `edge`: image

Fixed

maxxxxxdlp:
> This PR and `edge`: image
> 
> Maybe this should be something like "invalid integer" instead...

Fixed. It was literally printing out the Python exception message. Added a separate string. When testing. please test the case with decimal field and float field (they have different error messages)

grantfitzsimmons:
For the record, all references to `meta data` are incorrect and should be `metadata` or `meta-data`

grantfitzsimmons:
I feel this is ready to merge if you are satisfied @maxxxxxdlp

grantfitzsimmons (approved):
Feel free to review or ask another person to take a look.

maxxxxxdlp:
🎉

Issue #2518, opened 2022-11-30
[testability] Saving changes to Agent record with certain fields filled out results in error


chanulee1:
Attempting to save changes to an agent record with agent type "Other" and an associated address record results in error. Error logs say address cannot be found, but a query showed that the address record was created.

https://user-images.githubusercontent.com/103299204/204930222-e9f8b782-4ab2-4e33-b572-d5d30a86125e.mp4

https://kufishnotests-testability.test.specifysystems.org/specify/view/agent/4443/

[Specify 7 Crash Report - 2022-11-30T23_32_24.663Z.txt](https://github.com/specify/specify7/files/10128092/Specify.7.Crash.Report.-.2022-11-30T23_32_24.663Z.txt)

Console:
<img alt="image" width="500px" src="https://user-images.githubusercontent.com/103299204/204930715-180da789-c733-4860-81a2-18b9e4e6fc01.png"/>

realVinayak:
I can replicate the error on my local instance. The problem seems to be the incorrect address id. It looks like making a request to `/api/specify/address/{id}/` is giving an error.  In my case, frontend is trying to access `addressid=776` in `address table`. However, when I check the address table using SQL, there is no `addressid 776`. That is, there is no address in the address table with id=776.

melton-jason:
Some additional context:
The error only occurs on PUT requests (i.e. updating an already existing Agent) and only for agent types `Other` and `Group`: seemingly behaves as intended otherwise.  

@realVinayak where are you seeing this request? On a local instance of Specify7 (and through the API for `kufishnotests`), I was able to find the PUT request, and the addressID in the request correctly corresponds to the address it _should_ be adding/updating (and yet, the error is still being thrown).

maxxxxxdlp:
Potentially related to https://github.com/specify/specify7/issues/1702 or https://github.com/specify/specify7/issues/929

@realVinayak @melton-jason this is a concerning issue.

Also, see how the error message in the video includes `[object Object]` - that's a separate bug - should be fixed to.

Let me know if you have questions, need help figuring this out

grantfitzsimmons:
https://herbrbge6802-testability.test.specifysystems.org/specify/query/new/agent/

I was able to recreate this on `testability`

```
Error: [object Object]
    at i (https://herbrbge6802-testability.test.specifysystems.org/static/js/main.89a990d7a6b36380deab.bundle.js:2:84452)
[...]

PR #2702, opened 2022-12-22
Improve error response from backend

### LINKED ISSUE #2668: Report limited exceptions in production
Server errors are not reported at all. Back-end just says "500 error occurred"

The reasoning was that the error overlay when in debug mode includes variable values, and thus could include plaintext passwords

Still, a better compromise would be to display the exception itself but not the variable values. This would be of insane help when debugging errors submitted by users

melton-jason:
Fixes #2668 
Preview:

error_message


maxxxxxdlp:
This looks awesome!
Please note that Ben began work on improving business rule exceptions - https://github.com/specify/specify7/pull/2219
On that branch, back-end sends business rule exceptions in a structured form, which can them be interpreted by the front-end and properly attributed to a particular field/relationship.

Thus, you should make your changes on top of that branch - you should make all other 500 errors be formatted more user friendly (but only when DEBUG=false - in debug mode, we want to see the full information)

melton-jason:
I structured all of the business rule improvements based off of his changes. The only data not preserved from his branch and mine  is the 'type' of business rule, which I can easily include if needed. All of the data would be stored under the `Data` details tag.
Do you want a separate formatting in the front end for BusinessRuleExceptions? 

While I have not tested it yet, the backend should only send the response when `DEBUG` mode is off. When Debug is on, the behavior should (theoretically) be the same as before. 

Also, I assume permission exceptions are handled already? There seems to already be a middleware system in place that seems pretty complex, and I do not want to touch it if it is already done.

maxxxxxdlp:
Yes, permission errors are handled well by the front-end.
Besides business rules, we also need to better handle random 500 errors that can occur anywhere (i.e, MySQL exceptions). At the very least, back-end should send an error message ot the front-end

melton-jason:
The error dialog themselves are mostly done.
Attached below is a preview of what to expect. At a bare minimum (as seen in general exceptions), each error dialog message contains the exception that was raised, the exception message (if applicable), and the whole stack trace.

As @maxxxxxdlp and I discussed, these error dialogs will be a LAST RESORT. The next step in this project is to separate out the errors that can be resolved by form changes and instead of showing the dialog, raise a form error that points the user to the affected field(s) so they can make the required changes and hopefully not encounter the error dialog.

Let me know if there are any changes to these error dialogs you would like to see. 

### General Exceptions:
Everything that is not a Specify-specific exception will belong in this group- MySQL, Django, miscellaneous Python exceptions, etc.

general_message


### BusinessRule Exceptions
In addition to everything a general exception has, a BusinessRule Exception contains the affected table's icon, as well as a 'type' that is displayed before the exception message.

business_rule_exception

maxxxxxdlp:
Is this ready for a code review or not yet?

melton-jason:
Feel free to do a code review! The earlier we can iron out problems the better. 

>  The next step in this project is to separate out the errors that can be resolved by form changes and instead of showing the dialog, raise a form error that points the user to the affected field(s) so they can make the required changes and hopefully not encounter the error dialog.

This is not implemented yet, but if there are changes that need to be done on the current system, that will alleviate some headaches in the future, 
I know the frontend code will require changes to make it more accessible and dynamic

grantfitzsimmons:
This actually gets me giddy

grantfitzsimmons:
I'm so excited. How can I help test this @melton-jason ?

specifyweb/businessrules/accessionagent_rules.py
```diff
@@ -5,4 +5,7 @@
 @orm_signal_handler('pre_save', 'Accessionagent')
 def agent_division_must_not_be_null(accessionagent):
     if accessionagent.agent_id is None:
-        raise BusinessRuleException(_("AccessionAgent -> Agent relationship is required."))
```
maxxxxxdlp, line 8: Not required for now, but we are looking into getting rid of back-end localization in favor of having just front-end localization. That way, back end just sends the string key and arguments, whereas the front-end resolves that to a localized string.

Benefits:
- instead of having two localization systems, you have only one
- minor, but back-end responses may be smaller

I began doing that on `l10n-improvements`. After the branches are merged, you can look into doing this if you are interested.

specifyweb/businessrules/agent_rules.py
```diff
@@ -6,20 +6,41 @@
 def agent_delete_blocked_by_related_specifyuser(agent):
     try:
         Specifyuser.objects.get(agents=agent)
+        user = Specifyuser.objects.get(agents=agent)
     except Specifyuser.DoesNotExist:
         return
-    raise BusinessRuleException("agent cannot be deleted while associated with a specifyuser")
+    raise BusinessRuleException(
```
maxxxxxdlp, line 12: (optional) Is there a way to get rid of this business rule in favor of adding this information to the 
response from `/api/delete_blockers/`?

Since front-end already has a nice UI for delete blockers, doing it though that will save us some work

specifyweb/businessrules/agent_rules.py
```diff
@@ -6,20 +6,41 @@
...
 
 @orm_signal_handler('pre_save', 'Agent')
 def agent_types_other_and_group_do_not_have_addresses(agent):
     from specifyweb.specify.agent_types import agent_types
     if agent.agenttype is None:
-        raise BusinessRuleException("agenttype cannot be null")
-    if agent_types[agent.agenttype] in ('Other', 'Group'):
-        agent.addresses.all().delete()
+        raise BusinessRuleException(
+            "agenttype cannot be null",
```
maxxxxxdlp, line 34: Can you check if this is always enforced on the front-end? If so, we won't need special handling for this case on the front-end

specifyweb/businessrules/groupperson_rules.py
```diff
@@ -6,7 +6,11 @@
 @orm_signal_handler('pre_save', 'Groupperson')
 def agent_cannot_be_in_self(groupperson):
     if groupperson.member_id == groupperson.group_id:
-        raise BusinessRuleException('a group cannot be made a member of itself')
+        raise BusinessRuleException(
+            'a group cannot be made a member of itself', 
+            {"table" : "GroupPerson",
```
maxxxxxdlp, line 11: Whenever possible, add a `fieldName` to the response payload. That way front-end can attribute error to a single field. In cases when fieldName is missing, front-end will display a dialog

specifyweb/businessrules/interaction_rules.py
```diff
@@ -34,20 +34,39 @@ def loanprep_quantity_must_be_lte_availability(ipreparation):
         quantity = ipreparation.quantity or 0
         quantityresolved = ipreparation.quantityresolved or 0
         if available < (quantity - quantityresolved):
-            raise BusinessRuleException(f"loan preparation quantity exceeds availability ({ipreparation.id}: {quantity - quantityresolved} {available})")
+            raise BusinessRuleException(
```
maxxxxxdlp, line 37: A big ask, but can you check if front-end already has all of these business rules?

Handling business rules on the front-end has benefits:
- Faster error feedback
- We know exactly which field the error related to

Of course, we still need to have back-end business rules for direct API calls and for workbench

specifyweb/frontend/js_src/lib/components/Errors/FormatError.tsx
```diff
@@ -111,8 +112,7 @@ export function formatError(
 /** Format error message as JSON, HTML or plain text */
 function formatErrorResponse(error: string): JSX.Element {
   try {
-    const json = JSON.parse(error);
-    return <pre>{jsonStringify(json, 2)}</pre>;
+    return formatJSONBackendResponse(error);
```
maxxxxxdlp, line 115: Could you rename this to `formatJsonBackendResponse` to follow the strictCamelCase convention? Don't want to do what Internet Explorer did with `XMLHttpRequest`

specifyweb/frontend/js_src/lib/components/Errors/JSONError.tsx
```diff
@@ -0,0 +1,129 @@
+import React from 'react';
```
maxxxxxdlp, line 1: Rename file to `JsonError`

specifyweb/frontend/js_src/lib/components/Errors/JSONError.tsx
```diff
@@ -0,0 +1,129 @@
+import React from 'react';
+import { jsonStringify } from '../../utils/utils';
+import { className } from '../Atoms/className';
+import { TableIcon } from '../Molecules/TableIcon';
+
+type JSONResponse = {
```
maxxxxxdlp, line 6: Rename to `JsonResponse`

specifyweb/frontend/js_src/lib/components/Errors/JSONError.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+type JSONResponse = {
+  exception: string;
+  message: string;
+  data: any;
+  formattedData: string;
+  traceback: string;
+};
+
+function createJsonResponse(error: string): JSONResponse {
```
maxxxxxdlp, line 14: One of the reasons we follow strictCamelCase convention is that it's unambiguous and consistent. In this case, you had `JSON` and `Json` in the same file. Make sure all usages in the file are `Json` instead of `JSON`

Another reason for this: https://medium.com/@thomasreggi/enforcing-strict-camelcase-5332c53a7484

specifyweb/frontend/js_src/lib/components/Errors/JSONError.tsx
```diff
@@ -0,0 +1,129 @@
...
+export function formatJSONBackendResponse(error: string): JSX.Element {
+  const response = createJsonResponse(error);
+  if (response.exception == 'BusinessRuleException')
+    return formatBusinessRuleException(error);
+  else if (response.exception == 'TreeBusinessRuleException')
+    return formatTreeBusinessRuleException(error);
+  else return formatBasicResponse(error);
+}
+
+function JSONBackendResponseFooter({
```
maxxxxxdlp, line 35: You could also configure ESLint to report naming convention errors for you - https://github.com/specify/specify7/tree/production/specifyweb/frontend/js_src#plugins

specifyweb/frontend/js_src/lib/components/Errors/JSONError.tsx
```diff
@@ -0,0 +1,129 @@
...
+  response,
+  hasData = true,
+  isDataOpen = true,
+}: {
+  readonly response: JSONResponse;
+  readonly hasData?: boolean;
+  isDataOpen?: boolean;
```
maxxxxxdlp, line 42: Forgot `readonly`

specifyweb/frontend/js_src/lib/components/Errors/JSONError.tsx
```diff
@@ -0,0 +1,129 @@
...
+}: {
+  readonly table: string;
+  readonly response: JSONResponse;
+}): JSX.Element {
+  const type: string = response.data.type;
+  return (
+    <>
+      <div className="flex">
+        <TableIcon name={table} label={false}></TableIcon>
+        <span>{'  '}</span>
```
maxxxxxdlp, line 72: What is this used for? Can you accomplish this with CSS? Or tailwind class names?

specifyweb/frontend/js_src/lib/components/Errors/JSONError.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+function formatBusinessRuleException(error: string): JSX.Element {
+  const response = createJsonResponse(error);
+  const table: string = response.data.table;
+  return (
+    <>
+      <BusinessRuleExceptionHeader
+        table={table}
+        response={response}
+      ></BusinessRuleExceptionHeader>
```
maxxxxxdlp, line 106: Use shorthand syntax. Replace `></BusinessRuleExceptionHeader>` with `/>` (here and in other places

maxxxxxdlp: Once you configure ESLint, you would be able to just press the keyboard shortcut for `ESLint: fix all errors` and it will fix small things like this for you

maxxxxxdlp (changes requested):
Looks quite good. Good job! Most suggestions are regarding extending this further, but we can merge it as is and extend it in the future

maxxxxxdlp:
You could also try triggering a business rule

specifyweb/businessrules/agent_rules.py
melton-jason: Added Agent.specifyuser to deletion_blockers in 56cb36f.

specifyweb/businessrules/agent_rules.py
melton-jason: Yes, looks like this is already handled on the front-end!

specifyweb/businessrules/groupperson_rules.py
melton-jason: Added in 93b1277. 
I did not yet add 'fieldName' to TreeBusinessRule Exceptions. Most of the time the error can not be easily distilled to a single field. We can probably separate the frontend implementation of TreeBusinessRule Exceptions from BusinessRule Exceptions to provide enough information in 'form-error form' so the user can resolve the issues and does not see the error dialog

melton-jason:
Looks like with the addition of extending save blockers on the backend, all explicit BusinessRule Exceptions (except https://github.com/specify/specify7/issues/2518#issuecomment-1347039470) are already covered on the frontend. 

Besides things that maybe need localization, all of the logic is complete. 
@maxxxxxdlp, do we want to localize _everything_ we can in the backend? 
It is not feasible to translate every exception, particularly any non BusinessRuleException and any non AssertionError that we do not raise within the code (SQL, Django, and other exceptions)
Alongside the BusinessRuleExceptions, there are a lot of miscellaneous Assertion Errors that theoretically could be translated. 

If we do translations on errors, we would need to additionally send the untranslated English version within the crash report. So we would know what is going on if the file is ever sent to us

maxxxxxdlp:
> Looks like with the addition of extending save blockers on the backend, all explicit BusinessRule Exceptions (except [#2518 (comment)](https://github.com/specify/specify7/issues/2518#issuecomment-1347039470)) are already covered on the frontend.
> 
> Besides things that maybe need localization, all of the logic is complete. @maxxxxxdlp, do we want to localize _everything_ we can in the backend? It is not feasible to translate every exception, particularly any non BusinessRuleException and any non AssertionError that we do not raise within the code (SQL, Django, and other exceptions) Alongside the BusinessRuleExceptions, there are a lot of miscellaneous Assertion Errors that theoretically could be translated.
> 
> If we do translations on errors, we would need to additionally send the untranslated English version within the crash report. So we would know what is going on if the file is ever sent to us

Let's try to get rid of back-end localization completely and have all localization strings on the front-end (as they are integrated with weblate).

That means, if back-end has a string, you either:
- Don't localize it - this is completely fine if we don't expect to see the error often, of if the business rule is already handled on the front-end (and localized on the front-end)
- Or make back-end send the string key and make front-end resolve the sting key into an actual value. I already did this in some places - https://github.com/specify/specify7/blob/4f03cab1a425f1a579efa97d70b0e0d192d45ffa/specifyweb/frontend/js_src/lib/components/WorkBench/resultsParser.ts#L165

specifyweb/businessrules/accessionagent_rules.py
melton-jason: Added localization to the backend in 6e5af9f and to the frontend in a291cb3

melton-jason:
This can be tested more heavily now. With a291cb3 (Localizing things on the front end) and 6e5af9f (Adding a 'localizationKey' property in the backend repsonse), error messages we can anticipate will be localized. 

More localization can be done (such as localizing the data and exception type), and if that is a blocker to releasing this then those can be added.

specifyweb/businessrules/agent_rules.py
```diff
@@ -6,20 +6,45 @@
 def agent_delete_blocked_by_related_specifyuser(agent):
     try:
         Specifyuser.objects.get(agents=agent)
```
maxxxxxdlp, line 8: Remove this line. It's redundant with the next line

specifyweb/frontend/js_src/lib/components/Errors/JsonError.tsx
```diff
@@ -0,0 +1,200 @@
...
+        <summary>Traceback</summary>
+        <pre>{response.traceback}</pre>
+      </details>
+    </>
+  );
+}
+
+function BusinessRuleExceptionHeader({
+  table,
+  response,
```
maxxxxxdlp, line 66: (totally optional)
Could replace `response,` with `response: {data, exception, message}`

specifyweb/frontend/js_src/lib/localization/backEnd.ts
```diff
@@ -10,6 +10,11 @@ import { createDictionary } from './utils';
 // Refer to "Guidelines for Programmers" in ./README.md before editing this file
 
 export const backEndText = createDictionary({
+
+  ///////////////////////
```
maxxxxxdlp, line 14: note, these comments (however pretty they look) would be removed by weblate when it updates the local strings from the remote.

If you need to add a comment for an individual string, add a `comment` key. Note, the value of the `comment` key is visible to translators.

Example:

```js
{
  ...
  failedParsingBoolean: {
    comment: `
     Some comment.
     Visible to translators.
            not white     space        sensitive

     Blank line is treated as a multi-line.
    `,
    'en-us': '...',
    ...
  },
  ...
}

File: specifyweb/specify/tree_extras.py [hunk 17]
```diff
@@ -69,10 +69,40 @@ def save():
         try:
             model.objects.get(id=self.id, parent__rankid__lt=F('rankid'))
         except model.DoesNotExist:
-            raise BusinessRuleException("Tree node's parent has rank greater than itself.")
+            raise TreeBusinessRuleException(
+                "Tree node's parent has rank greater than itself",
+                {"tree" : self.__class__.__name__,
+                 "type" : "TREE_RANK_INVARIANT_PARENT",
```
maxxxxxdlp, line 75: Why do you need both type and localizationKey? Can we just have localizationKey?

maxxxxxdlp: If you need it just so that the error in the UI is in upper case, then just add a function called `camelToUpper` in here and call it when printing the error code: https://github.com/specify/specify7/blob/c8a6773f746491141a02f0af216a008c12c63b90/specifyweb/frontend/js_src/lib/utils/utils.ts#L29-L42

maxxxxxdlp (changes requested):
Minor changes. Looking good!

specifyweb/frontend/js_src/lib/components/Errors/JsonError.tsx
melton-jason: Is this needed (I can include it if need be)? Right below this I define response to be of type JsonResponse.

specifyweb/specify/tree_extras.py
melton-jason: Fixed in 44f6b8e

specifyweb/businessrules/agent_rules.py
melton-jason: Fixed in 44f6b8e

specifyweb/frontend/js_src/lib/components/Errors/JsonError.tsx
maxxxxxdlp: Response type shouldn't be affected by this change. Full example:

```ts
{
  table,
  response: {data, exception, message}
}: {
  readonly table: string;
  readonly response JsonResponse;
}

maxxxxxdlp (approved):
👍

grantfitzsimmons:
@maxxxxxdlp @melton-jason 

Wow. I'm very impressed with this branch.

## Debug Mode ON

image

I could recreate https://github.com/specify/specify7/issues/2786 with this branch with debug=true

Useless 500 error

## Debug Mode OFF

image

It even shows the exact ID of the CO that already uses that catalog number.

http://localhost/specify/view/collectionobject/134579/

maxxxxxdlp:
2nd screenshot looks great.
Regarding the 1st screenshot, there is no Django error page on the first screenshot because of a bug?

grantfitzsimmons:
Presumably a bug. Same behavior in production, so maybe even with debug mode on it should fall back on this instead

grantfitzsimmons:
Trying to delete a schema mapping field results in this error:
image

melton-jason:
> Trying to delete a schema mapping field results in this error:
> 
> ```
> Object of type QuerySet is not JSON serializable
> ```
> 
> image
> 
> _Just an observation. Not tremendously helpful or informative, just reporting._

There are some cases where something in the original error is not Json Serializable and thus can not be directly sent as a Json response. To get around this, I manually catch the error that is thrown when trying to serialize the object and serialize that instead to json and sent it to the frontend. Based on the traceback, I still know you were trying to delete something. 

I can try and include more information for these errors specifically if we need, but regardless the information we send will probably be limited. 

Actually, I can write up our own way of serializing specifically django's QuerySet. We can not anticipate every object, but QuerySet might appear often so it might be worth the effort.

specifyweb/frontend/js_src/lib/components/Errors/JsonError.tsx
melton-jason: Ah ok! I see what you mean now. Did not know you could do that!

melton-jason:
> Also, looks like it didn't display the collection object icon on the 2nd screenshot correctly (it generated one on the fly)

I apologize for this. Currently, I am using the [TableIcon](https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/components/Molecules/TableIcon.tsx#L39-L95) function to generate the icons. Is there a better way of doing this?

maxxxxxdlp:
That's the correct function to use. It might be that it is case-sensitive in regard to the table name.
The `<TableIcon>` component is rendered in the Data Entry dialog. Looks like it is using `CollectionObject` as the table name.

If it's hard to get a PascalCase table name on the back end, then you can modify TableIcon to call `getIcon(getModel(name)?.name ?? name)` to try to resolve a lowercase table name to PascalCase, and if fails (because name is not a table name, then use the name as it)

melton-jason:
> That's the correct function to use. It might be that it is case-sensitive in regard to the table name. The `<TableIcon>` component is rendered in the Data Entry dialog. Looks like it is using `CollectionObject` as the table name.
> 
> If it's hard to get a PascalCase table name on the back end, then you can modify TableIcon to call `getIcon(getModel(name)?.name ?? name)` instead of `getIcon(name)`. This way it will try to resolve a lowercase table name to PascalCase, and if fails (because name is not a table name), then use the name as it

Looks like every BusinessRule besides uniqueness rules pass the table name correctly to the front end. I should be able to resolve this on the backend. 
I can still make the changes to `TableIcon` if it will be useful in the future.

maxxxxxdlp:
Not a high priority. There is actually a better solution than the one I proposed: make `getIcon` treat name as case-insensitive

melton-jason:
> > That's the correct function to use. It might be that it is case-sensitive in regard to the table name. The `<TableIcon>` component is rendered in the Data Entry dialog. Looks like it is using `CollectionObject` as the table name.
> > If it's hard to get a PascalCase table name on the back end, then you can modify TableIcon to call `getIcon(getModel(name)?.name ?? name)` instead of `getIcon(name)`. This way it will try to resolve a lowercase table name to PascalCase, and if fails (because name is not a table name), then use the name as it
> 
> Looks like every BusinessRule besides uniqueness rules passes the table name correctly to the front end. I should be able to resolve this on the backend. I can still make the changes to `TableIcon` if it will be useful in the future.

This should be fixed now

specifyweb/frontend/js_src/lib/components/Errors/JsonError.tsx
melton-jason: Fixed in c1ee332

melton-jason:
Before this is merged, I would like to resolve [#2518](https://github.com/specify/specify7/issues/2518#issuecomment-1347039470) first. 
We need to consider if we even want to keep this business rule in Specify 7; if so, we need to have a better way of handling it.

grantfitzsimmons:
@tlammer Do you think we need to preserve this business rule? https://github.com/specify/specify7/issues/2518#issuecomment-1347039470

tlammer:
I don't see a reason to keep the rule, but testing would be required and it may cause inconsistency issues.  In Specify 6 you cannot see the address subform for Other type agents.  If you can see it in 7 and not in 6, that is going to cause confusion.

melton-jason:
> I don't see a reason to keep the rule, but testing would be required and it may cause inconsistency issues. In Specify 6 you cannot see the address subform for Other type agents. If you can see it in 7 and not in 6, that is going to cause confusion.

-@tlammer 

How adverse are we to this inconsistency? 
If it is too much concern, we can (preferably) show an error on the form or raise an error instead.

grantfitzsimmons:
lets do it!!!!!!!!!!!!!!!!!!

melton-jason:
Here is the decision we have reached regarding the problematic Business Rule Exception which disallows Addresses being added to Agents of type Other and Group (https://github.com/specify/specify7/issues/2518#issuecomment-1347039470)

The Business Rule will be removed in Specify 7. Before the Specify 6 release (https://github.com/specify/specify6/pull/1189), I will remove it there as well (if the removal does not cause major problems). 
The BusinessRule will be removed from Specify 7 even if we can not safely/easily remove it in Specify 6

PR #2710, opened 2022-12-25
Try out Qodana code quality checker

maxxxxxdlp:
Not sure how useful it would be, but we fit into a free plan so it doesn't cost us much to try it.

Most of the things Qodana checks are already available from inside of JetBrains IDEs. The most important Qodana addition is that in can enforce that project must contain fewer than `x` code quality errors when merging code.

Qodana can also be integrated with ESLint and other tools, making it potentially useful as a UI for viewing results from all linters.

Thus, I can see us using Qodana as a "meter" of code smell, and when the number of Qodana errors surpasses a threshold, we take the time to refactor the code or improve our coding practices - quite important with new developers coming in


maxxxxxdlp:
After looking at the code suggestions further, they don't look to be so useful (or maybe our code just doesn't have any issues 🤗 )

Main things reported:
1. False positive (including false typo). What is worse, some suggestions were downright misleading and could cause bugs (i.e, "Can be simplified to ..." or "Unnecessary check ..."). There also are bugs (i.e, with shorthand syntax for boolean jsx attributes)
2. Rules that can be replaced by ESLint, thus not binding us to a single tool (there are a LOT of good ESLint front-ends)
3. Formatting - can be ignored as Prettier handles our formatting
4. A ton of errors reported in files that are not yet converted to TypeScript (old js, backbone and jquery files) - but we know that those files are bad and need to be rewritten
5. Some valid suggestions, but not super important (line length)

I was really excited it would find some real bugs or suggest super cool coding practices, but it fell short of that. TypeScrpit+Prettier+ESLint are of a ton more use.

It's a bit unfair to judge it on the UI as it's still in beta, but I found it to have bad user experience (resets search filters when going between tabs, there is a way to hide filters, but no way to a list of which filters are currently hidden and others)

My new idea is to fine-tune our ESLint rules and then add strict enforcement for them. Also, integrate with some ESLint UIs to visualizer ESLint results in pull request comments.

PR #2726, opened 2023-01-03
Disambiguate global prefs on the app resources sidebar

### LINKED ISSUE #2430: Preferences need to be uniquely identified by name
image

I know we have discussed this several times before, but I could not find an issue for this.

**Requested By:** @acbentley and myself

_Even if a UI is on the way, can this be resolved sooner? It is impossible to tell the difference until they are opened._

maxxxxxdlp:
Fixes #2430


melton-jason:
Whenever I try to add a new AppResource, I get the following error:

pull_2726_issue

Replicated in `newmexico`, `pripaleo`, and a local instance

grantfitzsimmons (changes requested):
jason is mad

maxxxxxdlp:
This is actually a bug from production. Added in 7.8.2. Fixed here. Would recommend merging this in as soon as possible

maxxxxxdlp:
Fixed, for good.

I also made typescript detect this type of bug so in the future it should never happen again

PR #2728, opened 2023-01-03
Fix the get database name error since django version update issue-2727

acwhite211:
In the update-deps branch, since the django version update, the function to get the current database name in the agent replace view code no longer works and needs to be updated to the new django standards.


maxxxxxdlp:
Fixes #2728

maxxxxxdlp:
Back-end tests passed. Front-end tests failed due to an issue that is already fixed on production. Dependency changes were already tested. Thus merging this in

maxxxxxdlp (approved):

PR #2752, opened 2023-01-05
Improve favicon resolution & add manifest.json

grantfitzsimmons:
Make sure that the higher resolution favicon is displaying in all browsers.

**Firefox:**
image

**Chrome:**
image

and Safari should have a full icon when added to favorites.


grantfitzsimmons:
Safari:
image

image

specifyweb/urls.py
```diff
@@ -20,7 +20,8 @@
 from .workbench import urls as wb_urls
 
 urlpatterns = [
-    url(r'^favicon.ico', RedirectView.as_view(url='/static/img/fav_icon.png')),
+    # First draft, need to see how this is handled
+    url(r'^favicon.ico', RedirectView.as_view(url='/static/img/short_logo.svg')),
```
maxxxxxdlp, line 24: This just tells to redirect requests to /favicon.ico` to `/static/img/short_logo.svg`.

Need to test if browsers can handle favicon.ico being an SVG. If they can't, you would have to use a PNG here instead

specifyweb/frontend/templates/specify.html
```diff
@@ -3,6 +3,11 @@
 {% get_current_language as LANGUAGE_CODE %}
 <html lang="{{ LANGUAGE_CODE }}">
   <head>
+    <link rel="apple-touch-icon" sizes="180x180" href="/static/img/apple-touch-icon.png">
+    <link rel="mask-icon" href="/static/img/safari-pinned-tab.svg" color="#F79244">
+    <link rel="shortcut icon" href="/static/img/favicon.ico">
+    <link rel="icon" href="/static/img/favicon-512x512.png" sizes="any">
+    <link rel="icon" href="/static/img/short_logo.svg" type="image/svg+xml">
```
maxxxxxdlp, line 10: Could you please move this to the end of the `</head>` tag (just after `<script>`?

This is because the `<meta charset="utf-8" />` line must be in the first KB of the file (there is a reason for it, don't remember). But also, we want the `main.js` line to begin executing as soon as possible as that's what actually displays the page

maxxxxxdlp (changes requested):

specifyweb/urls.py
grantfitzsimmons: The last 3 major releases of Chrome, Edge, Firefox, and Opera support this.

https://caniuse.com/link-icon-svg

Safari (of course) does not support this, but there is a special icon just for that browser ( see `safari-pinned-tab.svg`)

image

This introduces some odd behavior, since it must be a silouette. What do you think?

specifyweb/frontend/templates/specify.html
grantfitzsimmons: Done– https://github.com/specify/specify7/pull/2752/commits/b80c26af456cff28c2071e5774bc5bf3cbac3922

maxxxxxdlp:
It might get too technical, but you might be interested in creating a `manifest.json` file for Specify - https://web.dev/add-manifest/

That file would allow to configure site's theme color, appearance when downloaded as a standalone app and other cool things

You can put that file in this directory - https://github.com/specify/specify7/tree/production/specifyweb/frontend/static

maxxxxxdlp (approved):

maxxxxxdlp:
Not sure why we started getting this message on all pull requests, but until we have time to look into it, we can ignore the `test` entry (unlike the `Tests` lines)

Screenshot 2023-01-04 at 22 08 57


Seems to have started happening around the time localization changes were merged. Maybe the github actions from different branches are conflicting 🤷

grantfitzsimmons:
@maxxxxxdlp I am a little nervous to merge without waiting for requirements to be met. Do you want to merge this? We can work on the manifest at a later date

maxxxxxdlp:
@grantfitzsimmons these are the real tests:

Screenshot 2023-01-05 at 20 07 17

They all passed (if they fail you would receive an email)

The "Expected - Wating for status to be reported" is a phantom test.

After you update the manifest.json I can test this branch and then merge it

grantfitzsimmons:
@maxxxxxdlp -
https://web.dev/codelab-make-installable/

There is some more work to be done to make this installable from what I understand. 

We need to add a service worker to this branch to enable installing the app.

https://glitch.com/edit/#!/make-it-installable?path=service-worker.js%3A1%3A0

See this example `service-worker.js`:

```js
const CACHE_NAME = 'offline';
const OFFLINE_URL = 'offline.html';

self.addEventListener('install', function(event) {
  console.log('[ServiceWorker] Install');
  
  event.waitUntil((async () => {
    const cache = await caches.open(CACHE_NAME);
    // Setting {cache: 'reload'} in the new request will ensure that the response
    // isn't fulfilled from the HTTP cache; i.e., it will be from the network.
    await cache.add(new Request(OFFLINE_URL, {cache: 'reload'}));
  })());
  
  self.skipWaiting();
});

self.addEventListener('activate', (event) => {
  console.log('[ServiceWorker] Activate');
  event.waitUntil((async () => {
    // Enable navigation preload if it's supported.
    // See https://developers.google.com/web/updates/2017/02/navigation-preload
    if ('navigationPreload' in self.registration) {
      await self.registration.navigationPreload.enable();
    }
  })());

  // Tell the active service worker to take control of the page immediately.
  self.clients.claim();
});

self.addEventListener('fetch', function(event) {
  // console.log('[Service Worker] Fetch', event.request.url);
  if (event.request.mode === 'navigate') {
    event.respondWith((async () => {
      try {
        const preloadResponse = await event.preloadResponse;
        if (preloadResponse) {
          return preloadResponse;
        }

        const networkResponse = await fetch(event.request);
        return networkResponse;
      } catch (error) {
        console.log('[Service Worker] Fetch failed; returning offline page instead.', error);

        const cache = await caches.open(CACHE_NAME);
        const cachedResponse = await cache.match(OFFLINE_URL);
        return cachedResponse;
      }
    })());
  }
});
```

specifyweb/settings/__init__.py
```diff
@@ -141,6 +141,11 @@
     ('config', SPECIFY_CONFIG_DIR),
 )
 
+# Add web app manifest
+WEBPACK_LOADER = {
+    'MANIFEST_FILE': os.path.join(SPECIFY_CONFIG_DIR, "/static/manifest.json"),
```
maxxxxxdlp, line 146: What does this do? Do we need to add a python dependency for it to work? (https://github.com/django-webpack/django-webpack-loader)

Is this required? You might be able to achieve the same by defining a view that would serve the manifest.json page:

View: https://github.com/specify/specify7/blob/3dced0964721218de64659e8371b1c218c6d3fef/specifyweb/urls.py#L28

it's imported here: https://github.com/specify/specify7/blob/3dced0964721218de64659e8371b1c218c6d3fef/specifyweb/urls.py#L11

the view itself is defined like this:
(https://specifydev.slack.com/archives/D02LXH9PEUU/p1672952503649319)
https://github.com/specify/specify7/blob/3dced0964721218de64659e8371b1c218c6d3fef/specifyweb/frontend/views.py#L20-L29

maxxxxxdlp:
> Not sure why we started getting this message on all pull requests, but until we have time to look into it, we can ignore the `test` entry (unlike the `Tests` lines)
> 
> Screenshot 2023-01-04 at 22 08 57
> 
> Seems to have started happening around the time localization changes were merged. Maybe the github actions from different branches are conflicting 🤷

Finally fixed this thing!
Turns out in branch-protection settings I said that all pull requests must run the `test` github action before merging. Instead, it expected the following settings:

Screenshot 2023-01-10 at 21 38 53

specifyweb/settings/__init__.py
maxxxxxdlp: Did you plan to make any change related to this? I don't see the commit

This is what I get right now:
Screenshot 2023-01-10 at 22 28 01

maxxxxxdlp: Never mind, all is good. It's just that the static files directory in docker does not update automatically when you switch branches, so I had to manually copy the newly added files into it.

maxxxxxdlp: Though, Chrome fails to pick up the icon for me:

image

(even in incognito)

Looks like you told it in specify.html to look at `/static/img/favicon.ico`
Instead, it should be `/favicon.ico`.

I have just pushed that change, and also some template refactoring so that favicons are present on all pages, not just on the main page (the login page, and the choose collection)

grantfitzsimmons (commented):
Thank you!! I was nervous to modify the other templates but I wanted to have these applied to each without duplicating the info. Exactly what I was hoping for!

specifyweb/settings/__init__.py
grantfitzsimmons: I was under the impression anytime a URL ended in `favicon.ico` it would redirect to the one specified in the `urls.py` but I see that is not the case. Thank you!

grantfitzsimmons: `url(r'^favicon.ico', RedirectView.as_view(url='/static/img/short_logo.svg'))`
only redirects if the location is exactly `/favicon.ico` then

maxxxxxdlp: ah good point. that's a subtle difference.

If you remove `^`, then it would match any URL that contains `favicon.ico`

To many any URL that ends with it, you have to say `r'favicon.ico$'`

melton-jason (approved):
Looks so much nicer on Edge and Chrome
I have not tested Firefox or Safari.

specifysoftware:

Issue #2704, opened 2022-12-22
`bad tree structure` error when moving Genus to a different Family


grantfitzsimmons:
To recreate, attempt the following move:

> The Taxon node "Arceuthobium" will be placed, along with all of its descendants, under the new parent "Loganiaceae".

Move `Arceuthobium` to the `Loganiaceae` Family.

https://aafcaacdao-edge.test.specifysystems.org/specify/tree/taxon/?conformation=%7E%7E1%7E181107%7E160131%7E160170--%7E159752----

```
Error occurred fetching from https://aafcaacdao-edge.test.specifysystems.org/api/specify_tree/taxon/160170/move/

Invalid response code 500. Expected 200.

This error may indicate a misconfiguration or a bug in Specify. Please double check your configuration and report this issue.
```

```
AssertionError at /api/specify_tree/taxon/160170/move/
bad tree structure
```

**Request URL:** http://aafcaacdao-edge.test.specifysystems.org/api/specify_tree/taxon/160170/move/

**Reported By:** HeatherC at the Community Forum

https://discourse.specifysoftware.org/t/error-in-s7-tree-moves/949/2

grantfitzsimmons:
[Specify 7 Crash Report - 2022-12-22T17_53_24.586Z.txt](https://github.com/specify/specify7/files/10289337/Specify.7.Crash.Report.-.2022-12-22T17_53_24.586Z.txt)

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/error-in-s7-tree-moves/949/5

grantfitzsimmons:
Similar to https://github.com/specify/specify7/issues/2707

Might have the same underlying issue

melton-jason:
For this case, it looks like the tree is misconfigured. There is a tree node `Cornus canadensis f. dutillyi` (taxonid: `18574`)  which has a child which is the same rank as itself. 
The child is `Cornus canadensis f. viride` (taxonid: `186086`)

``` 
SQL Command> select t.taxonid, p.taxonid, t.fullname, p.fullname, t.parentid, p.parentid, t.rankid, p.rankid from taxon t join taxon p on t.parentid=p.taxonid where t.rankid <= p.rankid;
+---------+---------+-----------------------------+-------------------------------+----------+----------+--------+--------+
| taxonid | taxonid | fullname                    | fullname                      | parentid | parentid | rankid | rankid |
[...]

PR #2753, opened 2023-01-05
Ignore synonymized nodes when checking tree structure

### LINKED ISSUE #2707: `bad tree structure` error when creating a child under a family
1. Navigate to this section in the tree:

http://localhost/specify/tree/taxon/?conformation=%7E%7E1%7E204425-%7E181107%7E4690%7E7330-%7E5162-%7E7299%7E194257----%7E2%7E10%7E16%7E617------

Notice that "Chaetodermatidae" is selected.

image

2. Click "Add Child"
3. Create a child named "grant-test" and assign it the rank of Genus

image

4. Click "Save" and you will see this error:

```
AssertionError at /api/specify/taxon/
bad tree structure
```

See the expanded error below:
[Specify 7 Crash Report - 2022-12-23T21_06_26.336Z.txt](https://github.com/specify/specify7/files/10297333/Specify.7.Crash.Report.-.2022-12-23T21_06_26.336Z.txt)

**Traceback**

```
/opt/specify7/specifyweb/specify/tree_extras.py in validate_node_numbers
from specifyweb.businessrules.exceptions import BusinessRuleException
from  .auditcodes import TREE_MERGE, TREE_SYNONYMIZE, TREE_DESYNONYMIZE
@contextmanager
def validate_node_numbers(table, revalidate_after=True):
    try:
        validate_tree_numbering(table) …
    except AssertionError:
        renumber_tree(table)
    yield
    if revalidate_after:
        validate_tree_numbering(table)
```

Seems to be similar to the issue here: https://github.com/specify/specify7/issues/2704

melton-jason:
Fixes #2707
If synonymized nodes exist as pointers to non-synonymized nodes, is it _that_ important to take these nodes into consideration when accounting for tree structure? At a minimum, perhaps a warning (in some way) can be raised instead. 

>  it's not valid in a taxonomic sense to have a child that is at a higher rank than the parent.
> 
> However, this gets messy with synonymy, as we often see taxonomic changes result in the child of one taxon getting elevated to a different taxonomic rank, which could be equal to or higher than its former parent.
> In short, synonyms do not need to be at the same rank. Differences in opinion of the relative distinctiveness of a given taxon often result in one taxon being recognized at different ranks.

-From Garth at the University of Michigan.

### DIFF
--- a/specifyweb/specify/tree_extras.py
+++ b/specifyweb/specify/tree_extras.py
@@ -409,6 +409,7 @@ def validate_tree_numbering(table):
     cursor.execute((
         "select count(*) from {table} t join {table} p on t.parentid = p.{table}id\n"
         "where t.rankid <= p.rankid\n"
+        "and t.acceptedid is null" if table == 'taxon' else ""
     ).format(table=table))
     bad_ranks_count, = cursor.fetchone()
     assert bad_ranks_count == 0, \
@@ -454,6 +455,7 @@ def renumber_tree(table):
         "select count(*) from {table} t\n"
         "join {table} p on t.parentid = p.{table}id\n"
         "where t.rankid <= p.rankid\n"
+        "and t.acceptedid is null" if table == 'taxon' else ""
     ).format(table=table))
     assert (0, ) == cursor.fetchone(), "bad tree structure"
 


grantfitzsimmons:
image

Screenshot 2023-01-09 at 7 13 02 PM

I could fix in Specify 6 without any errors. Not the same story in Specify 7 unless you are using this branch.

grantfitzsimmons:
image

Is this supposed to fix moving a bad node to a higher rank?

I tried to move **viride** (Forma, child of **dutillyi** (Forma)) to be a child of **alpestris** (Variety)

https://daotest-issue-2707.test.specifysystems.org/specify/tree/taxon/?conformation=%7E%7E1%7E181107%7E132266%7E132345--%7E18522%7E18540%7E18572%7E18578-%7E18574%7E186086--%7E18577-------

image

melton-jason:
Regarding non-synonymized nodes (as in #2704), we can implement a different approach if maintaining the tree structure is important. Once #2702 is merged, it will display an error message containing a list of all nodes which caused the error; this will be helpful but probably not a permanent fix. Perhaps a way to override or toggle the tree checks would be the easiest way of completely fixing this problem (or disabling these checks by default, kind of similar to Specify 6).

grantfitzsimmons (approved):
Jason is 100% sure this won't introduce any issues ever

PR #2755, opened 2023-01-06
Added delete button in data set propreties in workbench

### LINKED ISSUE #1836: Ability to delete data sets from properties dialog
image

I would like to delete the many, many data sets I have imported without needing to load the workbench grid view for every one of them.

CarolineDenis:
Fixed #1836

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -1,0 +1,4 @@
+module.exports = {
+    ...require('@maxxxxxdlp/prettier-config'),
+    // your overwrites go here
+};
--- a/specifyweb/frontend/js_src/lib/components/WorkBench/DataSetMeta.tsx
+++ b/specifyweb/frontend/js_src/lib/components/WorkBench/DataSetMeta.tsx
@@ -36,6 +36,7 @@ import { LocalizedString } from 'typesafe-i18n';
 import { schema } from '../DataModel/schema';
 import { StringToJsx } from '../../localization/utils';
 import { getField } from '../DataModel/helpers';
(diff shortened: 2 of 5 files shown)


specifyweb/frontend/js_src/lib/components/WorkBench/DataSetMeta.tsx
```diff
@@ -36,354 +36,420 @@ import { LocalizedString } from 'typesafe-i18n';
...
-              <option key={user.id} value={user.id}>
-                {user.name}
-              </option>
-            ))}
-          </Select>
-        </Label.Block>
-      </Form>
-    </Dialog>
-  );
+                  return users === undefined ? null : isChanged ? (
```
maxxxxxdlp, line 370: Looks like the code was reformatted inconsistently.
Could you configure the Prettier plugin (https://github.com/specify/specify7/tree/production/specifyweb/frontend/js_src#plugins) in your IDE and reformat this file with it?

maxxxxxdlp (changes requested):

.prettierrc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
```
maxxxxxdlp, line 1: I have just noticed that this file was added. I don't think it is needed
@CarolineDenis does prettier work for you even if this file is removed?

I don't think it is looking at this file as long as we have a "prettier" key in package.json. You need to remove either one of them

specifysoftware:

PR #2769, opened 2023-01-09
Improve query exports to CSV

### LINKED ISSUE #1575: Exported CSV column headers should be their labels
When exporting query results to CSV, all column headers are the field names from the schema rather than the caption for those fields (e.g. “10.collectingevent.text4” rather than “Station number”). This request is to make the CSVs export with the captions so we don’t have to rename the columns every time we want to send data to someone.

**Requested By:** Plarson

### LINKED ISSUE #2761: Exporting to CSV while querying within a record set exports full collection query output
## CSIRO Reported:

> One of our users discovered that the csv export after querying a record set returns incorrect data. I tested this in two different instances and it appears that even though a record-set was used for the query, on csv export the entire database is queried and the corresponding result returned in the csv export file.
> 
> For example, I have a record set of 69 records which I limited to Contains ‘A 16’ in the catalognumber field – this results in 8 records. When I export this to csv, I get a file with 223 records – all of them having ‘A 16’ in the catalognumber field. Which is the same number of records I get when I run the query on the full dataset.

**Reported By:** Corinna P at CSIRO ([Asana Link](https://app.asana.com/0/1201972579963460/1203659671915887/f))

---

image

_I created a record set from the original output of [the query](https://sp7demofish.specifycloud.org/specify/query/56/) then ran the query on the newly created record set with "Kansas" changed to "River"._

image

Even though the query returned 274 results, after exporting the results to CSV, it exported the results as if I queried the entire database rather than the record set.

[query_results_2023-01-09T02_11_55.082573 (1).csv](https://github.com/specify/specify7/files/10369994/query_results_2023-01-09T02_11_55.082573.1.csv)

CarolineDenis:
Make export headings user friendly
Fixed #1575

Fix CSV export not being scoped to recordSet
Fixed #2761

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
@@ -4,6 +4,7 @@ import { commonText } from '../../localization/common';
 import { queryText } from '../../localization/query';
 import { ping } from '../../utils/ajax/ping';
 import type { RA } from '../../utils/types';
+import { keysToLowerCase } from '../../utils/utils';
 import type { SerializedResource } from '../DataModel/helperTypes';
 import type { SpecifyResource } from '../DataModel/legacyTypes';
 import type { SpQuery, SpQueryField, Tables } from '../DataModel/types';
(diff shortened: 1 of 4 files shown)


maxxxxxdlp (approved):

grantfitzsimmons:
Wow– this is _long_ overdue regarding the headers! Just tested & can verify the fix works as intended. Such a great fix! Thank you @CarolineDenis !!

grantfitzsimmons (approved):
Works great!

PR #2770, opened 2023-01-10
Make current collection fetching more efficient

maxxxxxdlp:
Turns out available collections were being fetched twice.

After this change, the collection switcher should be populated with collections as soon as the page loads.

Additional things to test:
- Sorting
- Switching collection
- Collection switcher for a guest user

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Header/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/index.tsx
@@ -5,14 +5,13 @@
 import React from 'react';
 import { useNavigate } from 'react-router-dom';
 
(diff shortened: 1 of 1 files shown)


grantfitzsimmons:
> Collection switcher for a guest user

Is this something I will need to establish in a local instance until the test panel supports actual guest users?

maxxxxxdlp:
Yeah, sorry about that. Test panel is so neglected by us 😢

grantfitzsimmons (approved):
I was able to switch collections correctly while being an anonymous user and a logged in user. Sorting using the tool in the preferences looks to be working correctly. Everything is working as expected!

specifysoftware:

PR #2771, opened 2023-01-10
Improve code quality checks

maxxxxxdlp:
- Automatically run Prettier and ESLint auto-fix after all new commits
- Reformat all files with Prettier and run ESLint on all existing files
   - This makes it look like this commit has a lot of changes, but most of them are automated, thus if something is broken, it would be consistently broken and broken everywhere.

Testing:
- High-level general testing of major features


chanulee1 (approved):
Major features appear to be working correctly.

PR #2777, opened 2023-01-10
Disable attachment when no attachment server is installed

### LINKED ISSUE #2585: Disable attachment features if there is no attachment server
> As a test, we hit our DEV instance of 7.7.5 at: http://herb.specify-dev.lsa.umich.edu/specify/attachments/ to see if attachments load.
We found that Specify's attachment view attempts to render and request images.
From the screenshot below the images do not load [as intended].

image

> Other attachment views also render attachment-related controls but operations fail or are not possible due to the attachment server being disabled.

image

> If you click save though it does not succeed in creating the empty attachments [this is good].
>
>I suppose this ticket is just to discuss whether all of these views should be disabled if the attachment server is not configured.

Matthew makes a good point– areas where attachments could be attached/viewed should either be disabled or be better at reporting there is no attachment server rather than attempting to upload or display attachments that cannot be reached. While this is uncommon, I imagine this would be helpful for users to communicate the disconnect.

**Reported By:** Matthew Cruz at UMich (Email)

CarolineDenis:
Fixed #2585

Added a dialog in attachment page to notify no server are installed.

Disable attachment edit in both data entry form and record set form.


specifyweb/frontend/js_src/lib/localization/attachments.ts
```diff
@@ -24,11 +24,14 @@ export const attachmentsText = createDictionary({
...
-    'fr-fr': 'Serveur de pièces jointes indisponible.',
-    'uk-ua': 'Сервер прикріплень недоступний.',
+    'en-us': 'Attachment server unavailable',
+    'ru-ru': 'Сервер прикрепленных файлов недоступен',
+    'es-es': 'Servidor de archivos adjuntos no disponible',
+    'fr-fr': 'Serveur de pièces jointes indisponible',
+    'uk-ua': 'Сервер прикріплень недоступний',
+  },
+  attachmentServerUnavailableDescription: {
+    'en-us': 'Setup your attachment server'
```
grantfitzsimmons, line 34: I would reccomend expanding this with instructions once they are written. We can make this change at a later date

grantfitzsimmons (approved):
It works as intended on my system! I am not able to add new attachments and it informs the user they should setup the attachment server.

specifyweb/frontend/js_src/lib/localization/attachments.ts
melton-jason: How complex are these instructions? They may be too long to warrant putting in a dialog. 
If/when we do add (summarized?) instructions, perhaps a link to a post on the [Speciforum](https://discourse.specifysoftware.org/) may also be helpful. 
@grantfitzsimmons, Do we have public (non-member) documentation detailing the process?

grantfitzsimmons: This will definitely link out to the forum! We cannot possibly localize an entire set of instructions 😂

Max asked me to write documentation for this since we do not have any user-facing docs on this. It's on my to-do list but I'll have to learn the system better myself before writing docs

maxxxxxdlp: Hehe, you will become the most knowledgeable not just in web portals, but also in asset servers 😂

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/localization/attachments.ts
grantfitzsimmons: I'll let you focus on the big stuff. I'm already writing an outline... now I just need to deploy some asset servers and make sure I know what's happening!

PR #2780, opened 2023-01-11
Cascade permissions for dependent relationships

### LINKED ISSUE #2007: Cascade permissions for dependent relationships
If user has no access to removing Determinations, which are a dependent relationship of the Collection Object table, they can still remove them by pressing the `-` button in the Determinations sub view.

Currently the `-` button is controlled only by the `Table -> Collection Object -> Update` permission, as it is editing the Collection Object record, not determination.
UI should be made to also check the `Table -> Determinations -> Delete` permission.

More info: https://discourse.specifysoftware.org/t/determiner-role-in-security-accounts/611

melton-jason:
Fixes #2007

The code for this was seemingly already in place (unless there is something about this issue I am understanding incorrectly), there was just a minor bug which had the `remove` button use the `create` permission rather than `delete` permission.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
@@ -227,7 +227,7 @@ export function IntegratedRecordSelector({
                 ) : undefined}
                 {hasTablePermission(
                   relationship.relatedModel.name,
-                  isDependent ? 'create' : 'read'
+                  isDependent ? 'delete' : 'read'
                 ) && typeof handleRemove === 'function' ? (
                   <DataEntry.Remove
                     disabled={


maxxxxxdlp:
Yeah, looks like a copy-paste error.

Though, it would be good to test this will different variations of create/modify/delete permissions for dependent/independent to see if the behavior makes sense.

chanulee1:
How do I test this?

maxxxxxdlp:
See https://github.com/specify/specify7/issues/2007#issue-1344706318

To test:
1. Create a user without access to removing Determinations but with edit access to Collection Object
2. Check that the user can not press the `-` button in the Determination subview on the Collection Object form on this branch, but that they can on `production`

chanulee1 (approved):
For a user without access to removing Determinations but with edit access to Collection Objects, the `-` button is not present in the Determination subview. It is present on `production`.

maxxxxxdlp (approved):

Issue #2784, opened 2023-01-11
`IntegrityError` when trying to delete a user


grantfitzsimmons:
(1451, 'Cannot delete or update a parent row: a foreign key constraint fails (`KUVertPaleo_1_11_23`.`spprincipal_sppermission`, CONSTRAINT `FK9DD8B2FA99A7381A` FOREIGN KEY (`SpPrincipalID`) REFERENCES `spprincipal` (`SpPrincipalID`))')

```
IntegrityError at /api/specify/specifyuser/10/
```

```
(1451, 'Cannot delete or update a parent row: a foreign key constraint fails (`KUVertPaleo_1_11_23`.`spprincipal_sppermission`, CONSTRAINT `FK9DD8B2FA99A7381A` FOREIGN KEY (`SpPrincipalID`) REFERENCES `spprincipal` (`SpPrincipalID`))')

```

https://kuvertpaleo11123-edge.test.specifysystems.org/specify/security/user/10/

current user: `spvertpaleoadmin`

[Specify 7 Crash Report - 2023-01-11T15_23_34.095Z.txt](https://github.com/specify/specify7/files/10393742/Specify.7.Crash.Report.-.2023-01-11T15_23_34.095Z.txt)

grantfitzsimmons:
See comment https://github.com/specify/specify7/issues/2783#issuecomment-1378944582

> So there were some layers to this issue–
> 
> When deleting a record that was blocking deletion, it would allow you to delete the record, and once that was done, it would immediately ask you whether you want to delete the user.
> 
> If you tried to delete the user and there was _another_ deletion blocker that was not visible initially, it would throw a 500 error.
> 
> Once you carefully deleted all blocking records and canceled the user delete in the subsequent dialog, you could try to delete the user once and for all.
> 
> Of course, once the blockers were removed, there is a new error: #2784

melton-jason:
I believe this is a duplicate of #1069. The issue occurs because Specify 7 does not also delete the Specify 6 permissions associated with that user. 
Will be fixed when #1694 is fixed.

melton-jason:
> **Describe the bug** Made a new guest user with no access to collections/etc. so no agent attached to it. Tried to delete it as an admin and got an integrity error.
> 
> **To Reproduce** Steps to reproduce the behavior:
> 
> 1. Log in as an admin user
> 2. Go to Security and Accounts and create a new user. Don't give it collection access, only give it a username, password and put it in the Guest user group.
> 3. Save the new user.
[...]

Issue #2511, opened 2022-11-29
`ProtectedError at /api/specify/borrow/1365/` when deleting borrow 


grantfitzsimmons:
```
ProtectedError at /api/specify/borrow/1365/
```
```
("Cannot delete some instances of model 'Addressofrecord' because they are referenced through a protected foreign key: 'Borrow.addressofrecord'", <QuerySet [<Borrow: Borrow object (143)>, <Borrow: Borrow object (417)>, <Borrow: Borrow object (722)>, <Borrow: Borrow object (795)>, <Borrow: Borrow object (808)>, <Borrow: Borrow object (859)>, <Borrow: Borrow object (869)>, <Borrow: Borrow object (870)>, <Borrow: Borrow object (954)>, <Borrow: Borrow object (1012)>, <Borrow: Borrow object (1025)>, <Borrow: Borrow object (1092)>, <Borrow: Borrow object (1245)>, <Borrow: Borrow object (1315)>, <Borrow: Borrow object (1334)>, <Borrow: Borrow object (1420)>, <Borrow: Borrow object (1519)>, <Borrow: Borrow object (1585)>, <Borrow: Borrow object (1660)>, <Borrow: Borrow object (1685)>, '...(remaining elements truncated)...']>)
```

RBGE reports being unable to delete this record due to the above error.

https://herbrbge-edge.test.specifysystems.org/specify/view/borrow/1365/

[Specify 7 Crash Report - 2022-11-29T21_41_04.770Z.txt](https://github.com/specify/specify7/files/10117579/Specify.7.Crash.Report.-.2022-11-29T21_41_04.770Z.txt)

maxxxxxdlp:
Back-end bug.

FYI for new developers:

There is a `/api/delete_blockers/` endpoint that tells front-end if a given resource can be deleted, or if some other resource depends on it.

That endpoint returns no blockers for this record:
https://herbrbge-edge.test.specifysystems.org/api/delete_blockers/borrow/1365/

Need to figure out why it doesn't report this dependency as a delete blocker

maxxxxxdlp:
Until then, a workaround is to manually delete the related records or make them reference a different borrow record.

grantfitzsimmons:
Not solved by #2806 

https://herbrbge6802-issue-1694.test.specifysystems.org/specify/view/borrow/1365/

[...]

Issue #1069, opened 2021-11-16
Can't delete users


tlammer:
Reported by Gothenburg, confirmed in several databases on the test server using the "next-release" version.
When you click to delete a user a dialog comes up saying they couldn't be deleted.  Here is the error reported by Gothenburg:
Log from assigning collections:

specify7_1         | Traceback (most recent call last):

specify7_1         |   File "/opt/specify7/ve/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner

specify7_1         |     response = get_response(request)

specify7_1         |   File "/opt/specify7/ve/lib/python3.6/site-packages/django/core/handlers/base.py", line 115, in _get_response

specify7_1         |     response = self.process_exception_by_middleware(e, request)

specify7_1         |   File "/opt/specify7/ve/lib/python3.6/site-packages/django/core/handlers/base.py", line 113, in _get_response

specify7_1         |     response = wrapped_callback(request, *callback_args, **callback_kwargs)

specify7_1         |   File "/opt/specify7/specifyweb/specify/views.py", line 20, in wrapped

specify7_1         |     return view(request, *args, **kwargs)

specify7_1         |   File "/opt/specify7/ve/lib/python3.6/site-packages/django/views/decorators/http.py", line 40, in inner

specify7_1         |     return func(request, *args, **kwargs)

specify7_1         |   File "/opt/specify7/ve/lib/python3.6/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func

specify7_1         |     response = view_func(request, *args, **kwargs)

specify7_1         |   File "/opt/specify7/specifyweb/context/views.py", line 77, in user_collection_access

specify7_1         |     set_users_collections(cursor, user, collections)

specify7_1         |   File "/opt/specify7/specifyweb/context/views.py", line 49, in set_users_collections

specify7_1         |     cursor.execute('delete from spprincipal where grouptype is null and spprincipalid not in ('

specify7_1         |   File "/opt/specify7/ve/lib/python3.6/site-packages/django/db/backends/utils.py", line 67, in execute

specify7_1         |     return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)

specify7_1         |   File "/opt/specify7/ve/lib/python3.6/site-packages/django/db/backends/utils.py", line 76, in _execute_with_wrappers

specify7_1         |     return executor(sql, params, many, context)

[...]

Issue #2584, opened 2022-12-12
Workbench Dataset deletes alongside user with no warning


melton-jason:
**Describe the bug**
If a user has ownership of one or more Datasets, when that user is deleted there is no warning dialog or deletion-blocker outside of the expected 'Are you sure you want to delete this SpecifyUser from the database?': the user is removed from the database, along with all datasets the user owned. 

**To Reproduce**
1. Transfer ownership of a dataset to other user
2. Go to Security and Accounts
3. Delete user that has access to dataset
4. See that Agent _AND_ datasets are deleted from the database

**Desktop:**
 - OS: [Windows 11]
 - Browser: [Chrome]
 - Specify 7 Version: [testability]

I have not tested the case where a user owns a resource where the deletion-blockers _do_ apply, but I would assume this bug would have no significant impact in that scenario. 

Are there deletion-blockers in place for the `spdataset` table?

maxxxxxdlp:
See also https://github.com/specify/specify7/issues/1694

emenslin:
Can recreate in edge (7.9.6)

PR #2806, opened 2023-01-12
Refactor and refine cascade deletion rules

### LINKED ISSUE #1694: Fine tune the deletion cascade rules
Can't delete SpAuditLog because of SpAuditLogField records:
Screen Shot 2022-05-26 at 21 28 34

The delete operation should probably cascade in this case.

A few other examples that should cascade, in case they don't already:
Deleting SpLocaleContainer should delete SpLocaleContainerItem and SpLocaleItemStr
Deleting SpLocaleContainerItem should delete SpLocaleItemStr
Deleting SpQuery should delete SpQueryField
Deleting RecordSet should delete RecordSetItem
Deleting SpAppResourceDir should delete SpAppResource, SpViewSetObjand SpAppResourceData
Deleting PickList should delete PickListItem

melton-jason:
Fixes https://github.com/specify/specify7/issues/1694, https://github.com/specify/specify7/issues/2784, https://github.com/specify/specify7/issues/1069
Additionally test #2511 to see if the error still occurs. If so, we can easily add AddressofRecord to Borrow deletion blockers. 

In addition to refactoring the code, the following rules were added:

While deleting any of the following:
- Picklist should delete Picklist Item
- Spquery should delete Spqueryfield
- Spauditlog should delete Spauditlogfield
- Specifyuser should delete Spprincipal

In addition, we now have the ability to extend deletion blockers as we see fit on the backend. 
Added:
- Specifyuser added to the Agent delete blockers (You should not be able to delete an Agent when associated with Specifyuser)


specifyweb/specify/build_models.py
```diff
@@ -112,11 +96,13 @@ def make_relationship(modelname, rel, datamodel):
         return None
 
     try:
-        on_delete = SPECIAL_DELETION_RULES["%s.%s" % (modelname.capitalize(), rel.name.lower())]
+        on_delete = SPECIAL_DELETION_RULES[f"{rel.name.capitalize()}"][f"{modelname.lower()}"]
```
maxxxxxdlp, line 99: It looks like you swapped the relationship and model name. Is this intentional?

specifyweb/specify/deletion_rules.py
```diff
@@ -0,0 +1,67 @@
+from django.db import models
+
+""" Special Deletion Rules for table relationships 
+    
+    Of the form: base_table : { relationship: action}
+
+    Possible Django actions (w/ django 2.2) are:
```
maxxxxxdlp, line 7: Thanks for documenting this!

maxpatiiuk (approved):

specifyweb/specify/build_models.py
melton-jason: I think I should change the variable names...

This was confusing even in production. 

Take this for example. 
https://github.com/specify/specify7/blob/741a5e34da50328ce8706d25c2071faba3976a2e/specifyweb/specify/build_models.py#L77
This operates as 'When a Specifyuser is deleted, set its relationship with Agent to null"
Where the on_delete functionality for these rules get defined, the `modelname` would be Agent, and `rel.name` would be `specifyuser`.
https://github.com/specify/specify7/blob/741a5e34da50328ce8706d25c2071faba3976a2e/specifyweb/specify/build_models.py#L115

I thought it might make more sense to have the table being deleted be the 'primary' key of the dictonary, so I changed it to `'Specifyuser' : {'agent': models.SET_NULL}`

As a result of this flip, I flipped the variables where `on_delete` was being defined to ensure the same functionality.

melton-jason: I can change it to `'Agent': {'specifyuser': models.SET_NULL}` if that makes more sense

maxxxxxdlp: I think adding an example breakdown you provided here as a comment on top of `SPECIAL_DELETION_RULES` (in addition to what you already have there) would be great

grantfitzsimmons:
Add a block for
- `spdataset`
- Queries
- Reports & Labels (if they aren't already covered by `spappreosurce`)

melton-jason:
> Add a block for
> 
> * `spdataset`

For more context, this is regarding adding a deletion blocker to SpecifyUser. 
This operation is not currently supported with how I extended deletion blockers, but I will add a way to support adding reverse relationships so this can be added.

melton-jason:
> > Add a block for
> > 
> > * `spdataset`
> 
> For more context, this is regarding adding a deletion blocker to SpecifyUser. This operation is not currently supported with how I extended deletion blockers, but I will add a way to support adding reverse relationships so this can be added.

`Spatasets` will more complicated than any other table(s) to work with on the backend. Many of Specify's internal helper functions rely on the datamodel being loaded when Specify starts up (this happens in [load_datamodel.py](https://github.com/specify/specify7/blob/production/specifyweb/specify/load_datamodel.py)). 
The issue is that to get the list of tables and fields in the database, Specify 7 reads the `specify_datamodel.xml` in the Specify 6 config directory (as seen below), and as far as I know the spdataset table is not included within that file.
https://github.com/specify/specify7/blob/4877f75686668cdd5b2d28c46881fb7628ce69a4/specifyweb/specify/load_datamodel.py#L222-L231

We could manually use SQL to get the correct information, but that will only be a short-term solution (which may be fine for now). Preferably, we would want to engineer a way to load/register these Specify 7 specific tables in the internal specify datamodel. (If such a way already exists and I am overlooking it, then feel free to point it out. I have checked all attributes on specify.models and spdataset does not exist)

melton-jason:
Additionally requested by @grantfitzsimmons:
Add a Report/Label deletion blocker to Specifyuser

melton-jason:
I have went through and cleaned up all of the deletion rules as they stand, so this is ready for testing. 

Adding reverse-relationship deletion blockers (i.e. the relationship we want to protect does not exist on the table we want to delete) will require a little more engineering and time: there is no reason to further delay the changes that have already been made on this branch. 

These deletion rules can probably be addressed/changed when #2847 and #2813 are being developed (once those two issues are being developed, #2584 can be fixed as well), but this branch should probably be merged before work on those issues begin. 

Also, along with testing #2511, can issues #1385, #1431, #88, and #150 additionally be tested to see if those issues still occur? If so, they can be fixed in this PR.

melton-jason:
| Base Table  | Relationship | Action |
| ------------- | ------------- | ------------- |
| Spappresource  | SpReports  | Cascade  |
| Recordset  | RecordSet Items  | Cascade  |
| Specifyuser | Agent  | Sets Null  |

grantfitzsimmons:
It looks like #1385 is not solved

> image
> 
> https://fwri2023-issue-1694.test.specifysystems.org/specify/query/48/
> 
> #2806 does not seem to help this

grantfitzsimmons:
Does not solve #1431 either

> Still able to recreate
> 
> https://fwri2023-issue-1694.test.specifysystems.org/specify/view/collectionobject/367315/
> 
> image
>

grantfitzsimmons:
I couldn't recreate #88 (likely because I do not understand the issue) and #150 (seems to be invalid/already solved).

grantfitzsimmons:
#2511 is not solved either

> https://herbrbge6802-issue-1694.test.specifysystems.org/specify/view/borrow/1365/
> 
> image
>

grantfitzsimmons (commented):
Please let me know if everything I discovered is intentional. I can formally aprove once I hear from you again

melton-jason:
These remaining issues should be fixed now.

- Spauditlogfield should no longer block the deletion of Spauditlog

- #1431
  - This will require some discussion. How should we handle LoanPreparation when it's associated Preparation is deleted?  For comparison purposes to other branches I have set the relationship to null between Preparation and LoanPreparation. (The data seems to be preserved within the LoanPreparation, but it is 'orphaned' )
  
  - On other branches, LoanPreparations are a deletion blocker for Preparations, but the 'Trash' icon can still be clicked from the CollectionObject form and the user can 'attempt' to remove the Preparation. If we want to preserve the deletion blocker, we would need an implementation in the front end

Some concerns:
- #2511 

Correctly solving this issue would be impossible with the current way the backend sets up its deletion rules (There is no `Addressofrecord.borrows` relationship). The code can be refactored to support handling cases like these, but I am not sure how bugprone the solution may be if not hardcoded. 

There is another concerning side-effect with this issue (currently in production). When Borrow is deleted, it tries to cascade and delete its associated Addressofrecord. If there are multiple records referencing the Addressofrecord, the error is thrown. However, if the Borrow is the only record referencing the Addressofrecord, both the Addressofrecord _and_ Borrow will be deleted. 
This actually holds true for all of the following relationships:

- Accession.addressofrecord
- Borrow.addressofrecord
- Exchangein.addressofrecord
- Exchangeout.addressofrecord
- Gift.addressofrecord
- Loan.addressofrecord
- Repositoryagreement.addressofrecord

Interactions are flagged as dependents of Addressofrecord (i.e., they require an Addressofrecord), but at a database level they do not require an Addressofrecord (similar to #2299). 
@maxxxxxdlp or @grantfitzsimmons  do you know why there is this discrepancy? 

- #2784 

Spprincipal and Specifyuser have a many-to-many relationship: a relationship type which is not created/supported on the backend. We would either need a proper implementation to handle creating Django many-to-many relationships (the on delete behavior would have to be [handled differently](https://stackoverflow.com/questions/6333068/django-removing-object-from-manytomany-relationship) than other fields) or use SQL directly to handle deletion rules on a case-by-case basis. 

- #1385 

To maintain compatibility with Specify 6, I will not touch the schema mapping fields in this pull request. This can be addressed again when the frontend gets a 'proper' implementation of schema mappings.

grantfitzsimmons:
@melton-jason I need to take a look at this again.

>Interactions are flagged as dependents of Addressofrecord (i.e., they require an Addressofrecord), but at a database level they do not require an Addressofrecord (similar to https://github.com/specify/specify7/issues/2299).
@maxxxxxdlp or @grantfitzsimmons do you know why there is this discrepancy?

I would remove that requirement– that's bizarre. What do you think @maxxxxxdlp ?

maxxxxxdlp:
I believe address of record is marked as dependent so that it can be displayed as a sub view on the form (otherwise it would only be available as a query combo box and all existing forms would have to be redone)

maxxxxxdlp:
Fixing https://github.com/specify/specify7/issues/114 would allow for editing independent resources in sub-views

grantfitzsimmons:
@melton-jason Is this something that just needs to be tested?

melton-jason:
Yes, testing can probably be done on this. There are still 2 cases I need further clarification/discussion on.

>  There is another concerning side-effect with this issue (currently in production). When Borrow is deleted, it tries to cascade and delete its associated Addressofrecord. If there are multiple records referencing the Addressofrecord, the error is thrown. However, if the Borrow is the only record referencing the Addressofrecord, both the Addressofrecord and Borrow will be deleted.
> This actually holds true for all of the following relationships:
> 
> Accession.addressofrecord
> Borrow.addressofrecord
> Exchangein.addressofrecord
> Exchangeout.addressofrecord
> Gift.addressofrecord
> Loan.addressofrecord
> Repositoryagreement.addressofrecord
>
> Interactions are flagged as dependents of Addressofrecord ..., but at a database level they do not require an Addressofrecord (similar to https://github.com/specify/specify7/issues/2299).
> @maxxxxxdlp or @grantfitzsimmons do you know why there is this discrepancy?

It has been awhile, so I would need to look into this specific issue again. Should deleting the interaction try and delete its address of record? I would immediately say no, because the address of record can be referenced by other interactions (or tables), hence the issue that caused #2511.

- - - 

(#1431)
> This will require some discussion. How should we handle LoanPreparation when it's associated Preparation is deleted? For comparison purposes to other branches I have set the relationship to null between Preparation and LoanPreparation. (The data seems to be preserved within the LoanPreparation, but it is 'orphaned' )
> 
> On other branches, LoanPreparations are a deletion blocker for Preparations, but the 'Trash' icon can still be clicked from the CollectionObject form and the user can 'attempt' to remove the Preparation. If we want to preserve the deletion blocker, we would need an implementation in the front end

We need to identify what _really_ needs to be done and what the issue is here. I would lean more towards this being a frontend issue with -to-many relationships. The issue is caused because a Preparation is removed from the Collection Object form, but that Preparation references a Loan Preparation, which causes the error to be thrown. 

@grantfitzsimmons For that view, I envision making any -to-many relationship like this display the warning symbol next to the trash can, which displays that resource's deletion blockers when clicked and prevents the deletion 
We can probably do a similar approach to the minus button in a subview representation of the resource.
Would this be an acceptable solution?

carlosmbe:
I'm going to request both UX and Dev Testing on this PR to conform with our new policy on needing 3 Reviews, Consisting of both Devs and UX people.

CarolineDenis (approved):

melton-jason:
There were changes to the delete blocker dialog and the corresponding backend endpoint in `v7.9.0` (specifically in the following commits: https://github.com/specify/specify7/commit/dd2710154230cd219e44e0c1c4a0d5f036b50a6b, https://github.com/specify/specify7/commit/80087a29f4547c5c9fbd9ca37fd40309241af3fb), which resulted in the Additional Delete Blockers not being resolved correctly. 
This issue has since been resolved. 

To test this fix:
- Identify an Agent which is associated with a `Specifyuser`
  - Here is a Query which identifies all Agents in the `lsumz mammals` database: https://lsumzmammals-issue1694.test.specifysystems.org/specify/query/53/
-  The SpecifyUser should appear in the Agent's delete blocker dialog

pashiav (changes requested):
**https://github.com/specify/specify7/issues/1385** could not reproduce error on `edge` with https://fwri2023-edge.test.specifysystems.org/specify/query/48/

**https://github.com/specify/specify7/issues/1431** 
- Deletes preparation linked to interaction without error! 
- When I clicked **Save** for the loan, a new tab opened with this: [ReportException at _report_runner_run_.pdf](https://github.com/specify/specify7/files/14089804/ReportException.at._report_runner_run_.pdf). Is this related or a separate issue?


https://github.com/specify/specify7/assets/118556884/19864d7c-44bc-42ec-bcd6-ee2df43bc913



**https://github.com/specify/specify7/issues/2511** not fixed
[Specify 7 Crash Report - 2024-01-29T21_16_00.444Z.txt](https://github.com/specify/specify7/files/14089839/Specify.7.Crash.Report.-.2024-01-29T21_16_00.444Z.txt)

https://herbrbge71423-issue1694.test.specifysystems.org/specify/view/borrow/1364/ 

**https://github.com/specify/specify7/pull/2806#issuecomment-1913346814** looks good!
Screenshot 2024-01-29 at 3 44 27 PM

PR #2807, opened 2023-01-13
Remove additional deletion blocker from import

melton-jason:
I have no words on how this left over import got through with my merge from #2702. 

The additional delete blockers were moved to #2806 in commit c25ab91 and were supposed to be cleaned up from #2702 in 0387638. 
I am amazed this got past me, the dockerize test, and backend tests.


maxxxxxdlp (approved):
Doesn't seem like a big deal.
Unlike on the front-end, where TypeScript will throw a type error for any unused import, the back-end does not currently have any such checks (though you are welcome to configure some if you want)

melton-jason:
I realize that now, unfortunately. It will not happen again and I will be more vigilant about these things

PR #2809, opened 2023-01-13
Fix race condition on multiple stats pages open

maxxxxxdlp:
To reproduce:
1. Open stats page
2. Open one more in a different tab
3. Click "Update" in one tab
4. Wait for it to finish
5. The tabs should not be making requests/callback calls in a loop

Able to reproduce the bug before the fix that is included in this branch. Unable to reproduce it after the fix. @realVinayak can you please test it to verify it's also fixed for you?

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Statistics/StatItems.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Statistics/StatItems.tsx
@@ -1,21 +1,24 @@
-import type { SpQueryField, Tables } from '../DataModel/types';
+import React from 'react';
+
+import { ajax } from '../../utils/ajax';
+import { throttledAjax } from '../../utils/ajax/throttledAjax';
 import type { RA } from '../../utils/types';
 import type { SerializedResource } from '../DataModel/helperTypes';
-import { StatsResult } from './StatsResult';
-import React from 'react';
+import type { SpecifyResource } from '../DataModel/legacyTypes';
+import type { SpQuery, SpQueryField, Tables } from '../DataModel/types';
 import {
   queryCountPromiseGenerator,
   useCustomStatsSpec,
   useResolvedSpec,
   useResolvedSpecToQueryResource,
 } from './hooks';
-import type { CustomStat, DefaultStat, StatsSpec } from './types';
-import { SpecifyResource } from '../DataModel/legacyTypes';
-import { SpQuery } from '../DataModel/types';
-import { useAsyncState } from '../../hooks/useAsyncState';
-import { throttledAjax } from '../../utils/ajax/throttledAjax';
-import { BackendStatsResult } from './types';
-import { ajax } from '../../utils/ajax';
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/Statistics/index.tsx
```diff
@@ -155,9 +155,9 @@ export function StatsPage(): JSX.Element | null {
   ]);
 
   /* Set Default Layout every time page is rendered*/
-  React.useEffect(() => {
-    setDefaultLayout(defaultLayoutSpec);
-  }, [setDefaultLayout, defaultLayoutSpec]);
+  // React.useEffect(() => {
```
maxxxxxdlp, line 158: Not sure why this was needed so I commented it out. Feel free to re-enable if it's needed

specifyweb/frontend/js_src/lib/components/Statistics/StatItems.tsx
```diff
@@ -133,37 +141,32 @@ function BackEndItem({
...
   readonly onRemove: (() => void) | undefined;
   readonly onItemRename: ((newLabel: string) => void) | undefined;
-  readonly onItemValueLoad:
-    | ((value: number | string, itemLabel: string) => void)
-    | undefined;
+  readonly onItemValueLoad: ((value: number | string) => void) | undefined;
 }): JSX.Element {
-  const [count] = useAsyncState<number | string | undefined>(
-    React.useCallback(
+  const shouldFetch =
```
maxxxxxdlp, line 146: This is a good way to do it as it prevent needless calls to useEffect.

Calls to useEffect may cause react to re-render even if setState wasn't called (don't remember the exact reason why, but I think they needed to do it to ensure consistency), thus it might be worth handling conditions outside useEffect, if it means useEffect has fewer dependencies.

But also, I think it's just cleaner (personal opinion)

specifyweb/frontend/js_src/lib/components/Statistics/StatItems.tsx
```diff
@@ -1,21 +1,24 @@
-import type { SpQueryField, Tables } from '../DataModel/types';
```
maxxxxxdlp, line 1: The imports sorting and props sorting are dony by ESLint, not explicitly me

melton-jason:
Nice! Glad we can get a testable PR for this out so quick!

grantfitzsimmons:
After freshly composing a new instance on this PR using the `herb_rbge_6802` database on my laptop (I have used it with the stats page before, both times ending in massive memory usage and major stuttering), I experienced _a ton_ of memory usage.

https://user-images.githubusercontent.com/37256050/212349875-9aba5dea-5344-4270-a29f-d49bfc8ed7ed.mp4

maxxxxxdlp:
Closing as these changes were manually copied by @realVinayak to `statspage` branch

Issue #2080, opened 2022-09-01
Schema Viewer does not show the ID field


maxxxxxdlp:
The database record id field is present in every table, but is not exposed by the Database Schema viewer in Sp7

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/specify-7-8-6-release-announcement/1064/1

PR #2816, opened 2023-01-13
Improve schema viewer user experience

### LINKED ISSUE #1831: Adapt schema viewer layout for printing and Ctrl+F
Make the layout better suited for printing (currently, you can only print one page at a time). This would also allow doing `Ctrl+F`. Possibly introduce a switch between continuous and single table view modes.

CarolineDenis:
Fixes #1831


melton-jason:
Is #2080 within the scope of this Pull Request, or will that be added at a later date?

melton-jason (approved):
Looks great, and supports using Find (CTRL + F) to quickly go to _any_ field in the database!


schemaViewer

While not important, would it also be possible to include a button above the `Accession` Table button in the sidebar to redirect to the user to the very top of the Schema Viewer? Currently the quickest way to back to the list of every table when you are anywhere else on the page is to click on `Accession` and then scroll up.

grantfitzsimmons (approved):
Nice work! 

@melton-jason, a jump to the start button would be nice but how often will that be necessary?

melton-jason:
> Nice work! 
> 
> @melton-jason, a jump to the start button would be nice but how often will that be necessary?

Not that necessary. Not necessary enough to keep this from production.

grantfitzsimmons:
https://user-images.githubusercontent.com/37256050/212574305-25bccd7d-3c8d-4e4e-9302-1baad2428599.mov

Unfortunately the first item I select after loading into the schema viewer is not properly highlighted.

grantfitzsimmons (changes requested):
> Unfortunately the first item I select after loading into the schema viewer is not properly highlighted.

maxxxxxdlp:
> > Unfortunately the first item I select after loading into the schema viewer is not properly highlighted.

Can you elaborate on what you mean by this / provide a video?

It works for me, albeit very laggy:

https://user-images.githubusercontent.com/40512816/212702291-36da9b0c-b26e-43ea-9ede-fd05109bedf8.mov

grantfitzsimmons:
> Can you elaborate on what you mean by this / provide a video?

https://github.com/specify/specify7/pull/2816#issuecomment-1383289115

maxxxxxdlp:
> > Can you elaborate on what you mean by this / provide a video?
> 
> [#2816 (comment)](https://github.com/specify/specify7/pull/2816#issuecomment-1383289115)

Sorry! I missed it

grantfitzsimmons:
When I’m on a data entry form, Specify uses ~500-600MB of memory – the schema viewer takes 1GB and clicking on a button spikes to using 25-35% of my CPU

The [schema print-out](https://files.specifysoftware.org/schema/version/2.10/) uses 100MB of memory and 0.6% of my CPU when I jump to a table

maxxxxxdlp:
yeah, it has a lot of nodes so that's about what I would expect

the printout page is static and does not use react (in the schema viewer there is a react component for each cell, and they all take memory)

See https://github.com/specify/specify7/pull/2816#issuecomment-1384172449 for potential solutions

grantfitzsimmons (approved):
It is up to the team to decide if we should implement a new solution to make this work without consuming a lot of memory. I would argue that it is not often used or referenced, making it a low priority to optimize.

maxxxxxdlp:
Let's push this as is before it gets too big.
#1831 mentions 3 more improvements (which are smaller) and can be implemented in the future.

maxxxxxdlp (commented):
👍

PR #2818, opened 2023-01-13
Add pref to display author name in Taxon tree viewer

### LINKED ISSUE #1121: Taxon author not displaying in the tree
**Describe the bug**
When Specify 6 has "Display Taxon Author in the tree view" checked in Preferences, the author of each taxon is displayed in the tree. In 7, this behavior is not replicated and there is no way to enable this.

**To Reproduce**
1. Enable "Display Taxon Author in the tree view" in Specify 6.
image
2. View tree in 6 to verify this change was successful.
3. View tree in 7.

**Expected behavior**
The author should display after each taxon entry.

**Screenshots**

sp6

sp7

 - OS: Windows 10
 - Browser: Edge
 - Specify 7 Version: 7.6.1

**Reported By**
Unitec

melton-jason:
Fixes #1121

Addes a new remote preference `TaxonTreeEditor.DisplayAuthor`, which when enabled displays the author next to the full name in the Taxon TreeViewer

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/InitialContext/remotePrefs.ts
+++ b/specifyweb/frontend/js_src/lib/components/InitialContext/remotePrefs.ts
@@ -189,6 +189,13 @@ export const remotePrefsDefinitions = f.store(
         parser: 'java.lang.Long',
         isLegacy: true,
       },
+      'TaxonTreeEditor.DisplayAuthor' : {
+        description:
+        "Display Authors of Taxons' next to nodes in the Tree Viewer",
+        defaultValue: false,
+        parser: 'java.lang.Boolean',
+        isLegacy: true,
+      },
       'attachment.is_public_default': {
         description: 'Whether new Attachments are public by default',
         defaultValue: true,
--- a/specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
@@ -104,10 +104,11 @@ function TreeView<SCHEMA extends AnyTree>({
   const baseUrl = `/api/specify_tree/${tableName.toLowerCase()}/${
     treeDefinition.id
   }`;
+  const doIncludeAuthorPref = getPref(`TaxonTreeEditor.DisplayAuthor`);
 
   const getRows = React.useCallback(
     async (parentId: number | 'null') =>
-      fetchRows(`${baseUrl}/${parentId}/${sortField}`),
+      fetchRows(`${baseUrl}/${parentId}/${sortField}/${doIncludeAuthorPref}`),
     [baseUrl, sortField]
   );
 
@@ -291,6 +292,7 @@ function TreeView<SCHEMA extends AnyTree>({
               row={row}
               setFocusedRow={setFocusedRow}
               synonymColor={synonymColor}
+              treeName={tableName}
               onAction={(action): void => {
                 if (action === 'next')
                   if (rows[index + 1] === undefined) return undefined;
--- a/specifyweb/specify/tree_views.py
+++ b/specifyweb/specify/tree_views.py
@@ -35,7 +35,8 @@ def tree_mutation(mutation):
 
 @login_maybe_required
 @require_GET
-def tree_view(request, treedef, tree, parentid, sortfield):
+def tree_view(request, treedef, tree, parentid, sortfield, includeAuthor=False):
+
     """Returns a list of <tree> nodes with parent <parentid> restricted to
     the tree defined by treedefid = <treedef>. The nodes are sorted
     according to <sortfield>.
@@ -60,6 +61,7 @@ def tree_view(request, treedef, tree, parentid, sortfield):
                               node.rankId,
                               node.AcceptedID,
                               accepted.fullName,
+                              node.author if (includeAuthor and tree=="taxon") else "NULL",
                               sql.functions.count(child_id)) \
                         .outerjoin(child, child.ParentID == id_col) \
                         .outerjoin(accepted, node.AcceptedID == getattr(accepted, node._id)) \
@@ -68,7 +70,6 @@ def tree_view(request, treedef, tree, parentid, sortfield):
                         .filter(node.ParentID == parentid) \
                         .order_by(orderby)
         results = list(query)
-
     return HttpResponse(toJson(results), content_type='application/json')
 
 @login_maybe_required
--- a/specifyweb/specify/urls.py
+++ b/specifyweb/specify/urls.py
@@ -32,7 +32,7 @@ urlpatterns = [
         url(r'^(?P<id>\d+)/desynonymize/$', tree_views.desynonymize),
         url(r'^(?P<parentid>\d+)/predict_fullname/$', tree_views.predict_fullname),
         url(r'^(?P<treedef>\d+)/(?P<parentid>\w+)/stats/$', tree_views.tree_stats),
-        url(r'^(?P<treedef>\d+)/(?P<parentid>\w+)/(?P<sortfield>\w+)/$', tree_views.tree_view),
+        url(r'^(?P<treedef>\d+)/(?P<parentid>\w+)/(?P<sortfield>\w+)/(?P<includeAuthor>\w+)$', tree_views.tree_view),
         url(r'^repair/$', tree_views.repair_tree),
     ])),
 
--- a/specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
@@ -9,6 +9,7 @@ import type { RA } from '../../utils/types';
 import { Button } from '../Atoms/Button';
 import { icons } from '../Atoms/Icons';
 import { useId } from '../../hooks/useId';
+import { getPref } from '../InitialContext/remotePrefs';
 
 export function TreeRow({
   row,
@@ -27,6 +28,7 @@ export function TreeRow({
   onAction: handleAction,
   setFocusedRow,
   synonymColor,
+  treeName,
 }: {
   readonly row: Row;
   readonly getRows: (parentId: number | 'null') => Promise<RA<Row>>;
@@ -49,6 +51,7 @@ export function TreeRow({
   readonly onAction: (action: Exclude<KeyAction, 'child' | 'toggle'>) => void;
   readonly setFocusedRow: (row: Row) => void;
   readonly synonymColor: string;
+  readonly treeName: string;
 }): JSX.Element {
   const [rows, setRows] = React.useState<RA<Row> | undefined>(undefined);
   const [childStats, setChildStats] = React.useState<Stats | undefined>(
@@ -124,6 +127,8 @@ export function TreeRow({
   const parentRankId = path.at(-1)?.rankId;
   const id = useId('tree-node');
   const isAction = actionRow === row;
+
+  const doIncludeAuthorPref = getPref(`TaxonTreeEditor.DisplayAuthor`);
 
   const handleRef = React.useCallback(
     (element: HTMLButtonElement | null): void => {
@@ -218,7 +223,11 @@ export function TreeRow({
                       : undefined
                   }
                 >
-                  {row.name}
+                  {doIncludeAuthorPref &&
+                  treeName === 'Taxon' &&
+                  typeof row.author === 'string'
+                    ? row.name + ' (' + row.author + ')'
+                    : row.name}
                   {typeof row.acceptedId === 'number' && (
                     <span className="sr-only">
                       <br />
@@ -295,6 +304,7 @@ export function TreeRow({
               row={childRow}
               setFocusedRow={setFocusedRow}
               synonymColor={synonymColor}
+              treeName={treeName}
               onAction={(action): void => {
                 if (action === 'next')
                   if (typeof rows[index + 1] === 'object')
--- a/specifyweb/frontend/js_src/lib/components/TreeView/helpers.ts
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/helpers.ts
@@ -22,7 +22,8 @@ export const fetchRows = async (fetchUrl: string) =>
         number,
         number | null,
         string | null,
-        number
+        string,
+        number,
       ]
     >
   >(fetchUrl, {
@@ -40,7 +41,8 @@ export const fetchRows = async (fetchUrl: string) =>
           rankId,
           acceptedId = undefined,
           acceptedName = undefined,
-          children,
+          author = undefined,
+          children
         ],
         index,
         { length }
@@ -53,6 +55,7 @@ export const fetchRows = async (fetchUrl: string) =>
         rankId,
         acceptedId,
         acceptedName,
+        author,
         children,
         isLastChild: index + 1 === length,
       })


grantfitzsimmons:
🤯

grantfitzsimmons:
Wow. It works great @melton-jason. 

image

You have 100% blown my mind with this change. Great job.

grantfitzsimmons (approved):

melton-jason:
If more customization is needed, we can expand the preference to allow the user to define how they want the author display. (Either with a separator or with an isolated 'format'. In this case, I would have the 'formatter' as ` (#)`. We would just have a fixed character act as a placeholder for the author name) 

In fact, this can be generalized even further for _every tree_ to enable this functionality for any field within the Tree's Table. (Like how the `sp7.allow_adding_child_to_synonymized_parent.TREENAME` pref can accept any tree). 

That may just be me succumbing to feature creep. Is this is fine for users, then no need to overcomplicate things.

melton-jason:
@maxxxxxdlp 
If this implementation is not satisfactory (I did not like adding a new parameter to the `tree_view` enpoint, that is sure to eventually get messy), we can have the backend read the remote prefs (as far as I could see, having the preferences in a format the backend could easily use does not exist currently) and use that instead of defining a new parameter for the endpoint.

If this might be better, I have already toyed with this idea and have a working solution given `Remote Preferences` and `Global Preferences` are unique per database (they are, correct?)

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -104,10 +104,11 @@ function TreeView<SCHEMA extends AnyTree>({
   const baseUrl = `/api/specify_tree/${tableName.toLowerCase()}/${
     treeDefinition.id
   }`;
+  const doIncludeAuthorPref = getPref(`TaxonTreeEditor.DisplayAuthor`);
 
   const getRows = React.useCallback(
     async (parentId: number | 'null') =>
-      fetchRows(`${baseUrl}/${parentId}/${sortField}`),
+      fetchRows(`${baseUrl}/${parentId}/${sortField}/${doIncludeAuthorPref}`),
```
maxxxxxdlp, line 111: This should be a query string parameter.
Consider refactoring to
```js
fetchRows(formatUrl(`${baseUrl}/${parentId}/${sortField}/`, {includeAuthor: doIncludeAuthorPref}))
```

specifyweb/frontend/js_src/lib/components/InitialContext/remotePrefs.ts
```diff
@@ -189,6 +189,13 @@ export const remotePrefsDefinitions = f.store(
         parser: 'java.lang.Long',
         isLegacy: true,
       },
+      'TaxonTreeEditor.DisplayAuthor' : {
```
maxxxxxdlp, line 192: Can we change this to accept a rank id instead?
I.e, display authors for all nodes in the tree at that rank id or further down.

Reasoning:
People might want to customize this for performance reasons.
People might want to customize this for purpose of reducing the visual clutter in the tree viewer

maxxxxxdlp (changes requested):
Quite good. Some changes

maxxxxxdlp:
@melton-jason yes, Globals&Remote prefs are unique for the entire database
There should be a util on the back-end already for reading them (take a look at other usages)

Though, making front-end pass that as a parameter is a better solution in my opinion (kind of like you are doing)
Because:
- Back-end does not need to parse the entire remote prefs file for each request
- Front-end can customize the behavior without the need for the back-end to be aware of it (it would just follow what the request says)
- third party API calls can choose on their own whether they want to see the authors

specifyweb/specify/urls.py
```diff
@@ -32,7 +32,7 @@
         url(r'^(?P<id>\d+)/desynonymize/$', tree_views.desynonymize),
         url(r'^(?P<parentid>\d+)/predict_fullname/$', tree_views.predict_fullname),
         url(r'^(?P<treedef>\d+)/(?P<parentid>\w+)/stats/$', tree_views.tree_stats),
-        url(r'^(?P<treedef>\d+)/(?P<parentid>\w+)/(?P<sortfield>\w+)/$', tree_views.tree_view),
+        url(r'^(?P<treedef>\d+)/(?P<parentid>\w+)/(?P<sortfield>\w+)/(?P<includeAuthor>\w+)$', tree_views.tree_view),
```
maxxxxxdlp, line 35: Could you also define OpenAPI schema for this endpoint?
Others too if you have time

specifyweb/frontend/js_src/lib/components/InitialContext/remotePrefs.ts
melton-jason: Yes I can change it to accept a rankid. Could we change this to accept a rank name instead? That may be more intuitive to users than an id. 
Currently, the first thought/easiest way I had of getting the rankid would probably be to setup a query like this. Even this may be too complicated to figure out.
taxonRankQuery

How much do we assume for users that can modify the remote preferences?

melton-jason: What would we want the default for this to be? 180 (Genus)? Or 220 (Species)?
Is it possible these rankids change if there are custom ranks?

maxxxxxdlp: I suggested using rank id only because it's already used by other remote prefs. Maybe because with rank IDs you can easily do a `>` comparison to see if a node is above or below the fold. Since rank id is a number, it also doesn't have case sensitivity problems. As far as getting rank id, I think Specify is using standard IDs for them that would be familiar to techy biologists.

It's your choice whether you want to use rank id or rank name

Have to ask @grantfitzsimmons for what the default value should be.

melton-jason: Since we are changing the functionality, perhaps the name of the pref could be changed as well. 
How about something like `TaxonTreeEditor.DisplayAuthor.ForRanks`? 
I am open for ideas. If the current name is ok, then no need to chang it.

melton-jason:
Now, instead of taking a boolean value, the preference accepts a number. The author should only display for nodes whose rankid is >= the preference value.

For example, 
`TaxonTreeEditor.DisplayAuthor=180`

This would -in a default tree- only display the author for ranks Genus and below.

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
melton-jason: Fixed in https://github.com/specify/specify7/commit/2517904783158a6c2eea240956937f9f7f9368d8

specifyweb/frontend/js_src/lib/components/InitialContext/remotePrefs.ts
grantfitzsimmons: Author should be shown for Genus and below by default

specifyweb/specify/tree_views.py
```diff
@@ -51,6 +51,23 @@ def tree_view(request, treedef, tree, parentid, sortfield, includeAuthor=False):
...
+    
+    """
+        If the request wants to include an author in the response, we use a case to determine
+        if the rankId is >= the supplied rank, and if so then we add node.author to the query statement.
+        If the tree is not the taxon tree or the rankId is less than the supplied rank, add NULL instead.
+
+        For more information, view the sqlalchemy docs on case:
+            https://docs.sqlalchemy.org/en/20/core/sqlelement.html#sqlalchemy.sql.expression.case
+    """
+    includeAuthorAfterRank= int(request.GET.get('includeauthorafterrank')) if 'includeauthorafterrank' in request.GET else 99_999
```
maxxxxxdlp, line 63: Can you rename this to `includeAuthor`? No need to be more verbose

specifyweb/frontend/js_src/lib/components/InitialContext/remotePrefs.ts
maxxxxxdlp: Current name is fine.
`TaxonTreeEditor.DisplayAuthor.ForRanks` is misleading as you only specify one rank id as a value.
`TaxonTreeEditor.DisplayAuthor.UnderRanks ` may be better expressed as `TaxonTreeEditor.DisplayAuthor.BelowRanks` - feel free to rename it if you want

specifyweb/specify/tree_views.py
melton-jason: Fixed in https://github.com/specify/specify7/pull/2818/commits/b4a084f0cfdb2c133782f709cb8be62e5cef98a0

melton-jason: Fixed in https://github.com/specify/specify7/pull/2818/commits/8bdc5b3e50c4355941ff7e544e53d2fb332f7e9f

specifyweb/specify/urls.py
melton-jason: Fixed in https://github.com/specify/specify7/commit/b4a084f0cfdb2c133782f709cb8be62e5cef98a0

maxxxxxdlp (approved):
Thanks for making the changes!

maxxxxxdlp:
@grantfitzsimmons can you test this again using the latest changes? If all is good we can merge this

grantfitzsimmons:
After the latest changes, I'm not seeing the author in the tree in my local copy of the `herb_rbge` database.

Screenshot 2023-01-15 at 5 31 28 PM

On `coldFish`, it works fine locally.
image

I have two takeaways from testing this on a variety of databases
- We should not place `( )` parenthesis around taxon authors. This is because some collections already have those in the author field, resulting in situations where it displays as `((Author))` across many nodes
- RBGE's database is not showing it correctly for whatever reason. Genus in their database is still `RankID` `180`.

grantfitzsimmons (changes requested):
After these changes, I think it is ready to merge. We need to ensure it works on a database as large as Edinburgh's.

melton-jason:
@grantfitzsimmons 
Do we want to follow Specify 6 logic then and just add a space between the taxon name and author?

grantfitzsimmons:
> @grantfitzsimmons Do we want to follow Specify 6 logic then and just add a space between the taxon name and author?

For this release, I think so. I think being able to customize this in the future would be great once we have a UI for customizing remote user prefs.

melton-jason:
> After the latest changes, I'm not seeing the author in the tree in my local copy of the herb_rbge database.

Interesting. What value did you have for the preference? It seems to be working correctly for this branch on the testpanel for  database `herbrbge6802`

grantfitzsimmons (approved):
Cannot recreate the issue anymore, might be related to system performance

grantfitzsimmons:
@melton-jason Can we merge this?

melton-jason:
There is still the issue regarding parsing the value near the top of some remote prefs, but that is the only concern. If that is not too big of a conern, then yes

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -101,13 +102,19 @@ function TreeView<SCHEMA extends AnyTree>({
...
 
+  const includeAuthor = getPref(`TaxonTreeEditor.DisplayAuthor`);
+
   const baseUrl = `/api/specify_tree/${tableName.toLowerCase()}/${
     treeDefinition.id
   }`;
 
   const getRows = React.useCallback(
     async (parentId: number | 'null') =>
-      fetchRows(`${baseUrl}/${parentId}/${sortField}`),
```
maxxxxxdlp, line 110: Oh, you added `/` at the end of the URL!
This fixes a small bug (needles redirect) and thus will improve performance.

maxxxxxdlp:
Actually there is a problem here:
Since sp6 expects `TaxonTreeEditor.DisplayAuthor` to be set to a boolean, not a number, looking at the Java code, it will parse a numeric value as `false`, which will mean you won't be able to have `TaxonTreeEditor.DisplayAuthor` enabled both in sp6 and sp7.

@melton-jason we need to either rename the pref in sp7 or revert it back to accepting a boolean

maxxxxxdlp:
The remote prefs parsing should be fixed now. Needs testing:
- try to replicate the original bug with prefs not being parsed correctly
- look at the list of prefs that sp7 looks at: https://github.com/specify/specify7/blob/b76572a358c8226109b490fdef5e66b5bc1850c6/specifyweb/frontend/js_src/lib/components/InitialContext/remotePrefs.ts#L115-L291
  try the following:
    putting a pref on the first/last line
    putting a pref immediately following a commented line

melton-jason:
Well will users usually just want to display the author for every node, or does only specific ranks matter?

Personally,I think there is some merit in keeping the rankid preference, so perhaps renaming it in 7? 
We _could_ have two prefernces then in Specify 7: one with the new functionality and a legacy one with the old. Although this may confusing and unnecessary.

maxxxxxdlp:
Since it never worked in sp7, users won't necessarily expect it to work in sp7 using the same remote pref (especially if it accepts a different argument in sp7)

As far as renaming it, would `TaxonTreeEditor.DisplayAuthor.BelowRankId` or `TaxonTreeEditor.DisplayAuthor.AboveRankId` work? (I don't know whether Below or Above makes more sense in this context)

Maybe explicitly `TaxonTreeEditor.DisplayAuthor.ForRankIdGreater` is the best choice

grantfitzsimmons:
The boolean value from 6 should be respected IMO.

The RankID option is a (mostly) unnecessary feature. If we are to keep it, it should just be an additional pref on top of whether it is enabled or not ideally.

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
maxxxxxdlp: Correct. In fact, if you find any other back-end API calls that don't have a trailing `/`, you should add it.

maxxxxxdlp:
Okay.
I was thinking that rank id would improve performance because back-end won't have to calculate the authors, but looking at the code, it seems like it won't make a performance difference.

The only remaining reason I see for wanting to specify a rank id, is if you have authors for Kingdom, Phylum and other top level tree nodes, but you don't actually want to see those in the UI (you only want to see them for Species/Subspecies). @grantfitzsimmons is that a valid use case?

grantfitzsimmons:
> The only remaining reason I see for wanting to specify a rank id, is if you have authors for Kingdom, Phylum and other top level tree nodes, but you don't actually want to see those in the UI (you only want to see them for Species/Subspecies).

I _could_ see that as a request, but I haven't heard anyone ask for that functionality

maxxxxxdlp:
Okay. Then how about we just implement it as a boolean and by the same name as in sp6 so that it "magically" starts working for people who have it configured in sp6.

And in the future, when we get rid of remoteprefs/globalprefs in favor of a more sane solution, there would be a migration step that could also convert the boolean to a rank id.

grantfitzsimmons:
> Okay. Then how about we just implement it as a boolean and by the same name as in sp6 so that it "magically" starts working for people who have it configured in sp6.

I think that's a great idea. Users who had the functionality before have complained that it was missing in Specify 7.

> And in the future, when we get rid of remoteprefs/globalprefs in favor of a more sane solution, there would be a migration step that could also convert the boolean to a rank id.

Yes please. That would be great!

melton-jason:
Ok, I reverted the preference back to accept a boolean and in both backend and frontend code comments linked this pull request if context is needed.

maxxxxxdlp:
Code looks good. Now just need someone to test this once again.

@chanulee1  would you have time to test this issue, and also https://github.com/specify/specify7/pull/2818#issuecomment-1386472569?

chanulee1 (changes requested):
The pref appears to be active on default. Can only be disabled by explicitly setting it to false.

melton-jason:
Should be fixed now

chanulee1:
Everything looks good. One request: Could we add something to differentiate between the taxon name and author name? It can be a little difficult to tell them apart. Disregard if it's not a priority or becomes too involved.

acbentley:
What do you mean - different font, different color?

Andy
    A  :             A  :             A  :
 }<(((_°>.,.,.,.}<(((_°>.,.,.,.}<)))_°>
    V                V                V
Andy Bentley
Ichthyology Collection Manager
University of Kansas
Biodiversity Institute
Dyche Hall
1345 Jayhawk Boulevard<x-apple-data-detectors://9/>
Lawrence, KS, 66045-7561<x-apple-data-detectors://9/>
USA<x-apple-data-detectors://9/>

Tel: (785) 864-3863<tel:%28785%29%20864-3863>
Fax: (785) 864-5335<tel:%28785%29%20864-5335>
Email: ***@***.******@***.***>
ORCID: https://orcid.org/0000-0002-3093-1258
http://ichthyology.biodiversity.ku.edu<http://ichthyology.biodiversity.ku.edu/>
    A  :             A  :             A  :
 }<(((_°>.,.,.,.}<(((_°>.,.,.,.}<)))_°>
    V                V                V

chanulee1:
Yes, either of those would be great. Just as a temporary solution.

grantfitzsimmons:
@acbentley @chanulee1 

Can we leave it as-is (just like Specify 6) for this first release? We can enhance this soon :smile:

chanulee1:
Yes that's fine with me!

grantfitzsimmons:
@chanulee1 If you have tested since the fix can you approve the PR?

chanulee1 (approved):
Appears to be working. Tested remote prefs parsing as well.

grantfitzsimmons (approved):
I agree, a better separator between the taxon name and author should be added. Maybe a dash, parenthesis as we had before, or a custom option? For now, matching 6 is good.

PR #2819, opened 2023-01-13
Replace PNG Table Icons with SVG icons

### LINKED ISSUE #2368: New table icons for Specify 7
Specify 7 still relies on the Specify 6 table icons that are stored at an extremely low resolution. Ideally, Specify 7 should store table icons in the front end for every table in better resolution.

### LINKED ISSUE #2772: Use dark scrollbars in dark mode
image

https://stackoverflow.com/questions/65940522/how-do-i-switch-to-chromes-dark-scrollbar-like-github-does

CarolineDenis:
Fixed #2368
Fixed #2772


grantfitzsimmons:
This seems to be a work in progress but it’s a good start!

maxxxxxdlp:
Preliminary autogenerated icons are now done. While we are still evaluating the new icons, the home page has been replaced by an icon gallery.

To revert the change, resolve all the `// FIXME: ` comments in the code and the fix typescript errors.

Here are the results:
[Sp7_icon_comparsion.pdf](https://github.com/specify/specify7/files/10416486/Sp7_icon_comparsion.pdf)
(the 100% zoom level will display them in the same size as would be most commonly used in specify)


Note, if an icon was previously missing, I didn't try to keep it consistent
Also, if icon was previously inconsistent (*preparation, *citation, *permit), I made them consistent


@grantfitzsimmons I invite you to make the edits:

This defines the names for icons:
https://github.com/specify/specify7/blob/2091468729d36c63ace6724e8e96d51880437fa5/specifyweb/frontend/js_src/lib/components/Molecules/SvgIcon.tsx#L150-L194

If name is not in the list, the following rule is used:
If table name begins with Sp, `Sp` is dropped
If table name ends with Attachment and it's not literally an `Attachment` table, `Attachment` is dropped.
If table name contains more than one capital letter, the first 3 capital letters are used as a name.
Otherwise, the first 3 letters of the name are used.

As far as colors, it's way harder to come up with such broad rules, so I ended up defining each table:
https://github.com/specify/specify7/blob/2091468729d36c63ace6724e8e96d51880437fa5/specifyweb/frontend/js_src/lib/components/Molecules/SvgIcon.tsx#L215-L337

grantfitzsimmons:
Here are the screenshots that stood out to me as icons that could use tweaking

IMG_1325
IMG_1326
IMG_1327
IMG_1328
IMG_1329
IMG_1330
IMG_1331
IMG_1332
IMG_1333
IMG_1334
IMG_1335

grantfitzsimmons:
I can fix all the items in the above comment

grantfitzsimmons:
Everything above has been resolved but I'm conflicted. I like the full "Form" text for view set objects but it definitely is too wide for the icon. What do you think? Worth the appearance for how much it communicates?

melton-jason:
> Everything above has been resolved but I'm conflicted. I like the full "Form" text for view set objects but it definitely is too wide for the icon. What do you think? Worth the appearance for how much it communicates?

On an information standpoint, I like the full `Form` as well. In my opinion, the most important thing to consider is if `Frm` will communicate enough. 

In these changes, there are other tables we have which display four characters: 
- CollectingEventAuthorization
- CollectingTripAuthorization
- DNAPrimer
- DNASequencingRun / DNASequencingRunAttachment

So having another display `Form` may be ok. Then again, those tables listed above do not include a full word, which does _somehow_ make it look like a tighter fit.  

Ultimately, I am ambivalent about either choice. Both decisions have their own pros and cons to consider,

maxxxxxdlp:
As far as I know, the SpViewSetObj table icon is not exposed anywhere outside of the data model viewer so either way whatever we decide won't have much impact.

@grantfitzsimmons you can try playing around with the font size if you don't want the text to overflow

grantfitzsimmons:
I think it looks great. Now I'd like to see how it looks in place of the existing PNGs everywhere in the application :smile:

specifyweb/frontend/js_src/lib/components/Molecules/SvgIcon.tsx
```diff
@@ -150,30 +150,44 @@ const endsWith = <T,>(
 const nameMapper = f.store<Partial<RR<keyof Tables, string>>>(() => ({
   ...startsWith('Accession', 'Acc'),
   ...startsWith('Attachment', 'Att'),
+  ...startsWith('Disposal', 'Dsp'),
+  ...startsWith('FieldNotebookPageSet', 'FnS'),
+  ...startsWith('CollectingEventAttr','CeA'),
   ...endsWith('Agent', 'Agt'),
   ...endsWith('Citation', 'Cit'),
-  ...endsWith('Authorization', 'Per'),
+  ...endsWith('Authorization', 'Acc'),
```
maxxxxxdlp, line 159: This conflicts with the shortcut for `Accession`

maxxxxxdlp:
@grantfitzsimmons the new icons are now enabled everywhere on this branch

grantfitzsimmons:
image

Is this misaligned icon inevitable when zoomed in?

grantfitzsimmons:
I think this looks great. What left is there to consider before bringing this into Specify?

maxxxxxdlp:
The only thing remaining is the font-size discussion

grantfitzsimmons:
I vote for leaving the font size system the same. Do we want to discuss this with the team tomorrow?

CarolineDenis:
Were you able to add a minimum margin around the text?

grantfitzsimmons:
> Were you able to add a minimum margin around the text?

I tried it but I prefer the small margins for readability. It is already hard to see the icons from a regular distance. Is it okay if we maintain the Specify 6 look for this release?

maxxxxxdlp (approved):

maxxxxxdlp:
@grantfitzsimmons could you take one last look at this branch now that it is production ready

To test:
1. Welcome page is returned to normal
2. New table icons are used everywhere
3. Old icons still work (in reports dialog and weblink)

After that, we can merge this

specifyweb/frontend/js_src/lib/components/Molecules/SvgIcon.tsx
maxxxxxdlp: @grantfitzsimmons did you want to fix this?

grantfitzsimmons (approved):

PR #2825, opened 2023-01-16
Fix incorrect comparison for "5A" being equal to "5B"

### LINKED ISSUE #2805: Pick list not allowing the selection of other items with the same first character
https://user-images.githubusercontent.com/37256050/212190227-a39e1b78-7523-4ada-9701-aec381234542.mov

https://herbrbge6802-xml-validation.test.specifysystems.org/specify/view/collectionobject/1178963/

_This can be recreated in `edge`_

See the "Filing Region" (`text4`) pick list in the Collecting Event table.

When you select an item such as `6E - New Guinea` and try to select another item that starts with `6`, you will be unable to select it. You need to first select an item without a 6 at the start and then select the item you wish to choose.

**Reported By:** Robyn at RBGE

maxxxxxdlp:
Fixes #2805

Caused by 24eec2ffc4 (from 8 months ago)
This happens for all fields, but only if the old value for that field starts with a number (`5A`) and the new value of that field starts with a number (`5B`).
**The correct new value is stored in the database**. The bug only affects the visual display.

To test:
- Change the value from `5A` to `5B`. Save button should be triggered
- Open a record that has whitespace in a text field. Trimming white space should not trigger the save button (front-end component might trim the white space automatically)
- Open any record that has a decimal field (with some value in it). The save button should not be triggered

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.js
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.js
@@ -225,8 +225,9 @@ function eventHandlerForToOne(related, field) {
                    * string to number (back-end sends certain numeric fields
                    * as strings. Front-end converts those to numbers)
                    * REFACTOR: this logic should be moved to this.parse()
+                   * TEST: add test for "5A" case
                    */
-                  Number.parseInt(oldValue) === Number.parseInt(newValue)
+                  oldValue.toString() === newValue.toString()
                 )
                     options ??= {silent: true};
                 else if(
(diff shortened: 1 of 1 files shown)


grantfitzsimmons:
https://user-images.githubusercontent.com/37256050/212772126-dd8b3681-6446-410c-8a19-9df1ff00d039.mov

Looks to be working perfectly. I tested this with existing decimal values and white space and it did not trigger the save button.

chanulee1 (approved):
Looks good. Tested whitespace and decimal values as well.

PR #2836, opened 2023-01-17
Display RSS export file name in lowercase

### LINKED ISSUE #1821: Text should not be autocapitalized for feed export
image

This should say `kuit-dwca.zip` rather than `Kuit-Dwca.Zip`

CarolineDenis:
Fixed #1821

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
@@ -23,7 +23,7 @@ export const notificationRenderers: IR<
       <>
         {notificationsText.feedItemUpdated()}
         <Link.Green
-          className="w-fit"
+          className="w-fit normal-case"
           download
           href={`/static/depository/export_feed/${filename}`}
         >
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -23,7 +23,7 @@ export const notificationRenderers: IR<
       <>
         {notificationsText.feedItemUpdated()}
         <Link.Green
-          className="w-fit"
+          className="w-fit normal-case"
```
maxxxxxdlp, line 26: This looks good. Are there any other links/buttons in that file that should also have `normal-case` added to them?

CarolineDenis:
In the notification dialog we can find 7 different links/buttons. 
- Feed item updated (=> displays the file name to be downloaded) 
- Feed update failed (=> displays "Error")
- Dwca export complete (=> displays "Download") 
- Dwca  failed (=> displays "Error")
- Query export to CSV complete (=> displays "Download") 
- Query export to KML complete (=> displays "Download") 
- Dataset ownership transferred (=> displays the dataset name that was set during creation which makes sense to me) 

I suggest that we keep the capital letter for Error and Download and the lower case for the feed file.

maxxxxxdlp:
That works

A note that `normal-case` class name just disables converting the first letter to upper case. But if the first letter was already in upper case, it won't put it into lower case.

grantfitzsimmons (approved):
Works as expected. Merge if this solution is satisfactory :smile:

PR #2837, opened 2023-01-17
Add one more link to Discourse localization post

maxxxxxdlp:
The link was there when picking a language that is not yet finished.

This commit also updates the "Help Localize Specify 7" dialog message to add a link to Discourse post regarding localization


chanulee1 (approved):
Looks good.

PR #2857, opened 2023-01-20
Improve WB import date parsing

### LINKED ISSUE #2027: Uploading data from Excel with formatted dates gives an error
I've experienced this many times, and Mandi has reported the same yesterday. 

If a cell is formatted as a date, the WorkBench ignores its configuration when its uploaded and the dates are read as invalid. I have had to convert cells to text to workaround this issue in the past.

maxxxxxdlp:
When importing a spreadsheet though WorkBench, date formats like `10/10/20` are now resolved into the format configured in remote prefs (using the `ui.formatting.scrdateformat` property)

Fixes #2027

Spreadsheet I used for testing: 
[date.format.testing (1).xlsx](https://github.com/specify/specify7/files/10467749/date.format.testing.1.xlsx)
(this is an extended version of the one sent by @danb213)

More information on the date & time parsing in the library we are using: https://docs.sheetjs.com/docs/csf/features/dates/

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/PartialDateUi.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/PartialDateUi.tsx
@@ -1,10 +1,6 @@
 import React from 'react';
 
-import {
-  databaseDateFormat,
-  fullDateFormat,
-  monthFormat,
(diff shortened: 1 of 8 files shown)


chanulee1 (changes requested):
See #2926. Other than that, looks good.

PR #2863, opened 2023-01-21
Allow for conditional rendering in forms

### LINKED ISSUE #2860: Allow conditionally rendering form fields
`<panel>` element should be extended to allow rendering form conditionally (display some fields based on the value in a different field)

[@melton-jason](https://github.com/melton-jason) Can you double-check what the form system in sp6 does if it encounters a panel cell with more than one <rows> tag?
Example of a panel cell: https://github.com/specify/specify7/wiki/Form-System#panel
Looking at the code, it looks like it would take the first one, but would be good to double-check.

maxxxxxdlp:
Fixes #2860

Documentation: https://github.com/specify/specify7/wiki/Form-System#conditional-rendering
Demo:

https://user-images.githubusercontent.com/40512816/213841543-a6a7eb0c-829a-41f9-949f-7fc1cc62fe2b.mov

To test:
1. Test that the behavior in sp7 matches what is mentioned in the documentation.
2. Double-check what the form system in sp6 does if it encounters a panel cell with more than one `<rows>` tag?
    [Looking at the code](https://github.com/specify/specify7/issues/2860#issuecomment-1399169461), it looks like it would take the first one, but would be good to double-check.
    Also, make sure there are no relevant warnings emitted in the console.


chanulee1 (changes requested):
See linked issues #2922, #2923, #2925.

chanulee1 (approved):
Looks good so far. Behavior is consistent with documentation. Sp6 treats multiple `rows` elements in a panel element as expected: it only accepts the first one.

grantfitzsimmons:
@maxxxxxdlp This is good to go once conflicts are resolved it seems

PR #2872, opened 2023-01-24
Do not trigger unload protect if value did not change

### LINKED ISSUE #2868: IntegrityError when creating a new taxon record
image

To recreate, fill the form out as follows:

image

[Specify 7 Crash Report - 2023-01-23T04 44 43.025Z.txt](https://github.com/specify/specify7/files/10476828/Specify.7.Crash.Report.-.2023-01-23T04.44.43.025Z.txt)

```
IntegrityError at /api/specify/taxon/
(1048, "Column 'IsHybrid' cannot be null")
```

This was introduced in `v7.8.4` as I can not recreate in 7.8.3.

@melton-jason

melton-jason:
Fixes #2868

A value should only be added to a Resource's attributes (the values sent to the backend) if it was changed. A bug in https://github.com/specify/specify7/commit/9e7e3cec2c8c5797915a0112b105e9bb303484f9 caused many more fields on all forms to register as 'changed' even if they have not, causing #2868. 

See https://github.com/specify/specify7/issues/2868#issuecomment-1401287927 for an example of how unchanged values were being added objects' responses.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.js
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.js
@@ -218,7 +218,8 @@ function eventHandlerForToOne(related, field) {
               typeof (oldValue??'') !== 'object' &&
               typeof (newValue??'') !== 'object'
             ) {
-                if(
+                if (oldValue === newValue) return this;
+                else if(
                   /*
                    * Don't trigger unload protect if:
                    *  - value didn't change


maxxxxxdlp (approved):

grantfitzsimmons (approved):

melton-jason:
@grantfitzsimmons @chanulee1 do we need to do some more testing on this pull request, or is it ready to merge? 
The sooner we can get that bug fixed, the better.

chanulee1 (approved):
Looks good. Able to create taxon record, picklist record without the null error.

PR #2878, opened 2023-01-24
Add field ID for each table with their scope

### LINKED ISSUE #2080: Schema Viewer does not show the ID field

CarolineDenis:
Fixed #2080

Notes: Scope is displayed when present in the relationship.


maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Toolbar/DataModel.tsx
```diff
@@ -234,8 +234,14 @@ function DataModelFields({
...
 }): JSX.Element {
   const data = React.useMemo(() => getFields(model), [model]);
+  const scope = React.useMemo(
+    () => model.getScopingRelationship()?.relatedModel.name,
+    []
+  );
   return (
     <>
+      <p>Field ID: {model.idField.name}</p>
+      {scope && <p>Scope: {scope}</p>}
```
maxxxxxdlp, line 244: Remember that this is bug-prone. Please refactor to use explicit check `typeof scope === 'string' && <p>...</p>`

You can also configure ESLint plugin to have it notify you of this automatically (https://github.com/specify/specify7/tree/production/specifyweb/frontend/js_src#plugins)

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Toolbar/DataModel.tsx
```diff
@@ -234,8 +235,24 @@ function DataModelFields({
   readonly model: SpecifyModel;
 }): JSX.Element {
   const data = React.useMemo(() => getFields(model), [model]);
+  const scope = React.useMemo(
```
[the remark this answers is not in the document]
maxxxxxdlp: (optional) Also, a `useMemo` here is not really needed at all as `model.getScopingRelationship()?.relatedModel.name` is not an expensive operation. You could just say `const scope = model.getScopingRelationship()?.relatedModel.name;`

maxxxxxdlp (changes requested):
Some small changes

specifyweb/frontend/js_src/.eslintrc.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
   root: true,
   parser: '@typescript-eslint/parser',
   parserOptions: {
-    project: './tsconfig.eslint.json',
+    project: '/Users/carolinedenis/Desktop/specify7/specifyweb/frontend/js_src/tsconfig.eslint.json',
```
maxxxxxdlp, line 11: Is this change necessary? It works for me without it.
And even if it is necessary, it shouldnt' be pushed as the path you provided only exists on your system

specifyweb/frontend/js_src/lib/components/Toolbar/DataModel.tsx
```diff
@@ -297,11 +311,35 @@ function DataModelRelationships({
   readonly model: SpecifyModel;
 }): JSX.Element {
   const data = React.useMemo(() => getRelationships(model), [model]);
+
+  const [filteredDependentData, setFilteredDependentData] =
+    React.useState(data);
+  const [dependentFilter, setDependentFilter] = React.useState(false);
+
+  React.useEffect(() => {
```
maxxxxxdlp, line 319: Refactor to replace useEffect with useMemo - try to avoid useEffect whenever possible
See https://beta.reactjs.org/learn/you-might-not-need-an-effect for more information

specifyweb/frontend/js_src/lib/components/Toolbar/DataModel.tsx
```diff
@@ -297,11 +311,35 @@ function DataModelRelationships({
...
+    if (dependentFilter === true) {
+      const dependentData = data
+        .filter((table) => table.isDependent === 'Yes')
```
maxxxxxdlp, line 322: 1. `table` is not a good varialbe names as tables can't be dependent - relationships can
2. This will likely not work when users change their interface language. Either modify `getRelationships` to have a two-item array for the `isDependent` option (1st item for sorting & filtering, 2nd for display), or get the relationship object from the relationship name: `model.strictGetRelationship(relationship.name[0]).isDependent`

specifyweb/frontend/js_src/lib/components/Toolbar/DataModel.tsx
```diff
@@ -297,11 +311,35 @@ function DataModelRelationships({
...
+      setFilteredDependentData(dependentData);
+    } else {
+      setFilteredDependentData(data);
+    }
+  }, [dependentFilter]);
+
   return (
     <>
-      <H3 id={model.name.toLowerCase()}>{schemaText.relationships()}</H3>
+      <div className="flex items-center">
```
maxxxxxdlp, line 332: Add `gap-4` or `justify-between` to separate the elements in the UI

specifyweb/frontend/js_src/lib/components/Toolbar/DataModel.tsx
```diff
@@ -310,6 +348,39 @@ function DataModelRelationships({
...
+  onClick: handleClick,
+  className: newClasses,
+}: {
+  readonly isPressed: boolean;
+  readonly children: LocalizedString;
+  readonly onClick: () => void;
+  readonly className?: string;
+}): JSX.Element {
+  return (
+    <button
```
maxxxxxdlp, line 363: Not good.
Don't use `<button>` - use `<Button.SomeButtonVariantHere>` instead - it comes with a bunch of styles baked in so you don't have to write the same styles every time.
With your solution, imagine if I need to change the border of the button in dark mode - now I have to look for every simple possible way where you may have used this button.

Even better - there already is a <RadioButton> component in `./components/AppResources/Filters.tsx`- you can just export it and reuse it here rather than have to duplicate the code

specifyweb/frontend/js_src/lib/components/Toolbar/DataModel.tsx
```diff
@@ -297,11 +311,35 @@ function DataModelRelationships({
...
   return (
     <>
-      <H3 id={model.name.toLowerCase()}>{schemaText.relationships()}</H3>
+      <div className="flex items-center">
+        <H3 id={model.name.toLowerCase()}>{schemaText.relationships()}</H3>
+        <DependentFilterButton
+          isPressed={dependentFilter}
+          onClick={(): void => setDependentFilter(!dependentFilter)}
+        >
+          {schemaText.showDependent()}
```
maxxxxxdlp, line 338: This name is too long for button label - especially since it would be even longer when localized to other languages.

Also, it looks like your solution does not allow to see independent relationships only.

Consider instead just having two buttons: `Dependent` / `Independent` with a similar toggle logic like in app resources.

specifyweb/frontend/js_src/lib/localization/schema.ts
```diff
@@ -289,4 +292,10 @@ export const schemaText = createDictionary({
     'ru-ru': 'Перейти к началу',
     'uk-ua': 'Перейти вгору',
   },
+  fieldID: {
```
maxxxxxdlp, line 295: All keys must be in strictCamelCase. Instead of `fieldID`, do `fieldId` - these should never be two consecutive capitalized letters.

See https://medium.com/@thomasreggi/enforcing-strict-camelcase-5332c53a7484 for some reasons why.

You can find more rules regarding localization here - https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/localization/README.md

specifyweb/frontend/js_src/lib/localization/schema.ts
```diff
@@ -289,4 +292,10 @@ export const schemaText = createDictionary({
...
+    'en-us': 'Field ID',
```
maxxxxxdlp, line 296: Change `Field ID` to `ID Field` (they mean different things)

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

PR #2904, opened 2023-01-26
Add the taxon node name in the delete dialog

### LINKED ISSUE #2854: Improve the delete taxon node confirmation dialog
> **UX FEATURE REQUEST:** Improve messaging in the "Delete a taxon node confirmation dialog":
> 
> It might improve the UX to list in the confirmation dialog the name of the Taxon node that is about to be deleted so the user can identify they did not click delete on the wrong node.
> The only indicator in the image below is that "test (0)" has an outline. I believe it is possible that this dialog might overlap other nodes, so the user would have to move the dialog around just to see. Putting the name in the dialog box would be a clearer way to confirm.

image

Moving the dialog box around:

image2

**Requested By:** Matthew Cruz at The University of Michigan

CarolineDenis:
Fixed #2854


specifyweb/frontend/js_src/.eslintrc.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
   root: true,
   parser: '@typescript-eslint/parser',
   parserOptions: {
-    project: './tsconfig.eslint.json',
+    project: '[./tsconfig.eslint.json]',
```
maxxxxxdlp, line 11: this looks like an accidental change. can you revert it?

specifyweb/frontend/js_src/README.md
```diff
@@ -82,6 +82,7 @@ development. Many IDEs have plugins for closer integration.
   Highly customizable code linter
 
   Config file is located in [`./.eslintrc`](./.eslintrc.js). The config file
+
```
maxxxxxdlp, line 85: revert this change

specifyweb/frontend/js_src/lib/components/Forms/DeleteButton.tsx
```diff
@@ -56,6 +57,7 @@ export function DeleteButton<SCHEMA extends AnySchema>({
   deferred: initialDeferred = false,
   component: ButtonComponent = Button.Gray,
   onDeleted: handleDeleted,
+  focusedName: focusName,
```
maxxxxxdlp, line 60: This works, but the solution is way too specific to this case.
Consider these things:
- DeleteButton does not know anything about the concept of "focused node" - and it shouldn't. At very least, rename the prop
- Adding a formatted node name would be great not just in the tree, but in all the places were DeleteButton is used - since you have access to the resource, you can call `format()` function to get a formatted string (see usages)

maxxxxxdlp (changes requested):

maxxxxxdlp:
Good idea!
@CarolineDenis could you also add a prop to `<FormattedResource>` to not display the formatted resource as a link (because the link does not make sense in this case - you would already have the resource you are trying to delete open)

grantfitzsimmons (approved):
Works well! Would like to see an icon if possible :smile:

maxxxxxdlp (approved):

grantfitzsimmons (changes requested):
image

The table icon is italicized just as the text is

maxxxxxdlp:
ha, it's a nice side effect of the dynamically generated icons!

maxxxxxdlp:
@CarolineDenis rather than putting the icon outside of `<i>`, can you add `not-italic font-normal` class name to the component that generates the class name so that this kind of bug does not happen anywhere else

grantfitzsimmons:
image

Looks good to me now.

grantfitzsimmons (approved):

PR #2927, opened 2023-01-31
Changed query builder buttons background in light mode

### LINKED ISSUE #2611: Query Builder's "+" button does not have background when in light mode
The background of the"+" and negate buttons for the selected query builder field clash with the selected field background:

Screenshot 2022-12-12 at 19 42 23

cdamyx:
Fixes #2611 

Simply changed background from 300 to 100 (see screenshot). Is this adequate? Would you prefer a border? Or, would it be more appropriate to change the "select" gray highlight color instead?
image
@grantfitzsimmons @maxxxxxdlp


maxxxxxdlp (changes requested):
While this change fixes this case, it may break all the other places where the gray button is used (as they expect the button to be gray, not white).

Instead, in all the places where `className.grayButton` is used in [this file](https://github.com/specify/specify7/blob/95ca93bd2d52acfdf6507f024e714ba491862078/specifyweb/frontend/js_src/lib/components/QueryBuilder/Field.tsx#L325), replace the usages with:

```js
`${className.grayButton} !bg-gray-100 dark:!bg-neutral-600`
```

(or even better - create a separate variable for that and reuse that variable)

(the `!` in the class names above is equivalent to css's `!important`)

cdamyx:
Thanks Max! I went with the new variable, but I had to add the rest of the styles for things like hover and dark mode to work properly. Please let me know if that variable needs to be located somewhere else :grin:

specifyweb/frontend/js_src/lib/components/QueryBuilder/Field.tsx
```diff
@@ -233,6 +233,9 @@ export function QueryLine({
 
   const hasAny = field.filters.some(({ type }) => type === 'any');
 
+  const lightGrayButton = `${className.grayButton} hover:!bg-gray-400 !bg-gray-100 !text-gray-800 !shadow-sm
```
maxxxxxdlp, line 236: Good job customizing all the other styles too!
Could you please move this line into `./components/Atoms/className.ts` right after the `grayButton`?

https://github.com/specify/specify7/blob/1f3adff00d01d09a3b572951f68bb52d14e2de9e/specifyweb/frontend/js_src/lib/components/Atoms/className.ts#L54

And then you can use it in the code as `className.lightGrayButton`

Also, it looks like you ended up overriding all the styles of the `className.grayButton`, so instead of overriding them, you could just start from scratch (i.e, copy the value of `grayButton` for `lightGrayButton`, but change the styles that differ, instead of using `!` to override them)

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Field.tsx
cdamyx: Thanks - I thought you might want it there :smiley: Got it changed over

maxxxxxdlp (approved):

grantfitzsimmons (approved):
Looks good!

PR #2930, opened 2023-01-31
Fix GeoLocate button having no effect

### LINKED ISSUE #2917: Cannot georeference a locality
Works in `7.8.4`, does not work in `edge`.

https://user-images.githubusercontent.com/103299204/215584398-38def5b3-b97f-4d84-b193-8b67b0c1ff16.mp4

https://coldfish-edge.test.specifysystems.org/specify/view/locality/new/

maxxxxxdlp:
Fixes #2917

Please test GeoLocate both on the locality form and in workbench.
Also, test GeoLocate in a record set (switch to a different item while geolocate window is open and see what happens)


grantfitzsimmons (approved):
💪

PR #2934, opened 2023-02-01
Allow pasting into modern date field and accept relative date input

### LINKED ISSUE #1910: Restore ability to copy dates
image

There should be a way to copy the entire date easily.

**Requested By:** Barry with the Minnesota Dragonfly Society

### LINKED ISSUE #2845: Pasting into modern date fields doesn't work
Related to #1910

There is code in place to handle Ctrl+V into a date field, but the code appears to not be working

CarolineDenis:
Fixed #2845
Fixed #1910


specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
```diff
@@ -161,46 +166,33 @@ export const Input = {
...
-            const value =
-              // @ts-expect-error globalThis.clipboardData does not have typings
-              (event.clipboardData ?? globalThis.clipboardData).getData(
-                'text/plain'
-              );
-            handleDatePaste(value);
-          } catch (error: unknown) {
-            console.error(error);
+          const parsed = parseDate('full', input.value);
+          if (parsed.isValid()) input.value = parsed.format(fullDateFormat());
```
maxxxxxdlp, line 181: Can someone please test if this works in the query builder
(i.e, whether it queries on the correct date). We might have to change `fullDateFormat()` back to `databaseFieldFormat` after all)

CarolineDenis:
Please test different places where date inputs are used.
The max length validation attribute has been removed. Test if no repercussions

maxxxxxdlp:
53/64/2011 is probably interpreted as 64 months after 2011, which is why you get 2016.

It seems like the date parsing library tries its best to interpret the date and clears the value only if the date is unknown. Not sure if that's good (makes it easier for the user) or bad (makes it harder to detect typos)

`13/2/2003` looks like a semi-valid date so probably should not be cleared.

maxxxxxdlp:
I don't know whether we should fix the remaining issues with the date parsing library we are using - there are some peculiarities with our current date parsing library. I.e, it doesn't work with the European date format WTF - I had to write a hack to make it work

We can't migrate to a new one just yet because when sys admins specify the date format in global prefs, they are using the format that is specific to this library (and compatible with sp6)

The best solution would be to let this issue be for now as it's not super urgent (and is present in production). Then, when we have time to migrate global prefs away from sp6, we do that and combine that with migration to new library that hopefully doesn't have these issues

I heard date-fns is a popular solution nowadays, but will need to do more research when we are ready to migrate

chanulee1 (approved):

specifyweb/frontend/js_src/lib/utils/parser/definitions.ts
```diff
@@ -168,7 +168,9 @@ export const parsers = f.store(
     'java.sql.Timestamp': {
       type: 'date',
       minLength: fullDateFormat().length,
-      // maxLength: fullDateFormat().length,
+      maxLength: fullDateFormat().length,
```
maxxxxxdlp, line 171: I think we needed to remove the `maxLength` constraint as part of this pull request - because it was preventing entering values like `today + 2 weeks` in the query builder. Can you remove this line?

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

PR #2948, opened 2023-02-06
Add a link to a specific data model table or user prefs category

### LINKED ISSUE #2898: Can't link to a specific data model table or user prefs category
A URL like this: https://sp7demofish.specifycloud.org/specify/data-model/#journal or https://sp7demofish.specifycloud.org/specify/data-model/journal/
does not scroll down to Journal table

Looks like we have the same bug for user preferences https://sp7demofish.specifycloud.org/specify/user-preferences/#form

CarolineDenis:
to scroll intoview

Fixed #2898

Need fix for "readonly references: React.Object<WritableArray<HTMLElement | undefined>>;"


maxxxxxdlp (approved):

grantfitzsimmons (changes requested):
https://user-images.githubusercontent.com/37256050/217884998-0a4c5cc5-f9df-41ad-942a-a18484d9221f.mov


It does not scroll properly into view– this applies to the schema viewer as well

https://coldfish-issue-2898.test.specifysystems.org/specify/user-preferences/#form

Once you load the page, it works correctly, but directly navigating to it focuses incorrectly.

FIrefox, macOS

specifyweb/frontend/js_src/lib/components/TreeView/helpers.ts
```diff
@@ -141,7 +141,7 @@ export function serializeConformation(
 const throttleRate = 250;
 export const scrollIntoView = _.throttle(function scrollIntoView(
   element: HTMLElement,
-  mode: ScrollLogicalPosition = 'center'
+  mode: ScrollLogicalPosition = 'start'
```
maxxxxxdlp, line 144: By making a change here, you change the behavior for every place that is using scrollIntoView().

Instead, when calling `scrollIntoView()`, just provide `start` as the second argument in the places where you need to

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

chanulee1:
Seems to be the same problem as before for the user prefs, but works for datamodel viewer. As stated before, it works when the page is already loaded but does not work when navigating to the url directly. Tested on Chrome, Firefox, Windows 10.

https://user-images.githubusercontent.com/103299204/219198027-7a5340f5-9a88-4fd2-86dd-0f56c1e25689.mp4

grantfitzsimmons:
After navigating to https://coldfish-issue-2898-2.test.specifysystems.org/specify/user-preferences/#header, selecting any option on the side brings me back to the Top Level Menu section:

https://user-images.githubusercontent.com/37256050/219259604-5edf3684-be42-4928-b15d-3f2f6f4466f1.mov

grantfitzsimmons (changes requested):
See my comment

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/Toolbar/DataModel.tsx
```diff
@@ -170,7 +170,7 @@ function DataModelTable({
   ) : (
     <section className="flex flex-col gap-4" ref={forwardRef}>
       <div className="flex items-center gap-4">
-        <div className="flex items-center gap-2">
+        <div className="flex items-center gap-2" id={tableName}>
           <TableIcon label={false} name={model.name} />
           <H2 className="text-2xl" id={model.name.toLowerCase()}>
```
maxxxxxdlp, line 175: Since you added ID to <div>, it should be removed from the header. (or maybe the ID should be removed from the div in favor of the header - like it was before)

maxxxxxdlp (changes requested):
Slight change

maxxxxxdlp (approved):

grantfitzsimmons (approved):

grantfitzsimmons:
Ready to go :smile:

Issue #2873, opened 2023-01-24
`origFilename` not displaying in 7


grantfitzsimmons:
> Just noticed a discrepancy in the way Specify 7 and Specify 6 are displaying attachment info.  Note the top field.  Is this intentional or a bug?

![image](https://user-images.githubusercontent.com/37256050/214329622-a6b46f19-447d-49aa-bad3-0d76e1024824.png)

![image](https://user-images.githubusercontent.com/37256050/214329638-4725904d-a73d-4965-94a7-4d0635bd7865.png)

maxxxxxdlp:
Looks like sp6 might have special treatment for the "file" input type. In sp7, it just renders the input as disabled - we will need to investigate it deeper to see what sp6 does.

Alternatively, this might be caused by https://github.com/specify/specify7/issues/415

grantfitzsimmons:
![image](https://user-images.githubusercontent.com/37256050/214396757-c32ccd17-42c8-49c2-b268-9c8c31c3a14c.png)
![image](https://user-images.githubusercontent.com/37256050/214396762-508ee0ac-c00b-4288-a98b-4390cbe1f01c.png)

**Reported By:** Gothenburg

grantfitzsimmons:
Ideally we would solve this with the PR #2961

Issue #415, opened 2017-06-30
Forms for *attachment tables are ignored.


benanhalt:
The form definitions for collectionobjectattachment etc tables are ignored because Specify 7 has to use a different plugin to display the attachment. There needs to be some way to customize these forms, though, so that fields can be added. Need to determine if it is possible to use the Specify 6 definition somehow, or introduce a way to define a special forms for 7 that are still customizable.

hexx870206helen:
Hey benanhalt,

Does that mean we can not customize the attachment forms, such as collection object attachment form etc  in Specify 7 currently? 

Kind Regards,
Helen

benanhalt:
@hexx870206helen, yes. I'm afraid that is currently the situation.

tlammer:
request for this feature added by Gothenburg

maxxxxxdlp:
Fixed in 7b1164e3

Issue #2955, opened 2023-02-07
Clicking and holding on the attachment size slider leads to an infinite loop


grantfitzsimmons:
https://user-images.githubusercontent.com/37256050/217329319-e3c3c581-4507-44e2-b51b-b58132935716.mp4

`Firefox`

grantfitzsimmons:
Could not recreate on chrome or firefox

maxxxxxdlp:
Replicated by @chanulee1:

> Dragging scale slider can result in this (not present in edge):
>
> https://user-images.githubusercontent.com/103299204/220204876-5d0a6640-59c4-4462-a3c2-516a20f9ef53.mp4

maxxxxxdlp:
Reproduced when two attachment viewers were open:

https://user-images.githubusercontent.com/40512816/220518862-6d05b00e-f24d-42d5-bed2-9e3eb09af1b4.mov

maxxxxxdlp:
Caused by local storage syncing between tabs. Need to research more what is the best way to fix this, while still preserving localstorage sync (i.e, maybe adding a 10ms throttle)

maxxxxxdlp:
FYI: this appears to be even easier to replicate if you have 3 tabs of the same page open.
Replicated in app resource editor too (fold/unfold the side bar)

Fixed on `xml-editor` branch

grantfitzsimmons:
https://github.com/user-attachments/assets/f41d3851-4b51-4e1d-91c4-0b1dec7b106d

This is still possible on `v7.12.0.6` and no changes have been made to address it since as far as I know. Encountered today on https://sp7demofish.specifycloud.org/specify/attachments/

PR #2961, opened 2023-02-09
Show attachment preview in a dialog with MetaData form

### LINKED ISSUE #2787: Attachments should show in a dialog preview
Clicking on an Attachment should open up a shadowbox with the attachment in a preview:
image

Like this:

image

CarolineDenis:
Fixed #2787

Need to improve Tailwind for iFrame. @maxxxxxdlp


maxxxxxdlp:
Fixes #415

maxxxxxdlp (approved):
I made some edits. Please review the comments

grantfitzsimmons:
Also fixes https://github.com/specify/specify7/issues/2873

grantfitzsimmons (changes requested):
https://coldfish-issue-2787.test.specifysystems.org/specify/attachments/

**Firefox:**
image

**Chrome:**
image

In Firefox, it stretches the image to fit the dimensions of the container.
In Chrome, it shows only the portion of the image that fits within the dimensions.

Can we make it so that it automatically fits to the dimensions of the frame and can be expanded when clicked on in the same frame?

grantfitzsimmons:
https://user-images.githubusercontent.com/37256050/219232502-3c1878f8-b0a8-4c35-a1af-837a2745230d.mp4

In Firefox here, it also seems that there is some odd scrollable space for no discernable reason

grantfitzsimmons (changes requested):
image

If you click on a table icon in the top left of an item, you see the following:

```
can't access property "get", t is undefined

Stack:

F@https://coldfish-issue-2787.test.specifysystems.org/static/js/7377.cf49b82643c9f5e5ea7d.bundle.js:1:5342
maybe@https://coldfish-issue-2787.test.specifysystems.org/static/js/main.dd443c3ed93a40be4542.bundle.js:2:579124
81631/P/onClick/<@https://coldfish-issue-2787.test.specifysystems.org/static/js/7377.cf49b82643c9f5e5ea7d.bundle.js:1:4955
```

maxxxxxdlp:
To test:
- Supported file formats have preview displayed (image, video, audio, pdf)
- Other formats have icon displayed (xlsx, zip)
- Dialog can be resizes/repositioned and it's position is remembered (close the dialog then open again and it should open in the same place). This change was made for all dialogs, and each dialog remembers it's own position, so please test 2-3 kind of dialogs just to make sure the feature works (https://github.com/specify/specify7/issues/3010)
- When viewing attachments in the attachments tool, left/right buttons are available for moving between attachments
- Attachments metadata form is visible and can be edited
- Fixes https://github.com/specify/specify7/issues/2873
- Fixes https://github.com/specify/specify7/issues/415
- Open in new tab/download/meta visibility toggle buttons work. Note, open in new tab acts equivalent to download for files that browsers can't preview (i.e, zip)
- The attachments plugin is displayed nicely when displayed as a subview button, subview or form table on the form

Screenshot 2023-02-17 at 15 50 46

chanulee1:
Opening preview of some attachments leads to this effect. Unsure of what causes this.


https://user-images.githubusercontent.com/103299204/220202621-5dcc1b9d-25c6-489b-988c-e499ca62e1f9.mp4


https://coldfish-issue-2787.test.specifysystems.org/specify/attachments/

chanulee1:
Dragging scale slider can result in this (not present in edge):


https://user-images.githubusercontent.com/103299204/220204876-5d0a6640-59c4-4462-a3c2-516a20f9ef53.mp4

maxxxxxdlp:
> Dragging scale slider can result in this (not present in edge):

Already covered by https://github.com/specify/specify7/issues/2955. It's present in production too(

Caused by cache sync between tabs. Need to research more how to fix it nicely, while preserving cache syncing

maxxxxxdlp:
> Opening preview of some attachments leads to this effect. Unsure of what causes this.

Fixed. This happened if there is no custom attachment view for that table (i.e, no AccessionAttachment view)

grantfitzsimmons:
**Download** and **Open In New Tab** do the same thing for me (macOS, Firefox)

https://user-images.githubusercontent.com/37256050/221668543-11680c84-d14b-49d0-8e97-ae4530e41f03.mov

**Expected Behavior:**

I would expect **Download** to download the jpg file rather than open it in a new tab. I'm sure this could be caused by my default image viewer on my system or some setting, but it should try to download it as a file

grantfitzsimmons:
I'd also like to see the ability to use arrow keys to navigate between items. That was my first instinct and I'm sure others will have the same request. 

https://github.com/specify/specify7/issues/3057

grantfitzsimmons:
image

This button should say **Hide Form** instead of **Show Metadata** when the form is visible (and conversely **Show Form** when it is opened) 

### "Form" should be used instead of metadata, since "Attachment Metadata" is its own form.

grantfitzsimmons:
image

The icon should not be ⚠️ in the top left. This should be the attachment icon (image)

grantfitzsimmons (changes requested):
See my comments above

grantfitzsimmons:
https://sp7demofish-issue-2787.test.specifysystems.org/specify/attachments/

If you want to try and see a lot of real-world attachments, the above link is very good

grantfitzsimmons:
image

There should also be a way to navigate to the record the attachment is connected to.

maxxxxxdlp:
> I'd also like to see the ability to use arrow keys to navigate between items. That was my first instinct and I'm sure others will have the same request.

A very important feature. Will be added when I work on keyboard accessibility. There is some infrastructure required to implement it in a nice and flexible way, so it's out of scope of this pull request, but it's important and I will try to get to it

maxxxxxdlp:
> This button should say Hide Form instead of Show Metadata when the form is visible (and conversely Show Form when it is hidden)

"Show Form" and "Hide Form" buttons labels are bad for accessibility.
I.e, a screen reader will read it like this:
`Show Form, unchecked, button`
`Hide Form, checked, button` (because button has an "is checked" attribute) - this label makes it seem like the "hide form" feature is checked, but the button actually means that form is visible and clicking it would hide the form.

Instead, can we just say `Show Form`
If checked, the button has a different style
Screen reader: `Show Form, checked, button`, and `Show Form, unchecked, button`

If you want, we could have just `Form`, or `Form Visible` instead.

maxxxxxdlp:
> The icon should not be ⚠️ in the top left. This should be the attachment icon (image) or something friendlier

Good catch. Fixed

maxxxxxdlp:
> There should also be a way to navigate to the record the attachment is connected to. Perhaps a button next to "Show Form" that says "Base Record" with the table icon? Just some way to get to it

Fixed

grantfitzsimmons (approved):
Working well in Chrome and Firefox! Seems to be some performance issues on the Attachments page in Safari, but that exists in production from what I can tell.

I would still like to be able to navigate using my keyboard, but I think this looks good!

Link to record not displaying when viewing from the forms as expected. Icons look good, icon and name are dynamic and based on the record. Can handle many large images one after another. Very satisfied with this initial implementation

PR #2974, opened 2023-02-10
Fix back-end tests not being run

maxxxxxdlp:


maxxxxxdlp:
The tests not being run bug has been fixed on a separate branch
Thus, only remaining thing is the schema version update

PR #2978, opened 2023-02-12
Improve WorkBench's "Disambiguate All" performance

maxxxxxdlp:
Clicking on "Apply All" in the disambiguation dialog in the WorkBench when there are ~1000+ ambiguities of that kind freezes the page for up to a minute.

This might fix it


grantfitzsimmons:
# Amazing!

This is a _major_ improvement.

You can see that the right side (`edge`) still wasn't finished by the end of that recording.

https://user-images.githubusercontent.com/37256050/218496171-f12ea9f3-8460-433c-ae1d-3e2d531174e6.mp4

grantfitzsimmons (approved):
👏

PR #2994, opened 2023-02-15
Add a warning when importing a query with hidden fields

### LINKED ISSUE #1318: Warn when importing a query that has hidden fields
Suggested by @tlammer:
> so it is more apparent when importing that there are differences and they may need to fix if they want to get results in their database

CarolineDenis:
Fixed #1318

### DIFF
--- a/specifyweb/frontend/js_src/lib/localization/query.ts
+++ b/specifyweb/frontend/js_src/lib/localization/query.ts
@@ -638,4 +638,7 @@ export const queryText = createDictionary({
       {maxLength:number|formatted}
     `,
   },
+  hiddenFieldsMess: {
+    'en-us':'The following fields are hidden in the query you imported: '
+  },
 } as const);
--- a/specifyweb/frontend/js_src/lib/localization/query.ts
+++ b/specifyweb/frontend/js_src/lib/localization/query.ts
@@ -639,6 +639,6 @@ export const queryText = createDictionary({
     `,
   },
   hiddenFieldsMess: {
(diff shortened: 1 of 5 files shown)


grantfitzsimmons (changes requested):
image

[NCBI Genbank list 2020.json.zip](https://github.com/specify/specify7/files/10751501/NCBI.Genbank.list.2020.json.zip)

It looks like I can only see the table name for _one_ of these fields.

https://herbrbge11723-issue-1318.test.specifysystems.org/specify/query/249/

A bit confusing– both should have the table icon, name, and then the field.

I would imagine this to have the table name for **each** field and have it be bold while the field name is not bold. 

image

CarolineDenis:
Having the name of the table in bold and not the field is not possible at the moment. 
This dialog will not be often shown, the work to implement that would be too big for what it is worth.

grantfitzsimmons:
Sounds good!

grantfitzsimmons (approved):

grantfitzsimmons:
@maxxxxxdlp Merge if everything checks out :smile:

specifyweb/frontend/js_src/lib/localization/query.ts
```diff
@@ -638,4 +638,7 @@ export const queryText = createDictionary({
       {maxLength:number|formatted}
     `,
   },
+  hiddenFieldsMess: {
```
maxxxxxdlp, line 641: @CarolineDenis this is not a good name for a localization string. Since this pull request is already merged, can you create a new pull request with a fix for this?

PR #3002, opened 2023-02-16
Use a tooltips library

maxxxxxdlp:
To test:
- Different places that use tooltips
   - Make sure tooltip is appropriate and useful
   - Make sure tooltip does not overflow the screen
- Tooltips appear after an appropriate amount of time (i.e, after a duration that makes sense)
- Tooltip text can be selected
- Tooltips can be activated with keyboard
- Tooltips can be used on a touchscreen device (mobile phone)
- Disabled elements fallback to browser default tooltips
- Tooltips look nice in dark mode and light mode
- Tooltips have sufficient contrast from the background elements to be easily visible and readable in all places. if this is not good enough, we can consider changing background color or adding some shadow in light mode
- Custom tooltips can be disabled in user preferences (Main -> UI -> Moder tooltips)

Some bugs found so far:

Overlaps with content:
Screenshot 2023-02-16 at 00 11 34

Causes page overflow:
Screenshot 2023-02-16 at 00 11 44


grantfitzsimmons:
https://user-images.githubusercontent.com/37256050/219370163-3db39c3f-5867-4dc0-bbb6-a0eb046ba3e3.mov

I'm not sure that having the tooltip jump from one item to another between hovering. I don't think it should be jumping around if possible

grantfitzsimmons:
image

Tooltips are still visible and at the forefront when a dialog is opened. Click on Reports and see the behavior

grantfitzsimmons (changes requested):
There are a number of usability concerns I have about the current implementation– many of which may be resolved by making the hover requirement a longer period of time before displaying the tooltips.

The navigation menu items should have the tooltip display at the current speed, but not every button or field in my opinion.

---

Another issue– if you have a tooltip open and the dialog changes or disappears the tooltip remains: See "Edit" in this instance

https://user-images.githubusercontent.com/37256050/219372849-4ad2e40a-df29-4b56-a8f5-7cd8429c5abb.mov

grantfitzsimmons:
> Causes page overflow: Screenshot 2023-02-16 at 00 11 44

Can we not have tooltips be many more lines long and have it wrap?

CarolineDenis (approved):

maxxxxxdlp:
> I'm not sure that having the tooltip jump from one item to another between hovering. I don't think it should be jumping around if possible.

It's a bug. Caused by animations being enabled (we have a global style saying all transitions should be animated - most of the time it's good - exceptions are the dialog drag and the tooltip move).

I had animations off so didn't detect this. Should be fixed now.

@grantfitzsimmons though, there is another question. I made tooltips have an 800ms delay before being shown for the first time, but then if you move your mouse to another element, the tooltip for the new element is displayed right away (assuming you moved you mouse to new element within the 1000ms threshold). Should I change that behavior to require 800ms for each hover? or maybe some smaller number?

Or would you like me to temporarily add user prefs for customizing these timeout durations so that we can fine-tune them and set good defaults?

maxxxxxdlp:
> Tooltips are still visible and at the forefront when a dialog is opened. Click on Reports and see the behavior

Fixed. As a side effect of the fix, fancy tooltips are no longer available on the login page, choose collection page and password change page. Hopefully not a big deal as we didn't have tooltips there before

grantfitzsimmons:
image

This is with a VERY long description

image


Maybe we need to add a shadow or outline to these tooltips

maxxxxxdlp:
Fixes #3009

maxxxxxdlp:
> Maybe we need to add a shadow or outline to these tooltips

Added a shadow

maxxxxxdlp:
If this is good now, we can merge it and release

grantfitzsimmons (approved):
Ok– you have won me over. There is more work to do here, so don't think this is the last time we will see tooltip tomfoolery

grantfitzsimmons:
@maxxxxxdlp Just need to resolve the conflict & we can release

specifysoftware:

https://discourse.specifysoftware.org/t/specify-7-8-6-release-announcement/1064/1

PR #3007, opened 2023-02-17
Set focus to parent node after deletion

### LINKED ISSUE #2781: Jump to node's parent after deletion
> **HeatherC:**
> 
> When working in the S7 tree, I really appreciate when it ‘jumps’ back to where I was. I often have many edits to do in the same part of the tree, or am working my way through. I have found on edits, merges, and synonomy, the tree display ‘returns’ to where I was. Love it.
>
> For some reason, on ‘delete’ it doesn’t go back, and I am left at the top of the tree, and have to enter another search, or guess where to scroll to. I understand you can’t jump back to the node I deleted, but I would find it very useful (and time-saving) if it would jump back to the parent node (of the node just deleted).

**Speciforum:** https://discourse.specifysoftware.org/t/tree-behaviours/978

I think a good solution would be to jump to the parent of the deleted node rather than losing the place in the tree.

CarolineDenis:
Fixed #2781


maxxxxxdlp (approved):

chanulee1 (approved):
Works!

Issue #123, opened 2015-06-26
"save and add another" Gift doesn't save


garrettgirard:
If you create a Gift and use the "Save and Add Another" feature, it grays out the "save" and "save and add another" features until you make any changes to the gift, however it doesn't actually save the gift.

So if you select the "save and add another" option and then make changes and save a second gift, only the second gift will be saved; any gifts saved using "save and add another" will not be saved.

benanhalt:
@timo11 Does "save and add another" make sense for interaction forms? Does Sp6 support it?

timo11:
I don't think it's necessary or sensible.
Especially not for loans and gifts.

Issue #1661, opened 2022-05-18
Warn when opening a query that has fields without read access


maxxxxxdlp:
When opening a query that contains relationships to related tables to which the current user does not have read permissions, some sort of warning dialog should be shown.

The user should see a list of relationships they don't have read permission to as well as a one-click option to remove those fields from the query

See also #1318

grantfitzsimmons:
This looks like more work than it's worth, what do you think?

maxxxxxdlp:
It's a good practice assignment before we get to more interesting tasks

maxxxxxdlp:
Also, this would become more useful once's shared queries mechanism is in place

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/specify-7-8-6-release-announcement/1064/1

PR #3013, opened 2023-02-17
 Warn if the query contains fields without read permission

CarolineDenis:
Issue 1661


maxxxxxdlp:
Don't forget the `#` symbol when mentioning an issue and use the `Fixes` keyboard. The reason is that github is setup to recognize `Fixes #123` and `Fixed #123` and automatically links the issue and pull request together. Then, if the pull request is closed, the issue is closed automatically

specifyweb/frontend/js_src/lib/components/QueryBuilder/CheckReadAccess.tsx
```diff
@@ -0,0 +1,66 @@
...
+  query,
+}: {
+  readonly query: SerializedResource<SpQuery>;
+}): JSX.Element | null {
+  const [noAccessTables, setNoAccessTables] = React.useState<RA<keyof Tables>>(
+    []
+  );
+
+  function accessTables(query: SerializedResource<SpQuery>): RA<keyof Tables> {
+    const info = query.fields.flatMap((field) => {
```
maxxxxxdlp, line 23: You said that you preferred to split big expressions into separate variables to make it easier. To follow that, `info` variable name is not description at all. Can you rename it to something like `tableNames`?

specifyweb/frontend/js_src/lib/components/QueryBuilder/CheckReadAccess.tsx
```diff
@@ -0,0 +1,66 @@
...
+      const fieldSpec = QueryFieldSpec.fromStringId(
+        field.stringId,
+        field.isRelFld ?? false
+      );
+      const relatedTable = fieldSpec.joinPath.map(({ model }) => model.name);
```
maxxxxxdlp, line 28: The `relatedTable` variable name is incorrect since it is assigned to an array. Also, you really don't need that variable and can just return right away

maxxxxxdlp: Can you replace `fieldSpec.joinPath.map(({ model }) => model.name);` with `fieldSpec.joinPath.map((field)=>field.isRelationship ? field.relatedModel.name : field.model.name)`?

This way, fields like `Accession -> (formatted)` would be checked correctly

specifyweb/frontend/js_src/lib/components/QueryBuilder/CheckReadAccess.tsx
```diff
@@ -0,0 +1,66 @@
...
+      return relatedTable;
+    });
+
+    const withoutDuplicates = new Set(info);
+
+    const noAccessTables = Array.from(withoutDuplicates).filter(
```
maxxxxxdlp, line 34: (optional) instead of `Array.from(new Set(someVar))` can use our util and just type `f.unique(someVar)`

specifyweb/frontend/js_src/lib/components/QueryBuilder/CheckReadAccess.tsx
```diff
@@ -0,0 +1,66 @@
...
+      (name) => !hasTablePermission(name, 'read')
+    );
+
+    setNoAccessTables(noAccessTables);
+    return info;
```
maxxxxxdlp, line 39: Return is unused

specifyweb/frontend/js_src/lib/components/QueryBuilder/CheckReadAccess.tsx
```diff
@@ -0,0 +1,66 @@
...
+export function CheckReadAccess({
+  query,
+}: {
+  readonly query: SerializedResource<SpQuery>;
+}): JSX.Element | null {
+  const [noAccessTables, setNoAccessTables] = React.useState<RA<keyof Tables>>(
+    []
+  );
+
+  function accessTables(query: SerializedResource<SpQuery>): RA<keyof Tables> {
```
maxxxxxdlp, line 22: Bad function name. It should either return the tables, in which case it should be called `getNoAccessTables`, or modify the tables, in which case it could be called `findNoAccessTables`

specifyweb/frontend/js_src/lib/components/QueryBuilder/CheckReadAccess.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const withoutDuplicates = new Set(info);
+
+    const noAccessTables = Array.from(withoutDuplicates).filter(
+      (name) => !hasTablePermission(name, 'read')
+    );
+
+    setNoAccessTables(noAccessTables);
+    return info;
+  }
+  React.useEffect(() => {
```
maxxxxxdlp, line 41: Try to avoid useEffect whenever possible. In this case, useMemo would work better. Or even simpler
```js
const [noAccessTables,setNoAccessTables] = React.useState(getNoAccessTables);
```

Where `getNoAccessTables` is a function that would return `RA<keyof Tables>`

maxxxxxdlp: This works because you can provide the initial value to React.useState as the value itself, **or as a function that returns default value**

specifyweb/frontend/js_src/lib/components/QueryBuilder/CheckReadAccess.tsx
```diff
@@ -0,0 +1,66 @@
...
+    accessTables(query);
+  }, []);
+
+  return noAccessTables.length > 0 ? (
+    <Dialog
+      buttons={commonText.close()}
+      header={queryText.noReadPermission()}
+      onClose={(): void => setNoAccessTables([])}
+    >
+      <>
```
maxxxxxdlp, line 51: `<>` is not needed here

specifyweb/frontend/js_src/lib/components/Toolbar/Query.tsx
```diff
@@ -170,18 +170,20 @@ function QueryList({
...
-                  getQuerySelectUrl?.(query) ?? `/specify/query/${query.id}/`
-                }
-              >
-                <TableIcon
-                  label
-                  name={getModelById(query.contextTableId).name}
-                />
-                {query.name}
-              </Link.Default>
+              <>
```
maxxxxxdlp, line 173: `<>` is not needed here

specifyweb/frontend/js_src/lib/hooks/useStateForContext 2.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
```
maxxxxxdlp: Not sure how these files keep getting created on your computer

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Toolbar/Query.tsx
maxxxxxdlp: Looks like redundant `<>` are still here. The other changes look good. As soon as this one is resolved we can merge

maxxxxxdlp (approved):

chanulee1 (commented):
Warning appears to be working. But, could not find the "one-click option to remove those fields" without read permission mentioned in #1661. Will this not be included in this PR?
image

CarolineDenis:
> Warning appears to be working. But, could not find the "one-click option to remove those fields" without read permission mentioned in #1661. Will this not be included in this PR?
> image

@chanulee1 indeed, deleting the fields is not a part of this PR

chanulee1:
In that case, this PR looks good.

chanulee1 (approved):

PR #3017, opened 2023-02-20
fix failing unit tests in prod

acwhite211:
This commit is a series of fixes from various failing unit tests in prod branch.  As of this first commit, the only remaining unit test that is still working on being resolved is specifyweb.context.tests.OpenApiTests.test_operations_spec.


acwhite211:
I believe the openapi schema change of specify/tree_views.py from [this pull request](https://github.com/specify/specify7/pull/2818/files#diff-1967ecfeac4d37f410b5ea281811229f5383fa0b725a8b7447e2d56731e94cde) is causing the issues for the last failing unit test.  Will investigate further

acwhite211:
Ok, all unit tests are working now, ready for reviewing and merging.

specifyweb/context/views.py
```diff
@@ -823,6 +823,14 @@ def languages(request):
         else:
             languages = languages.split(',')
 
+        temp = JsonResponse({
+            code:{
+                **get_language_info(code),
+                'is_current': code==request.LANGUAGE_CODE
+            } for code in languages
+        })
+        print(temp)
```
melton-jason, line 832: Is this `temp` variable used anywhere else? Can it be removed?

melton-jason (changes requested):
Overall looks good! One minor change.

specifyweb/context/views.py
acwhite211: good catch, removed that temp code

melton-jason (approved):

specifyweb/specify/tree_views.py
```diff
@@ -56,50 +56,50 @@ def wrapper(*args, **kwargs):
                         "schema": {
                             "type": "array",
                             "items": {
-                                "type": "array",
-                                "items": [
-                                    {
+                                "type": "object",
```
maxxxxxdlp, line 59: This doesn't look right. The endpoint returns an array:
image

specifyweb/context/views.py
```diff
@@ -722,19 +722,19 @@ def api_endpoints_all(request):
...
-            "languages": {
-                "name": "languages",
-                "in": "query",
-                "description": "Comma separate list of languages",
-                "example": "en-us,uk-ua,ru-ru",
-                "required": False,
-                "schema": {
-                    "type": "string",
-                },
+    "parameters": [
```
maxxxxxdlp, line 725: This change should be undone. The "languages" parameter is only for the "get" method, not for the "post" method

specifyweb/workbench/upload/tests/testuploading.py
```diff
@@ -173,15 +173,15 @@ def test_enforced_county(self) -> None:
             results[0].record_result,
             FailedBusinessRule(
                 message='missingRequiredTreeParent',
-                payload={'names':['County']},
+                payload={'names':"['County']"},
```
maxxxxxdlp, line 176: This doesn't look right. Front-end expects this to be a list, not a json string

specifyweb/workbench/upload/tests/testuploading.py
```diff
@@ -173,15 +173,15 @@ def test_enforced_county(self) -> None:
...
                 info=ReportInfo(tableName='Geography', columns=['City'], treeInfo=None)
             )
         )
         self.assertEqual(
             results[1].record_result,
             FailedBusinessRule(
                 message='missingRequiredTreeParent',
-                payload={'names':['County']},
+                payload={'names':"['County']"},
```
maxxxxxdlp, line 184: This doesn't look right. Front-end expects this to be a list, not a json string

maxxxxxdlp (changes requested):

specifyweb/specify/tree_views.py
acwhite211: ya, i thought so too, but was the only way I could get the schema validation test to work.  Feel free to play around with it

melton-jason: I originally defined the openapi schema for that endpoint. I could not find any other endpoint on which to model it from. I wanted to provide additional information for each of the items of the nested array. Do you know if there is a way to do this without losing any information?
I am willing to mess around with the schema structure for this endpoint and find a solution.

specifyweb/workbench/upload/tests/testuploading.py
acwhite211: ya, but the test fails with an assertion error without this change

specifyweb/specify/tree_views.py
maxxxxxdlp: I see what you mean. Looks like it's possible, but though a kind of a work-arround: https://stackoverflow.com/a/57465199/8584605 (the comment uses yaml, but we can write the same in json)

Converting the endpoint to return object is not recommended as that would increase the payload size significantly

specifyweb/context/views.py
acwhite211: this is the validation error that is thrown without the change to this schema:``

acwhite211: fixed

specifyweb/workbench/upload/parsing.py
```diff
@@ -26,7 +26,7 @@ class PicklistAddition(NamedTuple):
 
 class ParseFailure(NamedTuple):
     message: str
-    payload: Dict[str, Any]
+    payload: Dict[str, Union[str, int, List[str], List[int]]]
```
maxxxxxdlp, line 29: Are you sure mypy doesn't support `Any` type? https://mypy.readthedocs.io/en/stable/kinds_of_types.html#the-any-type

What error did you get?

Sorry, I can't get the back-end tests to run on my machine to see the error for myself

acwhite211: it's not mypy that is the problem, it's hypothesis that is not able to mock it for testing

acwhite211: to run a unit tests locally run ```docker exec -it specify7-specify7-1 ve/bin/python3 manage.py test specifyweb.workbench.upload.tests.test_upload_results_json.UploadResultsTests.testUploadResult```

acwhite211:
That change to the tree_view schema is causing a validation failure.  It's saying that the items array contains arrays that describe the properties of each node, but they should be objects instead.
```jsonschema.exceptions.ValidationError: [{'type': 'integer', 'description': 'The id of the child node'}, {'type': 'string', 'description': 'The name of the child node'}, {'type': 'string', 'description': 'The fullName of the child node'}, {'type': 'integer', 'description': 'The nodenumber of the child node'}, {'type': 'integer', 'description': 'The highestChildNodeNumber of the child node'}, {'type': 'integer', 'description': 'The rankId of the child node'}, {'type': ['integer', 'null'], 'description': 'The acceptedId of the child node. Returns null if the node has no acceptedId'}, {'type': ['string', 'null'], 'description': "The fullName of the child node's accepted node. Returns null if the node has no acceptedId"}, {'type': ['string', 'null'], 'description': 'The author of the child node. Returns null if <tree> is not taxon or the rankId of the node is less than <includeAuthor> paramter'}, {'type': 'integer', 'description': 'The number of children the child node has'}] is not of type 'object'```

**Comment by melton-jason** [comment]
Yes, I am looking into a workaround.
I believe the issue is caused by the definition for the `items` property to always be an object in [openapi_schema.py](https://github.com/specify/specify7/blob/production/specifyweb/context/openapi_schema.py ) (this also explains why `prefixItems` is not recognized as a valid key, even though it is in json schema draft 2019).

Validation works correctly with my changes in an external source using [jsonschemavalidator.net](https://www.jsonschemavalidator.net/ ). 
I will try and maintain the nested array structure and individual item description, but it will probably be a hacky solution regardless of the outcome.

**Comment by maxxxxxdlp** [comment]
We might be using OpenAPI 3.1 syntax, where as JSON schema is set to recognize OpenAPI 3.0 syntax or wise versa

**Comment by melton-jason** [comment]
> We might be using OpenAPI 3.1 syntax, where as JSON schema is set to recognize OpenAPI 3.0 syntax or wise versa

This may be true, but even a workaround that _should_ work in older versions is invalid. 

> In Draft 4 - 2019-09, tuple validation was handled by an alternate form of the items keyword. When items was an array of schemas instead of a single schema, it behaved the way prefixItems behaves.

- [json-schema.org#tuple-validation](https://json-schema.org/understanding-json-schema/reference/array.html#tuple-validation )

Even having an array of schemas will not work because in our case `items` is always expected to be an object (although an array of schemas is _exactly_ what we are looking for). We should be able to use `enum` to restrict the type of schema to integer, number, or null, but we would lose a lot of the information. 
If worse comes to worse, we can always include some information in an `example` key for the endpoint

**Comment by melton-jason** [comment]
I've fixed the endpoint (and in the process, have learned a _lot_ about jsonschema). 

In the future, it may be wise to expand our schema to allow definitions of more than one type. For instance, there is no way currently to say the type of something could be a number or a string, or a number or null, etc. 

The jsonschema way of doing this would be to define it as follows:
`"type": ["type1", "type2", ...]`

**Comment by maxxxxxdlp** [comment]
Good news.
Swagger UI is finally supporting OpenAPI 3.1 - https://swagger.io/blog/swagger-support-for-openapi-3-0-and-openapi-3-1/

That means we can migrate to using it - the biggest benefit for us is that it's more compatible with JSON schema, which we are already using a lot on the back edn

File: specifyweb/workbench/upload/tests/testuploading.py (hunk 4 above)
Comment (maxxxxxdlp) [reply specifyweb/workbench/upload/tests/testuploading.py:176]: This should be fixed somehow before we can merge the PR, or else it will cause front-end to fail

**Review by maxxxxxdlp — CHANGES_REQUESTED** [review]

File: specifyweb/workbench/upload/tests/testuploading.py (hunk 5 above)
Comment (acwhite211) [reply specifyweb/workbench/upload/tests/testuploading.py]: the payload value is an array now.  This change fixed that: [treerecord.py](https://github.com/specify/specify7/blob/eb377cae961fe8eb23b292173fcf401d782e5347/specifyweb/workbench/upload/treerecord.py#L297)

File: specifyweb/context/openapi_schema.py [hunk 7]
```diff
@@ -480,6 +480,19 @@
             }
           ]
         },
+        "prefixItems" : {
```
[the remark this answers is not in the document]
maxxxxxdlp: Tested. Looks like swagger can't display this well (I get an infinite spinner)

Screenshot 2023-02-24 at 15 55 09

Updating to swagger UI 4 didn't fix this.

maxxxxxdlp (approved):
Swagger has issues with new syntax, but let's not block this PR for that

specifyweb/context/openapi_schema.py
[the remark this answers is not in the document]
melton-jason: Interesting. I can remove prefixItems and use a similar schema for the tree_view endpoint if need be. 
I can also try to remove the reference to `#/definitions/Reference` and see if that alleviates the issue. 

Do we have code somewhere that tells swagger how to display the schema? All of the other properties are technically self-defined, as is this one. 
If this is the case, then maybe the infinite spinner is caused by the recursive nature of referencing  `"#/definitions/Schema"` (which `prefixItems` is defined to be a property of).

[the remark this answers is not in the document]
maxxxxxdlp: recursion could be it!

you can try updating the json schema we have with this - https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.json

> Do we have code somewhere that tells swagger how to display the schema?

Both swagger UI and we are following OpenAPI 3.0 (and now OpenAPI 3.1) standard - https://swagger.io/specification/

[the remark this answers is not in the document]
acwhite211: I'm going to finish this pull request, let's add this upgrade to a new issue.

Issue #2580, opened 2022-12-10
[testability] Deleting a saved query from inside of it does not navigate away immediately


chanulee1:
https://user-images.githubusercontent.com/103299204/206831560-6d655ede-14b6-48f1-86bf-8858272e7ff9.mp4

grantfitzsimmons:
@mc0822 recreated this issue again today.

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/specify-7-8-9-release-announcement/1131/1

PR #3022, opened 2023-02-20
Update the query list after a deletion

### LINKED ISSUE #3021: Deleting query does not update query list immediately
https://user-images.githubusercontent.com/103299204/220184179-a0767e71-07b2-4772-af1e-057b73387f7d.mp4

https://coldfish-edge.test.specifysystems.org/specify/overlay/queries/

branch: `edge`

CarolineDenis:
Fixed #3021

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
@@ -32,7 +32,9 @@ import type { Tables } from './types';
  * fetching the new one.
  */
 
-export const resourceEvents = eventListener<{ readonly deleted: SpecifyResource<AnySchema> }>();
+export const resourceEvents = eventListener<{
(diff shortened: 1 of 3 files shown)


maxxxxxdlp (approved):

melton-jason:
Is #2580 within the scope of this pull request? It is not specific to the query list, but is related to performing some action/update upon query deletion (specifically, not navigating away from a query when it is deleted, potentially cause a 404 page to be displayed).

chanulee1 (approved):
If the answer to Jason's question is no, then this looks good.

maxxxxxdlp:
@CarolineDenis you can use a resourceEvents just like you did in the query list to detect when the query that is currently displayed has been deleted and so redirect the user to home page/query list dialog.
Place that code into a hook and call it in QueryBuilder/Wrapped.tsx

PR #3029, opened 2023-02-21
Stop importing a query for  creator when imported by different user

### LINKED ISSUE #3020: Importing a query from different user in the same database imports in that user too
When importing a query from a different user in the same database, the query is imported to the queries of the user that the query is from as well.

Steps to recreate:
1. Export a query from a user
2. Import the query to a different user in the same database
3. Log back in as the original user and see that the query is imported there as well

https://coldfish-edge.test.specifysystems.org/specify/overlay/queries/

branch: `edge`

CarolineDenis:
Fixed #3020


maxxxxxdlp (approved):

chanulee1 (approved):
Appears to be fixed.

PR #3032, opened 2023-02-21
Support external URL in form definition for weblinks and separators

### LINKED ISSUE #2095: Images in the form definiton should support external URLs
Images for elements like separators, buttons, and other places in the form currently use the internal images within the database.

It would be nice to specify an external URL so you can use a custom icon for things like weblinks or separator headers

CarolineDenis:
Fixed #2095


maxxxxxdlp (approved):

maxxxxxdlp:
Also test it for subviews that are displayed as buttons - and that would be it

chanulee1:
Also works for subviews displayed as buttons.

chanulee1 (approved):

PR #3034, opened 2023-02-21
Forbid negative record set indexes

### LINKED ISSUE #3033: Adding a record to an empty record set is buggy
1. Open empty record set
2. It opens a new resource form
3. Save it
4. Should replace the `new` in the URL with the id of the resource, while still having `?recordsetid=someNumber` in the URL
5. In the records slider, current index should be shown as `1`, not `0`

maxxxxxdlp:
Fixes #3033


chanulee1 (approved):
Looks good.

PR #3036, opened 2023-02-21
Limit max year to 9999

### LINKED ISSUE #2353: Y10K problem: Date in far future causes back-end server error
Date that is beyond year 9999 (i.e `11-11-11111`) causes back-end to return 500 error:

Screen Shot 2022-10-25 at 12 51 45

Back-end should probably provide a better error message
Front-end should probably enforce dates to be before year `9999`

Similar to https://github.com/specify/specify7/issues/1072

### LINKED ISSUE #3035: Year in date field accepts 6 digits, throws error
image

https://sp7demofish.specifycloud.org/specify/view/collectionobject/51591/?recordsetid=14

[Specify 7 Crash Report - 2023-02-21T18_56_05.774Z (1).txt](https://github.com/specify/specify7/files/10796686/Specify.7.Crash.Report.-.2023-02-21T18_56_05.774Z.1.txt)

This causes a particular issue for users who have year first in their date format:

image

> Same day, different question, I have switched a little bit to filling in some dates, one of the things I really liked about S6, was that when entering dates, it would automatically moved you through the year/month/date field structure when entering the date components, so that you didn’t have to manually tab or mouse between them.
>
>However, when using the ‘browse in forms’ to edit collection objects, the system waits until I have entered 6 digits for year (instead of 4) before ‘moving’ over to the month section.

**Reported By:** Heather Cole at AAFC/AAC

maxxxxxdlp:
Fixes #3035
Fixes #2353
When testing, test partial dates and query builder date picker too


grantfitzsimmons (changes requested):
I can still enter 6 digits on Firefox, not on Chrome

maxxxxxdlp:
Seems to be a difference between Firefox and Chrome.

In Firefox, I get this validation error:
Screenshot 2023-02-23 at 11 27 01

The validation error would prevent submitting the form, thus would prevent the exception and the loss of data.

Whereas in Chrome I am prevented from entering the invalid year in the first place.

You can see the difference in this example too - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date (behaviors differently between firefox and chrome)

In addition, this behavior appears to be not specific to the date input - the number picker in firefox also allows entering a number higher than the limit.

grantfitzsimmons (approved):
I can't see a standard way to fix this in Firefox in particular

specifysoftware:

https://discourse.specifysoftware.org/t/specify-7-8-7-release-announcement/1079/1

PR #3037, opened 2023-02-22
Make WB new cell highlighting more conservative

### LINKED ISSUE #2966: Stop highlighting entire row if no new cells are returned
WorkBench

maxxxxxdlp:
If the back-end returned no columns, don't highlight the entire row

I.e, if you have `Collector -> Agent -> Last Name` column, and new Collector was created, but existing Agent was used, previously it would highlit that cell as new, but not it won't

Fixes #2966


grantfitzsimmons (approved):
Works as expected!

PR #3038, opened 2023-02-22
Disable default form table ordering

### LINKED ISSUE #2981: Collectors are sorted by `collectorID`, not `orderNumber` in grid view
image

The order is the default ordering by `collectorID`. It should be by `orderNumber`, as this is the intended use of that field.

```SQL
select c.CollectorID, c.CollectingEventID, c.OrderNumber, a.LastName, a.FirstName 
from collector c 
join agent a ON c.AgentID = a.AgentID
where c.CollectingEventID = 260;
```

image

This conflicts with Specify 6's behavior

image

https://coldfish-edge.test.specifysystems.org/specify/view/collectingevent/260/

maxxxxxdlp:
Turns our resources are already ordered on the back-end, so the default ordering on the front-end was only causing problems

Fixes #2981

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
@@ -48,11 +48,6 @@ const cellToLabel = (
 const cellClassName =
   'sticky top-0 bg-[color:var(--form-foreground)] z-10 h-full -mx-1 pl-1 pt-1';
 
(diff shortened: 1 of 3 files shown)


chanulee1 (approved):
Collectors are ordered by `orderNumber` in grid view and subform view. Also tested for Reference Work table.

specifysoftware:

https://discourse.specifysoftware.org/t/collector-order-in-groups-and-multiple-collectors/1012/5

Issue #2696, opened 2022-12-20
Cannot create new agent from the assign agent QCBX in the accounts panel


grantfitzsimmons:
**Reported By:** @grantfitzsimmons

grantfitzsimmons:
**Reported By:** Natural History Museum of Denmark

> 
> Our process was the following:
> 
> 1. Sp7 – Security and Accounts
> 
> 2. Choosing the collection
> 
> 3. Under ‘User Accounts Assigned to this Collection’ selecting a user
> 
> 4. Clicking on Enable Collection Access
> 
> 5. First problem: it doesn’t find an agent with this name
> OK, we are creating a new one for him
> 
> 6. Trying to save it
> 
> 7. Set User Agents
> 
> 8. Second problem: a pop-up window is appearing that says ‘Set user agents’
> Please set the following agents before proceeding:
> It show the he has agents at the collections, where he needs to have one, but doesn’t have, where he has no access. I cannot save it this way.
> I cannot save it if I fill out every fields (adding agents for him at each and every collections).
> 
> 9. Dead end.

grantfitzsimmons:
**Reported By:** Willem

> I tried to add a new user. It seems that there was an issue when I tried to create a new agent record from within the ‘create new user’ process. I was forced to create a new agent in the agent table first and then look it up in the create user process.

I can confirm this experience as reported above.

maxxxxxdlp:
I got good news and bad news.

Turns out, this is a bug with business rules, not with the security panel. Scoping rules were not applied to resources created outside of the Data Entry (or Interactions or Record Set) screen (on the front-end, the workbench creates resources on the back-end so it's ok) (all other business rules were still applied). Thus, agents created in the security panel were not scoped to division.

This has been a bug since the beginning of specify 7, but it didn't manifest itself for two reasons:
 - we didn't have any place where you created resources outside of Data Entry.
 - we didn't not load code dynamically. even if you were in query builder, the form code was loaded anyway.

Since scoping file was loaded dynamically, it was only loaded when you open Data Entry. And if you didn't open forms, it won't get loaded. Thus, a temporary workaround for this bug is to open any form, and then if you go to user preferences, the agent will be created correctly.

But I also pushed a temporary fix.

maxxxxxdlp:
#2696 and #2405 are manifestations of the same issue

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

[...]

Issue #2405, opened 2022-11-04
Cannot set batch of agents for collection access


grantfitzsimmons:
https://user-images.githubusercontent.com/37256050/200043897-c86ff644-c8fb-4594-b0b0-d86609b8127b.mp4

I tried to grant collection access to all collections for a user (`ksearcy`) in the `herb_umass` database. This resulted in a prompt asking for me to assign agents for all collections. Once I did, the dialog would not disappear. I then tried to save the record and was unable. Refreshing showed that the changes were not saved.

https://herbumass-testability.test.specifysystems.org/specify/security/user/2/

`testability`, macOS, chrome

grantfitzsimmons:
Performing the same set of activities on `edge` I recieved this warning:

<img width="708" alt="image" src="https://user-images.githubusercontent.com/37256050/200044335-6a909db5-1bdf-4045-afaf-84a17b705006.png">

Keep in mind that I created 3 agents using the QCBX during this process.

maxxxxxdlp:
The video you attached does not seem to load.

I wasn't able to replicate the issue in Herb_umass.
My steps:
1. Create a user with collection access in one collection
2. Create agent for it
3. Save - no issues
4. Make user institutional admin
5. Save - dialog about missing agents pops up
6. Set all agents to agents that are already in use
7. Press Save - dialog updates to show that agents are already in use
8. Select agents that are not in use
9. Save - saved without issues

maxxxxxdlp:
For the error on your screenshot, I literally have a comment in code "Not formatting this error nicely, as it shouldn't ever happen".
LOL

The error means you assigned more than one agent to the same user in the same division - not sure how that happened

grantfitzsimmons:
Here is the full video walking through how to recreate this bug in `edge`:

The issue is that if you can create agents with the QCBX, it should be supported. If existing agents are required, it should be enforced.


https://user-images.githubusercontent.com/37256050/202812303-78d86c12-f32b-4cc7-ba20-fdc2a6d6c391.mp4

maxxxxxdlp:
I was able to reproduce the bug from your last video in `edge`, but not in `testability`. Since it doesn't happen in next release, I won't investigate it further

I was able to reproduce the behavior you mentioned in the initial comment on `testability`.

It happens because sp7 does not set `Agent.division` when creating a new agent, and so when you choose an agent in the `Set User Agents` dialog, back-end does not detect that you set an agent in that division, because agent is not connected to the division.

[...]

PR #3039, opened 2023-02-22
Fix scoping rules not always being loaded

maxxxxxdlp:
The scoping rule file wasn't imported until some code needed
"fetchCollectionsForResource". Thus, scoping rules were not applied
unless a form was opened.

This means records created outside of a form weren't getting scoped.

Fixed.

Fixes https://github.com/specify/specify7/issues/2696


maxxxxxdlp:
@specify/ux-testing 
Can this be reviewed? This is a quite high priority

chanulee1 (approved):
Appears to be fixed. Unable to replicate #2696 and #2405.

PR #3044, opened 2023-02-23
Allow to populate number of prep in loans with bulk value

### LINKED ISSUE #2549: Allow setting default quantity of preps when selecting all preps
> For many of our loans we will have multiple loan preparations that we have to add (>30) but when you add the list of preparations (from a record set) it defaults to choosing the highest quantity for all preps (image 1). In our case we'd like the default quantity to be 1, can we change that anywhere?

image

**Requested By:** Dan Baker at CSIRO on [Asana](https://app.asana.com/0/1201972579963460/1203424280039616/f)

CarolineDenis:
Fixed #2549


specifyweb/frontend/js_src/lib/components/Interactions/PrepDialog.tsx
```diff
@@ -205,8 +209,29 @@ export function PrepDialog({
...
+              <td />
+              <td>{commonText.bulkValue()}</td>
+              <td>
+                <Input.Number
+                  aria-label={interactionsText.selectedAmount()}
+                  max={maxPrep}
+                  min={1}
+                  title={interactionsText.selectedAmount()}
+                  value={bulkValue}
+                  onValueChange={(newSelected): void =>
```
maxxxxxdlp, line 224: Can just replace this with `onValueChange={setBulkValue}`

specifyweb/frontend/js_src/lib/localization/query.ts
```diff
@@ -650,28 +650,8 @@ export const queryText = createDictionary({
   importNoReadPermission: {
     'en-us':
       'The query you imported contains tables you do not have read access to:',
-    'es-es': `
```
maxxxxxdlp, line 653: Was this removed accidentally?

specifyweb/frontend/js_src/lib/components/Interactions/PrepDialogRow.tsx
```diff
@@ -79,8 +83,19 @@ export function PrepDialogRow({
             max={preparation.available}
             min={0}
             title={interactionsText.selectedAmount()}
-            value={selected}
-            onValueChange={handleChange}
+            value={
```
maxxxxxdlp, line 86: This (very complicated) logic only affects what's visible in the field, not what's going to be sent to the server when user submits the form

maxxxxxdlp: You shouldn't have to change anything in this file.
Instead, in the `onValueChange` for the bulk value input field, call setSelected to change how many are selected in each

specifyweb/frontend/js_src/lib/localization/common.ts
```diff
@@ -594,4 +594,7 @@ export const commonText = createDictionary({
     'ru-ru': 'не в сети',
     'uk-ua': 'офлайн',
   },
+  bulkValue: {
+    'en-us': 'Bulk value',
```
maxxxxxdlp, line 598: Rename to `Bulk Select`

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):
@grantfitzsimmons can you take a look and see if you like the position of the bulk select input field? Or should it be above the table?

grantfitzsimmons:
I think it should be above the table, but it looks good!

My only request would maybe be that when "Select All" is selected the Bulk Select goes to its max value.

grantfitzsimmons (approved):

grantfitzsimmons:
Don't merge until @chanulee1 has had some time to test as well

chanulee1 (changes requested):
Does not select preparations for initial count of bulk value. Must increment and then decrement to select number of preps equal to the initial count.


https://user-images.githubusercontent.com/103299204/224203330-8e8b1c27-699c-4408-b630-f2998bf4d11a.mp4

https://coldfish-issue-2549.test.specifysystems.org/specify/

grantfitzsimmons:
> Does not select preparations for initial count of bulk value. Must increment and then decrement to select number of preps equal to the initial count.

The default bulk select value should be 0 instead of 1 so that this issue does not occur

chanulee1:
Default bulk select issue is fixed. But, "select all" and "deselect all" buttons do not affect bulk select value. Is this an issue?

https://user-images.githubusercontent.com/103299204/224378863-66585b59-51f8-46f9-9330-842e9a09791e.mp4

CarolineDenis:
@chanulee1 , The select all should not modify the bulk value

chanulee1 (approved):

maxxxxxdlp (changes requested):
I would echo what Grant said and suggest that this is put above the table as it's current position looks weird (because some cells are blank) and is confusing (because it changes the number of rows in the table - which is sort of data misrepresentation)

Can we put it above the table?
ie, something like: `Bulk Select: [_0_______]` (just a label and a field, excuse the ascii art)

maxxxxxdlp (approved):

PR #3048, opened 2023-02-24
Delete unnecessary separator in combo box

### LINKED ISSUE #2333: Agent Query Combo Box adding space before short names
**Describe the bug**
The Agent Query Combo Box seems to be adding a space to the box after an agent has been selected with a small(see below video). This then messes with the search algorithms (regardless of what the user is using) and the text has to be entirely removed to make a similar query. 
I am not sure what the effective cutoff is for name length, but this bug is seemingly not occurring for long names.

**Screenshots**

https://user-images.githubusercontent.com/64045831/196284139-70a6ab65-c827-463f-944c-093a95d14aa2.mp4

**Desktop:**
 - OS: Windows 11
 - Browser: Chrome
 - Specify 7 Version: testability

**Database Name:**
willemcatno

CarolineDenis:
Fixed #2333


specifyweb/frontend/js_src/lib/localization/query.ts
```diff
@@ -650,28 +650,8 @@ export const queryText = createDictionary({
   importNoReadPermission: {
     'en-us':
       'The query you imported contains tables you do not have read access to:',
-    'es-es': `
```
maxxxxxdlp, line 653: Not sure why this was removed once again

maxxxxxdlp: If you want, I can add it back myself

maxxxxxdlp (changes requested):
If

specifyweb/frontend/js_src/lib/localization/query.ts
CarolineDenis: ok thank you

maxxxxxdlp: Note, I also renamed the commit message to `Remove unnecessary separator in combo box`

(the commit message should be structured like `This commit will <commit message>` - `will` should not be part of the commit message)

Also `Remove` is better than `Delete` in this context (delete is better suited for files, resources)

maxxxxxdlp:
Also had to replace `.reduce(...) as LocalizedString` with `.reduce<LocalizedString>(...)` to fix the type error

chanulee1 (changes requested):
Issue as described in #2333 still appears to happen.


https://user-images.githubusercontent.com/103299204/221667368-4d042b7f-620b-4583-a225-17d8fbe837bc.mp4

https://willemcatno-issue-2333.test.specifysystems.org/specify/view/collectingevent/new/

chanulee1 (changes requested):
Seems to remove separator when it is only whitespace. Does not remove when it contains other characters.

https://user-images.githubusercontent.com/103299204/222232898-ce9eba28-11da-475f-b04f-c6d5dfdc3e14.mp4

specifyweb/stored_queries/views.py
```diff
@@ -79,6 +79,11 @@ def ephemeral(request):
 
     check_permission_targets(collection.id, request.specify_user.id, [QueryBuilderPt.execute])
     data = run_ephemeral_query(collection, request.specify_user, spquery)
+    for idx in range(len(data['results'])):
```
maxxxxxdlp, line 82: hmm, I'm not sure what this change is about.
I was expecting the change would be made in this file - https://github.com/specify/specify7/blob/production/specifyweb/stored_queries/format.py (to mirror the front-end change)

maxxxxxdlp:
@CarolineDenis here is the code that currently adds the separator on the back-end: https://github.com/specify/specify7/blob/68200c21651134fe2fb43f669793f1404f67b7f7/specifyweb/stored_queries/format.py#L126-L127

Looks like it is constructing a SQL query, thus fixing this would be quite a bit harder than it is on the front end.

Thus, please undo https://github.com/specify/specify7/pull/3048/commits/8e07c451a091e2d2c23b24083f8cc37be2ed592a  (as it's hardcoded to assume there is only one field in the query and that field is "(formatted)")
and then let's merge this with just the front-end side being fixed

maxxxxxdlp (approved):

chanulee1 (approved):
Looks good. Unnecessary separator removed on save.

PR #3052, opened 2023-02-26
fix issue 2399 Swagger UI try it out config

### LINKED ISSUE #2399: Swagger "try it" function broken.
Hey Folks

I’ve just been trying to write a simple API to specify.  

To start out, I thought I’d try to use the embedded swagger API docs, which include a really handy “try it out” function to try out the different API endpoints.  Ufortunately, it looks like these don’t actually work.  – maybe some mis-configuration in the swagger setup?

What I did was select the “POST” operation on the permits table, hit the try it out button, and left parameters at defaults.  (wasn’t expecting this to work fully – would expect to need to tweak a parameters, but was expecting it to at least pass authorization.  

What happened instead was I got a 403 – details below.  - key line is "<p>CSRF verification failed. Request aborted.</p>"

Is the swagger try it function meant to work?  
Thanks.  

image

seen on RBGE sandboc hosted instance, running 7.7.5

cdamyx:
Fixes #2399

Need to pass CSRF token for all requests to work. Ran frontend tests and rebased fork before submitting.


maxxxxxdlp:
To test:
Use swagger UI to make any POST or PUT request and make sure it is successful
GET requests should still be working, just like they were before the fix

maxxxxxdlp (approved):
Looks good. Thanks!

PR #3060, opened 2023-02-27
Tooltip fixes

### LINKED ISSUE #3041: Confusing behavior for tooltips
https://montreal-v7.test.specifysystems.org/specify/view/collectingevent/new/

https://user-images.githubusercontent.com/37256050/220759436-f4261156-b44a-4504-a68a-509bbf1b6667.mov

Why does it seemingly randomly appear? It seems like you can hardly go a second without some tooltip

Firefox, macOS

maxxxxxdlp:
Fixes the bugs with tooltips trigger and dismissal
Fixes #3041

Goals:
- Triggering tooltips unintentionally should be hard
- Triggering tooltips intentionally should be easy
- Dismissing tooltips intentionally should be easy
- Dismissing tooltips unintentionally should be hard

Given that, test the following edge cases:
- Tree dialog, hover over "Edit" until tooltip is visible, and then click on it
- SubView in a dialog -> hover over "Add" until tooltip is visible and then click on it
- SubView in a dialog -> click on "Add" before tooltip is visible
- Click on sub view button before tooltip is visible
- Click on sub view button after tooltip is visible
- Rapidly move the cursor around several elements that have tooltips (i.e, timestamp created column in queries dialog)

Possible bugs to look out for:
- Tooltip is visible, even though the element it was triggered on is now behind the overlay (i.e, dialog is opened, but the too
- After moving cursor between elements rapidly, can't trigger the tooltip anymore on some elements that you used to be able to


grantfitzsimmons (approved):
There are some minor issues that I encountered, such as modern tooltips not appearing at all after opening a dialog on some labels (though the browser default tooltips showed instead). I am very satisfied with this implementation because I'd rather be conservative (where you can mouse off and back on to trigger it intentionally) than show tooltips when you don't want to see them.

maxxxxxdlp:
@grantfitzsimmons that sounds like a bug. Can you please attach a video (or open a separate issue)
That is expected behavior for disabled elements, but all others should have modern tooltips

specifysoftware:

https://discourse.specifysoftware.org/t/specify-7-8-7-release-announcement/1079/1

PR #3065, opened 2023-02-28
Integrate Specify Network into Specify 7

### LINKED ISSUE #3064: Add Leaflet.GestureHandling plugin

maxxxxxdlp:
Integrate Specify Network into Specify 7 instead of it being a separate page

Also makes leaflet navigation easier on mobile:
Fixes #3064


maxxxxxdlp:
To test:
- Specify Network badges on Collection Object and Taxon form
  - Styling in dark mode and light mode
  - Behavior and UX (should match the features of the old specify network page, but present in a more flexible way)
- Go to taxon tree viewer. Click on any node that has collection objects. Click on "Query". Click on "GeoMap" - the map should have all the same features as the map in Specify Network badge
- In query builder, add some fields and sort on them and make sure that the headings in the query results table correctly display the sorting.
  - Clicking on the heading, should correctly change the sort position.
  - In particular, test a query with CO as a base table, and a single field to Locality, where that field is the first field in the query - make sure the sorting for the last field works
  - Do similar testing for audit log queries
- Switching base layers in leaflet map
  - Last used layer should be remembered between sessions
- Toggling visibility of overlay layers
  - Visibility of overlays should be remembered between sessions

maxxxxxdlp:
Note, a lot of the code in this PR was copied from https://github.com/specifysystems/specify_cache/tree/join_sn/sppy/frontend/js_src so that all of it is in-house. That code was used in Specify Network for the last year and thus have been tested. Thus the testing should be more focused on whether the code was integrated back into specify correctly, and that nothing was broken in the process

maxxxxxdlp:
Screenshots:

Screenshot 2023-03-28 at 20 56 23
Screenshot 2023-03-28 at 20 57 06

grantfitzsimmons (changes requested):
image

Icons for GBIF and iDigBio are not loading, as seen above.

image

Anytime that I click on the "Loc" button, I get this error.

[Specify 7 Crash Report - 2023-04-27T02_49_48.355Z.txt](https://github.com/specify/specify7/files/11339109/Specify.7.Crash.Report.-.2023-04-27T02_49_48.355Z.txt)

https://sp7demofish-specify-network-integration.test.specifysystems.org/specify/view/collectionobject/51591/?recordsetid=14

PR #3083, opened 2023-03-02
Allow to drag a query line in the query builder

### LINKED ISSUE #1282: Bring back drag sort in Query Builder
Add a draggable region to query lines like this:
image

CarolineDenis:
Fixed #1282

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
@@ -1,8 +1,13 @@
+import { event } from 'jquery';
 import React from 'react';
 
 import { useReadyEffect } from '../../hooks/useReadyEffect';
+import { commonText } from '../../localization/common';
 import type { RA } from '../../utils/types';
 import { Ul } from '../Atoms';
+import { Button } from '../Atoms/Button';
+import { className } from '../Atoms/className';
+import { icons } from '../Atoms/Icons';
 import type { Tables } from '../DataModel/types';
 import { ErrorBoundary } from '../Errors/ErrorBoundary';
 import { scrollIntoView } from '../TreeView/helpers';
@@ -62,9 +67,7 @@ export function QueryFields({
 }): JSX.Element {
   const fieldsContainerRef = React.useRef<HTMLUListElement | null>(null);
 
-  //************************************************** */
   const [draggedItem, setDraggedItem] = React.useState<QueryField | null>(null);
-  const [list, setList] = React.useState<RA<QueryField>>(fields);
 
   const handleDragStart = (
     event: React.DragEvent<HTMLDivElement>,
@@ -104,8 +107,6 @@ export function QueryFields({
     handleChangeFields(newItems);
   };
 
-  //************************************************** */
-
   // Scroll to bottom if added a child
   const oldFieldCount = React.useRef(fields.length);
   // REFACTOR: extract this into hook and use everywhere where applicable
@@ -131,6 +132,7 @@ export function QueryFields({
       {fields.map((field, line, { length }) => (
         <ErrorBoundary dismissible key={field.id}>
           <div
+            className="flex items-center gap-2"
             key={line}
             draggable={true}
             onDragStart={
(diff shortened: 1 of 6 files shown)


realVinayak:
Overall, a nice idea. It seems a bit glitchy when dragging items, they keep switching places if they are right below each other. Also, is there a reason why the dragging item is less faded towards the end?
image

realVinayak:
Can we also add a scroll? So if you have a long list of fields, if automatically scrolls down when the dragging item is held below the last item? Same with scrolling up.

realVinayak:
Also, it seems like the items between the drag source and destination are re-rendered. I see the 'Loading...' in the field filter. Is that a priority to fix?

CarolineDenis:
For the "glitching effect": 

- When dragging the element, the browser is constantly redrawing the page as the item is being dragged. This can cause visual glitches to appear as the browser tries to keep up with the movement of the item.

- Additionally, the dragging behavior may be affected by the browser's implementation of the drag-and-drop API, which can vary between browsers and operating systems. Some browsers may handle dragging and dropping more smoothly than others, and some may have limitations or bugs that can affect the behavior.

For the "loading": 
- I couldn't replicate it on my end so I would say it's a part of the recalculation and re-load that is part of the dragging feature. It will depend on the browser, computer

For the scroll to bottom: 
- I will look to see how to implement that

maxxxxxdlp:
Able to replicate "Loading..." consistently:

https://user-images.githubusercontent.com/40512816/223785445-ecddd643-d932-4eec-be03-140a2df3748f.mov


See https://github.com/specify/specify7/blob/22a5a0e879187a4308b65ef5b081e798e603ae64/specifyweb/frontend/js_src/lib/components/QueryBuilder/FieldFilter.tsx#L564

Happens because of https://github.com/specify/specify7/blob/22a5a0e879187a4308b65ef5b081e798e603ae64/specifyweb/frontend/js_src/lib/components/QueryBuilder/FieldFilter.tsx#L539-L550

Not a big deal, but something to look into in the future

specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
```diff
@@ -61,9 +62,71 @@ export function QueryFields({
...
   readonly onOpenMap: ((line: number) => void) | undefined;
+  readonly onChangeFields?: ((fields: RA<QueryField>) => void) | undefined;
 }): JSX.Element {
   const fieldsContainerRef = React.useRef<HTMLUListElement | null>(null);
 
+  const [draggedItem, setDraggedItem] = React.useState<QueryField | null>(null);
+
+  const [isDragging, setIsDragging] = React.useState(false);
+
+  React.useEffect(() => {
```
maxxxxxdlp, line 73: This is quite buggy:

https://user-images.githubusercontent.com/40512816/223826943-48baaf96-07db-4d91-b00b-08673c1f4684.mp4

Either make it less buggy, or look into using libraries that handle this very well (after doing a bit of research, found: react-dnd, https://dev.to/augani/how-to-create-a-smooth-draggable-list-with-react-in-10-mins-2fbn, https://codesandbox.io/examples/package/react-draggable-list, https://github.com/atlassian/react-beautiful-dnd, https://shopify.github.io/draggable/

specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
```diff
@@ -61,9 +62,71 @@ export function QueryFields({
...
+        listNode.scrollTop += bottom - windowHeight;
+      }
+    }
+  }, [isDragging]);
+
+  const handleDragStart = (
+    event: React.DragEvent<HTMLDivElement>,
+    index: number
+  ) => {
+    setDraggedItem(fields[index]);
```
maxxxxxdlp, line 94: I am assuming a lot of the code in this component is based on online/examples. If so, can you provide links to those places as comments before the code?

i.e, I often put in the code something like

```js
/**
 * Inspired by:
 * <link>
 */
```

```js
/**
 * Based on:
 * <link>
 */
```

or just docs that you think would be helpful for someone looking at this in the future

maxxxxxdlp (changes requested):

maxxxxxdlp:
Though, it works pretty well on mobile! (tested on ios)

CarolineDenis:
Handle the feature with @shopify/draggable library in this PR: 
https://github.com/specify/specify7/pull/3143

PR #3093, opened 2023-03-03
V7.8.7.1 fix

melton-jason:


specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
```diff
@@ -111,13 +111,15 @@ function Field({
...
         field?.isRelationship === true
           ? hasTablePermission(field.relatedModel.name, 'read')
-            ? (
-                resource?.rgetPromise(field.name) as Promise<
-                  SpecifyResource<AnySchema> | undefined
-                >
-              )
-                .then(format)
-                .then((value) => value ?? '')
+            ? resource?.specifyModel.getField(field?.name) === undefined
```
maxxxxxdlp, line 114: Weird. This should be caught during form parsing - i.e, instead of each plugin and each field checking in every place whether a field is valid, we just check it once during form parsing.
This fix is okay as a hotfix, but not a good long-term solution

specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts
```diff
@@ -27,10 +27,12 @@ export function getDomainResource<
   LEVEL extends keyof typeof schema.domainLevelIds
 >(level: LEVEL): SpecifyResource<Tables[Capitalize<LEVEL>]> | undefined {
   const id = schema.domainLevelIds?.[level];
-  if (id === undefined)
+  if (id === undefined) {
     console.error(
       `Trying to access domain resource ${level} before domain is loaded`
     );
+    return;
```
maxxxxxdlp, line 34: (optional) Can you change `return;` to `return undefined;` just to be explicit about it?

maxxxxxdlp (approved):

grantfitzsimmons (approved):
I tested these fixes on all of the problem deployments and the working ones:

- newmexico
- newmexico2
- rjd-madrid
- KU_FISH_NO_TESTS
- sp7demofish
- sp6fishtest2

It is working well!

specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
melton-jason: Yes, I would think this would be handled by form parsing. I figured it probably was in the past, and looled through some of the code but did not have much time to work out the 'right' solution.

Here is an example of something that would cause a crash:
A `locality.geography` field (QueryComboBox) is on the Collecting Event sub form definition.

This works fine for records which already have existing localities, but for records without Localities the `collectingEvent.rgetPromise(geography)` call returns undefined (rather than `Promise<undefined>`) which throws an error

maxxxxxdlp: ah, so just making rgetPromise function into an async function will make sure that the return value is always wrapped in a promise (i.e, instead of `rgetPromise(...)` do `async rgetPromise(...)` in resourceApi.js

maxxxxxdlp: Here is the real cause of the issue:
if `a` is undefined, `a?.b.c` does not throw an exception, but `(a?.b).c` does

So we just have to replace `(a?.b).c` with `(a?.b)?.c`

PR #3100, opened 2023-03-07
Update to Heroicons V2

### LINKED ISSUE #2842: Migrate to HeroIcons v2
We are currently on v1 solid.
v2 has come out.

Need to migrate to v2 20px (or 24px) solid
See changelog: https://github.com/tailwindlabs/heroicons/releases/tag/v2.0.0

List of icons - https://heroicons.dev/?iconset=v2-24-solid&size=24

No rush, but new icons would only be added to the `v2` version

grantfitzsimmons:
Fixes #2842

These are the 20px solid icons


grantfitzsimmons:
image

I find this more inconsistent in thickness.

grantfitzsimmons:
image

I'm not a huge fan of the new QCBX look with these icons either

maxxxxxdlp:
yeah, I also think v2 lost some of it's magic
other popular icon packs are quite ugly too (too much detail, edges too sharp)

we either should look for a different source of icons, or stay with v1 for as long as we can (and if absolutely needed, selectively use icons from v2 or other sources)

maxxxxxdlp:
> I find this more inconsistent in thickness.

Keep in mind that the trees icon was created by us. But true, others are incosistent

maxpatiiuk (changes requested):
See https://github.com/specify/specify7/pull/3100#issuecomment-1457481461

grantfitzsimmons:
> See [#3100 (comment)](https://github.com/specify/specify7/pull/3100#issuecomment-1457481461)

>yeah, I also think v2 lost some of it's magic
other popular icon packs are quite ugly too (too much detail, edges too sharp)
>we either should look for a different source of icons, or stay with v1 for as long as we can (and if absolutely needed, selectively use icons from v2 or other sources)

Maybe we should use some of v1 where the icons are dramatically nicer (chevrons, arrows, data entry icon, search, edit, plus, minus) and use v2 for the rest.

maxxxxxdlp:
Aren't they going to look inconsistent?
I also don't like the ambiguity of whether v1 or v2 needs to be used - it's best to commit to one and use it everywhere

PR #3112, opened 2023-03-08
Fix system information not being included in stack trace

maxxxxxdlp:
Because it's not loaded by the time stack trace is created
Along with a few other keys

To test:
Open http://localhost/specify/command/test-error/ (or any other way of generating a stack trace)
Download the error message
Make sure "systemInformation" key is present in the object


chanulee1 (approved):
'systemInformation' key is present in error message.

PR #3113, opened 2023-03-08
Improve error message for oversized attachments

### LINKED ISSUE #729: Check attachment file size before attempting to upload to asset server.
Currently the nginx proxy rejects uploads above a set size. The Sp 7 interface just reports this as an unknown error.

maxxxxxdlp:
Fixes #729

### DIFF
--- a/specifyweb/frontend/js_src/lib/utils/ajax/definitions.ts
+++ b/specifyweb/frontend/js_src/lib/utils/ajax/definitions.ts
@@ -10,6 +10,7 @@ export const Http = {
   NOT_FOUND: 404,
   FORBIDDEN: 403,
   CONFLICT: 409,
+  TOO_LARGE: 413,
   MISDIRECTED: 421,
   HUGE_HEADER: 431,
   SERVER_ERROR: 500,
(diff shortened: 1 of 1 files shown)


CarolineDenis (approved):

chanulee1:
How do I test this?

maxxxxxdlp:
Try to upload a 200mb file, and see that the error message includes:
```
    This error happened because you tried to upload a file that is larger than
    the configured server limit. Either contact your system administrator about
    increasing the limit, or try uploading a smaller file.
```

chanulee1 (changes requested):
Getting this error when trying to create a 300 mb collection object attachment.

image

[Specify 7 Crash Report - 2023-03-08T20_49_46.386Z.txt](https://github.com/specify/specify7/files/10925218/Specify.7.Crash.Report.-.2023-03-08T20_49_46.386Z.txt)

maxxxxxdlp:
I was able to replicate that (bad) error message.

Screenshot 2023-03-08 at 16 18 56


It appears to happen if the asset server run out of storage:

Screenshot 2023-03-08 at 16 18 17

And the reason the error is so bad, is that the asset server does not send proper headers in the response, which leads to chrome not allowing specify 7 to see it's response:

Screenshot 2023-03-08 at 16 19 40

Not much specify 7 can do about that, other than fixing this on the asset server side.

I am going to merge this PR and close the issue, but open another issue on the asset server's side

PR #3119, opened 2023-03-08
Add webpack-bundle visualizer option

### LINKED ISSUE #2610: Add webpack bundle visualizer to pull requests
Similar to #2595

When webpack build runs as part of a test in a pull request, it would be cool to generate a visualization of the bundle size.

Example:
Screenshot 2022-12-12 at 19 39 36

After that, would be good to add warnings when some bundle size is too size

CarolineDenis:
Fixed #2610

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,8 +26,10 @@ seed-database
 /specifyweb/frontend/static/js
 /specifyweb/frontend/static/css
 /specifyweb/frontend/static/manifest.json
-/specifyweb/frontend/js_src/node_modules
+node_modules
 /specifyweb/frontend/js_src/testBuild
 /specifyweb/frontend/js_src/docs
 /specifyweb/frontend/js_src/coverage
 /local_specifyweb_apache.conf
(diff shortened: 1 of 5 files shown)


maxxxxxdlp (approved):

maxxxxxdlp:
The visualization looks super cool!
@specify/dev-testing @grantfitzsimmons I am going to merge this, but encourage others to take a look out of curiosity

Really interesting that even though we use very few libraries, most of the bundle size is still comping from libraries (I added a red line around our code):

Screenshot 2023-03-08 at 10 53 28

The giant purple block at the top is handsontable (workbench spreadsheet library)
The big pink block at the top is for xlsx parsing
Thus, half of all shipped code is just 2 WB libraries!
The green is the main bundle - get's loaded before any page is displayed. Half of it is comprised of libraries, other half of our code
And then the small bundles at the bottom right are lazy-loaded for individual components (i.e, Security Page, Tree Viewer, Wb Import)

grantfitzsimmons:
That's so cool!

PR #3130, opened 2023-03-08
Fix export to KML not working

### LINKED ISSUE #3088: Export to KML does not work
Export a Locality query with coordinates to KML and see that it does not work.

Version 7.8.5 and 7.8.6 have been tested and it's not working.

**Reported By:** CSIRO

> Hey Grant, I've noticed potentially a big bug and I can't see a github for it - the KML export button seems to be broken? I just tested it on your demo site: https://sp7demofish.specifycloud.org/specify/query/61/

maxxxxxdlp:
Fixes #3088


melton-jason (approved):
Looks good! You can export to KML when the query contains coordinate data (latitude1 & longitude1).
Here is an example query that was tested on:

advancedQuery

When brought into Google Earth, the results and data from other fields are correctly imported:
advancedGoogleEarthInfo

PR #3132, opened 2023-03-09
Update "latlongaccuracy" to "geocoorddetails.maxuncertaintyest"

### LINKED ISSUE #1907: Leaflet error polygons use wrong field
When leaflets plots a locality, it retrives the error radius from the `Locality.latLongAccuracy` field.

`Locality.geoCoordDetails.maxUncertaintyEst` should be used instead, as that's the field that is used by GeoLocale.

cdamyx:
Fixes #1907

Went ahead and submitted PR so you can see where I'm at, and I'll correct after. Still a bit lost on this issue, and a couple of the tests are failing when I run npm test now. Sorry if I'm way off-base here...


maxpatiiuk (approved):
The code change looks good. Need to confirm whether it actually works. To test:
- Whether it can use the `geocoorddetails.maxuncertaintyest` when plotting localities of type "Point"
  - in the WorkBench
  - on the locality form
  - in the query builder
- Make sure GeoLocate is saving error radius into `geocoorddetails.maxuncertaintyest`

specifyweb/frontend/js_src/lib/components/WorkBench/GeoLocate.tsx
```diff
@@ -61,7 +61,7 @@ export function WbGeoLocate({
       const changes = [
         ['locality.latitude1', latitude],
         ['locality.longitude1', longitude],
-        ['locality.latlongaccuracy', uncertainty],
+        ['locality.geocoorddetails.maxuncertaintyest', uncertainty],
```
maxxxxxdlp, line 64: @grantfitzsimmons I was under impression that GeoLocate already used `geocoorddetails.maxuncertaintyest`. If it was already using `latlongaccuracy`, then there is no inconsistency with Leaflet. Which one should it be using?

cdamyx: When I try to recreate the steps from @grantfitzsimmons video, with the code changes from this PR still present on my local machine, I _do not_ get a value in the Max Uncertainty Est field...

maxxxxxdlp: Looks like the GeoLocate in forms was using `geoCoordDetails.maxUncertaintyEst` like expected, but the code in the workbench was incorrectly using `latlongaccuracy` (along with GeoMaps). This pull request fixes that

@grantfitzsimmons can you try out GeoMap in forms and WorkBench and GeoLocate in forms and WorkBench and check that the `locality.geocoorddetails.maxuncertaintyest` is used in all cases, rather than `latlongaccuracy`

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/Leaflet/config.ts
```diff
@@ -156,7 +156,7 @@ export const localityPinFields: RA<LocalityPinFields> = [
       ['Locality', 'latitude2'],
       ['Locality', 'longitude2'],
       ['Locality', 'latLongType'],
-      ['Locality', 'latLongAccuracy'],
+      ['Locality', 'geoCoordDetails', 'maxuUcertaintyEst'],
```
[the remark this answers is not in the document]
cdamyx: Good catch @maxxxxxdlp - unless I'm mistaken, I think there was also a typo in the change request above. Updated to `maxUncertaintyEst`. Thanks!

maxxxxxdlp:
A related comment from @acbentley on this (from a year ago):
> The Darwin Core concept coordinate uncertainty in meters is what should be (and is usually) mapped to maxUncertaintyEstimate in the Specify data model and what is used by Geolocate to insert the coordinate radius created in Geolocate in meters

maxxxxxdlp:
@realVinayak or @melton-jason could you please pick up this PR?
The only thing left is to do light testing of changes. The tricky bit is that this branch is not available on the test panel (because it's from an external contributor), which is why this PR can't be tested by Grant or Chanu - you will have to deploy this branch locally

grantfitzsimmons (changes requested):
Screenshot 2023-08-01 at 9 16 19 AM

When you have a record with coordinates, GEOLocate no longer shows the coordinates properly in the text box

grantfitzsimmons:
@carlosmbe Do you want to pick up on this sometime in the future?

carlosmbe:
> @carlosmbe Do you want to pick up on this sometime in the future?

In a testing or development sense?

grantfitzsimmons:
Development sense– it is not working for me in my last review

carlosmbe:
@grantfitzsimmons Should this be closed or will @pashiav pick this up later?

PR #3147, opened 2023-03-10
Display only name for new record set dialog

### LINKED ISSUE #2782: New Record Set form should only show name
image

I know it is based on which fields are hidden/unhidden, but this should _only_ ever show the **Name**. There isn't really a valid use case in the "new" dialog for anything else to display.

Maybe when navigating directly to the record set form this can display, but not in the new dialog

https://kuvertpaleo11123-edge.test.specifysystems.org/specify/overlay/record-sets/

CarolineDenis:
Fixed #2782


maxxxxxdlp (approved):

chanulee1 (approved):
Only name displayed in new records set dialog, even when other fields are unhidden.

maxxxxxdlp (approved):

PR #3149, opened 2023-03-10
Add copy/pasting detection

### LINKED ISSUE #2996: Add a code duplication scanner
Run it as part of GitHub tests (i.e, https://github.com/kucherenko/jscpd/pull/463)
Will help automate code-review process

CarolineDenis:
Fixed #2996

edit: [testing instructions](https://github.com/specify/specify7/pull/3149#issuecomment-2184248825)


chanulee1:
How do I test this?

maxxxxxdlp:
This is likely not ready for testing yet because the front-end tests are failing - but that's the intention of the change

The change will make front-end tests fail if code duplication is detected.

As soon as code duplication issues are fixed, the tests will be pass, and thus this will be ready for merging

maxpatiiuk (approved):

maxxxxxdlp:
I could see one issue with this though, not sure it's important enough to fix, but something to consider:

I made a change where back-end test are only run on back-end changes, and front-end tests are run on front-end changes.

However, the code duplication check is checking both front-end and back-end - yet is only run if the front-end changes.

The worst case is that a pull request that added code duplication on the back-end gets merged, and then front-end test will start failing for people.

The fix for this would be to modify `test.yml` to have a third category of tests - the tests that run always, regardless of what files where changed

maxxxxxdlp:
a nice side-effect of adding this library:
it prints a nice summary of the size of the code base - https://github.com/specify/specify7/actions/runs/4407952016/jobs/7722317168

Screenshot 2023-03-13 at 12 57 07



(and it ignores the test files and third-party libraries, so the numbers are not inflated like in sp6)

Shows that front-end is `23946+ 29135+ 49773+ 483 = 103,337` lines of code
Back-end is about `18352+15+755 = 19,122`

specifyweb/frontend/js_src/lib/components/DataModel/legacyTypes.ts
```diff
@@ -43,20 +58,22 @@ export type SpecifyResource<SCHEMA extends AnySchema> = {
...
+    //   | keyof SCHEMA['toOneDependent']
+    //   | keyof SCHEMA['toOneIndependent'],
+    // VALUE extends (CommonFields &
+    //   IR<never> &
+    //   SCHEMA['fields'] &
+    //   SCHEMA['toManyDependent'] &
+    //   SCHEMA['toManyIndependent'] &
+    //   SCHEMA['toOneDependent'] &
+    //   SCHEMA['toOneIndependent'])[FIELD_NAME]
+    FIELD_NAME extends AllowedFields<SCHEMA>,
```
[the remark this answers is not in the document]
maxxxxxdlp: Instead, here is what you could do:
Instead of creating a new type, AllowedField, use `keyof CommonFields | TableFields<SCHEMA>`
(TableFields is a type defined in `components/DataModel/helperTypes.ts`

Then, don't create a new type for AllowedValues, but leave it inlined as it was - since it's just 8 lines, the jscpd should not trip up anymore. If it does, then add a comment to silence it.

The reason I am against creating `AllowedValues` is that it would be more complicated to de-duplicate the code than it's worth (i.e, you would need to pass both SCHEMA and FIELD_NAME to AllowedValues - which might work, but TypeScript might not be able to handle that well as this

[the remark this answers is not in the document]
maxxxxxdlp: Actually, I am still getting typescript errors even with that change:

```ts
FIELD_NAME extends keyof CommonFields | keyof TableFields<SCHEMA>,
```

let's just leave this file as is and add a comment to make jscdp ignore the copy/paste. This file is called legacyTypes.ts because this will be deleted once we transition away from the legacy librarty

maxpatiiuk (changes requested):

maxpatiiuk (approved):
All good (besides the failing tests)

maxxxxxdlp:
@melton-jason can you please take a look into the following code duplication occurrences:

```
Clone found (python):
 - /home/runner/work/specify7/specify7/specifyweb/workbench/upload/tomany.py [59:20 - 70:3] (11 lines, 130 tokens)
   /home/runner/work/specify7/specify7/specifyweb/workbench/upload/upload_table.py [94:19 - 105:7]

Clone found (python):
 - /home/runner/work/specify7/specify7/specifyweb/workbench/upload/tomany.py [92:9 - 111:2] (19 lines, 249 tokens)
   /home/runner/work/specify7/specify7/specifyweb/workbench/upload/upload_table.py [182:9 - 201:2]
```

see how easy it is to fix them (if deduplicating them is adding a lot of complexity, then we can silence the duplication check for those two occurrences)

maxxxxxdlp:
@realVinayak can you also look into resolving the following. should be relatively simple

```
Clone found (python):
 - /Users/maxpatiiuk/site/python/specify7/specifyweb/context/collection_resources.py [14:68 - 40:2] (26 lines, 194 tokens)
   /Users/maxpatiiuk/site/python/specify7/specifyweb/context/user_resources.py [31:89 - 57:17]

Clone found (python):
 - /Users/maxpatiiuk/site/python/specify7/specifyweb/context/collection_resources.py [55:9 - 79:13] (24 lines, 205 tokens)
   /Users/maxpatiiuk/site/python/specify7/specifyweb/context/user_resources.py [58:9 - 82:14]

Clone found (python):
 - /Users/maxpatiiuk/site/python/specify7/specifyweb/context/collection_resources.py [95:71 - 118:72] (23 lines, 194 tokens)
   /Users/maxpatiiuk/site/python/specify7/specifyweb/context/user_resources.py [97:92 - 120:101]

Clone found (python):
 - /Users/maxpatiiuk/site/python/specify7/specifyweb/context/collection_resources.py [118:72 - 136:6] (18 lines, 149 tokens)
   /Users/maxpatiiuk/site/python/specify7/specifyweb/context/collection_resources.py [39:79 - 57:6]

```

melton-jason:
> @melton-jason can you please take a look into the following code duplication occurrences:
> 
> ```
> Clone found (python):
>  - /home/runner/work/specify7/specify7/specifyweb/workbench/upload/tomany.py [59:20 - 70:3] (11 lines, 130 tokens)
>    /home/runner/work/specify7/specify7/specifyweb/workbench/upload/upload_table.py [94:19 - 105:7]
> 
> Clone found (python):
>  - /home/runner/work/specify7/specify7/specifyweb/workbench/upload/tomany.py [92:9 - 111:2] (19 lines, 249 tokens)
>    /home/runner/work/specify7/specify7/specifyweb/workbench/upload/upload_table.py [182:9 - 201:2]
> ```
> 
> see how easy it is to fix them (if deduplicating them is adding a lot of complexity, then we can silence the duplication check for those two occurrences)

These duplicates are related to the fact that subclasses of Named Tuples are not meant to be subclassed themselves. 
More specifically, subclassing _is_ allowed, but new attributes/data can not be added to the NamedTuple, only methods.
We have conversed about this issue before, in https://github.com/specify/specify7/pull/3240#discussion_r1149507690

In short, the backend is extensively using Named Tuples to store information about the status of an upload.  
Named Tuples are convenient because it is very easy to convert between Named Tuples and JSON: 
https://github.com/specify/specify7/blob/393538b081eb797beb502204cdea9311179361f6/specifyweb/workbench/upload/upload_plan_schema.py#L228-L245)
and NamedTuples are immutable once created. 

Through the process of uploading or validating a workbench upload plan, the backend 'converts' NamedTuple classes into ones that have more information or are more 'specialized', up until the actual upload/validation. (A typical example of the conversion process is UploadTable -> ScopedUploadTable -> BoundUploadTable)

Because we can not subclass a generic UploadTable in ScopedUploadTable, BoundUploadTable, etc., we have to reuse a lot of the 'generic' functionality common to all of them.

maxxxxxdlp:
@melton-jason not a problem we can leave it as is given that it is presenting issues. We have the capability to silence those errors by adding comments like this - https://github.com/specify/specify7/pull/3149/commits/9c52034bc03b0c5ea65917c2de28a78a4267f749#diff-c8cfcd0e13a828573c0d195ee26ce1cbb513e3554e37dcb0860e9053ffa16568R73-R121


Though, the issues in https://github.com/specify/specify7/pull/3149#issuecomment-1516725371 would be easy to fix so should be fixed

CarolineDenis:
@melton-jason do we want to merge this at some point?

melton-jason:
Yes, we can look into merging this. 
We should probably silence the copy/paste detection as a result of https://github.com/specify/specify7/pull/3149#issuecomment-1516801273 and look into the errors from
https://github.com/specify/specify7/pull/3149#issuecomment-1516725371

carlosmbe:
Is this a feature that requires any UX Testing? @CarolineDenis

CarolineDenis:
@carlosmbe , no it does not

CarolineDenis:
@specify/dev-testing 
What do we want to do with this PR?

maxpatiiuk:
Would love to see this merged.
If you need help, I can work on this this Saturday

realVinayak:
regarding backend workbench, is there a way to silence the workbench ones for now? They should be fixed in this [PR](https://github.com/specify/specify7/pull/4929) - I reworked the matching, and deleted everything from the tomany.py file 😄 (made the logic completely generic, no need for that file anymore, see [diff](https://github.com/specify/specify7/pull/4929/commits/6821a887d26944fd42b2c9d71d77b252eb3d2a3d#diff-53998e45444c1bb212b1c80674fb0b7711d17175fd6ca0e0d2634d54552edc46)).

.gitignore
```diff
@@ -1,36 +1,43 @@
...
 *.pyc
-*~
-.vagrant
+
+# OS
+.DS_Store
+
+# Editors
+/.vscode/
+/.idea/
```
maxpatiiuk, line 15: I used to be against OS or editor specific gitignore patterns, and was recommending using [global gitignore for that](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) but I changed my mind.

Reasons:
- Setting up global gitignore is added friction, especially for new team members or external contributors.
- Automated tools (like jscpd, eslint, prettier) and some IDEs do not look into global gitignore file

maxpatiiuk:
I just realized that jscpd will also help us detect incorrect merge resolutions (i.e cases when a deleted file got incorrectly "reviewed" during merging). Pretty nice

maxpatiiuk:
Testing instructions:
- Ensure the `Form Editor/App Resource` buttons at the top of the app resource still work and look good
- Ensure the `Table Format/Table Aggregation` buttons at the top of a "Record Formaters" visual editor still look good
- Open User Tools -> Operations API and make sure the documentation for the `/context/user_resource/*` and `/context/collection_resource/*` endpoints looks similar or better to prodcution

specifyweb/frontend/js_src/lib/components/DataModel/legacyTypes.ts
```diff
@@ -42,20 +61,8 @@ export type SpecifyResource<SCHEMA extends AnySchema> = {
...
-      | keyof SCHEMA['toOneDependent']
-      | keyof SCHEMA['toOneIndependent'],
-    VALUE extends (CommonFields &
-      IR<never> &
-      SCHEMA['fields'] &
-      SCHEMA['toManyDependent'] &
-      SCHEMA['toManyIndependent'] &
-      SCHEMA['toOneDependent'] &
-      SCHEMA['toOneIndependent'])[FIELD_NAME]
+    FIELD_NAME extends SchemaFields<SCHEMA>,
```
maxpatiiuk, line 64: Re-run `npx tsc --extendedDiagnostics` several times with and without this change
seems like this change makes typescript type-checking of entire project faster between 0s and 1s (possibly because TypeScript can cache the resolved type?)

specifyweb/workbench/upload/upload_table.py
```diff
@@ -37,6 +37,7 @@ def get_cols(self) -> Set[str]:
             | set(col for u in self.toOne.values() for col in u.get_cols()) \
             | set(col for rs in self.toMany.values() for r in rs for col in r.get_cols())
 
+    # jscpd:ignore-start
```
maxpatiiuk, line 40: > regarding backend workbench, is there a way to silence the workbench ones for now? They should be fixed in this https://github.com/specify/specify7/pull/4929 - I reworked the matching, and deleted everything from the tomany.py file 😄 (made the logic completely generic, no need for that file anymore, see [diff](https://github.com/specify/specify7/pull/4929/files#diff-53998e45444c1bb212b1c80674fb0b7711d17175fd6ca0e0d2634d54552edc46)).

Sounds good. Leaving these to you

maxpatiiuk:
Testing instructions:
- Verify that the Tabs buttons in app resources still look good after the change
- Verify that the "Table Format"/"Table Aggregation" tabs in the

CarolineDenis:
@maxpatiiuk, 
Do you know if there is a possibility to have this ran when the dev runs a command locally instead of having it as failing test / passing test in our actions?

maxpatiiuk:
You can run the check locally at any point by calling this command from the `js_src` folder:

```sh
npx tsx lib/tests/jscpd.ts
```

We could have it run automatically in a pre-commit hook, however the command takes 11s to check entire specify7 repository, so this would slow down commits considerably.

I know that IntelliJ IDEs had this built-in live in the editor. Maybe there is a VS Code plugin for that too - but I imagine it would slow down the IDE a bit.

sharadsw (approved):

pashiav (commented):
- [X] Verify that the Tabs buttons in app resources still look good after the change
- [ ] Verify that the "Table Format"/"Table Aggregation" tabs in the
- [X] Go to User Tools -> Operations API -> Verify that the documentation for the `/context/user_resource/*` and `/context/collection_resource/*` endpoints looks good (or better) than production
---

Looks good so far! For the "Table Format"/"Table Aggregation" tabs, what exactly should we be focusing on testing here?

maxpatiiuk:
ups! looks like I lost my train of though when I typing that 😆
I meant to say that they are supposed to look identical on this branch to how they do on main branch

Issue #1055, opened 2021-11-08
Taxon.isAccepted shouldn't be mappable in the WB.


benanhalt:
Because it is a (redundant) system field that should be set automatically.

Related to #1054

PR #3157, opened 2023-03-13
Sets isAccepted field by default for wb tree uploads

### LINKED ISSUE #2361: Taxon uploads in the WorkBench do not have isAccepted set by default
Any taxon records uploaded through the WorkBench in Specify 7 do not have `isAccepted` set by default.

This can be seen in `newmexico2` currently, but that is subject to change. This has caused issues in cases where the full name display format is modified as the full name is not built for non-preferred taxon records.

chanulee1:
Fixes #2361


maxxxxxdlp:
@grantfitzsimmons are there cases where back-end should not be setting isAccepted to true?

I suppose the only one is if you added a "Prefered Taxon" relationship to the mapping, but that's not supported right now - though maybe the code should account for it now, so as not to create a bug in the future.

i.e, if Taxon -> preferredTaxon mapping is in the data set and it's not empty, then don't set isAccepted to try?

melton-jason:
Right now, this fix would be OK, as it doesn't look like there is any way to 'synonymize' a Taxon record through the Workbench.

> I suppose the only one is if you added a "Prefered Taxon" relationship to the mapping, but that's not supported right now - though maybe the code should account for it now, so as not to create a bug in the future.
> 
> i.e, if Taxon -> preferredTaxon mapping is in the data set and it's not empty, then don't set isAccepted to try?

If we do add support for synonymizing through the Workbench (and mapping Determination -> preferredTaxon), another use case we need to consider is both pre-existing Taxon records that may become synonymized, and Taxon records in the dataset which should become synonymized through other Taxon records in the dataset.

grantfitzsimmons:
> @grantfitzsimmons are there cases where back-end should not be setting isAccepted to true?

If `isAccepted` is set to false in the import mapping

melton-jason:
> > @grantfitzsimmons are there cases where back-end should not be setting isAccepted to true?
> 
> If `isAccepted` is set to false in the import mapping

@grantfitzsimmons Are you able to map `isAccepted` in the Workbench? If so, then that is a bug. The field should be hidden in the Workbench as of #1055. 
(Or are you speculating of a bigger, better Workbench?)

chanulee1:
@melton-jason @grantfitzsimmons I could add a check to see if the field is in the dataset before setting the default. Would that work?

grantfitzsimmons:
@chanulee1 That sounds perfect. What do you all think?

carlosmbe (commented):
For anyone curious why I closed the issue without approving this PR. It's because for some reason, all WB Taxon /Determination Taxon started uploading with `isAccepted` being true. I tested it in NHMD, MOR Paleo and inverts

melton-jason:
>  It's because for some reason, all WB Taxon /Determination Taxon started uploading with isAccepted being true. I tested it in NHMD, MOR Paleo and inverts

Likely because a business rule was added in `v7.9.0` via https://github.com/specify/specify7/commit/2c2faa9e9241753988d3e7067ccdb63a2b549f39 to set isAccepted based on whether the object is synonymized or not before the object is saved. 
This PR can likely be closed, as it was superseded by the aforementioned commit in #2965

PR #3172, opened 2023-03-14
Remove redudant scroll bar in schema config

### LINKED ISSUE #3167: Schema Config has redundant scroll bars
(under my user agent style sheet scroll bars are visible as 1px white strips, but the bug is reproducable even without that)

Notice the white line in bottom and right corner
Also, notice the text being cut of in first column if you scroll right

Screenshot 2023-03-13 at 19 48 21

Caused by this change:

https://github.com/specify/specify7/commit/1a7d7e568#diff-8bf9d9d2b07370b0ab9e7e9e68f3798b5a9b6542c065cc90c177c9a049993c53

@CarolineDenis do you remember why that was made?

Reverting that or removing `overflow-hidden` fixes this issue (but causes other?)

CarolineDenis:
Fixed #3167

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/SchemaConfig/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/SchemaConfig/index.tsx
@@ -112,7 +112,7 @@ export function SchemaConfigMain(): JSX.Element {
         languages={schemaData.languages}
         onSave={canSave ? handleSave : undefined}
       />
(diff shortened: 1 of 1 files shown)


maxxxxxdlp (approved):

chanulee1 (approved):
Unnecessary scroll bars removed (bottom and right).

PR #3183, opened 2023-03-15
Redirect to resource's form when record set has been deleted

### LINKED ISSUE #2558: Handle links to deleted records better
If the link refers to an existing record, but a non-existent record (i.e, because the record set was deleted), the form shows a 404 error.

http://localhost/specify/view/agent/3/?recordsetid=114

A better behavior would be to redirect the user out of the context of a record set.

CarolineDenis:
Fixed #2558


chanulee1 (approved):
Looks good. Links to records in a deleted record set redirect to the record.

maxxxxxdlp (approved):
Code looks good and works like expected

Issue #888, opened 2021-08-05
Prevent automapper from creating roundabout mappings


maxxxxxdlp:
Don't automap if the mapping path contains base table twice (E.x, if Collection Object is a base table, don't map fields whose mapping path include collection object again)

For example, this mapping for `Gear` is invalid:
<img width="1215" alt="Screen Shot 2021-08-05 at 10 06 09 PM" src="https://user-images.githubusercontent.com/40512816/128408821-941a49cf-b273-4d6b-ba1d-3f163e7bfbc6.png">

There should probably be a few exceptions:
Agent
Taxon
others?

maxxxxxdlp:
Disabled circular mappings in auto mapper for all tables except those that have relationships to themself:
```json
    [
      "[table Agent]",
      "[table Container]",
      "[table Geography]",
      "[table GeographyTreeDefItem]",
      "[table GeologicTimePeriod]",
      "[table GeologicTimePeriodTreeDefItem]",
      "[table LithoStrat]",
      "[table LithoStratTreeDefItem]",
      "[table ReferenceWork]",
      "[table Storage]",
      "[table StorageTreeDefItem]",
      "[table Taxon]",
      "[table TaxonTreeDefItem]",
    ]
```

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/specify-7-8-11-release-announcement/1174/1

PR #3214, opened 2023-03-17
Fix nested -to-manys in automapper suggestions

### LINKED ISSUE #3213: Fix automapper suggestions not working for nested -to-manys
1. Open WorkBench mapper in a data set that has a field whose name exactly matches the catalog number field label
1. Choose Collection Object as the base table
2. Map the field to `Accession`
3. Open the 2nd box, and see the suggested mappings.

This is how it looks with the bug:

image

This is how it should look:

Screenshot 2023-03-17 at 15 07 27

maxpatiiuk:
Fixes #3213
The particular mapping on the screenshot in the issue description won't be shown once #888 is fixed, but for now it's a good way to test this issue

Besides testing the fix, please do general testing of auto mapper suggestions
Focus on testing that the suggestions appear correctly when they are displayed, and there are no errors in the console, rather on whether a particular header should have resulted in a suggestion being shown
Include -to-many relationships and tree ranks in testing


chanulee1 (changes requested):
Is this related to this fix? Cannot use automapper suggestion to create mapping. Not present in `edge`.

https://user-images.githubusercontent.com/103299204/226427617-ae017b0f-386e-411a-8ba6-689e75896718.mp4

chanulee1 (changes requested):
"Not mapped" field within to-many mapping.

https://user-images.githubusercontent.com/103299204/230212162-69d61db5-0a74-4e80-8ad2-986c07d4b826.mp4

maxxxxxdlp:
I also had this happen (don't remember which branch though):

Screenshot 2023-04-13 at 22 01 46

grantfitzsimmons:
Let's revisit this soon

emenslin:
I'm not running into any of the issues and everything looks the same between the issue, edge, and xml editor. Was this fixed somehow?

CarolineDenis:
@specify/ux-testing 
can someone else test this and if same conclusion as emeslin let's close this

pashiav:
Came to the same conclusion as @emenslin

CarolineDenis:
@maxpatiiuk , can this be closed?

maxpatiiuk:
> can this be closed?

TLDR: most bugs this PR intended to fix were fixed by https://github.com/specify/specify7/pull/3215. https://github.com/specify/specify7/pull/3214#pullrequestreview-1349138656 got fixed by some other change not related to either of these.
So yeah, lets close

---

I am able to replicate https://github.com/specify/specify7/pull/3214#pullrequestreview-1373710385 and https://github.com/specify/specify7/pull/3214#pullrequestreview-1373710385, but only if I revert https://github.com/specify/specify7/pull/3215, as otherwise that mapping path is not given

Not able to reproduce https://github.com/specify/specify7/pull/3214#pullrequestreview-1349138656

hmm, I am unable to reproduce original bugs
but I am sure there was an issue back then - likely a change in a different PR fixed something that made this PR no longer needed

---

for future context, description of what this PR does:
when automapper suggestion is generated, it might be generated not on the first box in the row, but on the 2nd or 3rd or so on - thus some parts of the mappingPath are already given, and automapper only needs to guess the rest.

when displaying the mapping path preview, `getMappingLineData` is used (this function turns mapping path into objects that are then consuped by CustomSelectElement to render the UI), but it's result is sliced (`.slice(baseMappingPath.length - pathOffset)`) to not redundantly output boxes for path segments user already has selected

this PR modifies it so that `getMappingLineData` is called not with the complete path, but only with the path starting at what is the table in a current path position (i.e if base table is Collection Object, and we have a pending path `Accession -> `, then the table at current position is Accession)
this is supposively to work around the restriction of circular relationships not being allowed??? not sure though why did I make it handle that case correctly, given that https://github.com/specify/specify7/pull/3215 would make that case not happen

PR #3215, opened 2023-03-17
Forbid circular mappings in auto mapper

### LINKED ISSUE #888: Prevent automapper from creating roundabout mappings

maxxxxxdlp:
Fixes #888


chanulee1:
What is the best way to test this?

maxxxxxdlp:
1. Open the workbench mapper for a data set that contains a "Cat #" as a header
2. Choose Collection Object as base table
3. Manually change the mapping for "Cat #" column to "Accession"
4. Open the 2nd box in that line and make sure mapping back to Collection Object is not proposed

chanulee1 (approved):
Automapper does not suggest a circular relationship on the collection object table through the accession table.

Issue #2713, opened 2022-12-27
Duplicate entry '875-4458' for key 'AgentIDX' when merging agents


maxxxxxdlp:
<img width="1063" alt="Screenshot 2022-12-27 at 08 26 48" src="https://user-images.githubusercontent.com/40512816/209680500-0ce8e41f-1d59-4f25-b0c9-10fd6398c3c4.png">

[Specify 7 Crash Report - 2022-12-27T14_26_26.175Z.txt](https://github.com/specify/specify7/files/10308684/Specify.7.Crash.Report.-.2022-12-27T14_26_26.175Z.txt)

Agent 4458:

```json
{
  "id": 4458,
  "abbreviation": null,
  "agenttype": 1,
  "date1": null,
  "date1precision": null,
  "date2": null,
  "date2precision": null,
  "dateofbirth": null,
  "dateofbirthprecision": null,
  "dateofdeath": null,
  "dateofdeathprecision": null,
  "datetype": null,
  "email": null,
  "firstname": "Kyle",
  "guid": "2b9551ea-2f83-440a-ae74-c70a7a7838ab",
  "initials": null,
  "integer1": null,
  "integer2": null,
  "interests": null,
  "jobtitle": null,
  "lastname": "Orr",
  "middleinitial": "D",
  "remarks": null,
  "suffix": null,
  "text1": null,
  "text2": null,
  "text3": null,
  "text4": null,
  "text5": null,
  "timestampcreated": "2022-11-30T14:33:41",
  "timestampmodified": "2022-12-27T14:25:13.011361",
  "title": null,
  "url": null,
  "verbatimdate1": null,
  "verbatimdate2": null,
  "version": 1,
  "collcontentcontact": null,
  "colltechcontact": null,
  "createdbyagent": "/api/specify/agent/1514/",
  "division": "/api/specify/division/2/",
  "instcontentcontact": null,
  "insttechcontact": null,
  "modifiedbyagent": "/api/specify/agent/1514/",
  "organization": null,
  "specifyuser": null,
  "addresses": [],
  "orgmembers": "/api/specify/agent/?organization=4458",
  "agentattachments": [],
  "agentgeographies": [],
  "identifiers": [],
  "agentspecialties": [],
  "variants": [],
  "collectors": "/api/specify/collector/?agent=4458",
  "groups": [],
  "members": "/api/specify/groupperson/?member=4458",
  "resource_uri": "/api/specify/agent/4458/"
}
```

[...]

PR #3220, opened 2023-03-20
Record merging fixes and generic record recursion

acwhite211:
Adds generic record merging, recursion record merging, and dependent field deletion.  Solves [ISSUE-2713](https://github.com/specify/specify7/issues/2713)


maxxxxxdlp:
Code looks good

Before we can test whether this fixes the bug (https://github.com/specify/specify7/pull/2579#issuecomment-1399051961), I would have to merge these changes onto `agent-merging`. I will try to get to it today or tomorrow (there are merge conflicts)

When that is done, I will approve this and assign to Chanu for testing

maxxxxxdlp (approved):
Code looks good

Before we can test whether this fixes the bug (https://github.com/specify/specify7/pull/2579#issuecomment-1399051961), I would have to merge these changes onto `agent-merging`. I will try to get to it today or tomorrow (there are merge conflicts)

When that is done, I will assign this PR to Chanu for testing

maxxxxxdlp:
@chanulee1 can you please test this?
The changes on this branch are now included on the `agent-merging` branch so use that branch in the test panel when testing the changes.

In particular, test https://github.com/specify/specify7/pull/2579#issuecomment-1399051961 (I already tested and appears to work correctly, but you should try it too) and test https://github.com/specify/specify7/issues/2713

specifyweb/specify/views.py
```diff
@@ -399,86 +400,156 @@ class ReplaceRecordPT(PermissionTarget):
     update = PermissionTargetAction()
     delete = PermissionTargetAction()
 
+@transaction.atomic
+def record_merge_fx(model_name: str, old_model_id: int, new_model_id: int) -> http.HttpResponse:
+    """Replaces all the foreign keys referencing the old record ID
+    with the new record ID, and deletes the old record record.
```
melton-jason, line 406: There seems to be an extra `record` at the end of the function documentation (docstring)

specifyweb/specify/views.py
```diff
@@ -399,86 +400,156 @@ class ReplaceRecordPT(PermissionTarget):
...
+    """
+    # Confirm the target model table exists
+    target_model = apps.get_model('specify', model_name.lower())
```
melton-jason, line 409: This approach, while fine, is not consistent with getting models on the backend. 
The most common approach is to import `models` from `specifyweb.specify` and use `getattr(models, MODEL_NAME)`

Here are a few examples: https://github.com/specify/specify7/blob/fc70113acb58b91513af8d2a8025f742699cb36c/specifyweb/businessrules/uniqueness_rules.py#L2-L8
https://github.com/specify/specify7/blob/fc70113acb58b91513af8d2a8025f742699cb36c/specifyweb/accounts/views.py#L32
https://github.com/specify/specify7/blob/fc70113acb58b91513af8d2a8025f742699cb36c/specifyweb/workbench/tasks.py#L60

specifyweb/specify/views.py
```diff
@@ -399,86 +400,156 @@ class ReplaceRecordPT(PermissionTarget):
...
+    if target_model is None:
+        return http.HttpResponseNotFound("model_name: " + model_name + "does not exist.")
+
+    # Check to make sure both the old and new agent IDs exist in the table
```
melton-jason, line 413: This comment should probably be something like
`# Check to make sure both the old and new model IDs exist in the table`

specifyweb/specify/views.py
melton-jason: It looks like you already have this imported. In that case, it would be `getattr(spmodels, model_name)`

On a side note, how does the frontend send the model_name? If it sends the model name in camel case, you would probably need to do `model_name.lower().capitalize()`

specifyweb/specify/views.py
```diff
@@ -399,86 +400,156 @@ class ReplaceRecordPT(PermissionTarget):
...
+            if not hasattr(foreign_model, field_name_id):
+                continue
+
+            # Filter the objects in the foreign model that references the old target model
+            foreign_objects = foreign_model.objects.filter(**{field_name_id: old_model_id})
+
+            # Update and save the foreign model objects with the new_model_id
+            for obj in foreign_objects:
+                # If it is a dependent field, delete the object instead of updating it.
+                # This is done inorder to avoid duplicates
```
melton-jason, line 455: `inorder` should be `in order`

specifyweb/specify/views.py
```diff
@@ -399,86 +400,156 @@ class ReplaceRecordPT(PermissionTarget):
...
+    dependant_table_names = [rel.relatedModelName
+        for rel in target_object.specify_model.relationships
+        if api.is_dependent_field(target_object, rel.name)]
+
+    # Get all of the columns in all of the tables of specify the are foreign keys referencing AgentID
+    foreign_key_cols = []
+    for table in spmodels.datamodel.tables:
+        for relationship in table.relationships:
+            if relationship.relatedModelName.lower() == model_name.lower():
+                foreign_key_cols.append((table.name, relationship.name))
```
melton-jason, line 430: Is this the best way of doing this? I am not found of a for-loop in a for-loop (that's a lot of iterations!),  but I can not immediately think of a better solution, and if it works--- it works.  
Perhaps this should be looked into further, but I will put some more thought into it nonetheless.

specifyweb/specify/views.py
melton-jason: There are other cases in the comments/documentation that still reference Agent and should be changed. These occur on lines:
[425](https://github.com/specify/specify7/pull/3220/files#diff-13cfe16950398450bc9f1bf00605843b3ebbb22efca216859825059571711e1bR425)
[432](https://github.com/specify/specify7/pull/3220/files#diff-13cfe16950398450bc9f1bf00605843b3ebbb22efca216859825059571711e1bR432)
[500](https://github.com/specify/specify7/pull/3220/files#diff-13cfe16950398450bc9f1bf00605843b3ebbb22efca216859825059571711e1bR500)

melton-jason (changes requested):
Code looks good. 
There are some optional changes that could be done to fix some typos in code comments and maintain consistency with other areas of the backend. 

Feel free to mark conversations as resolved as you see fit. Most of the changes are just typos in the comments and would be trivial to fix

specifyweb/specify/views.py
acwhite211: ya, this could maybe have been written in a functional manner, but it would be doing essentially the same thing.

chanulee1 (changes requested):
Able to merge agents with respective addresses without address duplication on `agent-merging`. However, I cannot merge agents with respective member records. Is this within the scope of this PR? Also, for one of the agents, the member relationship is not displayed in the merging dialog (Test1).


https://user-images.githubusercontent.com/103299204/226975501-c6ea3eab-831d-47eb-859c-60cce08766b4.mp4

[Specify 7 Crash Report - 2023-03-22T16_09_47.228Z.txt](https://github.com/specify/specify7/files/11042507/Specify.7.Crash.Report.-.2023-03-22T16_09_47.228Z.txt)

Agent "Test1":  https://coldfish-agent-merging.test.specifysystems.org/specify/view/agent/20789/

maxxxxxdlp:
Looks like the changes Jason suggested have been made
For the issue Chanu reported I created a separate bug #3308 (as it is not specific to the changes made in this pr) - will work on fixing that next

PR #3223, opened 2023-03-20
Increase contrast for selection option in WorkBench mapper

### LINKED ISSUE #3212: Insufficient contrast for selection option in WorkBench mapper
1. Use Dark mode
2. Open WorkBench mapper
3. Select any required field
4. See that the contrast is insufficient (hard to read the text - need to use a different color for the text)

Screenshot 2023-02-11 at 20 55 04

CarolineDenis:
Fixes #3212


maxxxxxdlp (approved):
To test:
In both light mode and dark mode:
- Sufficient contrast between text and background
  - for required, optional, hidden and autocomplete suggestion option
     - Even when hovering or if the option is selected

Issue #3161, opened 2023-03-13
Create new specify user page missing "name" field


chanulee1:
`7.8.6`:
![image](https://user-images.githubusercontent.com/103299204/224777330-3890e7ee-8066-4a54-903f-c86283e51447.png)

`edge`:
![image](https://user-images.githubusercontent.com/103299204/224777466-dcf12ce6-73da-4d2c-9c8a-45924a45af87.png)

Cannot create new users because of this.

maxxxxxdlp:
I'm able to reproduce #3164, but not this one just yet
@chanulee1  Does this issue happen in any database or a specific database?

PR #3235, opened 2023-03-21
Fix dialogs not having a scroll bar

### LINKED ISSUE #3228: Dialogs are no longer scrollable
`v7.8.8` on Firefox

https://user-images.githubusercontent.com/37256050/226624974-e4845929-60cb-4626-9732-42c3b67fa053.mp4

See `v7.8.8` on the left, `v7.8.7.1` on the right

**Reported By:** Amanda at New Brunswick

CarolineDenis:
Fixes #3228

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
@@ -64,7 +64,7 @@ export const DataEntry = {
   >(
     'DataEntry.Grid',
     'div',
-    `overflow-x-auto items-center p-1 -ml-1 gap-2`,
+    `items-center p-1 -ml-1 gap-2`,
     ({
       viewDefinition,
       display,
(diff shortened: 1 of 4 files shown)


grantfitzsimmons (approved):

maxxxxxdlp (changes requested):
The fix is not correct as it makes dialog header/footer not be sticky anymore:

https://user-images.githubusercontent.com/40512816/226780331-0d3fc899-da93-4e8f-be3e-755f60f9ac53.mov

@CarolineDenis 
Instead, since Grant mentioned exactly when the bug started appearing, you can run this command to see all the changes between those versions:

```bash
git log -p v7.8.8 --not v7.8.7.1
```

You will see that `flex-shrink-0` was added to the form, removing which fixes this issue. 

From `git blame` you can see that it was added in e137789ef to fix a bug (https://github.com/specify/specify7/issues/3161).

Instead, of adding `flex-shrink-0` to fix the bug, looks like removing `overflow-x-auto` here fixes https://github.com/specify/specify7/issues/3161: 

https://github.com/specify/specify7/blob/48ee0c007aff33efec3a3a93c854b67ec5b4edae/specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx#L67

`overflow-x-auto` was added during the initial form rewrite (ef116336b) thus it might no longer be needed - needs testing to see the impact of it

Thus:
- Undo the change you made here
- Remove flex-shrink-0
- Remove overflow-x-auto
  - Test the impact of removing overflow-x-auto
- Read more about `git blame` and `git log` with `--not` flag to learn how to learn how to hunt down what change could have caused the breakage and why a change was made

maxxxxxdlp (approved):

PR #3239, opened 2023-03-21
Add rounded corners for menu items

### LINKED ISSUE #3224: Make navigation menu items rounded when selected
image

Add rounding to each item

CarolineDenis:
Fixes #3224

This PR was made to have a better visualization of idea. 
Screenshot 2023-03-21 at 3 59 16 PM

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Header/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/index.tsx
@@ -117,7 +117,7 @@ export function Header({
           <img
             alt=""
             className={`
-                hover:animate-hue-rotate
+                rounded-md hover:animate-hue-rotate
                 ${isCollapsed ? 'hidden' : ''}
               `}
             src="/static/img/logo.svg"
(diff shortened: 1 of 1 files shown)


maxxxxxdlp:
For comparison, could you try `rounded` instead of `rounded-md` to have the same border-radius as in all the other places (for consistency)? If it looks worse, than it's fine to keep it as it. If difference is not noticeable, then making it consistent is good

CarolineDenis:
Rounded instead of rounded-md: 
Screenshot 2023-03-23 at 12 12 19 PM

maxxxxxdlp:
can barely notice the difference between them (0.375rem vs 0.5rem) but I will sleep better knowing they are consistent 😂 💤

realVinayak:
Looks fine to me. @grantfitzsimmons

CarolineDenis:
> image
> 
> _See the left side is rounded, right is not_
> 
> We cannot round the logo as it is inconsistent with our brand identity unless we wish to entirely change the look.
> 
> We just need to remove the rounding on the logo in particular (even though it looks a little better when collapsed)

What about rounding the logo when collapsed to make sense with the menu items but no rounded when expanded to keep the logo as is

grantfitzsimmons:
> What about rounding the logo when collapsed to make sense with the menu items but no rounded when expanded to keep the logo as is

I am still against the rounding in general, but I think we need to keep the logo not rounded at all times.

We can discuss rounding the logo as a team, but consistency with our marketing materials/website is important! :smile:

Issue #2813, opened 2023-01-13
Front-end has no direct access to sp7-exclusive tables


grantfitzsimmons:
<img width="1565" alt="image" src="https://user-images.githubusercontent.com/37256050/212392190-b12a082f-b944-4fbe-a3dc-50734e6a0d9a.png">

grantfitzsimmons:
https://sp7demofish.specifycloud.org/specify/query/new/spdataset/

maxxxxxdlp:
I don't know why, but Ben made some tables not be available through the main specify API (tables for notifications, workbench, permissions and all other specify7-exclusive tables).

Thus, all of those tables are not available to the front-end, except though dedicated APIs. We can look into changing that in the future.

maxxxxxdlp:
Though, the query builder should not show a blank page - it should show a 404 page.
You could open a separate bug for that

grantfitzsimmons:
It does show a 404 page, so that's not a problem :smile:

maxxxxxdlp:
oh I see. it's not visible on the screenshot.
Yeah, then it's just a back-end bug - something that we can consider changing

melton-jason:
> > > > Add a block for
> > > > 
> > > > * `spdataset`
> > > 
> > > 
> > > For more context, this is regarding adding a deletion blocker to SpecifyUser. This operation is not currently supported with how I extended deletion blockers, but I will add a way to support adding reverse relationships so this can be added.
> > 
> > 
> > `Spatasets` will more complicated than any other table(s) to work with on the backend. Many of Specify's internal helper functions rely on the datamodel being loaded when Specify starts up (this happens in [load_datamodel.py](https://github.com/specify/specify7/blob/production/specifyweb/specify/load_datamodel.py)). The issue is that to get the list of tables and fields in the database, Specify 7 reads the `specify_datamodel.xml` in the Specify 6 config directory (as seen below), and as far as I know the spdataset table is not included within that file.
> > https://github.com/specify/specify7/blob/4877f75686668cdd5b2d28c46881fb7628ce69a4/specifyweb/specify/load_datamodel.py#L222-L231
> > 
> > We could manually use SQL to get the correct information, but that will only be a short-term solution (which may be fine for now). Preferably, we would want to engineer a way to load/register these Specify 7 specific tables in the internal specify datamodel. (If such a way already exists and I am overlooking it, then feel free to point it out. I have checked all attributes on specify.models and spdataset does not exist)
> 
[...]

PR #3240, opened 2023-03-22
Improve CollectionRelationship uploads in the WorkBench

### LINKED ISSUE #3089: Uploading Collection Object Rels not working in the WorkBench
> A big bug I may have found is to do with using the workbench to add collection relationships - before we went live last year I tested this and all I had to do was map the left and right catalog numbers:
> image.png
>  
> image
> 
> However now this doesn't work at all - if I do this from our specimen collection its is creating the right-side catalog number as a new CO in the specimen collection rather than just creating a link between. Has this behaviour changed in the recent upgrades?
> 
> This is a bit of a serious issue because Corinna is travelling to Hobart to get the fish collection voyage data uploaded into the production system

**Reported By:** Dan at CSIRO

melton-jason:
Fixes #3089 

---

This pull request includes:
- Autogenerated default forms for CollectionRelationship and CollectionRelType 
  - I believe Specify 6 contains a view definition for CollectionRelType. Which of course does not work in Specify 7. So primarily ensure that the default form for CollectionRelationship is OK. 

- Apply the proper schema overrides for CollectionRelationship and CollectionRelType. 
  - CollectionRelType was made requried for CollectionRelationship
  - name was made required for CollectionRelType 

- Makes `CollectionRelType.name` a frontend-only picklist with `name` as its field
  - CollectionRelType.name is how the collections are supposed to be determined during the upload
  - Because this is such an important field, this change will ensure that only existing names are used

- Adds the ability to override which collection a column will upload into
  - To accomplish this, go into the uploadplan (A video is attached detailing this process)
  - Find the correct relationship/table/column you want to override
  - As a key/value pair for an `uploadTable`, add `"overrideScope" : {"collection" : COLLECTION_ID}`, where COLLECTION_ID is your desired collection to upload into
  
  - NOTE: This override should work for _any_ table being uploaded through the WorkBench, not just CollectionRelationships

---

### To View Upload Plan

https://user-images.githubusercontent.com/64045831/226804158-2318935a-a77e-4e62-999f-86ca74e658e8.mp4

_* a video showing how to view the [DEV] Show Plan button_

---

### Collection Override Example

uploadPlan

_* an example of a collection override being applied to the `rightside` of a CollectionRelationship, with the additions marked by an orange box_


melton-jason:
The only remaining thing to accomplish with this Pull Request is to have the backend automatically override the `rightside` collection when uploading a CollectionRelationship. 

The intended behavior can also currently be accomplished by adding a manual collection override for the `rightside` field in the WorkBench upload plan 
(see my steps outlined in the intial PR comment under the **_Adds the ability to override which collection a column will upload into_** section for instructions to test this).

maxxxxxdlp (approved):
Looks great!

So if I understood correctly, front-end doesn't need to fill out overrideScope, as back-end will do it automatically for CollectionRelatinships, but if it would be necessary, someone can manually edit the upload plan?

Also, if I enter overrideScope, then edit upload plan in the mapper and save, the overrideScope won't be preserved? Not an issue, but something to be aware of

melton-jason:
> So if I understood correctly, front-end doesn't need to fill out overrideScope, as back-end will do it automatically for CollectionRelatinships, but if it would be necessary, someone can manually edit the upload plan?

Yes, to stay as compatible as possible with past workbenches, I made overrideScope optional in the upload plan. We will need to it automatically on the backend for things like CollectionRelationships. 

And yes, if needed someone can manually edit the upload plan and tell the backend which collection to validate and upload into. With this change, some use cases might be uploading Collection Objects in multiple collections with one WorkBench dataset, or upload/validate PickLists for a separate collection, etc... 
I tried to make it as easy to scale as possible, so in the future we can provide customizability for every scope. 
If this is a popular idea, we could provide some UI in the WorkBench that would just add the overrideScope to the upload plan.
 
> Also, if I enter overrideScope, then edit upload plan in the mapper and save, the overrideScope won't be preserved? Not an issue, but something to be aware of

The upload plan is stored in the database. Editing and then saving the upload plan saves it to the database, so overrrideScope will be preserved.

maxxxxxdlp:
I like how you implemented it, but I don't think there would be that much need for this beyond CollectionRelationships - most of the time, if users want to upload into a different collection, they would just switch to that collection

specifyweb/workbench/upload/upload_table.py
```diff
@@ -55,6 +56,28 @@ def to_json(self) -> Dict:
 
     def unparse(self) -> Dict:
         return { 'baseTableName': self.name, 'uploadable': self.to_json() }
+    
+class DeferredScopeUploadTable(NamedTuple):
```
maxxxxxdlp, line 60: could make this class extend `UploadTable` rather than `NamedTuple` since most of the stuff can be inherited, right?

specifyweb/workbench/upload/upload_table.py
```diff
@@ -55,6 +56,28 @@ def to_json(self) -> Dict:
...
+    toOne: Dict[str, Uploadable]
+    toMany: Dict[str, List[ToManyRecord]]
+
+    related_key: str
+    relationship_name: str
+    filter_field: str
+
+    def apply_scoping(self, collection, deffer: bool = True) -> "ScopedUploadTable":
+        if not deffer:
+            from .scoping import apply_scoping_to_uploadtable as apply_scoping
```
maxxxxxdlp, line 74: once inherited from UploadTable, could just call `super().apply_scoping(...)` in there

maxxxxxdlp, line 74: can you tell me what's the difference between doing imports inline vs at the top of the file?
is there a performance impact?
is it just for convenience?
is it a best practice?

maxxxxxdlp (changes requested):
The code for deferring scope looks tricky. Sorry can't comprehend it right now as I don't know well what's going on here.

If there is a way to simplify it then do it, otherwise document what the DeferedUploadPlan concept means and how it interacts with UploadPlan. Also, let's keep up the tradition of workbench have very good test coverage and add test cases for the newly added code - that will also help other explain what's the expected behavior

melton-jason:
This is ready to be tested. I will fix any failing tests and add comments/documentation either tonight or tomorrow, but all of the logic is in place.

When testing, please make sure that right-side collection objects get uploaded/matched correctly in the rightSideCollection of the collection rel type. (Also, test to ensure that it validates and uploads with the correct collection's Catalog Number formatting).

Please also do brief testing of common workbench uploads/functions, as well as for recordsets.

specifyweb/workbench/upload/upload_table.py
melton-jason: That is the very first thing I tried. Unfortunately, it seemed to break any new attributes for the named tuple: You could not create an instance of the object with any newly declared attributes as keywords in the constructor.

I just have a basic understanding of Named Tuples to begin with, so I am sure there is a way to do it. 
I will do some further investigation.

grantfitzsimmons:
I am so impressed just reading your summary and code. Excited to try this and so happy to see this!!!

melton-jason:
I was considering my changes and realized there is a bug right now with uploading CollectionRelationships with more than one CollectionRelTypes that have differing `rightSideCollections`. 

Take this WorkBench dataset for example:
example_colrel_dataset

If the `HostFish` CollectionRelType has a separate `rightSideCollection` than `TestRel`, it will be ignored and upload CollectionObject `111` into `TestRel`'s rightSideCollection. 

I am working on a fix, but expect more complicated WorkBench code!

melton-jason:
> I was considering my changes and realized there is a bug right now with uploading CollectionRelationships with more than one CollectionRelTypes that have differing `rightSideCollections`. 
> 
> Take this WorkBench dataset for example:
> example_colrel_dataset
> 
> If the `HostFish` CollectionRelType has a separate `rightSideCollection` than `TestRel`, it will be ignored and upload that row's rightSide CollectionObject `111` into `TestRel`'s rightSideCollection. 
> 
> I am working on a fix, but expect more complicated WorkBench code!
> 
> You can continue to test this as normal, I am simply pointing out I am aware of this bug to prevent redundancy

This should now be fixed.

Assuming no more bugs present themselves, all that is remaining is to add documentation in the code and write tests.

specifyweb/workbench/upload/upload_table.py
melton-jason: These imports are inline to prevent a circular import dependency. 

Because `scoping.py` has top_level/global imports from `upload_table.py`, if these imports were included at the top of `upload_table.py`, whenever scoping.py or upload.py would be 'loaded', it would go through its import statements and try and load the other python file--- which would require it to load the original file, etc...

grantfitzsimmons:
> I'm getting this error when I try to save this mapping. image

Can you request changes on the PR?

specifyweb/workbench/upload/upload_table.py
melton-jason: It seems that Named Tuples are not built/meant to be inherited from (see https://github.com/python/typing/issues/427 and this answer to a StackOverflow question https://stackoverflow.com/a/50369898).

Perhaps the easiest implementation is to create a new child class of NamedTuple whenever we need and manually add every function/attribute we need. 

However, this is not the 'cleanest' solution, and makes implementing future needed NamedTuples prone to bugs and 'harder' than it should be. 

There are some workarounds mentioned online, each with its own pros and cons. 
We could use dataclasses (described in the aforementioned StackOverflow question), the `@property` decorator (https://stackoverflow.com/a/55989641), or override each subclasses call to `__new__` (https://stackoverflow.com/a/42386174).

However, refactoring that bit of WorkBench code may require a decent amount of code changes and testing, so it is probably not in the scope of this pull request.

specifyweb/workbench/upload/upload_table.py
melton-jason: This is more tricky than it seems. 
See my breakdown in https://github.com/specify/specify7/pull/3240#discussion_r1149507690

specifyweb/workbench/upload/upload_table.py
maxxxxxdlp: Giving that this is a limitation of Mypy/python typing I have no problems with current solution given that workarounds are only adding more complexity

Though perhaps you could leave a brief comment in the code linking to relevant information so that if this get's fixed by python in the future, the code can be refactored accordingly

specifyweb/workbench/upload/tests/testuploading.py
```diff
@@ -328,11 +328,13 @@ def test_attachmentimageattribute(self) -> None:
         plan = UploadTable(
             name='Attachment',
             wbcols={'guid': parse_column_options('guid')},
+            overrideScope=None,
```
maxxxxxdlp, line 331: Not very familiar with Python, but is there support for default values so that you don't have to add this for all the tests?

maxxxxxdlp (approved):

maxxxxxdlp:
> NOTE: This override should work for any table being uploaded through the WorkBench, not just CollectionRelationships

Does it check for permissions when uploading into a different collection?

melton-jason:
> Saving mapping throws an error - image

Good catch!
This should be fixed now. I created the mapping I was testing on before some significant changes (which broke creating a new mapping) and was just testing validation/upload/unupload on the same, old mapping.

specifyweb/workbench/upload/tests/testuploading.py
melton-jason: Fixed in 48d6367

realVinayak:
Am I misinterpreting the discourse post?

melton-jason:
Here are some testing instructions regarding the expected behavior:

- Find/make a new Collection Relationship Type (Collectionreltype) 
  - Note its name and right side collection

_(This is a query for Collectionreltype on `coldfish`)_
collection_rel_type

- Create a new WorkBench Dataset with the Collection Relationship base table
- At least have the following mapping
- In the Data Mapper, ensure  `CollectionRelType -> name` is required  

collection_rel_workbench

- Consider the following sample data in the WorkBench

hot_collectionrel

- (Make sure that CollectionRelType displays as a Combo Box with all of the names for CollectionRelType)
- `124` in column `Cat # (2)` is the Right Side of the Collection Relationship. It should be validated/uploaded against other collection objects in the CollectionRelType's Right-Side Collection
  - In this case the Right-Side Collection is the KU Fish Tissue Collection 
- Test the following cases:
  - Make sure the catalog number for the Right-Side CollectionObject is validated using the Right-Collection's CollectionObject catalognumber format
  - Try and validate/upload a Collection Object whose catalognumber already exists in the Right-Side Collection
  - Try and validate/upload a Collection Object whose catalognumber does not exist in the Right-Side Collection

For convenience, I have saved a WorkBench on `coldfish` called `collection relationships` which has the correct mappings

Additionally, try and test all the above with uploading a HostTaxon

melton-jason:
> Based on https://discourse.specifysoftware.org/t/create-a-collection-relationship-in-specify-7/581, the right side collection object needs to have just one left side collection object. So, if I'm doing an import something like this: **Left** | **Right** 0 | 0 1 | 0 This shouldn't be considered a correct relationship, and workbench shouldn't validate this. But, currently, it does. Look at record set work bench created after I uploaded it https://coldfish-issue-3089.test.specifysystems.org/specify/view/collectionrelationship/7129/?recordsetid=149. Here is a screenshot of the dataset I uploaded: image. Forgot to include screenshot of the mapping in the initial comment: image

Yes, you are correct. It does seem **by that Discourse post** that there _shouldn't_ be the same Right-Side CollectionObject used in CollectionRelationships. 
However, doing this is currently supported even when manually adding Collection Relationships on the forms. 

You can even view all of the Left-Side Collection Objects from the Right-Side collection objects if you use a OneToManyPlugin. 

For testing purposes, I changed how the plugin is displayed in the KU Fish Tissue Colelction from
```
<!-- <cell type="field" id="colrel" uitype="plugin" name="this" initialize="name=ColRelTypePlugin;relname=Has Tissue;"/> -->
```
to
```
<cell type="field" id="colrel" uitype="plugin" name="this" initialize="name=CollectionRelOneToManyPlugin;relname=Has Tissue;" colspan="9"/>
```

I see no reason to restrict this arbitrarily, as some people may eventually find it useful. But I do not know much about how this relationship is used in practice, so we can add the restriction if it is sensible.

realVinayak:
When I met with Theresa, she insisted that the restriction be enforced.
Maybe you can convince her?

melton-jason:
> Host Taxon relationship uploads not working. In `edge`, this would result in a prompt to disambiguate the taxon record.
> 
> https://purherb-issue-3089.test.specifysystems.org/specify/workbench/3/

Thanks for catching this

melton-jason:
> When I met with Theresa, she insisted that the restriction be enforced. Maybe you can convince her?
@realVinayak 

Yes, I wrote my message before that decision was made. I will still enforce the restriction.

melton-jason:
> Host Taxon relationship uploads not working. In `edge`, this would result in a prompt to disambiguate the taxon record.
> 
> https://purherb-issue-3089.test.specifysystems.org/specify/workbench/3/

@chanulee1 
This should be fixed now. Can you test this please?

chanulee1:
Error is fixed! However, similar to on `edge`, the disambiguation options show taxon records from the "Botany" discipline, not from the "Host" discipline.

image

image

image

https://purherb-issue-3089.test.specifysystems.org/specify/workbench/4/

melton-jason:
@realVinayak 
Can you test the restriction? rightside in CollectionRelationship should now be unique to Collectionreltype

realVinayak:
The restriction works now! Another bug: currently, host taxon node is created in the logged in collection tree rather than the host taxon tree. For purdueherbaria11023-issue-3089.test.specifysystems.org, they have left side collection Arthur Fungarium and the right side as Host. The discipline on the left side is Botany and the right side is Host. So, if I'm using workbench on Arthur Fungarium (Botany) to make host taxons (https://purdueherbaria11023-issue-3089.test.specifysystems.org/specify/workbench/4/), it should make new taxons in the Host tree (not Botany tree). Right now, it is creating them in Botany tree. Here is workbench data I uploaded 
image
Here is mapping I used 
image
SQL query to check discipline after upload: 
image
. Querying using Query Builder confirms this.

realVinayak:
Can't believe I didn't catch this before. Can we have `Collection Rel Type` be required, or put a default value in? It throws this error otherwise - it should just highlight it red. 
image

realVinayak:
Also have it highlight the empty catalog number (or any other required cells) after validation if they are empty. I was trying to validate it without putting in a catalog number.
image

melton-jason:
> The validation isn't looking for catalog numbers in the correct collection. For example, consider `collectionreltypeid 1`, image where `left side collection id is 327680` and the `right side collection id is 294912`. The maximum catalog number in `collection 327680` is `000036839`. I'm currently logged in `collection 294912`. In the screenshot below, I chose `Plant-Fly` as collection rel type - left side of `Plant-Fly` isn't my current collection. image I've put `030002233` as left catalog number which is a valid catalog number in my current collection (but not in the left side collection). It doesn't show the blue box saying a new collection object will be created (because it should be in the collection `327680`). It looks like it is assuming that the left side collection of a relationship is the current logged in collection. After uploading this row, it takes the collectionobjectid from the current logged in collection and refers to the collection object in the wrong collection.

@realVinayak 
Yes, this was 'intentional'. I always assumed the left-side collection would be the one that is currently logged in. In theory, this allows greater flexibility because _any_ collection can be the left-side of the CollectionRelationship. 
This is an easy change (unlike adding support for Host Taxon) if we want to be strict about the left-side collection and _only_ validate and upload into the left side collection of the Collection Rel Type

melton-jason:
> > The validation isn't looking for catalog numbers in the correct collection. For example, consider `collectionreltypeid 1`, image where `left side collection id is 327680` and the `right side collection id is 294912`. The maximum catalog number in `collection 327680` is `000036839`. I'm currently logged in `collection 294912`. In the screenshot below, I chose `Plant-Fly` as collection rel type - left side of `Plant-Fly` isn't my current collection. image I've put `030002233` as left catalog number which is a valid catalog number in my current collection (but not in the left side collection). It doesn't show the blue box saying a new collection object will be created (because it should be in the collection `327680`). It looks like it is assuming that the left side collection of a relationship is the current logged in collection. After uploading this row, it takes the collectionobjectid from the current logged in collection and refers to the collection object in the wrong collection.
> 
> @realVinayak Yes, this was 'intentional'. I always assumed the left-side collection would be the one that is currently logged in. In theory, this allows greater flexibility because _any_ collection can be the left-side of the CollectionRelationship. This is an easy change (unlike adding support for Host Taxon) if we want to be strict about the left-side collection and _only_ validate and upload into the left side collection of the Collection Rel Type

@realVinayak 
Fixed. Now it should always validate/upload into the leftside collection as specified in the collectionreltype

melton-jason:
Furthermore, should the fix for allowing uploading HostTaxon even be included in this pull request? 

It will be tricky to implement and here is why:
HostTaxon is a many-to-one relationship from the CollectingEventAttribute table to the Taxon table. 
A HostTaxon inherently does not know anything about CollectionRelationships: it is not a CollectionRelationship and CollectionRelationship is not included anywhere in the key constraints.

Specify only knows which discipline to associate the Host Taxon to on the forms because the CollectionRelType name is defined _in the form definition_. 

This presents a dilemma for the WorkBench. 
The WorkBench would need to access the form definition for Collecting Event Attribute (this would _have_ to be the form definition in the logged-in collection) to get the name of the CollectionRelType while evaluating the row. From the name, it can run a query and get the left-side/right-side collections. From the right-side collection it can get the discipline and its associated Taxon Tree, which it can evaluate the HostTaxon against. 

Essentially, I am unsure how long this implementation would take, and HostTaxon has nothing to do with CollectionRelationships, which is what this PR sought out to fix (#3089).

Because of these reasons, I feel it would be better if the fix for HostTaxon uploads are included in its own pull request. I do not want to continually add features/bug fixes to a pull request (especially ones that seem like they are out of the scope of the PR). The fix this PR sought out to do would never be merged!

Of course, I am open to thoughts and discussions on this topic: this is just my two cents

realVinayak:
Not necessary. Can you disable inserting host taxon for now? Don't want
people putting in wrong nodes in trees

grantfitzsimmons:
> Furthermore, should the fix for allowing uploading HostTaxon even be included in this pull request?

As we discussed more in person, no.

In summary, for all of the reasons @melton-jason described above, it is simpler to disable it at this time without something in the database indicating what relationship the HostTaxon is using.

melton-jason:
I've (finally) resolved all of the failing tests. 
Can this briefly be tested again? If everything is still good, then this can be merged! 🥳

grantfitzsimmons:
@chanulee1 Can you test this today?

melton-jason:
> Can't believe I didn't catch this before. Can we have `Collection Rel Type` be required, or put a default value in? It throws this error otherwise - it should just highlight it red. image

> Also have it highlight the empty catalog number (or any other required cells) after validation if they are empty. I was trying to validate it without putting in a catalog number. image

Collection Rel Type and the Left/Ride side Collection Objects _are_ required, but only in the frontend (as schema overrides). Thus they are not actually required on a database level. 

The cell highlighting is done by checking if the SpLocaleContainerItem associated with each column of the upload is required in the database. 
https://github.com/specify/specify7/blob/e96288c5e8cd530b31ceb841fbc6326731418822/specifyweb/workbench/upload/parsing.py#L60-L71

Preferably, we run a Django migration which properly marks the correct fields as required in the database. However, to be thorough, we would additionally need to modify the Specify6 datamodel xml (The django fields are built from the xml). We may even 

Alternatively, we can make schema overrides for the backend. However, I do not like this solution because it fragments the schema even further. There would be front-end overrides, back-end overrides, and then what the value is in the database.

melton-jason:
I have talked with @grantfitzsimmons and we have reached a conclusion about how to approach this. 
The solution will be painful regardless of what we do. 

If we solve the root of the problem, alongside a django migration, we would need to modify Specify's 6 `specify_datamodel.xml`, as Specify 7 relies on it to build the datamodel. However, the `specify_datamodel.xml` is autogenerated by Specify 6 when it is compiled, so this would required a Specify 6 update. 
(Preferably we move away from relying on the outdated xml. Having a Specify 7-only representation of the datamodel would also fix #2813)

Alternatively, adding schema overrides to the backend would needlessly make debugging in the future more difficult, as the schema would be split into 3 pieces (the frontend representation, backend, and database) and cost more development time in the future. The solution would be 'hacky', and not long term. 

For now, unless any further ideas are brought forward, it may be wise to leave this as is for now and provide documentation explaining the behavior. A proper solution-regardless of the way we go-will warrant further discussion.

grantfitzsimmons:
@realVinayak and @chanulee1 – can you test this when you get a chance?

chanulee1 (approved):
Looks good. Right side collection object validated using right collection's catalog number format. Right side collection object is created when catalog number does not exist in right side collection. Collection Relationship uploads seem to be working well.

realVinayak (approved):

realVinayak:
See https://github.com/specify/specify7/pull/4929 for an alternate implementation. The main difference is that I consider scoping completely different than binding. Instead of scoping upload plan "constantly" (which then required defer logic per-row), I scope it for every row (so defer isn't needed -- just scope). Simplifies things this way. 

For performance, observe that upload plan only needs to be rescoped if there are collection relationships (or any other "defer" fields). So, it checks for them when doing scoping. If not found, it is safe to cache it. Hence, the little caching mechanism in that implementation. That is, the upload plan gets scoped just once (for the first row, the next rows reuse it as long as no defer fields are found).

Issue #2999, opened 2023-02-15
Implement an attachment picker


maxxxxxdlp:
Would be useful in user preferences when user wants to use some attachment as welcome page image (currently they have to put the URL)

Can also be used in the future to allow people to set a profile picture.

The attachment picker should still allow putting an arbitrary URL, while also allowing searching/filtering of attachments and ability to upload a new attachment

realVinayak:
I don't think 'Pick an attachment' is a good title for this dialog as it also includes other image selection options.
![image](https://user-images.githubusercontent.com/61122018/228354032-645be722-617f-4c57-9f2e-e9f0a92ab831.png)

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/specify-7-8-11-release-announcement/1174/1

PR #3241, opened 2023-03-22
Add ability to customize buttons'color in preferences

### LINKED ISSUE #2091: Ability to set button colors
Being able to set the color of buttons specifically would be very advantageous. 

This could be for all buttons, but even better, being able to give them color in the form definition.

CarolineDenis:
Fixes #2091


maxxxxxdlp (approved):
👍

grantfitzsimmons:
image

I don't consider the save button to be a "Warning Button"– maybe "action" button or "save" button is more accurate?

image

grantfitzsimmons (changes requested):
Nomenclature changes may be necessary. We have a lot of button names, and little indication of what they are

maxxxxxdlp (approved):
Looks good

Optionally consider separating the button colors into a separate section in user preferences (as the current section would get too big if we ever get more colors in the future). It is especially important to make this decision before release, as renaming category/moving items into a different category would revert changes for people who customized the colors - thus we try to think about best placement of items, and do bare minimum changes once released

A better long term solution would be to implement a versioning and migration mechaims for preferences, but that would add complexity and was not needed so far

specifyweb/frontend/js_src/lib/components/UserPreferences/ApplyPreferences.tsx
```diff
@@ -100,28 +100,28 @@ export function SetCssVariables(): null {
       saveButtonColor: getUserPref('general', 'appearance', 'saveButtonColor'),
       secondaryButtonColor: getUserPref(
         'general',
-        'appearance',
+        'buttonColor',
```
maxxxxxdlp, line 103: consider renaming this to `button` to match the UI label and to be more general in case we add non-color button customization options in the future (and like you mentioned, to remove redundancy with the item name - as each name includes "color" in it)

specifyweb/frontend/js_src/lib/localization/preferences.ts
```diff
@@ -296,6 +296,9 @@ export const preferencesText = createDictionary({
     'fr-fr': 'Apparence',
     'uk-ua': 'Зовнішній вигляд',
   },
+  buttonColor: {
```
maxxxxxdlp, line 299: Change the key from `buttonColor` to `buttons` - try to have the language key match the localized English string. Exceptions:
- Localized string is long/has special characters (in this case key should be summarizing it)
- The string is a description for another string/used as a tooltip text. I.e, `searchAlgorithm` and `searchAlgorithmDescription` - this helps translators see the relationship between the button/heading and the accompanying tooltip

More details: https://github.com/specify/specify7/tree/production/specifyweb/frontend/js_src/lib/localization#front-end-localization

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

grantfitzsimmons (changes requested):
image

The current color changed for categories in preferences:
image

I don't consider the major action buttons to be "warning buttons":


https://user-images.githubusercontent.com/37256050/234306812-6e35e760-f3d5-49b6-9107-3531d1455ea6.mov

I don't consider the "Copy to Clipboard" to be a "success" button either:

https://user-images.githubusercontent.com/37256050/234307044-045ad080-e853-4ee3-944f-e74a450688f5.mov

I also don't consider the trash can in a query an "info button":

https://user-images.githubusercontent.com/37256050/234307617-67eaa034-df51-42cc-ab0f-2614860440a7.mov

specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
```diff
@@ -241,7 +241,7 @@ export function QueryLine({
           aria-label={commonText.remove()}
           className="print:hidden"
           title={commonText.remove()}
-          variant={className.redButton}
+          variant={className.infoButton}
```
maxxxxxdlp, line 244: why was this changed?

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Atoms/__tests__/__snapshots__/Button.test.tsx.snap
```diff
@@ -73,6 +50,17 @@ exports[`Button.Icon renders without errors 1`] = `
 </DocumentFragment>
 `;
 
+exports[`Button.Info renders without errors 1`] = `
+<DocumentFragment>
+  <button
+    class="button rounded cursor-pointer active:brightness-80 px-4 py-2
+    disabled:bg-gray-200 disabled:text-gray-500 dark:disabled:!bg-neutral-700 gap-2
+    inline-flex items-center capitalize justify-center shadow-sm dialog-icon-info hover:brightness-150 bg-[color:var(--info-button-color)] dark:bg-[color:var(--info-button-color-dark)] text-white"
```
maxxxxxdlp, line 58: instead of `bg-[color:var(--info-button-color)] dark:bg-[color:var(--info-button-color-dark)]` here, modify ApplyPreferences.tsx to get the value of the `--info-button-color` CSS variable from either dark more or light mode user pref depending on the result of `useDarkMode()`

This way, if you need to use `--info-button-color` in multiple places, you don't have to provide separate dark mode styles in every place.

Do the same for all the other colors you added. If you need an example, take a look at how this was done for other colors in ApplyPreferences.tsx

specifyweb/frontend/js_src/lib/components/SchemaConfig/Components.tsx
```diff
@@ -33,7 +34,7 @@ export function SchemaConfigHeader({
         {schemaText.changeBaseTable()}
       </Button.Small>
       <span className="-ml-2 flex-1" />
-      <Button.Small onClick={handleSave}>{commonText.save()}</Button.Small>
+      <Submit.Save onClick={handleSave}>{commonText.save()}</Submit.Save>
```
maxxxxxdlp, line 37: while this is a good change you did, keep in mind that submit buttons should not be used outside of `<Form>`. Please add `<Form>` element around the entire schema config page

specifyweb/frontend/js_src/lib/components/UserPreferences/index.tsx
```diff
@@ -129,7 +138,7 @@ export function usePrefDefinitions() {
             ] as const
         )
         .filter(([_name, { subCategories }]) => subCategories.length > 0),
-    []
+    [preferencesVisibilityContext]
```
maxxxxxdlp, line 141: Careful here. This created a bug or a performance problem.
`preferencesVisibilityContext` is re-created on every render.
Thus, useMemo is re-run on every render.
If you had ESLint enabled, it would have reported this as a bug - please spend some time looking into fixing your IDE's ESLint issues as ESLint would be crucial for catching bugs once I'm not here to review every line of code

specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
```diff
@@ -385,12 +386,12 @@ export function Mapper(props: {
             {props.isReadOnly ? wbText.dataEditor() : commonText.cancel()}
           </Link.Small>
           {!props.isReadOnly && (
-            <Button.Small
+            <Submit.Save
```
maxxxxxdlp, line 389: Please review **ALL THE PLACES** where you made a change to use a .Save button to check for the following:
1. If you are using Submit where it used Button before, you need to wrap the entire page in a `<Form>`. Alternatively, use `Button.Save` instead.
2. Make sure that after your change did not affect the button size (looks like you went from Button.Small to regular sized button in here)

specifyweb/frontend/js_src/lib/localization/preferences.ts
```diff
@@ -296,6 +296,12 @@ export const preferencesText = createDictionary({
     'fr-fr': 'Apparence',
     'uk-ua': 'Зовнішній вигляд',
   },
+  buttonLight: {
```
maxxxxxdlp, line 299: be consistent. if localization key says "button", than language string should say "Button". or  the opposite, if you want language string to say "Buttons", then use "buttons" in the localization key.

This way, when looking at the localization string in the code, you can easily guess what it would print

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/SchemaConfig/Components.tsx
CarolineDenis: Do you mean around Container.Full line 109 in SchemaConfig/index.tsx?

specifyweb/frontend/js_src/.eslintrc.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
```
maxxxxxdlp, line 1: While this file is a good example of the default eslint config, we already have an eslint config  so you don't have to create a new one. Take a look at `eslint.config.js`

specifyweb/frontend/js_src/package.json
```diff
@@ -88,6 +87,8 @@
     "@types/react-modal": "^3.13.1",
     "@types/react-router-dom": "^5.3.3",
     "@types/underscore": "^1.10.24",
+    "@typescript-eslint/eslint-plugin": "^5.59.1",
```
maxxxxxdlp, line 90: please undo all changes to this file (with exception of removing `eslint-plugin-react` dependency - it's not needed here)
1. first, this is not applicable to this branch
2. secondly, you don't need to manually add dependencies for eslint plugins in here as that's already done though `@maxxxxxdlp/eslint-config` and `@maxxxxxdlp/eslint-config-react` dependencies

specifyweb/frontend/js_src/lib/components/Atoms/__tests__/__snapshots__/Button.test.tsx.snap
maxxxxxdlp: example for you from ApplyPreferences.tsx:
```
  const background = darkMode ? prefs.darkBackground : prefs.background;
  React.useEffect(
    () => document.body.style.setProperty('--background', background),
    [background]
  );
```

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/WorkBench/Template.tsx
```diff
@@ -169,13 +169,13 @@ function WbView({
                 >
                   {wbText.revert()}
                 </Button.Small>
-                <Button.Small
+                <Button.Save
                   aria-haspopup="dialog"
-                  className="wb-save"
+                  className="wb-save !py-1 !px-2"
```
maxxxxxdlp, line 174: please don't hardcode `!py-1 !px-2`.
Those are defined in className.ts and should not be duplicated here.

But even besides that, there is a better solution.
Remember that Button.Small accepts a `variant` attribute?
Take a look at the code to see how you can use Button.Small, but make it look like Button.Save

<details>
  <summary>Solution</summary>
 
  ```
  <Button.Small ... variant={className.saveButton}>...</...>
  ```
</details>

specifyweb/frontend/js_src/lib/components/UserPreferences/ApplyPreferences.tsx
maxxxxxdlp, line 1: These changes look good now. Does everything in this file make sense or should we meet and talk about this file a bit?

Also, I would like you to do a small refactor in this file.
Instead of calling getUserPref for both light mode and dark mode and later using only one of those values based on darkMode variable, how about you call getUserPref only once, for the correct pref right away (for dark more or for light mode)

<details>
  <summary>Partial solution</summary>

  ```tsx
  saveButtonColor: getUserPref('general', darkMode ? 'buttonDark' : 'buttonLight', 'saveButtonColor'),
  ```
</details>

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/UserPreferences/ApplyPreferences.tsx
maxxxxxdlp, line 1: very nice. thanks for refactoring it.
but can you do the refactor not just for the colors that you added but for other colors that were in this file before too?

I.e, disabledFieldBackground and darkDisabledFieldBackground

maxxxxxdlp (approved):

maxxxxxdlp (approved):

CarolineDenis:
@grantfitzsimmons could you test this PR this week please?

grantfitzsimmons (approved):
https://github.com/specify/specify7/assets/37256050/b7c07953-e36b-4349-bfa9-d49c98d91450

@maxpatiiuk Chrome has a major issue where when I use the color picker, it takes a _long_ time before it settles down.

maxxxxxdlp:
> Chrome has a major issue where when I use the color picker, it takes a long time before it settles down.

This is related to the bug with localStorage synchronization between multiple tabs causing an infinite loop

This is fixed on the use-dynamic-ranks branch

specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
```diff
@@ -42,7 +40,42 @@ import {
...
+  defineItem,
+  GenericPreferences,
+  PreferenceItem,
+  PreferencesVisibilyContext,
+} from './types';
+import { StatLayout } from '../Statistics/types';
+import { statsText } from '../../localization/stats';
+
+// Custom Renderer for a preference item
+export type PreferenceItemComponent<VALUE> = (props: {
```
maxxxxxdlp, line 53: This is not merged correctly. PreferenceItemComponent was removed from this file, but added back accidentally when resoling merge conflicts
I will push a fix to production for this

specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
```diff
@@ -378,6 +533,10 @@ export const userPreferenceDefinitions = {
             renderer: WelcomePageModePreferenceItem,
             container: 'div',
           }),
+          /*
```
maxxxxxdlp, line 536: This is not merged correctly. This comment was removed once https://github.com/specify/specify7/issues/2999 was fixed, but got added back during merge conflict resolution

specifyweb/frontend/js_src/lib/components/WorkBench/Template.tsx
```diff
@@ -171,7 +171,7 @@ function WbView({
                 </Button.Small>
                 <Button.Small
                   aria-haspopup="dialog"
-                  className="wb-save"
```
maxxxxxdlp, line 174: The wb-save class name was important here - without it workbench save button won't be clickable anymore (the class name was used for attaching the on click handler)
Careful when removing class names - be sure to look for usages

maxxxxxdlp: @melton-jason thanks for keeping track!

PR #3245, opened 2023-03-23
Allow taxontiles to resize with browser window

### LINKED ISSUE #2235: [testability] Taxon tiles display incorrectly when window resizes
https://user-images.githubusercontent.com/37256050/192672906-c8c81760-c38a-45ce-9a39-aee79a988e28.mov

db: `freshfish`
browser: Chrome on macOS
user: `abentley`

https://freshfish-testability.test.specifysystems.org/specify/

CarolineDenis:
Fixes #2235

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/HomePage/TaxonTiles.tsx
+++ b/specifyweb/frontend/js_src/lib/components/HomePage/TaxonTiles.tsx
@@ -1,4 +1,6 @@
+import { treemap } from 'd3';
 import React from 'react';
+import _ from 'underscore';
 
 import { useAsyncState } from '../../hooks/useAsyncState';
 import { welcomeText } from '../../localization/welcome';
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/HomePage/TaxonTiles.tsx
```diff
@@ -30,17 +31,37 @@ export function TaxonTiles(): JSX.Element {
       treeData === undefined
     )
       return undefined;
+
+    const resizeObserver = new ResizeObserver(_.debounce(render, 50));
+
+    if (container !== null) resizeObserver.observe(container);
```
maxxxxxdlp, line 37: this if is redundant as you already check this on line 29

specifyweb/frontend/js_src/lib/components/HomePage/TaxonTiles.tsx
```diff
@@ -30,17 +31,37 @@ export function TaxonTiles(): JSX.Element {
...
-      .on('mouseover', (_event, node) => setTitle(titleGenerator(node)))
-      .on('click', (_event, node) =>
-        window.open(`/specify/query/fromtree/taxon/${node.data.id}/`, '_blank')
-      );
+
+    let chart: ReturnType<typeof makeTreeMap> | undefined = undefined;
+
+    function render(): void {
+      if (container === null || treeData === undefined) return;
+      void chart?.remove();
```
maxxxxxdlp, line 46: (optional) I don't think this would be an issue, but could you double-check that there is no error if this line is executed twice? just in case

i.,e `void chart?.remove(); void chart?.remove();`

maxxxxxdlp (changes requested):
one small change and one small test

maxxxxxdlp (approved):

grantfitzsimmons (approved):

PR #3246, opened 2023-03-23
Navigate to home page after deletion of a query in query builder

### LINKED ISSUE #2580: [testability] Deleting a saved query from inside of it does not navigate away immediately

CarolineDenis:
Fixes #2580

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Forms/DeleteButton.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/DeleteButton.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { useNavigate } from 'react-router-dom';
 
 import { useAsyncState } from '../../hooks/useAsyncState';
 import { useBooleanState } from '../../hooks/useBooleanState';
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/Forms/DeleteButton.tsx
```diff
@@ -124,6 +127,7 @@ export function DeleteButton<SCHEMA extends AnySchema>({
                      */
                     overwriteReadOnly(resource, 'needsSaved', false);
                     loading(resource.destroy().then(handleDeleted));
+                    navigate('/specify/', { replace: true });
```
maxxxxxdlp, line 130: This is not good.

Note that this change affected all delete buttons across the application, not just the query builder (i.e, in record sets it used to go to previous/next resource, but because of this change it will likely always go to homepage)

Also, note that the delete button calls handleDeleted callback when the resource is deleted. So if you need to add redirect to homepage in this single place, you could use rather than modify this code.

But even better, remember we modified the query list dialog to remove the query from the list when a query was deleted? Take a look at how we implemented that and do the same in query builder - make the query builder listen for current query to get deleted, and when it is, redirect to home page

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

PR #3247, opened 2023-03-23
Allow longer list of collections to be displayed on login page

### LINKED ISSUE #2588: Collection list should have a scrollbar at all times
> If a user has access to multiple collections, then the list might be so large that it exceeds the bounds of the panel, however the scrollbars do not automatically appear to indicate there are more collections to pick from. The user is potentially going to not know additional collections are available if they scroll the list.

**Problem:**

image

**Solution:**
Show a scroll bar at all times when the list exceeds the visible frame
image

**Requested By:** Matthew Cruz at UMich

CarolineDenis:
Fixes #2588

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/ChooseCollection/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/ChooseCollection/index.tsx
@@ -125,7 +125,7 @@ function Wrapped({
         {errors.length > 0 && <ErrorMessage>{errors}</ErrorMessage>}
         {hasAccess ? (
           <>
-            <div className="-ml-1 flex max-h-56 flex-col gap-2 overflow-y-auto pl-1">
+            <div className="-ml-1 flex max-h-[50vh] flex-col gap-2 overflow-y-auto pl-1">
               {sortedCollections.map(({ id, collectionname }) => (
                 <Label.Inline key={id}>
                   <Input.Radio


specifyweb/frontend/js_src/lib/components/ChooseCollection/index.tsx
```diff
@@ -125,7 +125,7 @@ function Wrapped({
         {errors.length > 0 && <ErrorMessage>{errors}</ErrorMessage>}
         {hasAccess ? (
           <>
-            <div className="-ml-1 flex max-h-56 flex-col gap-2 overflow-y-auto pl-1">
+            <div className="-ml-1 flex max-h-[50vh] flex-col gap-2 overflow-y-auto pl-1">
```
maxxxxxdlp, line 128: This would be bad on mobile in the landscape. Can you disable it on less tall screens? Or leave it in rem like it was, but increase the number

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/ChooseCollection/index.tsx
maxxxxxdlp: Edit:
It does look fine actually. I didn't realize how much space was taken by logo, label and button, so collection switcher being just 50% of the screen is fine

maxxxxxdlp (approved):

chanulee1 (approved):
Was only able to view a maximum of about 7 collections before the change, can now view much more depending on window size.

PR #3258, opened 2023-03-27
Implement an attachment picker

### LINKED ISSUE #2999: Implement an attachment picker

CarolineDenis:
Fixes #2999


specifyweb/frontend/js_src/lib/components/Attachments/Gallery.tsx
```diff
@@ -23,12 +23,14 @@ export function AttachmentGallery({
   scale,
   isComplete,
   onChange: handleChange,
+  onClick,
```
maxxxxxdlp, line 26: Replace with `onClick: handleClick` here and in other places

specifyweb/frontend/js_src/lib/components/UserPreferences/AttachmentPicker.tsx
```diff
@@ -0,0 +1,116 @@
...
+      {url !== undefined ? (
+        <img
+          className={`max-h-full max-w-full object-contain`}
+          src={url}
+          style={{
+            width: `10rem`,
+            height: `10rem`,
+          }}
+        />
+      ) : null}
```
maxxxxxdlp: please don't use null 😢

maxxxxxdlp (changes requested):
Few tiny changes, mostly variable naming
Looking good!

maxxxxxdlp (changes requested):
Few tiny changes, mostly variable naming
Looking good!

maxxxxxdlp (approved):

realVinayak:
I don't think 'Pick an attachment' is a good header for this dialog as it also includes other options
image

specifyweb/frontend/js_src/lib/components/Molecules/AttachmentPicker.tsx
```diff
@@ -0,0 +1,122 @@
...
+  isReadOnly,
+}: {
+  readonly url: string | undefined;
+  readonly onChange: (url: string) => void;
+  readonly isReadOnly: boolean;
+}): JSX.Element {
+  const [attachment, setAttachment] = useAttachment(undefined);
+  useErrorContext('attachment', attachment);
+
+  const [isOpen, , , handleToggle] = useBooleanState();
```
realVinayak, line 32: Currently, handleToggle isn't getting called when you click on an attachment. Would be better if dialog also closes when user clicks on an attachment.

CarolineDenis: Instead of choosing an attachment preview than closing dialog?

maxxxxxdlp:
> I don't think 'Pick an attachment' is a good header for this dialog as it also includes other options

@realVinayak 

It can also be though of as picking a file from your system to upload or picking a file from the internet to paste it's URL.

Do you have a suggestion for a better more inclusive heading?

specifyweb/frontend/js_src/lib/components/Molecules/AttachmentPicker.tsx
maxxxxxdlp: Good catch. The `onUploaded` callback for `<UploadAttachment>` and `onClick` callback for `<AttachmentsView>` should both call `handleClose` to close the dialog afterward

realVinayak:
Pick an image?

maxxxxxdlp (approved):

chanulee1 (changes requested):
Looks good. Able to set welcome page image using url, uploading an image, or from attachments. Found one UI bug:

Dialog resizes while scale is being dragged:

https://user-images.githubusercontent.com/103299204/234698360-a78ac1c5-d2b7-43ea-8fb8-0fac3810951a.mp4

specifyweb/frontend/js_src/lib/components/Attachments/index.tsx
```diff
@@ -72,7 +79,13 @@ export function AttachmentsView(): JSX.Element | null {
   );
 }
 
-function Attachments(): JSX.Element {
+function Attachments({
+  onClick,
+  displayScale = true,
```
maxxxxxdlp, line 84: Instead of `displayScale`, could you call this `isEmbedded` or `isInDialog`?

specifyweb/frontend/js_src/lib/components/Attachments/index.tsx
```diff
@@ -223,16 +236,18 @@ function Attachments(): JSX.Element {
...
-          {attachmentsText.scale()}
-          <Input.Generic
-            max={maxScale}
-            min={minScale}
-            type="range"
-            value={scale}
-            onValueChange={(value) => setScale(Number.parseInt(value))}
-          />
-        </Label.Inline>
+        {displayScale && (
```
maxxxxxdlp, line 239: For the future, when you do a change like this (i.e, removing some feature or any other change the reasons for which are not obvious without context) please include a brief explanation for why this was done.
Do that either in the commit message, or leave a comment in the code

As a quick hack, if the fix is in response to a particular github issue/command, instead of providing a full explanation of the issue you could just say `Fixes <link to comment or #issueNumber>`

specifyweb/frontend/js_src/lib/components/Attachments/index.tsx
maxxxxxdlp: or actually, you don't even need to provide a prop. Instead you could just check if React.useContext(DialogContext) is not undefined - that will tell you that you are in a dialog

specifyweb/frontend/js_src/lib/components/Molecules/AttachmentPicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+        >
+          {commonText.delete()}
+        </Button.Gray>
+      ) : undefined}
+
+      {url !== undefined && (
+        <img
+          className="max-h-full max-w-full object-contain"
+          src={url}
+          style={{
```
maxxxxxdlp, line 79: use tailwind class names for width and height

specifyweb/frontend/js_src/lib/components/Molecules/AttachmentPicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+            height: `10rem`,
+          }}
+        />
+      )}
+
+      {isOpen && (
+        <Dialog
+          buttons={commonText.close()}
+          header={preferencesText.pickAttachment()}
+          onClose={(): void => {
```
maxxxxxdlp, line 90: don't need `{` `}` `;`

specifyweb/frontend/js_src/lib/localization/common.ts
```diff
@@ -616,4 +616,10 @@ export const commonText = createDictionary({
     'ru-ru': 'Времени осталось',
     'uk-ua': 'Час, що залишився',
   },
+  change: {
+    'en-us': 'Change',
+  },
+  pickImage: {
```
maxxxxxdlp, line 622: can you move `pickImage` into `preferencesText` localization dictionary. it doesn't really belong in `commonText`

`change` is fine in here as it's something that could theoretically be used anywhere

specifyweb/frontend/js_src/lib/components/Molecules/AttachmentPicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+        <Button.Gray onClick={() => handleToggle()}>
+          {url === undefined ? commonText.pickImage() : commonText.change()}
+        </Button.Gray>
+      )}
+
+      {url !== undefined && !isReadOnly ? (
+        <Button.Gray
+          onClick={() => {
+            setAttachment(undefined);
+            handleChange('default');
```
maxxxxxdlp, line 68: sorry that I overlooked this earlier.
instead of `'default'` it should use `undefined` here - this component should not now anything about `'default'` as that's something specif to the welcome page pref

specifyweb/frontend/js_src/lib/components/UserPreferences/Renderers.tsx
```diff
@@ -237,17 +231,17 @@ export const WelcomePageModePreferenceItem: PreferenceItemComponent<WelcomePageM
...
+            displayScale={false}
             isReadOnly={isReadOnly}
-            item="source"
-            subcategory="general"
-            value={source}
-            onChange={setSource}
+            url={url === 'default' ? undefined : url}
+            onChange={(url): void => {
+              setUrl(url);
+              if (url === 'default') handleChange('default');
```
maxxxxxdlp, line 241: after you modify `handleChange('default')` to `handleChange(undefined)` in `<AttachmentPicker>` don't forget to change this `if` condition here

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

chanulee1 (approved):
Looks good!

PR #3260, opened 2023-03-28
Fix scroll bar when exporting to TSV 

### LINKED ISSUE #2733: `statspage`: Scrollbar appears when exporting to TSV
https://user-images.githubusercontent.com/37256050/210431469-a27ec23d-70a2-4608-8735-00be052f8f60.mov

CarolineDenis:
Fixes  #2733

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Molecules/FilePicker.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/FilePicker.tsx
@@ -140,6 +140,7 @@ export const downloadFile = async (
 ): Promise<void> =>
   new Promise((resolve) => {
     const iframe = document.createElement('iframe');
+    iframe.classList.add('absolute');
     iframe.addEventListener('load', () => {
       if (iframe.contentWindow === null) return;
       const element = iframe.contentWindow.document.createElement('a');
--- a/specifyweb/frontend/js_src/lib/components/Molecules/FilePicker.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/FilePicker.tsx
@@ -140,7 +140,7 @@ export const downloadFile = async (
 ): Promise<void> =>
   new Promise((resolve) => {
     const iframe = document.createElement('iframe');
-    iframe.classList.add('absolute');
+    iframe.classList.add('absolute', 'hidden');
     iframe.addEventListener('load', () => {
       if (iframe.contentWindow === null) return;
       const element = iframe.contentWindow.document.createElement('a');


grantfitzsimmons (changes requested):
https://user-images.githubusercontent.com/37256050/229659718-223d4e01-f754-4b5f-b637-009eb7da8ffb.mov

I still see the issue

https://herbrbge33123-issue-2733.test.specifysystems.org/specify/stats

Firefox, macOS

CarolineDenis:
@grantfitzsimmons can you test this to see if it makes any difference?

specifyweb/frontend/js_src/lib/components/Molecules/FilePicker.tsx
```diff
@@ -140,6 +140,7 @@ export const downloadFile = async (
 ): Promise<void> =>
   new Promise((resolve) => {
     const iframe = document.createElement('iframe');
+    iframe.classList.add('absolute', 'hidden');
```
maxxxxxdlp, line 143: I was hesitant to add `'hidden'` as some browsers might not trigger file downloads from hidden iframes.
Just adding `absolute` might be sufficient to fix this issue.

maxxxxxdlp: `'hidden'` only affects visibility, not the placement. Presence or absence of scroll bar should not be affected by it.

Doesn't the scroll bar disappear as soon as you add `'absolute'`? If not, you might need to add `top-0` or something like that

CarolineDenis:
@maxxxxxdlp 
With absolute + hidden: 

https://user-images.githubusercontent.com/108160931/230107942-38d887b4-a1fc-4f30-b138-a0a095545454.mov

with absolute: 
Screenshot 2023-03-30 at 2 36 26 PM

specifyweb/frontend/js_src/lib/components/Molecules/FilePicker.tsx
maxxxxxdlp: the 2nd video in your comment doesn't load for me, but I trust you

Interesting that there is not a single mention of scroll bars on the wiki page for visibility:hidden - https://developer.mozilla.org/en-US/docs/Web/CSS/visibility

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/Molecules/FilePicker.tsx
maxxxxxdlp: keep the hidden. as otherwise there is no point in these changes if they don't fix the scroll bar

grantfitzsimmons (approved):

Issue #2619, opened 2022-12-13
Test nested dialogs for accessibility compliance


maxxxxxdlp:
1. Open a modal dialog
2. From inside the modal dialog, click on a button that opens another dialog

Test cases where the 2nd dialog is modal and where 2nd dialog is nonmodal.

CarolineDenis:
Tested all modal dialogs and nested modal/non-modal dialogs. 
=> no issue found 

@maxxxxxdlp we might close this one

maxxxxxdlp:
Can confirm there are no issues (at least none that are detected by automated tools - there could be some issues that are detectable by using a screen reader only)
There are however some issues detected with forms. @CarolineDenis feel free to look into them if interested (https://github.com/specify/specify7/issues/966). I.e, the <QueryComboBox> does not properly set id attribute, thus the field is not linked to an associated label

PR #3264, opened 2023-03-28
Add [inert] attribute to #root when in modal dialog

### LINKED ISSUE #2618: Add `[intert]` attr to `#root` when in modal
New HTML attribute has been added for modal dialogs:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert

It should be added to the `#root` element when in modal dialog.

Besides the improvements for screen readers, it would also reduce the need for focus management and (in some browsers), make `Cmd+F` ignore the content behind the dialog

CarolineDenis:
Fixes #2618

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Core/Contexts.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Core/Contexts.tsx
@@ -123,6 +123,7 @@ export function Contexts({
                   header={commonText.loading()}
                   isOpen={isLoading}
                   onClose={undefined}
+                  modal={false}
                 >
                   {loadingBar}
                 </Dialog>
--- a/specifyweb/frontend/js_src/lib/components/Molecules/Dialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/Dialog.tsx
@@ -167,6 +167,18 @@ export function Dialog({
 }): JSX.Element {
   const id = useId('modal');
 
+  const root = document.getElementById('root');
+  React.useEffect(() => {
+    if (modal === true) {
+      root?.setAttribute('inert', 'true');
+    } else {
+      root?.removeAttribute('inert');
+    }
+    return () => {
+      root?.removeAttribute('inert');
+    };
+  }, [modal]);
+
   const [modifyTitle] = usePref('general', 'dialog', 'updatePageTitle');
   useTitle(modal && isOpen && modifyTitle ? header : undefined);
 


specifyweb/frontend/js_src/lib/components/Molecules/Dialog.tsx
```diff
@@ -167,6 +167,18 @@ export function Dialog({
 }): JSX.Element {
   const id = useId('modal');
 
+  const root = document.getElementById('root');
+  React.useEffect(() => {
+    if (modal) {
```
maxxxxxdlp, line 172: can be replaced with
```
React.useEffect(()=>{
  root?.toggleAttribute('inert',modal);
  return ()=>root?.removeAttribute('inert');
},[modal]);`

File: specifyweb/frontend/js_src/lib/components/Molecules/Dialog.tsx [hunk 2]
```diff
@@ -167,6 +167,18 @@ export function Dialog({
 }): JSX.Element {
   const id = useId('modal');
 
+  const root = document.getElementById('root');
```
maxxxxxdlp, line 170: By having this line in here, getElementById is executed on every render of every dialog - that would be wasteful in terms of performance - put this line outside the component

specifyweb/frontend/js_src/lib/components/Core/Contexts.tsx
```diff
@@ -122,6 +122,7 @@ export function Contexts({
                   className={{ container: dialogClassNames.narrowContainer }}
                   header={commonText.loading()}
                   isOpen={isLoading}
+                  modal={false}
```
maxxxxxdlp, line 125: Why was this change mad?

maxxxxxdlp (changes requested):
As part of this change, the developer should also test the implications of the change (whether Ctrl+F searches content behind the dialog when root is inert, and how does accessibility work, and how does this work when you have two dialogs)

maxxxxxdlp (approved):
The code looks good, but need to test the implications ([see my previous comment](https://github.com/specify/specify7/pull/3264#pullrequestreview-1362020587)) and like you mentioned test https://github.com/specify/specify7/issues/2619.

Before we do those things though, let's wait and see how maintainers of the dialog library we use respond to this feature request - https://github.com/reactjs/react-modal/issues/1011

(I subscribed on that issue so will be notified of updates. Until then, marking this as a draft to prevent accidental merging)

carlosmbe:
How can I test this from a UX Point of View?

CarolineDenis:
> How can I test this from a UX Point of View?

@carlosmbe you can compare the behavior between this branch and normal production.
You can open a dialog and use the tab and see how it acts and also use ctrl + f. You can test those behavior as well when 2 dialogs are open. 

Thank you!

carlosmbe (approved):

PR #3272, opened 2023-03-29
Allow to enter data into any search field after dragging

### LINKED ISSUE #3269: After moving an element in a query, you cannot enter data into any search field
>There seems to be an issue with the query builder where it will prevent you being able to enter data into any of the search fields and the only way to be able to work with it again is to refresh the page, which means starting the query process again from scratch. One way to trigger it seems to be to ‘move’ an element of the query by clicking and holding accidentally. 

**Reported by:** Robyn at RBGE

CarolineDenis:
Fixes #3269

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
@@ -81,6 +81,15 @@ export function QueryFields({
       distance: 4,
     });
 
+    sortable.on('sortable:start', (event) => {
+      const target = event.dragEvent.originalEvent.target as HTMLElement;
+      if (
+        ['A', 'BUTTON', 'INPUT', 'SELECT', 'TEXTAREA'].includes(target.tagName)
+      )
+        event.cancel();
+      console.log(event);
+    });
+
     sortable.on('sortable:stop', (event) => {
       const newIndex = event.newIndex;
       const oldIndex = event.oldIndex;
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
@@ -81,18 +81,21 @@ export function QueryFields({
       distance: 4,
     });
 
+    const interactiveElements = [
+      'button',
+      'a',
+      'input',
+      'select',
+      'textarea',
+      '.custom-select',
+    ];
+
     sortable.on('sortable:start', (event) => {
       const target = event.dragEvent.originalEvent.target as HTMLElement;
       if (
-        target.closest('button') !== null ||
-        target.closest('a') !== null ||
-        target.closest('input') !== null ||
-        target.closest('select') !== null ||
-        target.closest('textarea') !== null ||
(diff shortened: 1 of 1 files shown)


chanulee1 (changes requested):
Able to drag query line when hovering over a button.


https://user-images.githubusercontent.com/103299204/228647424-0536f0e8-a208-424a-ad90-ca5facd1f3ef.mp4

https://coldfish-issue-3269.test.specifysystems.org/specify/query/128/

maxxxxxdlp:
Drag and drop when hovering over list boxes was allowed because it didn't cause bugs (in our testing)
Drag and drop for buttons should have been disabled - @CarolineDenis take a look at why it's not fixed

realVinayak:
I can still reproduce it 
image

CarolineDenis:
> I can still reproduce it
> image

I think this was fixed in the new commit

maxxxxxdlp:
> I can still reproduce it

@realVinayak you should reject the commit in such cases. That then gives owners of the PR option to request re-review which would send you a notification

specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
```diff
@@ -81,6 +81,20 @@ export function QueryFields({
       distance: 4,
     });
 
+    sortable.on('sortable:start', (event) => {
+      const target = event.dragEvent.originalEvent.target as HTMLElement;
+      if (
+        target.closest('button') !== null ||
```
maxxxxxdlp, line 87: Change to loop over an array

specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
```diff
@@ -81,6 +81,20 @@ export function QueryFields({
...
+        target.closest('a') !== null ||
+        target.closest('input') !== null ||
+        target.closest('select') !== null ||
+        target.closest('textarea') !== null ||
+        target.closest('article') !== null
```
maxxxxxdlp, line 92: 'article' can be used for several things, not just by combo box. if you want to match combo box, replace `article` with `.custom-select`

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

realVinayak:
Not sure if this should be fixed in the PR or not but when using drag query fields in a dialog, the fields are behind the dialog. 
Screenshot (66)

realVinayak:
Also, sometimes, when using drag in the dialog, the field lines sometimes jump outside of the dialog. Like this:
image
I'm aware that is the `edge`, but that is irrelevant because you are just canceling  the events in the PR, so this will still be reproducible. I haven't found a conclusive way to reproduce this. @CarolineDenis can you go through the code and see why this might happen? Also, this remains on the screen even if go to a different page 
image
image
image

realVinayak:
Extensively tested this, works for me. @chanulee1 can you do another round and see if you catch anything weird related to this PR?

chanulee1 (approved):
Did not find anything weird. Cannot drag query lines by buttons or inputs.

PR #3273, opened 2023-03-29
Fix form and field sizing issues

### LINKED ISSUE #3255: User name field is missing
image

https://coldfish-edge.test.specifysystems.org/specify/security/user/new/

`edge`

@CarolineDenis first discovered on `statspage`, also verified on the demo instance

maxxxxxdlp:
Fixes #3255

Please do testing in the security panel on different screens

Also, please test that scroll bars appear correctly when form is rendered in a dialog

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
@@ -259,7 +259,7 @@ export const Select = wrap<
 >(
   'Select',
   'select',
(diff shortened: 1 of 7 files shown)


chanulee1 (approved):
Looks good. User name field is present in security panel (tested on multiple different screens). Did not find any issues with scroll bars.

realVinayak:
Can confirm that user name is always visible.

maxxxxxdlp:
For the future:
This PR removes min-width from fields as a hotfix
After more  testing is done, it should be added back (but to a smaller value - ie 5rem, not 10rem). Should also consider container queries

PR #3280, opened 2023-03-31
Disable user roles if collection access disabled

### LINKED ISSUE #1994: Collection access should be required when a role is selected
image

> Then I tried again and saw that there is an ‘enable collection access’ button, which you have to check before selecting ‘full access user’.
> 
> Perhaps the design of this interface just needs to be improved a bit.

**Reported By:** Willem

CarolineDenis:
Fixes #1994


maxxxxxdlp (approved):

grantfitzsimmons (approved):
Looking good!

PR #3296, opened 2023-04-04
Add navigation backwards and forwards among records

### LINKED ISSUE #511: Support navigation backwards and forwards among records added by with "Save and Add Another".

CarolineDenis:
Fixes #511


specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
```diff
@@ -45,6 +45,7 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
   onAdd: handleAdd,
   onClone: handleClone,
   onDelete: handleDelete,
+  createNewRecordSet,
```
maxxxxxdlp, line 48: This is not used so must be removed

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
```diff
@@ -183,6 +186,10 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
         recordSetTable: schema.models.RecordSet.label,
       })
     : commonText.delete();
+
+  // Determine if in scope of new data entry and set title header
+  const [recordsetid] = useSearchParameter('recordsetid');
```
maxxxxxdlp, line 191: Don't do this. It's a bad assumption that the query string argument would always be available when this component is called and that it is not used by any other component for a different reason.

I.e, imagine a feature where you can preview a record set in a dialog - your code won't work in that case.

Instead of getting record set id from a query string, it should be a prop. Or even better, find a way to not add new props - the more props there are the less maintainable the code and the more things can break

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
```diff
@@ -197,37 +204,42 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
                   !isDependent && dialog !== false ? resource : undefined
                 }
               />
+              {recordsetid === undefined ? undefined : (
```
maxxxxxdlp, line 207: Instead of adding redundant checks, exploit the checks that are already in place. I.e, the `Add` button is not rendered if `handleAdding` is undefined. So why not just pass `handleAdding={undefined}` instead of adding new checks?

maxxxxxdlp: Also, this is a bug. This assumes that the component is only used for record sets and removes the buttons if it's not in a record set. This is wrong because the component is also used for displaying query results and will be used in more places in the future.

When making changes, please also review all usages of the current component to see what is affected or else you risk introducing bugs

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -249,7 +256,12 @@ function RecordSet<SCHEMA extends AnySchema>({
   // Fetch ID of record at current index
   const currentRecordId = ids[currentIndex];
   React.useEffect(() => {
-    if (currentRecordId === undefined) handleFetch(currentIndex);
+    if (currentRecordId === undefined) {
+      if (recordSet.isNew()) {
+        setIds([currentRecord.id]);
```
maxxxxxdlp, line 261: redundant `{` and `}`. There is no need to add them when you have just a single statement.

Given how easy it is to confuse `}` and add extra ones or remove some accidentally, keeping the code clean and avoiding `{` whenever possible is advisable

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -283,24 +297,52 @@ function RecordSet<SCHEMA extends AnySchema>({
       })
     );
 
+  async function createNewRecordSet(
+    ids: RA<number | undefined>
+  ): Promise<void> {
+    await recordSet.save();
+    await Promise.all(
+      ids.map(async (resource) =>
```
maxxxxxdlp, line 305: the `resource` variable is misleading. Rename to `recordId` or just `id`

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -283,24 +297,52 @@ function RecordSet<SCHEMA extends AnySchema>({
...
+      )
+    );
+    navigate(`/specify/record-set/${recordSet.id}/`);
+  }
+
   return (
     <>
       <RecordSelectorFromIds<SCHEMA>
         {...rest}
+        createNewRecordSet={createNewRecordSet}
```
maxxxxxdlp, line 319: This is not used so should be removed

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -283,24 +297,52 @@ function RecordSet<SCHEMA extends AnySchema>({
...
         defaultIndex={currentIndex}
         dialog={dialog}
-        headerButtons={<EditRecordSetButton recordSet={recordSet} />}
+        headerButtons={
+          recordSet.isNew() && ids.length > 0 && !currentRecord.isNew() ? (
```
maxxxxxdlp, line 323: Is it possible to have the`<Slider />` and the create record set button appear only after the total count is more than one (for new record sets only)?

This way the slider won't be visible if you are creating just one resource (so it will look like just like it was before this pull request) but as soon as the user clicks on `Add`, the slider will re-appear

maxxxxxdlp (changes requested):
I know that we talked about how this branch has some bugs that I should look into resolving. I will take a look at fixing them after the things I commented about get resolved

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
maxxxxxdlp, line 1: You probably already know about this but putting it here just as a reminder:

Need to remove this button for new record sets:
Screenshot 2023-04-05 at 13 11 58

maxxxxxdlp: And this button should be positioned better (ideally, just right to the `<Slider>`)

Screenshot 2023-04-05 at 13 12 43

maxxxxxdlp: Also, the meta menu (the gear icon that is normally in the top right corner) has disappeared

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
maxxxxxdlp: and replace the giant `New Record Set` button with something more sublet - it's currently larger and more prominent than the "Save" button!

we try to not have any text in the form header besides the heading - so instead of buttons with text, use icons

see my earlier slack message:
>though, that button can't be there. it draws too much attention (heck, it's bigger than the "save" button)
propose replacing it with a small icon (similar to the + and - buttons that are there) except that it has to be distinct from the normal + button that appears there
> how about one of these?
> image
> (https://heroicons.dev/?search=plus)

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -315,14 +315,15 @@ function RecordSet<SCHEMA extends AnySchema>({
...
-        createNewRecordSet={createNewRecordSet}
         defaultIndex={currentIndex}
         dialog={dialog}
         headerButtons={
-          recordSet.isNew() && ids.length > 0 && !currentRecord.isNew() ? (
-            <Button.Orange onClick={() => loading(createNewRecordSet(ids))}>
-              {commonText.newRecordSet()}
-            </Button.Orange>
+          recordSet.isNew() ? (
+            ids.length > 0 && !currentRecord.isNew() ? (
```
maxxxxxdlp, line 322: don't display this button until there is more than 1 record

maxxxxxdlp:
I pushed several fixes for bugs
@CarolineDenis please review the changes, do more testing, and fix the remaining issues

maxxxxxdlp (approved):
To test:
- Check if the "Form Meta" menu still appears
- Test with browser-back arrow
- Test creating a record set from entered records

chanulee1 (approved):
Looks good. Navigation through record sets using browser arrows or record set buttons working. Form meta menu appearing. Able to create record set from entered items.

PR #3302, opened 2023-04-05
Add http2 and http3 to nginx container

### LINKED ISSUE #2608: Add HTTP2 (and HTTP3) support in NGINX containers
These will bring performance improvements and future-proofing

Enabling them in dev containers and production containers would be great

CarolineDenis:
Fixes #2608

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -1,5 +1,5 @@
 server {
-    listen 80;
+    listen 80 http2 http3 quic;
     server_name localhost;
     root /usr/share/nginx;
     client_max_body_size 128M;


maxxxxxdlp (changes requested):
Did this change work for your?

I am getting an error in my nginx:

Screenshot 2023-04-05 at 09 23 45

maxxxxxdlp:
Closed because this PR is not complete, and http2/3 requires SSL, which we don't want to enable in development just yet because it adds additional overhead to getting containers running
The issue is still open and will come back to it in the future

PR #3304, opened 2023-04-05
Scope all front-end requests where needed

### LINKED ISSUE #2292: Some front-end requests are not properly scoped to current collection
I.e, the `http://localhost/specify/bycatalog/KUIT/000005083/?recordsetid=78` URL will ask for the `000005083` collection object, but does not scope the results to KUIT collection, thus could display a record from a different collection

There are probably other cases like this. Need to audit them

maxpatiiuk:
Fixes #2292

To test:
- Attachment counts are scoped to the current collection (can be checked by making sure the attachment count matches the number of displayed attachments in the attachment viewer) (they were not scoped before)
- Pick lists of type "Entire table" or "Single Field" have their pick list items scoped (they were not scoped before, @grantfitzsimmons do you see a problem with this change?)


grantfitzsimmons:
> Pick lists of type "Entire table" or "Single Field" have their pick list items scoped (they were not scoped before, @grantfitzsimmons do you see a problem with this change?)

I do see potential problems with that change, as items that were once visible will no longer be (and previously valid pick list values would be invalid as well).

I'm not sure if this has a large scope, however, as it is not common to use pick-list items based on an entire table or field, but it might be worth asking Andy or Theresa.

carlosmbe (commented):
Attachment Counts are still not scoped to current collection

Screen shots
Two different collections, same counts

Screenshot 2023-05-23 at 4 46 09 PM
Screenshot 2023-05-23 at 4 46 46 PM

maxxxxxdlp:
@melton-jason could you please pick up this PR?

Remaining things:
- Fix the issue discovered by Carlos
- Add a collection pref for making pick lists based on a table not scoped (like they were before this change) - the default should be scoped. Relevant place in code: https://github.com/specify/specify7/pull/3304/files#diff-d1cc9c7e699ae9b07df06cd17a5127059d55991ac2f6a3d0ebc92c557bee15d7

melton-jason:
> Fix the issue discovered by Carlos

The problem here is that `fetchCollection` would always set the `domainFilter` property of the query to undefined (and thus default to false). This is because Specify could not resolve the scoping relationship of Attachment as it doesn't have a field titled collection, collectionobject, discipline, division, or institution. When the scoping relationship can not be resolved, the `fetchCollection` would always 'remove' the domainfilter. 
Instead of being defined in a relationship, the 'scope' of an Attachment is defined in two columns on the Attachment table: `scopetype` (which is a number which represents/maps to the correct scoping level) and `scopeid` (the id of the scopetype).

The backend already has handling for scoping Attachments when `domainfilter=true` is provided:

https://github.com/specify/specify7/blob/05ffff3c457e2986d54d76664a1cf8b96c32776f/specifyweb/specify/filter_by_col.py#L18-L25

While the changes in https://github.com/specify/specify7/pull/3304/commits/05ffff3c457e2986d54d76664a1cf8b96c32776f are 'hacky', I think any generic solution on the frontend would be quite messy (if not impossible without hardcoding around the attachment case). 

> Add a collection pref for making pick lists based on a table not scoped (like they were before this change) - the default should be scoped. 

The collection preference can be added to the Remote Properties App Resource with the pattern: `sp7_scope_table_picklists_<COLLECTION_ID>`. (similar to the other Collection Preferences defined in this App Resource: CO_CREATE_COA, CO_CREATE_PREP, CO_CREATE_DET). 
The default value is `true`, which means Specify will try and scope the picklist items by default. 

Judging by the previous code and that in production, the picklistitems would only be unscoped (domainfilter set to false) when the table that the picklist is based off of is one of:
- Collection
- Discipline
- Division
- Institution

See https://github.com/specify/specify7/blob/52176a4bfdfefaba9b9afff2b99f862f7f8cd4ac/specifyweb/frontend/js_src/lib/components/PickLists/fetch.ts#L115-L117

This behavior has been preserved.

specifyweb/frontend/js_src/lib/components/DataModel/collection.ts
```diff
@@ -104,7 +104,8 @@ function mapValue(
   if (key === 'orderBy') return (value as string).toString().toLowerCase();
   else if (key === 'domainFilter') {
     const scopingField = schema.models[tableName].getScopingRelationship();
-    return value === true && typeof scopingField === 'object'
+    return value === true &&
+      (tableName === 'Attachment' || typeof scopingField === 'object')
```
maxpatiiuk, line 108: could this problem be better solved by using .getScope() instead of .getScopingRelationship()?
it seems to better resolve scoping for many (though not all - i.e storageattachment) attachment tables - https://github.com/specify/specify7/blob/issue-4307/specifyweb%2Ffrontend%2Fjs_src%2Flib%2Fcomponents%2FDataModel%2F__tests__%2F__snapshots__%2FspecifyTable.test.ts.snap#L1248C10-L1436

melton-jason: That would most likely work if Specify were getting the attachment counts on the `table` attachment tables, however the the collection fetch which does so is using the `Attachment` table only, and `scema.models.Attachment.getScope()` returns undefined. 

https://github.com/specify/specify7/blob/332069df197d376ed2f82b557c8598e66d12cd04/specifyweb/frontend/js_src/lib/components/Attachments/index.tsx#L105-L118

Although the suggestion would still better handle other cases in the future, so for now I will include it with this 'fix'.

melton-jason: See https://github.com/specify/specify7/pull/3304/commits/97bf9a80747a5cc4b1bbbe2ce8a082e88484cdce

CarolineDenis (approved):

emenslin (approved):
Looks good, attachments and picklists seem to be scoped properly 
Screenshot 2024-01-25 114034
Screenshot 2024-01-25 114104

PR #3312, opened 2023-04-06
Add a cancel button in search dialog

### LINKED ISSUE #3274: Close button on Express Search dialog
Express Search dialog doesn't have a close button. Only way to close it is to click outside of the dialog. But, the user can disable closing when clicked outside. So now, they can never close the dialog.

image

CarolineDenis:
Fixes #3274


maxxxxxdlp (changes requested):

maxxxxxdlp (approved):
Looks good.

I forgot to mention that `<Button.DialogClose>` is identical to `<Button.Gray>`, except it automatically finds the onClose callback of the dialog it is in, and calls that when the button is clicked

grantfitzsimmons (approved):

PR #3313, opened 2023-04-06
Fixes query line being behind dialog

### LINKED ISSUE #3279: Draggable query lines are displayed behind the dialog
See https://github.com/specify/specify7/pull/3272#issuecomment-1491986486
See https://github.com/specify/specify7/pull/3272#issuecomment-1491994486

it is a z-index problem
and it's a bad bug
but we can't fix it in this release because we are supposed to have it released early next week, where as we haven't even merged anything yet
there is trickiness to this:
you need to use z-index higher than a dialog when inside a dialog
you need to use z-index lower than a dialog when outside a dialog, but a dialog is rendered on top (i.e, opened a notifications dialog, and have query builder in the background)
you need to have z-index in between when you have two dialogs
z-index of a dialog increases every time you click on it:

https://user-images.githubusercontent.com/40512816/229149070-c538fdc3-ecd3-4e6d-a97c-d22cad18bcb9.mov

CarolineDenis:
Fixes #3279

To test: 
- start to drag a query line and do not release 
- use tab until you focus close button
- press enter 
- the dialog should close and the line dragged should disappears

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
@@ -10,6 +10,7 @@ import { scrollIntoView } from '../TreeView/helpers';
 import type { MappingPath } from '../WbPlanView/Mapper';
 import type { QueryField } from './helpers';
 import { QueryLine } from './Line';
+import { f } from '../../utils/functools';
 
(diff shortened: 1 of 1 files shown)


chanulee1 (approved):
Dialog closes and dragged query line disappears as per testing instructions. Is there anything else that needs to be tested?

PR #3317, opened 2023-04-07
Focus first field when opening form + possibility to disable in pref

### LINKED ISSUE #1543: Opening forms should highlight first field for data entry
**Describe the bug**
As in Specify 6, Specify 7 should place the cursor in the fist field on the form ready for data entry rather than opening with nothing selected.

**To Reproduce**
1. Go to any data entry form
2. Note that cursor is not active in any field
3. See error

**Expected behavior**
Would expect top left hand most field to have cursor active for data entry on opening form

**Screenshots**

 - OS: [e.g. Windows 10]

**Reported By**
Name of your institution

**Additional context**

CarolineDenis:
Fixes #1543


specifyweb/frontend/js_src/lib/components/Forms/SpecifyForm.tsx
```diff
@@ -82,6 +82,29 @@ export function SpecifyForm<SCHEMA extends AnySchema>({
...
+
+  const [focusFirstFieldPref] = userPreferences.use(
+    'form',
+    'behavior',
+    'focusFirstField'
+  );
+  const formRef = React.useRef<HTMLDivElement | null>(null);
+
+  React.useEffect(() => {
+    if (focusFirstFieldPref) {
```
maxxxxxdlp, line 94: (optional) Just a personal opinion

In cases like this you can make the code more readable by replacing your condition with the following:
```js
if (!focusFirstFieldPref) return undefined;
```

That way you don't need `{` and `}`, and the has one less indentation level

maxxxxxdlp (approved):

chanulee1 (changes requested):
When navigating between records in a record set and when using "Browse In Forms" option in querybuilder, the first field of the last subview seems to become focused if the subview contains a record.

https://user-images.githubusercontent.com/103299204/233201679-6cf5afe9-cb62-470d-b3e2-1b866e340e48.mp4

Also, opening record sets is slightly epileptic.

https://user-images.githubusercontent.com/103299204/233201906-d64fd43c-320e-4c45-b2e9-39bba092e4de.mp4

https://coldfish-issue-1543.test.specifysystems.org/specify/view/collectionobject/6/?recordsetid=44

specifyweb/frontend/js_src/lib/components/Forms/BaseResourceView.tsx
```diff
@@ -57,7 +57,7 @@ export function useResourceView<SCHEMA extends AnySchema>({
             if (resource === undefined) return undefined;
             format(resource)
               .then((title) => {
-                setFormatted(title ?? '');
+                setFormatted(title ?? ('' as LocalizedString));
```
maxxxxxdlp, line 60: This change is not necessary.
The error here is created by your ide being misconfigured. It's not an actual typescript error

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
```diff
@@ -171,6 +173,10 @@ export function ResourceView<SCHEMA extends AnySchema>({
     resource,
     viewName,
   });
+  console.log(formElement);
```
maxxxxxdlp, line 176: remove console log

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
```diff
@@ -171,6 +173,10 @@ export function ResourceView<SCHEMA extends AnySchema>({
...
+  React.useEffect(() => {
+    formRef?.(formElement);
+  }, [formElement]);
```
maxxxxxdlp, line 179: (design decision) add `formRef` to the dependency list?

maxxxxxdlp (changes requested):

CarolineDenis:
Focus first field should apply to: 
- New form opening 
- Opening Subform 
- Opening Subview

Is not applicable in : 
- record set 
- when navigating in new entry form subview with slider arrows

CarolineDenis:
@maxxxxxdlp , if you have time, can you take a look at the error attached to forwardRef line 124 in Forms > SpecifyForm.tsx?

chanulee1 (changes requested):
Does not seem to work for grid-type subviews. Works for everything else.


https://github.com/specify/specify7/assets/103299204/ec16a368-3bb9-4265-95aa-c8941a271193

https://kufish51623-issue-1543.test.specifysystems.org/specify/view/collectionobject/new/

CarolineDenis:
> It seems that forwardRef that we use for the other forms to focus the first field is already used for scrollRef in FormTable.
> We should decide whether to change the code or to accept that the first field is not focused on grids.

After discussing this in the meeting we will merge this code and open a new issue to focus first field when grid view.

chanulee1 (approved):
Looks good.

CarolineDenis:
@maxxxxxdlp can this be merged?

maxxxxxdlp:
> @maxxxxxdlp , if you have time, can you take a look at the error attached to forwardRef line 124 in Forms > SpecifyForm.tsx?

ChatGPT:
image

Yes, this issue gets me too sometimes.
Basically, typescript is being too smart/ too safe and doesn't allow us to do something that isn't completely type-safe - we are find with that as in this case we don't care whether the element is div or not, as long as it's an element - thus type assertion is the thing to do in this case.

After fixing that, could you also check if this null can be removed or if it's required? (it's redundant as `null` is always added automatically by React.RefObject type)
 
Screenshot 2023-06-14 at 20 00 10

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -200,11 +200,12 @@ export function IntegratedRecordSelector({
       relationship={relationship}
       onAdd={handleAdd}
       onDelete={handleDelete}
-      onSlide={(index): void =>
+      onSlide={(index): void => {
+        focusFirstField();
```
maxxxxxdlp, line 204: are you sure you want to call focusFirstField here?
I belive we talked about it and decided that it's a bad idea.

I.e if you want to go to the record that is 5 records to the right of current one, every time your press next it will move your focus and you will have to move it back - not very keyboard friendly.

Perharps the best solution is to call focusFirstField() if the next/previous button was pressed with a mouse, and don't call focusFirstField() if next/previous was pressed with a keyboard enter

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
```diff
@@ -204,7 +204,9 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
                   aria-label={addLabel}
                   disabled={mode === 'view'}
                   title={addLabel}
-                  onClick={handleAdding}
+                  onClick={() => {
+                    handleAdding();
```
maxxxxxdlp, line 208: (optional) undo this change - seems like it was made accidentally

specifyweb/frontend/js_src/lib/components/Forms/BaseResourceView.tsx
```diff
@@ -139,7 +144,7 @@ export function useResourceView<SCHEMA extends AnySchema>({
         </>
       ),
     title,
-    formElement: form,
+    formElement: formRef.current,
```
maxxxxxdlp, line 147: Could you please add a comment here regarding the following?
Remember how we talked about never using a value of a ref during render as updates to ref don't trigger re-render, thus this is likely to cause a bug.
Well, this place is an exception because we update the ref on each render (line 127), so the value is always up to date.
Please summarise/paraphrase the above information and put it into a comment

maxxxxxdlp:
> It seems that forwardRef that we use for the other forms to focus the first field is already used for scrollRef in FormTable. We should decide whether to change the code or to accept that the first field is not focused on grids.

Are you sure? It seems we just forgot to focusFirstField() call here:

https://github.com/specify/specify7/blob/2f16eb9321d4ba70aee653e5ee14b1912fdf0ad8/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx#L185-L188

maxxxxxdlp (changes requested):

CarolineDenis:
> > It seems that forwardRef that we use for the other forms to focus the first field is already used for scrollRef in FormTable. We should decide whether to change the code or to accept that the first field is not focused on grids.
> 
> Are you sure? It seems we just forgot to focusFirstField() call here:
> 
> https://github.com/specify/specify7/blob/2f16eb9321d4ba70aee653e5ee14b1912fdf0ad8/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx#L185-L188

Only adding focusFirst field() in the onAdd doesn't work, if you remember we tried it together and the focusableElement was undefined. I just tried it again and this doesn't fix it unfortunately

maxxxxxdlp:
> If we want to add the focusFirstField() to the add button We also would need to pass down containerRef until reaching DataEntry.Grid and DataEntry.Grid already has a ref no?

Ah you are right, I remember that now (and see it in the code).

Yes, you will have to merge two refs together - which is not difficult. Take a look at what `<Dialog />` component does with externalContainerRef - it's just a few lines of code. But it's fine if you delay this to a future fix - in that case please open an issue and mention this PR or what's left to do in that issue (though opening the issue may take longer than fixing this 😄 )

maxxxxxdlp (approved):

maxxxxxdlp:
Regarding https://github.com/specify/specify7/pull/3317#discussion_r1230365181, you marked it as resolved. Does it mean you won't change the code, or that you haven't gotten around to changing the code?

Besides that all is good

CarolineDenis:
> Regarding [#3317 (comment)](https://github.com/specify/specify7/pull/3317#discussion_r1230365181), you marked it as resolved. Does it mean you won't change the code, or that you haven't gotten around to changing the code?
> 
> Besides that all is good

I did change the code. I think you are right and we shouldn't call the focus when sliding. I therefore remove the function call from the onSlide.

PR #3334, opened 2023-04-11
Implement skeleton loader (forms, query builder, attachment)

### LINKED ISSUE #2998: Reduce loading dialogs in favor of skeleton loaders
They provide much better UX and can disappear incrementally as parts of content get loaded.

Example:

image

FYI: there are libraries that provide skeletons (some of them even work with tailwind). But even without them, it shouldnt' be hard to create reusable skeleton elements (I,e circles, lines and rectangles that are pulsing synchronously)

CarolineDenis:
Fixes #2998


CarolineDenis:
To test: 
=> Change in the dev tool network tab: 'No throttling' to 'Slow 3G'

- Open Query Builder, while loading, the skeleton should be displayed 
- Open a record set that contains attachments, click on attachment button icon, while attachment is loading the skeleton should be displayed. 
- Open attachment page and see skeleton while attachments are loading 
- Open an App Resource and while loading on the right, see skeleton 

PS: If 3G not slow enough, you will have to pull the branch and replace the expected data by a temporary hardcoded 'undefined'.

CarolineDenis:
Preview of a skeleton: 

https://user-images.githubusercontent.com/108160931/231287065-077b8170-6aca-4ab3-ae09-406ef92c4342.mov

specifyweb/frontend/js_src/lib/components/SkeletonLoaders/AttachmentGallery.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  function createRectangles() {
+    const rectWidth = 22;
+    const rectHeight = 22;
+    const rectRadius = 2;
+    let rectX = 5; // Starting x position
+    let rectY = 0; // Starting y position
+    const rectangles = []; // Array to hold the generated rectangles
+
+    for (let i = 0; i < DEFAULT_FETCH_LIMIT; i++) {
```
maxxxxxdlp, line 19: there are just 3 `for (let ...` loops in the entire sp7 codebase. we generally try to avoid mutation whenever possible

Could replace you code with something like this:
```ts
Array.from({length:DEFAULT_FETCH_LIMIT}, (_,index)=>
  <rect
    ...
    // example logic, not sure if it matches what you had:
    x={rectWidth * index + 5}
    y={rectHeight * index}
    // you can also use something like `(index % 4)` or `Math.ceil(index / 4)` if needed
  />
)
```

maxxxxxdlp: Benefits of not mutating:
- When debugging you can execute the code in your console, without fearing side-effects messing up things
- Each iteration of the loop does not depend on the previous one, thus way less stuff to keep in the head when trying to understand what it is doing
- Matches the general react philosophy of not mutating data
- More more and more benefits (look up online if you are interested)

maxxxxxdlp: kind of related, but I don't like how this library requires you to hardcode all these pixel values.

i.e, something like this would have been so much simpler:
```tsx
<div className="flex flex-wrap gap-4">
  {Array.from({length: DEFAULT_FETCH_LIMIT}, ()=>
    <div className="skeleton-square w-64 h-64 rounded"  />
  )}
</div>
```

(flexbox takes care of all the positioning, rather than dealing with ugly pixel values and doing math)

Isn't that simpler and less error prone and easier to implement?

specifyweb/frontend/js_src/lib/components/SkeletonLoaders/AppResource.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+import { userPreferences } from '../Preferences/userPreferences';
+
+export function AppResourceSkeleton(): JSX.Element {
+  const [motionPref] = userPreferences.use('general', 'ui', 'reduceMotion');
+  return (
+    <div className="h-full w-full">
+      <ContentLoader
+        animate={motionPref !== 'reduce'}
+        backgroundColor="#333"
```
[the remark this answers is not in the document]
maxxxxxdlp: also, whatever colors you settle on, refactor the code in some way so that you don't have to specify these colors for every skeleton.

i.e, either assign them to a variable, or create a wrapper component around <ContentLoader> that would also provide the value for animate and speed props

if you were on the jscpd branch, I'm sure it will complain about all this code duplication

specifyweb/frontend/js_src/lib/components/SkeletonLoaders/QueryBuilder.tsx
```diff
@@ -0,0 +1,42 @@
...
+  return (
+    <div className="h-full w-full">
+      <ContentLoader
+        animate={motionPref !== 'reduce'}
+        backgroundColor="#333"
+        foregroundColor="#999"
+        speed={3}
+        viewBox="0 0 240 400"
+      >
+        <rect height="5" rx="2" ry="2" width="5" x="9" y="5" />
```
maxxxxxdlp, line 17: is there no better way to do this than a wall of copy pasted code?
i.e, a map?

maxxxxxdlp (changes requested):
looks very nice in the UI
code could use some improvement though

grantfitzsimmons (changes requested):
The skeleton is too stark in contrast to the background (black with a while gradient, should be   something more like this:
image

Than this:
image

Also it is used in place of the loading indicator when the loading circle should still be used after attachments begin to display.

**Full video:**
https://user-images.githubusercontent.com/37256050/231314507-d62041a1-d632-4f04-8962-fd69fa7eab99.mp4

grantfitzsimmons:
It would be very nice to have skeleton loading dialogs instead. Forms generally do not have any loading time & some of the components that take a long time to load are:
- The list of queries
- The list of record sets
- Creating an interaction

grantfitzsimmons:
image

Similar to my last feedback, this looks really odd with the black outline & makes the dialog **massive** when it should be very small

grantfitzsimmons:
See how this loads into a giant dialog that **includes a loading wheel** (not needed) then shrinks back to a small window upon completion? When things load quickly, this flashes by and looks bad

https://user-images.githubusercontent.com/37256050/231315147-168ccd80-303e-4e4d-aaee-0d5e60ac474d.mov

specifyweb/frontend/js_src/lib/components/SkeletonLoaders/AttachmentGallery.tsx
```diff
@@ -1,55 +1,25 @@
...
+  const rectWidth = 22;
+  const rectHeight = 22;
+  const rectRadius = 2;
+  const rectangles = Array.from({ length: DEFAULT_FETCH_LIMIT }, (_, index) => (
+    <rect
+      height={rectHeight}
+      rx={rectRadius}
+      ry={rectRadius}
+      width={rectWidth}
+      x={rectWidth * index + 5}
```
maxxxxxdlp, line 20: Looking closer, this should be replaced by `x={rectWidth * (index % 4) + 5}`

and `y={rectHeight + (rectHeight + 10) * (Math.floor(index / 4))}`

maxxxxxdlp: but again, I would urge you to find a solution that does not involve hard-coded pixel values
i.e, this fails completely on very large screens and on mobile screens.
Please look for a solution that involves using `<div>` and flexbox rather than `<SVG>` to avoid having to deal with hardcoded pixel values

maxxxxxdlp (changes requested):

CarolineDenis:
@grantfitzsimmons can you check this version for the previous components and send me your thoughts?

grantfitzsimmons (changes requested):
image

https://coldfish-issue-2998.test.specifysystems.org/specify/

Open the queries dialog and see that it appears very condensed

Screenshot 2023-04-13 at 9 03 57 AM

maxxxxxdlp:
> Open the queries dialog and see that it appears very condensed

@grantfitzsimmons This is likely a user error

Dialogs remember their size and position.

You likely resized it like that intentionally or accidentally

grantfitzsimmons:
> You likely resized it like that intentionally or accidentally
> 
> If this is the case, then it would be claned when you wipe your localStorage or use incognito

No way! I refuse to believe it.

specifyweb/frontend/js_src/lib/components/SkeletonLoaders/Form.tsx
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+
+import { Skeleton } from './Skeleton';
+
+export function FormSkeleton() {
+  return (
+    <div className="w-full">
```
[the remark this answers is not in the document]
maxxxxxdlp: consider creating one loader for Collection Object form as it's usually tall and very common, and a different, much smaller loader for all the other forms to avoid the issue Grant mention where skeleton is way larger than the form

specifyweb/frontend/js_src/lib/components/Toolbar/RecordSets.tsx
```diff
@@ -208,7 +211,16 @@ export function RecordSetsDialog({
...
+  ) : displayLoading ? null : (
+    <Dialog
+      buttons={<Button.DialogClose>{commonText.cancel()}</Button.DialogClose>}
+      header={commonText.recordSets()}
+      icon={<span className="text-blue-500">{icons.collection}</span>}
+      onClose={handleClose}
+    >
+      <DialogListSkeleton />
+    </Dialog>
+  );
```
maxxxxxdlp, line 223: feel free to merge this branch with paginator as there is overlap in the code. Plus you might need the dialog sizing fixes you did on that branch for some of the skeleton sizing issues. I imagine there would be some merge conflicts, but even more if you make even more changes.

For the future:
1. The life span of pull requests should be reduced - they shouldn't be open for a month or two (in this case it's mostly my fault because I was slow to review it - I will dedicate more time for pull request reviews going forward)
2. Try not to work on the same file on different branches. Either delay working until the other pull request is merged, or do both changes on the same branch, OR create a new branch on top of the other branch (i.e, create A on top of production, and create B on top of A - in this case when creating pull request for branch B, it's important to pick the destination branch as `A`, not `production`)

maxxxxxdlp (changes requested):
lots of small suggestions, nothing major

maxxxxxdlp:
All looks good and I will approve this pull request as soon as https://github.com/specify/specify7/pull/3334/files#r1173934603 is fixed. Tell me if you want to meet to talk about that

CarolineDenis:
> All looks good and I will approve this pull request as soon as https://github.com/specify/specify7/pull/3334/files#r1173934603 is fixed. Tell me if you want to meet to talk about that

yes would be nice

CarolineDenis:
> All looks good and I will approve this pull request as soon as https://github.com/specify/specify7/pull/3334/files#r1173934603 is fixed. Tell me if you want to meet to talk about that

Added the aria prop to Skeleton.Root

maxxxxxdlp:
There are still two unresolved comments. Besides those all looks good

specifyweb/frontend/js_src/lib/components/SkeletonLoaders/AttachmentPlugin.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <Skeleton.Root className="flex-col flex-wrap md:flex-row">
+      <Skeleton.TallRectangle className="min-w-[30vh] md:flex-1" />
+      <div className="flex flex-col items-center gap-4">
+        <div className="flex items-center gap-2 ">
+          <Skeleton.Rectangle />
+          <Skeleton.SmallSquare />
+        </div>
+        {Array.from({ length: 6 }, () => (
+          <div className="flex max-w-[80%] gap-4 p-2">
```
maxxxxxdlp, line 15: forgot key prop.
(ESLint can detect this error)

specifyweb/frontend/js_src/lib/components/SkeletonLoaders/DialogList.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+import { Skeleton } from './Skeleton';
+
+export function DialogListSkeleton() {
+  return (
+    <Skeleton.Root className="flex-col">
+      <div className="m-2 flex justify-center gap-6">
+        <Skeleton.Rectangle className="!w-80" />
+      </div>
+      {Array.from({ length: 5 }, () => (
```
maxxxxxdlp, line 11: forgot key prop

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

grantfitzsimmons (changes requested):
_Almost ready!_

Just one more thing, the form dialog is really wide by default, can we made this narrower initially?

https://user-images.githubusercontent.com/37256050/236219183-4a68d6b1-e589-4a55-b0df-e1977fb919ab.mov

CarolineDenis:
@grantfitzsimmons could you test this PR this week please?

grantfitzsimmons (approved):

carlosmbe (approved):

PR #3336, opened 2023-04-12
Fixes query builder bugs causing js render issues and autorun stat queries

realVinayak:


maxxxxxdlp (approved):

grantfitzsimmons:
@realVinayak How should we test this?

maxxxxxdlp:
1. try to run a query that has a field with incomplete mapping - the field should be filtered out instead of crash
2. clicking on a query stat, should auto-run the query, rather than require the user to click on "Query"

realVinayak:
Yes, Max describes it perfectly. However, I've made these changes on use-dynamic-stats branch and I'd rather have this tested it on that branch.

grantfitzsimmons:
@realVinayak If that is the case, can you close this PR and describe this on the appropraite PR?

Issue #641, opened 2020-03-26
Fix back-end not supporting time zones


benanhalt:
https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-TIME_ZONE

https://docs.djangoproject.com/en/3.0/topics/i18n/timezones/

maxxxxxdlp:
This affects the front-end. More info:

https://github.com/specify/specify7/blob/4d007df018f61b003128d74bcdbf808083e35886/specifyweb/frontend/js_src/lib/components/internationalization.tsx#L177-L188

maxxxxxdlp:
From @chanulee1:

> The "Date Edited" field on the CO form is populated with tomorrow's date. Current date at time of writing this issue: 8:37 pm 11/24/2022 CST.
>
>
> https://user-images.githubusercontent.com/103299204/203888983-fa30572c-8922-4735-a7b5-553fad3f2653.mp4
>
> https://freshfish-edge.test.specifysystems.org/specify/view/collectionobject/new/

maxxxxxdlp:
@melton-jason I know you looked into this issue. If you faced a roadblock preventing you from fixing this issue, please post the details here for future reference

maxxxxxdlp:
From @grantfitzsimmons:

> <img width="802" alt="image" src="https://github.com/specify/specify7/assets/37256050/60a4a74e-7cff-4a5b-a73c-3c42ef81987e">
>
> The timestamps shown (`now`, `last month`, `yesterday`, etc.) are relative to the current user's time, but these should be relative to the system time so that the dates are not in the future.
>
> The database stores the timestamp for the system, not the user accessing Specify in the browser.

emenslin:
This is still an issue with the batch attachment upload. The time uploaded says the correct time but the timestamp created and modified are showing the incorrect time. 
<img width="958" alt="Screenshot 2024-01-18 115950" src="https://github.com/specify/specify7/assets/150940840/c56c7ec6-7268-47b9-a7b0-d5701bc9f56c">

maxpatiiuk:
@melton-jason 

> The best-case solution for this would be to create a middleware which catches any naive datetime objects and converts them to aware datetime objects.

This sounds good

> In both of these solutions, we would have to assume some timezone information (most likely midnight).

emenslin:
From @lexiclevenger 

> **Describe the bug**
[...]

PR #3346, opened 2023-04-14
Add since=<date> to notifcations fetching

### LINKED ISSUE #2606: Make Notifications fetching and stats reporting more efficient
Currently, notifications are fetched from the back-end using polling (front-end calls back-end regularly - regularity depends on how long the page is open for)

Instead, there are some more high-tech solutions:
- Web Sockets
- Push API - https://developer.mozilla.org/en-US/docs/Web/API/Push_API
- Server-sent events - https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events
- Or at the very least, fix the bug introduced in 7.7, where the front-end does not send the `?since=<date>` query string argument to the back-end when fetching notifications so as to only fetch the notifications that haven't been fetched yet.
- Web Notifications

CarolineDenis:
Fixes #2606


specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -51,7 +51,10 @@ export function Notifications({
 
     let timeout: NodeJS.Timeout | undefined = undefined;
 
-    function doFetch(): void {
+    function doFetch(since?: Date): void {
+      const queryString = since
```
maxxxxxdlp, line 55: Replace to use `formatUrl()` utility function

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -109,7 +112,7 @@ export function Notifications({
           timeout =
             document.visibilityState === 'hidden'
               ? undefined
-              : globalThis.setTimeout(doFetch, pullInterval);
+              : globalThis.setTimeout(() => doFetch(new Date()), pullInterval);
```
maxxxxxdlp, line 115: This is wrong behavior. When fetching again, it should only fetch notifications that have been created since the last time we fetched. Thus, you should store the date it was last fetched in some variable and use that next time you fetch, rather than `new Date()`

maxxxxxdlp: Also, you need to update the code to append newly fetched notifications to the notifications state, rather than reply all existing notifications with newly fetched

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -72,25 +83,35 @@ export function Notifications({
...
+        .then(({ data: newNotifications }) => {
           if (destructorCalled) return undefined;
-          setNotifications(
-            notifications
-              .map(
+          setNotifications((existingNotifications) => {
+            const filteredNewNotifications = newNotifications.filter(
+              (newNotif) =>
+                !existingNotifications?.some(
+                  (existingNotif) =>
```
maxxxxxdlp, line 108: don't use `Notif` 🙁. doesn't your editor highlight that as a spelling error?
use `Notification` instead

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -99,23 +120,25 @@ export function Notifications({
                   type,
                   payload: rest as IR<LocalizedString>,
                 })
+              ),
+            ].sort(
```
maxxxxxdlp, line 124: (optional) the `// Make most recent notification first` comment was lost in the process. not a big deal, just a FYI in case that wasn't intentional

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -124,7 +147,9 @@ export function Notifications({
     return (): void => {
       document.removeEventListener('visibilitychange', handler);
       destructorCalled = true;
-      if (timeout !== undefined) globalThis.clearTimeout(timeout);
+      if (timeout !== undefined) {
```
maxxxxxdlp, line 150: @CarolineDenis @melton-jason can you weight in on whether you prefer single statement `if` to be with `{` `}` or without?

I generally find it's easier to make types and harder to see how to fix a typo when there are a lot of `}` and `)` near each other, so as a general rule try to avoid them as much as possible. That also makes code cleaner, looking more like Python.

Though I could see the benefit of always using `{` `}` as that would reduce the git diff when changing to have one/more than one statement.

Whatever you chose, I would just advise to be consistent with it. We have an eslint rule that removes all redundant  `{` `}`, but that is not running at the moment due to tests failing

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -99,23 +120,25 @@ export function Notifications({
...
-              ? undefined
-              : globalThis.setTimeout(() => doFetch(new Date()), pullInterval);
-          return undefined;
+            );
+          });
         })
         .catch(console.error);
+
+      // Stop updating if tab is hidden
+      timeout =
```
maxxxxxdlp, line 135: I don't think this should be outside the `.then()` or else you risk starting fetching notifications again even before previous notifications have finished fetching

maxxxxxdlp: https://www.techtarget.com/searchstorage/definition/race-condition?Offer=abMeterCharCount_var1

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -55,6 +64,8 @@ export function Notifications({
       const queryString = since
         ? `?since=${encodeURIComponent(since.toISOString())}`
         : '';
+      lastFetchedTimestamp.current = since;
```
maxxxxxdlp, line 67: hmm, you set `lastFetchedTimestamp.current` to `since`, and then you call `doFetch` again with the value of `lastFetchedTimestamp.current` - but you never call `new Date()` to get a new timestamp???

maxxxxxdlp, line 67: also, you should get the current timestamp when you begin fetching, but only update the last fetched timestamp variable once fetch has completed successfully - so that if one of the fetches fails (i.e due to network error), but the next one succeeds, the next fetch doesn't miss any notifications that happened in the meanwhile

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -42,16 +43,34 @@ export function Notifications({
...
 
-    function doFetch(): void {
+    function doFetch(since?: Date): void {
+      const startFetchTimestamp = new Date();
+      const queryString = since
+        ? encodeURIComponent(since.toISOString())
+        : // : '';
+          encodeURIComponent(new Date().toISOString());
+      //should look like this: `/notifications/messages/?since=2022-12-09%2020:35:23.000`
+      const url = formatUrl(`/notifications/messages/`, {
```
CarolineDenis, line 69: @maxxxxxdlp , the date format passed to the back-end url is not correct, what do you advise me to use?

maxxxxxdlp: does back-end accept `new Date().toJSON()`?


also, see at what the main `/api/specify/` endpoint does to the date it receives from the front-end: https://github.com/specify/specify7/blob/207df9a091c59cd04e4e4b1c3eb111c75c74f68a/specifyweb/specify/api.py#L670-L673

thus, does this work: `new Date().toISOString().replace('T', ' ')`?

also, see where back-end uses the date: https://github.com/specify/specify7/blob/207df9a091c59cd04e4e4b1c3eb111c75c74f68a/specifyweb/notifications/views.py#L20-L21. if you need to, you can modify that code

also, you can construct the date from pieces on the front-end. i.e `${date.getFullYear()}-${date.getMonth()+1}-...`

---

I also looked at the history to see what was used in there in the past (see line 34) - https://github.com/specify/specify7/commit/c34aa8e84502b325e3c9333781ba58f6fc47e694#diff-da092c6e0e273c53b4c8a86465d61826b5752a74dd6aa7cf03848be7d7aeb32f - it was just using the date from the last notifications - simple, smart.

in the process, beware of the time zones. Until https://github.com/specify/specify7/issues/641 is fixed, it's possible that there is no clean way to pass a date from front-end to back-end. If you do see the timezone causing a problem, try using the above solution of just sending back to the back-end the date of the most recent notification

maxxxxxdlp: also, feel free to involve @melton-jason in the process as he did his share of fighting with django and dates 😊 so is a seasoned date samurai now 🥷.

melton-jason: I'm unfortunately quite the time traveler now :hourglass_flowing_sand: :runner:

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -42,16 +43,52 @@ export function Notifications({
...
+
     document.addEventListener('visibilitychange', handler);
 
     let timeout: NodeJS.Timeout | undefined = undefined;
 
-    function doFetch(): void {
+    function doFetch(since?: Date): void {
+      const startFetchTimestamp = new Date();
+
+      const now = new Date();
```
maxxxxxdlp, line 65: does `new Date().toISOString().replace('T', ' ')` or `new Date().toJSON().replace('T', ' ')` not work?
it works in other places where front-end communicates with back-end so should work here too - and would be cleaner than this

or at very least extract this logic into some sort of function! don't you see how much code duplication there is here? 🫨

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -42,16 +43,52 @@ export function Notifications({
...
+      const sinceYear = sinceDate?.getFullYear();
+      const sinceMonth = sinceDate?.getMonth().toString();
+      const sinceDay = sinceDate?.getDate().toString();
+      const sinceHours = sinceDate?.getHours().toString();
+      const sinceMinutes = sinceDate?.getMinutes().toString();
+      const sinceSeconds = sinceDate?.getSeconds().toString();
+
+      const formattedSince = `${sinceYear}-${sinceMonth}-${sinceDay} ${sinceHours}:${sinceMinutes}:${sinceSeconds}`;
+
+      const queryString = since ? formattedSince : formattedDate;
```
maxxxxxdlp, line 85: improve variable name. this variable is not a query string.
a query string looks like `?name=value` - this isn't that

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -42,16 +43,52 @@ export function Notifications({
...
+      const url = formatUrl(`/notifications/messages/`, {
+        since: queryString,
+      });
+
+      lastFetchedTimestamp = new Date();
```
maxxxxxdlp, line 90: this line shouldn't be here

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -85,11 +122,20 @@ export function Notifications({
...
-          setNotifications(
-            notifications
-              .map(
+          setNotifications((existingNotifications) => {
+            const filteredNewNotifications = newNotifications.filter(
+              (newNotification) =>
+                !existingNotifications?.some(
+                  (existingNotification) =>
+                    existingNotification.messageId ===
+                    newNotification.message_id
```
maxxxxxdlp, line 133: can you reorder this code?
first .map() over notifications to convert `message_id` to `messageId` and the rest
and only afterward do `.filter()` and `.some()` - to avoid having to use message_id more than absolutely necessary

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
maxxxxxdlp: take some time to do research on how the front-end currently sends dates to the back-end in other places (i.e when you set a `cataloggedDate` field on the collection object - will be helpful for you both for this issue and for future

CarolineDenis: Unfortunately no, new Date().toISOString().replace('T', ' ') or new Date().toJSON().replace('T', ' ') do not work

CarolineDenis: I tried: 
- Date().toISOString().replace('T', ' ') 
- new Date().toJSON().replace('T', ' ') 
- encodeURIComponent(new Date().toISOString())
- new Date().toISOString()
- new Date().toJSON()
- `${new Date().getFullYear()}-${new Date().getMonth()}-${new Date().getDay()} ${new Date().getHours()}:${new Date().getMinutes()}:${new Date().getSeconds()}}
          `
- new Date().toISOString().replace('T', ' ')
- const now = new Date();
const formattedDate = now.toLocaleString();

Having errors with all of those I tested with the long version using get(...

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -59,36 +58,25 @@ export function Notifications({
...
-      const startFetchTimestamp = new Date();
-
-      const now = new Date();
-      const year = now.getFullYear();
-      const month = now.getMonth().toString();
-      const day = now.getDate().toString();
-      const hours = now.getHours().toString();
-      const minutes = now.getMinutes().toString();
-      const seconds = now.getSeconds().toString();
+    function formatDate(date: Date) {
```
maxxxxxdlp, line 61: since this function does not rely on any variable declared in this component, it would be best to move it outside the component (i. right after this component, or into a separate file)

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -59,36 +58,25 @@ export function Notifications({
...
-      const sinceSeconds = sinceDate?.getSeconds().toString();
+      return `${year}-${month}-${day} ${hours}:${minutes}:${seconds}`;
+    }
 
-      const formattedSince = `${sinceYear}-${sinceMonth}-${sinceDay} ${sinceHours}:${sinceMinutes}:${sinceSeconds}`;
+    function doFetch(since?: Date): void {
+      const startFetchTimestamp = new Date();
 
-      const queryString = since ? formattedSince : formattedDate;
+      const sinceValue = since ? formatDate(since) : formatDate(new Date());
```
maxxxxxdlp, line 75:
```suggestion
const singeValue = formatDate(since ?? new Date())
```

or replace
```ts
function doFetch(since?: Date): void {
```

with
```ts
function doFetch(since? = new Date()): void {
```
and then this line will be simpler:
```suggestion
const sinceValue = formatDate(since);
```

at which point, if you want, you can even remove the `sinceValue` variable, and just inline the `formatDate(since)` when you call formatUrl()

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/Header/helpers.ts
```diff
@@ -0,0 +1,10 @@
+export function formatDate(date: Date) {
```
maxxxxxdlp, line 1: (optional) move this to `utils/parser/dateFormat.ts`
(optional) rename the function to `formatDateForBackEnd` or `dateToPython` or something like that (or provide a comment)
that will make this function more discoverable in case someone would need it in the future

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -85,34 +99,47 @@ export function Notifications({
...
+                ({ timestamp }) => new Date(timestamp).getTime(),
+                true
               )
-          );
+            );
+          });
+          lastFetchedTimestamp = startFetchTimestamp;
           // Stop updating if tab is hidden
-          timeout =
+          return (timeout =
```
maxxxxxdlp, line 136: this is not very readable - try to avoid assigning and returning on the same line

maxxxxxdlp (approved):
looking good, though the logic is complicated, and important, thus I would encourage to take a bit of time to write tests for this (Feel free to ask chatgpt or Jason or me for help) - would also be good practice for the future

i.e mock the network requests using `overrideAjax()`, also mock INITIAL_INTERVAL to replace it with a smaller duration (i.e 1ms - so that the test doesn't take long time to run) - then inspect that the state is updated on first fetch and on second fetch like you would expect

melton-jason:
Wouldn't it make sense for the vast majority of these tests to be on the backend rather than frontend? 
At least as far as testing the functionality the endpoint is concerned. 

I agree, frontend tests would be helpful as well. 

>  then inspect that the state is updated on first fetch and on second fetch like you would expect

For further clarification, you mean rendering the `Notifications` component in the Jest tests (probably via a function in [reactUtils.tsx](https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/tests/reactUtils.tsx)) after having mocked the network requests that would be made from the `doFetch` function. 
Then we can check the useState via mocking it with Jest or inspect the generated HTML?

maxxxxxdlp:
We change front-end code more often, and it does more complicated logic here (back-end just fetches from the database, where as front-end does merging, sorting and duplication of notifications - and if that is wrong, it's not obvious, if back-end is wrong it's usually more obvious via crash or empty response) so I would say front-end tests should be prioritized

> For further clarification, you mean rendering the Notifications component in the Jest tests (probably via a function in [reactUtils.tsx](https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/tests/reactUtils.tsx)) after having mocked the network requests that would be made from the doFetch function.
Then we can check the useState via mocking it with Jest or inspect the generated HTML

As you wish. My suggestion would be to extract the fetching logic into a separate hook and test that hook in isolation - more readable code, and easier testing.
Then testing UI is lesser priority as this UI is pretty simple - but if you were to test it, then the way I would do it is mock the hook, render the component in a test, and then either take snapshot of the HTML or take screenshot of the rendered page (but would need to add screenshot tests support - which could be useful in the future anyway)

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -133,13 +133,14 @@ export function Notifications({
...
-          return (timeout =
+          timeout =
             document.visibilityState === 'hidden'
               ? undefined
               : globalThis.setTimeout(
                   () => doFetch(lastFetchedTimestamp),
                   pullInterval
-                ));
+                );
+          return timeout;
```
maxxxxxdlp, line 143: (optional) you don't actually need to return anything here as it's not being read by anyone?

maxxxxxdlp (approved):
The back-end test looks good, but I would say the front-end test would be more useful

also note that we have way more front-end code (and it is changed often), so I would say writing tests for the front-end is needed more

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
CarolineDenis: I had to add `return timeout` because of the error 'Not all code paths return a value.' on line 102 in the .then
Is there another solution for it?

CarolineDenis:
> The back-end test looks good (thanks for adding them), but I would say the front-end test would be more useful
> 
> also note that we have way more front-end code (and it is changed often), so I would say writing tests for the front-end is needed more

The back-end are still failing unfortunately, a problem with the ms

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
maxxxxxdlp: yes,
see the eslint doc for that page  -https://eslint.org/docs/latest/rules/consistent-return

to fix, replace `return undefined;` with `return` earlier in this function - then you won't need `return undefined`

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -42,10 +38,21 @@ export function Notifications({
     if (notificationCount === 0) handleClose();
   }, [notificationCount, handleClose]);
 
-  React.useEffect(() => {
-    let pullInterval = INITIAL_INTERVAL;
-    let lastFetchedTimestamp: Date | undefined;
+  let pullInterval = INITIAL_INTERVAL;
+  let lastFetchedTimestamp: Date | undefined;
```
maxxxxxdlp, line 42: this is not going to work
remember how react component work?

this function is called on each render! your variable is re-declared on each render
if you need your value to persist between renders you have to either:
1. use useState
2. use useRef
3. OR put the value in a closure (declare your `let` variable inside of useEffect like it was before!)

option 3 is the way to go here

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -1,28 +1,24 @@
...
 import { ErrorBoundary } from '../Errors/ErrorBoundary';
 import { DateElement } from '../Molecules/DateElement';
 import { Dialog, dialogClassNames, LoadingScreen } from '../Molecules/Dialog';
-import { formatUrl } from '../Router/queryString';
+import { useNotificationsFetch } from './hooks';
 import { MenuButton } from './index';
 import type { GenericNotification } from './NotificationRenderers';
 import { notificationRenderers } from './NotificationRenderers';
 
 const INITIAL_INTERVAL = 5000;
```
maxxxxxdlp, line 20: move these two into hooks.tsx

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
maxxxxxdlp: all of these variables should be moved into the hook

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -42,10 +38,21 @@ export function Notifications({
...
+  let destructorCalled = false;
+  const timeout: NodeJS.Timeout | undefined = undefined;
+
+  const { doFetch } = useNotificationsFetch({
+    pullInterval,
+    freezeFetchPromise,
+    destructorCalled,
+    notifications: [notifications, setNotifications],
+    lastFetchedTimestamp,
+    timeout,
```
maxxxxxdlp, line 52: move all of these into the hook

the hook should return just `notifications` and `doFetch` OR the hook can also take care of fetching (move the useEffect into the hook) and just return `notifications` and nothing else - you see how nicely that would move all of the notification fetching business into a hook, thus it can be tested independently of all the rendering logic (which would be in this file)

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
CarolineDenis: Do you want to move the state notifications in the hook?

melton-jason: Yes, the notifications state can be moved into the hook. 
I think the most important concept here is the decoupling of the logic which fetches the Notifications and the logic which renders them. 
Thus this can be done in a few ways (of which @maxxxxxdlp has mentioned): 
- the hook can return a (probably asynchronous) function (like `doFetch`) which returns the fetched notifications
  - This gives the most flexibility to the consumer of the hook, but that flexibility comes at a price and requires more work to be done _by_ the consumer
  
- the hook can initially fetch the notifications and return the fetched notifications and a way to re-fetch them
  - A more "laissez-faire" approach to the hook. The concept here might be to allow the consumer immediate access to the fetched notifications, but provide  control over when future fetching is needed
  
- the hook can only return the fetched notifications and handle all fetching logic
  - _All_ of the fetching logic should be included in the hook. This is the easiest for a consumer to use, but does sacrifice a little in the way of flexibility when compared to the other methods. (Consistency and rigidity probably also means this is the least bug-prone of the cases, at least once all of the initial bugs are ironed out)

maxxxxxdlp (approved):
All lloks very good.
Also, agree with what Jason said

however given how many changes have been made here, and that there are still no front-end tests, please make sure to test all of this thoroughly (or make someone else test it). Including:
- notifications are fetched
- new notifications are correctly appended to the existing results and sorted
- can delete notifications
- can mark notifications as deleted
- and optionally also test that notifications are not fetched while tab is not open (because another browser tab is open), and while tab is open, the fetch timeout scales

CarolineDenis:
> looking good, though the logic is complicated, and important, thus I would encourage to take a bit of time to write tests for this (Feel free to ask chatgpt or Jason or me for help) - would also be good practice for the future
> 
> i.e mock the network requests using `overrideAjax()`, also mock INITIAL_INTERVAL to replace it with a smaller duration (i.e 1ms - so that the test doesn't take long time to run) - then inspect that the state is updated on first fetch and on second fetch like you would expect

Would you have a good example in mind I could follow to write a test for the useNotifications hook?

maxxxxxdlp:
very good question, and thanks for agreeing to learn more about testing

this guide is a very nice and comprehensive tutorial: https://www.toptal.com/react/testing-react-hooks-tutorial (it's a long article, but has a lot of good details and foundational knowledge - if you have time, encourage you to read it, if not, skip down to where you see `[1]`)

That tutorial also creates a small test component via which you can test the hook - that is somewhat simple to understant, but involves more code as you need to forward the result of the prop out of the component somehow (either by returning from component, or via ref, or callback)

That tutorial than proceeds with creating a helper for testing hook - nice exercise

note, in the tutorial they mock the fetch function - mocking is useful to know and is used in many advanced tests, but you don't need to mock fetch in particular as it's already mocked in all specify 7 tests (if you use `ajax`) - thus instead of mocking it, use `overrideAjax` function (see usages in tests for how to use it - pretty simple - just give it a URL, and what should be the response from that URL)

---

[1]

And then finalizes by telling you that such helper already exists - and the library it suggests is by far the most popular option for testing hooks, and we already have it installed:

```tsx
import { renderHook } from "@testing-library/react";
```

(note, the name of the library is different than what is in the tutorial as it got [renamed recently](https://github.com/testing-library/react-hooks-testing-library#a-note-about-react-18-support))

The library has short 3 pages of documentation that will get you from 0 to hero: https://react-hooks-testing-library.com/usage/basic-hooks

---

My advice for you would be to first create a test for some simple hook like useTriggerState/useBooleanState/useAsyncState, and then write tests for the notifications hook

and when using overrideAjax(), provide the smallest test data you need to test something (to keep the test small, more maintainable)

----

@realVinayak @melton-jason I would highly encourage you to follow the above tutorials too

specifyweb/frontend/js_src/lib/components/Header/hooks.tsx
```diff
@@ -0,0 +1,145 @@
...
+import type { LocalizedString } from 'typesafe-i18n';
+
+import { ajax } from '../../utils/ajax';
+import { formatDateForBackEnd } from '../../utils/parser/dateFormat';
+import type { IR, RA } from '../../utils/types';
+import { sortFunction } from '../../utils/utils';
+import { formatUrl } from '../Router/queryString';
+import type { GenericNotification } from './NotificationRenderers';
+
+export const INITIAL_INTERVAL = 5000;
```
maxxxxxdlp, line 11: this no longer needs to be exported

specifyweb/frontend/js_src/lib/components/Header/__tests__/useNotificationsFetch.test.ts
```diff
@@ -61,3 +67,175 @@ test('Verify setNotifications function works', () => {
...
+        message_id: 8,
+        read: true,
+        timestamp: '2023-09-22T01:22:00',
+        type: 'query-export-to-csv-complete',
+        file: 'query_results_2023-05-19T01:22:00.782784.csv',
+      },
+    ]
+  );
+
+  test('Url makes correct address with since param and response parsed correclty', async () => {
```
maxxxxxdlp, line 215: typo. should be `correctly`

maxxxxxdlp (approved):
Looking very good!

I see that you mastered the basics of testing (and even some more advanced features like mocking, and waitFor)

see my slack message for more details and advise

I hope that you can apply what you learned in the process to write tests for features you add in the future (with priority given to testing code that is complicated, but easy to test)

grantfitzsimmons (changes requested):
@bronwyncombs discovered that the JSON responses were the same as in `v7.9`. 

I found this to be strange, given that when I call the API when entering a valid date, I only see notifications added _since_ that date. 

That's when I realized that the notification was fetching notifications since **this day, last month**. It is one month behind...

```
https://fwri6123-issue-2606.test.specifysystems.org/notifications/messages/?since=2023-8-27+15%3A37%3A48
```

image

🤨

maxxxxxdlp (approved):
looking great!

(except for the fact that older notifications are not fetched - thanks for catching that - that would have not been nice)

grantfitzsimmons (changes requested):
https://github.com/specify/specify7/assets/37256050/74366136-fa63-43dc-80f1-5cf2d0015032

grantfitzsimmons (approved):
Looks like the issue with all notifications being fetched was fixed after redeploying. 
#4050 is still a big prob– the fix might end up changing how this is implemented.

specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
```diff
@@ -10,6 +10,7 @@ import type { GenericNotification } from './NotificationRenderers';
 
 const INITIAL_INTERVAL = 5000;
 const INTERVAL_MULTIPLIER = 1.1;
+let addSinceToNotification = true;
```
maxxxxxdlp, line 13: this is a wrong location. it should be as a ref inside of `useNotificationsFetch()`
the notifications state is initialized to undefined when the hook is first called - so this boolean flag should be initialized at the same time

the Notifications component may be re-rendered completely (maybe when user changes the menu position)

specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
```diff
@@ -0,0 +1,161 @@
...
+        .then(({ data: newNotifications }) => {
+          addSinceToNotification = false;
+
+          if (destructorCalled) return;
+
+          setNotifications((existingNotifications) =>
+            mergeAndSortNotifications(existingNotifications, newNotifications)
+          );
+
+          lastFetchedTimestamp = startFetchTimestamp;
```
maxxxxxdlp, line 82: this should probably also be stored in a ref - right now lastFetchedTimestamp value is lost on useEffect re-executions (happens every time the user opens/closes the notifications component)

maxxxxxdlp:
Two issues. Can you please open a small follow-up pr for this?

CarolineDenis:
> Two issues. Can you please open a small follow-up pr for this?
> 
> Also, if needed, I would be happy to explain in more detail why these things are problem - so that hopefully you see these things in the future

new issue created called issue-2606-bis

specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
maxxxxxdlp: Explanation on why this could lead to a bug, and should be changed to a ref:

This value is used to denote whether it's a first fetch of notifications (thus need to fetch all) or if we are just appending to what we have fetched already (just fetch only new).

But, the notifications themself, are stored in a state in the hook - imagine this hook is instantiated again (meaning rendered from scratch, and state values are not preserved) (will mention why that may happen later, just assume it does for now), then notifications state is initialized from scratch with the undefined value.

yet addSinceToNotification would be false, because you set it to false on the useEffect execution in the previous instance of this hook. Now, you have a bug - no previous notifications will be fetched.

The solution is to store the "addSinceToNotification" value in the same place as where the notifications state is, so that if the hook is instantiated again, the notifications state is reset to undefined AND addSinceToNotification is reset to true at the same time.

That is why I suggested using a ref for "addSinceToNotification" - you don't care about this value for rendering UI components (unlike "notifications" state), and you don't care about being notified when this value changes - all you care about is having the newest value of this variable when you want check it's value in doFetch. That is why useRef is perfect for it, rather than useState.

Does this make sense?

Now for the reasons why hook may be instantiated again:
- Imagine some other code elsewhere wants to use this hook in another place. Possible reasons:
  - maybe you want an admin panel that lists notifications per user?
  - or the ability to see notifications in user preferences (so that you can customize them and see impact right away)
  - or someone wants to reuse this logic for something totally different (just like @realVinayak did with [drag and drop hook](https://github.com/specify/specify7/blob/298f7ca5d7d709e3a9b9ad6813e17e9a237b5c08/specifyweb/frontend/js_src/lib/components/Molecules/useDragDropFiles.ts) - this logic was intended originally to only be used in the FilePicker, but now the hook is used outside of FilePicker too). By not assuming that there is only one instance of this code executing at one time, reusing code becomes easy, without having to double check all logic to make sure no unsafe assumptions were made)
- The entire header component may be recreated. Possible cases:
  - User changes header position in user preferences
  - Or an error happens, which un-renders the entire app, but then we implement error recovery that leads user back to the home page
- Probably other cases too. But the point is it's unsafe, and confusing to those reading this code

---


looking at this again, you might be able to reduce the need for this variable by just checking inside of useEffect if notifications are undefined (without adding notifications to the dependency array) - not sure if that would work, but if it would that would be neat to try

specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
maxxxxxdlp: this one is simpler to explain.

lastFetchedTimestamp stores the timestamp of last fetch
yet this is a variable you created in useEffect
so every time useEffect re-runs, the value of this variable is lost, and new variable is created
and useEffect has `[isOpen]` dependency array, thus it re-runs every time user opens or closes notifications panel

but it is not right to clear the value of that variable on user open/close of notifications as that shouldn't be related to lastFetchedTimestamp. Instead, lastFetchedTimestamp is related to "notifications" state (since lastFetchedTimestamp roughly describes when notifications state was last updated), so this variable should have the same lifecycle as the "notifications" hook

PR #3363, opened 2023-04-17
Implement attachment viewer gallery in record sets

### LINKED ISSUE #2132: Ability to see all attachments associated with a record set
Specify 6 has that feature

CarolineDenis:
Fixes #2132


specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
```diff
@@ -218,15 +227,25 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
...
-
               {typeof newResource === 'object' ? (
                 <p className="flex-1">{formsText.creatingNewRecord()}</p>
               ) : (
                 <span
                   className={`flex-1 ${dialog === false ? '-ml-2' : '-ml-4'}`}
                 />
               )}
-
+              {hasAttachments && (
```
maxxxxxdlp, line 237: Consider extracting this attachment-related logic into a separate component, just to keep this file as devoid of attachment-related logic as possible.

For example, imagine this file just has this line:
```
<RecordSetAttachments records={records} onFetch={handleFetch} />
```

And then the actual button, the state that holds whether the attachment dialog is open and the actual RecordSetAttachments dialog would be handled in that component



(`RecordSetAttachments` would be a new component, and currently named `RecordSetAttachments` component could be renamed to something like `AttachmentsDialog` and won't need to be exported anymore as it would be in same file as `RecordSetAttachments`)

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { AttachmentGallery } from './Gallery';
+
+export function RecordSetAttachments<SCHEMA extends AnySchema>({
+  records,
+  onClose: handleClose,
+  onFetch: handleFetch,
+}: {
+  readonly records: RA<SpecifyResource<SCHEMA> | undefined>;
+  readonly onClose: () => void;
+  readonly onFetch?:
```
maxxxxxdlp, line 25: Why do you need both `?` and `| undefined`?
Given that `handleFetch` is an important prop to have, in order to not risk accidentally forgetting to provide it, it should not be marked as optional, thus you should remove `?`

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,113 @@
...
+    | ((index: number) => Promise<RA<number | undefined> | void>)
+    | undefined;
+}): JSX.Element {
+  const recordFetched = React.useRef<number>(0);
```
maxxxxxdlp, line 29: `recordFetched` variable names makes it sounds like it either holds the fetched records, or a boolean indicating whether a record was fetched.

a better name would be `fetchedCount`

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,113 @@
...
+      attachments={attachments?.map(({ attachment }) => attachment) ?? []}
+      isComplete={recordFetched.current === records.length}
+      scale={scale}
+      onChange={(attachment, index): void =>
+        void attachments?.[index].related.set(`attachment`, attachment)
+      }
+      onFetchMore={
+        attachments === undefined || handleFetch === undefined || halt
+          ? undefined
+          : async () => handleFetch?.(recordFetched.current)
```
maxxxxxdlp, line 99: try our the impact of replacing `recordFetched` with `records.length` and see if everything still works correctly

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,113 @@
...
+          }}
+        >
+          {attachmentsText.fetchNextAttachments()}
+        </Button.Orange>
+      </div>
+    )
+  ) : (
+    <AttachmentGallery
+      attachments={attachments?.map(({ attachment }) => attachment) ?? []}
+      isComplete={recordFetched.current === records.length}
```
maxxxxxdlp, line 91: try our the impact of replacing `recordFetched` with `records.length` and see if everything still works correctly

in that case we might be able to get rid of `recordFetched` ref entirely

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,113 @@
...
+        )
+      );
+
+      const fetchCount = records.findIndex(
+        (record) => record?.populated !== true
+      );
+
+      recordFetched.current = fetchCount === -1 ? records.length : fetchCount;
+
+      return Promise.all(
```
maxxxxxdlp, line 50: Try out replacing

```ts
      return Promise.all(
        filterArray(relatedAttachementRecords.flat()).map(
          async (collectionObjectAttachment) => ({
            attachment: await collectionObjectAttachment
              .rgetPromise('attachment')
              .then((resource) => serializeResource(resource)),
            related: collectionObjectAttachment,
          })
        )
      );
```

with
```ts
      return filterArray(relatedAttachementRecords.flat()).map(
         (collectionObjectAttachment) => ({
          attachment: serializeResource(collectionObjectAttachment.getDependentResource('attachment'))
          related: collectionObjectAttachment,
        })
      );
```

and check if things still work correctly

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,113 @@
...
+              .then((resource) => serializeResource(resource)),
+            related: collectionObjectAttachment,
+          })
+        )
+      );
+    }, [records]),
+    true
+  );
+
+  const [haltValue, setHaltValue] = React.useState(300);
```
maxxxxxdlp, line 64: live a comment in the code explaining why `halt` was added and what it does

maxxxxxdlp: future you will thank you for that

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const children = halt ? (
+    haltValue === records.length ? (
+      <>{attachmentsText.noAttachments()}</>
+    ) : (
+      <div className="flex flex-col gap-4">
+        {attachmentsText.attachmentHaltLimit({ halt: haltValue })}
+        <Button.Orange
+          onClick={() => {
+            if (haltValue + 300 > records.length) {
```
maxxxxxdlp, line 77: replace this imperative code with declaritive:
```js
onClick={():void => setHaltValue(Math.min(haltValue+300, records.length))}
```

also can you please create a variable for 300 above or above the component. could call it `haltSize` or `haltIncrementSize` or something like that. it's considered a code smell to have random numbers in the code. See https://en.wikipedia.org/wiki/Magic_number_(programming)

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,113 @@
...
+            }
+          }}
+        >
+          {attachmentsText.fetchNextAttachments()}
+        </Button.Orange>
+      </div>
+    )
+  ) : (
+    <AttachmentGallery
+      attachments={attachments?.map(({ attachment }) => attachment) ?? []}
```
maxxxxxdlp, line 90: This line could cause performance problems when there are a lot of attachments. Instead of doing .map on every render, can you move it into the useAsyncState above?

i.e, up there instead of `return filterArray...` do:
```
const attachments = filterArray...;
return {
  attachments: attachments.map(({ attachment }) => attachment),
  related: attachments?.map(({ related }) => related)
};
```

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,113 @@
...
+    />
+  );
+
+  return (
+    <Dialog
+      buttons={<Button.DialogClose>{commonText.close()}</Button.DialogClose>}
+      header={attachmentsText.attachments()}
+      onClose={handleClose}
+    >
+      {children}
```
maxxxxxdlp, line 110: not a big deal, but do you know why you created a `children` variable, rather than directly putting the children code in here?

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,113 @@
...
+          ? undefined
+          : async () => handleFetch?.(recordFetched.current)
+      }
+    />
+  );
+
+  return (
+    <Dialog
+      buttons={<Button.DialogClose>{commonText.close()}</Button.DialogClose>}
+      header={attachmentsText.attachments()}
```
maxxxxxdlp, line 107: feature: display attachment count in here. think of how to indicate that not all attachments might have been fetched yet. or if you want to keep it simple just display the `records.length`, but only once all attachments are fetched.

consider using `commonText.countLine()` for this (see other usages in the code)

maxxxxxdlp (changes requested):

maxxxxxdlp:
To test:
(besides the obvious things)
test editing attachments and make sure the edits are preserved when you click "save" - it's possible that we have a bug where attachments would only be saved by the "save" button on the form if you edited attachments for the record that is currently displayed on the form

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
```diff
@@ -218,15 +227,25 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
...
+                <Button.Blue onClick={() => setAttachmentState(true)}>
+                  {attachmentsText.attachments()}
```
maxxxxxdlp, line 239: Make this button less prominent. notice that we don't have any buttons with text up in this area as space is precious, and clutter is discouraged. (specify network logo would be replaced by small badges too)

Replace Button.Blue with Button.icon icon="photos"

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

chanulee1 (changes requested):
Fetching more attachments results in obtrusive loading dialog.

https://user-images.githubusercontent.com/103299204/234928122-aac4538c-dde4-4b93-b367-606de4570e3d.mp4

https://coldfish-issue-2132.test.specifysystems.org/specify/view/collectionobject/162/?recordsetid=157

chanulee1:
Cannot edit attachments through attachment viewer gallery. Must navigate to the associated record and make edits there to see changes in attachment viewer gallery, and then reload to see changes in the record set.

https://user-images.githubusercontent.com/103299204/234935017-64001218-e7a7-4ea6-87fe-68bd9ff9aa16.mp4

https://coldfish-issue-2132.test.specifysystems.org/specify/view/collectionobject/162/?recordsetid=157

CarolineDenis:
> Fetching more attachments results in obtrusive loading dialog. Also, the number of attachments displayed at the top of the viewer seems to count the number of attachments currently fetched, not total number of attachments. Expected?
> 
>  Cm6o3fwtSF.mp4 
> 
> https://coldfish-issue-2132.test.specifysystems.org/specify/view/collectionobject/162/?recordsetid=157

For the count: it shows what is loaded already, so the behavior is normal. If you scroll down and wait a little you will see the total number be updated. 
Reading the loading gif, I agree that it's obtrusive. It will be replaced with a skeleton once the PR (https://github.com/specify/specify7/pull/3334) is merged

grantfitzsimmons:
> > Fetching more attachments results in obtrusive loading dialog. Also, the number of attachments displayed at the top of the viewer seems to count the number of attachments currently fetched, not total number of attachments. Expected?
> > Cm6o3fwtSF.mp4
> > https://coldfish-issue-2132.test.specifysystems.org/specify/view/collectionobject/162/?recordsetid=157
> 
> For the count: it shows what is loaded already, so the behavior is normal. If you scroll down and wait a little you will see the total number be updated. Reading the loading gif, I agree that it's obtrusive. It will be replaced with a skeleton once the PR (#3334) is merged

The loading dialog needs to be removed entirely. It should not interrupt the user's ability to view the images that are already loaded while the new ones are loaded in

chanulee1 (changes requested):
Fetching more attachment records takes away ability to view already loaded records, and takes user to the top of the attachment list.


https://user-images.githubusercontent.com/103299204/235256319-d97824ca-ad77-4ec8-bcd5-274e776ce6b3.mp4

chanulee1:
Not sure if this is an issue, but also noticed some flickering when first opening a record set and when more attachment records are being fetched. (Difficult to capture on video)

https://user-images.githubusercontent.com/103299204/235257631-f27070bd-d098-4964-ac0d-01571d82fba0.mp4

specifyweb/frontend/js_src/lib/components/Attachments/RecordSetAttachment.tsx
```diff
@@ -0,0 +1,165 @@
...
+      const newAttachments = {
+        attachments: attachements.map(({ attachment }) => attachment) as RA<
+          SerializedResource<Attachment>
+        >,
+        related: attachements.map(({ related }) => related) as RA<
+          SpecifyResource<CollectionObjectAttachment>
+        >,
+      };
+
+      attachmentsRef.current = newAttachments;
```
maxxxxxdlp, line 83: this can be simplified. just return new attachments like you did before, but right after the useEffect add this code:
```
if (typeof attachments === 'object') attachmentsRef.current = attachments;
```

maxxxxxdlp: Then, you don't need to define `currentAttachments` and can just use `attachmentsRef.current`

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

grantfitzsimmons (dismissed):
image

https://sp7demofish-issue-2132.test.specifysystems.org/specify/view/collectionobject/37484/?recordsetid=21

This is not very visually appealing. Several nested layers for a single tiny attachment thumbnail.

image

This is not good either while loading. Maybe a skeleton loader and fewer shadows would make this look better!

image

Could we also make this be 3-4 columns wide by default? 2 is a bit restrictive, but not a big deal.

image

We need to remove the shadow box behind the images!

## Example:

I still don't like how small the default thumbnails are, but here's what my proposed changes would look like:

image


image

specifyweb/frontend/js_src/lib/components/Attachments/Gallery.tsx
```diff
@@ -74,7 +74,7 @@ export function AttachmentGallery({
     <>
       <Container.Base
         className="grid flex-1 grid-cols-[repeat(auto-fit,minmax(var(--scale),1fr))]
-          items-center gap-4"
```
maxxxxxdlp, line 73: this shadow should be there when rendered as a standalone page. only remove shadow when in a dialog.

when making changes please be careful to look at all the usages of a component to check the reputation of a change

specifyweb/frontend/js_src/lib/components/Attachments/Preview.tsx
```diff
@@ -22,7 +22,7 @@ export function AttachmentPreview({
   return (
     <div
       className={`
-        flex items-center justify-center rounded bg-white shadow-xl
```
maxxxxxdlp, line 25: these changes affect both the dialog and the attachment tool. are you sure they are needed?
The bugs Grant reported are related to container shadow, not individual item shadow, right?

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Attachments/Gallery.tsx
maxxxxxdlp: Dismissing this as the change was intentional

specifyweb/frontend/js_src/lib/components/Attachments/Preview.tsx
maxxxxxdlp: Dismissing this as the change was intentional

maxxxxxdlp (approved):

chanulee1 (changes requested):
Visuals are looking good. But, fetching more records leads to many loading dialogs that flicker and can be distracting.

https://github.com/specify/specify7/assets/103299204/e61069b8-ad2b-496d-854f-03327e37ccb4

https://coldfish-issue-2132.test.specifysystems.org/specify/view/collectionobject/162/?recordsetid=157

CarolineDenis:
> Visuals are looking good. But, fetching more records leads to many loading dialogs that flicker and can be distracting.
> 
>  LauduwCbP4.mp4 
> 
> https://coldfish-issue-2132.test.specifysystems.org/specify/view/collectionobject/162/?recordsetid=157

Now that production as been merged in it, the loading gifs are replaced by skeletons. Can you let me know if it's better?

chanulee1 (changes requested):
Skeleton loader looks really nice. But fetching more records seems to fetch every remaining record. Also, the loading gifs appear right before records are finally loaded and noticed slight cutoff on the right side of the third column of attachments.

https://github.com/specify/specify7/assets/103299204/7150650b-f999-4830-a2ae-5d352c3aaf05

CarolineDenis:
@chanulee1: regarding your comment "noticed slight cutoff on the right side of the third column of attachments." 
Do you think you might have resized your dialog when testing this? I tried to reproduce it on my end and also launched the branch you created on the test panel but didn't have a cutoff. See recording below. 

https://github.com/specify/specify7/assets/108160931/8449b33b-eddd-4e1a-b161-e5392292e739

chanulee1:
@CarolineDenis I see. It seems like it's browser specific. I was able to reproduce it in `Chrome` but not in `Firefox`.

maxxxxxdlp:
Yes, the cut-off was likely caused by manually resizing the dialog.
The dialog sizes are stored in the browser, thus changing the browser, or accessing the same specify user from a different computer  would result in different behavior

chanulee1:
Hmm. The cutoff seems to happen for me no matter how it is resized. It looks like it is because of the thickness of the scrollbar? @CarolineDenis For some reason the scroll bar appears different in our videos. Not sure what determines this.

https://github.com/specify/specify7/assets/103299204/512f72a5-2020-4849-b561-0cd7c9a13cc0

melton-jason:
> Hmm. The cutoff seems to happen for me no matter how it is resized. It looks like it is because of the thickness of the scrollbar? @CarolineDenis For some reason the scroll bar appears different in our videos. Not sure what determines this.


As far as I know, by default the scrollbars are determined by your Operating System. 
I do not believe Specify has altered or changed the appearance of Scrollbars in any way, so it should be using the default Scrollbar for your OS.

pashiav (approved):
Looks good!


https://github.com/specify/specify7/assets/118556884/c04fefac-33a1-4429-bc2b-015e35da65b7

acwhite211 (approved):

PR #3364, opened 2023-04-18
Align notifications to the left

### LINKED ISSUE #3227: Stop centering text when it overflows
image

I am not a fan of centered text in general, but especially in menus that have a bunch of left-aligned text. This change was introduced recently (likely related to the changes for 400% zoom) and it has perturbed me every time it happens. It should remain left-aligned even on overfill.

CarolineDenis:
Fixes #3227

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
@@ -142,7 +142,7 @@ export function Notifications({
           'aria-live': 'polite',
           className:
             unreadCount > 0 && !isOpen
-              ? '[&:not(:hover)]:!text-brand-300 [&:not(:hover)]:dark:!text-brand-400'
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -142,7 +142,7 @@ export function Notifications({
           'aria-live': 'polite',
           className:
             unreadCount > 0 && !isOpen
-              ? '[&:not(:hover)]:!text-brand-300 [&:not(:hover)]:dark:!text-brand-400'
+              ? '[&:not(:hover)]:!text-brand-300 [&:not(:hover)]:dark:!text-brand-400 text-left'
```
maxxxxxdlp, line 145: The issue affects not just this place, but every link in specify.
See a better fix in fc00e2b34 and e9d2d2ea9

Since it's already fixed on agent-merging branch, we can close this pull request

maxxxxxdlp (changes requested):

PR #3367, opened 2023-04-18
Allow spaces in Lat Long boxes

### LINKED ISSUE #3354: Space characters cannot be typed in LatLonUI at the end
Our official documentation claims the following:

> The only characters that are allowed in the input field for coordinate fields are:
> 
> - A whitespace character (tab, space)
> - A digit
> - Any of the following: "'-.:ensw° (case-insensitive)

This was demonstrably true some time ago (and @maxxxxxdlp made changes to what was accepted the last time I tested it), but now spaces are not enterable at the end of the line.

https://user-images.githubusercontent.com/37256050/232167722-1323a8a2-8967-46c4-9c0d-5b1ebdb8e958.mov

In this video, I tried typing `30 30 30 N`, but I could only go back to add space characters.

https://sp7demofish.specifycloud.org/specify/view/locality/new/

CarolineDenis:
Fixes #3354

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
@@ -37,7 +37,8 @@ function Coordinate({
     useResourceValue(
       resource,
       schema.models.Locality.strictGetField(coordinateTextField),
-      undefined
+      undefined,
+      false
     );
   const isChanging = React.useRef<boolean>(false);
   React.useEffect(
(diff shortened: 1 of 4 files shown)


maxxxxxdlp (changes requested):

CarolineDenis:
@maxxxxxdlp:
No problem, I can add the trimStart.

Regarding the useResourceValue function that accepts the new prop trim (a boolean), if we remove it, users will no longer be able to enter spaces in the lat long fields. The useResourceView function calls parseValue, which has a default argument of trim: true. If we don't set trim to false, the value will always be trimmed in line 50 of parseValue: trim ? value.trim() : value.

maxxxxxdlp:
> @maxxxxxdlp: No problem, I can add the trimStart.
> 
> Regarding the useResourceValue function that accepts the new prop trim (a boolean), if we remove it, users will no longer be able to enter spaces in the lat long fields. The useResourceView function calls parseValue, which has a default argument of trim: true. If we don't set trim to false, the value will always be trimmed in line 50 of parseValue: trim ? value.trim() : value.

you are right. I am not sure why it still worked for me when I removed that

maxxxxxdlp (approved):

chanulee1:
Is this ready for UI review?

chanulee1 (approved):
Looks good. Able to enter spaces at the end of a string in Lat Long boxes.

PR #3370, opened 2023-04-18
Allow to change cat # after failure in loan dialog

### LINKED ISSUE #2280: Unable to create a loan from a set of catalog numbers after entering invalid cat #s
https://user-images.githubusercontent.com/37256050/193710054-f2bbe566-79da-4690-b3e2-ed117b08a47d.mov

1. Click on **Interactions**
2. Click **Loan**
3. Enter a catalog number not matching the collection format
5. Try to enter valid catalog numbers in the collection
6. See you are unable to create a loan

db: `herb_rbge`
user: `britchie`

CarolineDenis:
Fixes #2280

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
@@ -307,7 +307,9 @@ export function InteractionDialog({
                   disabled={catalogNumbers.length === 0}
                   onClick={(): void => handleProceed(undefined)}
                 >
-                  {commonText.next()}
+                  {state.type === 'PreparationSelectState' && state.problems
+                    ? commonText.update()
+                    : commonText.next()}
                 </Button.Blue>
               </div>
               {state.type === 'PreparationSelectState' &&
--- a/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
@@ -79,7 +79,18 @@ export function InteractionDialog({
         'PreparationSelectState',
         {
           readonly entries: Preparations;
-          readonly problems: IR<RA<string>>;
+        }
+      >
+    | State<
+        'InvalidState',
+        {
+          readonly invalid: RA<string>;
+        }
+      >
+    | State<
+        'MissingState',
+        {
+          readonly missing: RA<string>;
         }
       >
     | State<'LoanReturnDoneState', { readonly result: number }>
@@ -96,7 +107,9 @@ export function InteractionDialog({
   function handleProceed(
     recordSet: SerializedResource<RecordSet> | undefined
   ): void {
-    const items = catalogNumbers.split('\n');
+    const catalogNumbers = handleParse();
+    if (catalogNumbers === undefined) return undefined;
+
     if (model.name === 'Loan')
       loading(
         ajax<readonly [preprsReturned: number, loansClosed: number]>(
@@ -106,15 +119,15 @@ export function InteractionDialog({
             headers: { Accept: 'application/json' },
             body: {
               recordSetId: recordSet?.id ?? undefined,
-              loanNumbers: recordSet === undefined ? items : undefined,
+              loanNumbers: recordSet === undefined ? catalogNumbers : undefined,
             },
           }
-        ).then(({ data }) =>
+        ).then(({ data }) => {
           setState({
             type: 'LoanReturnDoneState',
             result: data[0],
-          })
-        )
+          });
+        })
       );
     else if (typeof recordSet === 'object')
       loading(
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
```diff
@@ -304,10 +304,7 @@ export function InteractionDialog({
               />
               <div>
                 <Button.Blue
-                  disabled={
-                    catalogNumbers.length === 0 ||
-                    inputRef.current?.validity.valid !== true
-                  }
+                  disabled={catalogNumbers.length === 0}
```
maxxxxxdlp, line 307: The bug occurred because I relied on a ref (`inputRef`) during render. You would think I would know better than to do that

Refs must never be used during render (or at least not without a very good reason). They should only be used inside of hooks. The reason is that changes to refs don't cause re-render, thus a stale value could be replaced

@CarolineDenis your change would make the button be enabled even if there are some invalid catalog numbers. @grantfitzsimmons is that the desired change?

grantfitzsimmons: I would think all errors should be resolved before proceeding– it would be bad if some catalog numbers were left out because it didn't warn them.

CarolineDenis:
The next button being enabled allows the check for the validity for the new entry. It will not send the user to the next page (showPrepSelectDlg) until there is no error. If error, the user will be blocked on it with a red background and a list of the problematics cat #. 
Maybe we could change the next button to 'check again' or something similar when there is wrong cat #. 
PS: the 'missing' title is changed in another branch by: 'No preparations were found for the following Catalog Numbers.'

https://user-images.githubusercontent.com/108160931/233166644-f5d8d182-d7a9-41cc-875c-cf7daef99cf0.mov

CarolineDenis:
https://user-images.githubusercontent.com/108160931/233169802-3c4fae0c-0d8c-4bca-8100-637ff5d91331.mov

maxxxxxdlp (approved):
The behavior is good by me. @grantfitzsimmons can you review if you are ok with it too?

grantfitzsimmons:
https://user-images.githubusercontent.com/37256050/233483656-c760fe88-9043-44b6-88c1-710b195907f4.mov

It reports catalog numbers in the field comma-separated as being entirely invalid instead of the specific one that is, which would be confusing to the user

grantfitzsimmons (changes requested):
See above ^

> It reports catalog numbers in the field comma-separated as being entirely invalid instead of the specific one that is, which would be confusing to the user

CarolineDenis:
@grantfitzsimmons I agree it's confusing. Note that its the behavior in place before this PR. We probably need to rethink this component

specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
```diff
@@ -106,15 +119,15 @@ export function InteractionDialog({
             headers: { Accept: 'application/json' },
             body: {
               recordSetId: recordSet?.id ?? undefined,
-              loanNumbers: recordSet === undefined ? items : undefined,
+              loanNumbers: recordSet === undefined ? catalogNumbers : undefined,
             },
           }
-        ).then(({ data }) =>
+        ).then(({ data }) => {
```
maxxxxxdlp, line 125: don't need `{` and `;` and `}` in here.
are those being added accidentally, or is it your preference to always add them?

I usually don't add them unless necessary as excessive curly braces make code refactoring harder (it's easy to miss one curly brace somewhere or add an extra one and suddenly you get a syntax error and don't know where)

specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
```diff
@@ -136,39 +149,40 @@ export function InteractionDialog({
...
-          while (++index < prepsData.length && prepsData[index][0] == value);
-        } else {
-          missing.push(entries[offsetIndex]);
-        }
-        offsetIndex += 1;
-      }
-      if (offsetIndex < entries.length)
-        missing = [...missing, ...entries.slice(offsetIndex)];
+
+    let catalogNumbers: WritableArray<string> = Array.from(
```
maxxxxxdlp, line 154: This code is looking a lot better than what was better but can be improved even further

think about how you could refactor it to not have to use `let` or `WritableArray` at all

specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
```diff
@@ -136,39 +149,40 @@ export function InteractionDialog({
...
+      const item = new itemCollection.model.specifyModel.Resource();
+      f.maybe(toTable(item, 'LoanPreparation'), (loanPreparation) => {
+        loanPreparation.set('quantityReturned', 0);
+        loanPreparation.set('quantityResolved', 0);
+      });
+      itemCollection.add(item);
+      handleClose();
+      return;
+    }
+    if (missing.length > 0) {
```
maxxxxxdlp, line 178: don't need `{` and `}` here

specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
```diff
@@ -273,74 +303,38 @@ export function InteractionDialog({
...
               />
               <div>
                 <Button.Blue
-                  disabled={
-                    catalogNumbers.length === 0 ||
-                    inputRef.current?.validity.valid !== true
-                  }
-                  onClick={(): void => handleProceed(undefined)}
+                  disabled={catalogNumbers.length === 0}
+                  onClick={(): void => {
```
maxxxxxdlp, line 312: don't need `{` and `}`

maxxxxxdlp (changes requested):
minor changes

maxxxxxdlp (approved):

PR #3371, opened 2023-04-19
Improve text in loan dialog and disable ignore button when nothing found

### LINKED ISSUE #2881: Improve the "Create New Loan" dialog
> ## New Loan Dialogue
> 
> In Specify 7, When creating a new Loan, the following steps are taken.
> 
> Interactions > Loan > Create Loan popup with opens to choose recordset of enter Catalog Numbers.
> 
> image
> 
> If Catalog Numbers are added.
> image
> 
> And then ‘**Next**’ is selected, the following result appears.
> image
> 
> From the text following ‘**Next**’, there is a statement ‘There are problems with the entry:’ and then on the following line, ‘_Missing_:’
> This statement seems to imply that the Catalog Numbers themselves are missing, which is misleading.
> The following statement, that ‘_No preparations were found_’ is much more informative in this case.
> 
> From the previous popup, if ‘**Next**’ is selected, nothing happens. If ‘**Ignore**’ is selected, the following popup appears:
> 
> image
> 
> In this case, there are no preparations to select from, so this popup is a deadend. The only functional option from this popup is ‘Cancel’.
> 
> # Suggested Improvement
> 
> If no preparations are found, it would be helpful to replace the ‘Missing:’ text with ‘No preparations were found:’ for the following Catalog Numbers, and omit the ‘Ignore’ option. 

**Requested By:** Garth at The University of Michigan

Please let me know if you have any questions so I can pass them along.

CarolineDenis:
Fixes #2881

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
@@ -315,7 +315,10 @@ export function InteractionDialog({
                     catalogNumbers.length === 0 ||
                     inputRef.current?.validity.valid !== true
                   }
-                  onClick={(): void => handleProceed(undefined)}
+                  onClick={(): void => {
+                    handleProceed(undefined);
+                    setDisableIgnore(false);
+                  }}
                 >
                   {commonText.next()}
                 </Button.Blue>
--- a/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
@@ -339,13 +339,13 @@ export function InteractionDialog({
                   )}
                   <div>
                     <Button.Blue
+                      disabled={disableIgnore}
                       onClick={(): void =>
                         setState({
                           ...state,
                           problems: {},
                         })
                       }
-                      disabled={disableIgnore}
                     >
                       {commonText.ignore()}
                     </Button.Blue>
--- a/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
@@ -86,6 +86,8 @@ export function InteractionDialog({
     | State<'MainState'>
   >({ type: 'MainState' });
 
+  const [disableIgnore, setDisableIgnore] = React.useState(false);
+
   const { parser, split, attributes } = useParser(searchField);
   const { validationRef, inputRef, setValidation } =
     useValidation<HTMLTextAreaElement>();
@@ -161,7 +163,10 @@ export function InteractionDialog({
           loanPreparation.set('quantityResolved', 0);
         });
         itemCollection.add(item);
-      } else showPrepSelectDlg(prepsData, formatProblems(prepsData, missing));
+      } else {
+        showPrepSelectDlg(prepsData, formatProblems(prepsData, missing));
+        setDisableIgnore(true);
+      }
     } else showPrepSelectDlg(prepsData, {});
   }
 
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/localization/interactions.ts
```diff
@@ -37,20 +37,20 @@ export const interactionsText = createDictionary({
...
-  },
   preparationsNotFound: {
     'en-us': 'No preparations were found.',
     'ru-ru': 'Никаких препаратов не обнаружено.',
     'es-es': 'No se encontraron preparados.',
     'fr-fr': "Aucune préparation n'a été trouvée.",
     'uk-ua': 'Препаратів не виявлено.',
   },
+  preparationsNotFoundFor: {
+    'en-us': 'No preparations were found for the following Catalog Numbers.',
```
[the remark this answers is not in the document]
maxxxxxdlp: sorry for not being clear. I meant you need to replace
```
No preparations were found for the following Catalog Numbers.
```
with
```
No preparations were found for the following Catalog Numbers:
```

notice the `:` (colon) at the end instead of period

maxxxxxdlp (changes requested):

CarolineDenis:
PS: the disableIgnore back to true will only work once merge with this branch(https://github.com/specify/specify7/pull/3370) 
When the next button is unable after entering wrong data.

specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
maxxxxxdlp, line 1: (optional)

Regarding the `// This is a really ugly piece of code:` comment, ChatGPT suggested the following (which can be cleaned up still more and have types added and jest unit tests too) if you are interested.

The ideal way to do it is you encapsulate the logic into a function export it, write tests for it. Then refactor the code in the function to be not ugly, and verify that all the tests still pass - as a result, the code is nice, we know that behavior is correct and we have tests for the future

```js
const findNextIndex = (prepsData, value, startIndex) => {
  const index = prepsData.findIndex(
    (element, idx) => element[0] === value && idx >= startIndex
  );
  return index === -1 ? prepsData.length : index;
};

const findMissingEntries = (entries, prepsData) => {
  if (!Array.isArray(entries)) return [];

  let index = 0;
  return entries.filter((entry) => {
    const newIndex = findNextIndex(prepsData, entry, index);
    if (newIndex === prepsData.length) return true;

    index = newIndex + 1;
    return false;
  });
};

let missing = findMissingEntries(entries, prepsData);
```

specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
```diff
@@ -161,7 +163,10 @@ export function InteractionDialog({
           loanPreparation.set('quantityResolved', 0);
         });
         itemCollection.add(item);
```
maxxxxxdlp, line 165: There looks to be a bug in here (was present before). When the code follows this branch the dialog is not closed

Not sure if that's related to https://github.com/specify/specify7/pull/3389/files

maxxxxxdlp: If you are interested, here is how that code looked like before I rewrote it to typescript and react - https://github.com/specify/specify7/blob/b5bde53cb6a3178212752b2e4e31949888778cd8/specifyweb/frontend/js_src/lib/interactiondialog.js#L288-L298

specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
```diff
@@ -329,6 +339,7 @@ export function InteractionDialog({
                   )}
                   <div>
                     <Button.Blue
+                      disabled={disableIgnore}
```
maxxxxxdlp, line 342: Can you please clarify the behavior you are going for (under which conditions the button should be enabled or disabled). It would also be useful for testers.
Feel free to double-check with Grant for what's expected.
Looking at the code, it looks like `setDisableIgnore(true);` is called if there are any missing items. It is not called if no preparations where found.

maxxxxxdlp (changes requested):
I am confused by the intended behavior here. Not your fault

CarolineDenis:
Fixed in this PR (https://github.com/specify/specify7/pull/3371)

PR #3390, opened 2023-04-20
Fix a maximum height for attachment image preview

CarolineDenis:


specifyweb/frontend/js_src/lib/components/Attachments/Viewer.tsx
```diff
@@ -98,7 +98,7 @@ export function AttachmentViewer({
         {originalUrl === undefined ? (
           loadingGif
         ) : type === 'image' ? (
-          <img alt={title} src={originalUrl} />
+          <img alt={title} className="max-h-[70vh]" src={originalUrl} />
```
maxxxxxdlp, line 101: This is not a great fix. Please avoid hardcoded sizing whenever possible.
A nice fix would make the image occupy the full height of the container, regardless of container size (your fix assumes the container is always near the full screen in size)

Here is a preview of what desired behavior may look like - try making it look like that (except for the image dimensions distortion - which can be fixed using object-fit CSS property) - not showing you the code to leave it as an exercise

https://user-images.githubusercontent.com/40512816/233509885-7c0d0f31-72e0-451d-ae88-2180b8c39057.mp4

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Attachments/Viewer.tsx
```diff
@@ -94,11 +94,15 @@ export function AttachmentViewer({
   const table = f.maybe(serialized.tableID ?? undefined, getAttachmentTable);
   return (
     <div className="flex h-full gap-8">
-      <div className="flex min-h-[30vw] w-full min-w-[30vh] flex-1 items-center">
+      <div className="flex max-h-screen min-h-[30vw] w-full min-w-[30vh] flex-1 items-center">
```
maxxxxxdlp, line 97: the image sizing behaves nicely now, but the image is still not horizontally centered

maxxxxxdlp: also a separate improvement:
`min-h-[30vw] min-w-[30vh]` is fine on a regular screen, but if you zome out a lot (to simulate a large screen), 30% of the screen is way too much.
Can you try out some rem-based value instead?
From quick testing, `min-w-[theme(spacing.60)] min-h-[theme(spacing.60)]` seems to work, but you should try and see what works best. 
Example way to test this: upload a pdf and shrink dialog size and make sure the preview does not shrink beyond so much that it's not useful anymore (could also try this when the attachment plugin is on the form)

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

chanulee1:
How to test this?

grantfitzsimmons (changes requested):
image

This is not working

https://sp7demofish-attachment-size.test.specifysystems.org/specify/view/collectionobject/51737/

CarolineDenis:
Seems to be an issue with the dialog size?

https://user-images.githubusercontent.com/108160931/234895578-d44d95fb-9996-4d47-a63c-c7fefa18399a.mov

bronwyncombs (approved):
Looks great 😄 

Screenshot 2023-08-03 at 11 17 01 AM
Screenshot 2023-08-03 at 11 22 23 AM
Screenshot 2023-08-03 at 11 20 24 AM

Issue #3406, opened 2023-04-24
Reword relative dates


grantfitzsimmons:
<img width="842" alt="image" src="https://user-images.githubusercontent.com/37256050/234093754-8abc2a74-aed0-44eb-b238-fff3934a18ac.png">

This should be `13` `Days` in the `Future` instead of `Future` `13` `Day`.

This is the same comment I've made several times, but I think it's worth fixing. It currently looks like you are specifying a range (in my opinion).

PR #3408, opened 2023-04-24
Reorder relative dates

grantfitzsimmons:
Solves #3406


maxxxxxdlp (approved):
Code looks good. But before we merge this, @CarolineDenis can you take a look at making the order of the 3 boxes be customizable by translators?

See example usage at https://github.com/specify/specify7/blob/51ea5579edffeb1c36edb93612efe140f5046c98/specifyweb/frontend/js_src/lib/components/AppResources/Aside.tsx#L227-L235 and https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/localization/README.md#syntax-for-string-parameters-and-plurals

grantfitzsimmons:
> Code looks good. But before we merge this, @CarolineDenis can you take a look at making the order of the 3 boxes be customizable by translators?

Since we would like to release in the next day or so, can we spin this off into another issue if @CarolineDenis does not have time before release?

CarolineDenis:
> > Code looks good. But before we merge this, @CarolineDenis can you take a look at making the order of the 3 boxes be customizable by translators?
> 
> Since we would like to release in the next day or so, can we spin this off into another issue if @CarolineDenis does not have time before release?

yes let's do that

grantfitzsimmons:
Resolve in https://github.com/specify/specify7/issues/3421

PR #3283, opened 2023-03-31
Fix query combo box not scoping created resources to other collection

maxpatiiuk:
Fixes https://github.com/specify/specify7/issues/2405
Fixes https://github.com/specify/specify7/issues/3078
Also, allows unhiding all fields in autogenerated forms (in the meta menu when you say "use autogenerated form", there is now also a new checkbox)

Theresa was correct - there is a generic solution to this to fix it in all places

PR #3431, opened 2023-04-28
Restrict creating Host Taxons

melton-jason:
This restriction needs to be put into place until a proper fix for the Host Taxon relationship is implemented

The issue is that any Host Taxon uploaded through the WorkBench or created using the `+` button on the QueryCombobox will always be created in the current discipline. 

This was solved by removing the HostTaxon relationship from CollectingEventAtrribute in the Workbench and removing the `+` button for the HostTaxon QueryCombobox when the rightSideCollection as specified in the form definition is different then the current collection

Related: https://github.com/specify/specify7/pull/3240#issuecomment-1492001799, https://github.com/specify/specify7/issues/2675

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/HostTaxon.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/HostTaxon.tsx
@@ -44,19 +44,19 @@ export function HostTaxon({
               .maybe(records[0], deserializeResource)
               ?.rgetPromise('rightSideCollection')
           )
-          .then((collection) => collection?.get('id')),
+          .then((collection) => collection),
       [relationship]
     ),
     false
   );
-  return rightSideCollection === undefined ? (
+  return rightSideCollection === null || rightSideCollection === undefined ? (
     <Input.Text isReadOnly />
   ) : hasTreeAccess('Taxon', 'read') ? (
     <QueryComboBox
       field={schema.models.CollectingEventAttribute.strictGetRelationship(
         'hostTaxon'
       )}
-      forceCollection={rightSideCollection}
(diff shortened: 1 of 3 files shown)


specifyweb/frontend/js_src/lib/components/FormPlugins/HostTaxon.tsx
```diff
@@ -44,26 +44,29 @@ export function HostTaxon({
...
-      forceCollection={rightSideCollection}
+      forceCollection={rightSideCollection.get('id')}
       formType={formType}
       id={id}
       isRequired={isRequired}
       mode={mode}
       relatedModel={schema.models.Taxon}
       resource={resource}
       typeSearch={hostTaxonTypeSearch}
+      disableCanAdd={
```
maxxxxxdlp, line 67: This is not a good fix.
The query combo box receives `forceCollection` prop, which it can use to figure out if current collection is equal to `forceCollection`

In fact, this fix is already implemented. See https://github.com/specify/specify7/pull/3283

In there, I added support for properly creating resource into a correct discipline: https://github.com/specify/specify7/pull/3283/files#diff-11908f1e0fbcca3761cf5d0d0635a57bbedbb34f38675283d64fa843a9af23f0R263-R268

But also, it disables the add feature (temporary). It's disabled only because tree rank pick list is loading data from a wrong discipline:
https://github.com/specify/specify7/pull/3283/files#diff-11908f1e0fbcca3761cf5d0d0635a57bbedbb34f38675283d64fa843a9af23f0R359-R364

Can you please review the code change on that pull request, and undo the query combo box changes on this branch

maxpatiiuk (changes requested):

PR #3125, opened 2023-03-08
Allow independent subviews

maxpatiiuk:
Fixes #114
Fixes #3127

but only if sub view is displayed as a button (because button opens a dialog, and dialog has a save button, and dialog creates a separate `<form>` - thus validation errors in independent sub-view are scoped to that sub-view)

This is just a one-line change because 50% of the work was done in 7.7
Now, the other 50% is testing - **please test the hell out of this**

Suggestions for testing:
- Collection Object -> Accession as a subview button
- Collecting Event -> Collection as a subview
   - try opening Collection Object -> Collecting Event -> Collection Object that way
- Save main form before saving sub view
   - Especially when sub view has validation errors
- Also do testing in record sets (especially with saving and validation errors and going between record set records)
- Test #3127

PR #3441, opened 2023-05-03
Add prepType to query table to edit them in same fictitious record set

### LINKED ISSUE #2754: Introduce visual "PrepType" editor
In Specify 7, the best way to modify prep type options is to create a query on the `preptype` table, create a record set, then modify the options within the record set.

There should be an automatic way of doing this, potentially automatically creating an invisible record set of all `preptype` records and allowing the user to use this interface to modify the available options.

image

CarolineDenis:
Fixes #2754


maxxxxxdlp (approved):

grantfitzsimmons (changes requested):
1. The Preparation Type table is sorted at the bottom rather than where it should be alphabetically. Not sure why this is.

    image


    image

2. How is this any different from default behavior? This is still not satisfactory & does not really make it any easier to create new prep types

    image

    This kind of thing really needs its own UI (as it has in 6) where the user doesn't _feel_ like it's detached from the schema configuration process. It should appear like this:
   
    image


     image

    Where you can add new prep types from the dialog & this should be accessible automatically when editing the preparation type picklist in the schema configuration tool.

CarolineDenis:
Screenshot 2023-05-17 at 11 56 29 AM
What about displaying a PreType form with a slider that allows to navigate between preType and + / - buttons in the edit pick list dialog when Entire Table is selected?

CarolineDenis:
@maxxxxxdlp 
When you have time, could you please look into the following issues:
- There is a TypeScript error attached to'resource' line 197 in FormCells/index.tsx 
- in the subview, when 'entire table' is selected, the add button currently opens a query dialog. Instead, it should display a new empty form, replacing the last opened resource. Additionally, the number in the slider should increase . 
- when selecting 'institution' in the table name, a error dialog appears 
- when selecting 'project' in table name, nothing is displayed

maxxxxxdlp:
> There is a TypeScript error attached to 'resource' line 197 in FormCells/index.tsx

Fixed. See c1b084aae

> when selecting 'institution' in the table name, a error dialog appears

We didn't use domain filter correctly. See the fix in 19666c9ab

> in the subview, when 'entire table' is selected, the add button currently opens a query dialog. Instead, it should display a new empty form, replacing the last opened resource. Additionally, the number in the slider should increase

This one is quite tricky. Fortunately, it's either already fixed by https://github.com/specify/specify7/pull/3125, or needs to be fixed on that branch anyway. Basically, we need to allow for use case of creating independent records from inside of a subview, rather than just linking to existing independent records. Thus, it would be simplest to just make this PR wait till that PR is fixed

> when selecting 'project' in table name, nothing is displayed

You were correct - it's simply because there are no projects in the kufish database

maxxxxxdlp:
> If the feature in readOnly is satisfying we can merge this PR. The add/edit feature will be implemented with #3125

Could you please open a feature request for that so that we don't forget about it?

grantfitzsimmons (approved):
@CarolineDenis looks good! 

I am sad only because we cannot add new prep type records from the form, but I know it is due to the issues with independent subviews.

grantfitzsimmons (approved):
@CarolineDenis looks good! 

I am sad only because we cannot add new prep type records from the form, but I know it is due to the issues with independent subviews.

PR #3454, opened 2023-05-04
Allow to collapse long query line

### LINKED ISSUE #2479: Collapse long query lines
Like this:

Screenshot 2022-11-17 at 13 37 00

Clicking on the dots should expand the path

CarolineDenis:
Fixes #2479

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
@@ -231,6 +231,8 @@ export function QueryLine({
 
   const hasAny = field.filters.some(({ type }) => type === 'any');
 
+  const [collapse, setCollapse] = React.useState(false);
+
   return (
     <div
       aria-current={isFocused ? 'location' : undefined}
@@ -284,12 +286,32 @@ export function QueryLine({
             field.filters.length > 1 ? 'flex flex-wrap gap-2' : 'contents'
           }
         >
-          {join(
-            mappingLineProps.map((mappingDetails) => (
-              <MappingElement {...mappingDetails} role="listitem" />
-            )),
-            mappingElementDivider
-          )}
+          {collapse === true
+            ? join(
+                mappingLineProps
+                  .filter(
+                    (_, index) =>
+                      index === 0 ||
+                      index === mappingLineProps.length - 1 ||
+                      index === 1
+                  )
+                  .map((mappingDetails, index) =>
+                    index === 1 ? (
+                      <Button.Small onClick={() => setCollapse(!collapse)}>
+                        {icons.dotsVertical}
+                      </Button.Small>
+                    ) : (
+                      <MappingElement {...mappingDetails} role="listitem" />
+                    )
+                  ),
+                mappingElementDivider
+              )
+            : join(
+                mappingLineProps.map((mappingDetails) => (
+                  <MappingElement {...mappingDetails} role="listitem" />
+                )),
+                mappingElementDivider
+              )}
         </div>
         {filtersVisible && (
           <div
@@ -474,6 +496,9 @@ export function QueryLine({
             {icons.locationMarker}
           </Button.Small>
         ) : undefined}
+        <Button.Small onClick={() => setCollapse(!collapse)}>
+          {icons.minus}
+        </Button.Small>
         <Button.Small
           aria-label={queryText.showButtonDescription()}
           aria-pressed={field.isDisplay}


specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -247,6 +247,14 @@ export function QueryBuilder({
   );
   const resultsShown = state.queryRunCount !== 0;
 
+  const [collapseAllPref] = userPreferences.use(
+    'queryBuilder',
+    'behavior',
+    'collapseQueryLine'
+  );
+  const [isAllCollapsed, setIsAllCollapsed] =
+    React.useState<boolean>(collapseAllPref);
```
maxxxxxdlp, line 256: Don't need to specify `<boolean>` - it can infer that automatically

maxxxxxdlp, line 256: (optional) Consider getting rid of React.useState and using only `userPreferences.use`
Benefits:
1. the value will be preserved between sessions (consider if this is desirable behavior in this case)
2. you won't have to pass the state down to other components - other components can just call the `userPreferences.use` once again (any changed to the same `userPreferences.use` in one component is synced with all other components)

specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx
```diff
@@ -116,6 +118,9 @@ export function QueryHeader({
         </ProtectedAction>
       )}
       <div className="flex flex-wrap justify-center gap-2">
+        <Button.Small onClick={() => setIsAllCollapsed(!isAllCollapsed)}>
```
maxxxxxdlp, line 121: Add `aria-pressed={isAllCollapsed}` attribute. besides accessibility, it will also provide styling

specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
```diff
@@ -230,6 +232,11 @@ export function QueryLine({
     availableFilters.length > 1 || availableFilters[0][0] !== 'any';
 
   const hasAny = field.filters.some(({ type }) => type === 'any');
+  const [collapse, setCollapse] = React.useState(true);
+  React.useEffect(() => {
+    if (isAllCollapsed === undefined) return;
+    setCollapse(isAllCollapsed);
+  }, [isAllCollapsed]);
```
maxxxxxdlp, line 239: A shorter way to write this logic:
```
const [collapse, setCollapse] = useTriggerState(isAllCollapsed);
```

Take a look at the description of that hook to see what it does. If you have time, please also look over other hooks in that directory as they all are quite useful in some cases

specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
```diff
@@ -230,6 +232,11 @@ export function QueryLine({
     availableFilters.length > 1 || availableFilters[0][0] !== 'any';
 
   const hasAny = field.filters.some(({ type }) => type === 'any');
+  const [collapse, setCollapse] = React.useState(true);
```
maxxxxxdlp, line 235: rename `collapse` to `isCollapsed` or something like that - that's a more appropriate name for a boolean

`collapse` makes it seem like it's a function that will collapse when called rather than a boolean indicating whether we are collapsed right now

Thus, generally it's a good practice to make boolean variable names start with `is`

specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
```diff
@@ -284,12 +291,36 @@ export function QueryLine({
...
+            ? join(
+                mappingLineProps
+                  .filter(
+                    (_, index) =>
+                      index === 0 ||
+                      index === mappingLineProps.length - 1 ||
+                      index === 1
+                  )
+                  .map((mappingDetails, index) =>
+                    index === 1 ? (
```
maxxxxxdlp, line 304: if there are only two boxes total, it should not collapse

specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
```diff
@@ -284,12 +291,36 @@ export function QueryLine({
...
+                        }}
+                      >
+                        {icons.dotsHorizontal}
+                      </Button.Small>
+                    ) : (
+                      <MappingElement {...mappingDetails} role="listitem" />
+                    )
+                  ),
+                mappingElementDivider
+              )
```
maxxxxxdlp, line 317: A more readable way to write all of this:
```tsx
{
  collapse && mappingLineProps.length > 2
    ? <>
        <MappingElement {...mappingLineProps[0]} role="listitem" />
        {mappingElementDivider}
        <Button.Small onClick={() => setCollapse(!collapse) }>{icons.dotsHorizontal}</Button.Small>
        {mappingElementDivider}
        <MappingElement {...mappingLineProps.at(-1)!} role="listitem" />
      </>
    : ...
}
```

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

grantfitzsimmons (changes requested):
@CarolineDenis 

1. You did this perfectly! Just how the issue described! :smile:
2. I don't like the initial idea very much as it is.

Too much context is lost, and it probably shouldn't be collapsed by default.

**Here's an idea I mocked up–**

Table icons for the last mapped field should be used instead of a check mark (which communicates nothing) when collapsed.

image

This is how it looks currently:
image

maxxxxxdlp:
@grantfitzsimmons how about when collapsed we just display the last 2 fields instead of last 1?

grantfitzsimmons:
https://github.com/specify/specify7/assets/37256050/6cebe6e4-2f7c-4c8d-93e0-e1a6925d5655

Also this button should be gray at all times

maxxxxxdlp:
@grantfitzsimmons table icon is only used if it's a relationship everywhere across specify. might not be easy to modify it to show table icons for fields

grantfitzsimmons:
> @grantfitzsimmons how about when collapsed we just display the last 2 fields instead of last 1?

I can live with this :smile:

maxxxxxdlp:
> Also this button should be gray at all times

@grantfitzsimmons why?
This indicates the current state of the button - whether it's pressed or not. we use the same in many places. i.e, "Tools" button in the workbench

grantfitzsimmons:
> @grantfitzsimmons why?
> This indicates the current state of the button - whether it's pressed or not. we use the same in many places. i.e, "Tools" button in the workbench

I see! If the button had static text (like a real button) I would understand more intuitively that it should show itself as "pressed in", but the text changes, which makes it do a different thing the next time its pressed. This makes me think it should just be the same color. 

Just my thoughts on it. Not a big deal

maxxxxxdlp:
@CarolineDenis Summary:
- Remove aria-pressed from button
- Display last two boxes when collapsed

grantfitzsimmons (changes requested):
https://github.com/specify/specify7/assets/37256050/6d909696-9b81-4134-b7d8-7aefa9e86a3c

Once again, not a criticism of execution, but a question of usefulness/confusion.

For most of these items, collapsing them did not take up any less space but it does add confusion and complexity for the user. 

From a user perspective, these questions arise:
"Why would some items be hidden if there is room for them?"
"Why is there a new button to unhide one or two items?"

Keep in mind many users do not know what a data model is or even that there is a relational database underlying their data.

On top of this, the recent changes make it look like there are collapsed items when they do not exist:
https://github.com/specify/specify7/assets/37256050/ecf4f264-5d59-4a2a-ae5b-93d74ec92c29

I would reccomend either implementing the table icon next to the field name with one item (as it was in my previous concept) or we should likely stop developing this PR without feedback from a group of users/staff.

maxxxxxdlp:
@grantfitzsimmons agree with your criticism
Let's discuss the usefulness of this feature with the group tomorrow
I belive this was Andy's proposal

maxxxxxdlp:
@CarolineDenis sorry about this.
This is a common part of the process and when you take on larger tasks, you will see this type of thing happen more
One way to minimize it is to use Figma to create a mock-up of how the feature will look and work, before writing any actual code, and then discuss the prototype. Once all ideas/criticisms have been factored in, you can begin to turn the Figma mock-up into a actual code - at that point, few design decisions will remain, and you would be able to focus more on the technical decisions

grantfitzsimmons (changes requested):
1. Since the basic view is not flexible, when it occupies half the screen I can't see the operators anymore.

Screenshot 2023-05-26 at 12 35 24 PM

2. This text should not be bold, it should be regular weight. Text is rarely bold anywhere in the application.

Screenshot 2023-05-26 at 12 36 02 PM

3. When dragging an item, you have to select the field name. On top of that, you lose the icon when it is being dragged.


https://github.com/specify/specify7/assets/37256050/bb9b863f-3efa-4a52-86f3-f8c584716e51

4. When adding an "OR", it breaks the entire query

Screenshot 2023-05-26 at 12 37 34 PM

https://github.com/specify/specify7/assets/37256050/10340050-6ed5-4158-a250-2503e5d23fc0

5. Actually running the query results in errors.

```
KeyError at /stored_query/ephemeral/ 'field' Request Method: POST Request URL: http://coldfish-issue-2479.test.specifysystems.org/stored_query/ephemeral/
```

[Specify 7 Crash Report - 2023-05-26T17_39_18.202Z.txt](https://github.com/specify/specify7/files/11578532/Specify.7.Crash.Report.-.2023-05-26T17_39_18.202Z.txt)
[Specify 7 Crash Report - 2023-05-26T17_39_27.051Z.txt](https://github.com/specify/specify7/files/11578533/Specify.7.Crash.Report.-.2023-05-26T17_39_27.051Z.txt)

Screenshot 2023-05-26 at 12 38 27 PM

carlosmbe:
Extra padding when only a few items are being displayed in the basic view:

Screenshot 2023-05-26 at 12 38 04 PM
Screenshot 2023-05-26 at 12 39 52 PM

carlosmbe:
State not being preserved after saving a new query.
I.E: I create a new query and I'm in basic view throughout the entire process. After saving it, detailed view is applied.


https://github.com/specify/specify7/assets/53784701/f156a2ab-4c2e-4336-a2ba-3c0b43bf720f

specifyweb/frontend/js_src/lib/components/Core/Contexts.tsx
```diff
@@ -10,7 +10,7 @@ import { ErrorBoundary } from '../Errors/ErrorBoundary';
 import { loadingBar } from '../Molecules';
 import { Dialog, dialogClassNames, LoadingScreen } from '../Molecules/Dialog';
 import { TooltipManager } from '../Molecules/Tooltips';
-import { ReportEventHandler } from '../Reports/Context';
+// import { ReportEventHandler } from '../Reports/Context';
```
maxxxxxdlp: remove

specifyweb/frontend/js_src/lib/components/QueryBuilder/FieldFilter.tsx
```diff
@@ -199,7 +202,7 @@ export function QueryInputField({
       className={`
         relative min-w-[theme(spacing.40)] after:invisible
         after:block after:px-2 after:leading-[0px] after:content-[attr(data-value)]
-      `}
+      ${isBasic ? 'flex-1' : ''}`}
```
maxxxxxdlp, line 205: can you indent this correctly?
(indent by 1 to match the lines before it, and also put a new line:
```ts
    ${isBasic ? 'flex-1' : ''}
  `}
```

specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
```diff
@@ -163,18 +162,27 @@ export function QueryFields({
     }, [fields.length])
   );
 
+  const isBasic = React.useContext(IsQueryBasicContext);
+
   return (
-    <Ul className="flex-1 overflow-y-auto" forwardRef={fieldsContainerRef}>
+    <Ul
+      className={`flex-1 items-center overflow-y-auto
```
maxxxxxdlp, line 169: Could you please indent this properly? A new line after first \` and before last \`

specifyweb/frontend/js_src/lib/components/Atoms/Icons.tsx
```diff
@@ -70,6 +70,9 @@ export const icons = {
...
   documentReport: <svg aria-hidden className={iconClassName} fill="currentColor" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><path clipRule="evenodd" d="M6 2a2 2 0 00-2 2v12a2 2 0 002 2h8a2 2 0 002-2V7.414A2 2 0 0015.414 6L12 2.586A2 2 0 0010.586 2H6zm2 10a1 1 0 10-2 0v3a1 1 0 102 0v-3zm2-3a1 1 0 011 1v5a1 1 0 11-2 0v-5a1 1 0 011-1zm4-1a1 1 0 10-2 0v7a1 1 0 102 0V8z" fillRule="evenodd" /></svg>,
   documentSearch: <svg aria-hidden className={iconClassName} fill="currentColor" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg"><path d="M4 4a2 2 0 012-2h4.586A2 2 0 0112 2.586L15.414 6A2 2 0 0116 7.414V16a2 2 0 01-2 2h-1.528A6 6 0 004 9.528V4z" /><path clipRule="evenodd" d="M8 10a4 4 0 00-3.446 6.032l-1.261 1.26a1 1 0 101.414 1.415l1.261-1.261A4 4 0 108 10zm-2 4a2 2 0 114 0 2 2 0 01-4 0z" fillRule="evenodd" /></svg>,
+  dotsHorizontal:<svg aria-hidden className={iconClassName} fill="currentColor" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg">
```
maxxxxxdlp, line 73: is this still used or can be removed?

specifyweb/frontend/js_src/lib/components/Core/Contexts.tsx
```diff
@@ -10,7 +10,7 @@ import { ErrorBoundary } from '../Errors/ErrorBoundary';
 import { loadingBar } from '../Molecules';
 import { Dialog, dialogClassNames, LoadingScreen } from '../Molecules/Dialog';
 import { TooltipManager } from '../Molecules/Tooltips';
-import { ReportEventHandler } from '../Reports/Context';
+// Import { ReportEventHandler } from '../Reports/Context';
```
maxxxxxdlp, line 13: remove

specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
```diff
@@ -231,10 +239,19 @@ export function QueryLine({
 
   const hasAny = field.filters.some(({ type }) => type === 'any');
 
+  const fieldSpec = QueryFieldSpec.fromPath(baseTableName, field.mappingPath);
```
maxxxxxdlp, line 242: This might be expensive to run on every render (especially since there can be many QueryLine renders at once
Refactor to use React.useMemo

specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
```diff
@@ -231,10 +239,19 @@ export function QueryLine({
...
+    fieldSpec.toMappingPath()
+  );
+
+  const isBasic = React.useContext(IsQueryBasicContext);
+
   return (
     <div
       aria-current={isFocused ? 'location' : undefined}
-      className="flex flex-1 gap-2 border-t border-t-gray-500 bg-[color:var(--form-foreground)] py-2"
+      className={`flex flex-1 gap-2 border-t border-t-gray-500 bg-[color:var(--form-foreground)] py-2
```
maxxxxxdlp, line 253: please insert newlines after first \` and before last \` to make it more readable

specifyweb/frontend/js_src/lib/components/QueryBuilder/QueryLineTools.tsx
```diff
@@ -0,0 +1,113 @@
...
+  readonly field: QueryField;
+  readonly isFieldComplete: boolean;
+  readonly onChange: ((newField: QueryField) => void) | undefined;
+  readonly onMoveUp: (() => void) | undefined;
+  readonly onMoveDown: (() => void) | undefined;
+}): JSX.Element {
+  const isBasic = React.useContext(IsQueryBasicContext);
+  return (
+    <div
+      className={`${
```
maxxxxxdlp, line 36: please put newlines to make it nicer

specifyweb/frontend/js_src/lib/components/WbPlanView/CustomSelectElement.tsx
```diff
@@ -543,8 +543,7 @@ export function CustomSelectElement({
     header = (
       <header
         className={`
-          flex items-center gap-y-2 gap-x-1 rounded rounded-b-none
-          border border-brand-300 bg-brand-100 p-2 dark:bg-brand-500
+          flex items-center gap-y-2 gap-x-1 rounded rounded-b-none border border-brand-300 bg-brand-100 p-2 dark:bg-brand-500
```
maxxxxxdlp, line 546: this line is too long. make sure no line is longer than 80 chars

maxxxxxdlp (changes requested):
Looking good!
Only code formating comments

CarolineDenis:
@carlosmbe I just tested it several times and it remembers the last choice on my hand.. Could you delete the branch on the test panel and recreate one please? 
For the spacing between the line, it seems that it occurs on the test panel, I am still trying to look into this. 

@grantfitzsimmons, 
> there is a new table display allowing to fit a possible second row when the + button is triggered
> the tools buttons at the end of the lines and the trash button at the beginning take the whole height available 
> the table names are not bold anymore 
> we decided on a gap-4 between col and gap-3 between rows 
> the lines need to be dragged by the table name in the simple view version, we cannot have the same following effect as we have in the detailed as it is from the library we use
> we decided to have a scroll bar rather than a wrap when the width of the browser in small to access to right hand buttons

CarolineDenis:
Regarding the spacing between the rows that is too wide when not enough items: 
It seems that the front-end being off is not just related to the query lines / builder. By example we can see in the screen shot bellow that the test panel does not shrink the schema config as it should. There is issue a little bit every  where it seems. 
Need to investigate. 
Screenshot 2023-05-26 at 4 40 46 PM

specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx
```diff
@@ -80,7 +80,10 @@ export function QueryHeader({
   const [isBasic, setIsBasic] = useQueryViewPref(query.id);
 
   return (
-    <header className="flex flex-col items-center justify-between gap-2 overflow-x-auto whitespace-nowrap sm:flex-row sm:overflow-x-visible">
+    <header
+      className="flex flex-col items-center justify-between gap-2
```
maxxxxxdlp, line 84: fyi, using `"` for multi-line strings is discouraged as it's not well supported by all tools. If you need multi-line, use **{\`** and **\`}** instead (also, please indent everything correctly - that helps with readability and consistency - unfortunately prettier and eslint don't do that automatically in this caes

specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
```diff
@@ -239,7 +236,11 @@ export function QueryLine({
 
   const hasAny = field.filters.some(({ type }) => type === 'any');
 
-  const fieldSpec = QueryFieldSpec.fromPath(baseTableName, field.mappingPath);
+  const fieldSpec = React.useMemo(
+    () => QueryFieldSpec.fromPath(baseTableName, field.mappingPath),
+    [baseTableName, field.mappingPath]
+  );
+
   const rowTableName = generateMappingPathPreview(
```
maxxxxxdlp, line 244: (optional) you could put this into a useMemo too

specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
```diff
@@ -509,17 +510,21 @@ export function QueryLine({
             ))}
           </div>
         ) : (
-          <span className={`${isBasic ? 'col-span-2' : 'contents'}`} />
+          <span
+            className={`
+            ${isBasic ? 'col-span-1' : 'contents'}
```
maxxxxxdlp, line 515: sorry to be so pedantic about this, but could you please maintain consistent indentation?

i.e, the line starting with `$` should be indented
the line **\`}** should be indented too. Like this:

```ts
    className={`
      ${isBasic ? 'col-span-1' : 'contents'}
    `}
```

See that first line and last line has the same indentation, and all others are indented one more.

Or, if there is enough space (to fit everything below 80 chars), you could do it all in one line:

```ts
    className={`${isBasic ? 'col-span-1' : 'contents'}`}
```

specifyweb/frontend/js_src/lib/components/QueryBuilder/QueryLineTools.tsx
```diff
@@ -50,9 +54,11 @@ export function QueryLineTools({
       <Button.Small
         aria-label={queryText.showButtonDescription()}
         aria-pressed={field.isDisplay}
-        className={`${className.ariaHandled} ${
-          isFieldComplete ? '' : 'invisible'
-        }`}
+        className={`
+         ${className.ariaHandled} ${isFieldComplete ? '' : 'invisible'} ${
```
maxxxxxdlp, line 58: you could make this quite easier to read by indenting it better:

```ts
    className={`
      ${className.ariaHandled}
      ${isFieldComplete ? '' : 'invisible'}
      ${isBasic ? 'h-full' : ''}
    `}
```

(i.e, a separate line for each ${ block)

isn't this easier to read?

maxxxxxdlp (approved):
let's get class name indentation right
otherwise, all is good!
hope you can figure out what's causing the issue on the test panel - you could try creating a new user both locally and on the test panel (this way, you would be testing under default user preferences in both cases)

CarolineDenis:
Query line having wrong spacing is fixed.

specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
```diff
@@ -165,64 +165,70 @@ export function QueryFields({
...
+                }
+                onOpen={handleOpen?.bind(undefined, line)}
+                onOpenMap={handleOpenMap?.bind(undefined, line)}
+                onRemove={handleRemoveField?.bind(undefined, line)}
+              />
+            </li>
+          </ErrorBoundary>
+        ))}
+      </Ul>
+      {isBasic ? <span className="flex-1" /> : null}
```
maxxxxxdlp, line 230: Are you sure that the issue can't be fixed by just adding one of these tailwind class names to the grid container? (instead of adding a flex-1 hack)
https://tailwindcss.com/docs/align-content
https://tailwindcss.com/docs/align-items

maxxxxxdlp (approved):

grantfitzsimmons (changes requested):
**Issue 1**

Editing the Taxon rank in a query line throws an error:

```
Error: Tried to get unknown field: 0
```

[Specify 7 Crash Report - 2023-06-05T14_11_29.096Z.txt](https://github.com/specify/specify7/files/11653412/Specify.7.Crash.Report.-.2023-06-05T14_11_29.096Z.txt)

https://github.com/specify/specify7/assets/37256050/6be8b8d3-5fab-439b-be5b-9b7354414e57

https://sp7demofish-issue-2479.test.specifysystems.org/specify/query/55/

---

**Issue 2**

Some fields are not discernable from others. For instance, Taxon Name and Taxon Full Name both show as "Taxon"

https://github.com/specify/specify7/assets/37256050/8bcf201c-6853-45fb-9a5f-c0ced46bc242



---

**Issue 3**

Selecting a row does not highlight it like in the detailed view. It should have a gray background just like it does in the other view.

https://github.com/specify/specify7/assets/37256050/ac6e0812-4bee-4ffe-9449-3076ca74d77e

---

**Issue 4**

There should be horizontal lines separating query item lines like in detailed view to make it more clear where one starts and another ends

image

image

CarolineDenis:
@grantfitzsimmons 
Issue 1: should be fixed
Issue 2: a new issue will be open for that to change it in results as well 
Issue 3: is fixed 
Issue 4: we can investigate for a new solution but at the moment the className 'contents' that is added to the line does not allow us to have the separator line

specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
```diff
@@ -164,71 +166,76 @@ export function QueryFields({
 
   const isBasic = React.useContext(IsQueryBasicContext);
 
+  const fieldSpecsAndFields = queryFieldsToFieldSpecs(baseTableName, fields);
```
maxxxxxdlp, line 169: Consider wrapping this in a React.useMemo

specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
```diff
@@ -164,71 +166,76 @@ export function QueryFields({
...
+  const fieldSpecs = fieldSpecsAndFields.map(
+    ([_field, fieldSpec]) => fieldSpec
+  );
+  const fieldName = fieldSpecs
```
maxxxxxdlp, line 173: Wrap this in a React.useMemo

specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
```diff
@@ -237,14 +239,20 @@ export function QueryLine({
...
   );
 
-  const rowTableName = generateMappingPathPreview(
-    fieldSpec.baseTable.name,
-    fieldSpec.toMappingPath()
-  );
+  const rowTableName =
+    fieldSpec === undefined
+      ? fieldName
+      : generateMappingPathPreview(
```
maxxxxxdlp, line 252: please move this into a React.useMemo
remember that there can be many query lines rendered at once, but the user can only modify one query line at a time, thus all others will be re-rendered uselessly and re-run generateMappingPathPreview - which is a somewhat expensive function

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
```diff
@@ -166,15 +166,27 @@ export function QueryFields({
...
+    () =>
+      fieldSpecs
+        .map((field) =>
+          generateMappingPathPreview(
+            field.baseTable.name,
+            field.toMappingPath()
+          )
+        )
+        .join(' '),
+    [fieldSpecs]
```
maxxxxxdlp, line 188: careful! you need eslint to tell you about bugs like this
fieldSpecs is re-created on every render, so this useMemo is counter-productive.
you should either move fieldSpecs inside this useMemo, or have fieldSpecs call it's own useMemo

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

carlosmbe (approved):

grantfitzsimmons:
> Issue 2: a new issue will be open for that to change it in results as well

> Issue 4: we can investigate for a new solution but at the moment the className 'contents' that is added to the line does not allow us to have the separator line (need to open new feature/issue)

I think that these are both too important to leave unresolved. ☹️

grantfitzsimmons:
https://github.com/specify/specify7/assets/37256050/667bf7ea-4939-4860-9783-d5fe4987c8de

maxxxxxdlp:
Yes, you can't add border to elements with `contents` - but there is a workaround

for example, look at what workbench mapper does - it adds borders bottom to each elements of the line, rather than to the line container as a whole

there is also a simpler workaround than that:
take a look at this:
https://tailwindcss.com/docs/grid-row#spanning-rows

think of what you can do with it

<details><summary>Hind</summary>
<p>

you could do something with `row-span-full`

</p>
</details> 

<p>

you can create an element with `row-span-full` class name and 1px border or 1px background to emulate a vertical line - and the just repeat this element after every line

</p>
</details>

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingPreview.ts
```diff
@@ -110,6 +106,7 @@ export function generateMappingPathPreview(
         : (optionLabel as string);
     }),
   ];
+  const isAnyRank = mappingLineData[0].fieldsData['$-any'] !== undefined;
```
maxxxxxdlp, line 109: please don't use `$-any` once again 😢 
there exists `anyTreeRank`

maxxxxxdlp: There already is a (possibly more robust) check for `any` in this file:
https://github.com/specify/specify7/blob/041259050c024debfa26bef7a0a68df01330dd6d/specifyweb/frontend/js_src/lib/components/WbPlanView/mappingPreview.ts#L175

Can you extract that into a variable and reuse it?

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingPreview.ts
```diff
@@ -27,11 +27,7 @@ import {
 import { getMappingLineData } from './navigator';
 
 /** Use table name instead of field name for the following fields: */
-const fieldsToHide = new Set<string>([
-  'fullName',
-  'localityName',
-  formattedEntry,
-]);
+const fieldsToHide = new Set<string>(['localityName', formattedEntry]);
```
maxxxxxdlp, line 30: This file is very complicated. Please add/update the test cases after any change (i.e add test cases for taxon name, full name, with any rank and non-any rank)

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

grantfitzsimmons (changes requested):
We just need to reduce the padding (I know we've discussed this before, but I have had an epiphany):

## Currently:
image

image


## Proposed

Set `row-gap` and `column-gap` to `0.5rem` 

image

image

grantfitzsimmons (approved):

PR #3460, opened 2023-05-05
Disable transparent background for Firefox

### LINKED ISSUE #3449: Remove semi-transparent tree headings
In tree viewer, text behind the headings is blurred in browsers that support it
Screenshot 2023-05-04 at 10 33 42

In browsers that don't support it, we added transparency instead:
image

However, as @beach53 mentioned, it looks confusing (like a UI) bug, so for browsers that don't support blur we should probably just disable transparency

CarolineDenis:
Fixes #3449


specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -252,7 +252,9 @@ function TreeView<SCHEMA extends AnyTree>({
                   ${index === 0 ? '-ml-2 rounded-bl pl-4' : ''}
                   ${index + 1 === length ? '-mr-2 rounded-br pr-4' : ''}
                   ${
-                    reduceTransparency || !supportsBackdropBlur
+                    reduceTransparency ||
+                    !supportsBackdropBlur ||
+                    navigator.userAgent.includes('Firefox')
```
maxxxxxdlp, line 257: This is not a correct fix.
This assumes that all versions of firefox don't support this feature and never will. It also assumes that all other browsers support that feature and always did.

Instead, we can specifically check if current browser supports a given CSS feature using the `@supports` keyword in CSS or `CSS.supports` function in javascript. See more - https://developer.mozilla.org/en-US/docs/Web/API/CSS/supports_static

Please look at that article to familiarize yourself with how this works.
You don't have to actually write `CSS.supports` as it has already been done - that's what the `supportsBackdropBlur` variable right above this line is for.

In this case, `supportsBackdropBlur` happens to be true even in firefox, and yet blur doesn't work. Thus the next task here is to figure out why blur is not working (i.e, check other examples of backdrop blur where it works in firefox and see what are we doing differently)

maxxxxxdlp (changes requested):

CarolineDenis:
It appears that the problem is related to compatibility issues with Firefox. Specifically, when attempting to apply backdrop-blur to the title, there is no visible effect. However, if backdrop-blur is applied to the node instead, the title becomes blurred behind the passing node when scrolling up. This suggests that there may be an issue with the z-index or that the effect needs to be applied to the child rather than the parent element. Further investigation will be necessary once Firefox has resolved its compatibility issues.

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -250,14 +248,9 @@ function TreeView<SCHEMA extends AnyTree>({
...
                   className={`
-                  sticky top-0 whitespace-nowrap border border-transparent p-2
-                  ${index === 0 ? '-ml-2 rounded-bl pl-4' : ''}
-                  ${index + 1 === length ? '-mr-2 rounded-br pr-4' : ''}
-                  ${
-                    reduceTransparency || !supportsBackdropBlur
-                      ? 'bg-gray-100 dark:bg-neutral-900'
-                      : 'bg-gray-100/60 backdrop-blur-sm dark:bg-neutral-900/60'
-                  }
+                sticky top-0 whitespace-nowrap border border-transparent border-b-[color:var(--accent-color-300)] bg-[color:var(--background)] p-2 brightness-95 dark:brightness-105
```
maxxxxxdlp, line 251: looks like formatting here is completely broken
unfortunately prettier doesn't fix it automatically
can you fix it please and be on the lookout for this in the future?

Mainly:
1. Lines should not be longer than 80 chars
2. The string should be indented (`sticky...` should be one more indentation level than `className={`, not one less)

maxxxxxdlp: Also, the header has a totally different look now, it's just completely black?

Screenshot 2023-07-03 at 19 08 00

not sure I like the new look. I think we can both get rid of the problematic blur effect, and have the header look somewhat similar to how it was before.

but I would defer to product engineers on this. @grantfitzsimmons

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
CarolineDenis: yes we agreed with @grantfitzsimmons to make the header a solid color. The background of the header will be a slightly darker color than what the user choose for his normal SP background in light mode and slightly lighter color than what the user choose for his normal SP background in dark mode. We could choose to always have same color as a background but I thought I wasn't looking great when the user was customizing his b-g (we could increase the brightness more in dark-m and decrease more in light-m). 
The bottom border is also depending on the custom color 3 that the user is choosing. Again to have it harmonized with the user choices. 
I personally like the new look.

CarolineDenis: Decision made with @grantfitzsimmons : 
- Increase brightness of the solid color b-g in dark-mode keep the solid color in light-mode

maxxxxxdlp (approved):
code looks good. will defer to others for design in light mode and dark mode

bronwyncombs (approved):
Looks good 👍 

https://github.com/specify/specify7/assets/135047322/c9029295-c8ca-4a9d-a002-e718518fc66b

https://github.com/specify/specify7/assets/135047322/256499d8-88be-4476-88b2-5e248583d252

PR #3463, opened 2023-05-08
Add alt text for attachment preview, add label for share record box

### LINKED ISSUE #2615: Evaluate ARC Toolkit for accessibility checking
https://www.tpgi.com/arc-platform/arc-toolkit

CarolineDenis:
Fixes #2615

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormMeta/ShareRecord.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormMeta/ShareRecord.tsx
@@ -45,12 +45,12 @@ export function ShareRecord({
         );
   const url = new URL(rawUrl, globalThis.location.origin).href;
   return (
-    <div className="flex flex-col gap-2">
+    <label className="flex flex-col gap-2">
       <h4>{formsText.shareRecord()}</h4>
       <div className="flex gap-2">
         <Input.Text className="!cursor-auto" defaultValue={url} isReadOnly />
         <CopyButton text={url} />
       </div>
-    </div>
+    </label>
   );
 }
--- a/specifyweb/frontend/js_src/lib/components/Molecules/AttachmentPicker.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/AttachmentPicker.tsx
@@ -76,6 +76,7 @@ export function AttachmentPicker({
         <img
           className="h-40 max-h-full w-40 max-w-full object-contain"
           src={url}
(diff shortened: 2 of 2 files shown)


specifyweb/frontend/js_src/lib/components/FormMeta/ShareRecord.tsx
```diff
@@ -45,12 +45,12 @@ export function ShareRecord({
         );
   const url = new URL(rawUrl, globalThis.location.origin).href;
   return (
-    <div className="flex flex-col gap-2">
+    <label className="flex flex-col gap-2">
       <h4>{formsText.shareRecord()}</h4>
```
maxxxxxdlp, line 49: headings must not be used inside of `<label>`. remove `<h4>`

maxxxxxdlp (changes requested):

chanulee1:
How do I test this?

CarolineDenis:
@chanulee1 
1)
> Open a record
> Open the meta menu 
> Inspect the shared record and verify it is a Label and not a h4
2)
> Open pref 
> Scroll to change home page image
> Inspect the image and verify it has an alt value

chanulee1:
Are these the correct `img` elements? In `preferences`, the image's alt value is `9k=` and on the homepage it does not have an alt value.

https://github.com/specify/specify7/assets/103299204/124f05a8-f4f5-41ce-9759-6d93d95499fc

CarolineDenis:
> Are these the correct `img` elements? In `preferences`, the image's alt value is `9k=` and on the homepage it does not have an alt value.
> 
>  lyQldoB1n8.mp4

It was for the preferences so it's good the alt text is there. 
For the '9k' alt value, it's correct too. That the name of your image which is what the alt value needs to be. 
Screenshot 2023-06-08 at 8 36 20 AM

chanulee1 (approved):
Looks good. Shared record is a label, image has an alt value.

maxxxxxdlp:
ah, the code doesn't handle the base-64 case properly.
Not a big deal, but it is a bug

Since base 64 string beings with something like this:

```
data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAASwAAAEBCAMAAAD1kWivAAADAFBMVEWysrL5nCYYGBj7...
```

We could potentially have an alt text say something like "Image Png" (by dynamically getting that part from the start of the base 64 string.

Still, likely not worth the trouble as this is a very minor bug

PR #3470, opened 2023-05-08
Display Specify logo with custom accent color

### LINKED ISSUE #2210: Logo should respect user's accent color
User's can change the accent color from the default orange.

Since logo is no an SVG, it can change automatically to match the accent color.

See also https://github.com/specify/specify7/issues/2082#issuecomment-1256423412

CarolineDenis:
Fixes #2210


specifyweb/frontend/js_src/lib/components/Atoms/Logos.tsx
```diff
@@ -0,0 +1,98 @@
...
+export const logos = {
+  shortSpecify: (
+    <svg
+      aria-hidden
+      className={logoClassName}
+      viewBox="0 0 142.34 142.32"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <defs>
+        <style />
```
maxxxxxdlp, line 15: can remove this empty `<style />` - it has not effect

specifyweb/frontend/js_src/lib/components/Atoms/Logos.tsx
```diff
@@ -0,0 +1,98 @@
...
+        fill="currentColor"
+        stroke="currentColor"
+        strokeMiterlimit="10"
+        strokeWidth="1.5"
+      />
+    </svg>
+  ),
+};
```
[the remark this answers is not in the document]
maxxxxxdlp: actually, nevermind, I have a better idea:
what if we need to display the logo somewhere outside of react?
I.e, the back-end is using this image right now

how about instead of creating this file, you try out the following:
- keep the short_logo.svg and logo.svg file
- in those files, replace the hardcoded colors with CSS variables, thus allowing you to customize the color. BUT: provide the hardcoded color as the default value for a color. See https://developer.mozilla.org/en-US/docs/Web/CSS/var. So for example, replace `#123456` with `var(--some-variable-name, #123456)`

Now, if back-end uses the image, the image will show up correctly using default colors
But, if displayed on the front-end, the colors will be customizable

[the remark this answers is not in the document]
maxxxxxdlp: okay, now I have an even better idea:
don't need to modify short_logo.svg or logo.svg at all
Instead, apply a CSS filter for the image to hue rotate the colors to be of correct accent color - https://tailwindcss.com/docs/hue-rotate
To calculate the degree to which you need to rotate the color, convert the user's accent color and our original brand color to hsl (create a JS helper function for that conversation - look online how to convert from hex to hsl). Then, the hue values of both values can be compared and converted to degrees

Can provide more details if this is not clear

specifyweb/frontend/js_src/lib/components/Header/index.tsx
```diff
@@ -107,30 +108,28 @@ export function Header({
...
             `}
           href="/specify/"
         >
           {/* Both logs are loaded to prevent flickering on collapse/expand */}
-          <img
-            alt=""
-            className={`
-                hover:animate-hue-rotate
+          <span
+            className={`w-full
```
maxxxxxdlp, line 119: you removed `alt=""` - it was important
read online the difference between having empty alt and no alt

maxxxxxdlp (changes requested):

specifyweb/frontend/static/img/logo.svg
maxxxxxdlp, line 1: was this file changed or just reformatted?
(optional) if no changes were made, consider reverting back the reformatting just to not pollute git history needlessly

specifyweb/frontend/static/img/short_logo.svg
maxxxxxdlp, line 1: same as above, was this file changed or just reformatted?

specifyweb/frontend/js_src/lib/components/Header/index.tsx
```diff
@@ -108,7 +124,7 @@ export function Header({
       <h1 className="contents">
         <a
           className={`
-              flex items-center
+            flex items-center
```
maxxxxxdlp, line 127: undo this change

maxxxxxdlp (changes requested):
Quite good

specifyweb/frontend/static/img/logo.svg
CarolineDenis: I change it

specifyweb/frontend/js_src/lib/utils/__tests__/hexToHsl.test.ts
```diff
@@ -0,0 +1,39 @@
+import { hexToHsl } from '../utils';
+
+describe('hexToHsl', () => {
```
maxxxxxdlp, line 3: Looking very good!

This style of tests (with `describe` and `test`) is the most common and most flexible. It allows for testing asynchronous functions, mocking functions, and having multiple `expect()` assertions.

That is why it's good to know how to write this style of tests.
However, to make writing tests simpler and quicker, we have an alternative style - using a `theories` function - that will allow you to replace the entire `describe` blog with just ~7 lines!

Please take a look at our existing tests to see how `theories` is used. I also pushed a commit on this branch that improved `theories` documentation.

Main advantage of `theories` style is how much smaller the test definitions are (and thus it takes less time to write them). Also, you don't have to write a test case names, as `theories` generates test case names on the fly for you! But, they are limited in features. Including:
- Don't integrate with IDE test running plugins as well
- Don't support async functions
- Don't support multiple assertions
- Don't support mocking (unless doing a mock for an entire file, rather than a single test)

Thus, it's best to know both styles and use whichever is most appropriate in a given case

maxxxxxdlp (changes requested):

grantfitzsimmons (changes requested):
Brand consistency is important, so I believe we should keep the white background behind the Sp

I think that the white gives it character and plays into the “tile” aesthetic. For branded t-shirts, posters, flyers, etc. we would lose legibility for the “Sp” when its on another color background. The favicon relies on the white to be clearly visible, and we can ensure enough contrast for the icon in any position or location.

image
image

grantfitzsimmons:
image

Even here, legibility is reduced. The text is just too thin to not have the background

maxxxxxdlp:
Very good points!
Thanks for catching this

grantfitzsimmons:
Another comment–
I don't think that this is an accurate representation of the color selected:
image

Once we have the white background back– I think we need to make sure that the color selected is much closer to the logo color. This should be a dark purple color, not light pink.

I am colorblind so I am not the person for this.

maxxxxxdlp:
In the current implementation, we only allow them to change the hue, not the saturation or luminosity, thus there are only 256 colors they could set the logo to, not 16.7 million like the accent color

This is to prevent setting the logo to an inappropriate color - but we could remove this prevention if you want

grantfitzsimmons:
> This is to prevent setting the logo to an inappropriate color - but we could remove this prevention if you want

Dark (dark purple), "inappropriate" colors are OK, light (neon green) "inappropriate" colors are not. Can we make a distinction?

maxxxxxdlp:
We could release this anyway and remove that limitation when fixing https://github.com/specify/specify7/issues/3503 if that's not a deal breaker

grantfitzsimmons:
That is OK with me! Just need to add back the white background behind Sp & I'm OK with merging

carlosmbe:
Tested PR and works. Just the white background remaining

grantfitzsimmons:
Another thing, the logo color remains the same in loading dialogs & on the home screen:

https://github.com/specify/specify7/assets/37256050/a5e5f1fb-5e38-4b53-bb2a-98c3afeac9d5

These should be consistent with it as well

CarolineDenis:
As the Specify front-end developer, I strongly believe that the logo looks better without the white background. It not only looks better with default settings but also when the user changes the color. With the availability of dark mode and the option to change the background color, the white background becomes unnecessary and removing it gives a cleaner and more modern look. I disagree with the notion that the logo becomes less visible in tabs. If you take a look at the screenshot below, you'll see that transparent backgrounds are quite common among big brands nowadays.

Furthermore, I prioritize having a better-looking logo on the specified 7 app over investing in better t-shirts or other merchandise. I agree that maintaining consistency across our platforms is important, and I strongly believe that the logo should be changed to this new design on all our platforms. This includes the Discourse platform, where the current logo is not even being used on Sp7.

Regarding the tiles, I also agree that we need consistency, and once again, removing the white background makes them look significantly better. Please refer to the attached screenshot to see the difference.

@acwhite211 @beach53 What are your thoughts on this? 

MENU => 
menuCustom
menuDefault1
Screenshot 2023-05-12 at 9 31 06 AM

LOGIN PAGE 
=> 
login

TABS => 
tabs

TILES=> 
tilesBlack:White
tilesBlack
tilesPink:white
tilesPink

acwhite211:
I think it looks a lot better without the white background.  I don't think it's any less recognizable, so I don't think consistency is a significant issue.  I think the page looking good outweighs the con of minor inconsistency.

grantfitzsimmons (changes requested):
image

The gradient should be on both sides, but it is missing on the left

grantfitzsimmons (changes requested):
image

The gradient should be on both sides, but it is missing on the left

CarolineDenis:
@grantfitzsimmons let me know how this new favicon looks like on your end

specifyweb/frontend/js_src/lib/components/HomePage/index.tsx
```diff
@@ -49,16 +50,24 @@ function WelcomeScreenContent(): JSX.Element {
...
+  const hueDifference = useHueDifference();
+  return (
+    <div className="relative">
+      <img
+        alt=""
+        className="w-[800px]"
+        src={defaultWelcomePageImage}
+        style={{ filter: `hue-rotate(${hueDifference}deg)` }}
+      />
+      <div className="absolute top-0 h-full w-[20%] bg-[linear-gradient(to_right,var(--background),transparent)]" />
```
maxxxxxdlp, line 69: I know it was that way before, but can you please add a comment in the code above this line saying that the following 2 gradients do a "fade out" effect for the left and right edge of the SVG?

because I already forgot what this does, and it's not obvious from the code

maxxxxxdlp (approved):

bronwyncombs (commented):
Color selecting functionality is great. I also like the no background look for the logo and home screen.

I noticed that changing the color to a dark grey/black does translate to the accent buttons, but not to the logo; same thing for light grey/white (Both in either mode). Further, the logo color that appears when Accent 3 is white or black is a salmon-ish coral-toned color. I think that if we're overriding logo color choice it should just be the original Specify orange. 
I understand the need for visibility given the user has a choice of light/dark mode, but it also allows you to set a white colored button where the button text is white in both modes. Maybe remove the white option altogether since it doesn't translate the logo anyway?

Screenshots with button color set to white in light and dark mode:
msedge_n0cj6JaYzJ
msedge_FOMgjzzqXr

What do you all think?

CarolineDenis:
@maxxxxxdlp for this comment "Further, the logo color that appears when Accent 3 is white or black is a salmon-ish coral-toned color." 
This is due to the fact that we are applying a filter over the svg image depending on the difference of hue. 
We can decide that 
a) it is ok to have a orange/salmon color for the logo when white is chosen as accent 3 custom color 
or 
b) we need to have the logo as an icon and change its color to accent--color-3 (we will not use then the function useHueDifference we will only do i.e icon.logo and text-color:accent--color-3.

maxxxxxdlp:
1. a white button with white text is not practical and if it were easy to do so, should be disallowed by us (unfortunately, it's not too easy to implement)
2. we could allow any arbitrary color for the logo. the hue thing you implemented should work for normal uses cases - feel free to consult with the product engineers and decide what is preferred (i.e less choice or more choice, but also allowing users to shoot themself in the foot)

CarolineDenis:
After discussion with @grantfitzsimmons : 
- A new issue will be open for the white color problem. 
- this PR can merged

grantfitzsimmons:
Rebasing to `v7.9-dev` and will test on here

grantfitzsimmons (changes requested):
I'm still not 100% sold on the thin orange on black, but I will approve

grantfitzsimmons (approved):

Issue #3343, opened 2023-04-14
Improve formatting of back-end exceptions


maxxxxxdlp:
![Screenshot 2023-04-07 at 15 53 01](https://user-images.githubusercontent.com/40512816/231913435-7b95633c-158a-4cf7-997b-ead9aad152e1.png)

Back-end sends a stack trace properly formatted, but front-end glues all lines into one big mess. Need to wrap it in `<pre></pre>` to fix this. Better yet, @melton-jason is there a way to get back the original django exception page (when in development) - it was quite helpful

See https://github.com/specify/specify7/pull/3315#pullrequestreview-1383384367 for another example

melton-jason:
I believe this issue was not always the case. Some update (maybe the Django/Python update) or change in the code caused this problem. 
Even now there are times when I come across the original Django error messages in Debug mode, but I am unsure exactly what causes it. 

I have done some looking into this before: 

The middleware _is_ properly returning nothing (None) when Debug mode is enabled
https://github.com/specify/specify7/blob/393538b081eb797beb502204cdea9311179361f6/specifyweb/middleware/general.py#L45

So Django is sending its own error message to the frontend

Perhaps changing the order in which Django handles the middleware will alleviate the problem ([See the Django docs on Middleware layering](https://docs.djangoproject.com/en/4.2/topics/http/middleware/#middleware-order-and-layering)) or maybe Django has since migrated the error message into another Middleware option ([here is a list of all available Middleware](https://docs.djangoproject.com/en/3.2/ref/middleware/#available-middleware))

Currently, our Middleware is layered as:

https://github.com/specify/specify7/blob/393538b081eb797beb502204cdea9311179361f6/specifyweb/settings/__init__.py#L186-L197

I will do some further investigation so we can get back to the Django responses!

melton-jason:
I have done further investigation into this issue and here is what I have found so far:
- This issue is not caused by the middleware introduced in #2702
  - Completely removing the middleware and rebuilding the backend containers still has this exhibit the same behavior
- Templated views (like the 404 Page Not Found view) still appear as normal, without the weird formatting. 

![django_template](https://github.com/specify/specify7/assets/64045831/06ebb7b8-64af-4bff-bb75-3d5a4c350a3e)

- The backend response is being parsed as HTML by the frontend. This is occurring from the following code snippet
[...]

PR #3474, opened 2023-05-09
Extend Record Merging to multiple records

acwhite211:
Extend Record Merging to multiple records to allow for the handling of businesses rules and keeping the whole merging process inside of one ACID transaction.


acwhite211:
Still working on creating more unit tests, but please go ahead and start testing with front-end integration. 
The old url `/api/specify/old_model_id/new_model_id/` still exists, the new extended record merging url is `/api/specify/new_model_id/` with a json body containing old_model_ids and new_record_data

acwhite211:
@melton-jason feel free to add more unit tests

specifysoftware:

https://discourse.specifysoftware.org/t/record-merging-in-specify-7/939/11

specifyweb/specify/views.py
```diff
@@ -502,9 +510,29 @@ def update_record(record: models.Model):
...
 
     # Dedupe by deleting the agent that is being replaced and updating the old model ID to the new one
-    target_model.objects.get(id=old_model_id).delete()
-
-    return http.HttpResponse('', status=204)
+    for old_model_id in old_model_ids:
+        target_model.objects.get(id=old_model_id).delete()
+
+    # Update new record with json info, if given
+    has_new_record_info: bool = False if new_record_info is None else True
```
maxxxxxdlp, line 517: are type annotations required by `mypy` even when can be trivially inferred?

specifyweb/specify/views.py
```diff
@@ -555,5 +583,85 @@ def record_merge(request: http.HttpRequest, model_name: str, old_model_id: int,
...
+    },
+})
+
+@login_maybe_required
+@require_POST
+def record_merge_extended(request: http.HttpRequest, model_name: str, new_model_id: int) -> http.HttpResponse:
+    """Replaces all the foreign keys referencing the old record IDs
+    with the new record ID, and deletes the old records.
+    """
+    request_params = http.QueryDict(request.META['QUERY_STRING'])
```
maxxxxxdlp, line 640: there should be an easier way to do this.
Something like this:
```py
if 'version' in request.GET:
  version = request.GET['version']
else:
  version = getattr(spmodels, model_name.title()).objects.get(id=new_model_id).version
```

or maybe even this, to avoid needless if/else:
```python
default_version = getattr(spmodels, model_name.title()).objects.get(id=new_model_id).version
version = request.GET.get('version', default_version)
```

specifyweb/specify/views.py
```diff
@@ -502,9 +510,29 @@ def update_record(record: models.Model):
...
+    if has_new_record_info:
+        obj = api.put_resource(new_record_info['collection'],
+                               new_record_info['specify_user'],
+                               model_name,
+                               new_model_id,
+                               new_record_info['version'],
+                               new_record_info['new_record_data'])
+
+    # Return http response
+    if has_new_record_info:
```
maxxxxxdlp, line 527: should permissions be checked before `.delete()` and `.put_resource()` were called?

specifyweb/specify/views.py
```diff
@@ -555,5 +583,85 @@ def record_merge(request: http.HttpRequest, model_name: str, old_model_id: int,
...
+            "204": {"description": "Success",},
+            "404": {"description": "The ID specified does not exist."},
+            "405": {"description": "A database rule was broken."}
+        }
+    },
+})
+
+@login_maybe_required
+@require_POST
+def record_merge_extended(request: http.HttpRequest, model_name: str, new_model_id: int) -> http.HttpResponse:
```
maxxxxxdlp, line 636: I am not too familiar with django/python/sqlalcheme, but it this code being run as a transaction? i.e., if some records get deleted, but then it falls mid-way, (potentially because of missing permissions), do those records get un-deleted?

additionally, if those records have attachments, does the back-end send requests to the asset server to delete the attachments? (it might have triggers doing that) that might cause issues if request doesn't go though fully

maxxxxxdlp (changes requested):

specifyweb/specify/urls.py
```diff
@@ -10,6 +10,7 @@
 urlpatterns = [
     # replace record
     url(r'^specify/(?P<model_name>\w+)/replace/(?P<old_model_id>\d+)/(?P<new_model_id>\d+)/$', views.record_merge),
```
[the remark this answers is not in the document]
maxxxxxdlp: having a UI that makes requests would also make it easier for you to construct test cases

specifyweb/specify/views.py
acwhite211: ya, I made sure to research it.  The `@transaction.atomic` decorator encapsulates a database actions into one ACID transactions.  Even in the case of recursively calling the function, creating nested transactions, either all the actions occur, or none of them.  I'm not sure about the attachments, it's not something I can unit test easily.  We should make sure to test that testcase in the integration testing.  I'll look to see if I can find any triggers for deleting attachments, but I don't think it would cause a trigger to occur until the full atomic transaction is completed.

specifyweb/specify/views.py
```diff
@@ -501,36 +509,55 @@ def update_record(record: models.Model):
...
+    # Return http response
     return http.HttpResponse('', status=204)
-
+    
 @openapi(schema={
     'post': {
         "requestBody": {
             "required": True,
-            "description": "Replace a new model for an old model.",
+            "description": "Rplace a list of old records with a new record.",
```
melton-jason, line 533: Looks like there's a typo here (missing an e in 'Replace')

specifyweb/specify/views.py
```diff
@@ -545,15 +572,28 @@ def update_record(record: models.Model):
...
+    with the new record ID, and deletes the old records.
     """
-    check_permission_targets(None, request.specify_user.id, [ReplaceRecordPT.update, ReplaceRecordPT.delete])
-
-    # TODO: Test reocord merging targeting records other than Agent before allowing generic merging
-    # if model_name.lower() != 'Agent'.lower():
-    #     return http.HttpResponseForbidden(f'Record Merging for the {model_name} table is not yet been fully tested.')
-
-    response = record_merge_fx(model_name, old_model_id, new_model_id)
+    request_params = http.QueryDict(request.META['QUERY_STRING'])
```
melton-jason, line 579: What is the purpose of this `request_params` object? It's not used anywhere else in the function.

specifyweb/specify/api_tests.py
```diff
@@ -688,11 +698,104 @@ def test_agent_address_replacement(self):
...
+            id=5,
+            agenttype=0,
+            firstname="agent",
+            lastname="005",
+            specifyuser=None)
+
+        # Create mock addresses
+        models.Address.objects.create(
+            id=1,
+            timestampcreated="22022-11-30 14:34:51.000",
```
melton-jason, line 741: I'm not terribly familiar with this, but does Django properly convert this to a Date Time object, or leave it as a string? 
It may matter when we make Django time-aware, and it's good to have the data as closely representative to what it would be in the 'real-world' application

melton-jason (changes requested):
I will need to take a more indepth look at this and add a unit test for when Agents have Group Person records. 
For a quick thrice-over, everything looks good, but I do have some questions.

grantfitzsimmons (changes requested):
image
https://kufish51623-agent-merging-extremos.test.specifysystems.org/specify/overlay/merge/Agent/?records=3%2C5%2C6%2C7%2C8

After waiting for a few minutes, it seems that the merging process failed. It showed a red bar at the top of the dialog, but no error was reported to the user.

When opening the console, it seems there were some errors reported there.

Console Log:
[kufish51623-agent-merging-extremos.test.specifysystems.org-16842.zip](https://github.com/specify/specify7/files/11490881/kufish51623-agent-merging-extremos.test.specifysystems.org-16842.zip)

maxxxxxdlp:
@melton-jason
Three cases:
- None of the agents are connected to any users - no issue
- One of the agents is connected to a user - link the user to the merged agent instead
- More than one agent is assigned to a (same or different) user - quite unlikely. bail out with a nice error message

melton-jason:
Yes that would be a nice way of going about this particular case. 
Some more context on the `More than one agent is assigned to a (same or different) user` case; I think having an error message in this case would be fine. 
This operation is not supported in the UI at all (I do not know if it is supported in Specify 6 or not): you can not assign an agent to more than one user. 

We will need to think of ways to generalize cases such as these (if broad generalization can exist to cover _all_ backend deletion Business Rules). 
Here are some options we can take a look into:
- Utilize Django Signals
- Extend Deletion Rules
- Behavior based on attributes in the data model

maxxxxxdlp:
> This operation is not supported in the UI at all (I do not know if it is supported in Specify 6 or not): you can not assign an agent to more than one user.

I'm afraid you misunderstood my comment. Indeed, you can't assign the same agent to multiple users.
But, the agent merging tool deals with merging multiple agents into one, thus it's entirely possible (though hopefully not likely) that you would try to merge two agents who both are assigned to a user

maxxxxxdlp:
@melton-jason the simplest solution for this issue:
The front-end for agent merging has code for picking which agent should be considered "main"
We should just change it to consider agent connected to a user to be the main one
See the (untested) fix I pushed

melton-jason:
> @melton-jason the simplest solution for this issue: The front-end for agent merging has code for picking which agent should be considered "main" We should just change it to consider agent connected to a user to be the main one See the (untested) fix I pushed

https://github.com/specify/specify7/assets/64045831/26ff18e4-07e5-4644-81a7-e2f7931161de

@maxxxxxdlp 

Unfortunately, this is not a complete fix. 
What about the case when more than one Agents are associated to separate Specify Users? 
Regardless of what it chooses as the 'main' Agent, there will always be another Agent associated with a Specifyuser that the backend will try and delete.

melton-jason:
However, this falls under the third `More than one agent is assigned to a (same or different) user - quite unlikely. bail out with a nice error message` case. 
How likely is this in a production database?

maxxxxxdlp:
> What about the case when more than one Agents are associated to separate Specify Users?

I don't think this is very likely
@grantfitzsimmons what do you think?
And even if this does occur, a business rule error would occur, which the back-end business error handler code has to handle anyway

grantfitzsimmons:
> > What about the case when more than one Agents are associated to separate Specify Users?
> 
> I don't think this is very likely @grantfitzsimmons what do you think?

It is not very likely at all

melton-jason:
> image https://kufish51623-agent-merging-extremos.test.specifysystems.org/specify/overlay/merge/Agent/?records=3%2C5%2C6%2C7%2C8
> After waiting for a few minutes, it seems that the merging process failed. It showed a red bar at the top of the dialog, but no error was reported to the user.
> 
> When opening the console, it seems there were some errors reported there.
> 
> Console Log: [kufish51623-agent-merging-extremos.test.specifysystems.org-16842.zip](https://github.com/specify/specify7/files/11490881/kufish51623-agent-merging-extremos.test.specifysystems.org-16842.zip)
> 
> image

This is still concerning, and I have been unable to identify the cause of the issue thus far. 
Besides the issue I pointed out regarding Specify Users, I have been able to merge Agents without incidence, both on a local instance and for simple Agents (which I always created beforehand) on the `kufish_5_16_23` database.

Perhaps the _most_ concerning and perplexing symptom of this issue is how _little_ it gives in terms of information to debug the problem. The only information that can be gleaned directly from the issue is the 405 (Method Not Allowed) response.  
Usually this might be caused by a misconfigured url or the wrong request type, but everything seems correct as far as these are concerned. 

@maxxxxxdlp 
Did I make a mistake in structuring the request in 03f11a1 and subsequentially 8d5f8f6?

maxxxxxdlp:
Code looks ok. This could be caused by test panel restarting. That would explain the 421 error
405 would happen in that case because while test panel is restarting, the requests were sent to the nginx container or to test panel container, which doesn't accept POST

maxxxxxdlp:
https://github.com/specify/specify7/pull/3474#issuecomment-1551330763 - I really don't like how the error message is formatted in the video

If I remember correctly, this is only happening when in development. And we still don't know why that's happening?

melton-jason:
> [#3474 (comment)](https://github.com/specify/specify7/pull/3474#issuecomment-1551330763) - I really don't like how the error message is formatted in the video
> 
> If I remember correctly, this is only happening when in development. And we still don't know why that's happening?

In my free time in the past two days I have been looking into why this formatting is happening (#3343). I will update the issue shortly with what I have found in the interim.
If getting back to the correct Django error messages for development is a priority, I can focus on doing so. Or at the very least, I can devote more time to doing so.

maxxxxxdlp:
You have a ton of other important things!
As long as it's better in production, it's not too bad

melton-jason:
> Code looks ok. This could be caused by test panel restarting. That would explain the 421 error 405 would happen in that case because while test panel is restarting, the requests were sent to the nginx container or to test panel container, which doesn't accept POST

Yes, I first thought it might be an issue with the test panel restarting. However, I have attempted to perform that exact merge multiple times, and every time encountered the same behavior. 
It would make sense if the requests were being to NGINX or Test Panel containers, where POST requests are not allowed. If this is the case, then how do we manage other POST requests like creating records? The chances of the test panel being down _every_ time @grantfitzsimmons and I have attempted the merge is unlikely. Moreover, I did not encounter any `502 Gateway Time Out` or `TypeError: Failed to Fetch` errors during these attempted merges.

I have successfully merged a variety of Agents both locally and through the same `kufish_5_16_23` database, so the endpoint _does_ work correctly. 
I am just baffled by the fact that the error encountered by @grantfitzsimmons in https://github.com/specify/specify7/pull/3474#pullrequestreview-1429171474 provides very little helpful information. The 405 error is the _only_ information the application provides. There is no Business Rule, Django, or SQLAlchemy error. 
Trying to uncover why _that_ merge went wrong would be like trying to find a needle in _every_ haystack. 

There is a _lot_ that could go wrong, and we don't have a lot to go off of.

melton-jason:
I have identified a case on a local copy of  `umich-mollusks` which exhibited the same behavior as in  https://github.com/specify/specify7/pull/3474#pullrequestreview-1429171474. I discovered the issue, and was able to apply a fix. 

Here is a video showing two agents which I tried to merge which showed the aforementioned behavior. 

https://github.com/specify/specify7/assets/64045831/04c67bd8-96f6-43b3-9320-1cae2ed7aefe

Assuming the error was likely due to the references, I investigated the 13 Collecting Events which reference `ex Rush`. 
In the Collecting Events, there were multiple instances where `R.C. Rush` was a Collector alongside `ex Rush`. 
I removed `R.C. Rush` from being a Collector in every Collecting Event which also referenced `ex Rush`. 

Then I attempted the merge again, and it worked! 🥳 

https://github.com/specify/specify7/assets/64045831/1ea9d9ff-fba8-4b66-acc0-e042f15dd57d

Our phantom issue is what appears to be an Integrity Error. 
I will need to investigate how the endpoint works currently, and I can work on a fix.

melton-jason:
@maxxxxxdlp 
Do we want to merge this back to `agent-merging` before conflicts get out of hand? 😅 
You could argue that this change is within the scope of the pull request, but this pull request is already getting a tad too long.
It has accomplished what it sought out to accomplish: modify the endpoint so agent merging is only a single request, and the needed data (the JSON of the record to save, and an Array of id's from the clones')  is contained within the POST body.

maxxxxxdlp:
yes please, let's do that

melton-jason:
Ok. I'd say we can merge this back to `agent-merging` once @acwhite211 has gone through the review comments

acwhite211:
Let's go ahead now and merge this with the `agent-merging` branch and change the front-end code referencing the old api with the new record-merging api so that we can start testing again.

CarolineDenis (approved):

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -208,45 +207,46 @@ function Merging({
...
-              undefined,
-              'dismissible'
-            ).then(async () => {
-              /*
-               * Make requests sequentially as they are expected to fail
-               * (due to business rules). If we do them sequentially, we
-               * can leave the UI in a state consistent with the back-end
-               */
-              // eslint-disable-next-line functional/no-loop-statement
+            /*
```
maxxxxxdlp, line 221: This comment should no longer be here as we are no longer making requests sequentially

specifyweb/specify/views.py
```diff
@@ -501,36 +509,55 @@ def update_record(record: models.Model):
...
 
-    # Dedupe by deleting the agent that is being replaced and updating the old model ID to the new one
-    target_model.objects.get(id=old_model_id).delete()
-
+    # Dedupe by deleting the record that is being replaced and updating the old model ID to the new one
+    for old_model_id in old_model_ids:
+        target_model.objects.get(id=old_model_id).delete()
+
+    # Update new record with json info, if given
+    has_new_record_info = False if new_record_info is None else True
```
maxxxxxdlp, line 517: can be simplified to `has_new_record_info = new_record_info is not None`

maxxxxxdlp (approved):
Looks good. Two small refactoring things - can be resolved after merging this

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
melton-jason: Fixed in 2b10695

specifyweb/specify/views.py
melton-jason: Fixed in 2b10695

melton-jason:
> Perhaps the most concerning and perplexing symptom of this issue is how little it gives in terms of information to debug the problem. The only information that can be gleaned directly from the issue is the 405 (Method Not Allowed) response.
> Usually this might be caused by a misconfigured url or the wrong request type, but everything seems correct as far as these are concerned.

https://github.com/specify/specify7/pull/3474#issuecomment-1551534512

This was caused because the backend was returning an `http.HttpResponseNotAllowed()`

```python
# Determine which of the records will be assigned as old and new with the timestampcreated field
foreign_record_lst = foreign_model.objects.filter(**{field_name_id: new_model_id})
if foreign_record_lst.count() > 1:
    # NOTE: Maybe try handling multiple possible row that are potentially causes the conflict.
    # Would have to go through all constraints and check records based on columns in each constraint. 
    return http.HttpResponseNotAllowed('Error! Multiple records violating uniqueness constraints in ' + table_name) 
```
    
image

Raising an error instead of returning the HttpResponseNotAllowed would make the frontend and toasts behave normally

maxxxxxdlp:
@melton-jason I don't think back-end is at fault here. It's front-end's fault for not handling the error correctly.

There is code that tries to figure out if the error was generated by a permission system (thus display the permission error dialog) or by session time out (thus, prompt the user to sign in instead). It's concerning that this error didn't trigger either error dialog.

Relevant code:

https://github.com/specify/specify7/blob/bd2ba160a43bfb555757a8a0cc7c82095ac85e0e/specifyweb/frontend/js_src/lib/components/Errors/FormatError.tsx#L153-L177

Can you please take a look at what's going on there?

PR #3476, opened 2023-05-10
Create CSV export with only selected query rows

### LINKED ISSUE #2348: Query builder's Export to CSV ignores selected rows
From CSIRO:

> If you make a sub-selection on a query list and then click "create csv" it will export all rows, not just the selected ones (see images). This was tested on the current edge version of Specify 7

CarolineDenis:
Fixes #2348


CarolineDenis:
@acwhite211 could you create a new backend function to create CSV file that accepts an array of recordSet items instead of a record set id?

maxxxxxdlp:
@CarolineDenis This feature is not just about record sets, but for any query results

Also, there is no reason for this to be a back-end API - it would be easier to do this on the front-end:
The user can only select the query results that are already loaded, thus the front-end has access to them and can trigger a file download for user (`downloadFile()` function)

grantfitzsimmons:
@maxxxxxdlp @CarolineDenis Can this change be transferred to production as we have no timeline for release yet for `specify-network-integration`? That may mean creating a new PR

maxxxxxdlp:
@grantfitzsimmons
It's on this branch specifically because this branch modified the relevant code
specify network integration can be released as soon as it's tested - it has all the features I wanted it to have for v1
the collection-wide gbif maps are also there, but they are not linked to from anywhere yet

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -59,6 +65,33 @@ export function QueryExportButtons({
     });
   }
 
+  const [separator] = userPreferences.use(
+    'queryBuilder',
+    'behavior',
+    'exportFileDelimiter'
+  );
+
+  const selectedResults = resultsArray?.filter((item) => {
```
maxxxxxdlp, line 74: this all should not be done on every render - wasteful performance wise
put it into a function and only call that function when the export button was pressed

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -536,6 +545,8 @@ export function QueryBuilder({
                   getQueryFieldRecords={getQueryFieldRecords}
```
maxxxxxdlp, line 545: Looking at the code now, it might be simpler and better to call `<QueryExportButtons>` inside of `<QueryResultsWrapper>` or even `<QueryResuts>`

the `resultsArray` thing is a cludge - you are storing the same state in two different places
(FYI: it doesn't really need to be a state, a ref would be fine - still, it should probably be avoided entirely)

can you try out and see if that makes things simpler? that would also add the csv export feature to express search (although, only when some search results are selected)

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -59,6 +65,33 @@ export function QueryExportButtons({
...
+    if (item && typeof item[0] === 'number') {
```
maxxxxxdlp, line 75: please avoid implicit boolean conversion (`item && `)
this should have been written simpler as `typeof item?.[0] === 'number'`)

maxxxxxdlp: checking for first item being a number is a potential bug - if first query column is a numeric field, this will fail
you should instead check if query is distinct (using queryResource - see other usages for example)

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -59,6 +66,33 @@ export function QueryExportButtons({
     });
   }
 
+  const [separator] = userPreferences.use(
+    'queryBuilder',
+    'behavior',
+    'exportFileDelimiter'
+  );
+
+  function handleSelectedResults(): string {
```
maxxxxxdlp, line 75: Could you put a comment saying this is only called if the query is not distinct because you can't select rows in distinct queries?

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -87,16 +121,21 @@ export function QueryExportButtons({
...
+              ? doQueryExport(
+                  '/stored_query/exportcsv/',
+                  userPreferences.get(
+                    'queryBuilder',
+                    'behavior',
+                    'exportFileDelimiter'
+                  )
+                )
+              : downloadFile(
+                  'querySelectionExport.tsv',
```
maxxxxxdlp, line 135: Can you come up with more helpful name? camelCase name is for programmers, not for regular users 😄 

Take a look at something like this for inspiration: https://github.com/specify/specify7/blob/60dd907699ea82e2aeb42a1eaa69d352e745398e/specifyweb/frontend/js_src/lib/components/HomePage/AboutSpecify.tsx#L48
 be sure to include query name (and handle the case when query doesn't have a name because it's a new query) and include date to keep file name unique

Ideally, you should use the same naming convention as the back-end.
Though, I just noticed that back-end's naming convention is quite bad too.

Thus, afterward, could you also check all the places in the back-end where files are created/downloaded?

<details>
<summary>Hint on how to find them</summary>
In python, to create a file, you would do something like `with open(` - do global search on that.

</details>

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -31,6 +36,8 @@ export function QueryExportButtons({
     | (() => RA<SerializedResource<SpQueryField>>)
     | undefined;
   readonly recordSetId: number | undefined;
+  readonly resultsArray: RA<QueryResultRow | undefined> | undefined;
```
maxxxxxdlp, line 39: You don't need to include `Array` in variable names - instead, if variable name is plular, it's clear from that. I.e, you have `results` instead of `resutls` which signifies an array

maxxxxxdlp (changes requested):
After thinking futher about this code, I do like your solution of making selectedRows accessible to parent components as this might even be useful for other use cases in the future.

The only advice I have here is to change `resultsArray` from being a state to being a ref. Please think though implications of this change (i.e, what's the difference between state and ref, especially when it comes to triggering react re-renders) and why it's beneficial to do that change in here

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsWrapper.tsx
```diff
@@ -99,13 +103,16 @@ export function useQueryResultsWrapper({
...
+  selectedRows: [selectedRows, setSelectedRows],
+  resultsRef,
 }: ResultsProps): PartialProps | undefined {
   /*
    * Need to store all props in a state so that query field edits do not affect
    * the query results until query is reRun
    */
   const [props, setProps] = React.useState<
-    Omit<PartialProps, 'totalCount'> | undefined
+    | Omit<PartialProps, 'resultsArrayRef' | 'selectedRows' | 'totalCount'>
```
maxxxxxdlp, line 114: `resultsArrayRef` is no longer there. did you forget to rename it to `resultsRef`?

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -537,6 +546,8 @@ export function QueryBuilder({
...
                   queryResource={queryResource}
                   recordSetId={recordSet?.id}
+                  results={resultsRef.current}
```
maxxxxxdlp, line 549: Remember that changes to a ref don't trigger a re-render?
Thus, whatever prop value you give it, it might become stale.

Take a look at the `<QueryExportButton>` component and you will see that it only needs query results when the export button is clicked - it is not used during render, thus it doesn't need to be a state - that's exactly the reason why were able to convert from using state to a ref

<p>

Remember that you mutated the resultsRef object directly (by doing `resultsRef.current = ...`.
Thus, the object is mutated, but it's still the same object - the reference to an object remains unchanged, only the contents were updated - which means, you can pass the reference, not the value of the reference, and access the value of the reference only when you need it - when the button is clicked

</p>
</details> 

<p>

Replace
```ts
results={resultsRef.current}
```

with
```ts
resultsRef={resultsRef}
```

and fix the resulting type errors

</p>
</details>

maxxxxxdlp (changes requested):

carlosmbe:
Works. But exported files with only selected rows were .tsv
Not sure if collection managers are picky about that but maybe change the button to `Create TSV` in those situations

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsWrapper.tsx
CarolineDenis: yes

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -87,18 +127,27 @@ export function QueryExportButtons({
...
+                  userPreferences.get(
+                    'queryBuilder',
+                    'behavior',
+                    'exportFileDelimiter'
+                  )
+                )
+              : downloadFile(
+                  `${
+                    queryResource.isNew()
+                      ? 'new query export'
```
maxxxxxdlp, line 143: instead of `new query export`, which is not localized, and not informative, can you do something like this:

```ts
`${queryText.newQueryName()} - ${schema.models[baseTableName].label}`
```

maxxxxxdlp:
> Works. But exported files with only selected rows were .tsv Not sure if collection managers are picky about that but maybe change the button to `Create TSV` in those situations

Good catch! The code should not be hardcoded to use `.tsv` extension - you should either use `.csv` (which is a universal file extension for this type of data), or have a different file extension depending on the delimiter used (`.tsv` - tab, `.csv` - comma, `.psv` - pipe, and others)

The button however should always say `Create CSV` as CSV is far more recognizable to an average computer user than more obscure formats like TSV or PSV

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -59,6 +68,37 @@ export function QueryExportButtons({
...
+      subArray?.slice(1).join(separator)
+    );
+
+    const resultToExport = [
+      fields
+        .map((field) =>
+          generateMappingPathPreview(baseTableName, field.mappingPath)
+        )
+        .join(separator),
+      ...(joinedSelected ? joinedSelected : []),
```
maxxxxxdlp, line 96: no implicit conversion to boolean please

maxxxxxdlp: why not just do `joinedSelected ?? []` - safer and shorter

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -87,18 +127,27 @@ export function QueryExportButtons({
...
-                'behavior',
-                'exportFileDelimiter'
-              )
-            )
-          }
+          onClick={(): void => {
+            selectedRows.size === 0
+              ? doQueryExport(
+                  '/stored_query/exportcsv/',
+                  userPreferences.get(
```
maxxxxxdlp, line 134: since you now call `userPreferences.use()` you don't need to call `userPreferences.get` here anymore and can just use `separator` variable here

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -87,18 +127,27 @@ export function QueryExportButtons({
...
+                    queryResource.isNew()
+                      ? 'new query export'
+                      : queryResource.get('name')
+                  } - ${new Date().toDateString()}.tsv`,
+                  handleSelectedResults()
+                );
+          }}
         >
-          {queryText.createCsv()}
+          {commonText.export()}
```
maxxxxxdlp, line 150: not sure how I feel about this change. @grantfitzsimmons opinions?

maxxxxxdlp (changes requested):

carlosmbe:
> Works. But exported files with only selected rows were .tsv
> Not sure if collection managers are picky about that but maybe change the button to `Create TSV` in those situations

Also the naming format of the exported files differs:

query_results_2023-05-22T20_21_21.052320.csv

vs

taxon query - Mon May 22 2023.tsv

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
grantfitzsimmons: Create CSV is much clearer. I think we should keep that.

carlosmbe (commented):
Mainly related to maintaining consistency with the old `Create CSV`
Two things:
1- The old `Create CSV` lets users download the file through the notifications while selected query are automatically downloaded.
2- Naming format of the exported files still differs from the old `Create CSV`

maxxxxxdlp:
Yes, we should improve the back-end file name. But, Caroline created a separate issue for that, so it doesn't necessarily have to be fixed as part of this PR - https://github.com/specify/specify7/issues/3525

As far as using notifications:
1. The notifications system was used due to technical limitations (because it needs time to fetch all query results and generate the file) - here there is no reason to wait because the query results are already fetched (because user can only select query results that are fetched).
2. Front-end has not technical capability to send someone notifications, so we can't use that - only back-end has the power to create notifications at the moment.

Thus, this inconsistency is justified

carlosmbe (approved):

bronwyncombs:
`v7.8.12-prerelease` : The `Create CSV` button adds selected rows only to the download file. The file name format is Query Name - Day Date.

melton-jason:
@bronwyncombs can you expand upon this please? What exactly did you find while testing this?
Can you describe what you were doing (so we can recreate an issue)? 
Screenshots, Videos, or direct links are always helpful for this process as well!
Essentially, try and format your response like an Issue (if you don't know how issues are formatted, check out any of them from the Issues tab!). 
Because of this, it may be easier to open an Issue and link the associated Pull Requests (or other Issues) with `#` and then entering the ID or the name. 

I tested this on `v7.8.12-prerelease` and here are my results. 

When exporting a query to CSV that has not been saved, it exports correctly (only the selected rows).
However, there was an issue related to the name of the Query (most likely related to #3525) . 

The Query name is `New Query -_                        Collection Object - Fri Jun 02 2023.csv`. 
There seems to be extra space characters.
new_co_query

[New Query -_                        Collection Object - Fri Jun 02 2023.csv](https://github.com/specify/specify7/files/11637975/New.Query.-_.Collection.Object.-.Fri.Jun.02.2023.csv)

Exporting a saved Query seems to work as intended. The selected rows export properly, and the file is named correctly. 

saved_co_query

[CO Query - Fri Jun 02 2023.csv](https://github.com/specify/specify7/files/11637974/CO.Query.-.Fri.Jun.02.2023.csv)

PR #3480, opened 2023-05-11
Change the string in show form button depending on status

### LINKED ISSUE #3479: Attachment preview display wrong string in button

CarolineDenis:
Fixes #3479


specifyweb/frontend/js_src/lib/components/Attachments/Dialog.tsx
```diff
@@ -80,7 +80,9 @@ export function AttachmentDialog({
         <>
           <span className="-ml-4 flex-1" />
           <Button.Blue aria-pressed={showMeta} onClick={toggleShowMeta}>
-            {attachmentsText.showForm()}
+            {showMeta === true
```
maxxxxxdlp, line 83: Be careful about this, it's bad for accessibility.
After the change, consider how the screen reader will read this button:
when showMeta:
`Hide Form, checked`
when !showMeta:
`Show Form, not checked`

This is confusing. Compare this to prior behavior:
when showMeta:
`Show Form, checked`
when !showMeta:
`Show Form, not checked`
Much more clear about whether we are currently showing the form or not.

I understand why you made this change and it's easy to overlook this accessibility consideration. As a compromise, instead of your change, consider renaming `Show Form` to `Toggle Form`

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Attachments/Dialog.tsx
maxxxxxdlp: I was a bit wrong about this.

> Do not change the contents of the label on a toggle when the state changes. If a button label says "pause", do not change it to "play" when pressed. In this example, when the pressed state is true, the label remains "Pause" so a screen reader would say something like "Pause toggle button pressed".
>
> `<button aria-pressed="false">Pause</button>`
>If you want the label to toggle between "Paused" and "Play", don't use aria-pressed.

From https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed

Thus, we should either have aria-pressed and static label (like we did before), or changing label, but no aria-pressed

Also, there might be an even better alternative: https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/pressed/#alternative-technique-using-checkbox

maxxxxxdlp (approved):

CarolineDenis:
> How do I test this?

> Open attachment Gallery
> Open an attachment 
> Click on "hide form" in the top right corner
> See the changes

PR #3481, opened 2023-05-11
Highlight interaction menu item when in interaction forms

### LINKED ISSUE #3459: Side Bar Menu highlights the wrong item
- When clicking on: Interactions > new repository agreement => Data Entry is highlighted 
- When Clicking on Data Entry => Record Set is highlighted (due to the new feature introduced in (https://github.com/specify/specify7/pull/3296)

CarolineDenis:
Fixes #3459

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Forms/ShowResource.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/ShowResource.tsx
@@ -63,10 +63,11 @@ export function ShowResource({
   useErrorContext('resource', resource);
 
   useMenuItem(
-    typeof recordSet === 'object'
-      ? 'recordSets'
-      : interactionTables.has(resource.specifyModel.name)
+    typeof recordSet === 'object' &&
+      interactionTables.has(resource.specifyModel.name)
       ? 'interactions'
+      : typeof recordSet === 'object'
+      ? 'recordSets'
       : 'dataEntry'
   );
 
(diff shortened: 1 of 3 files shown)


specifyweb/frontend/js_src/lib/components/Forms/ShowResource.tsx
```diff
@@ -63,10 +63,11 @@ export function ShowResource({
   useErrorContext('resource', resource);
 
   useMenuItem(
-    typeof recordSet === 'object'
-      ? 'recordSets'
-      : interactionTables.has(resource.specifyModel.name)
+    typeof recordSet === 'object' &&
```
maxxxxxdlp, line 66: Is this change supposed to fix `When Clicking on Data Entry => Record Set is highlighted (due to the new feature introduced in`?

Your code change did not fix it. The `typeof recordSet === 'object'` condition doesn't make sense anymore since the record set is always defined. Also, your code change does not fix this still incorrectly highlighting the record set when opening any data entry table.

Think more about how to fix this issue. Think about what we changed in https://github.com/specify/specify7/pull/3296 and how it's changed should be reflected here

<p>

1. Undo your code change
2. Replace `typeof recordSet === 'object'` with `typeof recordSetId === 'number'`

</p>
</details>

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Forms/ShowResource.tsx
CarolineDenis: No this change was for the interactions. I didn't do the recordSet/dataE try yet

grantfitzsimmons (approved):

specifyweb/frontend/js_src/lib/components/Forms/ShowResource.tsx
maxxxxxdlp: again the code change in here is not correct. please read my comment again, including the
`The typeof recordSet === 'object' condition doesn't make sense`
part and the proposed solution

maxxxxxdlp (changes requested):
Code has faulty logic
As a result, this doesn't work correctly for record sets based on an interaction table

maxxxxxdlp (approved):

Issue #2136, opened 2022-09-12
Improve performance for pick lists based on another table


grantfitzsimmons:
> Pick lists that have records from another table as their items are very slow when the table has a lot of records (i.e, the taxon table)
> Pick lists like this should probably behave like a query combo box, and only start fetching items when you start typing into them

maxxxxxdlp:
From CSIRO:

> We saw that a ticket recently improved the performance of picklists, but we have one with ~10,000 values that is breaking the data entry form - it takes 10-30 seconds to load the option for the dropdown but when you click on it everything freezes and you have to close the browser tab. Were the tests done on picklists this large?
We also noticed when using the query builder that if you select an "equal" search on this field the dropdown loads fine there.
It would be nice to have the option to have something like the query combo box feature for picklists of this size - so rather than having the full list being able to start typing and have any matching values appear - would this be a reasonable enhancement request?
In the meantime are there any suggestions for changing the form to be able to cope with the large list?

maxxxxxdlp:
Adding virtual scrolling or infinite loading to Query Combo Boxes would be really good

Then add a way to render a pick list as a query combo box (thus, pick list will gain a new feature, and this issue will be fixed for pick lists)

Also, consider changing the code so that pick lists whose items are based on another table are rendered as a query combo box by default, unless the form definition explicitly says otherwise - this is because pick lists based on another table can contain thousands of records

@CarolineDenis for more context, see `Entire Table` and `Field From Table` pick list types - https://discourse.specifysoftware.org/t/user-defined-pick-lists/205

maxxxxxdlp:
Note, as a (bad) workaround, you can limit the maximum number of items that are retrieved by a pick list that is based on a field or table using the size limit option

In the code, it's used in here: https://github.com/specify/specify7/blob/4ab1ab97dd02079607113341312c89e15ab215fe/specifyweb/frontend/js_src/lib/components/PickLists/fetch.ts#L81-L84

Consider removing that option in the future when pick lists would be made to load items dynamically as needed

PR #3484, opened 2023-05-12
Remove front-end pick list size limit checks

### LINKED ISSUE #3482: Remove size limit enforcement for pick lists
Today, the size limit had a bug related to it, and that refocused our attention on whether the size limit is useful at all. The limit should just be removed if it only serves as a vessel for bugs!

from @maxxxxxdlp:
> Size limit was useful in sp6 as pick list items were added automatically
> not useful in sp7 at all
> and like Ben said, it's an "antifeature"

maxxxxxdlp:
It was buggy (see #3482) and didn't serve a purpose anymore. Also, back-end already removed these checks.

Fixes #3482


maxxxxxdlp:
Note, there is still one place where size limit is used. See https://github.com/specify/specify7/issues/2136#issuecomment-1545897213

chanulee1:
How do I test this?

maxxxxxdlp:
1. Create a pick list with size limit of 5
2. Make pick list not read only
3. Add 6 items to the pick list using the pick list form
4. Add the pick list field to the form
5. In the pick list field, enter a value that does not exist and see that the "Add" option is available
6. Click on "Add", save changes, and next time you use the field, the newly added value should be visible among pick list items

carlosmbe (approved):
Based on the test left in the comment. It works from a UX Pov

CarolineDenis (approved):

Issue #256, opened 2016-04-08
Shared Queries


jobdiogenes:
Its not really a bug, but a feature that could be very useful.
Since the database is very big, and we couldn't teach everyone hot to make queries.
Then, would be nice if we could share some queries.

maxxxxxdlp:
The current workaround for sharing a query is:
1. Open a query
2. Copy the URL from the address bar
3. Send it to another user

Then, a set person can execute the query, or clone it in order to make changes.

benritchie:
thanks for above workaround.  
Agree though, enhancing this, so queries can be shared is a key enhancement.  Doesn't need to be really complex - could we just share a default set of queries as an XML fragment, similarly to simple search config or something? - and manage through resource manager?

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/diplay-all-collection-object-records/791/2

grantfitzsimmons:
> Is it possible for all users to see the record sets and queries created by the administrator/manager of the database?

**Requested By:** Soraya at Eawag

grantfitzsimmons:
Another request for sharing record sets on the Community Forum

[Speciforum](https://discourse.specifysoftware.org/t/querying-record-set-created-by-a-different-user/946), requested by **RBGE**

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/querying-record-set-created-by-a-different-user/946/2

maxxxxxdlp:
This likely requires a schema change to add a boolean field indicating whether the query should be shared or not

mcruz-umich:
Sharing queries, record sets, and workbenches would be great!

grantfitzsimmons:
Apologies for the belated response, @philippeverley! High-level priorities are discussed internally by our development team and presented annually to our committees and board members. Occasionally, we have the capacity to take on smaller projects that receive broad support from member collections, like this one. However, our primary focus remains on the priority areas we committed to with our members.

I put some issues into a special "Grant's issue list" milestone, which serves only as a repository for other members on the team to see which issues I find particularly important to resolve!

PR #3501, opened 2023-05-16
Unhide string comparison operators in QueryBuilder

### LINKED ISSUE #1035: Catalog Number field does not offer expected operators in query
Reported by Santa Barbara: 

Interesting issue when working with non-integer indices. One of our databases has a “CatalogNumber” field that is populated with mostly integers, but it is in fact a string type (leading zeros and infrequent alphanumerics). On Specify6, we were able to use operators like “<“ with this field. In specify7, that query flags with a red field, and returns a set of records that aren’t readable in the records window. It looks like it gets passed to the database, which returns the same number of records each time. 

  More serious, it won’t work with “=“ either. Screen shot:

image

The way the same field looks in 6 in the query:
image

melton-jason:
Fixes #1035

### Overview/Description

Adds a new Preference in the `QueryBuilder -> Behavior` section of User Preferences called `Show comparison operators for text-based fields` (default is unechecked/false)

new_query_pref

When checked, the preference unhides the following operators for text fields in the Query Builder:

```
>, Greater than
<, Less than
>=, Greater than or equal to
<=, Less than or equal to
```

All of these were able to be accomplished previously using the BETWEEN operator, but doing so required knowledge of how MySQL used the BETWEEN operator 'behind the scenes' (which itself would require knowledge of ASCII characters and their values). 

This preference can lead to potentially nonsensical results, especially when numeric strings are sorted lexicographically  (see https://github.com/specify/specify7/pull/3501#pullrequestreview-1431053184). 
This potential confusion is the primary motivation behind putting this functionality behind a preference and making it non-default. 

#### Edge Case Behaviors/Notes

_**Viewing existing queries which have comparison-based filters on a text field when the preference is disabled**_

If you do not have the preference checked but encounter an existing comparison filter on a text field, then Specify respects the filter on the field: leaving it untouched. However, if you then switch the filter to a non-comparison operator, then the previous comparison filter will be removed from the list of available filters; only reloading the page, enabling the preference, or generally leaving the page and coming back (i..e, reloading the Query) will allow the user. to"go back" to the previous filter.  

https://github.com/user-attachments/assets/8eeb034e-b4b1-4262-ae0a-6d64b6a673a4

Let me know if this edge case deserves more attention! A more user-friendly implementation for example might keep the previous comparison-based filter in the list of possible filters to choose from: until the query is saved or the web-page is left. 

_**Viewing queries which have comparison-based filters on a text field in older versions**_

Support for this feature is not retroactive: if such a filter exists on a query, the filter will be replaced with `Any` when viewing the query in older versions of Specify 7. 
When viewing the Query in Specify 6, Specify 6 will replace the filter with `Contains` (see https://github.com/specify/specify7/pull/3501#pullrequestreview-1431047000). The Query, however, should be hidden in Specify 6 by default (see https://github.com/specify/specify7/pull/3501#issuecomment-1553170095)

@specify/ux-testing @specify/dev-testing 
Feel free to give any feedback on the wording of the preference, the default value, etc.! 😄 

We can also consider defining this User Preference as a Remote Preference instead, if the scope of the preference should be broader than user (which might become important when/if [Shared Queries (#256)](https://github.com/specify/specify7/issues/256) are implemented)

### Checklist

- [ ] Add automated tests

### Testing instructions

- With the `Show comparison operators for text-based fields` preference set to false (the default behavior)
- Open a new or existing query with a table containing a text field which is not CatalogNumber (e.g., `text1`, `text2`, `remarks`)
  - CatalogNumber should already contain the comparison operators: regardless of preference value 
  - If you do not have a text field in mind, https://files.specifysoftware.org/schema/version/2.10/ 
- Add the text field to the query if it is not already present
- Save the Query
- [ ] Click on/open the Filter Type dropdown (Any, Like, Equal, Between, etc.) and ensure that the Greater Than and Less Than filters are not shown
- Go into User Preferences and under `QueryBuilder -> Behavior` check/enable the `Show comparison operators for text-based fields` preference
  - Make sure the title and description of the preference makes sense!  
- Without reloading the page, go back to the previous Query and click on/open the Filter Type dropdown
- [ ] Ensure the Greater Than and Less Than filters are present in the dropdown
- Select one of the Greater Than or Less Than filters on a text field and save the Query
  - Optionally, ensure that results are returned as expected* when using comparison-based filters on text fields
- Navigate back to User Preferences and disable the `Show comparison operators for text-based fields` preference 
- [ ] Navigate back to the previously used Query and ensure that the comparison-based filter is present (and still respected by the query results) until changed


maxxxxxdlp (approved):

grantfitzsimmons:
Interestingly, the test panel keeps saying this is the Git SHA anytime I deploy this PR:
https://github.com/specify/specify7/commit/8133b0fa8233f37a98b6c404fd0e2d85e38959b5

image

grantfitzsimmons (commented):
image

Works as expected. Displays "Contains" in Specify 6, so this would further stray from what we consider "interoperability" with 6.

grantfitzsimmons (commented):
image

Another comment– string searching is still non-sensical to non-technical users. See this example.

melton-jason:
> Works as expected. Displays "Contains" in Specify 6, so this would further stray from what we consider "interoperability" with 6.

Interesting. I thought Specify 6 had the `>` and `<` symbols for strings, as shown by the screenshot in https://github.com/specify/specify7/issues/1035#issue-1030804570. Or does 6 change what is available for Query operations based on the 'type' of Catalog Number? 
`>=` and `<=` would not be supported in 6 as well. 

---

> image
> Another comment– string searching is still non-sensical to non-technical users. See this example.

Yes, it is based on comparing the ASCII values of each character, which leads to results like this. 
Perhaps a warning explaining this behavior would be okay? 
I am not sure how useful this would be, but maybe a feature to treat a field in the Query Builder as a different data type could be considered. As in, for this query, treat this text field as numeric (or vice versa).

maxxxxxdlp:
> Works as expected. Displays "Contains" in Specify 6, so this would further stray from what we consider "interoperability" with 6.

Interesting. Could it be that the same filter id means different things for different data types in sp6

melton-jason:
> Interesting. Could it be that the same filter id means different things for different data types in sp6

The comparator lists are stored in an Array in Specify 6. There are a lot of helper functions in [QueryFieldPanel.java](https://github.com/specify/specify6/blob/master/src/edu/ku/brc/specify/tasks/subpane/qb/QueryFieldPanel.java) to get the correct comparison operators. (If you are interested, [here](https://github.com/specify/specify6/blob/945d28f00288d2d64979848fb2fb63fda34bdd1b/src/edu/ku/brc/specify/tasks/subpane/qb/QueryFieldPanel.java#L911-L1049) are the relevant methods. Though I don't recommend going through them). 

The comparison operators are determined on the Data Class (type) for the field (when the field is not a Picklist or Catalog Number. There is special logic for both of those cases). 
There are hardcoded Operators for Strings, Booleans, and Timestamps. Every other data class gets the same operators. 

Here is the Specify 6 logic for determining the comparison operators for the Catalog Number field...

https://github.com/specify/specify6/blob/945d28f00288d2d64979848fb2fb63fda34bdd1b/src/edu/ku/brc/specify/tasks/subpane/qb/QueryFieldPanel.java#L921-L929
```java
        //CatalogNumber needs special treatment - works better as a number.
        //And other fields? Not sure how to tell. Maybe the formatter?????
        if (fieldInfo != null && fieldInfo.getName().equalsIgnoreCase("catalognumber") 
                && fieldInfo.getTableInfo().getClassObj().equals(CollectionObject.class))
        {
            if (fieldInfo.getFormatter() != null && fieldInfo.getFormatter().isNumeric())
            {
                return getComparatorListForClass(Number.class);
            }
            OperatorType[] stringCmps = getComparatorListForClass(String.class);
            OperatorType[] result = new OperatorType[stringCmps.length + 2];
            int c = 0;
            for (OperatorType ot : stringCmps)
            {
                result[c++] = ot;
            }
            result[c++] = SpQueryField.OperatorType.GREATERTHAN;
            result[c++] = SpQueryField.OperatorType.LESSTHAN;
            //result[c++] = SpQueryField.OperatorType.BETWEEN;
            return result;
        }
```
It quite literally is an amalgamation of hardcoded operator types based on the formatter for the field. 

I am not sure on the logic, but more than likely, it is simply receiving a number which it interprets as the index for the Operator List. The 'index' in this case results in `Contains` being shown...

Because of this behavior, there will be a different operator in the query (for Catalog Number) depending on the formatter for the field. 

All other string fields look like they'll display `Contains`

maxxxxxdlp:
🤦 🤦 🤦 🤦 🤦

melton-jason:
We would not be able to make this compatible with Specify 6 without an update to 6. 
Specify 6 only supports the [following comparison operations for string fields](https://github.com/specify/specify6/blob/945d28f00288d2d64979848fb2fb63fda34bdd1b/src/edu/ku/brc/specify/tasks/subpane/qb/QueryFieldPanel.java#L953-L962) :

Contains
Like
Equals
In
Between
Empty

We can close this pull request if the incompatibility is enough to warrant not implementing other operators in Specify 7.

maxxxxxdlp:
Sp7 Query builder is already not fully compatible (relative dates, OR filters, "current user" placeholder, ability to explicitly pick formatter aggregator)
Vinny added code that detected if the query has any sp7 exclusive features, and if so, hides the query in sp6 - if you think that's appropriate here, we could do the same - https://github.com/specify/specify7/blob/e54d84a3f437f0dd1f89b8ddd9b160981fe74cf2/specifyweb/frontend/js_src/lib/components/QueryBuilder/helpers.ts#L357

bronwyncombs (approved):
Operators are understandable and working as expected!
Screen Shot 2023-06-07 at 10 28 02 AM

melton-jason:
> Operators are understandable and working as expected! Screen Shot 2023-06-07 at 10 28 02 AM

Hi @bronwyncombs thank you for the review!
However, could you be a little more thorough in your testing, please? The catalog number field is a strange field in that it is very commonly treated as numeric (even though it is a string data type in the database). 
This pull request is specifically about the comparison operators for string-type fields. You can see the data type of a field in the Schema Config or Datamodel Viewer. Fields like `text` and `remarks` are almost always string fields.

Can you test the following operators on a string field that is not Catalog Number and ensure they work? DO they make sense? 
```
>, Greater than
<, Less than
>=, Greater than or equal to
<=, Less than or equal to
```

bronwyncombs:
@melton-jason thanks for catching that and letting me know! 

Good News: I did the test again on the string fields used for text in the forms and I believe it still checks out. 

Screen Shot 2023-06-07 at 6 15 15 PM
https://github.com/specify/specify7/assets/135047322/bde700c3-d66d-4c95-a714-a345b07a11d1

Does that look better?

grantfitzsimmons:
> Yes, it is based on comparing the ASCII values of each character, which leads to results like this.
> Perhaps a warning explaining this behavior would be okay?
> I am not sure how useful this would be, but maybe a feature to treat a field in the Query Builder as a different data type could be considered. As in, for this query, treat this text field as numeric (or vice versa).

@bronwyncombs mentioned this as being questionable as-is. Can we discuss this as a group before merging?

bronwyncombs (changes requested):
I don't think this should be implemented without including a preference to unhide or hide these operators for string fields. Unhiding them all by default may impact the non-technical users' ability to understand query building, causing unnecessary confusion and support requests.

grantfitzsimmons:
Another request for this ability has come in from The College of Idaho– can this be added as a preference and included in `v7.9.2`?

grantfitzsimmons:
This should only be merged _if_ it is a preference that is not enabled by default.

Areyes42 (approved):
### Testing instructions
-  [x] test that the preference "Show comparison operators for string" in QB section works
***
Looks good! Query search still works as expected with the option enabled/disabled.
Screenshot 2024-05-24 at 1 04 11 PM

melton-jason:
> This branch seems to have broken #4941 again. Screenshot 2024-05-24 at 2 46 15 PM [Specify 7 Crash Report - 2024-05-24T19_46_19.305Z.txt](https://github.com/specify/specify7/files/15439305/Specify.7.Crash.Report.-.2024-05-24T19_46_19.305Z.txt)
> 
> It was fixed by #4951 and isn't present in edge anymore, just this branch.

@lexiclevenger 
https://github.com/specify/specify7/pull/3501#pullrequestreview-2077799688

Should be fixed

lexiclevenger (approved):
- test that the preference "Show comparison operators for string" in QB section works
---
Looks good now! The preference is present and functioning.

specifyweb/frontend/js_src/lib/components/QueryBuilder/FieldFilter.tsx
```diff
@@ -396,6 +397,12 @@ function In({
   );
 }
 
+const showComparisonOperatorsForString = userPreferences.get(
```
maxpatiiuk, line 400: (optional) since this is accessed in global scope, front-end will try to access this value before user preferences are fetched from the back-end. that means:
- if this is the first time specify is opened, `showComparisonOperatorsForString` will equal to the default value , regardless of what is actually set in the preferences
- if this is not the first time specify is opened (or it was opened for a different user before), the preferences that were previously cached will be used
  - that means, if you switched user account, the `showComparisonOperatorsForString` will equal to the pref value for the previous user
  - if the pref was changed on a different machine (or default prefs app resource was changed), the `showComparisonOperatorsForString` would be stale until page is reloaded again

solution:
- wrap `queryFieldFilters` in `f.store()` and replace `showComparisonOperatorsForString` variable with `showComparisonOperatorsForString` function - this way, by the time `queryFieldFilters` is accessed for the first time (which should be after preferences are loaded - but double check that), the preferences would already be loaded

specifyweb/frontend/js_src/lib/components/QueryBuilder/helpers.ts
```diff
@@ -376,6 +376,20 @@ const containsRelativeDate = (
       fieldSpec.datePart === 'fullDate'
   );
 
+const containsExpandedStringComparisons = (
```
maxpatiiuk, line 379: hmm, I though this was already supported in sp6 too but removed by me in 7.6
could you double-check?

maxpatiiuk (changes requested):

emenslin (approved):
- [x] test that the preference "Show comparison operators for string" in QB section works 
- Logout and login as another user
- [x] Make sure the preference value updates to that of the logged in user and does not use that of the previous user's

Preference does not carry over to other users!

lexiclevenger:
- [x] test that the preference "Show comparison operators for string" in QB section works

- Logout and login as another user

- [x] Make sure the preference value updates to that of the logged in user and does not use that of the previous user's
---
Looks good! The preference value updates in accordance with each user

alesan99:
- [x] test that the preference "Show comparison operators for string" in QB section works 
- Logout and login as another user
- [x] Make sure the preference value updates to that of the logged in user and does not use that of the previous user's
Loading up a Query with that uses one of these operators after you've disabled the option sets it to 'Any' which could break the saved query, but this seems like an option you would enable once and never disable again so it seems fine.

combs-a (dismissed):
- [X] test that the preference "Show comparison operators for string" in QB section works 
- Logout and login as another user
- [X] Make sure the preference value updates to that of the logged in user and does not use that of the previous user's

Looking good! Tested the comparisons with numbers in the Text3 field and it worked perfectly, as well as the preferences being respected.

I did get this result as well when mixing numbers and letters since we're comparing ASCII.

image

### Requested changes:
Maybe adding to the note under the preference to make sure people know about the strange behavior to prevent unnecessary bug reporting or confusion? I'm not sure how to word it concisely, though. 😅

If others don't agree w/ this idea, then I'm fine with approving as-is. 

Good work! 👍

specifyweb/frontend/js_src/lib/components/QueryBuilder/FieldFilter.tsx
```diff
@@ -440,16 +449,21 @@ export const queryFieldFilters: RR<
...
-    types: ['text', 'number', 'date', 'id'],
   },
   greater: {
     id: 2,
     label: queryText.greaterThan(),
     description: undefined,
     renderPickList: false,
-    types: ['number', 'date', 'id'],
+    types: filterArray([
+      showComparisonOperatorsForString() ? 'text' : undefined,
```
maxpatiiuk, line 462: `showComparisonOperatorsForString` using an f.store is nice, but it's not enough to fix the bug
the `queryFieldFilters` is still created only once, and in the global scope - so it will call `showComparisonOperatorsForString` before preferences are fetched

you need to wrap `queryFieldFilters` in f.store too

maxpatiiuk (changes requested):

grantfitzsimmons:
@CarolineDenis Is this still going to be in 7.9.9?

specifyweb/frontend/js_src/lib/components/QueryBuilder/FieldFilter.tsx
melton-jason: Fixed in [`d2845cf` (#3501)](https://github.com/specify/specify7/pull/3501/commits/d2845cf9016b442404f090d568477497f17a8cde)

pashiav (approved):
- [X] Click on/open the Filter Type dropdown (Any, Like, Equal, Between, etc.) and 
- [x] Ensure the Greater Than and Less Than filters are present in the dropdown
- [X] Navigate back to the previously used Query and ensure that the comparison-based filter is present (and still respected by the query results) until changed
---

Nice job!

---

>   - Make sure the title and description of the preference makes sense!  


I think it's good! The description in preferences explains it well.


<br>

✅ Did not run into https://github.com/specify/specify7/pull/3501#pullrequestreview-2083414167:

> I did get this result as well when mixing numbers and letters since we're comparing ASCII.

CarolineDenis (approved):

emenslin (approved):
- [x] Click on/open the Filter Type dropdown (Any, Like, Equal, Between, etc.) and ensure that the Greater Than and Less Than filters are not shown
- [x] Ensure the Greater Than and Less Than filters are present in the dropdown
- [x] Navigate back to the previously used Query and ensure that the comparison-based filter is present (and still respected by the query results) until changed

Looks good!

Areyes42 (approved):
-   [x] Click on/open the Filter Type dropdown (Any, Like, Equal, Between, etc.) and ensure that the Greater Than and Less Than filters are not shown
-  [x]  Ensure the Greater Than and Less Than filters are present in the dropdown
-  [x]  Navigate back to the previously used Query and ensure that the comparison-based filter is present (and still respected by the query results) until changed
***
👍👍

melton-jason:
Thanks for the reviews everyone! I will merge this PR with production now, given that there are three UX/UI testing approvals and one other Developer approval. 
Given that there are still three old reviews requesting changes, I will dismiss these for now, but if there are any problems with the PR feel free to leave a comment here and open an Issue!

PR #3516, opened 2023-05-19
Allow to sort fields in schema config by their hidden value

### LINKED ISSUE #2738: Sort by hidden/unhidden fields in schema config
When editing the schema config, there are occassionally situations where I need to see all unhidden fields first. Having an option in the schema config menu to display those at the top would be nice. 

This would also make it easy to hide fields you don't want to be visible.

CarolineDenis:
Fixes #2738

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/SchemaConfig/Fields.tsx
+++ b/specifyweb/frontend/js_src/lib/components/SchemaConfig/Fields.tsx
@@ -7,7 +7,7 @@ import { schemaText } from '../../localization/schema';
 import type { RA } from '../../utils/types';
 import { sortFunction, split } from '../../utils/utils';
 import { H3 } from '../Atoms';
-import { Select } from '../Atoms/Form';
+import { Input, Label, Select } from '../Atoms/Form';
 import type { SerializedResource } from '../DataModel/helperTypes';
 import type { SpecifyTable } from '../DataModel/specifyTable';
 import type { SpLocaleContainerItem } from '../DataModel/types';
@@ -32,6 +32,7 @@ export function SchemaConfigFields({
     sortedItems,
     (item) => table.getField(item.name)!.isRelationship
   );
+  const [isHiddenFirst, setIsHiddenFirst] = React.useState(false);
   return (
     <SchemaConfigColumn header={schemaText.fields()} id={id('fields-label')}>
       <Select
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/SchemaConfig/Fields.tsx
```diff
@@ -32,6 +32,7 @@ export function SchemaConfigFields({
     sortedItems,
     (item) => table.getField(item.name)!.isRelationship
   );
+  const [isHiddenFirst, setIsHiddenFirst] = React.useState(false);
```
maxxxxxdlp, line 35: Your change only fixed this for fields, not for relationships.
Also, your change caused some code duplication.
Consider how this can be simplified and improved by sorting the fields instead

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

maxxxxxdlp:
Looking at the code, there should be a new checkbox below the list of fields in the schema config

CarolineDenis:
@carlosmbe 
Screenshot 2023-05-22 at 2 27 57 PM

carlosmbe:
Thanks for the pointer Max and Caroline. Greatly appreciated <3
It works fine. 
If I had a complaint it would be that it's a bit hard to tell the order initially(Descending vs Ascending)
And where the split between hidden and not hidden is

carlosmbe (approved):
It solves the issue #2738
But since it was an enhancement. I'm not sure if @grantfitzsimmons has any changes they'd like so I added them as a reviewer.

maxxxxxdlp:
Yes, we should consider adding ` (hidden)` to the names of fields that are hidden

maxxxxxdlp (changes requested):
See https://github.com/specify/specify7/pull/3516#issuecomment-1558414842

specifyweb/frontend/js_src/lib/components/SchemaConfig/Fields.tsx
```diff
@@ -24,14 +24,23 @@ export function SchemaConfigFields({
   readonly onChange: (index: number) => void;
 }): JSX.Element {
   const id = useId('schema-fields');
-  const sortedItems = Object.values(items ?? []).sort(
-    sortFunction(({ name }) => name)
-  );
+  const [isHiddenFirst, setIsHiddenFirst] = React.useState(false);
```
maxxxxxdlp, line 27: 1. store this in local storage (see other examples in specify for how to do it)
2. default should be true

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/SchemaConfig/Fields.tsx
```diff
@@ -91,6 +81,23 @@ export function SchemaConfigFields({
   );
 }
 
+export function SchemaConfigFieldsList({
+  fields,
+}: {
+  fields: readonly SerializedResource<SpLocaleContainerItem>[];
```
maxxxxxdlp, line 87: for consistency, consider using `RA<T>` instead of `readonly T[]` syntax

a blurb from Google Bard on consistency:

> Consistency in code is important because it leads to predictability. Consistency makes code more readable and understandable, which is important for both the developer who wrote the code and any other developers who may need to work with it in the future. Consistency in naming conventions, indentation, and formatting can also make it easier to spot errors and bugs in the code.
>
> Consistent coding standards also facilitate seamless collaboration among team members, improving productivity and reducing conflicts. When code is consistent, maintaining and updating it becomes easier, even for developers who weren't involved in the initial development.
>
>A consistent codebase is easier to work with because you get used to its patterns. Your brain creates shortcuts that help you read its code faster.

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

PR #3527, opened 2023-05-22
Resize Save button in Schema Config and Mapper

### LINKED ISSUE #3524: Wrong size `Save` button used in query builder
During the recent refactoring of button names, the save buttons were made to use a different component. Unfortunately, in the process, the sizes of save buttons changed in some places. One example is the schema config header.

Screenshot 2023-05-22 at 13 00 43

Please fix the sizing of this button (see older version of specify if need to compare how it looked) and review all the other save buttons to check where size might be been changed mistakingly

CarolineDenis:
fixes #3524

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/SchemaConfig/Components.tsx
+++ b/specifyweb/frontend/js_src/lib/components/SchemaConfig/Components.tsx
@@ -10,6 +10,7 @@ import { H2 } from '../Atoms';
 import { Button } from '../Atoms/Button';
 import { Select } from '../Atoms/Form';
 import type { SchemaData } from './schemaData';
+import { className } from '../Atoms/className';
 
 export function SchemaConfigHeader({
   languages,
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/SchemaConfig/Components.tsx
```diff
@@ -33,7 +33,9 @@ export function SchemaConfigHeader({
         {schemaText.changeBaseTable()}
       </Button.Small>
       <span className="-ml-2 flex-1" />
-      <Button.Save onClick={handleSave}>{commonText.save()} </Button.Save>
+      <Button.Save className="!py-1 !px-2" onClick={handleSave}>
```
maxxxxxdlp, line 36: noo, we talked about this. don't duplicate styles like that - if we change them (and we do from time to time), this place would remain unchanged.

Remember how you could use Button.Small and variant={} instead?

specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
```diff
@@ -389,6 +389,7 @@ export function Mapper(props: {
             <Button.Save
               disabled={!state.changesMade}
               onClick={(): void => handleSave(false)}
+              className="!py-1 !px-2"
```
maxxxxxdlp, line 392: same suggestion

maxxxxxdlp (changes requested):

maxxxxxdlp:
Wrong size:
image

Correct size:
Screenshot 2023-05-23 at 14 31 34

(It should have less padding to be consistent with how toolbars in other places look)

carlosmbe (approved):

Issue #3530, opened 2023-05-24
Log in screen improvememts


grantfitzsimmons:
> Can we change this login screen to shorten the dialog text to:
> Username and/or password not recognized.
> Both fields may be case sensitive.
> (so that the text appears on two lines only-- that's the main thing)
> Should the word "Language"
> be on top of the pick list at the top, like Username and Password are below, for consistency?

Change from this:
![image](https://github.com/specify/specify7/assets/37256050/482aaeb0-f941-4197-a4b6-728b97008f94)

To this:
![image](https://github.com/specify/specify7/assets/37256050/2547763b-f3bb-45a3-842d-7561b5c67acb)

Make sure spacing is correct on the padding between language and the language chooser if this is implemented

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/specify-7-8-12-release-announcement/1202/1

PR #3531, opened 2023-05-24
Add language subheading to login screen

grantfitzsimmons:
Solves #3530


maxxxxxdlp (approved):
We already had a "Language" string in a different file, so no need to add another one. Fixed in dda7524

The rest is good!
Let's make a full-time developer out of you! you would do great

chanulee1 (approved):
Language subheading is present on login screen. Login screen functionality working as expected.

Issue #3126, opened 2023-03-08
Auto-generated query title has only 1 quote


grantfitzsimmons:
<img width="1271" alt="image" src="https://user-images.githubusercontent.com/37256050/223848450-7acdc7ab-a1a4-45a9-bb51-10b4aadf0471.png">

```
Query: Collection Object using "Myxinidae
```

Should have another quote at the end

https://sp7demofish.specifycloud.org/specify/query/fromtree/taxon/250/

melton-jason:
Has this issue been fixed?

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/specify-7-8-12-release-announcement/1202/1

PR #3532, opened 2023-05-24
Show quotes after auto-generated query names

grantfitzsimmons:
Solves #3126

Changed
```
    'en-us': '{tableName:string} using "{nodeFullName:string}"',
```

to this

```
    'en-us': `{tableName:string} using "{nodeFullName:string}"`,
```

Will this have any implications for Weblate? In my testing it solves the quote issue.


maxxxxxdlp (approved):
Looks good. Weblate won't have a problem with you manually changing localization files

chanulee1 (approved):
Quotes present after auto-generated query names.

PR #3538, opened 2023-05-25
xml-editor bug fixes

melton-jason:
Fixes the following bugs:
- (Agent Merging) Can no longer save the merged resource from the Record Preview view
- Catalog Number formatter validation is no longer always the 'Catalog Number' UIFormatter
-  Filters out cases when a dependent resource/collection is null or undefined when getting its save blockers
- Use `getField()` rather than `strictGetField()`  because `strictGetField()` was spamming the console with errors getting fields that don't exist on the table, like 'resource_uri'
-  Do not show error toast when WebLink is defined on the form but not in the schema config 

Refactoring Changes:
- Rename the component to carry over fields for a merge 'TransferButton' (was MergeButton). We already have a component named MergeButton, the button to go initiate the merge. 

I will add more fixes to this branch as I encounter and fix them.


specifyweb/frontend/js_src/lib/components/DataModel/schemaExtras.ts
```diff
@@ -121,6 +122,8 @@ export const schemaExtras: {
       catalognumber.getFormat = (): string | undefined =>
         schema.catalogNumFormatName ||
         LiteralField.prototype.getFormat.call(catalognumber);
+      catalognumber.getUiFormatter = (): UiFormatter | undefined =>
```
maxxxxxdlp, line 125: hmm, this doesn't seem necessary

The default implementation of getUiFormatter is:
```js
  // Returns the UIFormatter for the field specified in the schema config.
  public getUiFormatter(): UiFormatter | undefined {
    return getUiFormatters()[this.getFormat() ?? ''];
  }
```

Which is identical to your implementation.
Is `this.getFormat()` referring to a different function

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -112,7 +109,8 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
     fieldName: string & keyof SCHEMA['fields'],
     results: RA<BusinessRuleResult<SCHEMA>>
   ) {
-    const field = this.resource.specifyTable.strictGetField(fieldName);
+    const field = this.resource.specifyTable.getField(fieldName);
```
maxxxxxdlp, line 112: Instead of this change, check if field is in `specialFields` and if so, don't do anything

maxxxxxdlp (changes requested):

maxxxxxdlp:
Should we go ahead and merge this before too many merge conflicts happen?

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
melton-jason: Is the change in b91ad63 sufficient?

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -109,8 +110,8 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
     fieldName: string & keyof SCHEMA['fields'],
     results: RA<BusinessRuleResult<SCHEMA>>
   ) {
-    const field = this.resource.specifyTable.getField(fieldName);
-    if (field === undefined) return;
+    if (specialFields.has(fieldName)) return;
```
maxxxxxdlp, line 113: Not a big deal, but a note for the future:
To follow a more functional-programming style, instead of having a function decide on its own that it doesn't want to do anything (and just `return;`), the code that calls the function should decide whether the function should be called.
In this case, it means moving this check from here to checkField

Of course this is not always practical. i.e, if processCheckFieldResults where called in 5 different places (which it isn't), then it would be easier to have the check in processCheckFieldResults rather than duplicate it 5 types. But, functional programming has a solution even for that case - either abstract away the check into a function (thus de-duplicating the logic) OR create a function that is build on top of processCheckFieldResults and will decide whether to call processCheckFieldResults

Again, it's a minor thing, so not worth fixing here, just something to keep in mind for the future

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
melton-jason: Thank you for the advice! I greatly appreciate it! :smiley: 
I'm going through and cleaning up a lot of the functions in `businessRules.ts` as many of them (especially `uniqueIn(...)`) had a lot of components and were obscure. 
We can go ahead and merge this then if you want, as I can include this requested change in that commit (and we don't have to worry about file conflicts).

Issue #3341, opened 2023-04-13
Formatters containing aggregators split into different results in querybuilder


chanulee1:
`Discipline` formatter containing `Collection` aggregator is split into different results in the querybuilder, one for each collection. Queried  for formatted discipline records.

https://user-images.githubusercontent.com/103299204/231821923-f2bba9c2-b2ce-42c3-a429-9d8f76b2fae9.mp4

https://coldfish-xml-editor.test.specifysystems.org/specify/query/209/

maxxxxxdlp:
Looks like aggregator does not get called when it should
Not sure if this was intentional or not.
@chanulee1 can this be replicated on production?
@realVinayak could you please take a look at this (since it's about query results, it's related to back-end)

chanulee1:
Yes, I was able to replicate this on production.

https://coldfish-edge.test.specifysystems.org/specify/query/209/

maxxxxxdlp:
It's very important to fix this issue before stats page release.
This bug resulted in query on collection object returning 3 billion records!

realVinayak:
Fixed on use-dynamic-ranks

Issue #3266, opened 2023-03-29
Detect circular formatters/aggregators (front-end)


maxpatiiuk:
I.e, if accession includes accession agent formatter, and accession agent formatter includes accession formatter.

The fix:
If the same resource appears the 2nd time in the formatting, exclude relationship fields from the formatter (thus, Accession will be formatted, Accession -> Accession Agents will be formatted, Accession -> Accession Agents -> Accession -> Accession Number will be formatted, but Accession -> Accession Agents -> Accession -> Accession Agents won't be

With the formatter/aggregator editor being available and so easy to use, and showing a live preview of formatter on any change, this issue is far more likely to happen. And once it does, the front-end will be rapidly sending requests to the back-end.

PR #3545, opened 2023-05-30
Apply formatters and aggregators properly

### LINKED ISSUE #4444: -to-many relationships are not being aggregated
-to-many relationships are not being aggregated when a table format references them (for example, `determinations` or `preparations` do not display aggregated when added to a table format).
Recursive formatters are also not limited, but once aggregation can be displayed, this must be handled to prevent future errors.

realVinayak:
Fixes #4444 

Testing instructions:
Multiple cases that need to be tested, so I'll recommend doing them incrementally. I'll add the test cases for all of them soon, but here are they for the first case:

**To-many not being aggregated when referenced in a formatter**
1. Open record formatter visual editor for a table.
2. Add a field which is to-many (for example, Determinations for Collection Object). It would appear as (aggregated)
3. If using Determinations for Collection Object, make a new collection object with more than 1 Determination.
4. Query that collection object, using (formatted) as the field. You should see that both of the determinations have been included in a single line.
Example:
image
My collection object has two determinations, one with Taxon named "Abantennarius will be added" and other with name "Aborichthys aa" and they both appear (since formatter of my determination is taxon)
image


specifyweb/stored_queries/format.py
```diff
@@ -89,96 +101,125 @@ def pseudo_sprintf(self, format, expr):
...
             return query, literal(_("<Formatter not defined.>"))
 
         if single:
             value, expr = cases[0]
         else:
-            control_field = getattr(orm_table, switchNode.attrib['field'])
-            expr = case(cases, control_field)
-
+            query, formatted, swith_field_spec = make_expr(query, switchNode.attrib['field'], {})
+            def case_value_convert(value): return value == 'true' if swith_field_spec.get_field().type == 'java.lang.Boolean' else value
```
maxxxxxdlp, line 182: huh, didn't know python allowed that! I though it required return to be on a new line

specifyweb/stored_queries/format.py
```diff
@@ -89,96 +101,125 @@ def pseudo_sprintf(self, format, expr):
...
+            cases = [(case_value_convert(value), expr) for (value, expr) in cases]
+            expr = case(cases, formatted)
         return query, blank_nulls(expr)
 
-    def aggregate(self, query: QueryConstruct, field: Union[Field, Relationship] , rel_table: SQLTable, aggregator_name) -> ScalarSelect:
-        
-        logger.info('aggregating field %s on %s using %s', field, rel_table, aggregator_name)
+    def aggregate(self, query: QueryConstruct,
+                  field: Union[Field, Relationship], rel_table: SQLTable,
```
maxxxxxdlp, line 188: weird formatting here. can you make each attribute start on a new line, just like prettier would do that in javascript?

maxxxxxdlp (approved):
Not too knowledgeable on python code. I will defer to Jason for a deeper review

specifyweb/stored_queries/format.py
realVinayak: Just some weird things Python allows you to shoot your foot with

specifyweb/stored_queries/format.py
melton-jason: I'm not entirely sure, but I believe @realVinayak formatted the document with PEP 8 (the Python standard for formatting and style). 
If that is the case, then this change was done 'automatically'. 

@realVinayak which formatter did you use to format the document with? 
Or was it all done by hand? :grin:

realVinayak: Lol it was done by PEP 8. I can revert the formatting if that's a problem.

chanulee1:
How do I do UX testing for this?

specifyweb/stored_queries/format.py
maxxxxxdlp: Not a problem, just very strange - i.e, why does it have 2 arguments on first line and then 2 arguments on the next line rather than one argument per line (like in js) - that seems far more readable

Probably I'm just not used to the intricacies of the python world

specifyweb/stored_queries/format.py
```diff
@@ -113,20 +113,34 @@ def objformat(self, query: QueryConstruct, orm_table: SQLTable,
...
 
+        cycle_with_self = [*cycle_detector, (inspect(orm_table).class_.__name__, 'formatting')]
+
         def make_expr(query: QueryConstruct, fieldNodeText, fieldNodeAttrib) -> Tuple[
             QueryConstruct, blank_nulls, QueryFieldSpec]:
             path = fieldNodeText.split('.')
             path = [inspect(orm_table).class_.__name__, *path]
             formatter_field_spec = QueryFieldSpec.from_path(path)
             if formatter_field_spec.is_relationship():
+                next_table_name = formatter_field_spec.table.name
```
maxxxxxdlp, line 124: this looks like the same algorithm front-end uses for cycle detection 👍:

https://github.com/specify/specify7/blob/a4a66e63a8ec27c82c60bca0c01bc0e1410b0725/specifyweb/frontend/js_src/lib/components/Formatters/formatters.ts#L147-L154C9

maxxxxxdlp (approved):

pashiav:
Instructions for testers? @realVinayak

specifyweb/stored_queries/format.py
realVinayak: Yeah, except that you can detect cycles more definitely on the frontend rather than the backend. This is because frontend can check if the id is repeated. Backend cannot since the query hasn't been run yet.

pashiav:
HOW DO WE TEST THIS PLEASE? </3 @realVinayak

realVinayak:
**To-many not being aggregated when referenced in a formatter**
1. Open record formatter visual editor for a table.
2. Add a field which is to-many (for example, Determinations for Collection Object). It would appear as (aggregated)
3. If using Determinations for Collection Object, make a new collection object with more than 1 Determination.
4. Query that collection object, using (formatted) as the field. You should see that both of the determinations have been included in a single line.
Example:
image
My collection object has two determinations, one with Taxon named "Abantennarius will be added" and other with name "Aborichthys aa" and they both appear (since formatter of my determination is taxon)
image

bronwyncombs (changes requested):
aggregator limits aren't enforced </3
Screenshot 2024-01-03 at 3 52 06 PM
Screenshot 2024-01-03 at 3 51 45 PM

CarolineDenis (approved):

bronwyncombs (changes requested):
Probably shouldn't allow for users to set a formatter with an aggregated field, seems to make the software unhappy. 
Screenshot 2024-01-10 at 2 16 57 PM

Screenshot 2024-01-10 at 2 18 03 PM

bronwyncombs:
Error not replicated on `xml-editor` branch

Screenshot 2024-01-10 at 2 55 20 PM
Screenshot 2024-01-10 at 2 53 49 PM

bronwyncombs:
Recursive formatters still not handled by backend ([issue #4292](https://github.com/specify/specify7/issues/4292))

Screenshot 2024-01-10 at 3 23 14 PM
Screenshot 2024-01-10 at 3 22 55 PM

---

Deletion blockers give this error, not totally sure how this is tied to this PR but issue #3576 is linked so I'll comment on it

image

---

I applied the example given by Max in the description of issue #3266 with accession and accession agent. 

1. added accession agents (aggregated) to accession formatter
2. opened accession agent formatter
3. I received this error message as a toast

Screenshot 2024-01-10 at 3 34 04 PM

---

I also used the example given in the description of issue #3341 

This seems to be fixed as I only got one result from my query, but the videos and links wouldn't work on that issue so I am not positive

Screenshot 2024-01-10 at 3 50 11 PM
Screenshot 2024-01-10 at 3 52 31 PM

emenslin (changes requested):
Still running into this issue that links to #3266 that Bronwyn mentioned above. 
Screenshot 2024-01-25 142245

realVinayak:
@acwhite211 @bronwyncombs 

Hierarchy Exception is an unrelated [issue](https://github.com/specify/specify7/issues/3564).

emenslin (approved):
Looks good, it seems like everything that was supposed to be fixed has been fixed!

bronwyncombs (approved):
All issues raised that are within the scope of this PR have been resolved. Thanks for working on and improving this!

specifyweb/stored_queries/query_construct.py
```diff
@@ -15,10 +15,12 @@ def get_treedef(collection, tree_name):
             if tree_name == 'Storage' else
             getattr(collection.discipline, tree_name.lower() + "treedef"))
 
-class QueryConstruct(namedtuple('QueryConstruct', 'collection objectformatter query join_cache param_count')):
+class QueryConstruct(namedtuple('QueryConstruct', 'collection objectformatter query join_cache param_count tree_rank_count')):
+
     def __new__(cls, *args, **kwargs):
         kwargs['join_cache'] = dict()
         kwargs['param_count'] = 0
+        kwargs['tree_rank_count'] = 0
```
melton-jason, line 23: Outside of these two usages, where is `QueryConstruct.tree_rank_count` used? I can not find another use in the code base for the attribute.

melton-jason (approved):
At some point, we will definitely need to migrate away from the old `collections.namedtuple` class and at the very least update these classes (and hopefully others throughout the backend) to be subclasses of `typing.NamedTuple`. 

There was a lot to go through here, but the changes and added tests look okay. 

Further tests could be created testing ObjectFormatter methods `aggregate` and `objformat` when supplied with a specific aggregator/formatter name (besides None), but I would not consider that a necessity before this PR is merged.

specifyweb/stored_queries/query_construct.py
realVinayak: Yeah, @acwhite211 you can rid of this. I was working on handling cases where formatter of taxon is defined with fields from the parent. In that case, the cycle will end (unlike other cyclical cases). Alec you can continue on that if you want!

acwhite211 (approved):

specifyweb/stored_queries/query_construct.py
acwhite211: Sounds good, I'll keep it in, I added a TODO note in the code on implementing this in the future.

PR #3551, opened 2023-06-01
Disable color picker in AppResources for readOnly users 

### LINKED ISSUE #3519: Disable Colour Picker When User Does Not Have Permission To Change App Resources
**Describe the bug**
Title. Everything else is disabled. I.E TextFields, Drop Downs and Save Button does not work. But the colour picker still shows up. It doesn't really affect anything(you can not save) but it would be better if it doesn't show up.

**To Reproduce**
1. Go to `App Resource: UserPreferences` as a user without permissions 
2. Click on any colour
3. See error

**Expected behavior**
Not interactive like everything else

**Screen Recording**

https://github.com/specify/specify7/assets/53784701/beb84390-d7bb-4746-9ad1-b61181646dc8

**Desktop:**
 - OS: MacOSx Ventura
 - Browser: Chrome
 - Specify 7 Version: 7.8.11

**Database Name:** Gree -> Logged in as `Carlos_Wilson`

CarolineDenis:
Fixes #3519

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Preferences/Renderers.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Preferences/Renderers.tsx
@@ -59,6 +59,7 @@ export function ColorPickerPreferenceItem({
         type="color"
         value={value}
         onValueChange={handleChange}
+        disabled={isReadOnly}
       />
     </div>
   );


maxxxxxdlp (approved):

melton-jason (approved):
I enabled Read-Only Mode and the color pickers are disabled in both App Resources and User Preferences.

Issue #3544, opened 2023-05-30
Dark/Light Mode Transitions Now Require Refresh


carlosmbe:
**Describe the bug**
Title

**To Reproduce**
1. Go to Edge
2. Change between Dark/Light mode

**Expected behavior**
Smooth transitions not the weird lingering effect that needs a browser refresh
This is the expected behaviour 

https://github.com/specify/specify7/assets/53784701/b805e74e-e706-4f51-95aa-c6f4161b2129

**Screen Recording**
This is the weird behaviour 

https://github.com/specify/specify7/assets/53784701/5cb0a44d-52ff-4cf0-a36f-73170122717b

https://github.com/specify/specify7/assets/53784701/acb32dc9-c981-4f74-9581-785302aedb5e

**Desktop:**
 - OS: Mac OSx Ventura
 - Browser: Chrome
 - Specify 7 Version: Edge

PR #3553, opened 2023-06-01
Add pref to have records in readOnly mode using existing context + edit btn

### LINKED ISSUE #2451: Add pref to automatically have records in read-only mode, then click an Edit button

CarolineDenis:
Fixes #2451

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
@@ -263,8 +263,7 @@ export function ResourceView<SCHEMA extends AnySchema>({
           <DataEntry.Footer>
             {deleteButton}
             {extraButtons ?? <span className="-ml-2 md:flex-1" />}
-            {saveButtonElement}
-            {editRecord}
+            {recordInReadOnly ? editRecord : saveButtonElement}
           </DataEntry.Footer>
         ) : undefined}
       </>
--- a/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
@@ -186,7 +186,7 @@ export function ResourceView<SCHEMA extends AnySchema>({
 
   const [recordInReadOnlyPref] = userPreferences.use(
     'form',
-    'recordSet',
+    'behavior',
     'isReadOnly'
   );
 
--- a/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
@@ -26,6 +26,7 @@ import { useResourceView } from './BaseResourceView';
 import { DeleteButton } from './DeleteButton';
 import { SaveButton } from './Save';
 import { propsToFormMode } from './useViewDefinition';
+import { useTriggerState } from '../../hooks/useTriggerState';
 
 /**
  * There is special behavior required when creating one of these resources,
@@ -107,8 +108,8 @@ export function ResourceView<SCHEMA extends AnySchema>({
    */
   isSubForm,
   isDependent,
-  isRecordInReadOnly,
-}: {
+}: // isRecordInReadOnly,
+{
   readonly isLoading?: boolean;
   readonly resource: SpecifyResource<SCHEMA> | undefined;
   readonly viewName?: string;
@@ -129,7 +130,7 @@ export function ResourceView<SCHEMA extends AnySchema>({
   readonly title?:
     | LocalizedString
     | ((formatted: LocalizedString) => LocalizedString);
-  readonly isRecordInReadOnly?: boolean;
+  // readonly isRecordInReadOnly?: boolean;
 }): JSX.Element {
   const [isDeleted, setDeleted, setNotDeleted] = useBooleanState();
   // Remove isDeleted status when resource changes
@@ -172,19 +173,29 @@ export function ResourceView<SCHEMA extends AnySchema>({
     viewName,
   });
 
-  const [recordInReadOnlyPref] = userPreferences.use(
+  const [openAsReadOnly] = userPreferences.use(
     'form',
     'behavior',
-    'isReadOnly'
+    'openAsReadOnly'
   );
 
-  const [recordInReadOnly, setRecordInReadOnly] =
-    React.useState(recordInReadOnlyPref);
+  const hasOwnButton =
+    !isDependent &&
+    !isSubForm &&
+    typeof resource === 'object' &&
+    formElement !== null;
 
-  React.useEffect(() => {
-    if (isRecordInReadOnly !== undefined)
-      setRecordInReadOnly(isRecordInReadOnly);
-  }, [isRecordInReadOnly]);
+  const [temporaryReadOnly, setTemporaryReadOnly] = useTriggerState(
+    !isReadOnly && openAsReadOnly && hasOwnButton && !resource.isNew()
+  );
+
+  // const [recordInReadOnly, setRecordInReadOnly] =
+  //   React.useState(recordInReadOnlyPref);
+
+  // React.useEffect(() => {
+  //   if (isRecordInReadOnly !== undefined)
+  //     setRecordInReadOnly(isRecordInReadOnly);
+  // }, [isRecordInReadOnly]);
 
   const navigate = useNavigate();
   if (isDeleted)
@@ -198,17 +209,21 @@ export function ResourceView<SCHEMA extends AnySchema>({
       </Dialog>
     );
 
-  const editRecord = recordInReadOnly ? (
-    <Button.Secondary onClick={() => setRecordInReadOnly(false)}>
+  // const editRecord = recordInReadOnly ? (
+  //   <Button.Secondary onClick={() => setRecordInReadOnly(false)}>
+  //     {commonText.edit()}
+  //   </Button.Secondary>
+  // ) : undefined;
+  const editRecord = (
+    <Button.Secondary onClick={(): void => setTemporaryReadOnly(false)}>
       {commonText.edit()}
     </Button.Secondary>
-  ) : undefined;
+  );
 
-  const saveButtonElement =
-    !isDependent &&
-    !isSubForm &&
-    typeof resource === 'object' &&
-    formElement !== null ? (
+  const saveButtonElement = hasOwnButton ? (
+    temporaryReadOnly ? (
+      editRecord
+    ) : (
       <SaveButton
         form={formElement}
         resource={resource}
@@ -225,7 +240,8 @@ export function ResourceView<SCHEMA extends AnySchema>({
         }}
         onSaving={handleSaving}
       />
-    ) : undefined;
+    )
+  ) : undefined;
 
   const deleteButton =
     !isDependent &&
@@ -253,17 +269,25 @@ export function ResourceView<SCHEMA extends AnySchema>({
       ? titleOverride(formatted)
       : titleOverride;
 
+  const formComponent = (
+    <ReadOnlyContext.Provider value={isReadOnly || temporaryReadOnly}>
+      {form(children, dialog === false ? 'overflow-y-auto' : undefined)}
+    </ReadOnlyContext.Provider>
+  );
+
   if (dialog === false) {
     const formattedChildren = (
       <>
-        {form(children, 'overflow-y-auto')}
+        {/* {form(children, 'overflow-y-auto')} */}
+        {formComponent}
         {typeof deleteButton === 'object' ||
         typeof saveButtonElement === 'object' ||
         typeof extraButtons === 'object' ? (
           <DataEntry.Footer>
             {deleteButton}
             {extraButtons ?? <span className="-ml-2 md:flex-1" />}
-            {recordInReadOnly ? editRecord : saveButtonElement}
+            {/* {recordInReadOnly ? editRecord : saveButtonElement} */}
+            {saveButtonElement}
           </DataEntry.Footer>
         ) : undefined}
       </>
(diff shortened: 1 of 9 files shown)


specifyweb/frontend/js_src/lib/components/Forms/SpecifyForm.tsx
```diff
@@ -136,7 +136,9 @@ export function SpecifyForm<SCHEMA extends AnySchema>({
         {formIsLoaded ? (
           <DataEntry.Grid
             aria-hidden={showLoading}
-            className={`${showLoading ? 'pointer-events-none opacity-50' : ''}`}
+            className={`w-full ${
+              showLoading ? 'pointer-events-none opacity-50' : ''
```
maxxxxxdlp, line 140: please format nicely 🙂

```ts
            className={`
               w-full
               ${showLoading ? 'pointer-events-none opacity-50' : ''}
            `}
```


isn't this easier to read and edit?

maxxxxxdlp: ah, this change might have been made automatically by prettier, in that case sorry about that

specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
```diff
@@ -1079,6 +1086,13 @@ export const userPreferenceDefinitions = {
               },
             ],
           }),
+          isReadOnly: definePref<boolean>({
```
maxxxxxdlp, line 1089: The read-only feature is enabled not just for record sets, right?
So why is it in the record sets section in preferences

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
```diff
@@ -182,6 +184,26 @@ export function ResourceView<SCHEMA extends AnySchema>({
...
+  React.useEffect(() => {
+    if (isRecordInReadOnly !== undefined)
+      setRecordInReadOnly(isRecordInReadOnly);
+  }, [isRecordInReadOnly]);
+
+  const editRecord = recordInReadOnly ? (
+    <Button.Secondary onClick={() => setRecordInReadOnly(false)}>
+      {commonText.edit()}
+    </Button.Secondary>
+  ) : undefined;
```
maxxxxxdlp, line 205: You shouldn't need to add any new props at all. `isRecordInReadOnly` shouldn't be needed at all

Just consider this code:

```tsx
const [temporaryReadOnly, _,_, handleToggleReadOnly] = useBooleanState(recordInReadOnly);
const button = temporaryReadOnly && !isReadOnly
   ? <Button.Secondary onClick={handleToggleReadOnly}>{commonText.edit()}</Button.Secondary>
   : saveButtonElement;
```

And then:

```tsx
<ReadOnlyContext.Provider value={temporaryReadOnly || isReadOnly}>
  {/* the form and rest of the stuff goes here */}
</ReadOnlyContext.Provider>
```

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
CarolineDenis: The issue with this is that it doesn't handle all the cases. By example when we do a new data entry it uses ResourceView as well now. In the attachment gallery I had to add some logic too. 
I tested this branch and it seems to work everywhere and to handle all the cases. 
I'm happy to discuss this more over a quick call if you have a few minutes at some point

CarolineDenis: I'll try that

maxxxxxdlp: Do we want to have a single edit button that controls everything or separate edit buttons for the main form and the sub view dialogs/attachment dialogs?
If we only need a single edit button, then there is no need for a prop. If you want to be able to independently enable read-only mode for the main form, but not the sub form, then you likely need a prop.

CarolineDenis: I was thinking having it independent. I like to be able to change the readonly default behavior with this prop

grantfitzsimmons: We really only need a single edit button on the main form (just like in Specify 6). It doesn't make sense to repeat it, and users would understand that they need to close the dialog to enable edit more.

grantfitzsimmons (changes requested):
I really don't think that this should be enabled by default– it would add an unnecessary step to record editing. I can imagine adminstrators wanting to set this in default collection preferences for new users but have it off for themselves.

image

I also think that the text should read:

"Make records read-only by default"

and it should apply to all records, not just record sets (like in Specify 6).

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
maxxxxxdlp: Sounds good, because a single edit button will simplify the code. @CarolineDenis is there a problem with just having an edit button in place of the main save button?

CarolineDenis: There is only a single button
Screenshot 2023-06-06 at 7 46 04 AM
Screenshot 2023-06-06 at 7 46 41 AM

maxxxxxdlp: yes, but if you open a subview in a dialog, or an attachment viewer, you have another edit button
if I read your code correctly, you can set subview to read only or not independently of the main form, which might be not a useful feature and just add complexity for the user and in the code

CarolineDenis: In a subview or attachment viewer I did it like this: 
Screenshot 2023-06-06 at 8 37 38 AM

maxxxxxdlp: I was thinking we could simpliy the code by removing the edit button from the attachments viewer and just have it listen to ReadOnlyContext
Then, whether you are in read only or not is controlled only by a single "Edit" button on the **main** specify form (which would be behind the dialog)

CarolineDenis: But that wouldn't work in the attachment gallery as there is no main form behind it

CarolineDenis: Or we say that the attachment form in the plugin opened from the gallery is not readonly from start

maxxxxxdlp: > But that wouldn't work in the attachment gallery as there is no main form behind it

Attachment gallery is used in two places if I remember correctly:
1. The form, in which case the ResourceView from the main form will call ReadOnlyContext with value true, so the attachment gallery will be read only
2. The attachment viewer page, which is less commonly used than the main form so if it does or doesn't have an edit button is not as much of a deal. Still, the ResoureView is called by Attachment gallery, so it might end up having an edit button anyway.

> Or we say that the attachment form in the plugin opened from the gallery is not readonly from start

Sorry, don't quite understand it.
When user opens let's say Collection Object, it is read only, and everything inside of it, including Attachment Viewer is read only.
When user clicks on "Edit" on the Collection Object form, everything inside of it becomes non-read only, including the ResourceView that is called inside of Attachment Viewer
No special behaviour is needed for attachment viewer at all, or am I missing something?

This PR should probably only need to modify the ResourceView file, with just a few line changes, and nothing else

CarolineDenis: If we click on an attachment in the AttachmentGallery it opens attachment preview with attachment form. If the user choose to have his recordSet readOnly by default the attachment form in the dialog is not editable. The only way to edit it is to go in pref than choose not read only by default than go back to attachments. Which is why I added a way to also edit this form in Viewer.tsx
Screenshot 2023-06-09 at 9 02 14 AM

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
maxxxxxdlp, line 1: There is a bug in this component (that is caught by eslint). You are calling hooks conditionally. Inspect the code and try to see why and how

<p>

See all the places were `return` is called in this component

</p>
</details> 

<p>

you have an `if` and `return` before these hooks, which means function could either return, or run the hooks, thus hooks could get called conditionally, which is not allowed by react and is a bug. Move the if/return after the hooks to fix

</p>
</details>

specifyweb/frontend/js_src/lib/localization/preferences.ts
```diff
@@ -1816,4 +1816,7 @@ export const preferencesText = createDictionary({
     'ru-ru': 'Цвет кнопки успеха',
     'uk-ua': 'Колір кнопки успіху',
   },
+  recordSetReadOnly: {
```
maxxxxxdlp, line 1819: The key should be renamed. Should not contain "recordSet" anymore

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
maxxxxxdlp: See #3612
I believe in that implementation you can't make attachment viewer in the attachments tool read-only, but most importantly everything on the main form works, and the code is way simpler

CarolineDenis:
TODO: remove readonly when opening resource from edit button

CarolineDenis:
NOTES: codes have been commented out and will be removed once the behavior is approved and edit bug fixed

specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
```diff
@@ -64,20 +67,22 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
...
+            onSaved={(): void => {
+              resourceView.onSaved?.();
+              handleClose();
+            }}
+            {...resourceView}
+            isDependent={false}
+            isSubForm={false}
+            resource={resource}
+            onClose={handleClosed}
+            // isRecordInReadOnly={false}
```
maxxxxxdlp, line 83: don't forget to remove commented out code

specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
```diff
@@ -11,6 +11,9 @@ import { getResourceViewUrl } from '../DataModel/resource';
 import { deserializeResource } from '../DataModel/serializers';
 import { ResourceView } from '../Forms/ResourceView';
 
+export const IsNotReadOnly = React.createContext(false);
```
maxxxxxdlp, line 14: (optional) Add a one sentance description of what this context stores/what it should be used for.

Preferably use the `/**  Something */` syntax rather than `/* something */` or `// something`

That special syntax should be used when describing methods, classes, functions and global variables. [See more](https://tsdoc.org/)

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
```diff
@@ -170,6 +174,36 @@ export function ResourceView<SCHEMA extends AnySchema>({
...
+    'openAsReadOnly'
+  );
+
+  const hasOwnButton =
+    !isDependent &&
+    !isSubForm &&
+    typeof resource === 'object' &&
+    formElement !== null;
+
+  const isNotReadOnlyContext = React.useContext(IsNotReadOnly);
```
maxxxxxdlp, line 189: (optional) instead of creating a context (Which is a fine solution) you might be able to get away with just using `dialog === false` (i.e, only do readonly for forms opened not in a dialog)

maxxxxxdlp (approved):
Looks good. Left optional suggestions. Otherwise, just remove commented out code and send it for testing

CarolineDenis:
FOR TESTING: 
- go in pref > form > check "open all records in read-only mode" >> save 
- open a record set, it should be read-only, click on edit button (bottom right) to make edits. + verify that combo box and subforms becomes editable after clicking "edit" 
- go in attachment gallery > click on an attachment > even if the pref is read-only the form in the attachment preview should not be read-only 
- open a new form entry > it shouldn't be read only
- open tree > click on the pencil next to the title > the form shouldn't be read-only as it has been opened from an edit button 
- uncheck the pref read-only, save and verify record sets are NOT read-only

grantfitzsimmons (changes requested):
https://herbrbge11723-issue-2451-b.test.specifysystems.org/specify/user-preferences/

`spadmin`, Herbarium, it does not have any effect.

**UserPreferences:**
```json
{"general":{"ui":{"theme":"light"}},"statistics":{"appearance":{"layout":[{"label":"Personal","categories":[{"label":"Collection","items":[{"type":"CustomStat","label":"Collection Objects Cataloged","querySpec":{"tableName":"CollectionObject","fields":[{"tableList":"1","stringId":"1.collectionobject.catalogNumber","fieldName":"catalogNumber","isRelFld":false,"sortType":0,"position":0,"isDisplay":true,"operStart":8,"startValue":"","isNot":false,"path":"catalogNumber"}],"isDistinct":false}},{"type":"DefaultStat","pageName":"user","itemName":"collectionObjectsDetermined","categoryName":"holdings","label":"Collection Objects Determined","itemType":"QueryStat"},{"type":"DefaultStat","pageName":"user","itemName":"collectionObjectInventorized","categoryName":"holdings","label":"Collection Objects Inventorized","itemType":"QueryStat"}]}]}]}},"form":{"preferences":{"disableClone":["CollectionObject"],"enableCarryForward":["CollectionObject"]},"behavior":{"openAsReadOnly":true}},"welcomePage":{"general":{"mode":"taxonTiles"}},"header":{"appearance":{"items":{"visible":["logOut","dataEntry","trees","interactions","queries","recordSets","reports","workBench","attachments","statistics"],"hidden":[]}}}}
```

image

macOS, Chrome

CarolineDenis:
> https://herbrbge11723-issue-2451-b.test.specifysystems.org/specify/user-preferences/
> 
> `spadmin`, Herbarium, it does not have any effect.
> 
> **UserPreferences:**
> 
> ```json
> {"general":{"ui":{"theme":"light"}},"statistics":{"appearance":{"layout":[{"label":"Personal","categories":[{"label":"Collection","items":[{"type":"CustomStat","label":"Collection Objects Cataloged","querySpec":{"tableName":"CollectionObject","fields":[{"tableList":"1","stringId":"1.collectionobject.catalogNumber","fieldName":"catalogNumber","isRelFld":false,"sortType":0,"position":0,"isDisplay":true,"operStart":8,"startValue":"","isNot":false,"path":"catalogNumber"}],"isDistinct":false}},{"type":"DefaultStat","pageName":"user","itemName":"collectionObjectsDetermined","categoryName":"holdings","label":"Collection Objects Determined","itemType":"QueryStat"},{"type":"DefaultStat","pageName":"user","itemName":"collectionObjectInventorized","categoryName":"holdings","label":"Collection Objects Inventorized","itemType":"QueryStat"}]}]}]}},"form":{"preferences":{"disableClone":["CollectionObject"],"enableCarryForward":["CollectionObject"]},"behavior":{"openAsReadOnly":true}},"welcomePage":{"general":{"mode":"taxonTiles"}},"header":{"appearance":{"items":{"visible":["logOut","dataEntry","trees","interactions","queries","recordSets","reports","workBench","attachments","statistics"],"hidden":[]}}}}
> ```
> 
> image
> macOS, Chrome

just tested and works as expected on my end

realVinayak:
@grantfitzsimmons can you double check this problem occurs on your computer? This might be a deeper problem with preferences refactor.

@CarolineDenis can you look at the code that updates preferences (it is in `specifyweb/frontend/js_src/lib/components/Preferences/BasePreferences.tsx`) and make sure it detects updates correctly? Feel free to ask questions about that. While you are at it, also see why changes to dark mode isn't detected (if it is, figure out why it doesn't get propagated). Context here is that the fix here (https://github.com/specify/specify7/issues/3544) is not correct. You shouldn't have it to put dark mode in the effect dependency. 

To both, feel free to not do any of this if it is not a high priority.

grantfitzsimmons (approved):
Works fine. Maybe it was an anomoly? :dizzy_face: I cleared my cache and gave it another shot and it worked fine. Can another tester take a look before we merge?

PR #3582, opened 2023-06-06
Disable find usages button when resource is undefined

### LINKED ISSUE #3549: 404 Error With `Find Usages` Button In Data Entry for Tree
**Describe the bug**
If you enter Form Meta while doing Data Entry for something with a Tree. IE Geography or Taxon. You can click on `Find Usages` but since we haven't entered anything we get a 404

**To Reproduce**
1. Go to Data Entry -> Geography( or Taxon)
2. Click on Form Meta
3. Click on Find Usages

**Expected behavior**
I guess it should be disabled 

**Screenshots**
Screenshot 2023-05-31 at 4 34 05 PM
Screenshot 2023-05-31 at 4 34 39 PM
Screenshot 2023-05-31 at 4 35 03 PM
Screenshot 2023-05-31 at 4 35 10 PM

**Desktop:**
 - OS: Mac OSx Ventura
 - Browser: Chrome
 - Specify 7 Version: 7.8.11

CarolineDenis:
Fixes #3549

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormMeta/QueryTreeUsages.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormMeta/QueryTreeUsages.tsx
@@ -13,19 +13,16 @@ export function QueryTreeUsages({
 }): JSX.Element {
   return (
     <ProtectedAction action="execute" resource="/querybuilder/query">
-      <Link.Small
-        className={
-          resource.id === undefined
-            ? 'pointer-events-none !bg-white !ring-1 !ring-gray-400 dark:!bg-neutral-700 dark:!ring-neutral-500'
-            : ''
-        }
-        href={`/specify/query/fromtree/${resource.specifyModel.name.toLowerCase()}/${
-          resource.id
-        }/`}
-        target="_blank"
-      >
-        {formsText.findUsages()}
-      </Link.Small>
+      {resource.id === undefined ? null : (
+        <Link.Small
+          href={`/specify/query/fromtree/${resource.specifyModel.name.toLowerCase()}/${
+            resource.id
+          }/`}
+          target="_blank"
+        >
+          {formsText.findUsages()}
+        </Link.Small>
+      )}
     </ProtectedAction>
   );
 }
--- a/specifyweb/frontend/js_src/lib/components/FormMeta/QueryTreeUsages.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormMeta/QueryTreeUsages.tsx
@@ -18,6 +18,11 @@ export function QueryTreeUsages({
           resource.id
         }/`}
         target="_blank"
+        className={
+          resource.id === undefined
+            ? 'pointer-events-none !bg-gray-200 !text-gray-500 !ring-1 !ring-gray-400 dark:!bg-neutral-700 dark:!ring-neutral-500'
+            : ''
+        }
       >
         {formsText.findUsages()}
       </Link.Small>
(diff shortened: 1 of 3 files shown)


maxxxxxdlp:
No no neat way to do that

Remember the two solutions we discussed for cases like this?
- When you need to render a disabled link, instead render an identically styled button in its place - this is best when the element can change from being disabled to not being disabled on the fly
- Simply not render the link when it should be disabled (simpler, and works in this case as link would only become visible after use clicks on "Save", which usually happens when the Form Meta dialog is closed)

specifyweb/frontend/js_src/lib/components/FormMeta/QueryTreeUsages.tsx
```diff
@@ -13,19 +13,16 @@ export function QueryTreeUsages({
...
-            : ''
-        }
-        href={`/specify/query/fromtree/${resource.specifyModel.name.toLowerCase()}/${
-          resource.id
-        }/`}
-        target="_blank"
-      >
-        {formsText.findUsages()}
-      </Link.Small>
+      {resource.id === undefined ? null : (
```
maxxxxxdlp, line 16: There is no point in rendering ProtectedAction if button is not displayed - ProtectedAction can display a dialog to the user if they don't have a permission, which is not helpful as there is no button, so no need for permission check.

Could you please modify the code by moving the ternary operator outside the ProtectedAction?

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/FormMeta/QueryTreeUsages.tsx
maxxxxxdlp: optionally, if you want you could even move the input outside of this component

specifyweb/frontend/js_src/lib/components/FormMeta/index.tsx
```diff
@@ -161,7 +161,7 @@ function MetaDialog({
                 <EditHistory resource={resource} />
               </ProtectedAction>
             </ProtectedTool>
-            {isTreeResource(resource) && (
+            {isTreeResource(resource) && resource.id !== undefined && (
```
melton-jason, line 164: Could we change `resource.id !== undefined` to `!resource.isNew()`? 
This would stay consistent with the rest of the code base and I personally think it is more explicit. 

To a new developer, or even someone going through the code `resource.isNew()` leaves no doubt as to _why_ its there. It might not be immediately clear that we are hiding the QueryTreeUsages component when the resource is new, or they might not have realized that an undefined id means a new resource. 
Alternatively, `isNew()` is a function from Backbone. If/when we decide to get rid of Backbone, it would be much easier to detect and thus replace this line.

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/FormMeta/index.tsx
```diff
@@ -161,7 +161,7 @@ function MetaDialog({
                 <EditHistory resource={resource} />
               </ProtectedAction>
             </ProtectedTool>
-            {isTreeResource(resource) && (
+            {isTreeResource(resource) && !resource.isNew() && (
```
maxxxxxdlp, line 164: Please try to avoid using the `&&` hack when there is more than one condition and prefer explicit ternary in cases like this.
The `&&` hack is fine if there is just one condition, but otherwise it get's a bit too hard to comprehend

maxxxxxdlp (approved):

bronwyncombs (approved):
Find Usages button working as expected!

PR #3583, opened 2023-06-06
Take off aria-pressed from toggleMapping as the content of btn changes

### LINKED ISSUE #3547: `Show Field Mapper` Has No Background. Similar to a disabled button
**Describe the bug**
The Button still works. It just looks like it's disabled. Similar to how the `Save Query` Button looks like when in an already saved query

Screenshot 2023-05-31 at 1 34 59 PM

**To Reproduce**
1. Go to Query
2. Click on `Hide Field Mapper`
3. See error

**Screenshots**

https://github.com/specify/specify7/assets/53784701/e4e4c096-c51b-4596-989a-35a7fcee1053

**Desktop:**
 - OS: Mac OSx Ventura
 - Browser: Chrome
 - Specify 7 Version: 7.8.11 and Edge as of 31 May 2023

CarolineDenis:
Fixes #3547

### DIFF
--- a/specifyweb/frontend/js_src/css/main.css
+++ b/specifyweb/frontend/js_src/css/main.css
@@ -159,7 +159,7 @@
   }
 
   .button:is([aria-pressed='true'], [aria-selected='true'], [aria-current]):not(.aria-handled) {
-    @apply brightness-[1.2] dark:brightness-[1.5];
+    @apply ring-[color:var(--accent-color-300)] brightness-[1.2] dark:brightness-[1.5];
   }
 
   button[aria-pressed='true']:not(.aria-handled).link {
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Components.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Components.tsx
@@ -135,7 +135,6 @@ export function ToggleMappingViewButton({
 }): JSX.Element {
   return (
     <Button.Small
-      aria-pressed={!showMappingView}
       disabled={fields.length === 0 && showMappingView}
       onClick={handleClick}
     >


maxxxxxdlp (changes requested):
@CarolineDenis can you please investigate what I said in https://github.com/specify/specify7/issues/3547#issuecomment-1572508147 and see if that's actually the case?

Because if the styles are actually quite similar, than this issue would affect many places in specify, and so just removing the indicator in this one place won't fix the issue, just mask it - bad way to go about it

bronwyncombs (approved):
Looks good 👍 
Screen Shot 2023-06-07 at 8 10 58 AM
Screen Shot 2023-06-07 at 8 11 01 AM

CarolineDenis:
> @CarolineDenis can you please investigate what I said in [#3547 (comment)](https://github.com/specify/specify7/issues/3547#issuecomment-1572508147) and see if that's actually the case?
> 
> Because if the styles are actually quite similar, than this issue would affect many places in specify, and so just removing the indicator in this one place won't fix the issue, just mask it - bad way to go about it

aria-pressed only adds brightness

maxxxxxdlp:
Okay, so added brightness makes background look whiter, which makes the button look similar to disabled, right?
So, my point is, this problem would happen in every place where aria-pressed is used, not just here, correct?
So while removing aria-pressed here might be ok, you should also consider experimenting with changing the CSS styles for aria-pressed and testing how that affects the other places where it's used

CarolineDenis:
> Okay, so added brightness makes background look whiter, which makes the button look similar to disabled, right?
> So, my point is, this problem would happen in every place where aria-pressed is used, not just here, correct?
> So while removing aria-pressed here might be ok, you should also consider experimenting with changing the CSS styles for aria-pressed and testing how that affects the other places where it's used

I added some subtility when aria-pressed. Although didn't add it (aria-pressed) back to the button in this PR as the content of the button changes.

maxxxxxdlp (approved):
The code looks good. Will leave to UX testers to decide if the design looks good

Issue #2339, opened 2022-10-20
Customizing tree accent color does not work quite right


maxxxxxdlp:
As discovered during a webinar
Customizing the accent color sometimes leads to tree still using the old accent color.
Or sometimes customizing the accent color for one tree also applies to another tree.
Seems like the code that checks which tree to use the accent color for is a bit broken

CarolineDenis:
The underlying issue seems to cause this as well: https://github.com/specify/specify7/issues/3500

maxxxxxdlp:
I think this might have actually been caused by the same issue as https://github.com/specify/specify7/issues/3578.

Could someone check if https://github.com/specify/specify7/pull/3586 fixes this issue?

Issue #3500, opened 2023-05-15
Empty Trees Should Not Be Interactive 


carlosmbe:
**Describe the bug**
In the event of a tree being completely empty and having it's root gone. The Tree should not be interactive as there is no direct easy way of creating a new root and performing certain task  like adding a child become impossible as there is no Parent. However, you can still try and either it throws a 404 or forever loads until you refresh the page.

Sometimes the expected behaviour is met but not always

**To Reproduce**
1. Go to `invertpaleo`
2. Enter the `Taxon Tree`
3. Enter another tree that has items and then go back to the `Taxon Tree`
4. Do something that requires a Parent. I.E Add a child
5. See errors

**Screenshots**

<img width="1792" alt="Screenshot 2023-05-15 at 12 36 59 PM" src="https://github.com/specify/specify7/assets/53784701/f04cd8c2-101b-455d-9c1c-1aea03704394">

<img width="1788" alt="Screenshot 2023-05-15 at 12 37 42 PM" src="https://github.com/specify/specify7/assets/53784701/20dc6157-3016-4737-b74e-1fd14a07b77b">
<img width="1792" alt="Screenshot 2023-05-15 at 12 37 54 PM" src="https://github.com/specify/specify7/assets/53784701/6aac939c-c4db-4555-8889-be598da8a375">
<img width="1792" alt="Screenshot 2023-05-15 at 12 38 08 PM" src="https://github.com/specify/specify7/assets/53784701/e24cc44b-25d5-4caf-aff0-c2a75db661b0">

This load never ends
<img width="1792" alt="Screenshot 2023-05-15 at 12 38 15 PM" src="https://github.com/specify/specify7/assets/53784701/d5dd712d-d07d-465d-8cb1-b3747cd16310">

CarolineDenis:
Fixed in https://github.com/specify/specify7/pull/3586

PR #3586, opened 2023-06-06
Reset the focused row in Tree View when Tree Type changes

### LINKED ISSUE #3578: 404 In Trees If You Don't Interact With One After Messing With Another
**Describe the bug**
Let's say I was in Taxonomy or Geography and added some children or anything really. The top buttons become active and interactive. If you go to another tree, the buttons are still interactive despite not selecting anything in this new tree. Using them will produce 404 Errors.

**To Reproduce**
1. Go to Geography and interact with the tree
2. Go to another tree, like Storage
3. Click on any buttons on top

**Expected behavior**
Buttons should be deactivated 

**Screen Recording**

https://github.com/specify/specify7/assets/53784701/090f37f5-0a37-4e62-9190-073ae2c5ee83

**Desktop:**
 - OS: Mac OSx Ventura
 - Browser: Chrome
 - Specify 7 Version: 7.8.11 and 7.8.12

**Database Name:** Any

melton-jason:
Fixes #3578

To Test:

- Focus on a node in any Tree
- Switch to another Tree of a different table
- Make sure the Tree Action Buttons (Query, Edit, Delete, etc.) are disabled until a node in that tree is focused

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
@@ -142,6 +142,10 @@ function TreeView<SCHEMA extends AnyTree>({
   const [focusedRow, setFocusedRow] = React.useState<Row | undefined>(
     undefined
   );
+  React.useEffect(() => {
+    setFocusedRow(undefined);
+  }, [tableName]);
+
   const [actionRow, setActionRow] = React.useState<Row | undefined>(undefined);
 
   const searchBoxRef = React.useRef<HTMLInputElement | null>(null);
--- a/specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
@@ -390,6 +390,8 @@ export function TreeViewWrapper(): JSX.Element | null {
     <ProtectedTree action="read" treeName={treeName}>
       {typeof treeDefinition === 'object' ? (
         <TreeView
+          treeDefinition={treeDefinition.definition}
+          treeDefinitionItems={treeDefinition.ranks}
           tableName={treeName}
           /**
            * We're casting this as a generic Specify Resource because
@@ -400,8 +402,6 @@ export function TreeViewWrapper(): JSX.Element | null {
           key={(treeDefinition.definition as SpecifyResource<AnySchema>).get(
             'resource_uri'
           )}
-          treeDefinition={treeDefinition.definition}
-          treeDefinitionItems={treeDefinition.ranks}
         />
       ) : null}
     </ProtectedTree>
--- a/specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
@@ -17,6 +17,7 @@ import { Button } from '../Atoms/Button';
 import { DataEntry } from '../Atoms/DataEntry';
 import { deserializeResource } from '../DataModel/helpers';
 import type {
+  AnySchema,
   AnyTree,
   FilterTablesByEndsWith,
   SerializedResource,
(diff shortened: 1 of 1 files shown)


carlosmbe (approved):

maxxxxxdlp (changes requested):
While this would likely fix the issue in this case, a similar problem might occur again in the future if we add more states to this component.

A better solution in this case might be to replace this:
```tsx
        <TreeView
          tableName={treeName}
          treeDefinition={treeDefinition.definition}
          treeDefinitionItems={treeDefinition.ranks}
        />
```

with this:
```tsx
        <TreeView
          tableName={treeName}
          key={treeDefinition.definition.resource_uri}
          treeDefinition={treeDefinition.definition}
          treeDefinitionItems={treeDefinition.ranks}
        />
```

(at the bottom of this file)

This way, react is forced to re-create the entire tree view from scratch when user goes to a tree for a different table (or to a different tree from same table - we want a future proof solution that doesn't break when we allow users to have e.g more than one Taxon tree). And when react re-creates the entire tree view, all states are reset

> At least that's how I understand how react works. Can you please test if my proposal actually fixes the issue?

If this does fix it, a similar fix can be applied in other places were a similar issue occurs.
We don't want to apply this everywhere as this decreases performance, and can't be used in places where state needs to be at least partially preserved (i.e, going between collection pages in the security panel)

melton-jason:
> While this would likely fix the issue in this case, a similar problem might occur again in the future if we add more states to this component.
> 
> A better solution in this case might be to replace this:
> 
> ```tsx
>         <TreeView
>           tableName={treeName}
>           treeDefinition={treeDefinition.definition}
>           treeDefinitionItems={treeDefinition.ranks}
>         />
> ```
> 
> with this:
> 
> ```tsx
>         <TreeView
>           tableName={treeName}
>           key={treeDefinition.definition.resource_uri}
>           treeDefinition={treeDefinition.definition}
>           treeDefinitionItems={treeDefinition.ranks}
>         />
> ```
> 
> (at the bottom of this file)
> 
> This way, react is forced to re-create the entire tree view from scratch when user goes to a tree for a different table (or to a different tree from same table - we want a future proof solution that doesn't break when we allow users to have e.g more than one Taxon tree). And when react re-creates the entire tree view, all states are reset
> 
> > At least that's how I understand how react works. Can you please test if my proposal actually fixes the issue?
> 
> If this does fix it, a similar fix can be applied in other places were a similar issue occurs. We don't want to apply this everywhere as this decreases performance, and can't be used in places where state needs to be at least partially preserved (i.e, going between collection pages in the security panel)

I didn't know React had that behavior for a `key` prop, thank you. I will definitely have to read up on this!

maxxxxxdlp (approved):

CarolineDenis:
@carlosmbe can you confirm that these changes also fix: 
https://github.com/specify/specify7/issues/2339
https://github.com/specify/specify7/issues/3500 
If yes, we can close those 2 issues when this one is merged

melton-jason:
I need to look into the failing test caused           `key={treeDefinition.definition.get('resource_uri')}`

I've briefly looked into the issue and looks like TypeScript doesn't like calling get() on the union of tree types. 

This is most commonly solved by defining a new, more explicit type. I need to see if that solution is viable in this case, or think of another one.

carlosmbe:
Can confirm that this PR fixes all 3 Issues
https://github.com/specify/specify7/issues/2339
https://github.com/specify/specify7/issues/3500
https://github.com/specify/specify7/issues/3578

maxxxxxdlp:
> key={treeDefinition.definition.get('resource_uri')}

Yeah, that bothers me to, and I didn't find a good solution yet. Any time a record is a union of several possible tables, typescript doesn't like it (unless it's explicitly "any table"). I was hoping to look into fixing that properly when the rewrite to new ORM happens, but run out of time. As a short term fix, a type assertion  might be best.

Note, the type assertion should say that treeDefinition.definition is a SpecifyResource<TaxonTreeDef> rather than treeDefinition.definition.get('resource_uri') is a string - the first one is a narrower type assertion (just telling typescript which of 6 possible types is the one we are looking at) - the 2nd one is when typescript completely failed to infter, and so didn't do type safe checks, and we tell it to "scrap it, use this type instead"

maxxxxxdlp (approved):
Looks good.

You might think typescript is smart enough to realize that every resource has resource_uri property, so it doesn't matter which resource we have, but unfortunately it doesn't do it - possibly because it could be really expensive - i.e, imagine instead of there being 7 tree tables, we have some union that is between a thousand items - typescript would have to check if every one of them has resource_uri property - of course they could add a limit where they would only check if there are less than x items in the union, but that means that if you ever add a new item to the union (i.e, added more tables to the datamodel), the code will suddenly begin to throw type errors and you will have to rewrite it in a lot of places

An even smarter way would probably be to detect that the presence of resource_uri and it's type does not depend on the `SCHEMA` generic, and so is present regardless of the generic, but I suppose that would add more complexity to an already complex typescript type checker engine

PR #3587, opened 2023-06-07
Don't trigger unload protect when setting lat/long

### LINKED ISSUE #3584: Locality not changed but being asked if I want to save changes
**Describe the bug**
Title

**To Reproduce**
1. Go to Data Entry, create and save a Locality 
2. Create another new item without leaving data entry and save it
3. Using the paginator navigation arrows. Travel between the locality and new item

**Expected behavior**
I did not make any changes. So I should not get this prompt

**Screen Recording**

https://github.com/specify/specify7/assets/53784701/e3e7f4c3-41f6-460b-80af-d606b70cb931

**Desktop:**
 - OS: Mac OSx Ventura
 - Browser: Chrome
 - Specify 7 Version: 7.8.11 and 7.8.12

melton-jason:
Fixes #3584

To Test:
- Create and save a new Locality
  - Refresh the page; the Save button should be disabled
  - Try inputting values in the Latitude and Longitude coordinate boxes and ensure there is no abnormalities in Form behavior
- Query on Locality and find one (or more) that has very precise Latitude 1 and Longitude 1 values
  - Ensure the Save button is disabled when opening the form
  - Repeat the steps under the (Create and save a new Locality) steps

@maxxxxxdlp As shown in the code comments, while not related to #3584, I stumbled across another issue. When rendering the Coordinate components, Specify parses the value (initially a string) to a Coord and then converts the Coord to a float and tries to set the once-string value to the new float. 
If the value was precise, there may be a loss of precision during the conversion. While most institutions would likely never _need_ values that precise, data integrity is still data integrity. 
I was looking at possible solutions, and am considering Number().toFixed(), because we know the database precision for the field. I'll need to look into this more if it is that important of a problem.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
@@ -98,8 +98,15 @@ function Coordinate({
     );
 
     isChanging.current = true;
-    resource.set(coordinateField, parsed?.asFloat() ?? null);
-    resource.set(coordinateTextField, trimmedValue || null);
+
+    /**
+     * Do not set unload protect because very precise coodinateFields
+     * may experience a change of precision during the conversion from
+     * string to float
+     */
+    resource.set(coordinateField, parsed?.asFloat() ?? null, { silent: true });
+
+    resource.set(coordinateTextField, trimmedValue);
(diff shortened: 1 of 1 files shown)


bronwyncombs (approved):

maxxxxxdlp (approved):
Feels like 50% of forms bug are in LatLongUi and PartialDateUi
It doesn't help that I didn't write any tests for those 😢

PR #3588, opened 2023-06-07
Add the possibility to see more than 1 attachment for a record 

### LINKED ISSUE #1048: Implement "icon" view type / Add ability to see more than one attachment thumbnail at a time
When looking at attachments in 7, users can only see one attachment thumbnail at a time.
image

The same record in 6 shows multiple attachment thumbnails at a time
image

CarolineDenis:
Fixes #1048

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
@@ -1,16 +1,17 @@
 import React from 'react';
-import { AnySchema } from '../DataModel/helperTypes';
-import { Collection } from '../DataModel/specifyModel';
-import { SpecifyResource } from '../DataModel/legacyTypes';
-import { RA } from '../../utils/types';
+
+import { useBooleanState } from '../../hooks/useBooleanState';
+import { useTriggerState } from '../../hooks/useTriggerState';
+import { attachmentsText } from '../../localization/attachments';
+import { commonText } from '../../localization/common';
+import type { RA } from '../../utils/types';
+import { Button } from '../Atoms/Button';
 import { useAttachment } from '../Attachments/Plugin';
 import { AttachmentViewer } from '../Attachments/Viewer';
-import { useTriggerState } from '../../hooks/useTriggerState';
+import type { AnySchema } from '../DataModel/helperTypes';
+import type { SpecifyResource } from '../DataModel/legacyTypes';
+import type { Collection } from '../DataModel/specifyModel';
 import { Dialog } from '../Molecules/Dialog';
-import { Button } from '../Atoms/Button';
-import { commonText } from '../../localization/common';
-import { attachmentsText } from '../../localization/attachments';
-import { useBooleanState } from '../../hooks/useBooleanState';
 
 export function AttachmentsCollection<SCHEMA extends AnySchema>({
   collection,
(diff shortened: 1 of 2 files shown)


CarolineDenis:
Here is the preview
Screenshot 2023-06-07 at 7 05 43 AM


Screenshot 2023-06-07 at 7 06 08 AM

specifyweb/frontend/js_src/lib/components/Attachments/Viewer.tsx
```diff
@@ -29,13 +29,15 @@ export function AttachmentViewer({
   related: [related, setRelated],
   showMeta = true,
   onViewRecord: handleViewRecord,
+  classNameProp,
```
maxxxxxdlp, line 32: just call it `className`? the fact that it's a prop is clear as it can't be anything other than a prop so is redundant

The only case you would use prop in a name is if you do something like `inputProps` (i.e, in AutoComplete) to indicate that all those props would actually be passed to a different component

specifyweb/frontend/js_src/lib/components/FormSliders/AllAttachments.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { useAttachment } from '../Attachments/Plugin';
+import { AttachmentViewer } from '../Attachments/Viewer';
+import { useTriggerState } from '../../hooks/useTriggerState';
+import { Dialog } from '../Molecules/Dialog';
+import { Button } from '../Atoms/Button';
+import { commonText } from '../../localization/common';
+import { attachmentsText } from '../../localization/attachments';
+import { useBooleanState } from '../../hooks/useBooleanState';
+
+export function AttachmentsCollection<SCHEMA extends AnySchema>({
```
maxxxxxdlp, line 15: Can you please remind me why attachment gallery wasn't used for this?
Seems like it would be provide the same functionality - in fact, didn't you already use it on the branch where you added ability to see attachments belonging to a record set?

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/FormSliders/AllAttachments.tsx
CarolineDenis: Because I preferred creating a new component with only what I needed. For the attachments in RecordSet I also created a new component.

maxxxxxdlp: > Because I preferred creating a new component with only what I needed. For the attachments in RecordSet I also created a new component.

😢 

This way you might be duplicating some code (thus 3 places to maintain and 3 places that could have a bug). Also, if there is a bug in one of the places, fixing it won't fix the other 3. That means if doing testing, all 3 places have to be tested - if the same component was reused, testing just one would most likely be sufficient.

Can you please look at using `<AttachmentGallery />` here and in record sets? Looking at the props it accepts, you should just be able to use it without any extra modifications, or did you run into some issues? We can also meet and try this together if you want

CarolineDenis: I'd be happy to read you suggested code implementing attachmentGallery. How would get RA<SerializedResource<Attachment>> to pass as attachments prop to Gallery in IntegratedRecordSelector.tsx?

maxxxxxdlp: Your new compoment accepts a Collection<...>, right? If you have a collection you can get RA<...> just like that:

```
collection.models
```

melton-jason:
@CarolineDenis can you provide a brief explanation as to why this pull request was closed? And provide links to any relevant Issues or Pull Requests?  
It may help you or other developers in the future!

Issue #3518, opened 2023-05-19
0 Results Should Disable Create KML 


carlosmbe:
**Describe the bug**
Title. 
When the blank KML is imported to Google Earth you just get a blank project. Which I assume, would be confusing as users will wonder whether it's a Specify issue or a Google Earth issue

**To Reproduce**
1. Run a Longitude and Latitude Query which has 0 Results
2. Click on `Create KML`
3. Download the empty KML and import it in Google Earth (or see screenshots)

**Expected behavior**
It shouldn't let you export an empty KML. 
**Screenshots**

<img width="1792" alt="Screenshot 2023-05-19 at 3 16 14 PM" src="https://github.com/specify/specify7/assets/53784701/55693586-a256-40b6-97d0-7a595478cc0f">
<img width="1791" alt="Screenshot 2023-05-19 at 3 16 26 PM" src="https://github.com/specify/specify7/assets/53784701/4024d9f0-a3a1-4915-9e74-7962cf38526e">
<img width="1792" alt="Screenshot 2023-05-19 at 3 16 43 PM" src="https://github.com/specify/specify7/assets/53784701/86a98c70-a89f-462b-a2c8-a5f4f0e20495">
<img width="1792" alt="Screenshot 2023-05-19 at 3 18 09 PM" src="https://github.com/specify/specify7/assets/53784701/acfbefdf-32fe-4891-a798-d701d0bdbe96">

grantfitzsimmons:
Should be fixed by #3590 once it is merged

Issue #3589, opened 2023-06-07
Query Builder export to KML ignores selected rows


melton-jason:
Similar to #2348, but with the `Export to KML` option.

Now, #2348 has been fixed in #3476, but there is still no way for the Query Builder to export only selected rows to KML. 

(As a reminder, the Export to KML button is only visible when there is coordinate data, like Latitude1 and Longitude1, in the Query)

emenslin:
Can recreate in edge (7.9.6)

Issue #3525, opened 2023-05-22
Check all file export naming


CarolineDenis:
In python, to create a file, you would do something like `with open(` - do global search on that.

Other ex: filename = 'query_results_%s.csv' % datetime.now().isoformat()

(Try to have file name in export name)

maxxxxxdlp:
Unless you plan on fixing this issue right away, you need to include more context information - "Check all file export naming" doesn't tell what to check for and why we have to check file export naming?

Unless the issue is going to be fixed very soon, you should assume it might not be fixed in 5 years - you won't remember what it was about 5 years later, and someone else looking at this would certainly not know.

Two solutions:
- Include more context information
- OR even simpler: just post a link [to the original comment that included more information](https://github.com/specify/specify7/pull/3476#discussion_r1200827112) (something like `More context - <link here>`)

melton-jason:
I apologize asking, but why was this issue closed? 
We _do_ have an open pull request (#3590) to partially address the issue here, but it is not comprehensive. 

Are the Query export names the _only_ file names exported from the backend?  
If so, then this issue can probably remain closed.

PR #3590, opened 2023-06-07
Improve backend query export names

melton-jason:
> Yes, we should improve the back-end file name. But, Caroline created a separate issue for that, so it doesn't necessarily have to be fixed as part of this PR - https://github.com/specify/specify7/issues/3525

Addresses https://github.com/specify/specify7/pull/3476#issuecomment-1560024885 and part of #3525 

Follows the more user-friendly Query Export file name introduced in #3476. 
All query export files should now be in the following format: `QueryName - Weekday (abbreviated) Month (abbreviated) Day Year`
Example: `New Query - Wed Jun 07 2023`

To Test:
Without selecting any rows:
- Export a Query to csv
  - Ensure the file name has the correct format (which should be similar to the file name when selected rows are exported)
- Export a Query to kml 
  - Ensure the file name has the correct format
  - The `Export to KML` button is only visible when the Query contains Coordinate data (like Longitude1 and Latitude1)
  - No testing needs to be done regarding exporting selected rows to kml: this is not currently supported (#3589)


carlosmbe (changes requested):
Two Things:
Exported Files are not CSV when using `All Items` Create CSV. It appears they have no extension and are treated as a "Document" on Mac. This happens to both Saved Queries and New Queries.

Secondly, when it's a New Query which is not saved. The Type is included with `Selected` and not with the `All Items` option.

Here is an example made from the same query using both options

- Selected Items `New Query Locality - Mon Jun 12 2023.csv`

- All Items `New Query - Mon Jun 12 2023`

**As for the KML Exports**

The Export to KML button is only visible when the Query contains Coordinate data. It would be great if we could resolve #3518
Need to test how Google Earth is accepting Exports

melton-jason:
> Two Things: Exported Files are not CSV when using `All Items` Create CSV. It appears they have no extension and are treated as a "Document" on Mac. This happens to both Saved Queries and New Queries.
> 
> Secondly, when it's a New Query which is not saved. The Type is included with `Selected` and not with the `All Items` option.
> 
> Here is an example made from the same query using both options
> 
> * Selected Items `New Query Locality - Mon Jun 12 2023.csv`
> * All Items `New Query - Mon Jun 12 2023`
> 
> **As for the KML Exports**
> 
> The Export to KML button is only visible when the Query contains Coordinate data. It would be great if we could resolve #3518 Need to test how Google Earth is accepting Exports

Both of these issues should be fixed. Can this be tested again? 

To be consistent, I also hid the Export to CSV button when there are no Query Results. 
If we still want the buttons visible on the Query Results, we can disable them instead of hiding them. 
@specify/ux-testing, let me know which option is preferable.

carlosmbe (changes requested):
Nice work fixing the lack of extension part. However this is still present:

When it's a New Query which is not saved. The Type is included with `Selected` and not with the `All Items` export option.

emenslin (approved):
Both csv and kml names have the correct format, looks good and this seems like it'll make things a lot easier for the users. 
Screenshot 2024-01-11 155102

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -100,9 +100,14 @@ export function QueryExportButtons({
     return resultToExport.join('\n');
   }
 
+  const containsResults =
+    (results.current?.length ?? 0) > 0 &&
```
maxpatiiuk, line 104: hmm, this line seems redundant with the next one
(.every() returns false for empty arrays)

melton-jason: The type of results is `React.MutableRefObject<RA<QueryResultRow | undefined> | undefined>`, thus the line `results.current?.every((row) => row !== undefined)` handles the case when `results.current` is an array and contains elements, but may have `undefined` elements.

maxpatiiuk: ups, I was wrong about `(.every() returns false for empty arrays)` 🤦

.every() returns TRUE for empty arrays

it's `.some()` that returns false for empty

---

thus, in this case, could you use `results.current?.some(row => row !== undefined)` instead?

melton-jason: Fixed in https://github.com/specify/specify7/pull/3590/commits/0a88e94fb07f3c92adf821397c97dca57f14d34e

maxpatiiuk (approved):

grantfitzsimmons (approved):
Fun note (not that I think it's worth holding up this fix for)– exports done on the front-end show Wednesday as the day, exports done on the back-end show Thursday as the day.

image

Screenshot 2024-01-17 at 9 02 32 PM

Screenshot 2024-01-17 at 9 03 21 PM

grantfitzsimmons:
> When it's a New Query which is not saved. The Type is included with `Selected` and not with the `All Items` export option.

@melton-jason Is this something that you have resolved?

carlosmbe (commented):
Not a bug (I think) but when I did an `All Export`, the day was Thursday 18 Jan. However, the selected export was Wednesday 17 Jan. 

I did this test at 11PM on Wednesday

The all export one is the 2MB file with 18 Jan as the file name 

Screenshot 2024-01-17 at 11 18 08 PM

melton-jason:
> > When it's a New Query which is not saved. The Type is included with `Selected` and not with the `All Items` export option.
> 
> @melton-jason Is this something that you have resolved?

This is now resolved. 



> Not a bug (I think) but when I did an `All Export`, the day was Thursday 18 Jan. However, the selected export was Wednesday 17 Jan.
> 
> I did this test at 11PM on Wednesday
> 
> The all export one is the 2MB file with 18 Jan as the file name
> 
> Screenshot 2024-01-17 at 11 18 08 PM

This is very likely caused by #641

emenslin (approved):
This isn't a big enough problem where I'm going to request changes however when you select rows (bottom) vs not select rows (top) one has a space between 'collection object' and the other doesn't 
Screenshot 2024-01-25 123055

melton-jason:
This discrepancy is due to the fact that the frontend uses the localized label for the table and the backend uses the `contextname` of the Query. 
The backend currently has no way to easily get the table localization without re-running the SQL to fetch it. 
(The localization used to be cached, but this was intentionally removed in https://github.com/specify/specify7/commit/b2c982506314cf3d75ff1f6881e0f42413d9ae88). 

At the moment, it may be unwise to re-implement a caching mechanism on the backend. 

So the easiest way to resolve this discrepancy would be to use the database table name on the frontend, rather than the localized table name. 
@specify/ux-testing, would this change be worth maintaining consistency for export file names, should it be kept the same, or should an implementation be made on the backend to easily fetch the localization without hitting the database when not needed?

carlosmbe (approved):

PR #3592, opened 2023-06-07
Change Edit Sp Query btn with up arrow

### LINKED ISSUE #3591: Change Edit Sp Query with up arrow

CarolineDenis:
Fixes #3591


specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx
```diff
@@ -88,17 +87,14 @@ export function QueryHeader({
               })}
         </H2>
         {!queryResource.isNew() && <QueryEditButton query={query} />}
-        <span className="ml-2 flex-1" />
         {!isScrolledTop && (
-          <Button.Small
+          <Button.Icon
+            icon="arrowCircleUp"
+            title="up"
```
maxxxxxdlp, line 93: You forgot to localize "up". And "up" is not very description. How about "Scroll Up" or even better, "Scroll to Editor" (because up and down should be avoided as those are discriminating against the non-visual users)?

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx
CarolineDenis: title in line 322 in RecordSet.tsx should be changed too than?

maxxxxxdlp: Ah, good catch! That is a bug. Can you please fix it on this branch?

maxxxxxdlp (commented):
Looks good, but please also fix the discovered issue in RecordSet.tsx

carlosmbe (approved):
It works fine. Although wouldn't it be better if it matched the colour of the pencil icon next to it? 
Screenshot 2023-06-13 at 12 57 16 PM

CarolineDenis:
> It works fine. Although wouldn't it be better if it matched the colour of the pencil icon next to it?
> Screenshot 2023-06-13 at 12 57 16 PM

@grantfitzsimmons @melton-jason @acwhite211  what do you think?

grantfitzsimmons:
I don't like the idea of it being orange, but I don't know if it fits next to the pencil. I don't know if it fits next to the buttons 😂

I don't have a strong feeling either way. I think I like the black/white color for it, but you all can decide!

maxxxxxdlp:
I don't see the point of this button at all, but at least it doesn't take up much space and is not complicated to implement

melton-jason (approved):
I think the code is fine. 
However, I am also conflicted about the appearance of the buttons. 
Because of this, I will defer to any opinions people on @specify/ux-testing (excluding me) have

I'm not without opinions however. My two cents is if we can tie the background color in with the other button preferences we probably should! I am all for customizability!

For convenience, here are some pictures on how the buttons appear in light and dark mode

light_mode
dark_mode

CarolineDenis:
We can have it displayed as our secondaryButton (which we can customize)
Screenshot 2023-06-20 at 10 05 00 AM
Screenshot 2023-06-20 at 10 05 06 AM

grantfitzsimmons:
> We can have it displayed as our secondaryButton (which we can customize) Screenshot 2023-06-20 at 10 05 00 AM Screenshot 2023-06-20 at 10 05 06 AM

I like the old appearance much better

CarolineDenis:
> > We can have it displayed as our secondaryButton (which we can customize) Screenshot 2023-06-20 at 10 05 00 AM Screenshot 2023-06-20 at 10 05 06 AM
> 
> I like the old appearance much better

same

maxxxxxdlp (approved):

PR #3594, opened 2023-06-08
Add formMeta menu in "View in Forms"

### LINKED ISSUE #2514: FormMeta menu is not available in "View in Forms"
Run query
Click on "View in Forms"
The gear button (Form Meta) menu is not available in the dialog heading

CarolineDenis:
Fixes #2514

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
@@ -313,17 +313,15 @@ export function ResourceView<SCHEMA extends AnySchema>({
       dimensionsKey={viewName ?? resource?.specifyModel.view}
       header={titleOverride ?? title}
       headerButtons={
-        typeof headerButtons === 'function' ? (
-          <>
-            {headerButtons?.(specifyNetworkBadge)} {formPreferences}
-          </>
-        ) : (
-          <>
-            <DataEntry.Visit resource={resource} />
-            <span className="-ml-4 flex-1" />
-            {headerContent}
-          </>
-        )
+        <>
+          {headerButtons?.(headerContent) ?? (
+            <>
+              <DataEntry.Visit resource={resource} />
+              <span className="-ml-4 flex-1" />
+              {headerContent}
+            </>
+          )}
+        </>
       }
       icon="none"
       modal={dialog === 'modal' || makeFormDialogsModal}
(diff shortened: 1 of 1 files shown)


maxxxxxdlp (changes requested):
Your fix might or might not fix the issue in this case, but is not ideal.
The whole point of headerButtons function is to give the component calling ResourceView control over how and where header elements are placed. By adding `{formPreferences}` outside of the function call, you are not giving headerButtons function any control over it.

That means for example, in IntegratedRecordSelector.tsx, the form meta menu would appear after the slider, rather than before like it is supposed to.

Besides being a bug, it's not consistent with how the code behaves in the other case (see line 248 in this file).

Thus, a way to fix the bug, and to make code between the two branches consistent is to undo the changes made in the last commit and then just replace `headerButtons?.(specifyNetworkBadge)` with `headerButtons?.(headerContent)` on line 317

maxxxxxdlp (approved):

melton-jason (changes requested):
Not sure how we would initially get around this, but the `Use localized field labels` checkbox forcing a refresh of the page may be destructive to some workflows in the Query Builder. 

Users would lose the temporary record set, as well as every selected row in the Query Builder (if any). 


Here is a very simple workflow which highlights the problem (The Autogenerated Forms work correctly however!)

https://github.com/specify/specify7/assets/64045831/8a028909-ea78-46fc-b0b6-2cb0ad459b02

CarolineDenis:
> Not sure how we would initially get around this, but the `Use localized field labels` checkbox forcing a refresh of the page may be destructive to some workflows in the Query Builder.
> 
> Users would lose the temporary record set, as well as every selected row in the Query Builder (if any).
> 
> Here is a very simple workflow which highlights the problem (The Autogenerated Forms work correctly however!)
> 
>  simplescreenrecorder-2023-06-13_12.18.29.mp4

Ha yes, since formPreferences change in useRecourceView (due to FormMeta changing) the dialog in ResourceView is re-rendered (due to a change in the headerButtons)?

maxxxxxdlp:
@melton-jason great catch!

It does this right now:

https://github.com/specify/specify7/blob/9d5cbf70e1f3873c1e8743d5eda8f738ea19012d/specifyweb/frontend/js_src/lib/components/FormMeta/Definition.tsx#L104

I though that would alert the user if there are save blockers. Doesn't it?

If not we can just add the code for checking for save blockers and handling that case appropriatly

bronwyncombs (changes requested):
I would recommend removing the checkboxes for `Show Clone button`, `Show Carry Forward button`, and `Show Add button` because they aren't functions in this view

https://github.com/specify/specify7/assets/135047322/75466b5a-9a5c-4751-bba7-b2659038a071

maxxxxxdlp:
Those checkboxes should be there if subview is for an independent relationship and is in a dialog - in that case all the buttons are present (though that is not fully supported yet. will be fixed by https://github.com/specify/specify7/pull/3125)

In all other cases the checkboxes can be removed - but it wasn't worth the trouble to remove them

PR #3602, opened 2023-06-09
Allow to delete a row in WB mapping

### LINKED ISSUE #1732: Delete row in Workbench mapping when not matched to a spreadsheet column
Reported by edinburgh
When creating a workbench mapping from scratch, there is no way to entirely remove a line added.  You can remove the mapping by clicking the x but it doesn't remove the line, which can quickly lead to clutter.
It would be nice if when a line represents a blank space rather than a spreadsheet column, if the line would disappear when you click the x, so where the line then says "New Column x" it would just get deleted.
image

CarolineDenis:
Fixes #1732

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/WbPlanView/LineComponents.tsx
+++ b/specifyweb/frontend/js_src/lib/components/WbPlanView/LineComponents.tsx
@@ -8,9 +8,11 @@ import React from 'react';
 import type { LocalizedString } from 'typesafe-i18n';
 
 import { useId } from '../../hooks/useId';
+import { commonText } from '../../localization/common';
 import { wbPlanText } from '../../localization/wbPlan';
 import type { IR, R, RA } from '../../utils/types';
 import { Button } from '../Atoms/Button';
+import { className } from '../Atoms/className';
 import { icons } from '../Atoms/Icons';
 import type { Tables } from '../DataModel/types';
 import type {
@@ -26,8 +28,6 @@ import {
 } from './CustomSelectElement';
 import type { AutoMapperSuggestion } from './Mapper';
 import type { MappingLineData } from './navigator';
-import { commonText } from '../../localization/common';
-import { className } from '../Atoms/className';
 
 export type HtmlGeneratorFieldData = {
   readonly optionLabel: JSX.Element | string;
--- a/specifyweb/frontend/js_src/lib/components/WbPlanView/LineComponents.tsx
+++ b/specifyweb/frontend/js_src/lib/components/WbPlanView/LineComponents.tsx
@@ -26,6 +26,8 @@ import {
 } from './CustomSelectElement';
 import type { AutoMapperSuggestion } from './Mapper';
 import type { MappingLineData } from './navigator';
+import { commonText } from '../../localization/common';
+import { className } from '../Atoms/className';
 
 export type HtmlGeneratorFieldData = {
   readonly optionLabel: JSX.Element | string;
@@ -46,6 +48,7 @@ type MappingLineBaseProps = {
   readonly onKeyDown: (key: string) => void;
   readonly onClearMapping: () => void;
   readonly isReadOnly: boolean;
+  readonly onRemoveLine: () => void;
 };
 
 export type MappingElementProps = {
@@ -130,6 +133,7 @@ export function MappingLineComponent({
   onFocus: handleFocus,
   onKeyDown: handleKeyDown,
   onClearMapping: handleClearMapping,
+  onRemoveLine: handleRemoveLine,
 }: MappingLineBaseProps): JSX.Element {
   const lineRef = React.useRef<HTMLDivElement>(null);
 
@@ -147,7 +151,18 @@ export function MappingLineComponent({
       aria-labelledby={id('header')}
       className="contents"
     >
-      <div className="border-t border-t-gray-500 py-2 print:hidden">
+      <div className="flex gap-2 border-t border-t-gray-500 py-2 print:hidden">
+        {typeof handleRemoveLine === 'function' && (
+          <Button.Small
+            aria-label={commonText.remove()}
+            className="print:hidden"
+            title={commonText.remove()}
+            variant={className.dangerButton}
+            onClick={handleRemoveLine}
+          >
+            {icons.trash}
+          </Button.Small>
+        )}
         <Button.Small
           aria-label={wbPlanText.clearMapping()}
           className="h-full w-full p-2"
--- a/specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
+++ b/specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
@@ -281,6 +281,12 @@ export function Mapper(props: {
     mappingPathIsComplete(state.mappingView) &&
     getMappedFieldsBind(state.mappingView).length === 0;
 
+  const handleDeleteLine = (line: number): void =>
+    dispatch({
+      type: 'UpdateLinesAction',
+      lines: state.lines.filter((_, index) => index !== line),
+    });
+
   return (
     <Layout
       buttonsLeft={
@@ -480,7 +486,7 @@ export function Mapper(props: {
       <Ul
         aria-label={wbPlanText.mappings()}
         className={`
-          grid flex-1 auto-rows-max grid-cols-[theme(spacing.8)_max-content_auto]
+          grid flex-1 auto-rows-max grid-cols-[theme(spacing.24)_max-content_auto]
           overflow-x-hidden
           print:grid-cols-[min-content_auto]
         `}
@@ -650,6 +656,7 @@ export function Mapper(props: {
                       line: line + 1,
                     });
                 }}
+                onRemoveLine={handleDeleteLine.bind(undefined, line)}
               />
             </ErrorBoundary>
           );
--- a/specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
+++ b/specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
@@ -287,19 +287,21 @@ export function Mapper(props: {
     dispatch({
       type: 'UpdateLinesAction',
       lines: removeItem(state.lines, line),
-      // lines: state.lines.filter((_, index) => index !== line),
+      // Lines: state.lines.filter((_, index) => index !== line),
       changesMade: true,
     });
     props.onDeleteLine(line, props.baseTableName);
-    // dispatch({
-    //   type: 'ChangeChangesMadeAction',
-    //   changesMade: true,
-    // });
-    // dispatch({
-    //   type: 'UpdateLinesAction',
-    //   lines: state.lines,
-    //   changesMade: true,
-    // });
+    /*
+     * Dispatch({
+     *   type: 'ChangeChangesMadeAction',
+     *   changesMade: true,
+     * });
+     * dispatch({
+     *   type: 'UpdateLinesAction',
+     *   lines: state.lines,
+     *   changesMade: true,
+     * });
+     */
   }
 
   return (
@@ -618,21 +620,12 @@ export function Mapper(props: {
                 headerName={headerName}
                 isFocused={line === state.focusedLine}
                 isReadOnly={props.isReadOnly}
-                lineData={fullLineData}
-                // Same key bindings as in QueryBuilder
-                onClearMapping={(): void =>
-                  dispatch({
-                    type: 'ClearMappingLineAction',
-                    line,
-                  })
-                }
                 onFocus={(): void =>
                   dispatch({
                     type: 'FocusLineAction',
                     line,
                   })
                 }
-                onRemoveLine={handleDeleteLine.bind(undefined, line)}
                 onKeyDown={(key): void => {
                   const openSelectElement =
                     state.openSelectElement?.line === line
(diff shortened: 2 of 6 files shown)


maxxxxxdlp (changes requested):
The issue was asking for ability to remove newly added column, not existing column. Newly added lines are those added by clicking on "Add new column" button, before the "Save" button is pressed, thus before they are added to the data set on the back-end. In cases like that, the "Clear" button should be replaced with "Delete" button (and perhaps only if there is no mapping for that line yet).

(not "column", "header" and "line" are used interchangingly to refer to the same thing)

Your code is adding an extra button for each row, thus reducing space available, and it doesn't take care of deleting the data for that column. As a result, all subsequent columns in the data set would be off by one (if I remember the code correctly)

CarolineDenis:
@maxxxxxdlp I investigated the needs during a meeting today for this issue, the request is to have that x button that deletes a row. The UI is what was expected. Although the code need to be changed as the results are not changed yet and the auto incrementing of rows title is not modify neither. There is probably a few other points I still need to work on

maxxxxxdlp:
The original issue is titled "Delete row in Workbench mapping when not matched to a spreadsheet column", which means only delete the row if it's not yet matched to any spreadsheet column - because it's a newly added row (like you see on the screenshot attached to that issue).

Are you sure that during the meeting they didn't misunderstand you, and really requested to be able to delete any column from any existing data set, rather than simply virtual columns created by the "Add new columns" button? There is little practical value from that (especially since you can delete the column before importing the data, or reorder the columns), and biologists are really against deleting data. At the same time, data sets are not intended to live long - import, fix issues, upload, delete - so allowing people to micromanage the data sets is against the intentions we had when designing the work bench.

specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
```diff
@@ -281,6 +282,18 @@ export function Mapper(props: {
     mappingPathIsComplete(state.mappingView) &&
     getMappedFieldsBind(state.mappingView).length === 0;
 
+  const handleDeleteLine = (line: number): void => {
+    dispatch({
+      type: 'UpdateLinesAction',
+      lines: state.lines.filter((_, index) => index !== line),
```
maxxxxxdlp, line 288: Use `removeItem` instead

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingReducer.ts
```diff
@@ -406,4 +412,8 @@ export const reducer = generateReducer<MappingState, MappingActions>({
       lines,
     };
   },
+  ChangeChangesMadeAction: ({ state, action }) => ({
```
maxxxxxdlp, line 415: Remove this -> Merge it with `UpdateLinesAction`

specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
```diff
@@ -281,6 +282,18 @@ export function Mapper(props: {
...
+    });
+    props.onDeleteLine(line, props.baseTableName);
+    dispatch({
```
maxxxxxdlp, line 291: Remove this action -> modify `UpdateLinesAction` to do this
Otherwise it's a bit bug-prone (adds a possibility of developers changing the lines, but forgetting to mark it as "changes made")

Or if you want to make it even better, get rid of the "changesMade" state entirely, and instead compute it dynamically, like it is in the other places based on storing the initial state in a ref and on state changes comparing if lines are the same

specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
```diff
@@ -480,7 +493,7 @@ export function Mapper(props: {
       <Ul
         aria-label={wbPlanText.mappings()}
         className={`
-          grid flex-1 auto-rows-max grid-cols-[theme(spacing.8)_max-content_auto]
+          grid flex-1 auto-rows-max grid-cols-[theme(spacing.24)_max-content_auto]
```
maxxxxxdlp, line 496: (optional) refactor line 487 in this file to replace `&#8594;` with `{'→'}` or maybe even `→`

specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
```diff
@@ -650,6 +663,7 @@ export function Mapper(props: {
                       line: line + 1,
                     });
                 }}
+                onRemoveLine={handleDeleteLine.bind(undefined, line)}
```
maxxxxxdlp, line 666: (optional) try to put smaller callbacks before longer callbacks. Otherwise, it's easy to loose this single line below the giant function that proceeds it.

Generally, callbacks are larger than other props, thus I also try to put callbacks last (consistent and more readable) - I believe we even have an ESLint rule for that

specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
maxxxxxdlp, line 496: The change from 8 to 24 could be a bug. Could you try making the side not be hardcoded but automatic? And if not, then this size should be different depending on whether one delete button or two buttons are displayed

specifyweb/frontend/js_src/lib/components/WbPlanView/helpers.ts
```diff
@@ -81,7 +85,7 @@ export async function savePlan({
       expectedResponseCodes: [Http.NO_CONTENT],
     }
   ).then(async () =>
-    newlyAddedHeaders.length === 0
+    newlyAddedHeaders.length === 0 && !hasDeletedLines
```
maxxxxxdlp, line 88: Is it possible to refactor this code to instead check if original line count != new line count?
A check that does computation like this is generally far less bag prone than a flag that you have to remember to check (the hasDeletedLines flag)

specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
```diff
@@ -281,6 +282,18 @@ export function Mapper(props: {
     mappingPathIsComplete(state.mappingView) &&
     getMappedFieldsBind(state.mappingView).length === 0;
 
+  const handleDeleteLine = (line: number): void => {
```
maxxxxxdlp, line 285: FYI: please use `function` keyword when function declaration requires `{ }` (i.e multiple statements) and `const` when function declaration does not require `{ }` (i.e single statement
OR enable eslint

specifyweb/frontend/js_src/lib/components/WbPlanView/Wrapped.tsx
```diff
@@ -113,6 +115,28 @@ export function WbPlanView({
     )
   );
   useErrorContext('state', state);
+  const [hasDeletedLines, setHasDeletedLines] = React.useState(false);
+
+  const handleDeleteLine = (
```
maxxxxxdlp, line 120: Once you remove `hasDeletedLines`, this entire function won't be needed anymore.

This is a bit confusing but the value of `lines` and `changesMade` between Wrapped and Mapper components may differ. Wrapped has those values initially and takes care of displaying base table selection or Mapper. Then when Mapper is rendered, it uses those as default values, but doesn't need to keep Wrapped informed of updates beyond that point up until onSaved callback.

Ideally, there won't be state duplication like this, but that's how it is (this code was refactored a lot to simplify it, because previous implementation DID have one big state, but that was even worse).

specifyweb/frontend/js_src/lib/components/WbPlanView/index.tsx
```diff
@@ -45,6 +48,48 @@ export function WbPlanViewWrapper(): JSX.Element | null {
   );
   useErrorContext('dataSet', dataSet);
 
+  const updateDataSetColumns = (lines: RA<MappingLine>) => {
```
maxxxxxdlp, line 51: None of this should be needed

WbPlanView calls savePlan, which makes the network request. AND right after that it calls navigate to go back to the workbench. Then, workbench fetches the data set from the back-end, thus you don't need to update the dataSet state in this file manually OR to make a back-end request

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/WbPlanView/Wrapped.tsx
CarolineDenis: if I remove handleDeleteLine in Mapper the <Ul> is not updated. How would you do?

specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
CarolineDenis: if I do this: 
`  function handleDeleteLine(line: number) {
    dispatch({
      type: 'UpdateLinesAction',
      lines: removeItem(state.lines, line),
      changesMade: true,
    });
    props.onDeleteLine(line, props.baseTableName);
  }`
  The delete button seems to have no effect

CarolineDenis:
TODO: 
- find a way to update visualOrder
- modify backend API to allow to delete columns

maxxxxxdlp:
1. savePlan() makes a network request to update the visual order
2. if you only allow deleting columns in new datasets, rather than existing data sets, then there won't be a need for modifying the back-end API

specifyweb/frontend/js_src/lib/components/WbPlanView/Wrapped.tsx
maxxxxxdlp: Are you sure? Can you double check?
I commented out this line, and the UI was still updated properly:
https://github.com/specify/specify7/blob/beb37fef591196b658459c6e9264e00516a48814/specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx#L293

CarolineDenis:
> * savePlan() makes a network request to update the visual order
> * if you only allow deleting columns in new datasets, rather than existing data sets, then there won't be a need for modifying the back-end API

the value of dataSet columns is all the added columns even if some have been deleted. 
Screenshot 2023-06-26 at 12 44 52 PM

specifyweb/frontend/js_src/lib/components/WbPlanView/Mapper.tsx
```diff
@@ -650,6 +651,15 @@ export function Mapper(props: {
                       line: line + 1,
                     });
                 }}
+                onRemoveLine={handleDeleteLine.bind(undefined, line)}
```
maxpatiiuk, line 654: callback uses word `remove`, but function uses `delete` - could you make it consistent?

maxpatiiuk: though, is ther even a reason for handleDeleteLine to exist? can't you just inline it's definition in here, or is it used in multiple places?

specifyweb/frontend/js_src/lib/components/WbPlanView/LineComponents.tsx
```diff
@@ -147,7 +151,18 @@ export function MappingLineComponent({
       aria-labelledby={id('header')}
       className="contents"
     >
-      <div className="border-t border-t-gray-500 py-2 print:hidden">
+      <div className="flex gap-2 border-t border-t-gray-500 py-2 print:hidden">
+        {typeof handleRemoveLine === 'function' && (
```
maxpatiiuk, line 155: according to your typescript type, onRemoveLine is always defined, so why this check in here?

specifyweb/frontend/js_src/lib/components/WbPlanView/helpers.ts
```diff
@@ -81,7 +83,7 @@ export async function savePlan({
       expectedResponseCodes: [Http.NO_CONTENT],
     }
   ).then(async () =>
-    newlyAddedHeaders.length === 0
+    newlyAddedHeaders.length === 0 && !isModified
```
maxpatiiuk, line 86: could you rename `isModified` to `columnCountChanged` to be more descriptive?
`isModified` is a wrong name as mapping could have been modified too - but the variable is only true if column count changed

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingReducer.ts
```diff
@@ -388,9 +388,10 @@ export const reducer = generateReducer<MappingState, MappingActions>({
     }),
     changesMade: true,
   }),
-  UpdateLinesAction: ({ state, action: { lines } }) => ({
+  UpdateLinesAction: ({ state, action: { lines, changesMade } }) => ({
     ...state,
     lines,
+    changesMade,
```
maxpatiiuk, line 394: like in other reducers in this file, changesMade is always set to true when reducer is called, rather than provided as an argument (i.e see line 389)
thus remove the changesMade argument

specifyweb/frontend/js_src/lib/components/WbPlanView/helpers.ts
```diff
@@ -81,7 +83,7 @@ export async function savePlan({
...
       ? Promise.resolve()
```
maxpatiiuk, line 87: code is going to be more complicated than this - rather than simply removing the column from the dataset, you also have to delete the data that was in that column (probably that has to be done on the back-end)

maxpatiiuk (commented):
several issues with the code

though, I am still sligtly against implementing this as it adds additional button to the UI for limited value

I would rather we have ability to remove the line that was added as a result of "Add new columnt", but not saved yet - until it's saved, it's simple to remove, and this is what https://github.com/specify/specify7/issues/1732 was asking for originally
In the UI this would look as "⌫" icon being replaced by trash can icons for such lines that are newly added, not saved yet and don't have a mapping

though even that is probably not worth the trouble as newly added headers are automatically cleaned up when saving if there is no data in them - so there is no need to change the current behavior IMHO

PR #3612, opened 2023-06-10
Issue 2451 c

### LINKED ISSUE #2451: Add pref to automatically have records in read-only mode, then click an Edit button

maxxxxxdlp:
Fixes #2451

Based on #3553, but simplified.

Unfortunately, my local docker is not working well for some reason, so I wasn't able to test that the code works like indented. @CarolineDenis can you take a look at the code to see if it makes sense, and then check if it works?

### DIFF
--- a/specifyweb/frontend/js_src/lib/localization/preferences.ts
+++ b/specifyweb/frontend/js_src/lib/localization/preferences.ts
@@ -1816,4 +1816,7 @@ export const preferencesText = createDictionary({
     'ru-ru': 'Цвет кнопки успеха',
     'uk-ua': 'Колір кнопки успіху',
   },
+  openAsReadOnly: {
+    'en-us': 'Open all record in read-only mode',
+  },
 } as const);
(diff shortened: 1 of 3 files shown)


specifyweb/frontend/js_src/lib/localization/preferences.ts
```diff
@@ -1816,4 +1816,7 @@ export const preferencesText = createDictionary({
     'ru-ru': 'Цвет кнопки успеха',
     'uk-ua': 'Колір кнопки успіху',
   },
+  openAsReadOnly: {
+    'en-us': 'Open all record in read-only mode',
```
CarolineDenis, line 1820: Open all record**s** in read-only mode.

maxxxxxdlp:
Those are all good suggestions. Thanks for testing this.

I opened this PR mainly to show what I meant a good solution might look like, rather than as a final bug-free code (i.e, I wasn't even able to test in the browser because of docker issue)

@CarolineDenis could you please look into resolving the issues you mentioned here - some of them are not exclusive to this implementation and are things we would have had to fix anyway.

@grantfitzsimmons could you also check this branch if the behavior is what we would like in the final version? (apart from the bugs Caroline mentioned)

CarolineDenis:
@maxxxxxdlp I will integrate you code in mine and will fix the issues from there. We can close this branch.

PR #3617, opened 2023-06-12
Allow to add preparation in loan

### LINKED ISSUE #2135: Adding loan preparations doesn't work
https://user-images.githubusercontent.com/37256050/189771842-d2df1fb0-daa4-4dfa-bad1-6a864071f712.mov

Adding a loan preparation creates a read-only item that does not allow linking to an existing CO record. 

In Specify 6, this works properly by having the ability to enter items or add items from a record set:

image

`v7.7.2`
`KUBirds`

CarolineDenis:
Fixes #2135

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormCells/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/index.tsx
@@ -4,7 +4,6 @@ import { useDistantRelated } from '../../hooks/resource';
 import { useAsyncState } from '../../hooks/useAsyncState';
 import { commonText } from '../../localization/common';
 import { formsText } from '../../localization/forms';
-import { f } from '../../utils/functools';
 import { DataEntry } from '../Atoms/DataEntry';
 import { toTable } from '../DataModel/helpers';
 import type { AnySchema } from '../DataModel/helperTypes';
@@ -22,7 +21,6 @@ import { SubView } from '../Forms/SubView';
 import { TableIcon } from '../Molecules/TableIcon';
 import { PickListTypes } from '../PickLists/definitions';
 import { relationshipIsToMany } from '../WbPlanView/mappingHelpers';
-import { FormTableInteraction } from './FormTableInteraction';
 import { PickListEditor } from './PickListEditor';
 
 const cellRenderers: {
@@ -191,7 +189,7 @@ const cellRenderers: {
 
     if (typeof pickList === 'object' && showPickListForm)
       return <PickListEditor relationship={relationship} resource={pickList} />;
-    else if (interactionCollection === false || actualFormType === 'form')
+    else
       return (
         <SubView
           formType={actualFormType}
@@ -203,17 +201,7 @@ const cellRenderers: {
           relationship={relationship}
           sortField={sortField}
           viewName={viewName}
-        />
-      );
-    else
-      return (
-        <FormTableInteraction
-          collection={interactionCollection}
-          dialog={false}
-          mode={mode}
-          sortField={sortField}
-          onClose={f.never}
-          onDelete={undefined}
+          isInteraction={interactionCollection === false ? false : true}
         />
       );
   },
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
@@ -241,105 +241,27 @@ export function IntegratedRecordSelector({
         isLoading,
       }): JSX.Element => (
         <>
-          {
-            typeof recordSetsPromise === 'object' && isInteraction && (
-              <InteractionDialog
-                action={{
-                  model: collection.related?.specifyModel as SpecifyModel<
-                    Disposal | Gift | Loan
-                  >,
-                }}
-                itemCollection={
-                  collection as Collection<
-                    DisposalPreparation | GiftPreparation | LoanPreparation
-                  >
-                }
-                model={schema.models.CollectionObject}
-                recordSetsPromise={recordSetsPromise}
-                searchField={getField(
-                  schema.models.CollectionObject,
-                  'catalogNumber'
-                )}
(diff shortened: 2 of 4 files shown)


maxxxxxdlp (changes requested):
This is actually a bug, where as you fixed it like a missing feature

See https://github.com/specify/specify7/blob/9d5cbf70e1f3873c1e8743d5eda8f738ea19012d/specifyweb/frontend/js_src/lib/components/FormCells/index.tsx#L143-L161

That code checks if it is a regular sub view or a special interaction subview - all good there. The issue is that the special subview is only used if you are in an interaction table **AND** you are displaying subview as a form table - that 2nd condition should not be there.

Find a way to refactor the existing code we have to display a special subview both for form and form table case OR delete that special code and handle both cases in IntegratedRecordSelector.tsx

CarolineDenis:
TODO: 

- fix issue where am empty loan record is added before the created one
- other subviews are not displayed anymore (i.e: loan agent, shipment)

maxxxxxdlp:
To test:
- LoanPreparation, GiftPreparation and DisposalPreparation subforms in grid view on the Loan/Gift/Dispoal forms respecively should behave identically on this branch as they do on production.
- The form view should not work properly, where as it didn't before (see the bug)

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -6,28 +6,46 @@ import { f } from '../../utils/functools';
...
   RecordSelectorProps,
   RecordSelectorState,
 } from './RecordSelector';
 import { useRecordSelector } from './RecordSelector';
-
+const defaultOrder: SubViewSortField = {
+  fieldNames: ['timestampCreated'],
+  direction: 'desc',
+};
 // REFACTOR: encapsulate common logic from FormTableCollection and this component
```
maxxxxxdlp, line 49: Remove that refactor statement after you are done
Also, dont' forget to delete the FormTableCollection file

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -54,6 +74,16 @@ function RecordSelectorFromCollection<SCHEMA extends AnySchema>({
     readonly relationship: Relationship;
     readonly defaultIndex?: number;
     readonly children: (state: RecordSelectorState<SCHEMA>) => JSX.Element;
+    readonly setRecordSetsPromise: React.Dispatch<
```
maxxxxxdlp, line 77: Why is this branch not based on xml-editor?
If it was, then `setRecordSetsPromise` wouldn't even be needed as that branch removed the need for it

specifyweb/frontend/js_src/lib/components/FormCells/index.tsx
```diff
@@ -189,33 +187,22 @@ const cellRenderers: {
...
-        />
-      );
+    return typeof pickList === 'object' && showPickListForm ? (
+      <PickListEditor relationship={relationship} resource={pickList} />
+    ) : (
+      <SubView
+        formType={actualFormType}
+        icon={icon}
+        isButton={isButton}
+        isInteraction={interactionCollection !== false}
```
maxxxxxdlp, line 197: Good catch with the fact that fetching interactionCollection again was redundant.

Since you are no longer using the actual value of the interactionCollection, you don't have to fetch it anymore.
Go ahead and replace the useAsyncState with something much simpler.

```tsx
    const isInteraction = typeof relationship === 'object' &&
          relationshipIsToMany(relationship) &&
          typeof data?.resource === 'object' &&
          [
            'LoanPreparation',
            'GiftPreparation',
            'DisposalPreparation',
          ].includes(relationship.relatedModel.name);
```
<p>



</p>
</details> 

And now that it's no longer async, you don't have to do place it in here and can instead place it much closer to where it's used (probably in `IntegratedRecordSelector`)

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -150,10 +192,11 @@ export function IntegratedRecordSelector({
   relationship,
```
maxxxxxdlp, line 192: Once this branch is based on xml-editor, could you please move this component to be above `RecordSelectorFromCollection`? That would make much more sense as the current order is wrong (i.e, naturally programmers read code from top to bottom, so components should be laid out in that order too) OR just move RecordSelectorFromCollection into a separate file

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -211,67 +247,103 @@ export function IntegratedRecordSelector({
...
-                ) && typeof handleAdd === 'function' ? (
-                  <DataEntry.Add
-                    disabled={
-                      mode === 'view' ||
-                      (isToOne && collection.models.length > 0)
+          {typeof recordSetsPromise === 'object' && isInteraction ? (
+            <InteractionDialog
+              action={{
+                model: collection.related?.specifyModel as SpecifyModel<
+                  Disposal | Gift | Loan
```
maxxxxxdlp, line 254: Moving the definition of `isInteraction` to this component might allow to simplify this typing (and simplify the definition of isInteraction).

<p>



```tsx
const interaction = typeof collection.related === 'object' ? (
  toTable(collection.related,'LoanPreparation') ??
  toTable(collection.related,'GiftPreparation') ??
  toTable(collection.related,'DisposalPreparation')
) : undefined;
```
</p>
</details>

maxxxxxdlp (changes requested):
Functionally looks good
And I love how much simpler the code would be after your refactor.
See my comments on how to simplify it futher

CarolineDenis:
Resolved in https://github.com/specify/specify7/pull/3659 using xml-editor as a base

PR #3624, opened 2023-06-15
Add the possibility to see more than 1 attachment for a record with Gallery

### LINKED ISSUE #1048: Implement "icon" view type / Add ability to see more than one attachment thumbnail at a time

CarolineDenis:
Fixes #1048


maxxxxxdlp (changes requested):
Code logic is valid, but could be a bit improved:

1. Please move as much of the attachment related logic into a separate component (which will still call AttachmentGallery like it does now). The logic for attachments does not belong here and this file is already too complicated. When I was talking about not creating new components for displaying attachments, I should be more clear I mean not recreating AttachmentGallery, but integrating with it (and if necessary creating a wrapper component for integrating with it).
2. Then, whatever component you come up, just call that component in the code only if table is an attachment table. i.e `const isAttachmentTable = ...;  /* later on ... */  return ... {isAttachmentTable && <AttachmentRecordSelectorOrSomeBetterName collection={collection} {/* other props here if you need them */} />} ...;`
3. serializerResource is a semi-expensive operation, and get's more expensive if there are a lot of attachments. wrap the code that maps and calls serializerResource in React.useMemo. And since the component is only called for attachments, you can simplify the code quite a bit

    <p>
    
      ```tsx
      const attachments = React.useMemo(()=>collection.models.map(serializerResource),[collection.models]);
      
      // assuming typescript will let you, that's all you need to do, though maybe you would still need a type hint:
      
      const attachments = React.useMemo<RA<SerializedResource<Attachment>>>(()=>collection.models.map(serializerResource),[collection.models]);
      
      // or even if that doesn't work, then a type assertion:
      
      const attachments = React.useMemo(()=>collection.models.map(serializerResource) as RA<SerializedResource<Attachment>>,[collection.models]);
      ```
    
    </p>
    </details> 

4. You used Array.from(). There are valid use cases for it, but it's not necessary in this case as simple .map() does the job. Array.from is needed when you want to iterate over something that does not have .map (document.querySelectorAll() or new Set()) or if you want to iterate x number of times: `Array.from({length: 10}, (_, index)=> ... )`

maxxxxxdlp (approved):
You are right, none of those versions worked. I pushed a commit that simplifies the typing a bit. The rest of the code looks good - though of course someone would still need to test it in the UI

specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
```diff
@@ -25,15 +23,11 @@ export function AttachmentsCollection({
 
   const [scale = defaultScale] = useCachedState('attachments', 'scale');
 
-  const attachments: RA<SerializedResource<Attachment>> = React.useMemo(
+  const attachments = React.useMemo(
     () =>
-      filterArray(
-        collection.models.map((model) => {
-          if (model.specifyModel.name === 'CollectionObjectAttachment') {
```
maxxxxxdlp: This wasn't correct because it assumed that only collection object attachments are used (many other tables have attachments too)

bronwyncombs (approved):
Working as expected! Once assets are up I will check how different sized images look.

https://github.com/specify/specify7/assets/135047322/011aff84-0c5e-45c1-a819-cbbb3c8b60f8

Perhaps we can change the hover label to say "Gallery" or "Gallery View"

Screenshot 2023-07-17 at 10 35 42 AM

bronwyncombs (changes requested):
Perhaps we can change the hover label to say "Gallery" or "Gallery View"

Screenshot 2023-07-17 at 10 35 42 AM

maxxxxxdlp:
> Perhaps we can change the hover label to say "Gallery" or "Gallery View"

Attachments match the name of the `Attachments` page in the main menu in sp6 and in sp7

We could rename the page too, but that would require redoing documentation too.

Also, `gallery`, at least in my mind is associated with pictures, where as `attachments` could mean any kind of file - as people can and often do attach audio files, pdf documents and maybe even videos

maxxxxxdlp (approved):

bronwyncombs (changes requested):
Images do not load in the gallery view dialog. 

https://github.com/specify/specify7/assets/135047322/72913459-ddb5-4944-adc3-f1af3b716c0e

Also, can we please simplify the shadowing in the gallery view dialog? The dimension around the image borders are fine, but I think we need to remove the border shadow of the intermediate dialog. Right now, it looks like there's three layers at the bottom and one or none at the top. If we could have a flat white dialog with dimension/shadow added to just the image borders that would be great! 

Screenshot 2023-07-19 at 10 24 37 AM

CarolineDenis:
@bronwyncombs totally agree on the shadows. This as been fixed in https://github.com/specify/specify7/pull/3363/files#diff-5e52f0d2188182a6b256bf141b4e71bd6fb7a2aa0c696dedac1e8fbc1d1191c8 so I didn't duplicate the changes.

CarolineDenis:
@bronwyncombs, for your comment: "Images do not load in the gallery view dialog." can you see them in the gallery page? I think the issue is related to gallery in general not this PR but I might be wrong

CarolineDenis:
@bronwyncombs I changed the shadows to what has been decided in https://github.com/specify/specify7/pull/3363

bronwyncombs:
@CarolineDenis 

> for your comment: "Images do not load in the gallery view dialog." can you see them in the gallery page? I think the issue is related to gallery in general not this PR but I might be wrong

They are loading in the attachment gallery :/  


https://github.com/specify/specify7/assets/135047322/a870df6a-c621-437a-bd6b-a6f7955c431a

---

> I changed the shadows to what has been decided in https://github.com/specify/specify7/pull/3363 

This part looks good! 

Screenshot 2023-07-19 at 1 20 55 PM
Screenshot 2023-07-19 at 1 22 35 PM

specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
```diff
@@ -23,13 +25,14 @@ export function AttachmentsCollection({
...
-      ),
-    [collection]
+  const attachments: RA<SerializedResource<Attachment>> = filterArray(
+    Array.from(collection.models, (model) => {
+      if (model.specifyModel.name.includes('Attachment')) {
+        const record = serializeResource(model) as SerializedResource<CollectionObjectAttachment>
+        return serializeResource(record.attachment) as SerializedResource<Attachment>
+      }
+      return undefined;
+    })
```
maxxxxxdlp, line 35: huh, why was this changed again to a more complex version?
does the simpler version not work?

maxxxxxdlp (changes requested):

bronwyncombs:
@CarolineDenis @grantfitzsimmons 

> Not related, but the positioning of the attachments icon is weird (see 1st screenshot) - it should either be to the left of the slider, next to the +/- icons, or should be styled to look like a button

@maxxxxxdlp's [comment on #3804](https://github.com/specify/specify7/issues/3804#issuecomment-1642900799)

specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
CarolineDenis: no it didn't... we have found this solution with Jason

bronwyncombs (changes requested):
Will you disable the right page arrow in the browsing view (Dialog when item is clicked on in gallery view) when on the last and/or only attachment in the gallery? 

https://github.com/specify/specify7/assets/135047322/8982bc8e-09f1-463a-ad09-877353d27464

Outside of that change, this is working as expected! Images are loading in dialog and formatting looks good. New button icon also looks good. 

https://github.com/specify/specify7/assets/135047322/c8d85469-ddb5-42b1-834c-d0ba3a51ba94

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
maxxxxxdlp: Got it. thanks for clarifying

bronwyncombs (approved):
Page arrows behave as expected! Everything looks great! 

Last page in gallery:
Screenshot 2023-07-21 at 10 14 03 AM

Only page in gallery: 
Screenshot 2023-07-21 at 10 15 04 AM

Styling in dark mode: 
Screenshot 2023-07-21 at 10 19 32 AM

specifyweb/frontend/js_src/lib/components/Attachments/Dialog.tsx
```diff
@@ -91,12 +95,14 @@ export function AttachmentDialog({
     >
       <div className="flex h-full gap-4">
         {/* FEATURE: keyboard navigation support */}
-        <Button.Icon
-          className="p-4"
-          icon="chevronLeft"
-          title={commonText.previous()}
-          onClick={handlePrevious}
-        />
+        {isFirst ? null : (
```
maxxxxxdlp, line 98: why not just hide the button if `handlePrevious` is undefined?
same for the other button - that will remove the need for the 2 new props

maxxxxxdlp: but really, I would advise against dynamically hiding or showing buttons in a dialog:
this can cause the dialog width to change, and so for example, if user was going to click the button twice rapidly to go left two times, if the button has disappeared on the 2nd click, the user may end up clicking outside the dialog accidentally - thus causing the dialog to close.
@grantfitzsimmons can tell you more about this UX issue

besides that there is a smaller annoyance but still an issue: buttons appearing and disappearing dynamically on the sides of the dialog, mean the dialog content can change it's position, which causes layout shift needlessly

overall, I don't think any of these issues are worth saving a few pixels. @grantfitzsimmons what do you think?

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Attachments/Dialog.tsx
grantfitzsimmons: The left and right buttons should be always visible, even when there is not another attachment to the right or left. It should just be disabled. Max is right about the usability issue with this!

maxxxxxdlp: actually, there might be another issue with buttons disappearing:
if an element is focused, and that element disappears, the focus is lost, and so get's reset to the first focusable element on the page.
in this case, if user clicks on the button using the keyboard, and the next record happens to be last, the button would disappear and focus is lost

maxxxxxdlp (approved):

PR #3628, opened 2023-06-16
Add additional instructions for django migrations

### LINKED ISSUE #3626: Upon first run of Specify 7 with a Specify 6 database, an error occurrs
```
{"exception": "ProgrammingError", "message": "1146", "data": "Table 'paleo.spuserpolicy' doesn't exist",
```

Migrations are never run the first time you run `just-specify7` with a new Specify 6 database. We need to solve this so that users do not need to follow this guide each time they start using Specify 7– it gives a bad impression.

https://discourse.specifysoftware.org/t/upgrading-to-specify-7-7-http-500-error/733

grantfitzsimmons:
Fixes #3626

Is this explanation satisfactory?

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,13 @@ services:
       - MYSQL_USER=master
       - MYSQL_PASSWORD=master
 
+# When running Specify 7 for the first time, make sure that the
+# MASTER_NAME and MASTER_PASSWORD are the root username and password.
+#
+# This ensures that the Django migrations are run properly the first
+# time. Once you are in the application, you can safely replace
+# these credentials with the master SQL user name and password.
+
   specify7:
     build:
       context: ./.


maxxxxxdlp (changes requested):
1. Small clarification: It's not that root is required, but that user with more permissions is required (my master user has all permissions, so I had no issues with this - of course that is not suitable for production). Worth clarifying or not worth the complexity?
2. When you modify docker compose, I believe you have to rebuild for changes to env variables to take effect - not sure. Could you try changing the MASTER_NAME and see if changes apply after restarting without rebuild?

grantfitzsimmons:
> * Small clarification: It's not that root is required, but that user with more permissions is required (my master user has all permissions, so I had no issues with this - of course that is not suitable for production). Worth clarifying or not worth the complexity?

I understood this before writing, but I was worried it would add too much complexity to explain this.

> When you modify docker compose, I believe you have to rebuild for changes to env variables to take effect - not sure. Could you try changing the MASTER_NAME and see if changes apply after restarting without rebuild?

I can try this at a later date. I need to catch up on 3 weeks of messages :wink:

maxxxxxdlp (approved):

Issue #1746, opened 2022-06-10
Add keyboard shortcuts for common actions


maxxxxxdlp:
Like this:
- Particular menu item
- Common actions on different pages (save query, open mapping editor, ...)
- Next/previous record selector item (suggested by @beach53)

To meet the WCAG 2.1 requirements, these shortcuts must be customizable

maxxxxxdlp:
@CarolineDenis This one might be interesting for you

I have the code for the keyboard picker component already written so can push it if you want to work on it
The remaining thing would be to create a hook like `useKey('category', 'subcategory', 'item', callback)`
where `'category', 'subcategory', 'item'` matches what you see in user prefs, and the callback is a function that will be called by the hook when the user presses the keyboard shortcut.

Edge cases to handle:
If user assigned a shortcut like `B` (without Ctrl/alt/or other modified), than you have to make sure to only trigger the keyboard shortcut if user is not currently typing into a text field

grantfitzsimmons:
>1. When browsing records in forms view, I miss a keyboard shortcut for moving to the next/previous record.

This would be so nice to have! 

**Requested By:** Eyal at The Hebrew University of Jerusalem

maxpatiiuk:
> Implementation idea:
Split user preferences into 3 ultra-sections:
Options (everything that is not in the next two)
Theme (for setting UI colors - and add ability to customize more colors)
Keyboard Shortcuts

I don't think this is the best solution anymore.
Better implementation:
- Keep user preferences as is
- Add a new "Keyboard Shortcuts" item to User Tools menu that will have UI identical to user preferences (will use the same react component under the hood), but will be used only for setting keyboard shortcuts
- Do not add ways to arbitrary customize every color used in specify. Reasons:
  - that may only be used by more advanced users
  - the customized colors may make the specify appear broken (i.e when contrast becomes very low or non existent in certain places)
  - it would be a pain to maintain this in the future as specify UI changes without breakages to customizations people did
  - Instead, if there would be desire to expand themeing, an approach similar to what Grant suggested in https://github.com/specify/specify7/issues/1460 would be more manageable - extend the "Dark" and "Light" mode themes with more variants (Blue, Pink, Beige, ...)
  - optionally in the future also reduce the ability to pick 6 accent colors in specify to just 2 accent colors, from which specify will derive the other 4 by means of increasing and decreasing luminosity or other HSL manipulations

maxpatiiuk:
Did a bit of research on this.

Complications:
[...]

PR #3632, opened 2023-06-16
Allow to slide between attachments in gallery attachment preview with…

### LINKED ISSUE #3057: Add option to use arrow keys to go between attachment previews
This would be the same on a record or in the Attachments viewer page

https://user-images.githubusercontent.com/37256050/221669063-d8cffb04-378e-48c1-af35-0d861c7abaad.mov

Navigation could be performed by pressing ⬅ or ➡️ keys rather than the buttons (since they shift place)

CarolineDenis:
… key board arrows

Fixes #3057

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Attachments/Dialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Attachments/Dialog.tsx
@@ -50,6 +50,22 @@ export function AttachmentDialog({
 
   const [showMeta, _, __, toggleShowMeta] = useBooleanState(true);
 
+  React.useEffect(() => {
+    const handleKeyPress = (event: KeyboardEvent) => {
+      if (event.key === 'ArrowLeft' && handlePrevious) {
+        handlePrevious();
+      } else if (event.key === 'ArrowRight' && handleNext) {
+        handleNext();
+      }
+    };
+
+    document.addEventListener('keydown', handleKeyPress);
+
+    return () => {
+      document.removeEventListener('keydown', handleKeyPress);
+    };
+  }, [handlePrevious, handleNext]);
+
   return (
     <Dialog
       buttons={
--- a/specifyweb/frontend/js_src/lib/components/Molecules/KeyboardShortcut.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/KeyboardShortcut.tsx
@@ -1,0 +1,196 @@
+/**
+ * Logic for setting and listening to keyboard shortcuts
+ */
+
+import React from 'react';
+
+import { useTriggerState } from '../../hooks/useTriggerState';
+import { commonText } from '../../localization/common';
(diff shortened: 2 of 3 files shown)


carlosmbe:
This works from a UX POV

melton-jason (changes requested):
Currently, this allows the user to navigate between attachments in the Gallery even when the Gallery is not the 'most recent' or 'topmost' dialog.

In the below video, I go to a different dialog and press the right arrow and left arrows. 
Pay close attention to the tab name, the Network section of chrome dev tools, and the console. 
https://github.com/specify/specify7/assets/64045831/86aad74c-7329-4615-ad15-4e1175f4fe8d

maxxxxxdlp (changes requested):
I would suggest instead of hardcoding a solution for this exact case, create a UI where users can set and customize the keyboard shortcuts for all sorts of actions.

See https://github.com/specify/specify7/issues/1746

In particular this part:
```
Split user preferences into 3 ultra-sections:
Options (everything that is not in the next two)
Theme (for setting UI colors - and add ability to customize more colors)
Keyboard Shortcuts
```

Then, create a hook `useShortcut` where first 3 arguments are category, subcategory and item names (just like any other pref), but the 4th argument is the callback that would be triggered when a keyboard shortcut assigned to that action is triggered:

```tsx
useShortcut( 'attachment', 'action', 'next', handleNext);
useShortcut( 'attachment', 'action', 'previous', handlePrevious);
```

Additional benefits of this approach:
- Once you implement the infrastructure for this once, it becomes trivially easy to add keyboard shortcuts in a lot of places in specify 7
- Keyboard shortcuts can be reassigned or disabled, which is a requirement of WCAG 2.1 AA standard
- Whatever keyboard shortcuts the user sets, they should be scoped to the OS (i.e, if the user opened prefs from Windows, and set some shortcut, and then opened prefs on macOS, the shortcut should be set again, same for linux). I.e, that would be stored something like this this: `Record<OsName, RA<KeyboardShortcut>>`. This is because system and browser shortcuts differ a lot between operating systems, so it's not safe to assume that one shortcut would work on all systems
- Also, when implementing useShortcut, there is one super important thing that you should handle that is not handled in your code at the moment: if the keyboard shortcut does not involve cmd/option/control keys, than it should only be activated if `<input>`, `<textarea>` and `<select>` element is not focused at the moment (i.e, if user is typing in a text box, and hits LeftArrow, you should not go to next attachment!)

Yes, that increases the scope of the task, but it would be SUPER GREAT to implement this very well once and reuse it in a lot of places.

But, things are a bit simpler, because I already implemented the UI for setting and editing keyboard shortcuts. See https://github.com/maxxxxxdlp/calendar-plus/blob/main/src/src/components/Molecules/KeyboardShortcut.tsx. I pushed a commit where I carried that file over to the sp7 repo.

Though, note that to minimize merge conflicts, I moved this feature to be based on xml-editor branch

PR #3639, opened 2023-06-17
Remove send email button from information request form

### LINKED ISSUE #2211: Send Email button should not be on the form as Specify 7 does not support that action
image

On the Information Request form, there is an option to "Send Email on Save" that doesn't function in Specify 7. This should be hidden as it has no function.

CarolineDenis:
Fixes #2211

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
+++ b/specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
@@ -110,8 +110,11 @@ const processFieldType: {
       console.error("Can't render a check box for a relationship field");
       return { type: 'Blank' };
     }
-    const isSendEmail = name === 'sendEmail';
-    if (isSendEmail) {
+    /**
+     * The send email button is not supported in Specify 7
+     * See https://github.com/specify/specify7/issues/2211
+     *  */
+    if (name === 'sendEmail') {
       return { type: 'Blank' };
     }
     return {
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
```diff
@@ -111,6 +110,10 @@ const processFieldType: {
       console.error("Can't render a check box for a relationship field");
       return { type: 'Blank' };
     }
+    const isSendEmail = name === 'sendEmail';
+    if (isSendEmail) {
+      return { type: 'Blank' };
+    }
```
melton-jason, line 116: Could you include a comment stating the sendEmail button is not supported in Specify 7 and provide a link to the issue?

I am also not sure if the `isSendEmail` variable is needed. I personally think the condition itself (when paired with a comment) would make it clear enough

melton-jason: I would imagine something like this would be sufficient

```js
/**
     * The send email button is not supported in Specify 7
     * See https://github.com/specify/specify7/issues/2211
     *  */
    if (name === 'sendEmail') {
      return { type: 'Blank' };
    }
```

melton-jason (changes requested):

carlosmbe:
Which Database should I use to test this?

CarolineDenis:
> Which Database should I use to test this?

I guess any should work as long as it is in the xml. 
The information request (name='InfoRequest') form should have: 
<row>
     <cell type="field" uitype="checkbox" id="5" name="sendEMail" label="Send Email on Save" ignore="true" default="true" colspan="6"/>
</row>

grantfitzsimmons:
It's already fixed!? Not sure how.

PR #3640, opened 2023-06-17
Vertically decenter subviews in forms

### LINKED ISSUE #2108: Cell that have subviews should not be vertically centered
image

db: `newmexico_august22`
CO form

CarolineDenis:
Fixes #2108


carlosmbe (approved):
Works in both `newmexico_august22` and `nhmdtest030823` Collection Object Data Entry Forms

maxxxxxdlp (approved):
Looks good. @CarolineDenis could you either allow users to customize this or open a feature request about being able to customize it?

I.e, that would be quite simple to implement, yet useful for users (and you know how to add new attributes to the form as we did it with the subview the other day - really, just look at how it listens for "align" right now and do something very similar for vertical align). Then look though **possible values** and see which of those we have to support (probably just 3 or 4 of the most common)

CarolineDenis:
> Looks good. @CarolineDenis could you either allow users to customize this or open a feature request about being able to customize it?
> 
> I.e, that would be quite simple to implement, yet useful for users (and you know how to add new attributes to the form as we did it with the subview the other day - really, just look at how it listens for "align" right now and do something very similar for vertical align). Then look though **possible values** and see which of those we have to support (probably just 3 or 4 of the most common)

So to be sure, you would like: 
- `readonly verticalAlign? : boolean` prop in DataEntry.Cell to be => `readonly verticalAlign? : string` 
- in the cell style, instead of: `alignSelf: verticalAlign ? 'stretch' : undefined` 
we would have: 
          `alignSelf: verticalAlign === 'strecth' 
                     ? 'stretch' 
                     : verticalAlign === 'center'
                     ? 'self-center'
                     : verticalAlign === 'start'
                     ? 'self-start'
                     : verticalAlign === 'end'
                     ? 'self-end'`
 and that would be it

maxxxxxdlp:
Yes that's on the right track.

Though instead of `readonly verticalAlign? : string` do something like `readonly verticalAlign? : 'stretch' | 'center' | 'start' | 'end'`
or even create a type for it `export type VerticalCellAlignment = 'stretch' | 'center' | 'start' | 'end';` because it would be used in multiple places

And then of course modify the parsing code for cells (not just sub view, but any cells -> not hard, just means that unlike when we added an attribute to SubView parser, you would add the attribute to a parent function (the cell parser)

Finally, after everything is ready, update the form wiki to describe the new attribute and possible values (http://github.com/specify/specify7/wiki)

specifyweb/frontend/js_src/lib/components/Atoms/__tests__/DataEntry.test.ts
```diff
@@ -35,29 +35,32 @@ describe('DataEntry.Cell', () => {
       children: 'Test',
       colSpan: 1,
       align: 'right',
+      verticalAlign: 'stretch',
       visible: true,
     },
-    'colspan 1, align right'
+    'colspan 1, align right, self stretch'
```
maxxxxxdlp, line 41: Nice! I like that you didn't forget to update the tests
Though, for best coverage, could you make each of these tests use a different value of verticalAlign?

specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
```diff
@@ -98,12 +98,13 @@ export const DataEntry = {
       readonly colSpan: number;
       readonly align: string;
       readonly visible: boolean;
+      readonly verticalAlign: string;
```
maxxxxxdlp, line 101: Replace `string` with `typeof cellVerticalAlign[number]`

specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
```diff
@@ -98,12 +98,13 @@ export const DataEntry = {
       readonly colSpan: number;
       readonly align: string;
```
maxxxxxdlp, line 99: Could you replace `string` with `typeof cellAlign[number]`
Thanks

specifyweb/frontend/js_src/lib/components/FormParse/cells.ts
```diff
@@ -329,6 +332,9 @@ export function parseFormCell(
       : cellType === 'Label'
       ? 'right'
       : 'left',
+    verticalAlign: f.includes(cellVerticalAlign, verticalAlign)
+      ? verticalAlign
+      : 'stretch',
```
maxxxxxdlp, line 337: So it looks like, before this PR, the vertical alignment was center.
But after this PR, it would be top instead.
Thus, you should either:
1. Test the implications of this change in a variety of scenarious
2. Do what line 332 in this file does, and apply 'stretch' only for SubView cells, and center for all others

maxxxxxdlp (changes requested):
All good. Just small things

specifyweb/frontend/js_src/lib/components/FormParse/__tests__/cells.test.ts
```diff
@@ -99,14 +99,15 @@ const cell = (
...
     >
 ): FormCellDefinition => ({
   id: undefined,
   colSpan: 1,
   align: 'left',
   visible: true,
   ariaLabel: undefined,
+  verticalAlign: 'stretch',
   ...cell,
 });
```
maxxxxxdlp, line 113: Could you modify any of the tests in this file to have 2-3 tests that include verticalAlign?

specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
CarolineDenis: why don't we do the same for line 100 align? ==> `typeof cellAlign[number]`?

CarolineDenis: sorry never mind just saw your next comment

CarolineDenis:
@maxxxxxdlp let me know if the changes made in cells.ts are what you wanted (center the cells except if subview, stretch or value define by user). 
PS: tests are failing at the moment because of verticalAlign

specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
maxxxxxdlp: yeah, sorry, I wrote these comments in the wrong order)

specifyweb/frontend/js_src/lib/components/FormParse/cells.ts
```diff
@@ -332,9 +332,12 @@ export function parseFormCell(
       : cellType === 'Label'
       ? 'right'
       : 'left',
-    verticalAlign: f.includes(cellVerticalAlign, verticalAlign)
-      ? verticalAlign
-      : 'stretch',
+    verticalAlign:
+      cellType === 'SubView'
```
maxxxxxdlp, line 336: This is wrong order of ternaries - you have to swap them.

First check if user explicitly specified a valid alignment, and if so use it, and only if they didn't specify an alignment, you should use a default, which would be determined by a ternary based on cellType

In your case, you prohibited the user from customizing the alignment for anything other than subview - defeats the point of the feature

maxxxxxdlp (changes requested):
One big issue, but the rest is good - and of course the failing tests

specifyweb/frontend/js_src/lib/components/FormParse/__tests__/cells.test.ts
```diff
@@ -129,7 +131,7 @@ describe('parseFormCell', () => {
       parseFormCell(
         schema.models.CollectionObject,
         strictParseXml(
-          '<cell invisible="true" type=" test2 " initialize="align=Center" colSpan=" 5 " id="test" />'
+          '<cell invisible="true" type=" test2 " initialize="align=center; verticalAlign:center" colSpan=" 5 " id="test" />'
```
melton-jason, line 134: > verticalAlign:center

This doesn't seem to be consistent with other attributes defined in a cell's `initialize` (it's also inconsistent with other verticalAlign (sub?)attributes used in this file)

Should it be `verticalAlign=center`?

So the initialize attribute would look like: 
```xml
initialize="align=center; verticalAlign=center"
```

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/FormParse/__tests__/cells.test.ts
```diff
@@ -393,10 +409,10 @@ describe('parseFormCell', () => {
       parseFormCell(
         schema.models.CollectionObject,
         strictParseXml(
-          `<cell type="panel" colDef="1px,2px,2px">
+          `<cell type="panel" colDef="1px,2px,2px" initialize="verticalAlign=center;">
```
maxxxxxdlp, line 412: Good job on updating tests to be comprehensive. All looks good.

A small lesson for the future in regard to tests:
It's advisable to keep tests small and not to test the same feature multiple times. Reasons:
1. Smaller tests are easier to maintain (i.e, less code to maintain) and easier to read (i.e, less code to read and to understand)
2. If for example, you rename the verticalAlign attribute or do some other change, if you only had 3 tests testing that feature, then there would be less refactoring work than if you had 9 tests
3. Each new tests means tests run slower, thus redundant tests could be waisting developer's time

Thus, it's very important to have a test for a feature, and in some cases to test that feature comprehensively in different scenarios (vertical align center, left, right), but it's also important not to test the same thing multiple times.

In this case it's not a big deal to test `verticalAlign=center` multiple times, but for some bigger things (i.e automappaer or testing a react component) try to have **comprehensive and succinct** tests.

maxxxxxdlp (approved):

melton-jason (approved):

PR #3642, opened 2023-06-19
Add collapsible feature in forms 

### LINKED ISSUE #1995: Collapsible subviews are missing
> The initialize="collapse=true" command in the form xml doesn’t appear to be working in S7. We have a subform in the collectionobject table that we’d like to open in a collapsed way. This works fine in S6, but not in S7.
>  
> `<cell type="subview" id="88" viewname="CollectionObjectProperty" name="collectionObjectProperties" colspan="14" initialize="collapse=true"/>`

In Specify 6, subviews could be collapsed. In Specify 7, this does not seem to be supported.

Apologies if this has already been mentioned.

**Reported By:** Corinna at CSIRO

CarolineDenis:
Fixes #1995


CarolineDenis:
https://github.com/specify/specify7/assets/108160931/afa78280-e4ec-4039-98c7-46d1dbff065a

CarolineDenis:
https://github.com/specify/specify7/assets/108160931/c50c289b-0eee-43c9-9d1e-f50fc4bb5ea3

specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
```diff
@@ -189,6 +196,20 @@ export function FormTable<SCHEMA extends AnySchema>({
     scrollerRef
   );
 
+  const resourceHasBlockersContext = React.useContext(HasBlockersContext);
```
maxxxxxdlp, line 199: Why not simplify it?

```tsx
const hasBlockers = React.useContext(HasBlockersContext);
React.useEffect(()=>{
  if(hasBlockers) handleExpand();
},[hasBlockers, handleExpand]);
```

maxxxxxdlp: Even more, you don't need context in this case (Though I do like that you know how to use it).

Simply add a `parentResource` prop to this component and then you can get hasBlockers like this:

```tsx
const blockers = useAllSaveBlockers(parentResource, relationship);
const hasBlockers = blockers.length > 0;
```

maxxxxxdlp (changes requested):
Can you please tell why `useBlockerHandler` didn't work for you? If there is a bug with useBlockerHandler then it would be important to know about it and fix it.

If it's just a matter of having to add `parentResource` prop, then surely it's simpler than creating a context and two React.useEffects with setTimeouts?

Please tell if there are issues with the following:

```tsx
useBlockerHandler(parentResource, relationship, hangleExpand);
```

Simply add this prop to FormTable:
```tsx
readonly parentResource: SpecifyResource<AnySchema> | undefined;
```

And in IntegratedRecordSelector, just add this to FormTableCollection:
parentResource={collection.related}

(and no need even to change anything in FormTableCollection)

CarolineDenis:
@maxxxxxdlp I used in this latest version "useAllSaveBlockers", it makes the code way more easy to read and I could get rid of the context and avoid a few line of code :+1:
I still had to use a useEffect and time out. One to avoid to many renderers, the other because it would uncollapse the first time but wouldn't if you add a record with a mistake than wait than collapse. 
The useBlockerHandler didn't work because when the element is collapsed it's hidden and therefore non-focusable.

grantfitzsimmons (changes requested):
image
image

I believe the Specify 6 symbols are more intuitive, where there is a right-facing arrow when collapsed and a downward facing arrow when expanded.

I don't know why I feel this way, since up to collapse and down to expand seems more "logical"– it just feels inverted somehow.

maxxxxxdlp:
@CarolineDenis the useBlockerHandler hook is intended to automatically uncollapse the subview if the user clicks on the "Save" button, rather than prevent the user from collapsing the subview if there is an error (based on a 2 second timer - a cludge).

I was thinking uncollapsing on save button click rather than timer would be less confusing to the user, but am open to suggestions.

maxxxxxdlp:
For example consider a case where a field in a collapsed subview becomes invalid as part of an edit to a field outside the subview. That could happen because of business rules. That could also happen because (on xml-editor branch) users can embed the fields from a dependent -to-one relationship directly on the main form, without using a sub view.

maxxxxxdlp:
And even if you want to go with the current approach, I still don't think a timer is the right approach.

Why not just disable the collapse button if there is an error OR even better, when user clicks on the button, focus the error field instead of collapsing? (that can be done by calling form.reportValidity(). You can add `readonly form: HTMLFormElement | undefined;` to FormContext->FormMetaType to accomplish this). That, while being a bit more complicated, results in a better UX.

Just look at how few usages of setTimeout there are in the codebase - and all of them are there for a good reason (i.e, for performance, or for auto-dismissing messages)

maxxxxxdlp:
I had this error even with the setTimeout hack:

Screenshot 2023-06-24 at 18 00 43

---

Feature request: either hide the add/remove buttons when view is collapsed, OR better: expand the subview when add/remove/slider is pressed.

maxxxxxdlp:
Feature request: either hide the add/remove buttons when view is collapsed, OR better: expand the subview when add/remove/slider is pressed.

specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
```diff
@@ -82,7 +86,12 @@ export function FormTable<SCHEMA extends AnySchema>({
   readonly onClose: () => void;
   readonly sortField: SubViewSortField | undefined;
   readonly onFetchMore: (() => Promise<void>) | undefined;
+  readonly isCollapsed: boolean | undefined;
+  readonly parentResource?: SpecifyResource<AnySchema> | undefined;
 }): JSX.Element {
+  const [isCollapsed, _, handleCollapsed, handleToggle] =
```
maxxxxxdlp, line 92: I would recommend not duplicating this state here and in IntegratedRecordSelector - duplicating state leads to bugs.
Instead, just have this component receive isCollapsed and handleToggle callbacks
Same for record selector

specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
```diff
@@ -189,6 +198,31 @@ export function FormTable<SCHEMA extends AnySchema>({
     scrollerRef
   );
 
+  const blockers = useAllSaveBlockers(parentResource, relationship);
```
maxxxxxdlp, line 201: This code shouldn't be here as you already have it in IntegratedRecordSelector, which covers both form table and record selector

maxxxxxdlp:
> ... but since there is no re-render the button stays disabled until reload

that seems like a bug (either with the hook or with your code). That shouldn't be happening as `useAllSaveBlockers` is supposed to trigger re-render on blockers change.

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -171,6 +176,52 @@ export function IntegratedRecordSelector({
     relationship.relatedTable.name
   );
 
+  const [isCollapsed, _, handleCollapsed, handleToggle] =
```
maxxxxxdlp, line 179: `handleCollapsed` is a confusing name. Shouldn't it be `handleExpand` instead?

maxxxxxdlp:
I fixed all of the remaining issues and fixed the suggested changed and opened a separate PR - https://github.com/specify/specify7/pull/3680

My fix didn't use setTimeout and didn't use useBlockerHandler.

I opened a separate PR so that you have a choice of fixing all the issues on your own on this branch and only afterward looking at that PR for reference/suggestions, OR continuing with that PR instead of this one.

If you want to carry over just that single commit from that branch to this, you could type `git cherry-pick 1417a8c`

In either case, the code should still undergo testing as I did only basic testing

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -176,52 +176,32 @@ export function IntegratedRecordSelector({
     relationship.relatedTable.name
   );
 
-  const [isCollapsed, _, handleCollapsed, handleToggle] =
+  const [isCollapsed, _handleCollapsed, handleExpand, handleToggle] =
```
maxxxxxdlp, line 179: FYI in case you were wondering why I used `_handleCollapsed` instead of `_` here.

There is no difference in behavior.

There is a very common convention in JavaScript that any variable that begins with `_` means unused - even if it's just `_` or `__`. Thus, the only reason to do something like `_handleCollapsed` instead of `_` is that the longer format is more explicit about what the unused variable is for - this makes it less guesswork for anyone reading the code, and if that unused variable will become used at a later point, naming it like this makes it easier to find

Not a big deal and most of the time I still go with `_` for simplicity, but just wanted you to know about the reasons behind it

maxxxxxdlp (approved):
Looks good
Hope the code makes sense - it's pretty much what you had, just a bit cleaned up

CarolineDenis:
> Looks good
> Hope the code makes sense - it's pretty much what you had, just a bit cleaned up

yes makes a lot of sense. Thank you for the push

grantfitzsimmons (changes requested):
The collapse behavior is inverted for the Collecting Event subview.

```xml
		<row>
			<cell type="subview" viewname="CollectingEventSub" id="21" label="Collecting Event" name="collectingEvent" colspan="13" rows="5"/>
		</row>
```

https://herbrbge71423-issue-1995.test.specifysystems.org/specify/view/collectionobject/349648/?recordsetid=178

https://github.com/specify/specify7/assets/37256050/160a53a4-8aaf-445c-96ea-bc3ec1ed5738

It is collapsed by default, though in the [form definition](https://github.com/specify/specify7/wiki/Form-System#subview) the `collapse` attribute (which would be inside of the `initialize` attribute) is not present.

grantfitzsimmons (commented):
I was so wrong... it was working correctly, the form definition tripped me up. 

Works well with the `initialize="collapse=true"` and `initialize="collapse=false"` attributes in the XML as well.

grantfitzsimmons (approved):

PR #3644, opened 2023-06-20
Allow to export a query from embeded query builder

### LINKED ISSUE #3616: Nested Query Builder Can Not Create All Items CSV
**Describe the bug**
If you open the Query Builder from let's say Data Entry and try to create a CSV from the query you just built. It will not export. I also tested this on the new Selected Query Branch and they work fine.

**To Reproduce**
1. Go to Data Entry -> Taxon -> Parent of Taxon
2. Click on the search on the side
3. Enter Query Builder
4. Run something and click on `Create CSV`
5. See Error

**Expected behavior**
It is exported without an issue.

**Beautiful Screen Recording**

https://github.com/specify/specify7/assets/53784701/b0491cf7-611a-4f88-b1c4-4a6f6db889ff

CarolineDenis:
Fixes #3616

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
@@ -145,7 +145,7 @@ export function QueryBuilder({
 
   const isResourceModified = useIsModified(queryResource);
   const saveRequired = isResourceModified || state.saveRequired;
-  const promptToSave = !isEmbedded || saveRequired;
+  const promptToSave = saveRequired && !isEmbedded;
 
   const unsetUnloadProtect = useUnloadProtect(
     promptToSave,
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
@@ -144,7 +144,9 @@ export function QueryBuilder({
   );
 
   const isResourceModified = useIsModified(queryResource);
-  const saveRequired = isResourceModified || state.saveRequired;
+  const saveRequired =
+    isResourceModified || (state.saveRequired && !isEmbedded);
+  const saveRequiredForEmbeded = isResourceModified || state.saveRequired;
 
   const unsetUnloadProtect = useUnloadProtect(
     saveRequired,
@@ -180,7 +182,7 @@ export function QueryBuilder({
    * That function does not need to be called most of the time if query
    * fields haven't changed yet. This avoids triggering needless save blocker
    */
-  const getQueryFieldRecords = saveRequired
+  const getQueryFieldRecords = saveRequiredForEmbeded
     ? (
         fields: typeof state.fields = state.fields
       ): ReturnType<typeof unParseQueryFields> =>
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
@@ -144,8 +144,7 @@ export function QueryBuilder({
   );
 
   const isResourceModified = useIsModified(queryResource);
-  const saveRequired =
-    (isResourceModified || state.saveRequired) && !isEmbedded;
+  const saveRequired = isResourceModified || state.saveRequired;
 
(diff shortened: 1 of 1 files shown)


maxxxxxdlp (changes requested):
Good job debugging this and finding a way to fix this!

For the best solution see https://github.com/specify/specify7/issues/3616#issuecomment-1601985002

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -145,7 +145,8 @@ export function QueryBuilder({
 
   const isResourceModified = useIsModified(queryResource);
   const saveRequired =
-    (isResourceModified || state.saveRequired) && !isEmbedded;
+    isResourceModified || (state.saveRequired && !isEmbedded);
```
maxxxxxdlp, line 148: A better variable naming + reuses the variable instead of stating logic twice:

```tsx
   const saveRequired =
      isResourceModified || state.saveRequired;
    const promptToSave = saveRequired && !isEmbeeded;
```

And then use `promptToSave` for useUnloadProtect, and use `saveRequired` everywhere else

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
CarolineDenis: isn't prompToSave everywhere else to keep the same behavior as before?

maxxxxxdlp: The only other usage of `saveRequired` is in QueryHeader to determine if the save button should be enabled or not

But QueryHeader is not rendered when isEmbedded anyway so it doesn't make a difference.

And in the future, once we do show enable QueryHeader for embedded queries, using saveRequired instead of promtToSave will result in behavior were user won't be required to save, but will be able to save if they want to - desirable behavior.

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -144,12 +144,11 @@ export function QueryBuilder({
   );
 
   const isResourceModified = useIsModified(queryResource);
-  const saveRequired =
-    isResourceModified || (state.saveRequired && !isEmbedded);
-  const saveRequiredForEmbeded = isResourceModified || state.saveRequired;
+  const saveRequired = isResourceModified || state.saveRequired;
+  const promptToSave = !isEmbedded || saveRequired;
```
maxxxxxdlp, line 148: careful, this should be an `&&`, not `||`.

I would also reorder the two conditions to make it more logical (you would prompt the user if "save is required as long as they are not in an embedded query" rather than if "the user is not in an embedded query, or and so happens they also need to save changed")

Thus:
```
const promptToSave = saveRequired && !isEmbedded;
```

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):
Looks good.
Needs testing with embeeded and standalone query builder:
- Test exporting modified query. Test exporting unmodified query
- Test that there is unload protect in standalone query builder, but not in embedded query builder

bronwyncombs (changes requested):
Exporting from embedded query does work as expected. 

Query builder with CO base table is crashing when (formatted) is added. Other fields on the CO table are working and (formatted) works on other tables. Seems to be specific to CO and the (formatted) field. 

Database: herb_rbge_3_13_23 
Browser: Edge
[Specify 7 Crash Report - 2023-07-03T18_55_11.854Z (1).txt](https://github.com/specify/specify7/files/11940384/Specify.7.Crash.Report.-.2023-07-03T18_55_11.854Z.1.txt)

CarolineDenis:
@bronwyncombs for your comment "Query builder with CO base table is crashing when (formatted) is added." 
This issue is not related to the changes made in this PR. If the export in embedded works, you can approve this PR. Could I also ask you to open a new issue for the CO with formatted if it doesn't exist yet?

bronwyncombs (approved):
Sorry about that! It can also be replicated on `edge`. @melton-jason is looking into this and I will create the issue now

melton-jason:
The aforementioned, associated issue is #3721

PR #3647, opened 2023-06-21
Allow to collapse side menu in App resource 

### LINKED ISSUE #3646: Add collapsible button for side menu in App resources
TODO3

CarolineDenis:
Fixes #3646

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Aside.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Aside.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { useParams } from 'react-router-dom';
 import type { LocalizedString } from 'typesafe-i18n';
+import { useBooleanState } from '../../hooks/useBooleanState';
 
 import { useErrorContext } from '../../hooks/useErrorContext';
 import { useId } from '../../hooks/useId';
@@ -67,6 +68,8 @@ export function AppResourcesAside({
   );
   useOpenCurrent(conformations, setConformations, resourcesTree);
 
+  const [isCollapsed, _, __, handleToggle] = useBooleanState(false);
+
   return (
     <aside
       className={`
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (changes requested):
image

The collapse button should be visible no matter where you are looking in a list of resources.

Additionally, do we need to display an entire column when the sidebar is collapsed?
image

See the difference between the live app resources display and the `issue-3646` version:

- The arrow is not right-aligned
- The sidebar now occupies the entire screen when it is narrow

image

Also, there is now an odd border on the right and bottom sides. See left is `issue-3646`, right is `v7.8.11`.

image

grantfitzsimmons:
I think the simplest and most elegant solution would be to just recommend the use of the full-screen button if you want to edit with fewer distractions. 

Once you hide the side menu of resources, is it functionally any different than just going full-screen?

CarolineDenis:
Recommend using full screen instead

grantfitzsimmons:
> Recommend using full screen instead

sorry 😞 I like the idea but I couldn't think of any reason it beats full-screen! I kind of wish full-screen had a sidebar in it and it was collapsible in that view.

maxxxxxdlp:
I think there are use cases when collapsing side bar is better than full screen (i.e, we could make it remember the collapsed state, or even auto collapse as soon as you open a file - can't do the same for full screen). But, it's not important enough in comparison to all the other things we are working on. It's more like a nice to have

> I kind of wish full-screen had a sidebar in it and it was collapsible in that view.

The full screen with sidebar is not much different than enabling full screen in your browser

grantfitzsimmons:
@maxxxxxdlp 

>The full screen with sidebar is not much different than enabling full screen in your browser

All of this red is wasted space (as far as I'm concerned when editing app resouces)

image

but the one thing I do often need to see is the sidebar with my position/which resource I'm editing at what level. By collapsing the sidebar, I lose the most important information while still being shown a large amount of unnecessary content on the top and left. 

In the scenario when you collapse the sidebar, you still have all of the red I have above, except for an extra 2 side menus of horizontal area. This kind of space-saving layout would be great:

image

The above environment is the real nice to have for me :heart:


---

For reference:

**Full screen:**
image

**Collapsed:**
image

**Expanded:**
image

maxxxxxdlp:
Point taken.

---

And you also switched to VS Code? This is an epidemic. I might be forced to switch too as in my team of 40 at Esri, of those that use any kind of editor, only 2 don't use vs code (one uses emacs, and the other one is me with PyCharm)

PR #3659, opened 2023-06-22
Allow to add preparations in loan form 

### LINKED ISSUE #2135: Adding loan preparations doesn't work

CarolineDenis:
Fixes #2135


specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromCollection.tsx
```diff
@@ -0,0 +1,128 @@
...
+    relatedResource: isDependent ? collection.related : undefined,
+    totalCount: collection._totalCount ?? records.length,
+    onAdd: (rawResources): void => {
+      const resources = isToOne ? rawResources.slice(0, 1) : rawResources;
+      if (isDependent && isToOne)
+        collection.related?.placeInSameHierarchy(resources[0]);
+      if (!isInteraction) collection.add(resources);
+      handleAdd?.(resources);
+      setIndex(
+        !isInteraction ? collection.models.length - 1 : collection.models.length
```
maxxxxxdlp, line 104: Try to avoid negated conditions. Remember that you can just swap the branches to resolve this.

Note, avoiding negated condition does not mean avoiding negation altogether. this only applies when you have two branches that can be easily swapped to remove negation.

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromCollection.tsx
maxxxxxdlp, line 1: FYI for the future: I am guilty of this too, but if you move something to a different file or rename a file, try not to change that file in the same commit, because that makes it hard to see if you simply moved code, or also modified it in the process.

(in this case I used https://www.diffchecker.com/text-compare/ too what's the actual changes).

So in a perfect world, moving the component would have been one commit, and then editing the component to add isInteraction prop would have been a separate commit

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromCollection.tsx
maxxxxxdlp, line 104: Line 104 looks like a bug.
If you call collection.add(), then collection.models.length should increase. If you don't call it, length shouldn't increase - thus, you should be able to just use `collection.models.length - 1` in both cases, no?

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -173,48 +58,61 @@ export function IntegratedRecordSelector({
...
+          onAdd: handleAdd,
+          onRemove: handleRemove,
+          isLoading,
+        }): JSX.Element => (
+          <>
+            {isInteraction &&
+            typeof collection.related === 'object' &&
+            isDialogOpen ? (
+              <InteractionDialog
+                actionTable={collection.related?.specifyTable}
```
maxxxxxdlp, line 106: You are checking for `typeof collection.related === 'object'` 3 lines above, so no need for `?` here

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -173,48 +58,61 @@ export function IntegratedRecordSelector({
...
-            typeof urlParameter === 'string'
-              ? setIndex(index.toString())
-              : undefined
+      <RecordSelectorFromCollection
+        collection={collection}
+        defaultIndex={isToOne ? 0 : index}
+        relationship={relationship}
+        isInteraction={isInteraction}
+        onAdd={(resource) => {
+          handleOpenDialog();
```
maxxxxxdlp, line 80: It would be better if you only call this function if `isInteraction`.
Two reasons:
- By doing the check, you are also documenting to anyone reading this code that the interaction dialog should only be open if in the interaction subview.
- While it won't make a difference right now, it might if code is refactored incorrectly - being extra cautious here is not difficult

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -1,143 +1,28 @@
...
 import { hasTablePermission } from '../Permissions/helpers';
 import { relationshipIsToMany } from '../WbPlanView/mappingHelpers';
-import type {
-  RecordSelectorProps,
-  RecordSelectorState,
-} from './RecordSelector';
-import { useRecordSelector } from './RecordSelector';
+import { RecordSelectorFromCollection } from './RecordSelectorFromCollection';
 
 // REFACTOR: encapsulate common logic from FormTableCollection and this component
```
maxxxxxdlp, line 24: Remove this REFACTOR statement as you now resolved it

maxxxxxdlp (changes requested):
All good, just small thing

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromCollection.tsx
CarolineDenis: if we keep only `collection.models.length - 1` the right side of the slider will increase but the record display will be the first one and not second one (after clicking on the + you will have 1/2 instead of 2/2). We need this to move the display to the next record. 
This is due to the fact that we do not do `collection.add(resources)`  if in 'Interactions'.

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
maxxxxxdlp: it looks like you applied the fix to a wrong function - I meant you should only call handleOpenDialog() if isInteraction.

though while you are here, the handleAdd call could be simplified: just do `handleAdd?.(resource)l` instead of the `if(...)` check

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromCollection.tsx
maxxxxxdlp: > if we keep only collection.models.length - 1 the right side of the slider will increase but the record display will be the first one and not second one (after clicking on the + you will have 1/2 instead of 2/2). We need this to move the display to the next record.

The thing is that until the user selects items in the interaction dialog, those records are not yet part of the subview, and so you can not setIndex to their future position - the code is not designed to handle that case and is either already broken in some ways in case of that or will break in the future - I don't feel safe doing that.

For example, see already what happens if user closes the interactions dialog:
An infinite spinner is not a proper behavior for this feature:
https://github.com/specify/specify7/assets/40512816/e07f7334-4c77-4b04-847b-2f349df9f070

Thus you either should find a way to set index only after the records are added to collection (i.e, probably add a prop to interaction dialog for that) or live with the issue where it goes to the index one before the correct one

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromCollection.tsx
```diff
@@ -0,0 +1,128 @@
...
+    onAdd: (rawResources): void => {
+      const resources = isToOne ? rawResources.slice(0, 1) : rawResources;
+      if (isDependent && isToOne)
+        collection.related?.placeInSameHierarchy(resources[0]);
+      if (!isInteraction) collection.add(resources);
+      handleAdd?.(resources);
+      setIndex(
+        isInteraction ? collection.models.length : collection.models.length - 1
+      );
+      handleSlide?.(collection.models.length - 1, false);
```
maxxxxxdlp, line 106: Careful! handleSlide is being called with "-1" here if isIntearction

I would change this code to this:
```ts
      const lastIndex = Math.max(0, collection.models.length - 1);
      setIndex(lastIndex);
      handleSlide?.(lastIndex, false);
```

That fixes all issue EXCEPT for the fact that the subview will show record one index before the one that will be added by interaction dialog - you can fix it by adding a prop to the prep select dialog

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):
Looks good.
The index -1 issue is something you could fix here or open a separate issue for.
Also, could compare to how it worked before - if we are not making it worse than before, then it's fine to leave it for the future

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -173,48 +57,61 @@ export function IntegratedRecordSelector({
...
+            typeof collection.related === 'object' &&
+            isDialogOpen ? (
+              <InteractionDialog
+                actionTable={collection.related.specifyTable}
+                itemCollection={
+                  collection as Collection<
+                    DisposalPreparation | GiftPreparation | LoanPreparation
+                  >
+                }
+                onClose={handleCloseDialog}
```
maxxxxxdlp, line 109: Regarding `you can fix it by adding a prop to the prep select dialog`:

When you display the InteractionDialog, two things could happen:
1. Either the user closes the dialog, without selecting anything - in this case only `handleCloseDialog` is called
2. OR, the user selects new props and then the dialog closes - in this case new records are added to `collection` and `handleCloseDialog` is closed.

What you need to do is when the dialog is closed, figure out what is the index of the first new added record (i.e collection had 3 records (thus last index is `2`), and user added 5 more, then the index of first added is `3`). And then you would open the record with that index in the subview

Some possible ways to do that
1. Add a callback i.e `onAdded={}` that would be called by InteractionDialog when it adds new records. the callback could be called right before new records are added (thus you could use `collection.models.length` as new index) or right after (in which case InteractionDialog will pass the new records, and you could do `collection.models.length - addedRecords.length` to get new index)
2. OR record the number of records in the collection before the interaction dialog is opened in a ref, and once the interaction dialog is closed, check if number of records changes and get the new index
3. OR, set an event listener for new items being added, and change the index when that happens - though this one would be tricky as if 4 records are added all at once, the event index would be called 4 times, thus, subview will end up displaying the last record

CarolineDenis: new issue created :+1:

bronwyncombs (approved):
Works! Added preparations to loan from record set and by entering catalog numbers. Added from subview and grid view. 

https://github.com/specify/specify7/assets/135047322/a0358c39-4641-4aeb-ae01-1a7f7898fec9

PR #3667, opened 2023-06-23
Properly set cellMeta of cells in WorkBench after validation

### LINKED ISSUE #3635: WorkBench errors not reported after validation cannot be navigated between
WorkBench errors not showing in the interface

After validating, I see no indication that there were errors and have to find them myself manually. See that there is no error in the navigation options at the bottom. 

I am using `v7.8.12` without Internet on a brand-new Paleo database.

melton-jason:
Fixes #3635 

The issue stemmed from the fact that `throttle` was now calling the `updateCellInfoStats` method, which re-bound its context and set it to undefined rather than WBView. 
Preferably, I would like to set the context of the function without passing it as a parameter (although this does allow for greater customization). 
I would love to hear other developers thoughts on this. 

To Test:
- Ensure the functionality of the Cell Navigation Buttons (Search Results, Modified Cells, New Cells, Error Cells) remain constant through the Workbench operations do not exhibit any unusual behavior
- Lightly ensure that the rendering behavior of the WorkBench when resizing your browser window remains the same as it did before Specify v7.8.12


grantfitzsimmons (approved):
Works as expected. So happy to see this working!

specifyweb/frontend/js_src/lib/utils/utils.ts
```diff
@@ -403,7 +404,7 @@ export function throttle<ARGUMENTS extends RA<unknown>>(
   function later(): void {
     previousTimestamp = Date.now();
     timeout = undefined;
-    callback(...previousArguments!);
+    callback.bind(thisArgument)(...previousArguments!);
```
maxxxxxdlp, line 407: See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply and https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/call

maxxxxxdlp (approved):
This bug was likely caused by me replacing underscore.js's throttle with our implementation of throttle - I didn't consider that some places might be relying on `this`

Though, you should might be able to just do this:
```ts
this.updateCellInfoStats = throttle(this.updateCellInfoStats.bind(this), throttleRate);
```
And no other changes

Still, not a big deal, and the bug is urgent, so let's merge

specifyweb/frontend/js_src/lib/utils/utils.ts
melton-jason: > this.updateCellInfoStats = throttle(this.updateCellInfoStats.bind(this), throttleRate);
I was going to do this, but I figured that writing the bind in the throttle function would prevent future bugs like this.

melton-jason: I would have preferred to use apply rather than bind, but Typescript did not like me for it. 
Typescript expects the arguments parameter in `function.apply(args)` to always be something like `RA<unkown>`, while the function in this case expects its arguments to match the type signatures of the definition (in this case the ARGUMENTS type). 

Even though the ARGUMENTS type definition is `extends RA<unknown>`, I was not able to make Typescript agree on a proper type

PR #3668, opened 2023-06-23
Focus first field when grid mode

### LINKED ISSUE #3630: Set focus on first field when subview is a grid.
See https://github.com/specify/specify7/pull/3317

CarolineDenis:
Fixes #3630

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
@@ -23,7 +23,7 @@ import type { FormMode } from '../FormParse';
 import type { FormCellDefinition, SubViewSortField } from '../FormParse/cells';
 import { attachmentView } from '../FormParse/webOnlyViews';
 import { SearchDialog } from '../Forms/SearchDialog';
-import { SpecifyForm } from '../Forms/SpecifyForm';
+import { SpecifyForm, useFirstFocus } from '../Forms/SpecifyForm';
 import { useViewDefinition } from '../Forms/useViewDefinition';
 import { loadingGif } from '../Molecules';
 import { Dialog } from '../Molecules/Dialog';
@@ -137,7 +137,11 @@ export function FormTable<SCHEMA extends AnySchema>({
     const lastRow: HTMLElement | null = rowsRef.current.querySelector(
       `:scope > :nth-child(${resourceIndex + 1}) > [tabindex="-1"]`
     );
-    lastRow?.focus();
+    const firstFocusableElement = rowsRef.current.querySelector<HTMLElement>(
+      'a, input:not([type="hidden"]), select, textarea, [tabindex]:not([tabindex="-1"])'
+    )!;
+    firstFocusableElement?.focus();
+    // lastRow?.focus();
   }, [resources]);
 
   const isToOne = !relationshipIsToMany(relationship);
--- a/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
@@ -141,7 +141,7 @@ export function FormTable<SCHEMA extends AnySchema>({
     lastRowRef.current = lastRow;
     const firstFocusableElement = findFirstFocusableElement(lastRowRef);
     firstFocusableElement?.focus();
-    // lastRow?.focus();
+    // LastRow?.focus();
   }, [resources]);
 
   const isToOne = !relationshipIsToMany(relationship);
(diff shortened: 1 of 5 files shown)


maxxxxxdlp (changes requested):
As per discussion earlier today, the `resourceIndex+1` fix is the only change required to make this work properly (unless testers find issues with this)

All other changes should be reverted

maxxxxxdlp (approved):

bronwyncombs (changes requested):
Unfortunately, this doesn't work. It doesn't focus the cursor on the first available field after changing from subform to grid view. 'Tab' is still functional, but it is not an automatic focus. 

**Screen Recording**

https://github.com/specify/specify7/assets/135047322/b6c7f0e7-b1ea-43ef-847a-e3c8a3d571c5

maxxxxxdlp:
Note, this is only expected to move focus when you add a record (click on `+`), not when you collapse/expand (though if we want that that could be added)

@CarolineDenis the other small issue is that the useEffect hook present in the FormTable component focuses the form container, not the first form field. You developed a function for finding the first focusable element - could you use that function here?

CarolineDenis:
> Note, this is only expected to move focus when you add a record (click on `+`), not when you collapse/expand (though if we want that that could be added)
> 
> @CarolineDenis the other small issue is that the useEffect hook present in the FormTable component focuses the form container, not the first form field. You developed a function for finding the first focusable element - could you use that function here?

With the last code I pushed (which doesn't work) 
I have a line     
`const lastRow: HTMLElement | null = rowsRef.current.querySelector(
      `:scope > :nth-child(${resourceIndex + 1}) > [tabindex="-1"]`
    );` that returns the last row in the grid 
And I have 
   `const firstFocusableElement = rowsRef.current.querySelector<HTMLElement>(
      'a, input:not([type="hidden"]), select, textarea, [tabindex]:not([tabindex="-1"])'
    )!;`
    That focuses the first input or select etc. 
    
    I would need to find a way to return the first input in the last line. 
    PS: doing `lastRow.querySelector('input', 'select',...)` always returns null

specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
```diff
@@ -135,9 +135,13 @@ export function FormTable<SCHEMA extends AnySchema>({
     addedResource.current = undefined;
     if (resourceIndex === -1 || rowsRef.current === null) return;
     const lastRow: HTMLElement | null = rowsRef.current.querySelector(
-      `:scope > :nth-child(${resourceIndex}) > [tabindex="-1"]`
+      `:scope > :nth-child(${resourceIndex + 1}) > [tabindex="-1"]`
     );
-    lastRow?.focus();
+    const firstFocusableElement = rowsRef.current.querySelector<HTMLElement>(
```
maxxxxxdlp, line 140: why duplicate this complex code here?

please never copy complicated code - especially when it would be so easy to extract it into a function and reuse that cunction

maxxxxxdlp: you need to look for first focusable in `lastRow`, not in `rowsRef.current` - that will solve your other problem

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
CarolineDenis: it was to try the code for now I will extract it to a function once it works. Finding the first input or text area etc on lastRow doesn't work for a reason I didn't figure out yet.

specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
```diff
@@ -137,9 +138,8 @@ export function FormTable<SCHEMA extends AnySchema>({
     const lastRow: HTMLElement | null = rowsRef.current.querySelector(
       `:scope > :nth-child(${resourceIndex + 1}) > [tabindex="-1"]`
     );
-    const firstFocusableElement = rowsRef.current.querySelector<HTMLElement>(
-      'a, input:not([type="hidden"]), select, textarea, [tabindex]:not([tabindex="-1"])'
-    )!;
+    lastRowRef.current = lastRow;
+    const firstFocusableElement = findFirstFocusableElement(lastRowRef);
```
maxxxxxdlp, line 142: (optional) lastRowRef shouldn't be needed
you could modify findFirstFocusableElement to accept `HTMLElement | null`
or pass `{current: lastRow}` as an argument

maxxxxxdlp: The logic looks good. Does this code still not work? Do you know why? if not, I can take a look at it

maxxxxxdlp (approved):

bronwyncombs (changes requested):
Still not working 

https://github.com/specify/specify7/assets/135047322/d49307bf-0461-45df-a167-2e8f6b0928d1

CarolineDenis:
> Still not working
> 
>  Screen.Recording.2023-07-06.at.1.35.50.PM.mov

Sorry I should have put this PR in draft. It's not finished yet.

PR #3669, opened 2023-06-23
Add the ability to hide plus button in forms

### LINKED ISSUE #3657: Add the ability to hide QCBX buttons
>Is it possible to selectively remove the ‘+’ function to add a new agent/taxon/geography etc from the form. This is for users that have full access otherwise. We found that users tend to click on the ‘+’ symbol when they want to select something from a QCBX, rather than use the type-ahead or search functionality. Of course, this partly a training exercise, but we were wondering if we can just remove that option from the form, so when they actually want to add something new they have to go via the tree or the relevant Data Entry form.
>
>image

This request is essentially asking for another attribute so that the `+` button could be hidden on the form.

Per the [form system documentation](https://github.com/specify/specify7/wiki/Form-System#querycbx), there is already an attribute named `cloneBtn` that can be set to true or false. Adding an attribute named `addBtn` that had a default of true and could be made false would solve this issue.

> - `cloneBtn` ('true' | 'false', optional, default: 'false')

**Requested By:** CSIRO on [Asana](https://app.asana.com/0/1201972579963460/1204830887216582/f)

CarolineDenis:
Fixes #3657

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
+++ b/specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
@@ -800,11 +800,6 @@ const queryComboBoxSpec = f.store(() =>
       syncers.maybe(syncers.toBoolean),
       syncers.default<boolean>(false)
     ),
-    showAddButton: pipe(
-      syncers.xmlAttribute('initialize addBtn', 'skip'),
-      syncers.maybe(syncers.toBoolean),
-      syncers.default<boolean>(true)
-    ),
     showEditButton: pipe(
       syncers.xmlAttribute('initialize editBtn', 'skip'),
       syncers.maybe(syncers.toBoolean),
--- a/specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
+++ b/specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
@@ -800,6 +800,11 @@ const queryComboBoxSpec = f.store(() =>
       syncers.maybe(syncers.toBoolean),
       syncers.default<boolean>(false)
     ),
+    showAddButton: pipe(
+      syncers.xmlAttribute('initialize addBtn', 'skip'),
+      syncers.maybe(syncers.toBoolean),
+      syncers.default<boolean>(true)
+    ),
     showEditButton: pipe(
       syncers.xmlAttribute('initialize editBtn', 'skip'),
       syncers.maybe(syncers.toBoolean),
--- a/specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
+++ b/specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
@@ -54,6 +54,7 @@ export type FieldTypes = {
     'QueryComboBox',
     {
       readonly hasCloneButton: boolean;
+      readonly hasAddButton: boolean;
       readonly typeSearch: string | undefined;
       readonly searchView: string | undefined;
     }
@@ -212,6 +213,7 @@ const processFieldType: {
         hasCloneButton: getProperty('cloneBtn')?.toLowerCase() === 'true',
         typeSearch: getProperty('name'),
         searchView: getProperty('searchView'),
+        hasAddButton: getProperty('addBtn')?.toLowerCase() === 'true' ?? true,
       };
     } else {
       console.error('QueryComboBox can only be used to display a relationship');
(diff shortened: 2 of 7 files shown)


specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
```diff
@@ -212,6 +213,8 @@ const processFieldType: {
         hasCloneButton: getProperty('cloneBtn')?.toLowerCase() === 'true',
         typeSearch: getProperty('name'),
         searchView: getProperty('searchView'),
+        hasAddButton:
+          (getProperty('addBtn')?.toLowerCase() ?? 'true') === 'true',
```
maxxxxxdlp, line 217: Can be simplified:
```ts
getProperty('addBtn')?.toLowerCase() !== 'false'
```

Plus, if user makes a typo and says `true ` or `truee`, your version would default to `false`, where as this version would default to `true`

specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
```diff
@@ -800,6 +800,11 @@ const queryComboBoxSpec = f.store(() =>
       syncers.maybe(syncers.toBoolean),
       syncers.default<boolean>(false)
     ),
+    showAddButton: pipe(
+      syncers.xmlAttribute('initialize addBtn', 'skip'),
+      syncers.maybe(syncers.toBoolean),
+      syncers.default<boolean>(true)
+    ),
```
maxxxxxdlp, line 807: On line 813 in this file you will see `showNewButton`
Sp6 already supported hiding the add button. Please refactor this PR to use that instead of adding new property

maxxxxxdlp (changes requested):
Logic is correct, except sp6 already has a property for this, so no need to add a new one.

(optional) Also, while you are at this, feel free to add support for other props too - https://github.com/specify/specify7/issues/3657#issuecomment-1605253142

Finally, don't forget to update the forms Wiki page

specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
```diff
@@ -213,8 +215,10 @@ const processFieldType: {
         hasCloneButton: getProperty('cloneBtn')?.toLowerCase() === 'true',
         typeSearch: getProperty('name'),
         searchView: getProperty('searchView'),
-        hasAddButton:
-          (getProperty('addBtn')?.toLowerCase() ?? 'true') === 'true',
+        hasNewButton: getProperty('showNewButton')?.toLowerCase() !== 'false',
+        hasSearchButton:
+          getProperty('showSearchButton')?.toLowerCase() !== 'false',
+        hasEditButton: getProperty('showEditButton')?.toLowerCase() !== 'false',
```
maxxxxxdlp, line 221: getProperty is being called with wrong arguments.
Should be `editBtn` `newBtn` `cloneBtn` `searchBtn` - carefully look at the spec again - you used the internal key names, not what is being read by the spec from the XML

```xml
   // showSearchButton is just an internal name - you can pick whatever you want
    showSearchButton: pipe(
   // searchBtn is what's actually read from the XML form from the "initialize" attribute
      syncers.xmlAttribute('initialize searchBtn', 'skip'),
      syncers.maybe(syncers.toBoolean),
      syncers.default<boolean>(true)
    ),
```

maxxxxxdlp (changes requested):
One big issue. The rest is good

maxxxxxdlp (approved):
Looks good.
(optional) it would be a good idea to also update the form parsing tests for query combo box to include these new attributes - that is not hard to do, but will prevent future bugs

bronwyncombs (approved):
Works as expected!  Edited Determination form for the 'Determiner' QCB. 
msedge_eZzmGG0kXc
msedge_7bjerLXwmY

CarolineDenis:
@maxxxxxdlp can you remind me why we wanted this one to be merged with xml and not prod?

maxxxxxdlp:
In general, avoid merging things to production unless it's urgent - every commit you merge is going to be more work for Alec or Jason later on due to merge conflicts (and of course, when resoling merge conflicts, bugs can be introduced)

But also, XML editing did changes to forms and form editing, so this was a better branch to do that change on.

If it is urgent to have this in production, feel free to cherry-pick the commits

specifysoftware:

https://discourse.specifysoftware.org/t/specify-7-9-4-release-announcement/1553/6

PR #3675, opened 2023-06-24
not allow to select several times a cell in WB

### LINKED ISSUE #3174: WorkBench allows selecting a row multiple times
https://user-images.githubusercontent.com/37256050/225083552-75b44762-5edb-4968-a5b6-359dc1d8be45.mov

Command + click any item and see that you can select it multiple times!

It should deselect when you click on it.

CarolineDenis:
Fixes #3174


CarolineDenis:
TODO: instead of adding newSelection to the array it needs to replace it

specifyweb/frontend/js_src/lib/components/WorkBench/hooks.ts
```diff
@@ -568,6 +568,16 @@ export function getHotHooks(wbView: WbView) {
...
     },
+    afterSelection: () => {
+      if (wbView.hot === undefined) return;
+      console.log(wbView.hot.getSelected());
+      const selection = wbView.hot?.getSelected() ?? [];
+      const newSelection = f
+        .unique(selection.map((row) => JSON.stringify(row)))
+        .map((row) => JSON.parse(row));
+      if (newSelection.length !== selection.length)
+        wbView.hot?.selectCells(newSelection);
```
maxxxxxdlp, line 579: in response to:

> TODO: instead of adding newSelection to the array it needs to replace it?

You need to double check if the call to selectCells on line 579 replaces all selected cells with the newSelection, OR adds newSelection to already selected cells.

I think it would replace the existing selection, but you have to check.
If it adds rather than replaces, then you would have to look though Handsontable's documentation for a way to deselect cells

specifyweb/frontend/js_src/lib/components/WorkBench/hooks.ts
```diff
@@ -568,6 +568,16 @@ export function getHotHooks(wbView: WbView) {
           10
         );
     },
+    afterSelection: () => {
```
maxxxxxdlp, line 571: Could you add one blank line before `afterSelection` to be consistent with other functions in this file?
Also, like you may see for other hooks in this file, you could please add a small comment explaining why this hook is needed?

maxxxxxdlp (changes requested):

CarolineDenis:
NOTES: currently works for cells but not row

specifyweb/frontend/js_src/lib/components/WorkBench/hooks.ts
```diff
@@ -568,6 +568,21 @@ export function getHotHooks(wbView: WbView) {
           10
         );
     },
+
+    /*
+     * Unallow user to select several times the same cell
```
maxxxxxdlp, line 573: `Disallow user to select`/`Forbit user from selecting` would be better than `Unallow` in this context

maxxxxxdlp (approved):

bronwyncombs (approved):
Working! Command + click now unselects the row(s) after being selected. 

https://github.com/specify/specify7/assets/135047322/b68354bd-11a7-4f98-a6eb-8ebe74c6aedf

PR #3680, opened 2023-06-25
Add collapsible feature in forms

maxxxxxdlp:
This is an alternative implementation of https://github.com/specify/specify7/pull/3642.
Only one of these two pull requests should be merged - the other should be closed

Fixes https://github.com/specify/specify7/issues/1995


CarolineDenis:
code changes have been implemented in https://github.com/specify/specify7/pull/3642.

PR #3682, opened 2023-06-26
Add additional debugging information

grantfitzsimmons:
@melton-jason This is based on the contents of your Slack message. Can you review and edit where necessary so that we can expand the information on this for the public?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,3 @@
-
 # [Specify 7](https://www.specifysoftware.org/products/specify-7/)
 
 The [Specify Collections Consortium](https://www.specifysoftware.org) is pleased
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -336,6 +335,23 @@ allow operation with the unoptimized Javascript files.
...
+**Backend:**
+Enabling `SP7_DEBUG` enables DEBUG mode for Django
+<https://docs.djangoproject.com/en/3.2/ref/settings/#debug>
+
+The Specify backend also special behavior while in DEBUG mode:
+- When the backend encounters an error, Django returns an error message containing comprehensive information of the error and potentially sensitive system information
+- Specify validates the WorkBench upload plan when fetching the WorkBench's upload results
+   - From the `/api/workbench/upload_results/(?P&lt;ds_id&gt;\d+)/` endpoint
+- Specify adds an indent to the datamodel returned from the `context/datamodel.json` endpoint
+- (DISABLED BY DEFAULT) Specify allows any endpoint/view to be profiled by adding a `/?prof` query parameter
```
melton-jason, line 347: I am not sure how thorough you wish to be, but just having it adds an indent and formats the datamodel returned from the endpoint might just be taking up space. 

Someone from @specify/dev-testing or @specify/developers will have to take a look at the profiling middleware and see if it's even functional when enabled. 
Theoretically, it just needs to be enabled by removing the comment in `specifyweb/settings/__init__.py`
https://github.com/specify/specify7/blob/52d0f1c12b294936382ef90cc24ec825ce0a6e9d/specifyweb/settings/__init__.py#L190

melton-jason (changes requested):

README.md
```diff
@@ -336,6 +335,23 @@ allow operation with the unoptimized Javascript files.
...
+
+**Frontend:**
```
maxxxxxdlp, line 349: Note, enabling DEBUG mode using SP7_DEBUG does not affect front-end at all!

Front-end debug mode is enabled separately:

https://github.com/specify/specify7/blob/bd5f5bc51fc03907936bcbc15e544afe72a36cc7/docker-compose.yml#L72

https://github.com/specify/specify7/blob/bd5f5bc51fc03907936bcbc15e544afe72a36cc7/Dockerfile#L37

(`-w` means watch for changes and rebuild front-end on changes)

See documentation at https://webpack.js.org/api/cli/

maxxxxxdlp: And unlike with python where debug mode can be enabled/disabled at will, front-end files are compiled from typescript into javascript and production/development mode must be specified during compilation.

In production, webpack is not running at all (it's only present at all when the docker image is being built by github action, and after that the build artifacts are copied into docker image, but not webpack itself as it's not needed anymore)

PR #3693, opened 2023-06-27
Don't add XML Declaration when running report

### LINKED ISSUE #3690: Reports and labels not working on `v7.8.12`
**Describe the bug**
Some (if not all) reports and labels are not working on `v7.8.12`. These work fine in `v7.8.11` and `v7.8.5`.

I could recreate across several other databases (UMich, coldfish).

**To Reproduce**
1. Run the `Mini Labels` report on catalog number `000161190` in the `Marine Invertebrates` collection.
2. See error on `v7.8.12` and see success on `v7.8.11`

**Screenshots**

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Specify 7 Version: v7.8.12

**Database Name:** `fwri_6_1_23`

## `v7.8.11`
image
https://fwri6123-v7811.test.specifysystems.org/specify/view/collectionobject/454378/?recordsetid=1798

## `v7.8.12`

image

https://fwri6123-v7812.test.specifysystems.org/specify/view/collectionobject/454378/?recordsetid=1798

[ReportException at _report_runner_run_.html.zip](https://github.com/specify/specify7/files/11875114/ReportException.at._report_runner_run_.html.zip)

melton-jason:
Fixes #3690 

The issue was caused because in https://github.com/specify/specify7/commit/d09f70f157#diff-1503b42bbfa34c58a6baefb81b117d23fe66c0b01b21d15ac9efce9151addbe0R29-R60 we try and add an xml declaration if one does not exist.

The problem is that we are calling a function (parseXML) to parse the xml from the report: https://github.com/specify/specify7/blob/6fd4d042b38e0743f15f88fd0becec537eccd8f9/specifyweb/frontend/js_src/lib/components/Reports/Report.tsx#L90-L96

parseXML uses the built-in DOMParser `parseFromString(...)` method. 
https://github.com/specify/specify7/blob/6fd4d042b38e0743f15f88fd0becec537eccd8f9/specifyweb/frontend/js_src/lib/components/AppResources/codeMirrorLinters.ts#L69-L73

A downside of using this built-in is that the XML Declaration is not included in the serialization. 
Thus, we will _always_ add an XML declaration when call `xmlToString(...)` because the report will never have a declaration. 

However, if the report had an XML declaration, it will now have _two_ declarations when sent to the report runner, [which is not allowed](https://stackoverflow.com/a/20251895) 

So reports that omitted the xml declaration ran correctly while those that included the declaration threw the error. 
While the latter is much more common the former, I've decided to lean in its favor.

- - - 

## To Test
Background: an XML Declaration should be present at the top of every XML file, and there should only ever be one. 
A Declaration looks like: `<?xml version = "1.0" encoding = "UTF-8"?>`

- Run a report which contains an XML declaration
- Run a report which _does not_ contain an XML Declaration

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Reports/Run.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Reports/Run.tsx
@@ -54,7 +54,7 @@ export function RunReport({
         type="hidden"
       />
       <input
-        defaultValue={xmlToString(definition.documentElement)}
+        defaultValue={xmlToString(definition.documentElement, false)}
         name="report"
         type="hidden"
       />


grantfitzsimmons (approved):
Works great. No more issues for me on `fwri_6_1_23` on both Mini labels with and without the declaration.

bronwyncombs (approved):
Working great for me on UOregonPaleo. Reports and labels both run correctly with and without declaration.

maxxxxxdlp:
I belive this is a non-issue on the xml-editor branch as the `xmlToString` function is much smarter and only adds a declaration if it's not already present - https://github.com/specify/specify7/blob/041259050c024debfa26bef7a0a68df01330dd6d/specifyweb/frontend/js_src/lib/components/Syncer/xmlUtils.ts#L54

maxxxxxdlp:
I believe this issue is fixed on the `xml-editor` branch as `xmlToString` is now much smarter - and along other things only adds declaration if it's not already there

https://github.com/specify/specify7/blob/041259050c024debfa26bef7a0a68df01330dd6d/specifyweb/frontend/js_src/lib/components/Syncer/xmlUtils.ts#L54

maxxxxxdlp:
So this fix would have to be reversed on the xml-editor branch or otherwise no XML declaration at all would be exposed

Thus, unless this is super urgent, I would recommend instead adding a check for existing declaration rather than not adding declaration (unless reports with no declarations work just fine)

melton-jason:
> I believe this issue is fixed on the `xml-editor` branch as `xmlToString` is now much smarter - and along other things only adds declaration if it's not already there
> 
> https://github.com/specify/specify7/blob/041259050c024debfa26bef7a0a68df01330dd6d/specifyweb/frontend/js_src/lib/components/Syncer/xmlUtils.ts#L54

How is this code different than the code already in production? 

https://github.com/specify/specify7/blob/52d0f1c12b294936382ef90cc24ec825ce0a6e9d/specifyweb/frontend/js_src/lib/components/AppResources/codeMirrorLinters.ts#L46-L51

For reports, we use `DOMParser().parseFromString(...)`which was not parsing the first XML declaration so we will _always_ attempt to add the declaration, even if one exists in the Report or Label.
If we are still using `DOMParser().parseFromString(...)`, for reports and labels, then this issue will still occur on the `xml-editor` branch.

grantfitzsimmons:
>Thus, unless this is super urgent, I would recommend instead adding a check for existing declaration rather than not adding declaration (unless reports with no declarations work just fine)

It is super urgent

maxxxxxdlp:
> How is this code different than the code already in production?

Ah I remember what happened.
I improved xmlToString on the `xml-editor` branch, but then needed that on the `specify-network-integration` branch. Then, the specify network branch got merged into production.

So this code wasn't in production for long

Issue #2349, opened 2022-10-25
Improve handling of nullable boolean fields


maxxxxxdlp:
From CSIRO:

> 1 - We would like to add an option to be able to choose what values are displayed in Specify query results for yesNo fields for the 3 possible options: NULL/0/1
2 - It would be very useful if we could configure what values are exported out into the csv file - it can be confusing for users to see numeric values when they expect a yes/no or true/false value
3 - Is it possible to set a default value for a yesNo tickbox? When viewing an unticked box on a form it is impossible to know if the database value is NULL or 0 - this can cause users to not find all the relevant data if they don't set up the query correctly. This could be fixed if we could set the default value to 0 for unticked boxes but we don't see a way to do that via the form xml? (Having three possible values for a binary result is a little counter-intuitive!)

maxpatiiuk:
From @grantfitzsimmons:

> This is a request to export a different value for yesNo fields:
>
> >Right now you get a 1 or a 0 in the csv which isn't always very helpful, but being able to pick a title value (true/false or yes/no) would be very nice as well
> 
> ## Current
> | Catalog Number | YesNo1 |
> |:--------------:|:------:|
> |   123456789    |   0    |
> |   234567890    |   1    |
> |   345678901    |   1    |
> |   456789012    |   0    |
> |   567890123    |   0    |
> |   678901234    |   1    |
> |   789012345    |   1    |
> |   890123456    |   0    |
> |   901234567    |   1    |
> 
> ## Requested
> | Catalog Number | YesNo1 |
> |:--------------:|:------:|
> |   123456789    | false  |
> |   234567890    |  true  |
> |   345678901    |  true  |
> |   456789012    | false  |
> |   567890123    | false  |
> |   678901234    |  true  |
> |   789012345    |  true  |
> |   890123456    | false  |
> |   901234567    |  true  |
> 
> **Requested By:** Dan B at CSIRO ([Asana](https://app.asana.com/0/0/1203981304440304/1204663198443053/f))

maxpatiiuk:
> 1 - We would like to add an option to be able to choose what values are displayed in Specify query results for yesNo fields for the 3 possible options: NULL/0/1
2 - It would be very useful if we could configure what values are exported out into the csv file - it can be confusing for users to see numeric values when they expect a yes/no or true/false value

While this is something we definitely should support, a short-term workaround they can use is to create a formatter for that field, and the formatter would print true or false based on the value of the field

[...]

PR #3696, opened 2023-06-27
Allow new picklists to be created 

### LINKED ISSUE #3692: Unable to create new picklist via schema config & URL
**Describe the bug**
When adding a pick list in Schema Config or by URL an error is thrown. 
It is still possible to add a pick list by going through Query > Pick List > Browse in Forms > Open in new tab > Add button

**To Reproduce**
1. Go to Schema Config
2. Select any field in any form to alter 
3. Change 'Field Format' to pick list
4. Click the '+' by QCB
5. See error

**Expected behavior**

Opens form to add and edit new pick list 

**Screenshots**
New Pick List error

**Desktop:**
 - OS: Windows 10
 - Browser: Edge
 - Specify 7 Version: 7.8.13-prerelease

**Database Name:** 
coloradopaleo

**Reported By**
Name of your institution

**Additional context**

TypeError: Cannot read properties of undefined (reading 'toLowerCase')
[Specify 7 Crash Report - 2023-06-27T14_49_33.597Z.txt](https://github.com/specify/specify7/files/11884360/Specify.7.Crash.Report.-.2023-06-27T14_49_33.597Z.txt)

melton-jason:
Fixes #3692

The issue occurs because of the following code:

https://github.com/specify/specify7/blob/52d0f1c12b294936382ef90cc24ec825ce0a6e9d/specifyweb/frontend/js_src/lib/components/FormCells/PickListEditor.tsx#L21-L33

`resource.get('tableName')` is returning undefined rather than null, and thus setting the constant of the React state `tableName` to undefined. 

The error is occurring because we are passing undefined to `getModel(tableName)` and getModel is expecting a string. 

I believe the get method in resourceApi.js is returning undefined because the picklist resource does not have an attribute of `tablename` (yet. Whichever code sets tableName has not been run when the error is thrown). 

My fix here is most likely not a long-term solution. 
@maxxxxxdlp I was hesitant to modify the get method of ResourceBase in resourceApi.js because it will have wide-reaching consequences, but since nearly every caller of resource.get() expects the value to be either string or null, a potential solution may be to return null if the internal Backbone call returns undefined. 
Of course, it _may_ be helpful to have the distinction between null and undefined: they mean two entirely different things and callers of resource.get() lose potentially helpful information (this information however, is very easily attainable from other means). 
- - - 

## To Test
- Ensure rendering/loading _all_ picklist forms does not result in an error (both new and existing picklists)
- Try changing the Table from the PickListTable combobox on the PickList form (on both new and existing picklists), and ensure it operates (saves!) correctly

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormCells/PickListEditor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/PickListEditor.tsx
@@ -30,7 +30,10 @@ export function PickListEditor({
     [resource]
   );
 
-  const table = tableName === null ? undefined : getModel(tableName);
+  const table =
+    tableName === null || tableName === undefined
+      ? undefined
+      : getModel(tableName);
 
   const collection = React.useMemo(
     () =>


maxxxxxdlp:
This is something that must be fixed in the new records API. Could you write it down?
I strongly belive that null should be abolished as a bug causing evil in favor of using `undefined` everywhere exclusively - the consistency would simplify code and reduce bugs.

Of course, the one cons of this is that when we receive data from back-end, we would have to convert nulls to undefined (because JSON uses nulls) - but that's not hard. Don't remember if converting back is necessary or not (undefined are just omited by JSON.stringify(), so it depends on whether back-end can handle missing keys)

---

The other more short term issue is that the typing for `.get()` is wrong - the return type at the moment is most often either `string` (for required fields) or `string | null` for optional fields (and number/boolean respectively instead of string)

But in reality, if you have just created a blank resource, every `.get()` call will return `undefined` - even for required fields.



This is something I have been trying to fix by creating a `addMissingFields` function, which could definitely be used by the new records API when creating new resources - it adds default values for all required fields (and optional too if you want), thus ensuring undefined is never returned.

---

> @maxxxxxdlp I was hesitant to modify the get method of ResourceBase in resourceApi.js because it will have wide-reaching consequences, but since nearly every caller of resource.get() expects the value to be either string or null, a potential solution may be to return null if the internal Backbone call returns undefined.
> Of course, it may be helpful to have the distinction between null and undefined: they mean two entirely different things and callers of resource.get() lose potentially helpful information (this information however, is very easily attainable).

My intuition says that making .get() always return null instead of undefined should be fine, but don't quote me on this - my suggestion would be to not change anything unless you have to until the migration to the new resources API - I,e bugs will be introduced during the migration, so it's better to introduce them all at once.

Yes, there is a difference between `null` and `undefined`, but I don't think front-end cares about that difference at the moment anywhere.

In fact, it's a bit worse and front-end often does not care about the difference between `false` and `null` (most notable in checkboxes and default values for checkboxes - CSIRO does not like that at all - https://github.com/specify/specify7/issues/2349)

maxxxxxdlp (approved):
An okay fix for an ungly issue - probably the best we can do at the moment but we should really set some time aside and to the resources API rewrite, or else we are slowing down development velocity on Sp7

bronwyncombs (approved):
Working as expected!

PR #3712, opened 2023-06-30
Display tables filter picklist in attachments only once total is fetched

### LINKED ISSUE #3606: Do Not Show Skeltons if Collection Has No Attachments
When a collection/institute has no attachments. You can change how the "Attachments" are sorted. This causes the skeleton loaders to flash for a moment. 

It would be great if this didn't happen. Since there's nothing to sort.
Or...
We could disable the picker if they are no attachments
Or...
You decide. ¯\_(ツ)_/¯
 
**Beautiful Pictures (Screen shoots and recordings)**

https://github.com/specify/specify7/assets/53784701/b997fcc3-76f9-45b5-bf71-22fc7def8e6c

**Reported By**
@carlosmbe from specify collections consortium

CarolineDenis:
Fixes #3606

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Attachments/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Attachments/index.tsx
@@ -212,7 +212,10 @@ function Attachments({
             )}
             <optgroup label={schemaText.tables()}>
               {tablesWithAttachments()
-                .filter(({ name }) => collectionSizes?.byTable[name] !== 0)
+                .filter(({ name }) => {
+                  if (collectionSizes?.byTable[name] === undefined) return;
+                  return collectionSizes?.byTable[name] !== 0;
+                })
                 .map(({ name, label }) => (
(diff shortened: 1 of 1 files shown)


CarolineDenis:
The issue mentioned by Carlos in https://github.com/specify/specify7/issues/3606 happened because we were displaying the pick list before knowing if the table had attachments. Therefore we were switching to a page table containing no attachments. With this implementation, only the tables with attachments are offered as options in the filter list. The down side is that fetching the attachments total amount takes time, which slower the pick list rendering.

maxxxxxdlp (changes requested):
I don't think this fix is worth the disadvantages it brings (for databases with a lot of collection object attachments, it could take a while to fetch the count)

Also, looking at the video Carlos attached - https://github.com/specify/specify7/issues/3606, it doesn't even seem that this fix addresses the issue. He has "All" selected, not any specific table.

Plus, his database doesn't have any attachments at all

PR #3715, opened 2023-07-03
Create function to export selection to kml

### LINKED ISSUE #3589: Query Builder export to KML ignores selected rows

CarolineDenis:
Fixes #3589

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
@@ -17,9 +17,9 @@ import { userPreferences } from '../Preferences/userPreferences';
 import { mappingPathIsComplete } from '../WbPlanView/helpers';
 import { generateMappingPathPreview } from '../WbPlanView/mappingPreview';
 import { QueryButton } from './Components';
-import type { QueryField } from './helpers';
-import { hasLocalityColumns } from './helpers';
+import { hasLocalityColumns, QueryField } from './helpers';
 import type { QueryResultRow } from './Results';
+import { create as createXMLBuilder } from 'xmlbuilder2';
 
 export function QueryExportButtons({
   baseTableName,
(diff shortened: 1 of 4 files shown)


CarolineDenis:
Notes: 
- the `placemark.ele('Point').ele('coordinates').txt(coordinatesValue);` does not provide the right output if the rows have been reordered. 

In the code when exporting selected rows we have:
from fields.filter((field)..)
i.e: field = {filters: [{…}], id: 3, isDisplay: true, mappingPath: (2) ['locality', 'latitude1'], sortType: undefined}
we also have result from selectedResults.forEach((result) ...)
i.e: result = [1, 'DE-N-236', 'Five locations on Geruwaa River along about 25 kilometer stretch between Chisapani and Kotiyaghat', '03/02/1996', '28.3999996000',  '81.1999969000', '28.5666999817', '81.2332992554', 'Line', null, 819]
The id of the field is created when added from the field mapper to the rows. When reordering the rows by dragging it or using the arrows, the fields are reordered in the result array but the id of the field stays the same. Results in retrieving the wrong item for the coordinates. 
In the example above, result[field.id] = result[3] but latitude is now, after reordering the rows, result[4] in the array.

specifyweb/frontend/js_src/tokml.d.ts
```diff
@@ -0,0 +1,17 @@
+declare module 'tokml' {
```
maxxxxxdlp, line 1: where are these definitions used?

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -100,6 +100,41 @@ export function QueryExportButtons({
     return resultToExport.join('\n');
   }
 
+  function handleSelectedResultsKml(): string {
+    const xmlBuilder = createXMLBuilder({ version: '1.0', encoding: 'utf-8' });
```
maxxxxxdlp, line 104: currious what features does this library provide that native DOM APIs don't?

also, if you don't like XML much (like me), I have a handy small function for converting json to XML:
https://github.com/specify/specify7/blob/4ad26fd93656e786765c9585214c65d47db4e97c/specifyweb/frontend/js_src/lib/components/Syncer/xmlToJson.ts#L54-L71

and if you want an even simpler to generate structure, see
https://github.com/specify/specify7/blob/4ad26fd93656e786765c9585214c65d47db4e97c/specifyweb/frontend/js_src/lib/components/Syncer/fromSimpleXmlNode.ts#L12

then SimpleXmlNode can be easily produced by your code

maxxxxxdlp (changes requested):

PR #3726, opened 2023-07-05
Unify color scheme on stat page 

### LINKED ISSUE #3687: Stats Page UI - Sections have a wrong colour
**Describe the bug**
The **default** colour of `Shared` and `Private` section buttons are brighter than the rest of the buttons in Specify. This is harder to notice in light mode but when in Dark Mode, the contrast is noticeable. 

**To Reproduce**
1. Go to Stats Page
2. Change to dark mode
3. Make sure all buttons are using the default colours

**Expected behavior**
The **Default** colour should match the other buttons.

**Beautiful Picture**
Stats Colors

**Desktop:**
 - OS: Mac
 - Browser: Chrome
 - Specify 7 Version: 7.9

CarolineDenis:
Fixes #3687

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Atoms/className.ts
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/className.ts
@@ -12,9 +12,9 @@ export const dialogIconTriggers = {
 };
 
 const buttonClassName = 'button';
-const secondaryButton = `${buttonClassName} hover:brightness-90 dark:hover:brightness-150 bg-[color:var(--secondary-button-color)] text-gray-800 shadow-sm
+const secondaryButton = `${buttonClassName} hover:brightness-90 dark:hover:brightness-125 bg-[color:var(--secondary-button-color)] text-gray-800 shadow-sm
 dark:text-gray-100`;
-const secondaryLightButton = `${buttonClassName} hover:brightness-90 dark:hover:brightness-150 bg-[color:var(--secondary-light-button-color)] text-gray-800 shadow-sm
+const secondaryLightButton = `${buttonClassName} hover:brightness-90 dark:hover:brightness-125 bg-[color:var(--secondary-light-button-color)] text-gray-800 shadow-sm
 dark:text-gray-100`;
 const containerBaseUnstyled = `flex flex-col gap-4 overflow-scroll
   overflow-x-auto [overflow-y:overlay] [scrollbar-gutter:stable]`;
--- a/specifyweb/frontend/js_src/lib/components/Atoms/__tests__/__snapshots__/Button.test.tsx.snap
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/__tests__/__snapshots__/Button.test.tsx.snap
@@ -5,7 +5,7 @@ exports[`Button.BorderedGray renders without errors 1`] = `
   <button
     class="button rounded cursor-pointer active:brightness-80 px-4 py-2
     disabled:bg-gray-200 disabled:text-gray-500 dark:disabled:!bg-neutral-700 gap-2
-    inline-flex items-center capitalize justify-center shadow-sm button hover:brightness-90 dark:hover:brightness-150 bg-[color:var(--secondary-button-color)] text-gray-800 shadow-sm
+    inline-flex items-center capitalize justify-center shadow-sm button hover:brightness-90 dark:hover:brightness-125 bg-[color:var(--secondary-button-color)] text-gray-800 shadow-sm
 dark:text-gray-100 ring-1 ring-gray-400 shadow-sm dark:ring-0
     disabled:ring-gray-400 disabled:dark:ring-neutral-500"
     type="button"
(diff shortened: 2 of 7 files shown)


CarolineDenis:
Change the color of headers to be unified with custom color 3
Before: 
Screenshot 2023-07-05 at 12 57 30 PM
With new push: 
Screenshot 2023-07-05 at 12 50 16 PM

CarolineDenis:
Regarding the button "Collection" being brighter. 
In specify we have a class that makes a button brighter when pressed. With the line `aria-current={isCurrent ? 'page' : undefined}` line 20 in Statistics/Buttons.tsx the button is pressed when selected. 
In other words, the behavior is correct with the current code.
@realVinayak what do think? We could choose to use another button that doesn't add brightness when pressed .

melton-jason (approved):

bronwyncombs (changes requested):
Functions correctly! 
Regarding the brightness, though, I did find colors like this. They are decent shades for the category headers and other areas, but too bright on the pressed button. For most colors this is fine, however, I think it should be changed if that can be done easily

Screen Shot 2023-07-06 at 9 15 54 AM
Screen Shot 2023-07-06 at 9 12 18 AM

grantfitzsimmons:
>Regarding the brightness, though, I did find colors like this. They are decent shades for the category headers and other areas, but too bright on the pressed button. For most colors this is fine, however, I think it should be changed if that can be done easily

I agree with @bronwyncombs – Could we make the color for a selected category the same as the headings? Does this introduce any UI concerns?

CarolineDenis:
> > Regarding the brightness, though, I did find colors like this. They are decent shades for the category headers and other areas, but too bright on the pressed button. For most colors this is fine, however, I think it should be changed if that can be done easily
> 
> I agree with @bronwyncombs – Could we make the color for a selected category the same as the headings? Does this introduce any UI concerns?

Agree with you, as I explained in my comment above, the color is the same as the headers just that the button is 'pressed'. I am waiting on Vinnie to know what he would like to do with this.

bronwyncombs:
> I am waiting on Vinnie to know what he would like to do with this.
Yes of course! Totally understand that. I really just wanted to add some more context to your comment for when he thinks it over  ◡̈

bronwyncombs (approved):
Looks good to me! 
- Button for selected page is no longer pressed automatically, making it more uniform. This also allows for a better view of a wider color selection. 
- A border color is now applied to selected page button (determined by `Secondary button color`) which helps supplement the visual distinction that the pressed button look had. 
- Text color of the page buttons still alternate black and white based on selection status. 
- `Accent color 3` now determines the color of the statistic name when scrolled over in addition to the category headings. (I think this looks nice)

**example of last bullet point**
Screenshot 2023-07-12 at 12 35 27 PM

**general overview**
Screenshot 2023-07-12 at 12 02 18 PM
Screenshot 2023-07-12 at 12 04 33 PM

bronwyncombs (changes requested):
Hi, so sorry about this. After testing with dark mode I realized that there is a necessary change. As you can see in this screenshot, the unselected button on the bottom, "Personal", is hard to read due to text color. Can you make it so the unselected button text is black like it is in light mode and system settings? Also, the border feature is missing from dark mode. 

Screenshot 2023-07-12 at 1 00 46 PM

Further, the color of the buttons when scrolled over is quite contrasting. This does its job, but could be considered too bright for the chosen color. This seems to be the case for all secondary buttons in dark mode. Could you remove this behavior and replace with the border feature of accent color 3 when hovered over?

https://github.com/specify/specify7/assets/135047322/267399d6-fea9-4c6d-a92b-4aeb247ce204

CarolineDenis:
@bronwyncombs, 

- I can make the text black when not selected :+1:
- We do not have rings in dark mode on the button small 
- We could change the value of the brightness added to the buttons when hovered but it needs to be the same value for all Sp. Note that depending on the color the brightness seems stronger or not. (We cannot handle all the possibilities as the user has all options to change them)

grantfitzsimmons:
@bronwyncombs Can you give this another look?

bronwyncombs (approved):
Looks great! 😌 

https://github.com/specify/specify7/assets/135047322/52fa573f-d181-42f1-95b2-49af7342c723

grantfitzsimmons (changes requested):
image

These buttons should not be gray. 
Instead of being the outline color, it should be the button color.

specifyweb/frontend/js_src/lib/components/Atoms/className.ts
```diff
@@ -12,9 +12,9 @@ export const dialogIconTriggers = {
 };
 
 const buttonClassName = 'button';
-const secondaryButton = `${buttonClassName} hover:brightness-90 dark:hover:brightness-150 bg-[color:var(--secondary-button-color)] text-gray-800 shadow-sm
+const secondaryButton = `${buttonClassName} hover:brightness-90 dark:hover:brightness-125 bg-[color:var(--secondary-button-color)] text-gray-800 shadow-sm
```
maxxxxxdlp, line 15: make sure to do quick testing of at least a few places where secondary and secondary light button is used to make sure the hover effect is visible enough even after this change

maxxxxxdlp (changes requested):

grantfitzsimmons (approved):
Beautiful.

PR #3727, opened 2023-07-05
Unify button's size in wb 

### LINKED ISSUE #3661: 7.9 Stats Page Buttons Do Not Match The Rest of Specify
**Describe the bug**
Title.
See Screen Shots of the Top Buttons.

**To Reproduce**
1. Go to Stats Page
2. See error

**Screenshots**
Screenshot 2023-06-22 at 2 32 49 PM

# VS 

Screenshot 2023-06-22 at 2 33 12 PM

**Desktop:**
 - OS: Mac OSx
 - Browser: Chrome
 - Specify 7 Version: 7.9 Dev
 
**Database Name:** NHMD

CarolineDenis:
Fixes #3661


melton-jason (approved):
This is OK if it is the desired behavior. 
This is doing more than changing the size, we are losing the 'pop' of the save button color when it can be saved. 

@specify/ux-testing 
What do we think of losing the save button color when the mapping can be saved? 
Is this better? Worse? 
Personally, I would imagine the save-button color to be useful to new users: a green button in a mountain of grey buttons _screams_ 'i'm important' and draws the eyes more. 
Please provide your thoughts!

Here's a comparison between this branch and production

`issue-3661`
Screenshot_2023-07-06_15-18-50


`production`
Screenshot_2023-07-06_15-13-15

CarolineDenis:
@melton-jason I agree that the save button should be colored as it was before

CarolineDenis:
> > The inconsistency is actually in more places:
> > 
> > * Preferences page
> > * Security Page
> > * Schema Viewer
> > 
> > I do like how larger buttons look, but they also take more space
> > Perharps we should consider using the larger buttons everywhere except for workbench because workbench (and workbench mapper) has a ton of buttons in the toolbar?
> 
> Are the other buttons mentioned in [#3661 (comment)](https://github.com/specify/specify7/issues/3661#issuecomment-1605261411) already using Button.Save?
> 
> If not, do you want to address the inconsistency issue in this Pull Request, or create new ones? I would lean more towards the former.

It was discussed to keep the big buttons everywhere except here in the WB.

grantfitzsimmons:
All save buttons should be the save button color !

bronwyncombs (approved):
Button is unified with preferences and correctly changes color in response to changes made in WB

https://github.com/specify/specify7/assets/135047322/2a935263-eccd-474a-a81d-a37b2a69a2bb

bronwyncombs:
The same issue solved by PR #3667 is present in this branch, though. Cells are not able to be navigated and the count is incorrect. 

https://github.com/specify/specify7/assets/135047322/11de18b0-a966-4835-91d6-6098f92ca289

(Database: NHMD_test_3_8_23)

grantfitzsimmons:
> The same issue solved by PR #3667 is present in this branch, though. Cells are not able to be navigated and the count is incorrect.

This is probably because `v7.9` had not been rebased to production when this branch was made. I just merged back in the base branch, so can you test this again after it finishes running tests?

bronwyncombs:
@grantfitzsimmons Yep, all good now. Thanks for your explanation!

PR #3730, opened 2023-07-06
Reload app resource page after saving to apply immediately changes

### LINKED ISSUE #3665: Does Not Apply Saved App Resources User Immediately
**Describe the bug**
Title.
Let's say I change my own User Preferences in App Resources. They are not applied until I refresh the page. Would be great if they were applied after saving.

**To Reproduce**
1. Go to `App Resources`
2. Go to `User Accounts`
3. Scroll down to yourself, make changes and save.

**Screen Recording**

https://github.com/specify/specify7/assets/53784701/2a9ef005-62da-44d6-be1e-7d6eccf008e3

**Desktop:**
 - OS: Mac OS
 - Browser: Chrome
 - Specify 7 Version: 7.9 Dev

CarolineDenis:
Fixes #3665

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
@@ -237,7 +237,7 @@ export function AppResourceEditor({
                     ...resourceDirectory,
                     scope: getScope(resourceDirectory),
                   });
-                  navigate(`/specify/resources/app-resource/${resource.id}`);
+                  globalThis.location.reload();
                 })
               );
 
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
@@ -40,6 +40,7 @@ import { useAppResourceData } from './hooks';
 import { AppResourcesTabs } from './Tabs';
 import { getScope } from './tree';
 import type { ScopedAppResourceDir } from './types';
+import { useNavigate } from 'react-router-dom';
 
 export function AppResourceEditor({
   resource,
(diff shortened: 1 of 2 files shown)


CarolineDenis:
@maxxxxxdlp what do you think of a reload here to see immediately the changes? Especially needed when in AppResources/UserPreferences

specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
```diff
@@ -126,7 +127,7 @@ export function AppResourceEditor({
       />
     </div>
   );
-
+  const navigate = useNavigate();
```
maxxxxxdlp, line 130: this is not used?

specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
```diff
@@ -236,6 +237,7 @@ export function AppResourceEditor({
                     ...resourceDirectory,
                     scope: getScope(resourceDirectory),
                   });
+                  globalThis.location.reload();
```
maxxxxxdlp, line 240: page reload alone won't always apply the changes.
this is because the browser can cache the old app resource.
Thus, the cache must be cleaned before page is reloaded.

fortunately, you don't have to do this as it's already done on the xml-editor branch

maxxxxxdlp (changes requested):

maxxxxxdlp:
I have mixed feelings about this.
Yes, it's good for seeing changes right away, preventing people from feeling like their changes didn't work
But, app resources are often edited by more experienced people - those can understand that it takes a reload to apply the changes.

The big reason against doing this change is that it would be real annoying if you try to edit several files - which would be common when setting up the database first time or editing forms in several collections

What is worse, the reload is only needed if the app-resource is applicable to current user - if you edited an app resource for a different user or for a different collection, then the reload is completely useless

grantfitzsimmons:
I agree with Max. Let's close this as not planned.

Issue #2731, opened 2023-01-03
Add a warning when clicking Delete All in `statspage`


grantfitzsimmons:
Add a warning when clicking Delete All in `statspage`

realVinayak:
Fixed a while ago

bronwyncombs:
Hi, could you add a warning or some kind of "Are you sure?" message to the `Delete Category` button please? It deletes immediately at this time

https://github.com/specify/specify7/assets/135047322/14230d05-c7e1-4c3d-88df-37473e6bf6cb

realVinayak:
We had that dialog pop up before but it was voted to not have that at the end, so I removed it. Maybe discuss this with team again before reimplementing.

grantfitzsimmons:
Add it back

grantfitzsimmons:
Add it back

Issue #3722, opened 2023-07-05
Apply button color preferences to statistics page save button


bronwyncombs:
**Describe the Bug**
Changing save button color in preferences has no effect on save button in stats page. Currently, the save button is still controlled by 'Secondary button color'. 

**To Reproduce**
1. Go to preferences and change button color
2. Go to statistics and click edit to show save button

**Expected behavior**
Secondary button color shouldn't apply to stats save button. Save color should apply to stats save button.

**Screenshots**
![Screen Shot 2023-07-05 at 11 44 08 AM](https://github.com/specify/specify7/assets/135047322/4d204cf2-e329-4f92-8ef4-199130a9b221)
![Screen Shot 2023-07-05 at 11 43 48 AM](https://github.com/specify/specify7/assets/135047322/17b759ce-54ed-4049-ae36-37920278ec14)

**Desktop:**
 - OS: macOS 
 - Browser: Chrome
 - Specify 7 Version: v7.9-dev

**Database Name:** If applicable
uoregonpaleo

**Additional context**
Database name or any other context about the problem here.

bronwyncombs:
Let's also unify the save button in the Edit Page dialog. Currently is set as an info button

![Screen Shot 2023-07-06 at 2 47 35 PM](https://github.com/specify/specify7/assets/135047322/83db1076-8773-4248-b66a-1b517aa9f3a9)

maxxxxxdlp:
Love all these colors!

makes it look like `material you`:

![image](https://github.com/specify/specify7/assets/40512816/45c1c831-f2b0-4064-896f-71869ac6d404)

grantfitzsimmons:
Solved by https://github.com/specify/specify7/pull/3747

PR #3732, opened 2023-07-06
Add warning dialog on deleting category

realVinayak:
Fixes https://github.com/specify/specify7/issues/2731


bronwyncombs (changes requested):
- Close and Delete should switch positions to match other warning dialogs
- `Close` should also read as `Cancel`
- Will add comment to #3722 about the `Delete` button color preferences

record set delete dialog
stats delete dialog

realVinayak:
Good points. Resolved in https://github.com/specify/specify7/commit/8368b5750d2a3ab5d97693ba16fee791f7d47e06

specifyweb/frontend/js_src/lib/components/Statistics/Categories.tsx
```diff
@@ -136,6 +137,10 @@ export function Categories({
     (items ?? []).some(
       (item) => item.type === 'CustomStat' || item.isVisible === undefined
     );
+
+  const [deletingCategory, setDeletingCategory] = React.useState<
+    number | undefined
+  >(undefined);
```
melton-jason, line 143: (Optional) 
Is there a better name for this variable? 
It is not initially clear what type `deletingCategory` is just by looking at the name. 
These may be too long, but something like `[deletingCategoryIndex, setDeletingCategoryIndex]` makes it _very_ clear exactly what the variables reference at a glance. 

However, other developers can simply look at the type, see how it is used, or use IDE tools to know the type. 

I'll leave it up to you which approach you think is better: the more concise current version, or the more verbose but marginally clearer version.

melton-jason (approved):
The code looks code!
One optional change.

grantfitzsimmons:
> * Close and Delete should switch positions to match other warning dialogs
> * `Close` should also read as `Cancel`
> * Will add comment to [Apply button color preferences to statistics page save button #3722](https://github.com/specify/specify7/issues/3722) about the `Delete` button color preferences

Great looking out. These are excellent observations and comments!

specifyweb/frontend/js_src/lib/components/Statistics/Categories.tsx
maxxxxxdlp: could also make it a bit more verbose (but clearer)

```tsx
const [state, setState] = React.useState<State<'Main> | State<'CategoryDelete',{readonly index: number}>>({type:'Main'});
```

OR even better:
move this state into a separate, smaller component - that component will draw the button, the confirmation dialog, and only call handle delete when deletion is confirmed (and handle delete will be passed to the component as a prop)

that solution works because the delete dialog is modal, meaning if one dialog is open, the user can't open another one - thus we don't need a single central state as there is no case when user needs to have two such dialogs open

specifyweb/frontend/js_src/lib/components/Statistics/Categories.tsx
```diff
@@ -292,6 +297,28 @@ export function Categories({
           {statsText.addACategory()}
         </Button.Secondary>
       )}
+      {typeof deletingCategory === 'number' && (
```
maxxxxxdlp, line 300: this component is already large - it would be easy to move the button and this dialog into a separate component

specifyweb/frontend/js_src/lib/components/Statistics/StatItems.tsx
```diff
@@ -257,8 +257,9 @@ function QueryItem({
           setStatState('valid');
           return formatNumber(data.count);
         }
-        if (status === Http.FORBIDDEN) setStatState('noPermission');
-        setStatState('error');
+        if (status === Http.FORBIDDEN) {
```
maxxxxxdlp, line 260: (optional)
```tsx
setState(status === Http.FORBIDDEN ? 'noPermission' : 'error')
```

specifyweb/frontend/js_src/lib/components/Statistics/Categories.tsx
```diff
@@ -292,6 +297,28 @@ export function Categories({
...
+        <Dialog
+          header={statsText.deleteCategory()}
```
maxxxxxdlp, line 302: would be good to include category name in the dialog (perhaps in the dialog header) so that it's clearer which category they are deleting - this is a good practice

see Asana:

Screenshot 2023-07-06 at 19 49 11

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Statistics/Categories.tsx
realVinayak: Moved to a new file!

specifyweb/frontend/js_src/lib/components/Statistics/Categories.tsx
realVinayak: Added category name in the dialog header.

specifyweb/frontend/js_src/lib/components/Statistics/Categories.tsx
realVinayak: Fixed.

specifyweb/frontend/js_src/lib/components/Statistics/Categories.tsx
bronwyncombs: Hiii, could you also add the red circle with exclamation point image in the top left corner? 

Additionally, to be consistent with other warning dialogs, we may want to make the question the dialog header and add the category name to the description. 

`example`
Screen Shot 2023-07-07 at 9 09 11 AM

specifyweb/frontend/js_src/lib/components/Statistics/Categories.tsx
```diff
@@ -268,14 +265,10 @@ export function Categories({
...
-                    variant={className.secondaryButton}
-                    onClick={(): void => {
-                      setDeletingCategory(categoryIndex);
-                    }}
-                  >
-                    {statsText.deleteCategory()}
-                  </Button.Small>
+                  <DeleteStatsCategory
+                    categoryLabel={label}
+                    onDelete={() => handleRemove?.(categoryIndex, undefined)}
```
maxxxxxdlp, line 270: is this a bug or intentional: you are only displaying the remove button if handleAdd is a function (rather than when handleRemove is a function)?

even if those are coupled at the moment (always both undefined or not), they may not be forever (i.e, imagine a user in the future that does not have delete permissions, but has add permission)

maxxxxxdlp (approved):

bronwyncombs (changes requested):
I think we need to have quotations around the category name in the header to distinguish it among the dialog. A colon would also work. Other than that, I think this is great. 
Screenshot 2023-07-13 at 12 45 07 PM

emenslin (changes requested):
I agree with Bronwyn's requested changes above, I think quotes or a colon would make this look better, I feel like without something there it looks a little weird.
Screenshot 2024-01-25 125323

emenslin (approved):
Looks a lot better, I like it! 
Screenshot 2024-02-20 150359

beach53:
Hmm.  I think this is a little too verbose and redundant.  I would recommend shortening the second sentence to:
This will permanently delete 'Attachments'.

If I were an invited editor, I would also suggest making the first sentence a little more formal/polite with:
Are you certain you would like to delete the category?

realVinayak:
Ayy Jim's here

beach53:
Just discovered this interesting site.

Issue #3725, opened 2023-07-05
Sorting by new query field heading adds duplicate field to right in stats


bronwyncombs:
**Describe the bug**

If a new field is added to query stat, sorting by clicking the column header of the newly added field produces a duplicate field column to the query. 
- No duplication if using the builder '⚫' button. No duplication when sorting pre-existing fields in the query. 
- Saving before sorting new field still duplicates. Closing and sorting the newly saved fields still duplicates. 
- Duplicated column does disappear after closed and re-opened (even if saved). 

**To Reproduce**
1. Go to statistics page
2. Click on a query stat 
3. Add a new field to the query
4. Click column title of new field to sort
5. See duplication error

**Expected behavior**
Column should be sorted without duplicated field added

**Screenshots**

https://github.com/specify/specify7/assets/135047322/542e43aa-55f3-468b-883f-5c768d2b08af

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Specify 7 Version: 7.9-dev

**Database Name:** If applicable
herb_rbge_3_31_23

**Additional context**
Database name or any other context about the problem here.

realVinayak:
Doesn't seem to be an internal bug but still needs fixing.

Looks like whenever you have a query with some not visible fields, sorting creates additional fields and replaces the previously not visible ones. So, even if you make a query like 
![Screenshot (182)](https://github.com/specify/specify7/assets/61122018/e4812a1c-9c3b-462c-865d-b13f45e49c9b)

and then click on sort, it would still replace that field. like below
![Screenshot (183)](https://github.com/specify/specify7/assets/61122018/08ddf3ab-b571-4dbb-b717-480aedf3c24c)
Not sure where this is in on the priority list @grantfitzsimmons but this could be a big performance issue because it takes away filters from the query, so the query will consider more rows

grantfitzsimmons:
Let's aim to have this fixed by `v7.9`.

Issue #3733, opened 2023-07-06
Sorting by locality column header reveals hidden search fields in mapper


bronwyncombs:
**Describe the bug**
When querying on any locality field (on any table) sorting by column header renders fields being searched in the background in the mapper. This then leads to the issue described in #3683 where those fields carry over after opening a new query through the side bar.

**To Reproduce**
1. Go to Query Builder
2. Choose locality table or other linked table
3. Add a locality field to mapper 
4. Click `Query`
5. Sort by the locality column header
6. See error

**Expected behavior**
Should simply sort the column without extra fields rendering in mapper.

**Screenshots**

https://github.com/specify/specify7/assets/135047322/df53956b-6594-4be9-9b80-1ff9f654ce7d

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Specify 7 Version: 7.9-dev

**Database Name:** If applicable
herb_rbge_3_31_23

**Additional context**
Database name or any other context about the problem here.
Crash Report thrown after switching tables:
[Specify 7 Crash Report - 2023-07-06T15_30_55.572Z.txt](https://github.com/specify/specify7/files/11969903/Specify.7.Crash.Report.-.2023-07-06T15_30_55.572Z.txt)
- Just tested and am able to recreate in `v7.8.13` (my fault)

bronwyncombs:
More context: I believe this issue is the root of the problem with issue #3683. If that is the case they can be closed together

CarolineDenis:
Issue introduced in V7.8.0

maxxxxxdlp:
See https://github.com/specify/specify7/issues/3683#issuecomment-1624553415

CarolineDenis:
duplicate of https://github.com/specify/specify7/issues/3725

grantfitzsimmons:
Solved by https://github.com/specify/specify7/pull/3741

Issue #3683, opened 2023-06-26
Opening new query from dialog after sorting in a current query auto-loads Locality fields


bronwyncombs:
**Describe the bug**
After querying on a CE with multiple fields and sorting by catalog number at the top (or any of the fields), trying to begin a new query causes a strange error.
Once first query is produced and results sorted, clicking on "new query" when prompted from sidebar causes the New Query page to load with 6 Locality fields/filters already set in place. Attempting to delete one of these fields with the button adds two "NOT MAPPED" fields (similar to WorkBench) and prompts multiple additional errors. 

**To Reproduce**
1. Go to Queries and add new CE query
2. Add these fields: CE Info (formatted), Locality (formatted), and CO > Cat Number (Any)
3. Click "Query"
4. Click one of the headers at the top to sort by this field
5. Now use side bar to open another new query of a different form
6. See error
        - Trying to delete one of the locality fields causes subsequent errors

**Expected behavior**
Should open a blank new query page without Locality fields present

**Screenshots**

**All of the fields in the screenshot automatically appeared after opening the new query**
![CT query Loc error](https://github.com/specify/specify7/assets/135047322/72913a8f-9d54-48bf-93d8-276b431a846f)

**NOT MAPPED fields after clicking delete button**
![CT query error SS](https://github.com/specify/specify7/assets/135047322/5a73975d-7ce8-49f4-aae3-c045d7b83cff)

**Desktop:**
 - OS: Windows 10
 - Browser: Edge
 - Specify 7 Version: 7.9-dev

**Database Name:** 
 UOregonPaleo (6.8.01) 

**Additional context**

bronwyncombs:
[Specify 7 Crash Report - 2023-06-26T14_10_07.815Z.txt](https://github.com/specify/specify7/files/11871161/Specify.7.Crash.Report.-.2023-06-26T14_10_07.815Z.txt)

Cazza1965:
I'm just a workbench user. Where do I create a support ticket? my workbench has stopped working. I sign in write an SOQL query but then it jumps straight back to the signin page

bronwyncombs:
@Cazza1965 You can reach our support team by sending an email to support@specifysoftware.org. Make sure you provide a description of the issue along with your institution name!

maxxxxxdlp:
FYI: this is likely related by query builder automatically adding several locality fields to a query that includes any locality field

those fields are queried, but not visible in the query fields list, unless the user explicitly added them

[...]

PR #3741, opened 2023-07-07
Fix hidden locality fields being added when sort config is changed via results table 

realVinayak:
Fixes https://github.com/specify/specify7/issues/3725, https://github.com/specify/specify7/issues/3733, and https://github.com/specify/specify7/issues/3683.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsWrapper.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsWrapper.tsx
@@ -13,6 +13,7 @@ import { mappingPathIsComplete } from '../WbPlanView/helpers';
 import type { QueryField } from './helpers';
 import {
   augmentQueryFields,
+  queryFieldIsPhantom,
   queryFieldsToFieldSpecs,
   unParseQueryFields,
 } from './helpers';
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsWrapper.tsx
```diff
@@ -196,15 +198,20 @@ export function useQueryResultsWrapper({
...
-                  const field = displayedFields[index];
+                  const displayField = displayedFields[index];
+                  const lineIndex = allFields.indexOf(displayField);
                   handleSortChange(
-                    replaceItem(allFields, index, {
-                      ...field,
-                      sortType,
-                    })
+                    replaceItem(
+                      allFields.filter((field) => !queryFieldIsPhantom(field)),
```
maxxxxxdlp, line 208: (optional) instead of adding phantom fields and then trying to remove them, you can try using the original fields array before more fields are added

I haven't looked to deep in the code to see if that is possible (if types match), but you can try to take a look around

maxxxxxdlp (approved):

bronwyncombs (approved):
Working as expected!
 
**Stats query**
Screenshot 2023-07-12 at 3 07 12 PM

**Locality base table query**
Screenshot 2023-07-12 at 3 08 21 PM

**Locality not as base table query**
Screenshot 2023-07-12 at 3 09 12 PM

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsWrapper.tsx
realVinayak: In the case of SpAuditLog query, unhidden fields will be added. On sorting in this case, we do want those fields so we cannot use the original fields directly (as they wouldn't contain additional sp auditlog query fields)

PR #3744, opened 2023-07-07
Specify MariaDB recommended version

### LINKED ISSUE #3743: Update `docker-compose.yml` to use `mariadb:10.11`
As we learned from the back-end test failing, we may be on borrowed time using Specify 6 with the latest version of MariaDB just as with MySQL 5.7.

MariaDB 11 is the first to begin deprecating `mysql` and other `mysql*` commands to distinguish itself from MySQL. The plan is to eventually drop support entirely, and it seems in some circumstances, that is already the case.

Specify 6 is working perfectly fine with MariaDB 11 _at this time_. In the code for that app, `mysqldump` and other commands that will be deprecated from MariaDB are used a number of times. There is a good chance it will break compatibility with 6 entirely at some point.

> *To clarify:* MariaDB 10.11 (LTS) has support until 16 February 2028). 
> We can continue recommending this for the next few years.
See https://github.com/specify/specify7/pull/3575/commits/32a8c56dedd4ebafc4f91a8d09933087724302d7|commit related to this https://github.com/specify/specify7/actions/runs/5252784968/jobs/9489281091
 due to us using `mysqladmin` instead of `mariadb-admin`. While this fix allows our back-end tests to run successfully, we should think about what version we are going to recommend.

In addition, MariaDB 11 (on macOS) reports the following when using `mysql`, `mysqladmin`, or similar commands:

`/opt/homebrew/Cellar/mariadb/11.0.2/bin/mysql: Deprecated program name. It will be removed in a future release, use 'mariadb' instead`

grantfitzsimmons:
Fixes #3743

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,7 @@
 version: "3.7"
 services:
   mariadb:
-    image: mariadb
+    image: mariadb:10.11
     volumes:
       - "database:/var/lib/mysql"
       - "./seed-database:/docker-entrypoint-initdb.d"


melton-jason (approved):

CarolineDenis (approved):

maxxxxxdlp:
two suggestions:
- add a comment in the docker-compose explaining why we have to stay on 10.11 for now
- do the same change in the docker-compositions repo

Issue #3000, opened 2023-02-15
Allow opening embedded queries in new tab


maxxxxxdlp:
Query builder is now invoked in a dialog in more and more places. The most notable example is the stats page.
Ability to open the query in a new tab would be very useful.
The difficulty is that the query might not be saved, and so the new tab should open the query with all the unsaved modifications preserved. Can be done by either forcing the user to save the query to open it in new tab, or doing some sort of cross-tab comunication.

grantfitzsimmons:
I'd vote for cross-tab communication–
Maybe this would be something that would make the dialog disabled (link to the other window/tab) and once the changes or options are saved or selected, it could return to the other window. Does that make sense?

grantfitzsimmons:
_This is issue #3000!!!!_

maxxxxxdlp:
Is it okay if both dialog and tab would still be editable, but they would be independent?

> it could return to the other window
you mean close the tab automatically?

grantfitzsimmons:
Yes, that works, and yes I do mean close the tab automatically

maxxxxxdlp:
No, I don't think closing the tab automatically (when they click the "Save" button or else) is a good idea.
That's something I only see in ancient java web applications

PR #3748, opened 2023-07-07
Add Hide Field Mapper button to dialog query 

### LINKED ISSUE #3745: Add Hide Field Mapper button to stats query dialog
**Describe the bug**
There's no button to hide the mapper in stats query dialog

**To Reproduce**
1. Go to Statistics page
2. Open stat with query

**Expected behavior**
Dialog should have `Hide Field Mapper` option somewhere

**Screenshots**

_**Statistics Page:**_
Screen Shot 2023-07-07 at 10 05 53 AM

**_Query Builder:_**
Screen Shot 2023-07-07 at 10 10 20 AM

**Desktop:**
 - OS: macOS
 - Browser: Chrome
 - Specify 7 Version: 7.9-dev

**Database Name:** If applicable
herb_rbge_3_31_23

**Additional context**

Since `Save` is at the bottom and in place of `Save Query` button, it could be better to add this below the mapper by `Count` and `Query`. Depends on what you guys think. Adding this alone at the top might take up space and look out of place/lonely.

CarolineDenis:
Fixes #3745

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Forms/SearchDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/SearchDialog.tsx
@@ -30,6 +30,7 @@ import { formatUrl } from '../Router/queryString';
 import { format } from './dataObjFormatters';
 import { SpecifyForm } from './SpecifyForm';
 import { useViewDefinition } from './useViewDefinition';
+import { wbPlanText } from '../../localization/wbPlan';
 
 const dialogDefinitions = load<Element>(
   formatUrl('/context/app.resource', { name: 'DialogDefs' }),
@@ -261,6 +262,8 @@ function QueryBuilderSearch<SCHEMA extends AnySchema>({
     [model]
   );
   const [selected, setSelected] = React.useState<RA<number>>([]);
+  const [showEmbeddedMappingView, setShowEmbeddedMappingView] =
+    React.useState(true);
   return (
     <Dialog
       buttons={
@@ -278,6 +281,18 @@ function QueryBuilderSearch<SCHEMA extends AnySchema>({
           </Button.Info>
         </>
       }
+      headerButtons={
+        <>
+          <span className="-ml-2 flex-1" />
+          <Button.Small
+            onClick={() => setShowEmbeddedMappingView(!showEmbeddedMappingView)}
+          >
+            {showEmbeddedMappingView
+              ? wbPlanText.hideFieldMapper()
+              : wbPlanText.showFieldMapper()}
+          </Button.Small>
+        </>
+      }
       className={{
         container: dialogClassNames.wideContainer,
       }}
@@ -291,6 +306,7 @@ function QueryBuilderSearch<SCHEMA extends AnySchema>({
         query={query}
         recordSet={undefined}
         onSelected={setSelected}
+        showEmbeddedMappingView={showEmbeddedMappingView}
       />
     </Dialog>
   );
--- a/specifyweb/frontend/js_src/lib/components/Forms/SearchDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/SearchDialog.tsx
@@ -31,6 +31,7 @@ import { format } from './dataObjFormatters';
 import { SpecifyForm } from './SpecifyForm';
 import { useViewDefinition } from './useViewDefinition';
 import { wbPlanText } from '../../localization/wbPlan';
+import { useCachedState } from '../../hooks/useCachedState';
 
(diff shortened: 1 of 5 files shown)


specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -363,7 +365,7 @@ export function QueryBuilder({
         }}
       >
         <div className="flex snap-start flex-col gap-4 overflow-y-auto">
-          {state.showMappingView && (
+          {state.showMappingView && showEmbeddedMappingView ? (
```
maxxxxxdlp, line 368: why add a new prop when you could already reuse existing one? `state.showMappingView`

specifyweb/frontend/js_src/lib/components/Forms/SearchDialog.tsx
```diff
@@ -278,6 +281,18 @@ function QueryBuilderSearch<SCHEMA extends AnySchema>({
           </Button.Info>
         </>
       }
+      headerButtons={
```
maxxxxxdlp, line 284: why add another hide/show mapping view button when the query builder already has one? (and all associated logic) and other buttons too

maxxxxxdlp (changes requested):
I would suggest to fix this by simply displaying the `<QueryHeader />` component for embedded queries (at the moment it's visible only if query is not embedded) - that component includes all the buttons we need.

That component was hidden because we haven't yet decided how to solve the save/save as/open in new tab situation (i.e, should users be able to save embedded query, or open it in new tab instead?)

thus, I would suggest resolving this issue alongside https://github.com/specify/specify7/issues/3000

CarolineDenis:
> I would suggest to fix this by simply displaying the `<QueryHeader />` component for embedded queries (at the moment it's visible only if query is not embedded) - that component includes all the buttons we need.
> 
> That component was hidden because we haven't yet decided how to solve the save/save as/open in new tab situation (i.e, should users be able to save embedded query, or open it in new tab instead?)
> 
> thus, I would suggest resolving this issue alongside #3000

That was the first fix I did but it was decided that having the QueryHeader bellow the Dialog Header wasn't a great design.

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
CarolineDenis: not sure what you mean, since I had a toggle btn in the parent I needed to pass down the boolean. showMappingView doesn't come from the parent of QueryBuilder.

specifyweb/frontend/js_src/lib/components/Forms/SearchDialog.tsx
CarolineDenis: <ToggleMappingViewButton/> needs the fields which we don't have in the dialog header right? 
It's not another show Mapping View button, it's a simple button with the string showFieldMapper for string. It's not a big component that I recreated.

maxxxxxdlp: we already have the code for this button in the header. in fact an entire component just for this button

https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx#L118-L122

maxxxxxdlp:
> having the QueryHeader bellow the Dialog Header wasn't a great design.

yes... that is an issue

let's than stand with the solution you proposed - EXCEPT, the code can be improved

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
maxxxxxdlp: ah, I overlooked that.

Still, you can refactor the code (and simplify it in the process) to fix this

take a look at where the initial value for `state.showMappingView` is coming from and what happens when user tries to call (what does the action handler do) and think of how you can improve that

<p>

Remove `showMappingView` from state and use `useCachedState` directly.
It is not using it at the moment because query builder was developed when I (mistakingly) believed that having a single big state instead of several smaller is best.

Then, you can use `useCachedState` outside of this component too, and any changes to the state will be synced between all usages

</p>
</details>

maxxxxxdlp (changes requested):

carlosmbe (approved):
From a UX Pov it seems fine.
Tested in Stats Page and different fields in Data Entry with a Query Builder Section

maxxxxxdlp (approved):

Issue #2305, opened 2022-10-07
Querying Agent Created By causes a crash


grantfitzsimmons:
<img width="1275" alt="image" src="https://user-images.githubusercontent.com/37256050/194636999-f8abe6d4-a984-458e-bc0c-fcd31dc8b4e9.png">

Run the query:

```
AttributeError at /stored_query/ephemeral/
'Relationship' object has no attribute 'otherSideName'
```

<img width="1237" alt="image" src="https://user-images.githubusercontent.com/37256050/194637047-fc4ef4c0-eae6-4f2b-967f-3ad9e88025c6.png">

[Specify 7 Crash Report - 2022-10-07T15_07_16.669Z.txt](https://github.com/specify/specify7/files/9736885/Specify.7.Crash.Report.-.2022-10-07T15_07_16.669Z.txt)

https://fwri-edge.test.specifysystems.org/specify/query/105/

**Reported By:** Eric at FWRI

realVinayak:
There was a bug in SQLAlchemy with cyclical queries. The previous versions v7.6.1 were doing the query wrong (no runtime errors but the columns are swapped in any cyclical query). Ben fixed the problem with cyclical queries but the side effect is that createdBy and modifiedBy relationships in the agent no longer work because those relationships are unidirectional. Schema change can be done to make the relationships bidirectional.

grantfitzsimmons:
https://fwri2023-favicon-improvements.test.specifysystems.org/specify/query/105/

user: `epost`

```
AttributeError at /stored_query/ephemeral/
'Relationship' object has no attribute 'otherSideName'
```

```
AttributeError at /stored_query/ephemeral/
'Relationship' object has no attribute 'otherSideName'
```

[Specify 7 Crash Report - 2023-01-05T01_27_42.148Z.txt](https://github.com/specify/specify7/files/10348617/Specify.7.Crash.Report.-.2023-01-05T01_27_42.148Z.txt)

grantfitzsimmons:
@maxxxxxdlp Eric at FWRI has reported this same issue again to @mc0822. What is involved in fixing this?

maxxxxxdlp:
@realVinayak do you remember how far you have gone in fixing this?

grantfitzsimmons:
<img width="1065" alt="image" src="https://user-images.githubusercontent.com/37256050/214905469-2cc2457c-daa4-432d-8dda-bc153eb607cb.png">

[...]

PR #3752, opened 2023-07-08
Generate proper SQLAlchemy classes and tables for self relationships

realVinayak:
See (https://github.com/specify/specify7/issues/2305#issuecomment-1627468923)


grantfitzsimmons (approved):
⚠️🚨 **USER REVIEW ONLY!**

It looks like it's working correctly for queries on `sp7demofish` and `fwri`. It very well could break many things, but as a fix for #2305, it seems to work. We need another UX tester in addition to a developer or two. Please read @realVinayak's comment linked at the top of this PR.

specifyweb/stored_queries/build_models.py
```diff
@@ -78,11 +76,11 @@ def make_relationship(reldef):
             column = getattr(table.c, reldef.column)
 
             relationship_args = {'foreign_keys': column}
+            if remote_class is cls:
+                relationship_args['remote_side'] = table.c[ tabledef.idColumn ]
```
melton-jason, line 80: I will have to read up a little more on the sqlalchemy docs to completely understand everything that is going on here, but when compared to how it was previously,  this look alright!

specifyweb/stored_queries/build_models.py
```diff
@@ -78,11 +76,11 @@ def make_relationship(reldef):
...
 
             if hasattr(reldef, 'otherSideName'):
                 backref_args = {'uselist': reldef.type != 'one-to-one'}
-                if remote_class is cls:
-                    backref_args['remote_side'] = table.c[ tabledef.idColumn ]
 
                 relationship_args['backref'] = orm.backref(reldef.otherSideName, **backref_args)
```
melton-jason, line 85: We will have to look at refactoring this code in the future. 
sqlalchemy has "deprecated" `orm.backref` in favor of `relationship.back_populates` explicitly using `relationship` objects. 
https://docs.sqlalchemy.org/en/20/orm/backref.html

I'm sure there are plenty of other things like this as well that we should address.

grantfitzsimmons:
@realVinayak Once you feel confident in this, please make it a full PR and request review as you see fit. Can you include testing instructions as well?

realVinayak:
@grantfitzsimmons I'd actually like to take some time to write automated test for it - to check if actual datamodel and SQLAlchemy datamodel are same or not. I'm half way done through that (in fact my tests also spotted more fields being flipped like this in previous versions of SQLAlchemy classes - for example Agent->Organization - which are now fixed in this branch). 
That automated test would take away the need of most UI testing for this - and would also catch errors like this in future (the current code makes some subtle assumptions about the datamodel - if datamodel changes, it would generate wrong relationships).
I'm on track to wrap automated testing by 7.9.1. Does that work?

specifyweb/stored_queries/build_models.py
melton-jason: Ah, yes, no worries then. 
This works for now.

specifysoftware:

PR #3755, opened 2023-07-10
Enable 'set collections' button when user is set as admin

### LINKED ISSUE #2695: Setting a user as admin does not set their 6 collections
image

Found this today when I set `zrogers` to admin in Specify 6. I was still encountering the "no collections" warning in 6 until I unset admin, selected the collection, then selected the collections.

This is all on a local install of 6, 7, and MariaDB on macOS.

CarolineDenis:
Fixes #2695

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Security/UserCollections.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Security/UserCollections.tsx
@@ -110,10 +110,8 @@ function UserCollectionsUi({
 
 export function UserCollections({
   user,
-  isAdmin,
 }: {
   readonly user: SpecifyResource<SpecifyUser>;
-  readonly isAdmin: boolean;
 }): JSX.Element {
   const [isOpen, handleOpen, handleClose] = useBooleanState();
   return (
@@ -122,12 +120,10 @@ export function UserCollections({
         className="w-fit"
         disabled={
           // Admin users have access to all collections
-          user === undefined || isAdmin || user.isNew()
+          user === undefined || user.isNew()
         }
         title={
-          isAdmin
-            ? userText.notAvailableOnAdmins()
-            : user === undefined
+          user === undefined
             ? commonText.loading()
             : user.isNew()
             ? userText.saveUserFirst()
--- a/specifyweb/frontend/js_src/lib/components/Security/UserComponents.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Security/UserComponents.tsx
@@ -250,7 +250,7 @@ export function LegacyPermissions({
           />
           {hasPermission('/admin/user/sp6/collection_access', 'read') &&
           hasTablePermission('Collection', 'read') ? (
-            <UserCollections isAdmin={isAdmin} user={userResource} />
+            <UserCollections user={userResource} />
           ) : undefined}
         </div>
       )}


specifyweb/context/views.py
```diff
@@ -119,7 +159,8 @@ def user_collection_access_for_sp6(request, userid):
         check_permission_targets(None, request.specify_user.id, [Sp6CollectionAccessPT.update])
         collections = json.loads(request.body)
         user = Specifyuser.objects.get(id=userid)
-        set_users_collections_for_sp6(cursor, user, collections)
+        if len(collections) > 0:
```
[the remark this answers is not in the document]
melton-jason: The logic for this, then, should be handled in the `set_users_collections_for_sp6` function as we will need to modify the SQL query when they want to _unset_ every collection for a user.

specifyweb/frontend/js_src/lib/components/Security/UserCollections.tsx
```diff
@@ -110,10 +110,8 @@ function UserCollectionsUi({
 
 export function UserCollections({
   user,
-  isAdmin,
```
melton-jason, line 113: The `isAdmin` boolean was here for a reason. 
We should make sure removing this functionality is what is intended!
I'm saying this because we may not be addressing the _real_ issue by scraping off this parameter!

Conceptually, the Specify 6 `SetCollections` button was disabled when the user is an admin because admins should _always_ have access to every Collection. 
A user didn't have to worry about setting Specify 6 collection access for admins because it was implicit. 

If this assumption should be maintained, then maybe there's a separate bug in how we set Specify 6 collection permissions for newly created admins!

melton-jason: @specify/ux-testing What are your thoughts on this? Should you (a user with the correct permissions) be able to modify which Specify 6 collections an administrator has access to? 
In terms of Specify 6 collection permissions and access, how do we define an admin?

I understand that flexibility might be nice, but should this rule be kept in place?

melton-jason (changes requested):
I am not sure that this is a correct fix, especially allowing users (with the right permissions) to modify the Specify 6 collection permissions of administrators (see my comment regarding this). 

Besides conceptual considerations, there is a bug related to these changes that should be addressed.

specifyweb/frontend/js_src/lib/components/Security/UserCollections.tsx
grantfitzsimmons: Specify 6 admins do not immediately have access to all collections. They are given permission to each collection one at a time. It is not like how it is in Specify 7.

It was not implicit even though admins had the ability to add themselves to any collection they chose. Admins had to give themselves (or have another admin grant) access to a collection as you have to have an agent in each discipline for it to work correctly.

It's not a rule, it's a mistake introduced because the development team did not know admins are not given access to all collections automatically in Specify 6 like they are in Specify 7. It's not a huge deal, it's just an oversight that has caused issues with many users since the introduction of the Security & Accounts panel.

grantfitzsimmons:
See [my comment](https://github.com/specify/specify7/pull/3755#discussion_r1262758830)

grantfitzsimmons:
We need to 
* Give the user access to set collections for admins
* Make sure they can give access to 0 or many collections

specifyweb/frontend/js_src/lib/components/Security/UserCollections.tsx
CarolineDenis: discussed: give access to set Collection button when 'is admin' + allow to choose 0 collection.

melton-jason: Got it. In this case, we should keep this change and allow users with the correct permissions to modify the Specify 6 collection access/permissions of administrators.

melton-jason (approved):
This is really what the Specify 6 message displays when a user doesn't have access to collections? :facepalm: 
no_collections

maxxxxxdlp:
`Specify will terminate` is not menacing sounding at all 🤖💣💥  ehem, Terminator

almost like `Specify will exterminate ...`

grantfitzsimmons (approved):
Works fine. I am still caught off guard by the fact that the "Set Collections" dialog has its own save button and acts independently of the rest of the options in the user page.

PR #3762, opened 2023-07-11
Allow to enter long query name without spacings

### LINKED ISSUE #3760: Remarkably long name without spaces can hide stats page item count
**Describe the bug**
If you have a remarkably long name. The number can no longer be seen. I expect this to be an issue if the Queries in a Database are very long descriptive titles written in Camel Case or with Underscores for whatever reason.

**To Reproduce**
1. Go to Stats Page
2. Click on Edit
3. Add a long title item without spaces

**Expected behavior**
Multi Line Title

**Screenshots**
Left is long string without spaces. Right is the full name of Bangkok with spaces
Screenshot 2023-07-10 at 4 58 20 PM

**Desktop:**
 - OS: Mac OS
 - Browser: Chrome
 - Specify 7 Version: 7.9

CarolineDenis:
Fixes #3760


maxxxxxdlp (changes requested):
I don't know if `break-all` is what you want to use in this case. Compare these two:

Screenshot 2023-07-12 at 18 19 28

I think `break-words` works better in almost all cases.

What do you think?

maxxxxxdlp:
(optional) @carlosmbe feel free to go over all the places in specify that display content coming from user, and test if similar bug is reproducable

carlosmbe (approved):

carlosmbe (approved):
Works in Stats Page and approving so that it gets merged. 

Following Max's suggestion I will investigate the issue in other sections of Specify but will give those their own issues as to not delay 7.9.

melton-jason (commented):
I agree that `break-all` (`word-break: break-all`) potentially handles more use cases than `break-words` (`overflow-wrap: break-word`).

Abruptly breaking a string mid-word never looks great, but it does guarantee every character is visible on the screen without scrolling. 

However, I think that most of the time, break-words will probably look and function _better_ than break-all. 
I can't imagine having a string long enough to encounter the issue will happen often. 
If it does, the query count is still visible by scrolling the page. 

I guess this is a question about how valuable the query count VS query name is. 
Should users _always_ be able to see the stat result, even if it means their exceptionally long query name may be wrapped to the next line? 

How long do we thing the longest word in a query will be? 
Personally, I think `break-words` might be slightly better. I feel that break-words will handle most query names very nicely, and not nicely for a select few names. 
While `break-all` handles the select few names marginally better, while treating some of the 'most query names' much worse. 

In the end, they both have a valid argument and use case, so I could go either way with enough justification.

maxxxxxdlp:
Interesting, I didn't expect `break-words` to produce such a result.

See https://tailwindcss.com/docs/word-break#normal where it correctly breaks only if necessary VS https://tailwindcss.com/docs/word-break#break-all where it breaks even if there is no need to - I think that will be bad in many non-extreme use cases.

for example `Collection Object Collected by John Doe` might be broken:

```
Collection Object Collected by John D
oe
```

it might be that some style change (i.e width auto, width fit, width min-content or overflow hidden) is required to make it work

Thus my suggestion is to try to make `break-words` work correctly and investigate why it doesn't, and only use `break-all` if all else fails

maxxxxxdlp:
I tried all supported properties, and it looks like you simply need to add this CSS style: `word-break: break-word;` - no other code changes are needed

Not sure why it's not in tailwind - feel free to investigate why

Afterward, just add a `style={{ word-break: break-word; }}` attribute to the span

CarolineDenis:
@bronwyncombs @carlosmbe could you test this one more time please?

melton-jason (approved):
I'm not a fan of using `break-word`, because it has been [deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#browser_compatibility). However, it looks like most browsers support it, and will most likely maintain backwards-compatibility. 
So if this is the best option for our use case, then we can probably get away sneaking it in.

@specify/ux-testing 
Can this be looked at and tested now? Just see if the word wrapping is sensible on multiple browsers (Chrome, Firefox, Safari, etc.). If so, then we can merge this.

bronwyncombs (commented):
On all browsers, formatting of long name in category view looks good. 
Chrome: 

https://github.com/specify/specify7/assets/135047322/39616805-fab6-4867-b950-fc5f396c3a0b

Firefox: 

https://github.com/specify/specify7/assets/135047322/1c013b0b-2961-489c-ae16-d84e3a2be5f4

Safari: 
Screenshot 2023-07-18 at 12 25 40 PM

Safari and Firefox do cut the name off in the query dialog on click, but Chrome formats this correctly. 
On Firefox and Safari, not Chrome: 

https://github.com/specify/specify7/assets/135047322/e284967b-4f78-4dcf-af51-03f2d9b4fd2d

Lastly, this issue branch has the blue "i" icon in the top left of the stat query dialog, but this isn't in v7.9-dev. Intentional?

Screenshot 2023-07-18 at 12 31 02 PM

CarolineDenis:
For the comment "Lastly, this issue branch has the blue "i" icon in the top left of the stat query dialog, but this isn't in v7.9-dev. Intentional?" 
This PR doesn't touch that component so I'm not exactly sure what is happening there

CarolineDenis:
@bronwyncombs, I just merged with the last v7.9-dev. The I blue icon is gone now :+1:

maxxxxxdlp:
> I'm not a fan of using break-word, because it has been [deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#browser_compatibility). However, it looks like most browsers support it, and will most likely maintain backwards-compatibility.

@melton-jason Very good catch. That explains why tailwind used `overflow-wrap: break-word;` instead of `word-break: break-word;`.

Though, it's still weird that `overflow-wrap` didn't work for us (despite being supported by all browsers). If anyone is curious, feel free to look more, but not a priority as this works and should work for the time to come

PR #3764, opened 2023-07-11
Display total count of records in a record set in the slider

### LINKED ISSUE #3756: Browse In Forms no longer allows navigating between all returned query items
**Describe the bug**
"Browse in Forms" used to show the entire number of records returned in the query. Now it only shows the count of the records loaded in the query results, removing the ability to navigate between all items in the query results.

**To Reproduce**
1. In `v7.8.10`, run a query with a large number of results
2. See that the total number of records returned are navigable using the buttons when browsing in forms: 
   image

3. In `v7.8.11`+, run a query with a large number of results
4. See that only the number of loaded query results are navigable using the buttons when browsing in forms:
   image

**Expected behavior**
It should behave as it did prior to `v7.8.11`

**Desktop:**
 - macOS, Chrome

**Reported By:** @bronwyncombs

CarolineDenis:
Fixes #3756

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
@@ -328,7 +328,6 @@ function RecordSet<SCHEMA extends AnySchema>({
           )
         }
         ids={ids}
-        totalCount={totalCount}
         isDependent={false}
         isInRecordSet
         isLoading={isLoading}
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
@@ -328,6 +328,7 @@ function RecordSet<SCHEMA extends AnySchema>({
           )
         }
         ids={ids}
+        totalCount={totalCount}
         isDependent={false}
         isInRecordSet
         isLoading={isLoading}
--- a/specifyweb/frontend/js_src/lib/components/Forms/DeleteBlocked.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/DeleteBlocked.tsx
@@ -159,7 +159,6 @@ function BlockerPreview({
           headerButtons={undefined}
           ids={resolvedIds}
           isDependent={false}
-          totalCount={resolvedIds.length}
           mode="edit"
           model={table}
           newResource={undefined}
--- a/specifyweb/frontend/js_src/lib/components/Forms/DeleteBlocked.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/DeleteBlocked.tsx
@@ -159,6 +159,7 @@ function BlockerPreview({
           headerButtons={undefined}
           ids={resolvedIds}
           isDependent={false}
+          totalCount={resolvedIds.length}
           mode="edit"
           model={table}
           newResource={undefined}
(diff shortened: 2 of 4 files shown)


melton-jason (changes requested):
This bug was not present in `v7.8.11` and was introduced in `v7.8.12`. 

I am not sure which changes caused this bug in `v7.8.12` and what changes were made to the `RecordSelectorFromIds`, but this fix doesn't seem to address the underlying issue. 

From what I can gather from the changes here, the `ids` were replaced from an array of _all_ resources, to just those that are fetched. If this is the case, was this change intentional? 

I would rather not add another prop if it is unnecessary, but if those changes are warranted then this may be unavoidable.

CarolineDenis:
@melton-jason I tried 7.8.10 and had the total number of resources. Than tried 7.8.11 and only the number of fetched resources was displayed. The changes from 10 to 11 was that we deleted that prop 'total count' that was passed from 'ToForms' and set the total count for the slider to be equal to ids.length which is not correct.

maxxxxxdlp:
> The changes from 10 to 11 was that we deleted that prop 'total count' that was passed from 'ToForms' and set the total count for the slider to be equal to ids.length which is not correct.

I made that change. I did it because `ids.length` matches the total number of records - but I overlooked the fact that that is only the case if you are in a record set. See where I create `ids` to have as many items as `totalCount`:

https://github.com/specify/specify7/blob/39cb18c84e94bd783fc10757b593039e683df8b8/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx#L199

Unfortunately, I forgot that similar does not exist for query results.

Solutions:

1. Either modify this line to create an array of length `totalCount`:
   
 
 
 
 https://github.com/specify/specify7/blob/39cb18c84e94bd783fc10757b593039e683df8b8/specifyweb/frontend/js_src/lib/components/QueryBuilder/ToForms.tsx#L89

2. OR add a `totalCount` prop (which could be made optional)

With either solution, a lot of testing would be appreciated, to make sure we don't overlook any other cases

maxxxxxdlp (changes requested):
Can you fix the typescript issues?
There are small issues with the code that might be fixed once you go over all typescript issues

<p>

Take a look carefully at how `resolvedTotalCount` was calculated before and how it is calculated now - there is some difference, which is incorrect


</p>
</details> 

<p>

On line 41, give default value to `totalCount`:

```tsx
totalCount = ids.length + (typeof newResource === 'object' ? 1 : 0)
```

Then go over though the fields you edited to add `totalCount={}` prop, and remove it everywhere where it's not needed (where `ids.length` always matches `totalCount` even if there are >80 records) since it's now optional

</p>
</details>

specifyweb/frontend/js_src/lib/components/Forms/DeleteBlocked.tsx
```diff
@@ -159,6 +159,7 @@ function BlockerPreview({
           headerButtons={undefined}
           ids={resolvedIds}
           isDependent={false}
+          totalCount={resolvedIds.length}
```
maxxxxxdlp, line 162: this line probably be removed as that component will already get total count from resolvedIds.length

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -328,6 +328,7 @@ function RecordSet<SCHEMA extends AnySchema>({
           )
         }
         ids={ids}
+        totalCount={totalCount}
```
maxxxxxdlp, line 331: this line can probably be removed as `ids` length should always equal `totalCount`

melton-jason (approved):

PR #3769, opened 2023-07-12
Add minimum width to queryBuilder overlay in stat page 

### LINKED ISSUE #3662: More Consistent Query Builder Overlay in Stats Page
It'd be awesome if we had one consistent size. Like maybe large or small. Perhaps it's a taste thing but the different sizes can be quite jarring.

Screenshot 2023-06-22 at 2 55 45 PM
Screenshot 2023-06-22 at 2 56 05 PM
Screenshot 2023-06-22 at 2 55 55 PM
Screenshot 2023-06-22 at 2 56 28 PM

CarolineDenis:
Fixes #3662

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Statistics/ResultsDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Statistics/ResultsDialog.tsx
@@ -86,6 +86,7 @@ export function FrontEndStatsResultDialog({
       className={{
         container: dialogClassNames.wideContainer,
       }}
+      dimensionsKey="QueryBuilder"
       header={label}
       onClose={handleClose}
     >


maxxxxxdlp (approved):
hmm, I didn't realize that the code already had a `className={}` and yet #3662 is still happening.

Still, your fix is a good improvement, that will fix the bug if user manually resizes the dialog at least once

melton-jason (approved):
I didn't know we had this prop for Dialog components!

Neither did I know about [Rezise Observers](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver)! 
Neat! :partying_face:

carlosmbe (commented):
Could I get a screenshot of what the approximate minimum looks like visually?
Not able to notice the difference right now. 
Testing on NHMD Amber

melton-jason:
> hmm, I didn't realize that the code already had a `className={}` and yet #3662 is still happening.
> 
> Still, your fix is a good improvement, that will fix the bug if user manually resizes the dialog at least once

Yes, it definitely looks like there should already be a minimum width to the dialogs already by the className definition:
https://github.com/specify/specify7/blob/22c79e9e4be7ef456c0a7c53535b146333ae427f/specifyweb/frontend/js_src/lib/components/Molecules/Dialog.tsx#L56

Should we increase this width? :thinking:

CarolineDenis:
Before: 
Screenshot 2023-07-13 at 1 24 42 PM
Now: 
Screenshot 2023-07-13 at 1 24 04 PM

maxxxxxdlp:
The main consequence of this fix is not a change in minimum width, but the fact that whatever width you now set manually for any stat query dialog, it would use the same width for all dialogs, so the original problem with each query opening dialog of different width is solved.

We are careful not to impose too much of a minimum width as that is needlessly limiting the user (for example, if they want to have two dialogs side by side, or have a narrower screen)

Issue #3259, opened 2023-03-28
Going to a record set triggers a form change


realVinayak:
**Describe the bug**
If you click on 'Record Sets', then I am shown the list of record sets for that user. Try going from one record set to another and I get the dialog message 'This form has not been saved' even if I didn't change anything. A major issue because I wanted to know if I accidentally changed anything or not. This is in production by the way.

**To Reproduce**
1. Click on 'Record Sets' on the menu
2. Click on any record set
3. Click on 'Record Sets' again and go to another record set
4. The dialog 'This form has not been saved.' shows up.

**Expected behavior**
Shouldn't trigger the form change. Or at least tell me what changed.

![image](https://user-images.githubusercontent.com/61122018/228246447-edc2aab1-1326-48ec-aff1-a97d323b1c18.png)

realVinayak:
The url in the screenshot: https://osuorton3823-edge.test.specifysystems.org/

realVinayak:
Happens in Chrome, Edge and Firefox

maxxxxxdlp:
You have an unsaved form behind the dialog.
Opening the dialog is fine and doesn't trigger the unload protect dialog.
Actually clicking on a record set would replace the form with a record set, thus unload protect dialog is triggered to warn you about this.
Not a bug.
Though, if you think this is a UX issue, do you have suggestions on improving this?

realVinayak:
Not showing the save dialog if there actually weren't changes! I discovered
this while on production database and thought I changed the form

maxxxxxdlp:
>  Not showing the save dialog if there actually weren't changes

Ah, that's a separate issue, not connected to record sets.

Two bugs here:
1. When collection object is created, collection relationship is set, which triggers save blocker
2. catalogedDatePrecision is set to 1 by default, which triggers save blocker

maxxxxxdlp:
TODO:
1. Check if the two bugs mention in https://github.com/specify/specify7/issues/3259#issuecomment-1487878114 are still reproducable on xml-editor
2. Fix them

maxxxxxdlp:
From @bronwyncombs:

> **Describe the bug** Using the `Browse in Forms` button in query dialog through stats page requires save before moving pages.
> 
> **To Reproduce** Steps to reproduce the behavior:
> 
> 1. Go to statistics page
> 2. Click on a statistic with query attached
> 3. Click on `Browse in Forms`
> 4. Use paginators
> 5. See error
> 
[...]

PR #3774, opened 2023-07-12
Don't trigger unload protect when initializing resources

melton-jason:
Addresses https://github.com/specify/specify7/issues/3259#issuecomment-1629254502

This was caused because of changes related to Business Rules. 
Beforehand, we were not triggering the Backbone `change` event for business rules while the resource was being fetched or saved by Backbone. I have restored this functionality. 

This has probably not solved the underlying cause of #3259 in its entirety, but should prevent business rules from triggering unload protect while the record is being loaded.  

To test: 
Navigate to an existing Determination record which has `taxon` data.
The Save Button should be disabled, and when navigating away from the Determination in a record set, the unload protect dialog should not be displayed.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
@@ -37,7 +37,7 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
   }
 
   private changed(resource: SpecifyResource<SCHEMA>): void {
-    if (resource.isBeingInitialized && typeof resource.changed === 'object') {
+    if (!resource.isBeingInitialized && typeof resource.changed === 'object') {
       Object.keys(resource.changed).forEach((field) => {
         this.checkField(field);
       });
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
@@ -37,7 +37,7 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
   }
 
   private changed(resource: SpecifyResource<SCHEMA>): void {
-    if (resource.changed !== undefined) {
+    if (resource.isBeingInitialized && typeof resource.changed === 'object') {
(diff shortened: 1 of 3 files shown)


melton-jason:
@CarolineDenis 
I should've separated the formatting of `resourceApi.js` into its own commit :facepalm:

Here are the lines in `resourceApi.js` related to this fix:
ttps://github.com/specify/specify7/pull/3774/files#diff-2f5423edccb3536d3e2c4fedda8f2513a4827fb943cbd51eb5536835f408f460R80-R87

CarolineDenis (approved):

maxxxxxdlp (commented):
Not sure if it's helpful, but you can make a field change not trigger a save blocker by adding `silent: true`:

https://github.com/specify/specify7/blob/39cb18c84e94bd783fc10757b593039e683df8b8/specifyweb/frontend/js_src/lib/components/DataModel/treeBusinessRules.ts#L13

maxxxxxdlp (approved):

melton-jason:
> Not sure if it's helpful, but you can make a field change not trigger a save blocker by adding `silent: true`:
> 
> https://github.com/specify/specify7/blob/39cb18c84e94bd783fc10757b593039e683df8b8/specifyweb/frontend/js_src/lib/components/DataModel/treeBusinessRules.ts#L13

This was the very first thing I tried, but does not quite cover every case. 
The resource was still triggering the `change` event when we were setting a `many-to-one` field (specifically, setting preferredTaxon on Determination). 
The following lines of code was the culprit:
https://github.com/specify/specify7/blob/22c79e9e4be7ef456c0a7c53535b146333ae427f/specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.js#L349-L352

bronwyncombs (approved):
Working as expected. Can now navigate `Browse in Forms` from query builder and stats query without unload protect dialog. 

https://github.com/specify/specify7/assets/135047322/d7681583-f510-4d09-b2a3-cc50e20ea4e2

- macOS, chrome, `uoregonpaleo`

grantfitzsimmons:
Let's merge this tomorrow.

PR #3776, opened 2023-07-12
Merge `production` back into `v7.9-dev`

grantfitzsimmons:
Some minor fixes we should bring back into `v7.9-dev` before it's too late.


grantfitzsimmons:
@melton-jason We can merge if everything checks out with you. **Very** safe.

PR #3778, opened 2023-07-13
Rename Security and Accounts Create Button 

### LINKED ISSUE #3716: Rename Security and Accounts `Create Button`
The work flow for adding roles and permissions to a new user is not very User Friendly right now. After discussing with a senior Sales team member they agree.

An area that we could easily improve is the `Collection User Roles` Section. It'd be great if we renamed its `Create` Button to something similar to `Add Role` and perhaps if you'd like to keep the `Create` Button. Use it in place of the `New` Role Button. The `Create Role` Title could also be `Add Role To Collection`.

Screenshot 2023-07-03 at 2 02 58 PM
Screenshot 2023-07-03 at 2 07 01 PM

CarolineDenis:
Fixes #3716

### DIFF
--- a/specifyweb/frontend/js_src/lib/localization/user.ts
+++ b/specifyweb/frontend/js_src/lib/localization/user.ts
@@ -1186,4 +1186,13 @@ export const userText = createDictionary({
     'fr-fr': 'Utilisateur actuel',
     'uk-ua': 'Поточний користувач',
   },
+  addRole: {
+    'en-us': 'Add Role',
+  },
+  addNewRole: {
+    'en-us': 'Add New Role',
+  },
+  addRoleToCollection: {
+    'en-us': 'Add Role To Collection',
+  },
 } as const);
--- a/specifyweb/frontend/js_src/lib/components/Security/CollectionComponents.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Security/CollectionComponents.tsx
@@ -63,12 +63,12 @@ export function CreateCollectionRoleButton({
   readonly collectionId: number;
 }): JSX.Element {
   return isDisabled ? (
-    <Button.Success onClick={undefined}>{commonText.create()}</Button.Success>
+    <Button.Success onClick={undefined}>{userText.addRole()}</Button.Success>
   ) : (
     <Link.Green
       href={`/specify/security/collection/${collectionId}/role/create/`}
     >
-      {commonText.create()}
+      {userText.addRole()}
     </Link.Green>
   );
 }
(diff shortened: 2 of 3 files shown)


maxxxxxdlp (approved):
Looks good to me, but let's have Grant or any other user-facing person check this

grantfitzsimmons:
I have asked @bronwyncombs to select the labels that she thinks are the most intuitive and clear for the user. I will review them once they have been implemented. 

We need to be consistent throughout all areas, so we should not interchangeably use `Create` and `Add`. I will review later!

bronwyncombs (changes requested):
I think that the buttons for User Roles are good as is for both Institution and Collections. 

I think we should change the `Create` button below "User Accounts Assigned to this Collection" to say `Assign User` or something similar instead. 

That would be one version of "consistent" by keeping buttons labeled `Create` in the Institution section. Another option would be to change both `Create` buttons below the User Accounts sections to `Assign` or `Assign User`. I believe I prefer the former to clearly differentiate Institution and Collection, but both options make sense. 

https://github.com/specify/specify7/assets/135047322/e4b9dc9e-dc97-44d9-94cc-e4436e2af7f8

carlosmbe:
Can someone ping or slack me when this is merged so I can update the documentation.

bronwyncombs (changes requested):
For the Add Role dialog, can the headers be different for Institution and Collection? I'd like to see the Institution Role dialog say "Add Role to Institution". If that is a complicated change, I think we should just have it read "Add Role". 

**Current Add Role dialog for Institution:**
Screenshot 2023-07-21 at 9 58 43 AM

---
Everything else is great!!! 

**Institution Buttons**
Screenshot 2023-07-21 at 10 02 27 AM

**Collection Buttons**
Screenshot 2023-07-21 at 10 02 53 AM

bronwyncombs (approved):
Looks good! 😄

specifyweb/frontend/js_src/lib/localization/user.ts
```diff
@@ -1186,4 +1186,16 @@ export const userText = createDictionary({
     'fr-fr': 'Utilisateur actuel',
     'uk-ua': 'Поточний користувач',
   },
+  addRole: {
+    'en-us': 'Add Role',
+  },
+  addNewRole: {
+    'en-us': 'Add New Role',
+  },
+  addRoleToCollection: {
```
grantfitzsimmons, line 1195: This is no longer used from what I can tell– can we remove it?

grantfitzsimmons (approved):
Looks like we could remove `addRoleToCollection` and then merge if it is not being used

PR #3782, opened 2023-07-14
Test from tree

CarolineDenis:


melton-jason (approved):
These snapshots needed to be updated due to the changes in https://github.com/specify/specify7/commit/edd6accc97f589d68ed652524bfef15c02e7f392 which now set the `_tableName` property of resources when `resource_uri` is not a string

grantfitzsimmons:
What does this change do?

PR #3789, opened 2023-07-14
Clone xml-editor

CarolineDenis:
Last version of xml with 7.9-dev merged in it


specifyweb/frontend/js_src/lib/components/Statistics/Categories.tsx
```diff
@@ -125,7 +125,7 @@ export function Categories({
     | ((
         categoryIndex: number,
         itemIndex: number,
-        newLabel: LocalizedString
+        newLabel: string
```
maxxxxxdlp, line 128: why was this change made? `LocalizedString` should be used instead of `string` for every string that is either localized by us or comes from the user (thus presumed to be localized)

by using `LocalizedString` type, TypeScript can catch cases when we forgot to localize a string

same applies for all the other cases where you replaced `LocalizedString` with `string`

maxxxxxdlp (changes requested):

CarolineDenis:
will keep original xml branch in which we merge production

Issue #2061, opened 2022-08-30
[sp7.7] Specify crashes when entering duplicate barcode


benritchie:
**Describe the bug**
If create a new preparation, with the same barcode as an existing prep, and hit save, then specify crashes.  
crash diags attached.   
Desired behaviour would be to display a useful popup to user explaining issue.  

Note, we are using preparations based workflow.  not sure if that's related?

**To Reproduce**
see above

**Expected behavior**
see above

**Desktop:**
seen on multiple browsers, windows 10 and 11
**Database Name:** rbge_herb

**Reported By**
RBGE

benritchie:
[Specify 7 Crash Report - 2022-08-30T11_59_44.850Z.txt](https://github.com/specify/specify7/files/9452999/Specify.7.Crash.Report.-.2022-08-30T11_59_44.850Z.txt)
here's the crash report.  
datestamp was 30/8 at 12.13pm BST

maxxxxxdlp:
@melton-jason related to what you are working on right now. We need to gradually make all errors on save dismissible starting with the ones that are safest.
Can you check if this is already caught by the front-end? Is it just a problem of user clicking "Save" before uniqueness checks were completed? In that case, we should make "Save" button wait for all uniqueness checks to pass before calling `.save()`

melton-jason:
This is not already caught on the frontend, and it is not a uniqueness rule on the backend either. The error is a SQLAlchemy Integrity Error. 
The frontend only respected Uniqueness Rules defined on the frontend. If a rule was missing on the frontend, but existed on the backend, then the backend BusinessRuleException would be thrown. If the value is unique at a database level but missing in both front and backend, then a SQLAlchemy error would be thrown. 
 
> ... Is it just a problem of user clicking "Save" before uniqueness checks were completed? In that case, we should make "Save" button wait for all uniqueness checks to pass before calling .save()

This is highly unlikely, as it updates the save blockers as soon as it detects the field change and runs the checks. I will concede this is a possibility, however,  and an implementation should (hopefully) not be to hard.

melton-jason:
The `Preparation.barCode` field is unique in the way it handles uniqueness (although there may be more cases I am unaware of), and will either require a separate system to be implemented to handle cases such as it, or be specifically hardcoded.

[...]

PR #3790, opened 2023-07-16
Stop cloning the `barcode` field 

### LINKED ISSUE #3787: Cloned records should erase values from unique fields
**Describe the bug**
Consider the case of Preparation `barcode`. It must be unique at the collection level, yet when you clone a record (and its associated preparations), the barcode value is preserved. When saving, you then see a destructive error dialog saying that the barcode is not unique to the collection.

These fields need to be cleared when cloning to prevent situations like this one from taking place!

grantfitzsimmons:
Fixes #3787

**Steps to test:**
1. Go to a record that has a barcode value on the preparation form
2. Clone the record
3. Verify that the barcode is not being cloned

Tested locally and it seems to do the job.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
@@ -234,6 +234,7 @@ export function AppResourceEditor({
         <AppResourcesTabs
           appResource={appResource}
           data={resourceData.data}
+          footer={footer}
           headerButtons={headerButtons}
           isFullScreen={isFullScreen}
           isReadOnly={isReadOnly}
@@ -241,7 +242,6 @@ export function AppResourceEditor({
           resource={resource}
           showValidationRef={showValidationRef}
           onChange={(data): void => setResourceData({ ...resourceData, data })}
-          footer={footer}
         />
       </Form>
       {isFullScreen ? null : footer}
--- a/specifyweb/frontend/js_src/lib/components/DataModel/legacyTypes.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/legacyTypes.ts
@@ -3,6 +3,7 @@
  */
 
 import type { IR, RA } from '../../utils/types';
+import type { BusinessRuleManager } from './businessRules';
 import type {
   AnySchema,
   CommonFields,
@@ -10,7 +11,6 @@ import type {
   SerializedResource,
   TableFields,
 } from './helperTypes';
-import { BusinessRuleManager } from './businessRules';
 import type { SaveBlockers } from './saveBlockers';
 import type { Collection, SpecifyModel } from './specifyModel';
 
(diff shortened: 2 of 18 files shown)


grantfitzsimmons:
Seems like this should also include `uniqueIdentifier`, no?

melton-jason:
Yes this works! 
Although, it shouldn't be an issue in the first place, at least for uniqueIdentifier :sweat_smile: 
Can you check and see if uniqueIdentifier is able to be cloned on the models which now have it as a uniqueness rule (Collection Object, Collecting Event, Locality)? If it _is_ able to be cloned, then it means there was some issue from my commit: https://github.com/specify/specify7/commit/d0e5c75ae11153cc37b1a0c1f29293ee09a71322

This is because Specify does not use the `uniqueFields` Array to directly determine whether a resource can Clone a particular field. The entire logic is found in the `getUniqueFields` function. 

https://github.com/specify/specify7/blob/eb0a310bbc59089dfe840e94a842a29b7a18cecf/specifyweb/frontend/js_src/lib/components/DataModel/resource.ts#L270-L289
- - - 

This change is fine for `barcode`  because no other Model has a field directly called `barcode`, so this would only apply to Preparation.barcode. 
If you wanted to, you can include a comment stating _why_ barcode is in the Array in the first place (see https://github.com/specify/specify7/issues/2061#issuecomment-1439391935)

Alternatively, we can add barCode as a uniqueness to Preparation in `uniqueness_rules.json`. 
Maybe not the best solution, as it would have no effect _other_ than to be included in the Preparation uniqueness rules when calling `getUniqueFields`

grantfitzsimmons:
@melton-jason Thanks for your comment Jason! 

Feel free to change how this is implemented any way you'd like, I just got tired of accidentally cloning barcodes. 😆

maxxxxxdlp (changes requested):
While this might fix it, it is probably not a correct solution.

Is barcode a unique field?

If not, then make it into a unique field - that will fix this issue.

If it is a unique field, the a programmer needs to review the code that determines if field is unique - that will fix this issue. That will fix it not just for this field, but for all fields like this in the future

maxxxxxdlp:
Of course, if you are looking for a very-short term solution, like something sp6 would go for, then what you did would work

maxxxxxdlp:
Reading the previous comments, I would be for editing the `uniqueness_rules.json` to designate the field as unique.

> Maybe not the best solution, as it would have no effect other than to be included in the Preparation uniqueness rules when calling getUniqueFields

@melton-jason can you explain why? If you designate a field as unique as far as front-end aware, then the uniqueness business rule will be active (in addition to fixing the clone issue)

You might have already mentioned it in https://github.com/specify/specify7/issues/2061#issuecomment-1439391935
If so, can you please double-check if this field is really the only one that handles uniqueness in this way?

Also, I heard you were working on ability to mark any field unique - would that feature create a better solution for this issue?

PR #3792, opened 2023-07-17
Fix business rule tests

melton-jason:
Rewrites the business rule tests so they are run correctly. The issue before was with the order of execution for the setup and teardown of Jest tests (https://jestjs.io/docs/setup-teardown).
For example, we would be trying to call  resource.fetch() before the associated `overrideAjax` function was called. 

My changes should have resolved these issues and make sure that the order of operations are correct for Jest tests.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
@@ -1,4 +1,3 @@
-import { overrideAjax } from '../../../tests/ajax';
 import { mockTime, requireContext } from '../../../tests/helpers';
 import { overwriteReadOnly } from '../../../utils/types';
 import { businessRuleDefs } from '../businessRuleDefs';
--- a/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
@@ -1,3 +1,4 @@
+import { overrideAjax } from '../../../tests/ajax';
 import { mockTime, requireContext } from '../../../tests/helpers';
 import { overwriteReadOnly } from '../../../utils/types';
 import { businessRuleDefs } from '../businessRuleDefs';
@@ -144,6 +145,8 @@ describe('DNASequence business rules', () => {
       id: 1,
     });
     dNASequence.set('geneSequence', 'aaa  ttttt  gg  c zzzz');
+
+    await dNASequence.businessRuleManager?.checkField('geneSequence');
 
     expect(dNASequence.get('compA')).toEqual(3);
     expect(dNASequence.get('compT')).toEqual(5);
--- a/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
@@ -117,12 +117,18 @@ describe('Collection Object business rules', () => {
       resource_uri: collectionObjectUrl,
     });
 
+  const orginalEmbeddedCollectingEvent = schema.embeddedCollectingEvent;
+
   beforeAll(() => {
     overwriteReadOnly(schema, 'embeddedCollectingEvent', true);
   });
 
   afterAll(() => {
-    overwriteReadOnly(schema, 'embeddedCollectingEvent', false);
+    overwriteReadOnly(
+      schema,
+      'embeddedCollectingEvent',
+      orginalEmbeddedCollectingEvent
+    );
   });
 
   test('CollectionObject customInit', async () => {
--- a/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
@@ -1,194 +1,144 @@
-import { overrideAjax } from '../../../tests/ajax';
 import { mockTime, requireContext } from '../../../tests/helpers';
+import { overwriteReadOnly } from '../../../utils/types';
 import { businessRuleDefs } from '../businessRuleDefs';
-import { SerializedModel } from '../helperTypes';
 import { getResourceApiUrl } from '../resource';
 import { schema } from '../schema';
-import { Determination } from '../types';
 
 mockTime();
 requireContext();
 
-test('uniqueness rules assigned correctly', async () => {
-  const accessionAgentUniquenessRules = {
-    role: [
+describe('uniqueness rules assigned correctly', () => {
+  test('otherField uniqueness rule assigned', async () => {
+    expect(businessRuleDefs.AccessionAgent?.uniqueIn).toMatchInlineSnapshot(`
       {
-        field: 'accession',
-        otherFields: ['agent'],
-      },
-      {
-        field: 'repositoryagreement',
-        otherFields: ['agent'],
-      },
-    ],
-    agent: [
-      {
-        field: 'accession',
-        otherFields: ['role'],
-      },
-      {
-        field: 'repositoryagreement',
-        otherFields: ['role'],
-      },
-    ],
-  };
-  expect(businessRuleDefs.AccessionAgent?.uniqueIn).toBe(
-    accessionAgentUniquenessRules
-  );
+        "agent": [
+          {
+            "field": "accession",
+            "otherFields": [
+              "role",
+            ],
+          },
+          {
+            "field": "repositoryagreement",
+            "otherFields": [
+              "role",
+            ],
+          },
+        ],
+        "role": [
+          {
+            "field": "accession",
+            "otherFields": [
+              "agent",
+            ],
+          },
+          {
+            "field": "repositoryagreement",
+            "otherFields": [
+              "agent",
+            ],
+          },
+        ],
+      }
+    `);
+  });
+
+  test('Standard rules assigned correctly', async () => {
+    expect(businessRuleDefs.CollectionObject?.uniqueIn).toMatchInlineSnapshot(`
+          {
+            "catalogNumber": [
+              "collection",
+            ],
+            "guid": [
+              "institution",
+            ],
+            "uniqueIdentifier": [
+              "collection",
+            ],
+          }
+      `);
+  });
+
+  test('JSON nulls are converted to undefined', async () => {
+    expect(businessRuleDefs.Permit?.uniqueIn).toMatchInlineSnapshot(`
+          {
+            "permitNumber": [
+              undefined,
+            ],
+          }
+      `);
+  });
 });
 
-const determinationId = 321;
-const determinationUrl = getResourceApiUrl('Determination', determinationId);
-const determinationResponse: Partial<SerializedModel<Determination>> = {
-  id: determinationId,
-  resource_uri: determinationUrl,
-};
+describe('Borrow Material business rules', () => {
+  const borrowMaterialId = 1;
+  const borrowMaterialUrl = getResourceApiUrl(
+    'BorrowMaterial',
+    borrowMaterialId
+  );
 
-const collectionObjectId = 220;
-const collectionObjectUrl = getResourceApiUrl(
-  'CollectionObject',
-  collectionObjectId
-);
-const collectionObjectResponse = {
-  id: collectionObjectId,
-  resource_uri: collectionObjectUrl,
-  catalognumber: '000022002',
-  collection: getResourceApiUrl('Collection', 4),
-  determinations: determinationUrl,
-};
-
-overrideAjax(collectionObjectUrl, collectionObjectResponse);
-overrideAjax(determinationUrl, determinationResponse);
-
-describe('business rules', () => {
-  test('collectionObject customInit', async () => {
-    const resource = new schema.models.CollectionObject.Resource({
-      id: collectionObjectId,
+  const getBaseBorrowMaterial = () =>
+    new schema.models.BorrowMaterial.Resource({
+      id: borrowMaterialId,
+      resource_uri: borrowMaterialUrl,
+      quantity: 20,
+      quantityreturned: 13,
+      quantityresolved: 15,
     });
-    await resource.fetch();
-    expect(resource.get('collectingEvent')).toBeDefined();
-    resource.save();
+
+  test('fieldCheck quantityReturned', async () => {
+    const borrowMaterial = getBaseBorrowMaterial();
+
+    borrowMaterial.set('quantityReturned', 30);
+    expect(borrowMaterial.get('quantityReturned')).toBe(15);
   });
 
-  describe('determination business rules', () => {
-    test('determination customInit', async () => {
-      const determination = new schema.models.Determination.Resource({
-        id: determinationId,
-      });
-      await determination.fetch();
-      expect(determination.get('isCurrent')).toBe(true);
+  test('fieldCheck quantityResolved', async () => {
+    const borrowMaterial = getBaseBorrowMaterial();
+
+    borrowMaterial.set('quantityResolved', 30);
+    expect(borrowMaterial.get('quantityResolved')).toBe(20);
+
+    borrowMaterial.set('quantityResolved', 5);
+    expect(borrowMaterial.get('quantityResolved')).toBe(13);
+  });
+});
+
+describe('Collection Object business rules', () => {
+  const collectionObjectlId = 2;
+  const collectionObjectUrl = getResourceApiUrl(
+    'CollectionObject',
+    collectionObjectlId
+  );
+
+  const getBaseCollectionObject = () =>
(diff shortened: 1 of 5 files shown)


CarolineDenis (approved):

specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
```diff
@@ -1,194 +1,144 @@
...
-      key: 'br-uniqueness-permitnumber',
-      valid: false,
-      reason: 'Value must be unique to Database',
+  const getBaseCollectionObject = () =>
+    new schema.models.CollectionObject.Resource({
+      id: collectionObjectlId,
+      resource_uri: collectionObjectUrl,
     });
+
+  beforeEach(() => {
```
maxxxxxdlp, line 120: should use `beforeAll` instead. although that doesn't make a difference in this case, `beforeEach` should be used when an action needs to be repeated for each test (i.e setup or teardown that needs to be reset for each test - possibly because the test might have modified something)

also, don't forget to use `afterAll` to change the value back afterward to original value (or optionally modify the mocked API response in the .json file, and see if it breaks any other tests)

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -73,36 +73,41 @@ export const nonUniqueBusinessRuleDefs: MappedBusinessRuleDefs = {
...
-        if (resolved && quantity && resolved > quantity) {
-          newVal = quantity;
-        }
-        if (resolved && returned && resolved < returned) {
-          newVal = returned;
-        }
 
-        if (typeof newVal === 'number')
-          borrowMaterial.set('quantityResolved', newVal);
+        const adjustedResolved =
```
maxxxxxdlp, line 98: (optional) this function and the one before it are very similar. if it's not too complicated, see if parts of it or entire thing can be abstracted away into a function

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -72,7 +74,7 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
       this.fieldChangePromises[fieldName as string].resolve('superseded');
     this.fieldChangePromises[fieldName as string] = thisCheck;
 
-    const checks = [
+    const checks: Array<Promise<BusinessRuleResult<SCHEMA> | undefined>> = [
```
maxxxxxdlp, line 77: for consistency, we don't use `Array` and `ReadonlyArray` anywhere in the codebase. Instead, we use `T[]` and `RA<T>`. feel free to change the convention, as long as you make all usages consistent

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -84,45 +86,46 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
...
-            result.reason
-          );
-        } else {
-          this.resource.saveBlockers!.remove(result.key);
-          if (typeof result.action === 'function') {
-            result.action();
+  ): RA<BusinessRuleResult<SCHEMA>> {
+    return filterArray(
+      results.map((result) => {
+        if (result !== undefined) {
```
maxxxxxdlp, line 102: (optional) I know that this code was this way before, but you could make it more readable if you want

instead of 

```ts
if(result === undefined) { ... }
return undefined;
```
do

```ts
if (result === undefined) return undefined;
// ...
```

this is a matter of preference, but I prefer to return as soon as possible, as that makes code less indented (this currently looks like a [pyramid of doom](https://en.wikipedia.org/wiki/Pyramid_of_doom_(programming)), with 3 nested if statements
plus, this gets rid of a negated condition (negation adds needless cognitive overhead)

maxxxxxdlp (approved):
nothing major, just code smell issues

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
melton-jason: Yes, a guard clause here would look sooo much cleaner!
Fixed in https://github.com/specify/specify7/pull/3792/commits/7a9d0b79016dad7adeb95f333e41479290935b13

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
melton-jason: Fixed in https://github.com/specify/specify7/pull/3792/commits/7a9d0b79016dad7adeb95f333e41479290935b13

specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
melton-jason: Fixed in https://github.com/specify/specify7/pull/3792/commits/7a9d0b79016dad7adeb95f333e41479290935b13

specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
```diff
@@ -117,10 +117,14 @@ describe('Collection Object business rules', () => {
       resource_uri: collectionObjectUrl,
     });
 
-  beforeEach(() => {
+  beforeAll(() => {
     overwriteReadOnly(schema, 'embeddedCollectingEvent', true);
   });
 
+  afterAll(() => {
+    overwriteReadOnly(schema, 'embeddedCollectingEvent', false);
```
maxxxxxdlp, line 125: can you check if jest would let you put

```ts
const originalEmbeddedCollectingEvent = schema.embeddedCollectingEvent
```

right before the `beforeAll(() => {` line and use that variable instead of hardcoding `false`?

if not, then this is ... fine

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
melton-jason: Fixed in 9a4ded6572ff7e1603638f6c61fabda74fb91daf

melton-jason:
@specify/dev-testing, can this be merged into v7.9-dev? 
The business rule tests have all been fixed and are running correctly. 
The remaining failing test I am pretty sure is related to Jest's setup/teardown, and i've yet to find a good solution. 
Increasing Jest's timeout for tests made them all run correctly, but that isn't a proper fix.

maxxxxxdlp:
@melton-jason I was debuggin this. Took 1.5hrs((. Here is the log of my debugging:

I ran `npm t`. There were several errors like this:

Screenshot 2023-07-27 at 19 55 45

they are saying that after the jest environment is torn down (because tests have finished running), some async code is still executing, and still trying to make requests/import ajaxMock function/access mocked DOM methods (document.location)

my hunch: modifying a resource, triggers a business rule, which makes a back-end request (likely the uniqueness check)

---

debugging...

you can always add

```ts
console.warn('MAKING REQUEST:', url, requestMethod);
```

here:

https://github.com/specify/specify7/blob/b5208a4d03a82043de47882048994f89d1bdc50c/specifyweb/frontend/js_src/lib/tests/ajax/index.ts#L94-L96

to figure out what requests are being (mocked) made

wasn't helpful here as there were too many requests

---

after seeing a message like this:
Screenshot 2023-07-27 at 20 06 28

I tried to isolate the error

running
```sh
npx jest lib/components/QueryBuilder/__tests__/fromTree.test.ts
```

error reproducible.

modifying `fromTree.test.ts` to only run the test on the Geography tree and error still reproducible.

---

Now let's look at what's going on in `businessRules.ts` ...

...going over all the places where async functions are called and checking how the promise is handled...

This is very suspicious:
https://github.com/specify/specify7/blob/b5208a4d03a82043de47882048994f89d1bdc50c/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts#L155-L155
- It sets an event listener, but never removes it?
- It calls an async function, but never listens to the result

Though, worryingly, that code has been there for quite a while. Added proper calls to `this.addPromise()`


Next, made a temporary change to the code:
Screenshot 2023-07-27 at 21 00 44

(`resources` is a global store containing all initialized resources - not for production use because of memory leaks)
now I am able to check if any resources have pending promises

Screenshot 2023-07-27 at 20 55 16

oh yeah, 13 resources with promises as a result of `requireContext()`:

Screenshot 2023-07-27 at 20 55 49

Tried to modify `requireContext()` to wait for all business rules to settle

Screenshot 2023-07-27 at 21 01 27

but that never resolves:

Screenshot 2023-07-27 at 21 02 05

could be because of reverse promises. could be because of a bug in my quick modifications

either way, some takeaways:
- This issue would affect not just `requireContext()`, but any test that makes changes to resources (resourceApi.test.ts, collectionApi.test.ts, businessRules.test.ts more prominently, but many tests are creating and modifying resources too)
- Not sure why this wasn't an issue before because the code looks quite fanky - i.e this should never have worked 😊- the tests were never waiting for business rules to finish. It's possible that 

Potential solutions:
- When creating resources, the "noBussinessRules" prop could default to true if running in test mode. Then if you want to explicitly test business rule, explicitly set `noBussinessRules: true` when creating a resource. Pros: simple. Cons: behavior in tests would not match runtime
- Create a util function and make sure to call it at the end of the test that modified resources. The util could either return a promise that resolves once all business rules are completed (pros: tests business rules, cons: would slow down tests) OR abort all pending business rules and returns a promise once all aborted (pros: doesn't slow down tests as much? cons: need to create a mechanism for aborting, doesn't test business rules). Another big con: easy to forget to wait for business rules to finish.
- A modification on a previous one: just like in `requireContext()`, we call `beforeAll()` to run code before all tests, we could also call `afterAll()`, which would take care of pending business rules (again, wait for them to complete, or abort them). Pros: all tests that work with business rules call `requireContext()`, thus no way to accidentally forget to cleanup business rules. Cons: similar to previous solution - either slows down tests, or doesn't run the code like it would run in the real world
- A solution to mitigate the slow down of tests, to be paired with one of the previous solutions: evaluate possibility of disabling business rules where safe (i.e some resources created by `requireContext()` are never modified, so business rules shouldn't be running, right?). OR, make business rules smarter and more efficient (see the "Also, this seems very inefficient:" comment below for one of the suggestions). Pros: will keep the behavior of code in tests closer aligned to behavior outside of tests. Cons: probably the most difficult solution as optimizing code and benchmarking code takes longer than just applying a hack-bandaid.
- A solution to be paired with disabling business rules when running in tests: add extra comprehensive tests for business rules, to be completely sure that everything is bug free. And then disable business rules when running all other tests. In effect, business rules would ever be running in the `businessRules.test.ts` file. I kind of like this solution, as it would speed up tests, improve business rule coverage, and resolve this jest error.

In summary:
sorry, wasn't able to solve this, and not fully sure why this didn't occur before (maybe the business rule refactor made more uniqueness rules run than before), but I proposed several solutions, and hopefully you can use them to resolve the issue.

not sure if you would have time to do that before release though.

Pushed the code changes I did while debugging this: https://github.com/specify/specify7/compare/fix-busrule-tests...debugging-business-rules?expand=1

please delete that branch after you are done to not leave useless branches lying arround

---

Also, this seems very inefficient:
```
  this.resource.on('change', this.changed, this);

  // ...

  private changed(resource: SpecifyResource<SCHEMA>): void {
    if (resource.isBeingInitialized && typeof resource.changed === 'object') {
      Object.keys(resource.changed).forEach((field) => {
        this.addPromise(this.checkField(field));
      });
    }
  }
```

if I read it correctly, on any change to a resource, all of it's changed fields will undergo business rule checks - is it safe to only re-check business rules for the field that triggered the change event?


Btw, look at this commit: https://github.com/specify/specify7/commit/3785fb8f98b924303906a08abff0df208b2eebb3
Very bad - does 4 unrelated, big, important changes. should have been 4+ separate smaller commits

---

melton-jason:
Wow! 
Thank you for spending the time to look into this. 
I REALLY appreciate the explanation behind the debugging process. 

> not fully sure why this didn't occur before (maybe the business rule refactor made more uniqueness rules run than before), but I proposed several solutions, and hopefully you can use them to resolve the issue.

I'm not entirely sure why this didn't happen either. It should be very close to identical the logic in the old Javascript file. Nevertheless, I'm glad you uncovered this issue with the tests and gave us some avenues to explore.

>  is it safe to only re-check business rules for the field that triggered the change event? 

Ideally, we should only check the fields in two scenarios: 
- Check fields with values when the resource is loaded/initialized
  - We don't currently do this, and this leads to potential problems when the backend error dialog appears because the frontend never checked the faulty field. Of course, any solution for this would need to take performance into consideration
- Check a field when it is changed
 
Right now, it functionally does just check the field that was changed, and nothing more. 
The `changed` object on a resource does not keep a running list of every attribute that has been modified since the resource was saved. Instead, it only holds the attributes that were changed in the last `set()` call. 
(At least, this is what I have seen using Chrome Developer Tools). 
I would need to look more into Backbone to determine if we can pass the field or field name along with an event, and then only check that specific field. 

Once again, another reason to migrate off of Backbone ASAP :face_exhaling: 


(I've never heard of the word 'fanky' before. I might steal that :joy: )

maxxxxxdlp:
ah, thanks! I didn't realize that about .changed!

> I've never heard of the word 'fanky' before. I might steal that

LOL. This is funny because I am pretty sure I learned this word from Ben. Passed on the torch of Specify down the generations 🤣

melton-jason:
The issue here was with `rgetPromise` in a testing environment. 
More specifically: https://github.com/specify/specify7/blob/55f42c34021b9e429487275358dd7bcd8335c94a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts#L318-L319

When initializing the resources in a test, there were two cases in which the resource as a result of `rgetPromise` did not have a static file or a corresponding `overrideAjax` (See 71411e44e44045f24a42b29aa46555166592a49f).
Normally, this would  lead to a warning, but it strangely did not in this case: the promise simply never resolved. 

Utilizing another BackBone method (`getRelated` in 31578c0520d563437ee0697cb443d142ea03fa77), restored the warning and allowed me to locate and add the proper overrideAjax calls. 
This is _very_ concerning because it implies there might be some fault in `rgetPromise`, a function which is used extensively in the frontend. Perhaps the promise is not going through our frontend `ajax` call and instead is being handled directly by BackBone. Further investigation will be needed. 

There are now only two remaining failing tests: both of which should be fixed in #3963

grantfitzsimmons:
im so excited!

acwhite211 (approved):

acwhite211:
nice job @melton-jason 7.9 is looking good!

maxxxxxdlp:
Great job debugging and fixing this!
yes, quite worrying.
part of the problem could be that backbone does not use native promises internally (it's too old for that - another reason to migrate off it as soon as possible) - 2 years ago I refactored resourcesApi.js a bit to use native promises (was using jquery loadable before), but since old school promises are still used internally, issues with errors not being reported could arise

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -172,21 +171,6 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
...
-        .filter((result) => result !== undefined)
-        .forEach((duplicate: SpecifyResource<SCHEMA> | undefined) => {
-          if (duplicate === undefined) return;
-          const event = duplicate.cid + ':' + (fieldName as string);
-          if (!this.watchers[event]) {
-            this.watchers[event] = () =>
-              duplicate.on('change remove', () => this.checkField(fieldName));
-          }
-        });
-    });
```
maxxxxxdlp, line 189: a lot of (potentially problematic) code got removed here. just double checking that that was intentional

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -84,45 +86,46 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
...
       );
-
-    Promise.all(checks)
-      .then((results) => {
-        return thisCheck === this.fieldChangePromises[fieldName as string]
-          ? this.processCheckFieldResults(fieldName, results)
-          : undefined;
-      })
-      .then(() => thisCheck.resolve('finished'));
+    return Promise.all(checks).then((results) =>
```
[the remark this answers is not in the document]
grantfitzsimmons: @realVinayak told us so

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
melton-jason: Yes this was intentional. 
But in hindsight I realize that removing this created a bug. 

Fixed in https://github.com/specify/specify7/pull/3976/files/55cdae94cd69a21d7326c92060c04d8e7e9f946e..b94e85ae2ae919174ab854ec1a27e5964d1d2dc5

PR #3800, opened 2023-07-18
Dissable browse to form when no results

### LINKED ISSUE #3796: Query builder from QCB doesn't disable Browse in Forms when 0 results returned
**Describe the bug**
The query builder dialog that's accessed from a query combo box allows to select `Browse in Forms` even when 0 results are returned. Upon click an infinite loading dialog begins. 

**To Reproduce**
1. Go to record with query combo box
2. Click on magnifying glass icon to the right
3. Click on `Query Builder`
4. Run a query that returns no results
5. Select `Browse in Forms`
6. See error

**Screenshots**

https://github.com/specify/specify7/assets/135047322/331885a0-cdec-4c0b-9c28-60419bdd99c8

https://github.com/specify/specify7/assets/135047322/1287d789-8f2d-448a-b778-81ce042f8980

**Desktop:**
 - macOS, Chrome, `v7.9-dev`
 
**Database Name:** sdnhmpaleo

CarolineDenis:
Fixes #3796

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
@@ -202,7 +202,7 @@ export function QueryResults(props: Props): JSX.Element {
         Array.isArray(results) &&
         Array.isArray(loadedResults) &&
         results.length > 0 &&
-        typeof fetchResults === 'function' && totalCount !== 0 ? (
+        typeof fetchResults === 'function'? (
           <>
             {hasPermission('/record/replace', 'update') &&
               hasTablePermission(model.name, 'update') && (
@@ -413,7 +413,9 @@ export function useFetchQueryResults({
     async (index?: number): Promise<RA<QueryResultRow> | void> => {
       const currentResults = resultsRef.current;
       const canFetch = Array.isArray(currentResults);
+
       if (!canFetch || fetchResults === undefined) return undefined;
+
       const alreadyFetched =
         currentResults.length === totalCount &&
         !currentResults.includes(undefined);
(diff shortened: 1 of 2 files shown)


maxxxxxdlp:
Wow, this pull request has 0 code changes - I didn't know that was possible!

CarolineDenis:
@bronwyncombs or @carlosmbe could yo test this to see if it's satisfying? 
@melton-jason what do you think of adding that condition?

bronwyncombs (approved):
Working for me! `Browse in Forms` option is not available for queries without results. Tested multiple tables and fields. Tested embedded query and query builder.

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -202,7 +202,7 @@ export function QueryResults(props: Props): JSX.Element {
         Array.isArray(results) &&
         Array.isArray(loadedResults) &&
         results.length > 0 &&
-        typeof fetchResults === 'function' ? (
+        typeof fetchResults === 'function' && totalCount !== 0 ? (
```
maxxxxxdlp, line 205: how is `totalCount !== 0` different from the `results.length > 0` condition that was already here (one line above)?

only one of those should be needed, unless something fishy is going on here - can you investigate? @CarolineDenis

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
CarolineDenis: because sometimes results = [undefined] instead of []

CarolineDenis: I'm looking into it now... 
handleSetResults is being with results=undefined and it shouldn't be called at all

CarolineDenis: seems to work every time

maxxxxxdlp: good catch - thanks for resolving this issue the proper way rather than adding a hack

maxxxxxdlp (approved):
thanks for resolving this issue the proper way rather than adding a hack

PR #3806, opened 2023-07-19
Automatically add ordinal to attachments

### LINKED ISSUE #3788: 7.9 Can Not Upload Attachments `IntegrityError: Ordinal Value`
**Describe the bug**
Grant believes this to be an Issue caused by Business Rules.

Not re-creditable in older versions. 

Whenever I try saving a new Record with an Attachment. I get an ordinal IntegrityError.

**To Reproduce**
1. Go to Data Entry for any Table
2. Enter Required Information 
3. Add attachment 
4. Save
5. See error

You could also add a Record with an Attachment to a Record Set

**Expected behavior**
Not a crash

**Beautiful Pictures**

https://github.com/specify/specify7/assets/53784701/23e948c7-dc74-423a-aa7d-ef033ac4f623

[Specify 7 Crash Report - 2023-07-14T18_51_57.710Z.txt](https://github.com/specify/specify7/files/12053192/Specify.7.Crash.Report.-.2023-07-14T18_51_57.710Z.txt)

**Desktop:**
 - OS: Mac Os
 - Browser: Chrome
 - Specify 7 Version: 7.9

**Database Name:** morpaleo 2 may

melton-jason:
Fixes #3788

This adds a backend business rule which sets the `ordinal` field of Attachment tables if `ordinal` is not provided. 
It does so via the following: 
- Get all of the other Attachments on the related record
- Gets the maximum `ordinal` value among those attachments
- Sets the new Attachment equal to 1 + the maximum, or 0 if there are no other attachments

This also adds a frontend business rule which sets the `ordinal` and `ordernumber` fields equal to the index of wherever they are in the collection when they're added. 
In other words, if I you add the 4th attachment to a CollectionObject, its ordinal field should be 3. 

TO TEST:
Try uploading Attachments to models that support Attachments.

For convenience, here's a list of all attachment tables (extracted from MariaDB):

```sql
MariaDB [specify]> show tables like "%_attachment";
+----------------------------------+
| Tables_in_specify (%_attachment) |
+----------------------------------+
| accessionattachment              |
| agentattachment                  |
| borrowattachment                 |
| collectingeventattachment        |
| collectingtripattachment         |
| collectionobjectattachment       |
| conservdescriptionattachment     |
| conserveventattachment           |
| deaccessionattachment            |
| disposalattachment               |
| dnasequenceattachment            |
| dnasequencerunattachment         |
| exchangeinattachment             |
| exchangeoutattachment            |
| fieldnotebookattachment          |
| fieldnotebookpageattachment      |
| fieldnotebookpagesetattachment   |
| giftattachment                   |
| loanattachment                   |
| localityattachment               |
| permitattachment                 |
| preparationattachment            |
| referenceworkattachment          |
| repositoryagreementattachment    |
| storageattachment                |
| taxonattachment                  |
| treatmenteventattachment         |
+----------------------------------+
```

### DIFF
--- a/specifyweb/businessrules/attachment_rules.py
+++ b/specifyweb/businessrules/attachment_rules.py
@@ -1,5 +1,7 @@
 import re
 
+from django.db.models import Max
+
 from .orm_signal_handler import orm_signal_handler
 from specifyweb.specify.scoping import Scoping
 from specifyweb.specify import models
@@ -24,6 +26,12 @@ def attachment_jointable_save(sender, obj):
     obj.attachment.tableid = attachee.specify_model.tableId
     obj.attachment.scopetype, obj.attachment.scopeid = Scoping(attachee)()
     obj.attachment.save()
+
+    if obj.id is None:
+        if obj.ordinal is None:
+            others = sender.objects.filter(**{attachee.specify_model.name.lower(): attachee})
+            top = others.aggregate(Max('ordinal'))['ordinal__max']
+            obj.ordinal = 0 if top is None else top + 1
 
 @orm_signal_handler('post_delete')
 def attachment_jointable_deletion(sender, obj):
--- a/specifyweb/businessrules/attachment_rules.py
+++ b/specifyweb/businessrules/attachment_rules.py
@@ -27,11 +27,6 @@ def attachment_jointable_save(sender, obj):
     obj.attachment.scopetype, obj.attachment.scopeid = Scoping(attachee)()
     obj.attachment.save()
 
-    if obj.id is None and obj.ordinal is None:
-        others = sender.objects.filter(**{attachee.specify_model.name.lower(): attachee})
-        top = others.aggregate(Max('ordinal'))['ordinal__max']
-        obj.ordinal = 0 if top is None else top + 1
-
 @orm_signal_handler('post_delete')
 def attachment_jointable_deletion(sender, obj):
     if sender in attachment_tables:
--- a/specifyweb/businessrules/attachment_rules.py
+++ b/specifyweb/businessrules/attachment_rules.py
@@ -27,11 +27,10 @@ def attachment_jointable_save(sender, obj):
     obj.attachment.scopetype, obj.attachment.scopeid = Scoping(attachee)()
     obj.attachment.save()
 
-    if obj.id is None:
-        if obj.ordinal is None:
-            others = sender.objects.filter(**{attachee.specify_model.name.lower(): attachee})
-            top = others.aggregate(Max('ordinal'))['ordinal__max']
-            obj.ordinal = 0 if top is None else top + 1
+    if obj.id is None and obj.ordinal is None:
+        others = sender.objects.filter(**{attachee.specify_model.name.lower(): attachee})
+        top = others.aggregate(Max('ordinal'))['ordinal__max']
+        obj.ordinal = 0 if top is None else top + 1
 
 @orm_signal_handler('post_delete')
 def attachment_jointable_deletion(sender, obj):
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
@@ -49,6 +49,10 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
     resource: SpecifyResource<SCHEMA>,
     collection: Collection<SCHEMA>
   ) {
+    /**
+     * REFACTOR: remove the need for this and the orderNumber check by
+     * implementing a general solution on the backend
+     */
     if (resource.specifyModel.getField('ordinal') !== undefined)
       (resource as SpecifyResource<CollectionObjectAttachment>).set(
         'ordinal',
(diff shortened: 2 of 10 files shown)


carlosmbe (approved):
It works.

Not sure if additional context helps but the Ordinal Value Error was never caused by uploading a new background via Preferences.

Tested in NHMD

bronwyncombs (approved):
Working!

Screenshot 2023-07-19 at 2 47 16 PM

Screenshot 2023-07-19 at 2 46 50 PM

Screenshot 2023-07-19 at 2 45 26 PM

bronwyncombs (changes requested):
Actually, the section for attachments in preferences is no longer there in this branch 😮 

https://github.com/specify/specify7/assets/135047322/4570065e-d1aa-4a8e-a952-96cf365fc19d

bronwyncombs (approved):
Preferences issue resolved! 5 gold stars for @melton-jason

grantfitzsimmons (approved):
Here's my unrequested review:

Works great.

grantfitzsimmons:
@melton-jason We can merge as long as you fix front-end tests :wink:

specifyweb/businessrules/attachment_rules.py
```diff
@@ -25,6 +27,12 @@ def attachment_jointable_save(sender, obj):
     obj.attachment.scopetype, obj.attachment.scopeid = Scoping(attachee)()
     obj.attachment.save()
 
+    if obj.id is None:
```
maxxxxxdlp, line 30: combine these two nested `if`s into a single one?

specifyweb/businessrules/attachment_rules.py
maxxxxxdlp, line 1: take a look at this:

https://github.com/specify/specify7/blob/22c79e9e4be7ef456c0a7c53535b146333ae427f/specifyweb/specify/build_models.py#L11-L25

it's possible that your code modifications are not necessary

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -44,6 +45,28 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
     }
   }
 
+  private added(
+    resource: SpecifyResource<SCHEMA>,
+    collection: Collection<SCHEMA>
+  ) {
+    if (resource.specifyModel.getField('ordinal') !== undefined)
```
maxxxxxdlp, line 52: this is a bit of a code smell. can you double check that?
- the code on the back-end does not make this code on the front-end redundant
- there is no simple way to accomplish this via the business rules definitions .json file that you created so that this is defined once and used both by back-end and front-end

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
melton-jason: > the code on the back-end does not make this code on the front-end redundant
Yes, the only reason I added the code to the frontend was because that's what was causing the original issue. 

I figured also having the code on the frontend would alleviate some other potential bugs because the frontend changes are more general. 
I can remove either the frontend or backend changes to cut down on redundancy

maxxxxxdlp (approved):
Don't want to be blocking this for no good reason.
Take a look at my comments, and if they are not valid, go ahead and merge this

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
maxxxxxdlp: as you see best
you know the business rules code better than me

melton-jason: That would also allow me to remove a lot of redundant code on the backend... 
There are multiple cases where we set the ordinal/ordernumber property if it doesn't exist. Might be nice to remove them

specifyweb/businessrules/attachment_rules.py
melton-jason: I was primarily following how the other business rules were structured, to stay consistent. 
We should run the backend files through PEP 8 soon...

specifyweb/businessrules/attachment_rules.py
melton-jason: Unfortunately, it's never that easy :smile: 
This list is here to set the `ordering` property within the Meta class of the Django object representation of a table. 

Here's where it's used:
https://github.com/specify/specify7/blob/be7468cf1e922fb809ad545187bb6d2cd088bde1/specifyweb/specify/build_models.py#L53-L57

Which just tells Django how the records should be ordered in a QuerySet. 

Here's the docs:
https://docs.djangoproject.com/en/4.2/ref/models/options/#ordering

grantfitzsimmons:
@melton-jason Please request my review if you'd like me to take another look :smile:

melton-jason:
@grantfitzsimmons feel free to test this again

bronwyncombs (approved):
Attachment uploads working! 👍 
Tested on `sp7demofish` and `coldfish` databases

PR #3809, opened 2023-07-19
let the exception pass through if not duplicate error

acwhite211:
This small change to agent merging will allow the exception pass through to be handled by the atomic.transaction if the error is not a duplicate error.


CarolineDenis (approved):

melton-jason (approved):
Nice catch! 
(Also see https://stackoverflow.com/a/42588043 and the Django docs at https://docs.djangoproject.com/en/3.2/topics/db/transactions/#django.db.transaction.atomic)

grantfitzsimmons (approved):
as @melton-jason said, nice catch (x2)!!!

PR #3819, opened 2023-07-20
Remove padding from detailed view

### LINKED ISSUE #3815: Extra padding in query items is unnecessary
image

Can we condense the buttons down once again so that they are not squares?

image

You can see that a considerable amount of space is used without any benefit

CarolineDenis:
Fixes #3815


carlosmbe (approved):
It removes the padding and it works. 

Any complaints may be minor stylistic choices that vary from person to person, but I personally think it's fine

grantfitzsimmons (approved):
Works as expected. Looks good!

melton-jason (approved):
A nice, simple Tailwind change :partying_face: 
We need more of these! :joy:

Issue #3801, opened 2023-07-18
Agent Merging - Integrity Error


carlosmbe:
**Describe the bug**
This Phantom error initially found by Jason in the development of Agent Merging has been resurrected and is in 7.9.

It's identifiable by the red line

https://github.com/specify/specify7/pull/3474#issuecomment-1552195207

**To Reproduce**
1. Go to NHMD -> Amber -> Queries
2. Click on Agent Semantics
3. Merge the Agents I listed below

Agents to Merge:
- Scharff, Nikolaj
- Hormiga, G
- Newton, Alfred

**Screenshots**
<img width="1792" alt="Screenshot 2023-07-18 at 3 23 29 PM" src="https://github.com/specify/specify7/assets/53784701/1178e5f4-db00-47b7-9958-581456ba9a25">

**Desktop:**
 - OS: Mac
 - Browser: Chrome
 - Specify 7 Version: 7.9

melton-jason:
As shown in https://github.com/specify/specify7/pull/3474, there are two issues here: one with what's causing the error (https://github.com/specify/specify7/pull/3474#issuecomment-1552195207), and how the error is displayed (https://github.com/specify/specify7/pull/3474#issuecomment-1554797350). 

The underlying cause of the issue here is all about the references to the agents.

To minimally recreate the issue: 
- Create two Agents
- Create a Collecting Event with both Agents as Collectors
- Create another Collecting Event with one of the Agents as a Collector
- Try and merge the first two agents together
- See the error

realVinayak:
This should be fixed though in https://github.com/specify/specify7/pull/3846. Let's keep it open, and then close as resolved when testers test this

grantfitzsimmons:
# To close this issue, test the following:

---

Testing instructions:

For Issue https://github.com/specify/specify7/issues/3801

I'll recommend running the below steps for at least some of Deaccession, Disposal, Loan, Accession, Gift, Borrow tables. Also test on both v7.9-dev and this branch. v7.9-dev shouldn't be able to merge these, and this branch should.

Later on, testers can go wild and merge anything into anything - and as long as it is not an actual business rule exception, it should work. 

Steps:
1. Make an interaction via the interactions form. I'll call it accession `001` for these steps. Make two new accession agents. I'll recommend creating two fresh new agents (but you can select a previous agent too). You can also make more than two accession agents. For this example, two new agents are `agent 001` and `agent 002`. **Important** that both of these agents should be assigned same role in the accession.
[...]

PR #3823, opened 2023-07-21
Properly display error message in Merging Dialog

melton-jason:
Addresses the frontend component of #3801 

The issue was that `http.HttpResponseNotAllowed` expected the first argument to be a list of permitted methods for the endpoint. 
The frontend was expecting a response with code 405, and was not going to construct the usual error message because we had said we would handle the error:  https://github.com/specify/specify7/blob/76878df9e9fa53a3e94b753ec7d11f19f9a9d6b6/specifyweb/frontend/js_src/lib/components/Merging/index.tsx#L233

The data for `http.HttpResponseNotAllowed` would have to be included as another argument, once the list of permitted methods was included. Thus the backend was not passing any data to the frontend. The proper way of using the response would be something like: 
```python
http.HttpResponseNotAllowed(["POST"], 'Error! Multiple records violating uniqueness constraints in ' + table_name)
```

https://docs.djangoproject.com/en/4.2/ref/request-response/#django.http.HttpResponseNotAllowed

TO TEST: 
- Follow the steps in https://github.com/specify/specify7/issues/3801#issuecomment-1640975722

### DIFF
--- a/specifyweb/specify/views.py
+++ b/specifyweb/specify/views.py
@@ -478,7 +478,7 @@ def record_merge_fx(model_name: str, old_model_ids: List[int], new_model_id: int
                     if foreign_record_lst.count() > 1:
                         # NOTE: Maybe try handling multiple possible row that are potentially causes the conflict.
                         # Would have to go through all constraints and check records based on columns in each constraint. 
-                        return http.HttpResponseNotAllowed('Error! Multiple records violating uniqueness constraints in ' + table_name)
+                        return http.HttpResponse('Error! Multiple records violating uniqueness constraints in ' + table_name, status=405)
 
                     old_record = obj
                     new_record = foreign_record_lst.first()


carlosmbe (changes requested):
Did not work

[Uncensored Version](https://drive.google.com/file/d/1md-4KU6SJe62P7U7zsA9sMMntljGpIYj/view?usp=share_link)

Screenshot 2023-07-21 at 4 13 33 PM

bronwyncombs (changes requested):
When the message appears it changes the names to the new merged record name. This is also in `v7.9-dev`, but only with this specific warning so if it's simple enough I hope this can be fixed in this PR. This is a temporary view and the names revert after canceling. It does work for me otherwise!


https://github.com/specify/specify7/assets/135047322/80e76129-252c-4834-ac8c-8998e9b5d51d

https://github.com/specify/specify7/assets/135047322/549641bd-76f7-463c-a89c-64b856bb5948

`v7.9-dev`: 

https://github.com/specify/specify7/assets/135047322/b61dcd3d-62bf-4a6e-a407-d347ed942f5c

Doesn't happen with 3+ agents: 

https://github.com/specify/specify7/assets/135047322/c19ab5e7-0789-4ea6-8987-7bc8eb0c6a7d

PR #3825, opened 2023-07-21
Display merge button in meta only for agent

### LINKED ISSUE #3824: "Merge Records" button shows up in the meta menu for any table
image

Tested on Collection Object and Locality.

The Office

CarolineDenis:
Fixes #3824 3824


carlosmbe (commented):
Based on first impressions it seems solid. Want to try something before approving

carlosmbe (approved):

specifyweb/frontend/js_src/lib/components/FormMeta/index.tsx
```diff
@@ -179,7 +179,9 @@ function MetaDialog({
                   action="update"
                   tableName={resource.specifyModel.name}
                 >
-                  <MergeRecord resource={resource} />
+                  {resource.specifyModel.name === 'Agent' && (
```
melton-jason, line 182: Instead of hardcoding the table name here, can you instead export and use the `recordMergingTables` Set in `*/lib/components/Merging/index.tsx`?

https://github.com/specify/specify7/blob/3bbc8378565eb66e9bdfa432d75e8ca6fb798339/specifyweb/frontend/js_src/lib/components/Merging/index.tsx#L42

That way, we can just add tables to this set and merging functionality will be enabled without having to find every place a table name is hardcoded for record merging. 

See this line for how to accomplish the logic:
https://github.com/specify/specify7/blob/3bbc8378565eb66e9bdfa432d75e8ca6fb798339/specifyweb/frontend/js_src/lib/components/Merging/index.tsx#L84

melton-jason (changes requested):

melton-jason (approved):

PR #3827, opened 2023-07-21
Correct form padding and round corners in Firefox

### LINKED ISSUE #3189: App resources corners are not rounded symmetrically
image

image

grantfitzsimmons:
Fixes #3189

**Instructions to test:**
- Make sure that all areas with scrollbars display properly in Firefox and Chrome.
- Make sure that corners on elements with scrollbars are still rounded
- Make sure that padding is equal on both the left and right on forms

**It shouldn't look like this:**

image

image


melton-jason (approved):
I would heavily defer the decision to say this is a good change or not to @specify/ux-testing:
It is fine as far as the code is concerned, though maybe `[scrollbar-gutter:auto]` might not be needed because it is the initial, default value. Need to look into this more, but it's not a _bad_ decision to be declarative. 
Maybe an approach like that taken in [this StackOverflow answer](https://stackoverflow.com/a/76029551) may be worth considering as well

I can describe what these changes _do_, to give @specify/ux-testing a better idea on what to test. 

[scrollbar-gutter](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) is a CSS property used to control how space is reserved for the scroll-bar. 
We changed it from `stable` to `auto`. 
Stable made it so that the space for the scroll-bar should have _always_ been reserved, even if a scroll-bar is not needed. 
See if these changes look okay on various screen widths: there should not be a drastic shifting of contents to accomodate a scroll that is suddenly needed

bronwyncombs (approved):
Looks good 👍 

https://github.com/specify/specify7/assets/135047322/92143d3c-557d-442e-aa6c-6b9e8f152c52

https://github.com/specify/specify7/assets/135047322/921fe59a-12a4-4d2f-833d-7229b42e08d4

maxxxxxdlp (changes requested):
`scrollbar-gutter:auto` is the [default value](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter), so you can just remove that class name

---

scrollbar-gutter was useful at one point to reduce needless UI shifts (expand/collapse something inside the container or go between records), but if rounded corners are a bigger issue then ok

maxxxxxdlp:
Also, perhaps `overflow:auto` would be a better solution - it would make scrollbars display on top of content in some modern browsers, thus removing the UI shift issue, and removing the persistent scroll bars issue

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#:~:text=a%20scroll%20container.-,auto,drawn%20on%20top%20of%20the%20content%20instead%20of%20taking%20up%20space.,-Description

maxxxxxdlp (approved):

PR #3831, opened 2023-07-24
Keep original data for records after merge failing

### LINKED ISSUE #3822: Merge records dialog changes formatted record data visually where no change is happening at the database level
https://github.com/specify/specify7/assets/37256050/a57f9a8b-04d2-420c-afa4-2f923f49bf12

Once an error occurs, the values in the merge records dialog are not only wrong, but make it even look like they are canonical when you preview the records.

CarolineDenis:
Fixes #3822

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Merging/Compare.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/Compare.tsx
@@ -16,7 +16,6 @@ import { MergingHeader } from './Header';
 
 export function CompareRecords({
   formId,
-  needUpdate,
   model,
   records,
   merged,
@@ -24,7 +23,6 @@ export function CompareRecords({
   onDismiss: handleDismiss,
 }: {
   readonly formId: string;
-  readonly needUpdate?: boolean;
   readonly model: SpecifyModel;
   readonly records: RA<SerializedResource<AnySchema>>;
   readonly merged: SpecifyResource<AnySchema>;
@@ -36,7 +34,7 @@ export function CompareRecords({
 }): JSX.Element {
   const resources = React.useMemo(
     () => records.map(deserializeResource),
-    [records, needUpdate]
+    [records]
   );
   const conformation = useMergeConformation(model, resources);
   return (
--- a/specifyweb/frontend/js_src/lib/components/Merging/Compare.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/Compare.tsx
@@ -16,6 +16,7 @@ import { MergingHeader } from './Header';
 
 export function CompareRecords({
   formId,
+  needUpdate,
   model,
   records,
   merged,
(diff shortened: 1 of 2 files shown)


maxxxxxdlp (changes requested):
hmm, your solution would work only once, right?

first time it will change the boolean from false to true - triggering the useMemo
second time it will change the boolean from true to true - not triggering the useMemo

perhaps a better option:

replace this line in the Merging component
```tsx
const records = useResources(model, ids);
```

with:
```tsx
const rawRecords = useResources(model, ids);
const [records,setRecords] = useTriggerState(rawRecords);
```

and then instead of
```tsx
setNeedUpdate(!needUpdate);
```

do:
```tsx
setRecords(Array.from(records))
```

that way there is no need to add a hacky prop (though, this solution is a bit hacky still)
@melton-jason what do you think?

of course, this issue will likely be fixed once we migrate off backbone...

CarolineDenis:
> hmm, your solution would work only once, right?
> 
> first time it will change the boolean from false to true - triggering the useMemo second time it will change the boolean from true to true - not triggering the useMemo
> 
> perhaps a better option:
> 
> replace this line in the Merging component
> 
> ```tsx
> const records = useResources(model, ids);
> ```
> 
> with:
> 
> ```tsx
> const rawRecords = useResources(model, ids);
> const [records,setRecords] = useTriggerState(rawRecords);
> ```
> 
> and then instead of
> 
> ```tsx
> setNeedUpdate(!needUpdate);
> ```
> 
> do:
> 
> ```tsx
> setRecords(Array.from(records))
> ```
> 
> that way there is no need to add a hacky prop (though, this solution is a bit hacky still) @melton-jason what do you think?
> 
> of course, this issue will likely be fixed once we migrate off backbone...

It works every time, the second time it will change from true to false. Always change to the opposite so triggers the memo. 
Will try your solution now :+1::skin-tone-2:

grantfitzsimmons (changes requested):
It seems to add create another dialog that adds another layer of gray over the previous, so the background becomes more obscured and the gray becomes more solid behind the dialog. Why is this happening?

maxxxxxdlp:
> It works every time

ah, good point!

sorry brain freeze. I seen `!needUpdate` and mentally translated to `false`

I don't like the need for the hack, and I don't like my proposed solution much better either, but neither would likely be needed once we migrate off backbone, so let it be

maxxxxxdlp (approved):

CarolineDenis:
@maxxxxxdlp, the issue @grantfitzsimmons is referring to happened when using the useTriggerState in index.tsx instead of my needUpdate prop. Are you ok to keep my original version? As you said it will be gone once we move out of backbone. 
See screen shot useTriggerState vs prop 

https://github.com/specify/specify7/assets/108160931/14048cd8-b89d-4258-8705-089a79d3fae0

https://github.com/specify/specify7/assets/108160931/944fdcc4-236c-4635-accf-4a3e7828c783

CarolineDenis:
> > It works every time
> 
> ah, good point!
> 
> sorry brain freeze. I seen `!needUpdate` and mentally translated to `false`
> 
> I don't like the need for the hack, and I don't like my proposed solution much better either, but neither would likely be needed once we migrate off backbone, so I am fine with either of those for now

Revert charges and went back to prop to avoid the issue Grant is referring to. (See previous comment)

bronwyncombs (approved):
Working! 

Screenshot 2023-07-27 at 10 34 41 AM

PR #3818, opened 2023-07-20
Add a warning dialog before merging

CarolineDenis:
Fixes #3816

PR #3832, opened 2023-07-24
Prevent merging agent with group members

realVinayak:
Prevent merging agents with group members

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
@@ -1,6 +1,6 @@
 import { RA } from '../../utils/types';
 import { AnySchema, SerializedResource } from '../DataModel/helperTypes';
-import { H2, Ul } from '../Atoms';
+import { Ul } from '../Atoms';
 import { FormattedResource } from '../Molecules/FormattedResource';
 import { deserializeResource } from '../DataModel/helpers';
 import React from 'react';
--- a/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
@@ -1,6 +1,6 @@
 import { RA } from '../../utils/types';
 import { AnySchema, SerializedResource } from '../DataModel/helperTypes';
-import { H3, Ul } from '../Atoms';
+import { Ul } from '../Atoms';
 import { FormattedResource } from '../Molecules/FormattedResource';
 import { deserializeResource } from '../DataModel/helpers';
 import React from 'react';
@@ -31,22 +31,20 @@ function InvalidMergeRecords({
       </h3>
       <p>{specificText}</p>
       <Ul className="gap-2">
-        {resources.map((resource, index) => {
-          return (
-            <li
-              key={(resource.id as number) ?? index}
-              className="flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
-            >
-              <TableIcon label={true} name={tableName} />
-              <FormattedResource
-                resource={React.useMemo(
-                  () => deserializeResource(resource),
-                  [resource]
-                )}
-              />
-            </li>
-          );
-        })}
+        {resources.map((resource, index) => (
+          <li
+            key={(resource.id as number) ?? index}
+            className="flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
+          >
+            <TableIcon label={true} name={tableName} />
+            <FormattedResource
+              resource={React.useMemo(
+                () => deserializeResource(resource),
+                [resource]
+              )}
+            />
+          </li>
+        ))}
       </Ul>
     </div>
   );
--- a/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
@@ -1,6 +1,6 @@
 import { RA } from '../../utils/types';
 import { AnySchema, SerializedResource } from '../DataModel/helperTypes';
-import { H2, Ul } from '../Atoms';
+import { H3, Ul } from '../Atoms';
 import { FormattedResource } from '../Molecules/FormattedResource';
 import { deserializeResource } from '../DataModel/helpers';
 import React from 'react';
@@ -13,6 +13,7 @@ import { commonText } from '../../localization/common';
 import { mergingText } from '../../localization/merging';
 import { Tables } from '../DataModel/types';
 import { TableIcon } from '../Molecules/TableIcon';
+import { className } from '../Atoms/className';
 
 function InvalidMergeRecords({
   resources,
@@ -27,7 +28,7 @@ function InvalidMergeRecords({
 }): JSX.Element {
   return (
     <div className="flex flex-1 flex-col gap-4">
-      <H2>{body}</H2>
+      <H3 className={className.headerPrimary}>{body}</H3>
       <p>{specificText}</p>
       <Ul className="gap-2">
         {resources.map((resource, index) => (
--- a/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
@@ -1,0 +1,78 @@
+import { RA } from '../../utils/types';
+import { AnySchema, SerializedResource } from '../DataModel/helperTypes';
+import { H2, Ul } from '../Atoms';
+import { FormattedResource } from '../Molecules/FormattedResource';
+import { deserializeResource } from '../DataModel/helpers';
+import React from 'react';
+import { OverlayContext } from '../Router/Router';
+import { Dialog } from '../Molecules/Dialog';
+import { recordMergingTableSpec } from './definitions';
+import { dialogIcons } from '../Atoms/Icons';
+import { Button } from '../Atoms/Button';
+import { commonText } from '../../localization/common';
+import { mergingText } from '../../localization/merging';
+import { Tables } from '../DataModel/types';
+
+function InvalidMergeRecords({
+  resources,
+  header = '',
+}: {
+  readonly resources: RA<SerializedResource<AnySchema>>;
+  readonly header?: string;
+}): JSX.Element {
+  return (
+    <div className="flex flex-1 flex-col gap-4">
+      <H2>{header}</H2>
+      <Ul className="gap-2">
+        {resources.map((resource, index) => (
+          <li
+            key={(resource.id as number) ?? index}
+            className="min-h-[theme(spacing.8)]"
+          >
+            <FormattedResource
+              resource={deserializeResource(resource)}
+            ></FormattedResource>
+          </li>
+        ))}
+      </Ul>
+    </div>
+  );
+}
+
+export function InvalidMergeRecordsDialog({
+  modelName,
+  recordsToIgnore,
+  onDismiss: handleDismiss,
+}: {
+  readonly modelName: keyof Tables;
+  readonly recordsToIgnore: RA<SerializedResource<AnySchema>>;
+  readonly onDismiss: (ids: RA<number>) => void;
+}) {
+  const handleClose = React.useContext(OverlayContext);
+  return (
+    <Dialog
+      header={recordMergingTableSpec[modelName]?.dialogHeader!}
+      icon={dialogIcons.warning}
+      buttons={
+        <>
+          <Button.DialogClose>{commonText.close()}</Button.DialogClose>
+          <Button.Small
+            onClick={() =>
+              handleDismiss(
+                recordsToIgnore.map((record) => record.id as number)
+              )
+            }
+          >
+            {mergingText.mergeRest()}
+          </Button.Small>
+        </>
+      }
+      onClose={handleClose}
+    >
+      <InvalidMergeRecords
+        resources={recordsToIgnore as RA<SerializedResource<AnySchema>>}
+        header={recordMergingTableSpec[modelName]?.dialogText}
+      />
+    </Dialog>
+  );
+}
--- a/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
@@ -25,7 +25,7 @@ function InvalidMergeRecords({
 }): JSX.Element {
   return (
     <div className="flex flex-1 flex-col gap-4">
-      <H2>{body}</H2>
+      <p>{body}</p>
       <Ul className="gap-2">
         {resources.map((resource, index) => (
           <li
--- a/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
@@ -31,15 +31,22 @@ function InvalidMergeRecords({
       <H3 className={className.headerPrimary}>{body}</H3>
       <p>{specificText}</p>
       <Ul className="gap-2">
-        {resources.map((resource, index) => (
-          <li
-            key={(resource.id as number) ?? index}
-            className="flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
-          >
-            <TableIcon label={true} name={tableName} />
-            <FormattedResource resource={deserializeResource(resource)} />
-          </li>
-        ))}
+        {resources.map((resource, index) => {
+          return (
+            <li
+              key={(resource.id as number) ?? index}
+              className="flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
+            >
+              <TableIcon label={true} name={tableName} />
+              <FormattedResource
+                resource={React.useMemo(
+                  () => deserializeResource(resource),
+                  [resource]
+                )}
+              />
+            </li>
+          );
+        })}
       </Ul>
     </div>
   );
--- a/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
@@ -16,24 +16,24 @@ import { TableIcon } from '../Molecules/TableIcon';
 
 function InvalidMergeRecords({
   resources,
-  header = '',
+  tableName,
+  body = '',
 }: {
+  readonly tableName: keyof Tables;
   readonly resources: RA<SerializedResource<AnySchema>>;
-  readonly header?: string;
+  readonly body?: string;
 }): JSX.Element {
   return (
     <div className="flex flex-1 flex-col gap-4">
-      <H2>{header}</H2>
+      <H2>{body}</H2>
       <Ul className="gap-2">
         {resources.map((resource, index) => (
           <li
             key={(resource.id as number) ?? index}
-            className="flex flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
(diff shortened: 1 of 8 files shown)


carlosmbe (changes requested):
If you merging only two agents. You shouldn't be able to "Merge Others". Rather it should be cancel. Or something else. Worked fine with multiple items though.

Edit:
Note to future testers (or self) 
See how this dialogue interacts with #3818

realVinayak:
Restricting this case only on the front-end. Users can still make API calls, but that may just result in an exception from the backend. 

To test,
1. Create multiple new agents, and add multiple members  to each of those agents.
2. In query results (for merging), select the above agents, along with agents which do not have members to merge for merging.
3. When you click on 'Merge Records', it should warn that agents with members are being merged. It should allow you to either cancel, or merge rest of the other agents - agents without group members. If no such filtered agents are present, it should close the dialog. Clicking on merge should take you to the actual dialog for merging, with only those agents which do not have group members

For example below 
image

carlosmbe:
This is a comment and not part of my review. Could we add the `Agt` Icon next to the names?

realVinayak:
> This is a comment and not part of my review. Could we add the `Agt` Icon next to the names?

Resolved

realVinayak:
> If you merging only two agents. You shouldn't be able to "Merge Others". Rather it should be cancel. Or something else. Worked fine with multiple items though.
> 
> Edit: Note to future testers (or self) See how this dialogue interacts with #3818

Clicking on 'Merge others' will have not allowed you to merge before (should have closed dialog). Removed button in those cases now. Resolved.

CarolineDenis (approved):

grantfitzsimmons:
I'm going to review this soon. PM me if you end up waiting on me @realVinayak

specifyweb/frontend/js_src/lib/localization/merging.ts
```diff
@@ -60,4 +60,14 @@ export const mergingText = createDictionary({
...
+  agentContainsGroupDescription: {
+    'en-us':
+      'The following agents cannot be merged since contain group members',
+  },
+  agentMergingWarning: {
+    'en-us': 'Agents contains group members',
+  },
+  mergeRest: {
+    'en-us': 'Merge others',
+  },
```
maxxxxxdlp, line 72:
```suggestion
  agentContainsGroupDescription: {
    'en-us':
      'The following agents cannot be merged since they contain group members',
  },
  agentMergingWarning: {
    'en-us': 'Agents contain group members',
  },
  mergeOthers: {
    'en-us': 'Merge others',
  },
```

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/definitions.ts
```diff
@@ -0,0 +1,24 @@
+import { Tables } from '../DataModel/types';
+import { SerializedResource } from '../DataModel/helperTypes';
+import { mergingText } from '../../localization/merging';
+
+export const recordMergingTableSpec: Partial<{
+  [P in keyof Tables]: {
```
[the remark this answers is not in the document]
realVinayak: > P?
> 
> did you just randomly hit a letter on the keyboard? 😊

C'mon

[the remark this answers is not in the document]
grantfitzsimmons: 😂

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -0,0 +1,82 @@
...
+      header={recordMergingTableSpec[modelName]?.dialogHeader!}
+      icon={dialogIcons.warning}
+      buttons={
+        <>
+          <Button.DialogClose>{commonText.close()}</Button.DialogClose>
+          {typeof handleDismiss === 'function' && (
+            <Button.Small
+              onClick={() =>
+                handleDismiss(
+                  recordsToIgnore.map((record) => record.id as number)
```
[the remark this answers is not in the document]
CarolineDenis: FYI, error without assertion: 
`Argument of type '(string | number | boolean | null)[]' is not assignable to parameter of type 'RA<number>'.
  Type 'string | number | boolean | null' is not assignable to type 'number'.
    Type 'null' is not assignable to type 'number'.`

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -0,0 +1,82 @@
...
+              {mergingText.mergeRest()}
+            </Button.Small>
+          )}
+        </>
+      }
+      onClose={handleClose}
+    >
+      <InvalidMergeRecords
+        resources={recordsToIgnore as RA<SerializedResource<AnySchema>>}
+        header={recordMergingTableSpec[modelName]?.dialogText}
```
melton-jason: (optional) I know it is used in another context, but find it a little weird that we are using the `dialogText` of the recordMergingTableSpec as the `header` value for the `InvalidMergeRecords` function. 
Might be a little confusing because one might expect the `dialogHeader` property to go here instead.

melton-jason (changes requested):
A few more things that could be clarified or changed! 
Looking good!

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -0,0 +1,82 @@
...
+  readonly header?: string;
+}): JSX.Element {
+  return (
+    <div className="flex flex-1 flex-col gap-4">
+      <H2>{header}</H2>
+      <Ul className="gap-2">
+        {resources.map((resource, index) => (
+          <li
+            key={(resource.id as number) ?? index}
+            className="flex flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
```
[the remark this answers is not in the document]
realVinayak: Not intentional

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -0,0 +1,82 @@
...
+          >
+            <TableIcon label={true} name={resource._tableName} />
```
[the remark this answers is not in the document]
realVinayak: since resource is SerializedResource and not SpecifyResource, `resource.specifyModel.name` won't work.

Good point about it always being defined. I just passed it down for sanity instead

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
realVinayak: Resolved

grantfitzsimmons (approved):
Looking great! Nice work @realVinayak!

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -16,24 +16,24 @@ import { TableIcon } from '../Molecules/TableIcon';
...
             key={(resource.id as number) ?? index}
-            className="flex flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
+            className="flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
           >
-            <TableIcon label={true} name={resource._tableName} />
-            <FormattedResource
-              resource={deserializeResource(resource)}
-            ></FormattedResource>
+            <TableIcon label={true} name={tableName} />
+            <FormattedResource resource={deserializeResource(resource)} />
```
maxxxxxdlp, line 36: (optional) use a useMemo. deserializeResource is somewhat expensive (serialize resource too, because it tries to convert from lower case back to camel case, which requires looping over every field in the table, for each field in the resource, thus `O(n^2)` at very least, but then you do this for every resource since you are in a loop, and you get `O(n^3)` 🤯)

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <Ul className="gap-2">
+        {resources.map((resource, index) => (
+          <li
+            key={(resource.id as number) ?? index}
+            className="flex flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
+          >
+            <TableIcon label={true} name={resource._tableName} />
+            <FormattedResource
+              resource={deserializeResource(resource)}
+            ></FormattedResource>
```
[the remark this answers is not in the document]
maxxxxxdlp: yet another thing Eslint can auto fix

specifyweb/frontend/js_src/lib/localization/merging.ts
```diff
@@ -60,4 +60,14 @@ export const mergingText = createDictionary({
   dismissFromMerging: {
     'en-us': 'Dismiss from merging',
   },
+  agentContainsGroupDescription: {
+    'en-us':
+      'The following agents cannot be merged since they contain group members',
```
maxxxxxdlp, line 65: add a colon `:` at the end of this string

---
OR:


(optional) it seems like you tried to make the solution generic, and not specific to agen table.
to that end, you can improve it a bit further:
instead of `The following agents cannot be merged since they contain group members`
use this string:
```
The following records cannot be merged. Reason:
Agents contain group members
```
and these would be two separate strings - the first one is not specific to Agents, thus can be reused. The second one will be just for agents.

of course, feel free to reword the string or ask Grant for better wording.

---

(optional) also, this string is a bit too long for a header in my opinion, so I would have displayed it as a `<p>` instead, but I would understand it if you chose `<H2>` just for styling

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -0,0 +1,83 @@
...
+  tableName,
+  body = '',
+}: {
+  readonly tableName: keyof Tables;
+  readonly resources: RA<SerializedResource<AnySchema>>;
+  readonly body?: string;
+}): JSX.Element {
+  return (
+    <div className="flex flex-1 flex-col gap-4">
+      <H2>{body}</H2>
```
maxxxxxdlp, line 28: the dialog renders an `<h2>`, thus this should be an `<h3>` or a `<p>`

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
realVinayak: Resolved!

specifyweb/frontend/js_src/lib/localization/merging.ts
realVinayak: Yeah I chose H2 because didn't want to think about styling it. We can switch to <p> and see how that works out.

@grantfitzsimmons 

> of course, feel free to reword the string or ask Grant for better wording.

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
maxxxxxdlp: this was marked as resolved but wasn't fixed.

if the only reason you are using H2 is styling, then you can get the same styling for p or h3 relatively easily:
```tsx
<h3 className={className.headerPrimary}>{body}</h3>
```

(just copied from the definition of H2:)
https://github.com/specify/specify7/blob/207df9a091c59cd04e4e4b1c3eb111c75c74f68a/specifyweb/frontend/js_src/lib/components/Atoms/index.tsx#L66C25-L67

specifyweb/frontend/js_src/lib/components/Merging/definitions.ts
```diff
@@ -18,6 +19,7 @@ export const recordMergingTableSpec: Partial<{
       return hasGroup ? resource : undefined;
     },
     dialogHeader: mergingText.agentMergingWarning(),
-    dialogText: mergingText.agentContainsGroupDescription(),
+    dialogText: mergingText.cannotBeMerged(),
```
maxxxxxdlp, line 22: (optional)

my point about making `dialogText` be generic was that you don't have to specify it here for every table - instead, you would just hardcode it in the `<InvalidMergeRecordsDialog>` since it would be the same for all tables

thus, you could remove it from `dialogText`
once that is removed, you can rename `dialogSpecificText` back to `dialogText`

sorry if this isn't clear

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/definitions.ts
CarolineDenis: Got it for the table name. Although we created two separate strings to have them displayed as 2 lines in the dialog with different styling. 
My suggestion is to have tableMergingWarning instead of agentMergingWarning and to be able to specify the table name in the string.

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -13,32 +13,40 @@ import { commonText } from '../../localization/common';
...
 }: {
   readonly tableName: keyof Tables;
   readonly resources: RA<SerializedResource<AnySchema>>;
-  readonly body?: string;
   readonly specificText?: string;
 }): JSX.Element {
   return (
     <div className="flex flex-1 flex-col gap-4">
-      <H2>{body}</H2>
+      <H3 className={className.headerPrimary}>
```
maxxxxxdlp, line 29: I think you meant to do `<h3 className={className.headerPrimary}>` not `<H3 className={className.headerPrimary}>`, right? (note the lowercase `h3`)

`H3` comes with it's own styles, and then you overwrite those by adding `className={className.headerPrimary}`, so it's kind of redundant

unless you intentionally meant to merge the two styles together, but why?

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -13,32 +13,40 @@ import { commonText } from '../../localization/common';
...
-          <li
-            key={(resource.id as number) ?? index}
-            className="flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
-          >
-            <TableIcon label={true} name={tableName} />
-            <FormattedResource resource={deserializeResource(resource)} />
-          </li>
-        ))}
+        {resources.map((resource, index) => {
+          return (
```
maxxxxxdlp, line 35: remove return and `{`

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -13,32 +13,40 @@ import { commonText } from '../../localization/common';
...
+            <li
+              key={(resource.id as number) ?? index}
+              className="flex min-h-[theme(spacing.8)] flex-1 items-center gap-2"
+            >
+              <TableIcon label={true} name={tableName} />
+              <FormattedResource
+                resource={React.useMemo(
```
maxxxxxdlp, line 42: what did we learn about hooks?
hooks can't be used in here! 🤣
move this useMemo to the top level
and enable ESLint so it can shout about this, instead of me shouting about this  🤣

specifyweb/frontend/js_src/lib/localization/merging.ts
```diff
@@ -61,17 +61,14 @@ export const mergingText = createDictionary({
     'en-us': 'Dismiss from merging',
   },
   agentContainsGroupDescription: {
-    'en-us': `Agents contain group members`,
+    'en-us': `Agent contain group members`,
```
maxxxxxdlp, line 64: either `Agents contain group members` or `Agent contains group members`. otherwise not gramatically correct

`Agents contain group members` is probably better as it counts for more than one agent having group members

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
realVinayak: I thought it was an anti-pattern but couldn't find a reason why this isn't allowed in this specific case. Maybe because I could have different number of hook calls between renders?

maxxxxxdlp: It's not an anti-pattern. It's simply not allowed.

There are only 3 rules of hooks and it pays to have them well memorized:
https://www.w3schools.com/react/react_hooks.asp#:~:text=Hook%20Rules,Hooks%20cannot%20be%20conditional

> Hooks can only be called inside React function components.
> **Hooks can only be called at the top level of a component.**
> Hooks cannot be conditional

Pretty sure @CarolineDenis tried to violate each of these at one point 🤗

CarolineDenis: :joy::see_no_evil:

maxxxxxdlp: Solutions:
- First thing that came to mind, but not the best: Move useMemo outside of the map. and useMemo will have a separate map inside of it, creating an array, which would be accessed by the .map that is in the return statement
- Best performance and the cleanest: extract the contents of the .map() into a separate component. Inside of that component, useMemo hook can be called (as it would be at the top level)
- A bit weird, and not best performance (because on change to any resource, all resources will be re-computed): `const li = React.useMemo(()=> /* copy the entire .map() as you have it right now, except remove useMemo() that you have inside of .map() right now */ ,[...]);`

realVinayak: Resolved. Another possible solution might be to use logic similar to useResource uses https://github.com/specify/specify7/blob/a4188f0e9b0b43f8188302698fec2fabc0678e17/specifyweb/frontend/js_src/lib/components/Merging/index.tsx#L404

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
```diff
@@ -37,19 +37,26 @@ function InvalidMergeRecords({
...
-            />
+            <FormattedMemoizedResource resource={resource} />
           </li>
         ))}
       </Ul>
     </div>
   );
 }
 
+function FormattedMemoizedResource({
```
maxxxxxdlp, line 44: what I meant is for the entire contents of the .map() to be turned into a component - that way the component will serve a bit more purpose than just useMemo

and then you would be able to rename `FormattedMemoizedResource` to just `Record` or `InvalidMergeRecord` (using record instead of resource because that's what you used in `InvalidMergeRecords` - consistency 🤵‍♂️)

please refactor - should be a simpler change

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/Merging/InvalidMergeRecords.tsx
realVinayak: Isn't that what it is right now?

realVinayak: All the contents of .map is a new component (https://github.com/specify/specify7/pull/3832/commits/9be6ce28a5e97585627c527c0e7f93465dd8a302)

maxxxxxdlp: you are right. looks like I was reviewing outdated code somehow

melton-jason (approved):
I only checked the commits since my last review (32766576700c635e7556a3b2549f72d8bba55832 to 9be6ce28a5e97585627c527c0e7f93465dd8a302), but the code looks fine!

PR #3842, opened 2023-07-26
Remove merging capability from forms

### LINKED ISSUE #3836: Agent Merging - Record Set Trying to Merge Agent Into Self leads to 404
**Describe the bug**
If you merge an Agent from a Record Set via the "Merge Records" Button. The list of possible Agents gets updated to include our newly merged Agent. Which means it's possible to merge an agent with itself. Which leads to a 404 Error.

**To Reproduce**
1. Go to a Agent Record Set
2. Click on the Gear Icon (I think it's official name is Form Meta)
3. Click on Merge Records
4. Search for an agent (or use * as a wild card)
5. Merge the Agents
6. Do it again with the same Agent you just merged
7. See error

This is probably easier on a super small Database or just one you intimately know.

**Expected behavior**
The agent should no longer be searchable since they're now in a better place.

**Beautiful Picture**

https://github.com/specify/specify7/assets/53784701/d8e179be-1e3c-44a9-826d-90a2e99c4764

**Desktop:**
 - OS: Mac OS Ventura
 - Browser: Chrome
 - Specify 7 Version: 7.9

CarolineDenis:
Fixes #3836

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormMeta/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormMeta/index.tsx
@@ -22,7 +22,6 @@ import { interactionTables } from '../Interactions/config';
 import { Dialog } from '../Molecules/Dialog';
 import {
   ProtectedAction,
-  ProtectedTable,
   ProtectedTool,
 } from '../Permissions/PermissionDenied';
 import { AutoNumbering } from './AutoNumbering';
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (approved):

melton-jason (changes requested):
Not a huge fan of simply removing something because of an issue. 
Finding a solution would be infinitely better.

However, this is not a major component of record merging, and there is considerable pressure to get the release out, which is why this is a barely acceptable change. 

If we do remove a feature from the code, it should be easy to identify what the previous implementation was, and/or why it was removed.

maxxxxxdlp (approved):
Just like Jason said, I don't like this solution, but for not delaying the release let's have it.

For a proper fix, you can just tell the Search Dialog to exclude the current record using an extra Filter:
https://github.com/specify/specify7/blob/207df9a091c59cd04e4e4b1c3eb111c75c74f68a/specifyweb/frontend/js_src/lib/components/SearchDialog/index.tsx#L56

---

edit: looks like that is already done
https://github.com/specify/specify7/blob/207df9a091c59cd04e4e4b1c3eb111c75c74f68a/specifyweb/frontend/js_src/lib/components/FormMeta/MergeRecord.tsx#L61-L66

the reason it doesn't work is that those extra filters were only respected by regular search, not query builder search until xml-editor branch (this branch likely doesn't have the fix)

thus the issue would likely fix itself in 7.10

if not, just call deduplicate selected resources once selected and abort if there is only one

realVinayak:
The actual fix would be on the backend - it is integrity error if trying to merge into self.

PR #3843, opened 2023-07-26
Disable link in stat item if value not valid

### LINKED ISSUE #3833: Stats Page Query Errors Do Not Use Properly Use Toast
**Describe the bug**
If you run into a Query Builder Error while in Stats Page the toast gets created alongside the dialogue.

**To Reproduce**
1. Go to Stats Page
2. Add a query that's guaranteed to throw an error on run. (Like UnMapped Queries)
3. Open the query

I'm logged in as petrifiedroadkill in NHMD -> NHMD Amber 

**Expected behavior**
The error should only appear as a Toast and not have the Dialogue open.

**Beautiful Pictures**

https://github.com/specify/specify7/assets/53784701/9470c6d6-0213-4499-b31f-8cedb793c6f7

**Desktop:**
 - OS: Mac OS
 - Browser: Chrome
 - Specify 7 Version: 7.9

CarolineDenis:
Fixes #3833

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Statistics/StatsResult.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Statistics/StatsResult.tsx
@@ -69,8 +69,14 @@ export function StatsResult({
         </>
       ) : (
         <li className="flex gap-2">
-          <Button.LikeLink className="flex-1" onClick={handleClickResolved}>
-            <span className="self-start" style={{ wordBreak: "break-word" }}>{label}</span>
+          <Button.LikeLink
+            className="flex-1"
+            onClick={handleClickResolved}
+            disabled={value === 'Error'}
+          >
+            <span className="self-start" style={{ wordBreak: 'break-word' }}>
+              {label}
+            </span>
             <span className="-ml-2 flex-1" />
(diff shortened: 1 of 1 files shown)


grantfitzsimmons:
To test this, you can import this query and add it as a statistic:
[Not.Mapped.Query.json.zip](https://github.com/specify/specify7/files/12177071/Not.Mapped.Query.json.zip)

realVinayak:
Actually, it maybe worth not disabling it. Otherwise, users can't see what the error itself is, right?

maxxxxxdlp (changes requested):
Like @realVinayak said, doing this is a bad idea.

Once again, this is a fix for a tiny issue, where the fix creates a larger problem than the original problem.

https://github.com/specify/specify7/issues/3833 should probably be closed as invalid - two error dialogs are returned because two separate network requests failed

instead of disabling the item, we could hide one of the error dialogs. but trying to hide one of the errors if error messages are the same or something like that is going to be complicated AND bug prone - and I don't want any bugs when it comes to displaying to the user what went wrong.

this is really a minor issue not with ANY developer time - we have bigger fish to fry

I would suggest closing this PR
@grantfitzsimmons what do you think?

PR #3845, opened 2023-07-26
Refresh relative date every x seconds

### LINKED ISSUE #3301: Make relative dates refresh as needed
**Describe the bug**
The Last Refreshed X time ago on the top of the Stats page only refreshes when another UI Element changes. This can be confusing since 10 minutes can pass but if the user does not change any other element. The timer will remain unchanged which can leave the user uncertain of how accurate their current stats are. For example, stats refreshed 30 minutes ago can be shown as 8 seconds ago.

**To Reproduce**
1. Go to any stats page and refresh the page
2. Leave the stats page alone and do not make any changes 
3. See error

**Expected behavior**
A more informative way of being aware of when these stats were refreshed.

More frequent auto updates of that timer.

A more vague title. Like "A few seconds ago", "Quite some time back". "A While Ago"

**Screen Recording**
It looks like a minute worth of a still image but that's how it little refreshs.

https://user-images.githubusercontent.com/53784701/229926835-190da0e5-9d8b-44a9-a72c-3a89a3f5cfa3.mov

**Desktop:**
 - OS:  Mac OSx Ventura
 - Browser: [Chrome]
 - Specify 7 Version: [statspage branch in test panel ]

**Database Name:** NHMD_TEST

**Reported By**
Carlos @carlosmbe 

**Additional context**
Logged in as petrifiedroadkill
IN NHMD AMBER Collection

CarolineDenis:
Fixes #3301


specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -19,14 +19,42 @@ export function DateElement({
   readonly fallback?: React.ReactNode;
   readonly flipDates?: boolean;
 }): JSX.Element {
+  const [relativeDate, setRelativeDate] = React.useState<string | undefined>(
+    date ? getRelativeDate(new Date(date)) : undefined
```
maxxxxxdlp, line 23: this is a possible bug:
any time the default value of your state depends on a prop, you are likely creating a bug
this is because the default value is only used first time the component is rendered - so if the `date` prop would change later, the code will be out of date.

---

edit:
in this case, the bug is fixed thanks to useEffect having `date` as a dependency

but in the future, whenever your default value depends on a prop, instead of `React.useState`, use `useTriggerState` or `useLiveState`

specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -19,14 +19,42 @@ export function DateElement({
...
+  );
+
+  React.useEffect(() => {
+    let intervalValue;
```
maxxxxxdlp, line 27: (optional) set default value for intervalValue to 0
the behavior won't change since setInterval's default is 0, but this will make code more clearer

specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -19,14 +19,42 @@ export function DateElement({
...
+    let interval: NodeJS.Timeout | undefined;
+
+    if (date) {
```
maxxxxxdlp, line 30: (optional)
can you play arround with refactoring this code to use setTimeout?
it might be simplified as a result or it might not.
but in either case it would be good exercise.
if code would be more complicated afterward, feel free to revert.

here is how that might look like:
you have a function that updates the state, and calls setTimeout with appropriate timeout value - the result of setTimeout will be stored in `interval` just like now
you call the function

specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -19,14 +19,42 @@ export function DateElement({
...
+      interval = setInterval(() => {
+        const dateObject = new Date(date);
+        const now = new Date();
+        const timeDifference = now.getTime() - dateObject.getTime();
+
+        if (timeDifference < 60000) intervalValue = 5000;
```
maxxxxxdlp, line 36: can you please refactor to use these time units instead of numbers?
https://github.com/specify/specify7/blob/xml-editor/specifyweb/frontend/js_src/lib/components/Atoms/timeUnits.ts

<p>

i.e, `60 * MILLISECONDS`
or `HOUR * MILLISECONDS`

</p>
</details> 

much more readable and less bug prone this way

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -33,9 +34,10 @@ export function DateElement({
         const now = new Date();
         const timeDifference = now.getTime() - dateObject.getTime();
 
-        if (timeDifference < 60000) intervalValue = 5000;
-        else if (timeDifference < 360000) intervalValue = 60000;
-        else intervalValue = 36000000;
+        if (timeDifference < 60 * MILLISECONDS)
+          intervalValue = 5 * MILLISECONDS;
+        else if (timeDifference < 6 * MINUTE) intervalValue = MINUTE;
```
maxxxxxdlp, line 39: did you mean `60 * MINUTE` instead of `6 * MINUTE`?
i.e update every minute, until reaches one hour

maxxxxxdlp: actually, using `HOUR` instead of `60 * MINUTE` would simplify this

maxxxxxdlp: but, you still need to add `* MILLISECONDS` in either case

specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -33,9 +34,10 @@ export function DateElement({
...
+        else intervalValue = HOUR;
```
maxxxxxdlp: now this is what I call readable 😊:

```suggestion
        const timeDifference = (now.getTime() - dateObject.getTime()) / MILLISECONDS;

        if (timeDifference < MINUTE)
          intervalValue = SECOND;
        else if (timeDifference < HOUR) intervalValue = MINUTE;
        else intervalValue = HOUR;
```

specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -19,14 +20,71 @@ export function DateElement({
...
+        const now = new Date();
+        const timeDifference = now.getTime() - dateObject.getTime();
+
+        if (timeDifference < 60 * MILLISECONDS)
+          intervalValue = 5 * MILLISECONDS;
+        else if (timeDifference < 6 * MINUTE) intervalValue = MINUTE;
+        else intervalValue = HOUR;
+
+        setRelativeDate(getRelativeDate(new Date(date)));
+      }, intervalValue);
```
maxxxxxdlp, line 43: wait, setInterval wouldn't work correctly here at all, right?
even though you modify the value of intervalValue, this is not react - setInterval will keep on using the value it was called with the first value - thus you would have to cancel the interval and set a new one - at which point it would be simpler to just use setTimeout

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -25,57 +25,34 @@ export function DateElement({
...
   React.useEffect(() => {
-    let intervalValue = 0;
-    let interval: NodeJS.Timeout | undefined;
+    let timeout: NodeJS.Timeout | undefined;
 
-    if (date) {
-      interval = setInterval(() => {
+    function updateRelativeDate() {
+      if (date) {
         const dateObject = new Date(date);
```
maxxxxxdlp, line 32: move this line outside the function - you don't need to create a new date object on every function call as the date doesn't change

specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -25,57 +25,34 @@ export function DateElement({
...
-          intervalValue = 5 * MILLISECONDS;
-        else if (timeDifference < 6 * MINUTE) intervalValue = MINUTE;
-        else intervalValue = HOUR;
+        let timeoutValue = 0;
+
+        if (timeDifference < MINUTE) timeoutValue = SECOND;
+        else if (timeDifference < HOUR) timeoutValue = MINUTE;
+        else timeoutValue = HOUR;
 
         setRelativeDate(getRelativeDate(new Date(date)));
```
maxxxxxdlp, line 43: why create a date once again here when you can just reuse the `dateObject`?

maxxxxxdlp (approved):
two optimizations. otherwise good

carlosmbe:
More testing instructions would be greatly appreciated. XoXo

CarolineDenis:
> More testing instructions would be greatly appreciated. @CarolineDenis

The current behavior on 7.9 (in the stat page, by example): 
- the relative date in the top right of the screen is not updated like it should, you can see 'Last refreshed 15 seconds ago' even if it was 35 min ago
On this new branch: 
- you should see the relative date be updated every second at first than every minute than every hour

carlosmbe:
@CarolineDenis 
Aside from Stats Page. Which areas are affected by this change? That was my main concern tbh

CarolineDenis:
@carlosmbe, you can find relative dates in the Work Bench and in the notifications

grantfitzsimmons:
> notifications

Screenshot 2023-09-29 at 9 27 22 AM

It is not showing in the notifications, but that is due to the future date of the system presumably

grantfitzsimmons (approved):
🤷 Works in WorkBench notifications, stats page...
not sure it's worth changing

CarolineDenis:
> 🤷 Works in WorkBench notifications, stats page...
> not sure it's worth changing

agree

PR #3839, opened 2023-07-25
Improvements to Uniqueness Rule handling, and ACID improvements to agent merging

realVinayak:
1. Fixes https://github.com/specify/specify7/issues/3801. Checkout @melton-jason's comment (https://github.com/specify/specify7/issues/3801#issuecomment-1640975722) to do an initial test. 
2. Makes ACID improvements to merging. Left comments explaining those decisions.

Priorities: 

- [x] Instead of returning `http.HttpResponseNotFound` and `http.HttpResponseNotAllowed`, raise an exception, and then catch it from outside recursive merge. Otherwise it won't rollback at all. Currently reviewing changes to celery worker's part for this.
- [x] Take least ordering as new record in dedup. Added test for this also. Passes!
- [x] Add more comprehensive unit tests for some of the changes, and rollbacks.
- [ ] Fix this `# BUG: timestampmodified could be null for one record, and not the other`
- [ ] Fix uniqueness rule mess. Implementing solution in my comment there https://github.com/specify/specify7/issues/3841 
- [ ] Do code cleanups post merge with celery worker.

PR #3846, opened 2023-07-27
Git and history improvements over PR #3839

realVinayak:
Previous branch and PR (https://github.com/specify/specify7/pull/3839) had considerable branch management issues with celery worker changes. This is a testing ready version of my changes, just without the celery worker. 

EDIT: The branch management issues were caused by me prematurely merging the changes from celery worker, and _then_ working on top of them

All list items from here are valid https://github.com/specify/specify7/pull/3839#issue-1820204543

EDIT: Merged v7.9-dev into this

### DIFF
--- a/specifyweb/businessrules/tests/accessionagent.py
+++ b/specifyweb/businessrules/tests/accessionagent.py
@@ -5,6 +5,7 @@ from specifyweb.specify.api_tests import ApiTests
 from ..exceptions import BusinessRuleException
 
 class AccessionAgentTests(ApiTests):
+    @skip("rule was removed in 17e82c6157")
     def test_no_duped_agents_in_accession(self):
         accession = models.Accession.objects.create(
             accessionnumber='a',
@@ -19,6 +20,7 @@ class AccessionAgentTests(ApiTests):
                 agent=self.agent,
                 role='Donor')
 
+    @skip("rule was removed in 17e82c6157")
     def test_no_duped_roles_in_accession(self):
         accession = models.Accession.objects.create(
             accessionnumber='a',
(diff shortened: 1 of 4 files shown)


realVinayak:
TODO: Need to carry over Caroline's commits to this branch

realVinayak:
Testing instructions:

For Issue https://github.com/specify/specify7/issues/3801

I'll recommend running the below steps for at least some of Deaccession, Disposal, Loan, Accession, Gift, Borrow tables. Also test on both v7.9-dev and this branch. v7.9-dev shouldn't be able to merge these, and this branch should.

Later on, testers can go wild and merge anything into anything - and as long as it is not an actual business rule exception, it should work. 

Steps:
1. Make an interaction via the interactions form. I'll call it accession `001` for these steps. Make two new accession agents. I'll recommend creating two fresh new agents (but you can select a previous agent too). You can also make more than two accession agents. For this example, two new agents are `agent 001` and `agent 002`. **Important** that both of these agents should be assigned same role in the accession.
2. Make another accession, and select `agent 002` and assign any role to this agent.
3. Go into query builder, select agents `agent 001` and `agent 002`. 
4. Merge `agent 001` into `agent 002`. The order is important to recreate issue.
5. This should fail in v7.9-dev, with an error saying `found multiple records violating uniqueness constraints`. In this branch, it should work without any errors.

realVinayak:
To test lower order is taken in collector and determiners in case of deletion. v7.9 should take the most recently created one.

1. Go to collectingevent form. Add a collector. Make a new agent for this (just for testing - can also be tested with existing ones). Calling this agent `agent 001`. In the remarks field, put some dummy value for this collector (this is to distinguish after merge). 
image
3. Save collecting event form. I'll recommend waiting a few seconds before going to next step.
2. Make another collector (with existing agent or new). Call this `agent 002`. Put some dummy value in remarks. Save the form.
4. Merge `agent 001` into `agent 002`. 
5. Go to collecting event form again. In the remarks, check the following: 
v7.9-dev should take the one most recently one (look at dummy value you put in remarks to confirm). This branch should take it by order (so the first one you created)

Try doing the exact same steps for determiners too.

PR #3848, opened 2023-07-27
Hide export to CSV button if no displayed field 

### LINKED ISSUE #3834: Disable Create CSV if All Columns are Hidden
**Describe the bug**
If you have all the columns hidden. You can still create a CSV. It's empty so I don't really see the use. Having a minimum of 1 column visible seems more practical.

**To Reproduce**
1. Go to any Query
2. Hide all columns
3. Run the Query and Create a CSV

**Expected behavior**
Disabled if visible columns == 0

**Beautiful Pictures**

https://github.com/specify/specify7/assets/53784701/de2818f7-c419-4d39-8787-f1bda11a5fe2

**Desktop:**
 - OS: Mac OS
 - Browser: Chrome
 - Specify 7 Version: 7.9

CarolineDenis:
Fixes #3834


specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -197,12 +204,12 @@ export function QueryResults(props: Props): JSX.Element {
           </Button.Small>
         )}
         <div className="-ml-2 flex-1" />
-        {extraButtons}
+        {props.displayedFields.length > 0 && extraButtons}
```
maxxxxxdlp, line 207: since all props in this component are restructured, you should restructure `displayedFields` too at the top of the component, so that you can use it here without `props.` - for consistency

grantfitzsimmons (approved):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
maxxxxxdlp: TLDR: (optional) consider renaming `extraButtons` to `exportButtons` or propose a better solution. See below for reasoning

This check does not belong here, but unfortunately, we can't easily put it higher up. Still, see my reasoning (no code changes required, just want you to look out for this in the future and understand why I called it out):

I think this is wrong from a code quality perspective. You need to invert the control.

If we are to think about the ideal case, this component receives an `extraButtons` prop, and displays it - it should now know anything about what `extraButtons` are, nor make any assumptions about when `extraButtons` should be displayed - for all it knows, the parent will provide different `extraButtons` depending on whether there are any fields or not.

Since both `displayedFields` and `extraButtons` are props from a higher level, you should move this to the highest level at which both props are created (which ideally would be the same level, but unfortunately not in this case)

To clarify, I mean that in the ideal case, this is the place where you would check if any fields are displayed:

https://github.com/specify/specify7/blob/207df9a091c59cd04e4e4b1c3eb111c75c74f68a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx#L537-L546

It would be logical there as that is where extra buttons are created, and would fit neatly as we already have one check there.

Unfortunately, we can't do that because `displayedFields` is not available, there, and making them available there would be too much trouble.

You could theoretically move this check to `<QueryResultsWrapper />`, but wouldn't make too much difference as it would still be a wrong place (it would result in the same problem, QueryResultsWrapper making assumption about what the contents of extraButtons are, is a code smell)

One possible way to avoid this issue is to make an **explicit** assumption about what content is expected. If you rename `extraButtons` to `exportButtons`, this issue is resolved, as the assumption about the content is made explicit.

@melton-jason @realVinayak you may find the above read useful too - helps designing solid components/functions. Maybe you can even suggest a solution that would work best in this case.

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -9,7 +9,14 @@ import { commonText } from '../../localization/common';
...
 import { f } from '../../utils/functools';
-import { filterArray, type GetOrSet, type GetSet, type IR, type R, type RA } from '../../utils/types';
+import {
+  filterArray,
+  type GetOrSet,
+  type GetSet,
+  type IR,
+  type R,
+  type RA,
+} from '../../utils/types';
```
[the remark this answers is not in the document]
CarolineDenis: it's my IDE

PR #3850, opened 2023-07-27
Add query builder header in embedded dialog

### LINKED ISSUE #3830: Add the ability to switch from detailed to basic view in query dialogs
image

At this time, there is no way to switch to the basic view, which would be very useful for these dialogs in particular as edits are not being made.

CarolineDenis:
Fixes #3830


specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx
```diff
@@ -61,6 +61,7 @@ export function QueryHeader({
   readonly unsetUnloadProtect: () => void;
   readonly onTriedToSave: () => void;
   readonly onSaved: () => void;
+  readonly toggleMapping: () => void;
```
maxxxxxdlp, line 64: `onToggleMapping`?
to be consistent with how we call functions that mutate state

specifyweb/frontend/js_src/lib/components/QueryBuilder/reducer.ts
```diff
@@ -52,6 +53,7 @@ export const getInitialState = ({
 }): MainState => ({
   type: 'MainState',
   fields: parseQueryFields(query.fields ?? []),
+  showMappingView: getCache('queryBuilder', 'showMappingView') ?? true,
```
maxxxxxdlp, line 56: would it be possible to call `useCachedState` in the components that need this cache instead and remove `showMappingView` from the MainState?

putting getCache/setState in a state is a bad pattern I used when I was designing first react components in Specify, but don't do that anymore. reasons: central state is more verbose (requires more boilerplate code), leads to worse performance (any change to the state triggers re-render of all components that use the state in any way) and is bug prone (useCachedState listens for cache being updated by some other component somewhere else. getCache/setCache does not listen, thus the state could get out of date)

maxxxxxdlp (changes requested):

maxxxxxdlp:
@specify/ux-testing when testing, please be extra careful to also test the save/save as buttons
test in different places where embedded query appears: Search Dialog, Meta->Edit History, (for a pick list record) Meta->Pick List Usages

specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx
CarolineDenis: just took the name that was in production but can change to onToggleMapping if it's better

specifyweb/frontend/js_src/lib/components/QueryBuilder/reducer.ts
CarolineDenis: I pushed a code using cachedState but not sure that is what you meant....

specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx
```diff
@@ -79,6 +80,11 @@ export function QueryHeader({
 
   const [isBasic, setIsBasic] = useQueryViewPref(query.id);
 
+  const [showMappingView = true, _] = useCachedState(
```
maxxxxxdlp, line 83: you don't need to pass down `onToggleMapping` at all

the 2nd argument of `useCachedState` in here is the callback to change the state - thus you can change the state from here, without the need for any props from the parent

remember that changes to `useCachedState` are synchronized between all the places where `useCachedState` is used

maxxxxxdlp (changes requested):

CarolineDenis:
@maxxxxxdlp all the query builder are synchronized atm, if we click Hide Mapping in one query it will be the case for all the queries. Is that what we want?

melton-jason:
> @maxxxxxdlp all the query builder are synchronized atm, if we click Hide Mapping in one query it will be the case for all the queries. Is that what we want?

I believe that would also be a question to ask @specify/ux-testing :wink:

maxxxxxdlp (approved):
remove commented out code

maxxxxxdlp:
> @maxxxxxdlp all the query builder are synchronized atm, if we click Hide Mapping in one query it will be the case for all the queries. Is that what we want?

Fine with me either way.
If you want, you can store in separate local storage keys depending on whether query builder is embedded or not - as most users would likely want to see the mapping explorer in the full query builder, but might want to hide it when in the dialog if on a smaller screen

bronwyncombs (changes requested):
When saving a query made in an embedded dialog (not via stats) it will prompt you to leave the current form and open the saved query in the main query builder.

It should just save the query and not show this dialog. Same applies to Save As button. 

https://github.com/specify/specify7/assets/135047322/ec58a46b-3c7b-4ead-927d-e0fcfe895f6b

---
**Comments:**

Embedded dialog header on stats page looks good! Switching to basic view reduces the width of the dialog, but I think this is fine. 

Embedded and full query builders in sync so detailed/basic view is carried over

CarolineDenis:
Decision: 
Keep "Basic View" and "Hide Field Mapper" bins in embedded header dialog. Remove "Save Query" btn from it

bronwyncombs (approved):
Looks good! Save/Save As issue resolved.

maxxxxxdlp:
For future reference, we could bring back the save button, or fix https://github.com/specify/specify7/issues/3000

Issue #133, opened 2015-06-26
Repository Agreement business rule trickiness


timo11:
Specify6 has no uniqueness constraint defined for accessionagent relative to repositoryagreement.
But it does enforce repositoryagreement, agentid, role uniqueness in the front end.
Specify7 also enforces that constraint in its front end business rules. 
Since that constraint is not 'built-in' to the db schema, Sp7 probably ought to have a backend businessrule for it, but its backend bus rule system doesn't current allow multiple field unique constraints, i.e. it doesn't support role AND agent being unique in repoagreement.

realVinayak:
Should be resolved in [Uniqueness Rule handling](https://github.com/specify/specify7/pull/3839)

PR #3851, opened 2023-07-27
Correctly enforce Collector -> agent uniqueness rule

### LINKED ISSUE #3841: Nested uniqueness don't work in frontend.
Things like Agent should be unique to CollectingEvent. So frontend doesn't check for this anymore. 
Before: 
image

You don't get that warning anymore. @melton-jason @grantfitzsimmons this used to work before but now don't. 

So, nested uniqueness rules that SQL and backend enforces are still enforced - but not the frontend ones.

The best solution will be to fix this as soon as possible. 

I intend to remove front-end only uniqueness from backend ones (I made them synced as part of https://github.com/specify/specify7/pull/3839). I also plan to make frontend and backend uniqueness actually separate now. Plan is to make a base json, and then another one that extends the base one for frontend. @melton-jason thoughts? So two json files. Ones that both share would be in base. Extra ones will be in another json, and frontend will read from both. But backend just from the base.

melton-jason:
Fixes #3841 

TO TEST: 
- Try adding a Collector to a CollectingEvent who is already a Collector in that event
- Test and make sure other 'types' of uniqueness rules work correctly (examples provided below)
  - Catalognumber must be unique in Collection 
  - Role and Agent must be unique to AccessionAgent

This PR also sets the uniqueIdentifier field in the CollectionObject, CollectingEvent, and Locality tables to be globally unique.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
@@ -250,7 +250,7 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
         fieldValue: string | number | null,
         fieldName: string
       ): boolean => {
-        if (typeof Number(other.id) !== 'number' && other.id === this.resource.id) return false;
+        if (other.id != null && other.id === this.resource.id) return false;
         if (other.cid === this.resource.cid) return false;
         const otherValue = other.get(fieldName);
 
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
@@ -1,4 +1,4 @@
-import { IR, overwriteReadOnly, RA } from '../../utils/types';
+import { filterArray, IR, overwriteReadOnly, RA } from '../../utils/types';
 import { AnySchema, AnyTree, CommonFields, TableFields } from './helperTypes';
 import { SpecifyResource } from './legacyTypes';
 import { BusinessRuleDefs, businessRuleDefs } from './businessRuleDefs';
@@ -138,7 +138,10 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
           fieldNames = fieldNames.concat(uniqueRule.otherFields);
           scope = uniqueRule.field;
         }
-        return this.uniqueIn(scope as string, fieldNames);
+        return this.uniqueIn(
+          ((scope ?? '') as string).toLowerCase(),
+          fieldNames
+        );
       }
     );
 
@@ -265,14 +268,17 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
     };
 
     if (scope !== undefined) {
-      const localCollection =
-        this.resource.collection?.models !== undefined
-          ? this.resource.collection.models.filter(
-              (resource) => resource !== undefined
-            )
-          : [];
+      const localCollection = this.resource.collection ?? { models: [] };
 
-      const duplicates = localCollection.filter((resource) =>
+      if (
+        typeof localCollection.field?.name === 'string' &&
+        localCollection.field.name.toLowerCase() !== scope
+      )
+        return { valid: true };
+
+      const localResources = filterArray(localCollection.models);
+
+      const duplicates = localResources.filter((resource) =>
         hasSameValues(resource)
       );
 
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
@@ -37,7 +37,7 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
   }
 
   private changed(resource: SpecifyResource<SCHEMA>): void {
-    if (!resource.isBeingInitialized && typeof resource.changed === 'object') {
+    if (resource.isBeingInitialized && typeof resource.changed === 'object') {
       Object.keys(resource.changed).forEach((field) => {
         this.checkField(field);
       });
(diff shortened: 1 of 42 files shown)


specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -250,7 +250,7 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
         fieldValue: string | number | null,
         fieldName: string
       ): boolean => {
-        if (other.id !== null && other.id === this.resource.id) return false;
+        if (typeof Number(other.id) !== 'number' && other.id === this.resource.id) return false;
```
maxxxxxdlp, line 253: hmm, could you double check `typeof Number(other.id) !== 'number'`?

I am pretty sure this is always true, regardless of value

```ts
Number('a')
> NaN
typeof Number('a')
> "number"

maxxxxxdlp: "non a number" is a number after all 🤪

maxxxxxdlp: also, I belive we have an eslint rule about not using the `Number` constructor - because usually a better alternative exists (be it `Number.parseInt/parseFloat`, `f.parseInt/parseFloat` or some other type of check)

don't forget that `Number.isNaN()` also exists

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
melton-jason: Ah, thank you for the alternatives. 
Well, the problem was that `other.id` was undefined rather than null, and while this would be fixed by removing the strict inequality, there may have been odd behaviors with Javascript's implicit casting. 

Now that I look into it, this condition _is_ always true. I assumed Number(undefined) would not be a number, but it is NaN, as you said. (And even worse, Number(null) is 0).

maxxxxxdlp: then either do:
`other.id !== null && other.id !== undefined`
or:
`other.id != null`
both are equivalent (2nd one is only ever going to do auto-casting for null and undefined)
or:
`(other.id ?? undefined) === undefined`
(paranthesis are important due to `??` having lower precedence than `===` 🤦‍♀️)

melton-jason: I'd have to look into this more, but interestingly `other.id != null` and `(other.id ?? undefined) === undefined` had different behaviors. 

I think i'm okay settling for  `other.id != null` since it only casts for null/undefined values.

melton-jason: Fixed in 91eb76b31086d66c876158ea2f533f0b360a40c8

melton-jason:
@specify/ux-testing, this can be tested now

bronwyncombs (changes requested):
Everything looks and works great except this one dialog! 

Screenshot 2023-07-28 at 11 22 30 AM

"Save Blocked" dialog for Accessions and Repository Agreements list the errors for both form names. It should only refer to errors for the form you're in.

realVinayak:
Also need to separate backend and frontend uniqueness rules, right? 

Repository agreement uniqueness in accessionagent with agent and role wasn't enforced on backend (or any uniqueness for more than two fields for that matter). Now they are as part of agent merging. So, all the frontend uniqueness rules that weren't enforced on backend should still not be enforced.

melton-jason:
> Repository agreement uniqueness in accessionagent with agent and role wasn't enforced on backend (or any uniqueness for more than two fields for that matter). Now they are as part of agent merging. So, all the frontend uniqueness rules that weren't enforced on backend should still not be enforced.

This PR was based off of `v7.9-bus-uniq-clean-hist`, which contains all of your commits which fixes #133. 
Should I revert those commits?

melton-jason:
> Everything looks and works great except this one dialog!
> 
> Screenshot 2023-07-28 at 11 22 30 AM
> 
> "Save Blocked" dialog for Accessions and Repository Agreements list the errors for both form names. It should only refer to errors for the form you're in.

@specify/ux-testing, this should be fixed now
Please test the uniqueness rule for both AccessionAgents, RepositoryAgreementAgents, and a miscellaneous uniqueness rule (like CatalogNumber being unique to Collection on CollectionObject)

melton-jason:
I'd say this can be merged if everything is good to go!

One side effect of this Pull Request is for the rule that Role and Agent must be unique for AccessionAgent in a Repository Agreement. Previously, this rule was only enforced on the frontend and not the backend. Unlike other similar rules, this was also not a constraint at the database level. This means that you could upload duplicate Agent and Role to an AccessionAgent on a Repository Agreement using the WorkBench. However, the frontend would not allow you to save the record when viewing the DataEntry form after the upload. 

@realVinayak's changes added the logic to enforce rules such as that one for the backend. Meaning that uploaded datasets through the WorkBench (or uploading AccessionAgents to RepositoryAgreements via the api) which now violated the new rule for the backend would not work as "expected".

I would say this new behavior is OK. It ensures consistency between the frontend and backend, and leaves no later surprises to the user. Consider the case where a duplicate AccessionAgent was uploaded to a Repository Agreement via the WorkBench. 
Everything would seem fine, until they _actually_ viewed the uploaded record and tried to make changes and/or save.

grantfitzsimmons:
I agree with your assessment. Enforcing the rule that the Role and Agent must be unique in a Repository Agreement on both the front-end and the back-end ensures consistency and prevents potential issues for users. It's ideal that this enforcement applies to both the API and WorkBench.

realVinayak:
Yeah, it probably is better to bite the bullet right now, rather than delaying it. Maybe prepare for some emails?

Didn't realize those changes will close this also (https://github.com/specify/specify7/issues/133).

PR #3852, opened 2023-07-27
Remove results table if empty due to all fields being hidden

### LINKED ISSUE #3829: Strange Rendering of Locality Results
**Describe the bug**

I performed these tests in Mor Paleo and NHMD. When performing a locality query and hiding every option in the results. There's a weird layout in the results.

**To Reproduce**
1. Go to Queries and Create (or Open) a Locality Query
2. Add some fields and make them all invisible 
3. Run Query
4. Scroll down to results
5. See error

**Expected behavior**
Similar behaviour to when you hide Collection Objects

**Screenshots**
Screenshot 2023-07-21 at 4 56 22 PM

Screenshot 2023-07-21 at 4 48 35 PM

https://github.com/specify/specify7/assets/53784701/9e22d0f7-e692-412b-82ae-b4691de76621

**Desktop:**
 - OS:  Mac OS 
 - Browser: Chrome
 - Specify 7 Version: 7.9 and 7.8.13

CarolineDenis:
Fixes #3829


specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -197,12 +204,15 @@ export function QueryResults(props: Props): JSX.Element {
           </Button.Small>
         )}
         <div className="-ml-2 flex-1" />
-        {extraButtons}
+        {props.displayedFields.length > 0 && visibleFieldSpecs.length > 0
```
maxxxxxdlp, line 207: This PR appears to be a superset of https://github.com/specify/specify7/pull/3848/files, and to that end the same comment applies

maxxxxxdlp: what is the difference between `props.displayedFields.length` and `visibleFieldSpecs.length`? can you check if they are ever not equal, and if so, in which cases?

then, double check all other usages of `visibleFieldSpecs.length` and `props.displayedFields.length` in this file. Ideally, both lengths are always the same, and so you would only use one of those two in all places in this component

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -312,55 +322,58 @@ export function QueryResults(props: Props): JSX.Element {
               )}
             </div>
           </div>
+        ) : null}
+        {visibleFieldSpecs.length > 0 && (
+          <div role="rowgroup">
+            {showResults &&
```
maxxxxxdlp, line 328: can you move the `visibleFieldSpecs.length > 0` condition here (`{showResults && ...`)?

this way, you can remove one extra level of indent

or even better, just add `visibleFieldSpecs.length > 0` to the definition of `showResults` and see if that works

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
melton-jason: > what is the difference between props.displayedFields.length and visibleFieldSpecs.length? can you check if they are ever not equal, and if so, in which cases?


https://github.com/specify/specify7/blob/2a259ca22b439ba71b9313705cb2a253ac8edd6e/specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsWrapper.tsx#L156

displayedFields seems to be the fields which are supposed to be shown in the Query Results (the ones that are explicitly marked to be shown in the results. This definition seems consistent for all usages of `displayedFields`)

https://github.com/specify/specify7/blob/2a259ca22b439ba71b9313705cb2a253ac8edd6e/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx#L96

visibleFieldSpecs are all of the fieldSpecs that are not 'Phantom' fields. 

https://github.com/specify/specify7/blob/2a259ca22b439ba71b9313705cb2a253ac8edd6e/specifyweb/frontend/js_src/lib/components/QueryBuilder/fieldSpec.ts#L75-L82

maxxxxxdlp (approved):

bronwyncombs (approved):
Looks good!

https://github.com/specify/specify7/assets/135047322/bb840380-8783-46ef-b807-6a25b8bc98ae

Screenshot 2023-09-29 at 10 33 19 AM

Screenshot 2023-09-29 at 10 33 53 AM

PR #3853, opened 2023-07-27
Use accent color 3 for menu item when hover

### LINKED ISSUE #3814: Use accent color 3 instead of 4 when hovering over an item in the navigation menuin light mode
## Accent color 3:
image

## Accent color 4:
image

CarolineDenis:
Fixes #3814


melton-jason (commented):
This will affect every link in Specify. Is that the intention? 

@specify/ux-testing, what do you think? Is this change for the better, or should _only_ the Navigation Menu items have the accent-color-3?  

For comparison purposes, one prominent places where links are used is the Tree Viewer. 
Every node in a tree is its own link, as well as the ranks in the 'header' of the viewer.

bronwyncombs:
> For comparison purposes, one prominent places where links are used is the Tree Viewer.
Every node in a tree is its own link, as well as the ranks in the 'header' of the viewer.

I just looked into this, and tree links are already Accent color 3! Same with query statistic links. 

However, we should consider that this would still change the "open new tab" icons (like by query results) and also most of the dialogs that the side bar opens (dialogs for links for saved content, table selecting, user tools, etc.). Personally, I feel like the trees and stats page plus all of those would be more than an accent. 

---

Also to consider, Notifications and Collapse in the side menu are already Accent color 3 to stand out. Would we keep that effect and switch those two to Accent color 4 now? 


https://github.com/specify/specify7/assets/135047322/e52bbd58-e00f-48f4-b1ec-a182d365d43d

melton-jason:
> I just looked into this, and tree links are already Accent color 3! Same with query statistic links.

I'll need to look into the classNames (css) of those links... I just quickly checked they were links and were using accent-color-3. 
Still, there might be other unintended links using the css which got changed (beside the open new tab icons and dialogs opened by the navigation menu), meaning they are now using accent-color-3 rather than accent-color-4. 
I suppose we'd have to look into the code to find all the specific usages.

maxxxxxdlp (commented):
I would suggest applying this only in the menu, not for all links - menu is the only place where dark background is used even in light mode, thus it deserves a bit of special styles.

Also, since you are close to release, I would discourage making any global style changes 😜

CarolineDenis:
@maxxxxxdlp how can I overwrite the hover color for the ActiveLink in Header/index.tsx
Is there a tailwind class I can add to the <span> children? (line 216)

maxxxxxdlp:
You can add this className to the `<header>` element:
```
hover:[&_a.link]:text-gray-500

carlosmbe (approved):

carlosmbe:
With the Navigation Menu it seems fine. I prefer the consistency. As for the other bits. I don't have any strong opinions. Requested a review from @bronwyncombs as they brought up a fair point earlier.

CarolineDenis:
> With the Navigation Menu it seems fine. I prefer the consistency. As for the other bits. I don't have any strong opinions. Requested a review from @bronwyncombs as they brought up a fair point earlier.

Which other bits are you referring to?

carlosmbe:
> > With the Navigation Menu it seems fine. I prefer the consistency. As for the other bits. I don't have any strong opinions. Requested a review from @bronwyncombs as they brought up a fair point earlier.
> 
> Which other bits are you referring to?

I was referring to this comment https://github.com/specify/specify7/pull/3853#issuecomment-1654882317

PR #3856, opened 2023-07-27
Format notifications for merge process

### LINKED ISSUE #3854: Properly format notifications from merge process
Screenshot 2023-07-27 at 1 13 07 PM

Types that need to be handled:
- `record-merge-starting`
- `record-merge-pending`
- `record-merge-succeeded`
- `record-merge-failed`

See the `record_merge` file for more information.

The succeeded notification should have a link to the newly merged record, with the URL constructed based on the table name and record ID.

CarolineDenis:
Fixes #3854

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
@@ -133,22 +133,13 @@ export const notificationRenderers: IR<
     const collection = userInformation.availableCollections.find(
       ({ id }) => id === collectionId
     );
-    const [collectionName, setCollectionName] = React.useState<
-      string | null | undefined
-    >('');
-
-    React.useEffect(() => {
-      setCollectionName(collection?.collectionName);
-    }, [collection]);
 
     return (
       <>
         {mergingText.mergingHasStarted()}
         <div className="flex items-center gap-2">
           <TableIcon label name={tableName} />
-          <p>
-            {collectionName} {'-'} {mergeName}
-          </p>
+          <p>{`${collection?.collectionName} - ${mergeName}`}</p>
         </div>
       </>
     );
@@ -157,18 +148,15 @@ export const notificationRenderers: IR<
     const tableName = notification.payload.table;
     const id = parseInt(notification.payload.new_record_id);
     const ids = [JSON.parse(notification.payload.old_record_ids), id];
+    const url = formatUrl(`/specify/overlay/merge/${tableName}/`, {
+      [mergingQueryParameter]: Array.from(ids).join(','),
+    });
     return (
       <>
         {mergingText.mergingHasFailed()}
         <div className="flex items-center gap-2">
           <TableIcon label name={tableName} />
-          <Link.NewTab
-            href={formatUrl(`/specify/overlay/merge/${tableName}/`, {
-              [mergingQueryParameter]: Array.from(ids).join(','),
-            })}
-          >
-            {mergingText.retryMerge()}
-          </Link.NewTab>
+          <Link.NewTab href={url}>{mergingText.retryMerge()}</Link.NewTab>
         </div>
       </>
     );
--- a/specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
@@ -5,6 +5,10 @@ import { notificationsText } from '../../localization/notifications';
 import { StringToJsx } from '../../localization/utils';
 import type { IR } from '../../utils/types';
 import { Link } from '../Atoms/Link';
+import { mergingText } from '../../localization/merging';
+import { TableIcon } from '../Molecules/TableIcon';
+import { FormattedResource } from '../Molecules/FormattedResource';
+import { getModel } from '../DataModel/schema';
 
 export type GenericNotification = {
   readonly messageId: string;
@@ -117,6 +121,32 @@ export const notificationRenderers: IR<
         }}
         string={notificationsText.dataSetOwnershipTransferred()}
       />
+    );
+  },
+  'record-merge-starting'() {
+    return mergingText.mergingHasStarted();
+  },
+  'record-merge-failed'() {
+    return mergingText.mergingFailed();
+  },
+  'record-merge-succeeded'(notification) {
+    const id = parseInt(notification.payload.new_record_id);
+    const tableName = notification.payload.table;
+    const model = getModel(tableName);
+    const resource =
+      typeof model === 'object' ? new model.Resource({ id }) : undefined;
+    return (
+      resource !== undefined && (
+        <>
+          {mergingText.mergingSucceeded()}
+          <Link.NewTab
+            href={`/specify/view/${notification.payload.table}/${notification.payload.new_record_id}`}
+          >
+            <TableIcon label name={notification.payload.table} />
+            <FormattedResource asLink={false} resource={resource} />
+          </Link.NewTab>
+        </>
+      )
     );
   },
   default(notification) {
--- a/specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
@@ -8,7 +8,8 @@ import { Link } from '../Atoms/Link';
 import { mergingText } from '../../localization/merging';
 import { TableIcon } from '../Molecules/TableIcon';
 import { FormattedResource } from '../Molecules/FormattedResource';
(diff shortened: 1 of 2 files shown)


grantfitzsimmons (approved):
I cannot find a way to create a notification with `record-merge-pending` anymore. I don't see it referenced in the code. Since this is potentially not a problem, I will approve. Not sure if it was removed during the dev process.

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -119,6 +123,30 @@ export const notificationRenderers: IR<
       />
     );
   },
+  'record-merge-starting'() {
+    return mergingText.mergingHasStarted();
+  },
+  'record-merge-failed'() {
+    return mergingText.mergingFailed();
+  },
```
melton-jason, line 131: (optional) Preferably, we would include more information for these two processes. Namely, an identifier that the user can use to keep track of which merging process is which. This can be something like the table and id of the merged record, or the celery task id itself. 

There can be multiple merges ongoing at the same time, and when the user just receives "Your merging process failed", it can be impossible to differentiate _which_ one failed.

melton-jason: (optional) 
On the topic of providing more information, I would love to have a link to the Merge Records dialog from the `record-merge-failed` notification. 
You should be able to construct the URL using the ids of the records (which is included in the payload). 
(The url is in the format `http://localhost/specify/overlay/merge/Agent/?records=9,10,11`, where records is an "array" of the ids)

See https://github.com/specify/specify7/blob/3b6a6ab6593bbb790a05a90d58bf35d2b484e51a/specifyweb/frontend/js_src/lib/components/Merging/index.tsx#L88-L90C12

Also, including the reason might be beneficial for debugging purposes. At the very least, I would hide the reason under a "dropdown" (details). 
If that is still too far, we should have the reason when the frontend is in DEBUG mode

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -119,6 +123,30 @@ export const notificationRenderers: IR<
...
+    const model = getModel(tableName);
+    const resource =
+      typeof model === 'object' ? new model.Resource({ id }) : undefined;
+    return (
+      resource !== undefined && (
+        <>
+          {mergingText.mergingSucceeded()}
+          <Link.NewTab href={`/specify/view/${tableName}/${id}`}>
+            <TableIcon label name={tableName} />
+            <FormattedResource asLink={false} resource={resource} />
```
melton-jason, line 144: `FormattedResource` as a link already opens the resource in a new tab. Why did you need to wrap it in a `Link.NewTab` and set the `FormattedResource` to not be a link. 
Was it it be consistent with other Notification links?

melton-jason (changes requested):

CarolineDenis:
> I cannot find a way to create a notification with `record-merge-pending` anymore. I don't see it referenced in the code. Since this is potentially not a problem, I will approve. Not sure if it was removed during the dev process.

@acwhite211 confirmed me this wasn't used

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
CarolineDenis: No you are right. I didn't have FormattedResource at first and didn't remove it once I implemented it

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -119,6 +124,67 @@ export const notificationRenderers: IR<
...
+    const tableName = notification.payload.table;
+    const collectionId = parseInt(notification.payload.collection_id);
+    const mergeName = notification.payload.name;
+    const schemaResource = new schema.models.Collection.Resource({
+      id: collectionId,
+    });
+    const [collectionName, setCollectionName] = React.useState<string | null>(
+      ''
+    );
+    schemaResource
```
maxxxxxdlp, line 137: did you forget to wrap this in a useEffect?

you are creating a new collection resource on each render, and then making a network request on each render, and that network request modifies the state. That's a recipe for a terrible performance or an infinite loop!

fetching data outside of useEffect is not allowed (though, recently added react server components would allow for this, but only on the back-end)
you don't know how many times the render function will get called

ideally, the render method should not have any side effects at all

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -119,6 +124,67 @@ export const notificationRenderers: IR<
...
+    const tableName = notification.payload.table;
+    const id = parseInt(notification.payload.new_record_id);
+    const ids = [JSON.parse(notification.payload.old_record_ids), id];
+    return (
+      <>
+        {mergingText.mergingHasFailed()}
+        <div className="flex items-center gap-2">
+          <TableIcon label name={tableName} />
+          <Link.NewTab
+            href={`/specify/overlay/merge/${tableName}/?${mergingQueryParameter}=${ids}`}
```
maxxxxxdlp, line 162: don't do this by hand. call `formatUrl` instead

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -119,6 +124,67 @@ export const notificationRenderers: IR<
...
+          </Link.NewTab>
+        </div>
+      </>
+    );
+  },
+  'record-merge-succeeded'(notification) {
+    const id = parseInt(notification.payload.new_record_id);
+    const tableName = notification.payload.table;
+    const model = getModel(tableName);
+    const resource =
```
maxxxxxdlp, line 174: do not create a new resource on each render! you muse call useMemo

`FormattedResource` makes a network request to fetch and format the resource every time the resource changes - which it does on every render here

maxxxxxdlp (changes requested):
Bad. Some basic React rules violated

grantfitzsimmons:
@melton-jason Can you review this PR sooner?

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -128,15 +130,17 @@ export const notificationRenderers: IR<
...
-      .fetch()
-      .then((resource) => setCollectionName(resource.get('collectionName')));
+    const [collectionName, setCollectionName] = React.useState<
+      string | null | undefined
+    >('');
+
+    React.useEffect(() => {
+      setCollectionName(collection?.collectionName);
+    }, [collection]);
+
```
maxxxxxdlp, line 143: why so complex?

```suggestion
const collectionName = collection?.collectionName;
```

and then you don't even need to create a variable and can use `collection?.collectionName` directly in the `return ...`

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -119,6 +126,74 @@ export const notificationRenderers: IR<
...
+    return (
+      <>
+        {mergingText.mergingHasStarted()}
+        <div className="flex items-center gap-2">
+          <TableIcon label name={tableName} />
+          <p>
+            {collectionName} {'-'} {mergeName}
```
maxxxxxdlp, line 150: Relying on JSX for spaces is very unsafe as Prettier does not try to preserve spaces when reformatting JSX, thus this can break in the future.

```suggestion
            {`${collectionName} - ${mergeName}`}
```

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -119,6 +126,74 @@ export const notificationRenderers: IR<
...
+    const tableName = notification.payload.table;
+    const id = parseInt(notification.payload.new_record_id);
+    const ids = [JSON.parse(notification.payload.old_record_ids), id];
+    return (
+      <>
+        {mergingText.mergingHasFailed()}
+        <div className="flex items-center gap-2">
+          <TableIcon label name={tableName} />
+          <Link.NewTab
+            href={formatUrl(`/specify/overlay/merge/${tableName}/`, {
```
maxxxxxdlp, line 166: (optional) this is a bit too complex. for readability, can you extract it into a variable, and just use that variable here?

that will make code more readable

maxxxxxdlp (changes requested):

melton-jason (approved):

PR #3864, opened 2023-07-28
Display a loading bar for the merging process

CarolineDenis:


specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -299,12 +307,16 @@ function MergeButton<SCHEMA extends AnySchema>({
...
 
+  const [loadingBar, setLoadingBar] = React.useState(false);
+
   return (
     <>
       {!saveBlocked ? (
         <>
           {noShowWarning ? (
-            <Submit.Blue form={formId}>{treeText.merge()}</Submit.Blue>
+            <Submit.Blue form={formId} onClick={() => setLoadingBar(true)}>
```
maxxxxxdlp, line 317: this isn't safe

if you click on a submit button, the form's onSubmit is only triggered if there are no validation errors
where as submit button's on click is called in every case.

thus if someone clicks this button while there are validation errors, they will be stuck with an infinite loading bar

maxxxxxdlp: why do you even call `<Status>` from this component?

calling it from `<Merging>` makes much more sense - that component knows when the merge process has started (`onMerge`) and has access to `mergeId`

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -198,12 +206,15 @@ function Merging({
...
           </Button.BorderedGray>
-          <MergeButton id={id} mergeResource={merged} mergeId={mergeId} />
+          <MergeButton id={id} mergeResource={merged} />
         </>
       }
       onClose={handleClose}
     >
       {typeof error === 'string' && <ErrorMessage>{error}</ErrorMessage>}
+      {loadingBar && (
+        <Status mergingId={mergeId} loadingBar={[loadingBar, setLoadingBar]} />
```
maxxxxxdlp, line 216: this doesn't need access to `loadingBar` at all - it's always true here

maxxxxxdlp: and then instead of `loadingBar` just call it `onStop` or `onAbort`

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -402,13 +416,19 @@ function Status({ mergingId }: { readonly mergingId: string }): JSX.Element {
   return (
     <Dialog
       buttons={
-        <Button.Danger onClick={undefined}>{wbText.stop()}</Button.Danger>
+        <Button.Danger onClick={() => setLoadingBar(false)}>
+          {wbText.stop()}
```
maxxxxxdlp, line 420: @grantfitzsimmons should it say "Stop" or "Abort" (like in the WB if I remember correctly)

maxxxxxdlp:
> @maxxxxxdlp , why is loadingBar doesn't pick up the change of mergeId and therefore doesn't display the Status dialog? It's like mergeId was never changing even if the setMergeId is triggered in the loading function...

in your useEffect, you set loadingBar to true on change to mergeId - you never set it to false (except if user clicks on the abort button), and so loading bar never disappears

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -361,6 +381,58 @@ function MergeButton<SCHEMA extends AnySchema>({
...
+          return undefined;
+        })
+        .catch(softFail);
+    fetchStatus();
+  }, [mergingId]);
+
+  return (
+    <Dialog
+      buttons={
+        <Button.Danger onClick={() => setLoadingBar(false)}>
```
maxxxxxdlp, line 419: when the button is clicked, a network request should be made notifying the back-end that user wants to abort, right?

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -361,6 +381,58 @@ function MergeButton<SCHEMA extends AnySchema>({
...
+          {wbText.stop()}
+        </Button.Danger>
+      }
+      className={{ container: dialogClassNames.narrowContainer }}
+      header={mergingText.mergeRecords()}
+      onClose={undefined}
+    >
+      {'test'}
```
maxxxxxdlp, line 427: a reminder to remove `"test"`

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -361,6 +381,58 @@ function MergeButton<SCHEMA extends AnySchema>({
   );
 }
 
+export function Status({
```
maxxxxxdlp, line 384: (optional) can you take a look WbStatus component and see if any code can be reused/encapsulated or if it's too diferent

maxxxxxdlp (changes requested):
You don't need `loadingBar` state at all - it only adds complication

loadingBar should only be visible if mergeId is not undefined, right?

so just do this:

```tsx
const [mergeId, setMergeId] = React.useState<string | undefined>(undefined);
// later:
mergeId === undefined
// and:
()=>setMergeId(undefined)
```

</details>

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -361,6 +381,58 @@ function MergeButton<SCHEMA extends AnySchema>({
...
+}: {
+  readonly mergingId: string;
+  readonly loadingBar: GetSet<boolean>;
+}): JSX.Element {
+  const [status, setStatus] = React.useState<string>();
+  const [total, setTotal] = React.useState<number>();
+  const [current, setCurrent] = React.useState<number>();
+
+  React.useEffect(() => {
+    const fetchStatus = () =>
```
maxxxxxdlp, line 396: why do you need to create a fetchStatus() function if you call it right away, and only in one place? just call ajax() right away, without creating a function

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -361,6 +381,58 @@ function MergeButton<SCHEMA extends AnySchema>({
...
+      void ajax(`/api/merging_status/${mergingId}/`, {
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        headers: { Accept: 'application/json' },
+      })
+        .then((data) => {
+          console.log(data);
```
maxxxxxdlp, line 402: remove console log before production

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
CarolineDenis: we decided abort, I just used this one until we have an abort API

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
CarolineDenis: yes but it doesn't exist yet...

CarolineDenis:
@maxxxxxdlp, I am facing the same issue where mergeId state is not being updated

CarolineDenis:
NOTES: the setError(undefined) on line 268 is not working neither. 
Error is undefined in this case as we hope it to be but only because the original value is 'undefined'. 
For the test we can do: setError('succeeded') and error will still be undefined.

CarolineDenis:
@melton-jason @realVinayak if you have input on this..

CarolineDenis:
NOTES: 
Tried: 
  `const wrapedSet = (value: string) => {
    setMergeId(value);
  };`
  added after `const [aborted, setAborted] = ...`
  And call `wrapedSet(response.data`) in the `.then` after `for (const clone of clones...)`

CarolineDenis:
NOTES: 
mergeId is now being set to the proper data :partying_face:
Alec is looking into fixing the issue with the API for the status as it is not working atm. 
I commented out 
`  React.useEffect(
    () => (ids.length < 2 ? handleClose() : undefined),
    [ids, handleClose]
  );`
  to not have the component unmounted, the issue now is that we can open the merging with only 1 agent. 
  Could we display the MerginDialog component directly from RecordMergingLink with a conditional instead of displaying it from an overlay route?

maxxxxxdlp (changes requested):

maxxxxxdlp:
sorry, I just realized that I left comments but forgot to publish them 😭

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -255,7 +268,7 @@ function Merging({
                 errorMode: 'dismissible',
               }
             ).then((response) => {
-              setMergeId(response.data);
+              setMergeId(JSON.parse(response.data));
               if (response.status === Http.NOT_ALLOWED) {
```
maxxxxxdlp, line 272: refactor to use if/else instead of if+return - will be cleaner and less bug-prone that way

maxxxxxdlp: how about instead of checking for Http.NOT_ALLOWED you check for everything that isn't what you expect - this way all errors will be caught

thus either `response.status !== Http.<what is the expected response code?>`
OR simply `!response.response.ok` (see https://developer.mozilla.org/en-US/docs/Web/API/Response/ok)

and use object destucturing to remove the need for weird looking `response.response` :
instead of ` ).then((response) => {` do ` ).then(({data,response}) => {`

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -265,8 +278,7 @@ function Merging({
               }
 
               setError(undefined);
-              // setLoadingBar(true);
-              handleClose();
+              // handleClose();
```
maxxxxxdlp, line 281: don't forget to remove commented out code

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -255,7 +268,7 @@ function Merging({
                 errorMode: 'dismissible',
               }
             ).then((response) => {
-              setMergeId(response.data);
+              setMergeId(JSON.parse(response.data));
```
maxxxxxdlp, line 271: don't do JSON.parse here
instead just change this:
```
	                headers: {
                  Accept: 'text/plain',
                },
```
to this

```
	                headers: {
                  Accept: 'application/json',
                },
```

this way ajax. will take care of parsing, and it will also ensure that back-end returned valid JSON response (and if not, take care of displaying proper error to the user)

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -122,19 +122,28 @@ export function MergingDialog(): JSX.Element | null {
   const handleDismiss = (dismissedId: number) =>
     setIds(ids.filter((id) => id !== dismissedId).join(','));
 
+  const [mergeId, setMergeId] = React.useState<string | undefined>(undefined);
```
maxxxxxdlp, line 125: you don't need this state in `<MergingDialog>` as you don't use it here
when I meant lifting the state, I meant lifting to `<Merging>` - because that's where it's used

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -361,6 +381,58 @@ function MergeButton<SCHEMA extends AnySchema>({
...
+  const [total, setTotal] = React.useState<number>();
+  const [current, setCurrent] = React.useState<number>();
+
+  React.useEffect(() => {
+    const fetchStatus = () =>
+      void ajax(`/api/merging_status/${mergingId}/`, {
+        // eslint-disable-next-line @typescript-eslint/naming-convention
+        headers: { Accept: 'application/json' },
+      })
+        .then((data) => {
```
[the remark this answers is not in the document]
maxxxxxdlp: edit: sorry for posting this late. but yeah, use object restructuring!

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -171,6 +183,16 @@ function Merging({
     true
   );
 
+  // const [mergeId, setMergeId] = React.useState<string | undefined>(undefined);
+
+  const [aborted, setAborted] = React.useState<boolean | 'failed' | 'pending'>(
```
maxxxxxdlp, line 188: try not mixing boolean and string types in the same state
why not use string types for all?
and also rename `aborted` to `state` to make it more appropriately named

maxxxxxdlp: edit: looks like this state is not used at all

I am not in your situation, so feel free to disregard it, but I would advise removing unused code as soon as possible instead of commenting it out.
if you think there is a chance you would need it, then create git commits, OR separate branches OR use git stash - all of those are excellent choices.
Reasons:
- Less code in the file that way while you are in the middle of debugging/futher development - thus either to comprehent
- No risk of forgetting to remove dead code before merging to production
- Easier for people who are reviewing as they don't have to spend time reviewing dead code

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -171,6 +183,16 @@ function Merging({
...
+    false
+  );
+
+  function handleAbort() {
```
maxxxxxdlp, line 192: don't need to create a function for this.
just create an inline function in the component:
```
onAbort={()=>setMergeId(undefined)}
```

maxxxxxdlp (changes requested):

maxxxxxdlp:
> Could we display the MerginDialog component directly from RecordMergingLink with a conditional instead of displaying it from an overlay route?

No. Reasons:
- By being on a separate URL, record merging is not dependent on query builer - you can call is easily from the Form Meta menu or the express search
- It has it's own URL, so it's integrated with the browser history - users can use back/forward arrows to close/reopen the dialog, as well as look in their browser history to see which records they merged
- Third-party devs could create tools for finding duplicate records, and then just linking to the record merging tool's URL - we don't provide proper ways of finding duplicates at the moment, but members already created some tools for finding duplicates, and they would appreciate not having to re-create the merging UI itself but just be able to easily reuse our interface

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -361,6 +381,58 @@ function MergeButton<SCHEMA extends AnySchema>({
...
+        .then((data) => {
+          console.log(data);
+          if (data === null) return undefined;
+          else {
+            // setStatus(data.status);
+            // setTotal(data.total);
+            // setCurrent(data.current);
+            globalThis.setTimeout(fetchStatus, 2000);
+          }
+          return undefined;
```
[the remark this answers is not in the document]
CarolineDenis: where does destructoredCalled is coming from?

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
CarolineDenis: Yes I wrote it already waiting for Alec to provide me with the abort API

CarolineDenis: "try not mixing boolean and string types in the same state
why not use string types for all?"
I took the example from the Status component we use for workbench.

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
CarolineDenis: I intended to call the abort API in the function handleAbort that's why I created that function

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -361,6 +393,99 @@ function MergeButton<SCHEMA extends AnySchema>({
...
+
+            setState({
+              status: data.taskstatus,
+              total: data.taskprogress.total,
+              current: data.taskprogress.current,
+            });
+            globalThis.setTimeout(fetchStatus, 2000);
+          }
+          return undefined;
+        })
```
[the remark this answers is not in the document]
melton-jason: When is `data` ever null? When there is no merging process corresponding to the `mergingId`?
If so, then please change the `ApiStatusResponse` type to reflect this.
You can still deconstruct the `data` object then, though it's benefit is more marginal in this case.
```ts
       .then(({ data }) => {
          if (data === null) return;
          const { taskstatus: taskStatus, taskprogress: taskProgress } = data;
          setState({
            status: taskStatus,
            total: taskProgress.total,
            current: taskProgress.current,
          });
          globalThis.setTimeout(fetchStatus, 2000);
        })
```

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -361,6 +393,99 @@ function MergeButton<SCHEMA extends AnySchema>({
   );
 }
 
+type StatusState = {
+  status: string;
```
melton-jason, line 386: This status can only be one of three strings, correct? 
Can you narrow its type then to:
```suggestion
  status: 'MERGING' | 'SUCCESS' | 'FAILED';
```

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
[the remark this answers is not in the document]
CarolineDenis: PS: I removed the data === null in one of the next commit

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
maxxxxxdlp: > I took the example from the Status component we use for workbench.

good reason. Workbench Mapper, Import and Status where the first React components in Specify. Despite code being refactored several times, some things are still not up to the code quality standard (i.e the use of `getState/setState` instead of `useCachedState` and the use of reducer/dispatch, and the use of custom select element, instead of a library)

specifyweb/frontend/js_src/lib/components/WorkBench/RemainingLoadingTime.tsx
```diff
@@ -38,6 +38,7 @@ export function RemainingLoadingTime({
     return undefined;
   }, [total, current]);
 
+  console.log(total, current, startTime, remainingTime);
```
maxxxxxdlp, line 42: don't forget to remove

maxxxxxdlp (commented):
Pushed a commit that cleans up things a bit

Left some comments, but most of them should already be fixed by the commit

Can you test and see if anything is not working? If so, ping me and I will launch specify for real and debug it (here I was doing code changes without opening specify)

also, I added a call to the abort_merging endpoint, but I belive that endpoint is not on this branch yet, and it might be renamed, so that request will likely fail for now

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -444,24 +428,29 @@ export function Status({
...
 
   return (
     <Dialog
       buttons={
-        <Button.Danger onClick={handleClose}>{wbText.stop()}</Button.Danger>
+        <Button.Danger
+          onClick={(): void =>
+            loading(
+              ping(`/api/specify/abort_merging/${mergingId}/`)
+                .then(handleClose)
```
CarolineDenis, line 440: Why (when?) using loading + ping over Ajax?

bronwyncombs (changes requested):
The size and placement of the progress bar dialog are remembered by the main merge records dialog. 

https://github.com/specify/specify7/assets/135047322/9971c3e4-cfb6-4e0f-bd92-c6a5b98591a1

---
Can we also...
- Automatically reload page when done so we don't see the dialog with abort button after completed. Will also solve going to 404 after completed merge. 
- Put "references" after count on dialog to explain the numbers

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
maxxxxxdlp: and because `ping` doesn't care about the response, you don't have to specify `headers: {Accept ...}`
so it's just used to write less code in this common case and to signal extra clear to those reading code that response does not matter (unless it's an error, in which case it will take care of it on it's own)

ping() does return one thing though (`number`) - the HTTP response code, just in case you need that

maxxxxxdlp:
> Automatically reload page when done so we don't see the dialog with abort button after completed. Will also solve going to 404 after completed merge.

hmm, it was supposed to re-run the query results after a merge. seems like that code broke

specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
```diff
@@ -65,6 +65,11 @@ export function Status({
 
   const loading = React.useContext(LoadingContext);
 
+  if (state.status === 'SUCCESS') {
+    handleClose();
+    globalThis.location.reload();
+  }
+
```
maxxxxxdlp, line 72: There already is a way for query builder to know that merging has completed:

https://github.com/specify/specify7/blob/8995d00014b85cbb8feccd0b8378208b910b3512/specifyweb/frontend/js_src/lib/components/Merging/index.tsx#L68-L73

It is supposed to re-run the query after merge. If it doesn't, it's a bug - please look into it

Also, you would notice that that code is rendered under the query builder URL, not merging URL, thus it has access to a callback for re-running the query and other goodies

maxxxxxdlp (changes requested):
almost!

specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
maxxxxxdlp: Also, re-running the query is far better than reloading the page:
- What if the query has unsaved changes?
- Plus, re-running the query takes far less time than reloading the page and then re-running the query

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -250,13 +246,17 @@ function Merging({
...
               }
             ).then(({ data, response }) => {
-              if (response.ok) {
-                setMergeId(data);
-                setError(undefined);
-                clones.forEach((clone) =>
-                  resourceEvents.trigger('deleted', clone)
-                );
-              } else setError(data);
+              setMergeId(data);
```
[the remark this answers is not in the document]
maxxxxxdlp: hmm, is this piece of code giving you errors?
do you need help?

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -60,14 +60,16 @@ export function RecordMergingLink({
...
     if (oldRecords.current === undefined && records !== undefined)
       needUpdateQueryResults.current = false;
     // Detect agent merging dialog getting closed after some records are merged
     else if (
       records === undefined &&
-      oldRecords.current !== undefined &&
-      needUpdateQueryResults.current
+      oldRecords.current !== undefined
+      // &&
+      // needUpdateQueryResults.current
```
maxxxxxdlp, line 72: is this not working correctly?
do you need help, or just time to work on this?

sorry for bothering you about this, just checking in...

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -246,17 +243,17 @@ function Merging({
...
-              if (!response.ok) {
-                setError(data);
-              } else {
-                return;
-              }
+              /*
+               * Because we can not pass down anything from the Query Builder
+               * as a prop, this is needed to rerun the query results once
+               * the merge completes.
+               * (the RecordMergingLink component is listening to the event)
```
maxxxxxdlp, line 252: (not a request for change, just commenting)

> (the RecordMergingLink component is listening to the event)
...and in the future ORM version, more things could be listening for deleted event. That would allow to solve cases when i.e you delete a query, but query list dialog still shows it - though in the best case, the query dialog will just call `useCollection(...)` hook, and that hook will take care of fetcing, fetching more (for pagination), errors and listening for records being deleted

maxxxxxdlp (approved):

grantfitzsimmons:
Merge this as soon as we can!

bronwyncombs (changes requested):
The progress bar doesn't appear for quick merges

https://github.com/specify/specify7/assets/135047322/92146cb1-6ca6-4d98-8a6b-486d050da552

---

Bigger merges do get the progress bar after a loading dialog. 

However, it will tell you success but then send a fail notification

https://drive.google.com/file/d/1zehhB45CYmOHkdRnEZ_vK3JtxUz90NCz/view?usp=sharing
(drive for privacy)

---

Also, the progress bar time countdown will display with dashes when resized:

Maybe move it to the top or bottom instead of the side? 

https://drive.google.com/file/d/1-BGoZFdP0lktlbZ1qybW10adoySrK9Q-/view?usp=sharing

melton-jason:
Can this be tested again? 
The bug causing the dialog to close immediately after beginning the merge with exactly two agents should be fixed. 

And the remaining time should now also be displayed below the loading bar

bronwyncombs (changes requested):
I get the correct dialog when merging two agents now! Remaining time doesn't show though

https://github.com/specify/specify7/assets/135047322/b3518949-38df-47b9-a2a9-b5254e3baa55

CarolineDenis:
> I get the correct dialog when merging two agents now! Remaining time doesn't show though
> 
>  Screen.Recording.2023-08-14.at.11.51.48.AM.mov

if you increase the height of the dialog can you see it? Or also sometime if the merge is too fast there is no remaining time showing up

CarolineDenis:
then I would say it's because it was too fast

maxxxxxdlp (approved):

melton-jason (approved):
Can this be merged after bronwyn approves it? 
As far as I know, the only remaining problem that needs to be addressed related to this is to fix the bug where the backend returns a SUCCESS when calling the merge status endpoint even when the merge fails. 
However, this PR has gotten quite lengthy and the fix for that should probably be its own PR, in my opinion.

bronwyncombs (changes requested):
Still not seeing the time remaining 

https://github.com/specify/specify7/assets/135047322/4cca1e25-336e-433c-8ea1-8efbcd709a21

melton-jason:
That part was not addressed. 

Here's the problem:
- When the `Proceed/Merge` button is clicked, the backend begins merging

- The frontend (due to a limitation of us being on HTTP 1.1) can only have 6 concurrent requests at a time (See #2608)

- Thus the merging can complete on the backend before the request to fetch the status (and thus the progress/time remaining) ever resolves

- On the frontend, it can not estimate the time remaining or progress without either making assumptions (which may be highly inaccurate) or _actually_ receiving a response from the backend while the agents are in the process of merging
  - In other words, the frontend goes straight from starting the merge to fetching the status, which is now finished

CarolineDenis:
> That part was not addressed.
> 
> Here's the problem:
> 
> * When the `Proceed/Merge` button is clicked, the backend begins merging
> * The frontend (due to a limitation of us being on HTTP 1.1) can only have 6 concurrent requests at a time (See [Add HTTP2 (and HTTP3) support in NGINX containers #2608](https://github.com/specify/specify7/issues/2608))
> * Thus the merging can complete on the backend before the request to fetch the status (and thus the progress/time remaining) ever resolves
> * On the frontend, it can not estimate the time remaining or progress without either making assumptions (which may be highly inaccurate) or _actually_ receiving a response from the backend while the agents are in the process of merging
>   
>   * In other words, the frontend goes straight from starting the merge to fetching the status, which is now finished

added a time out in the back end for testing and time remaining is displayed :ok_hand:

PR #3866, opened 2023-07-31
Catch case where no old_model_ids were given - [issue-3857]

acwhite211:
Add a guard clause to handle the case where no old_model_ids are given to the record merging API request.  The back-end should respond with an error explaining the need to include old record IDs.  This PR resolves issue-3857 where `old_model_ids=[]` caused issues when sent in the record merging request.  Further investigation might needed to determine how a situation on the front-end was formed so that a request was made to the back-end without

### DIFF
--- a/specifyweb/specify/views.py
+++ b/specifyweb/specify/views.py
@@ -733,6 +733,9 @@ def record_merge(
     data = json.loads(request.body)
     old_model_ids = data['old_record_ids']
     new_record_data = data['new_record_data'] if 'new_record_data' in data else None
+
+    if old_model_ids is None or len(old_model_ids) < 1:
+        return http.HttpResponseBadRequest('There were no old record IDs given to be replaced by the new ID.')
     
     background = True
     if 'background' in data:


grantfitzsimmons:
@bronwyncombs:

Can you try to recreate the issue again with this PR?

> Testing instructions:
> 
> For Issue https://github.com/specify/specify7/issues/3801
> 
> I'll recommend running the below steps for at least some of Deaccession, Disposal, Loan, Accession, Gift, Borrow tables. Also test on both v7.9-dev and this branch. v7.9-dev shouldn't be able to merge these, and this branch should.
> 
> Later on, testers can go wild and merge anything into anything - and as long as it is not an actual business rule exception, it should work. 
> 
> Steps:
> 1. Make an interaction via the interactions form. I'll call it accession `001` for these steps. Make two new accession agents. I'll recommend creating two fresh new agents (but you can select a previous agent too). You can also make more than two accession agents. For this example, two new agents are `agent 001` and `agent 002`. **Important** that both of these agents should be assigned same role in the accession.
> 2. Make another accession, and select `agent 002` and assign any role to this agent.
> 3. Go into query builder, select agents `agent 001` and `agent 002`. 
> 4. Merge `agent 001` into `agent 002`. The order is important to recreate issue.
> 5. This should fail in v7.9-dev, with an error saying `found multiple records violating uniqueness constraints`. In this branch, it should work without any errors.
> 
>

maxxxxxdlp (approved):

acwhite211:
Not sure why the front-end tests are failing, since there are no front-end changes in this PR.

acwhite211:
@melton-jason  Looks like this commit to the `v7.9-dev` branch is what caused the tests to fail.  Need to fix that so we can merge other PRs.

PR #3877, opened 2023-07-31
wrap items in detail view in query builder 

### LINKED ISSUE #3859: Query items are not forced to the current window width
Proper behavior can be seen in `v7.8.13`

## `v7.9`
See that the items are cut off on the right and it can be scrolled
image

## `v7.8.13`
image

macOS, Firefox

CarolineDenis:
Fixes #3859

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Fields.tsx
@@ -182,7 +182,7 @@ export function QueryFields({
   return (
     <Ul
       className={`
-          flex-1 items-center overflow-y-auto
+          items-center overflow-y-auto sm:flex-1
           ${
             isBasic
               ? 'grid grid-cols-[auto,auto,1fr,auto] content-start items-start gap-y-2 gap-x-2'
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
@@ -286,7 +286,7 @@ export function QueryLine({
         {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
         <div
           className={`
-          flex flex-1 flex-wrap gap-2 overflow-scroll print:gap-1
+          flex flex-1 flex-wrap gap-2 overflow-auto print:gap-1
           ${field.filters.length > 1 ? 'items-baseline' : 'items-center'}
           ${isFocused ? 'rounded bg-gray-300 dark:bg-neutral-700' : ''}
           ${isBasic ? 'contents' : ''}
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
@@ -286,7 +286,7 @@ export function QueryLine({
         {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
         <div
           className={`
-          flex flex-1 flex-wrap gap-2 print:gap-1
+          flex flex-1 flex-wrap gap-2 overflow-scroll print:gap-1
           ${field.filters.length > 1 ? 'items-baseline' : 'items-center'}
           ${isFocused ? 'rounded bg-gray-300 dark:bg-neutral-700' : ''}
           ${isBasic ? 'contents' : ''}
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
@@ -360,7 +360,7 @@ export function QueryLine({
               className={
                 field.filters.length > 1
                   ? 'flex flex-col gap-2'
-                  : 'flex items-center gap-2'
+                  : `flex items-center gap-2 ${isBasic ? '' : ' flex-wrap'}`
               }
             >
               {field.filters.map((filter, index) => (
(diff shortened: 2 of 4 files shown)


melton-jason (approved):
@specify/ux-testing get testing!
This change should only effect detailed view. 
Particularly, test QueryBuilder lines on narrow screens and see how the content overflows and wraps. 
Should be compatible on all major browsers!

[tailwind docs](https://tailwindcss.com/docs/flex-wrap#wrap-normally)
[css docs](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-wrap)

maxxxxxdlp (approved):

bronwyncombs (changes requested):
Firefox has some issues but Chrome and Safari look good! 

https://github.com/specify/specify7/assets/135047322/fb989ea2-0b6f-4211-9d4f-5efcd3726182

grantfitzsimmons (changes requested):
See @bronwyncombs's message

grantfitzsimmons:
@CarolineDenis Are you going to try and resolve this?

CarolineDenis:
Notes: 
The issue where the line doesn't seem to take the full width when in the narrowest screen is present on Firefox but also on Chrome. This wasn't happening in 7.13 because basic view wasn't a feature yet. 
Basic view changed the way the divs are organized.

grantfitzsimmons:
The issue that @bronwyncombs demonstrated is not in basic view– are we going to try and resolve that?

CarolineDenis:
NOTES: 
Met with @grantfitzsimmons, decision: 
==> hide up and down arrows only for narrow screen when in detailed view. 
Still present in all other screen sizes and always present for basic view

maxxxxxdlp (approved):

grantfitzsimmons:
@bronwyncombs found that the cut-off for small screen width may be too large– can we make it hide the up and down arrows when the screen width is smaller?

CarolineDenis:
> @bronwyncombs found that the cut-off for small screen width may be too large– can we make it hide the up and down arrows when the screen width is smaller?

I am not sure I get that, can I have a brief drawing on a screenshot?

bronwyncombs:
@CarolineDenis, this is the current behavior. We just think it's removing the buttons disappear a little too early, the window should be smaller before removing buttons. The size I make the window at the end of this clip is around where I imagine the buttons would go away. 


https://github.com/specify/specify7/assets/135047322/e0b961ed-0a87-42cd-9944-90752f1e8358

Let me know if you have any questions!

CarolineDenis:
> @CarolineDenis, this is the current behavior. We just think it's removing the buttons disappear a little too early, the window should be smaller before removing buttons. The size I make the window at the end of this clip is around where I imagine the buttons would go away.
> 
>  Screen.Recording.2023-08-04.at.3.56.37.PM.mov 
> 
> Let me know if you have any questions!

got it thanks

CarolineDenis:
There is a slight difference with what I just pushed but we might want to hard code a pixel break point and hide only when hit?

maxxxxxdlp:
> There is a slight difference with what I just pushed but we might want to hard code a pixel break point and hide only when hit?

I would prefer to not hardcode an arbitrary value, but if you really have to, here is the way to:

https://github.com/tailwindlabs/tailwindcss/discussions/8261#discussioncomment-3251527

bronwyncombs:
Right now, when the up and down arrows disappear, the bottom row functions (Reveal Hidden Form Fields, Distinct, Count, and Query) are moved down the page quite a bit. 

https://github.com/specify/specify7/assets/135047322/72fe1e50-700f-4e3c-bfbf-4807f5592c70

bronwyncombs (changes requested):
Browsers left to right: Chrome, Safari, Firefox (all in smallest size)

- Chrome and Safari still have the blank space between the builder and results. Button alignment is good, though.
- Firefox still causes buttons to misalign with small window. It also gets much smaller than the others so the Query button shifts to a row below the other functions. 
- As for when the up/down buttons disappear, it seems fine, but I'll leave that call to @grantfitzsimmons 

https://github.com/specify/specify7/assets/135047322/6b72416a-970a-474e-9c32-3b6e2851ea3e

up/down buttons:

https://github.com/specify/specify7/assets/135047322/9da732b2-bf34-4a39-bb16-44a87b9c49d5

CarolineDenis:
NOTES: I cannot reproduce=> Firefox still causes buttons to misalign with small window. It also gets much smaller than the others so the Query button shifts to a row below the other functions.

maxxxxxdlp (approved):

bronwyncombs (changes requested):
Blank space between mapper and results is fixed! 

Part of the problem: Firefox allows the window to get smaller than our longer field names (i.e. Collecting Event Attachments, Reference Work Attachments) which causes the staggered button appearance. I'm not sure how we would want work around that. 

https://github.com/specify/specify7/assets/135047322/389436f8-df3c-4e8d-93fd-7eb3b91b5ba0

CarolineDenis:
I can add an overflow-scroll @melton-jason @bronwyncombs here is an example: 

https://github.com/specify/specify7/assets/108160931/cf03643e-54cb-4efb-ad6f-353a00374c97

CarolineDenis:
@bronwyncombs, just pushed the code with the overflow for you to try :)

bronwyncombs (approved):
Overflow scroll looks good to me! Thank you for this solution! 😄 

https://github.com/specify/specify7/assets/135047322/3d1d26da-a757-4190-992a-86ada1f1da46

maxxxxxdlp (changes requested):
`overflow-scroll` should almost never be used
prefer `overflow-auto` instead

`overflow-scroll` forced the scrollbars to be present even if content is not overflowing - the only case when that's desirable is if content size changes, and you want to prevent the content shift from scrollbar appearing/disapearing dynamically - though even that can be fixed better by using a [scrollbar-gutter](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter)

see also more info on the issue - https://kilianvalkhof.com/2021/css-html/you-want-overflow-auto-not-overflow-scroll/

the reason scrollbar wasn't visible on the screen recordings posted in the last few comments is that recent versions of chrome make scrollbars invisible when not in use - but that behavior is not the case for all browsers on all systems, and can be customized by the user.

PR #3878, opened 2023-08-01
Remove saveblockers when -to-many relationship is removed when merging

### LINKED ISSUE #3872: Adding a -to-many relationship then removing it blocks merging
https://github.com/specify/specify7/assets/37256050/c4c37076-a9fb-4426-82d0-9de08b634c02

`herb_rbge` database

http://localhost/specify/overlay/merge/Agent/?records=47121%2C47120%2C47119%2C47118%2C47117%2C47116%2C47115%2C47114%2C47113%2C47112%2C47111%2C47110%2C47109%2C47108%2C47107%2C47106%2C47105

**Steps to Recreate:**
1. Uncheck "Show conflicting fields only"
1. Go to the **Agent Attachments** button
2. Add a blank record
3. Remove it
4. See merge is now blocked by the red button

melton-jason:
Fixes #3872 

> This issue occurred whenever a saveblocker was added to the dependent resource (in this case, ordinal and agent are required for an AgentAttachment).
> As soon as the AgentAttachment resource is created, it adds the saveblockers for all of the required fields.
> To handle -to-many resources, we create a list (array) based off of the resources in Backbone's internal array for that collection.
> 
> The trash icon just removed the resource from the array we created.
> We have to tell Backbone that we are deleting the Resource we created as well, which will remove the saveBlockers.

https://github.com/specify/specify7/issues/3872#issuecomment-1659355496

TO TEST: 
- Open up a -to-many relationship Subview in the Merging Dialog
- Get the record in a state where it has saveBlockers
  - An easy method, and the one showcased in https://github.com/specify/specify7/issues/3872#issue-1829828940 is to leave a required field blank
- Press the trash icon in the Subview to remove the record
- The Agent record should be able to be merged, and the subview button shouldn't be red/invalid

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
@@ -218,9 +218,10 @@ function MergeDialog({
                   : undefined
               }
               resources={children.map((record) => record[index])}
-              onRemove={(): void =>
-                setMergedRecords(removeItem(mergedRecords, index))
-              }
+              onRemove={(): void => {
+                mergedRecords[index].destroy();
+                setMergedRecords(removeItem(mergedRecords, index));
+              }}
               onSlide={(columnIndex, direction): void => {
                 if (columnIndex === 0)
                   setMergedRecords(moveItem(mergedRecords, index, direction));


maxxxxxdlp (approved):
good catch
and thanks for the detailed PR description

melton-jason:
The most ideal fix would be to pass the collection down to where it is accessibile to the `onMerge` function of the SubviewLine. Then we can call `collection.remove([resource])`-- this is currently what the Subview in the DataEntry Forms does. 
However, that might complicate the code a little, and the changes here seem functionally identical.

bronwyncombs (approved):
Working! Thanks! 


https://github.com/specify/specify7/assets/135047322/bc660364-70f4-43ef-8aaf-101db41ab806

PR #3893, opened 2023-08-02
Add clear all button in notifications

### LINKED ISSUE #3890: Clear All Notifications Button
It would be great if we had a clear all notifications button. It wasn't really an issue when only Work Bench and CSV exports were using it. However, with Agent Merging now using it and it being a fairly common thing to do. They can easily stack up. For example, using Edinburgh's Agent Merge Testing Spreadsheet easily gave me 10+

Screenshot 2023-08-02 at 11 52 32 AM

CarolineDenis:
Fixes #3890


specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -134,6 +134,21 @@ export function Notifications({
...
+
+  const onClearAll = () => {
+    if (notifications !== undefined) {
+      notifications.forEach((notification) => {
+        ping('/notifications/delete/', {
+          method: 'POST',
+          // eslint-disable-next-line @typescript-eslint/naming-convention
+          body: formData({ message_id: notification.messageId }),
+          errorMode: 'dismissible',
+        }).then(f.void);
```
melton-jason, line 146: Why is this `.then(f.void)` needed? The Promise containing the `ping` should already be considered a void function, because it does not return anything (undefined).

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -134,6 +134,21 @@ export function Notifications({
...
+      });
+      setNotifications([]);
+    }
```
melton-jason, line 149: For readability, it may be better to use a guard clause here instead of having all of the code under that if statement

```suggestion
   if (notifications === undefined) return;

    notifications.forEach((notification) => {
      ping('/notifications/delete/', {
        method: 'POST',
        // eslint-disable-next-line @typescript-eslint/naming-convention
        body: formData({ message_id: notification.messageId }),
        errorMode: 'dismissible',
      }).then(f.void);
    });
    setNotifications([]);
```

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -200,6 +215,11 @@ export function Notifications({
               />
             </ErrorBoundary>
           ))}
+          {notifications.length > 0 && (
+            <Button.Fancy onClick={(): void => onClearAll()}>
+              {'Clear All'}
```
melton-jason, line 220: Make sure to localize this string!
We already have a `Clear all` string localized that you can use. 

https://github.com/specify/specify7/blob/2a259ca22b439ba71b9313705cb2a253ac8edd6e/specifyweb/frontend/js_src/lib/localization/common.ts#L88-L95

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -134,6 +134,21 @@ export function Notifications({
           count: notifications.length,
         })
       : notificationsText.notificationsLoading();
+
+  const onClearAll = () => {
```
maxxxxxdlp, line 138: please use `function()` syntax instead for any function that requires `{` and `}` (more than one statement)
use `const` for any function that does not need `{` and `}` (single statement)

or is this consistency too much trouble? @melton-jason what do you think?

personally, I find it helpful as by just seeing const vs function I can immediately tell whether I am dealing with a small function or not.

Ben also preferred the `function` syntax as that way it's easier to see that this is a function, rather than just regular `const` variable

maxxxxxdlp, line 138: can you call this `handleClearAll` instead?
again consistency...

on*** when something emits events
handle*** when something handles an emitted event
quite common React convention

maxxxxxdlp (changes requested):
few things in addition to what Jason mentioned

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
melton-jason: I agree that consistency is important! All throughout the frontend code base, we _should_ maintain the consistency that has been upheld.
 
Personally, I would lean more towards Ben's convention: I associate `const` variables with non-function primitives and their derivatives (Can't explain this right now, but this is probably a remnant of my familiarity with languages like Java and C#). 
Then again, if the function is an arrow function without brackets, then you will be able to infer the type quickly because it shoul'd not take up much space.

melton-jason (approved):
Code looks good now!

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
maxxxxxdlp: > Personally, I would lean more towards Ben's convention: I associate const variables with non-function primitives and their derivatives (Can't explain this right now, but this is probably a remnant of my familiarity with languages like Java and C#).

I see merits in that change. Feel free to adapt any sane convention, as long as it's consistent (meaning if you want to change a convention, you better refactor all the places OR find an eslint rule that will do the refactor for you OR write a codemod to automate the refactor)

specifyweb/frontend/js_src/lib/components/Header/Notifications.tsx
```diff
@@ -200,6 +214,11 @@ export function Notifications({
               />
             </ErrorBoundary>
           ))}
+          {notifications.length > 0 && (
+            <Button.Fancy onClick={(): void => handleClearAll()}>
```
maxxxxxdlp, line 218: can replace `(): void => handleClearAll()` with just `handleClearAll`

maxxxxxdlp (approved):

carlosmbe (changes requested):
Can we put somewhere more easily noticeable. Like parallel the close button or on top. Otherwise, it's a bit hard to notice.

carlosmbe (changes requested):
If there's only one notification. We shouldn't have the button. Otherwise, it seems great.

maxxxxxdlp (approved):

bronwyncombs (changes requested):
- Could we make it less wide and add this to the bottom left of the dialog, in-line with the Close button? 
- It should be tied to button color preferences, maybe as a "Secondary"?
- Do we think it's consistent to keep this as a button that's highlighted when scrolled over?
- Could we also remove the slight border on the top?


https://github.com/specify/specify7/assets/135047322/644bf07b-bfb8-4245-bb6e-1cb934bdcb04

maxxxxxdlp (approved):

carlosmbe (approved):

PR #3894, opened 2023-08-02
Abort a currently running merge task

acwhite211:
An API for aborting a currently running record merging celery task.

### DIFF
--- a/specifyweb/specify/urls.py
+++ b/specifyweb/specify/urls.py
@@ -11,6 +11,7 @@ urlpatterns = [
     # replace record
     url(r'^specify/(?P<model_name>\w+)/replace/(?P<new_model_id>\d+)/$', views.record_merge),
     url(r'^specify/merging_status/(?P<merge_id>[0-9a-fA-F-]+)/$', views.merging_status),
+    url(r'^specify/abort_merge/(?P<merge_id>[0-9a-fA-F-]+)/$', views.abort_merge_task),
 
     # the main business data API
     url(r'^specify_schema/openapi.json$', schema.openapi),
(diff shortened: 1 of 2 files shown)


specifyweb/specify/views.py
```diff
@@ -926,3 +928,57 @@ def merging_status(request, merge_id: int) -> http.HttpResponse:
...
+    merge = Spmerging.objects.get(taskid=merge_id)
+    if merge is None:
+        return http.HttpResponseNotFound(f'The merge task id is not found: {merge_id}')
+
+    if merge.taskid is None:
+        return http.JsonResponse(None, safe=False)
+
+    task = record_merge_task.AsyncResult(merge.taskid)
+    
+    if task.state == 'PENDING' or task.state == 'STARTED':
```
maxxxxxdlp, line 973: (optional) use enums?

specifyweb/specify/urls.py
```diff
@@ -11,6 +11,7 @@
     # replace record
     url(r'^specify/(?P<model_name>\w+)/replace/(?P<new_model_id>\d+)/$', views.record_merge),
     url(r'^specify/merging_status/(?P<merge_id>[0-9a-fA-F-]+)/$', views.merging_status),
+    url(r'^specify/abort_merge/(?P<merge_id>[0-9a-fA-F-]+)/$', views.abort_merge_task),
```
maxxxxxdlp, line 14: why not put all merging APIs into a separate namespace?

`/specify/merge/abort/.../`?
`/specify/merge/status/...//`?
would also avoid the weird `_` in the path (we might be the only ones with `_` in the API URLs)

until we release merging for real it's easier to do such a change.

> REST API resource names
A resource in a REST API is always represented as the plural of an entity that is exposed by the API.
Resource names exposed in a REST API should use all lowercase characters.
Resource names may include hyphens.
Resource names should not include underscores or other punctuation (sole exception is the hyphen).

https://specs.openstack.org/openstack/api-wg/guidelines/naming.html

maxxxxxdlp (changes requested):

CarolineDenis (approved):

PR #3906, opened 2023-08-04
Change FormattedResource link to new tab link component to avoid fetching stale record

### LINKED ISSUE #3887: 404 in background when notifications are open with a deleted agent referenced
**Describe the bug**
After merging an agent and opening the notifications dialog from the navigation menu, the query page in the background changes to 404. 

**To Reproduce**
1. Merge agent
2. Click on Notifications
5. See error

**Expected behavior**
Should remain on query builder page 

**Screenshots**

https://github.com/specify/specify7/assets/135047322/8262902a-3f92-407c-b7a8-b581b1ac149f

**Desktop:**
 - macOS, chrome, v7.9

**Database Name:** `herb_rbge_3_31_23`

CarolineDenis:
Fixes #3887

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/legacyTypes.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/legacyTypes.ts
@@ -171,7 +171,10 @@ export type SpecifyResource<SCHEMA extends AnySchema> = {
     readonly errorOnAlreadySaving?: boolean;
   }): Promise<SpecifyResource<SCHEMA>>;
   destroy(): Promise<void>;
-  fetch(): Promise<SpecifyResource<SCHEMA>>;
+  fetch(options?: {
+    success?: (model: SpecifyResource<SCHEMA>, resoonse: unknown) => void;
+    error?: (model: SpecifyResource<SCHEMA>, response: unknown) => void;
+  }): Promise<SpecifyResource<SCHEMA>>;
   viewUrl(): string;
   isNew(): boolean;
   clone(cloneAll: boolean): Promise<SpecifyResource<SCHEMA>>;
--- a/specifyweb/frontend/js_src/lib/components/DataModel/legacyTypes.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/legacyTypes.ts
@@ -171,10 +171,7 @@ export type SpecifyResource<SCHEMA extends AnySchema> = {
     readonly errorOnAlreadySaving?: boolean;
   }): Promise<SpecifyResource<SCHEMA>>;
   destroy(): Promise<void>;
-  fetch(options?: {
-    success?: (model: SpecifyResource<SCHEMA>, resoonse: unknown) => void;
-    error?: (model: SpecifyResource<SCHEMA>, response: unknown) => void;
-  }): Promise<SpecifyResource<SCHEMA>>;
+  fetch(): Promise<SpecifyResource<SCHEMA>>;
   viewUrl(): string;
   isNew(): boolean;
   clone(cloneAll: boolean): Promise<SpecifyResource<SCHEMA>>;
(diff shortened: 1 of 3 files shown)


specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -164,22 +163,22 @@ export const notificationRenderers: IR<
   'record-merge-succeeded'(notification) {
     const id = parseInt(notification.payload.new_record_id);
     const tableName = notification.payload.table;
-    const model = getModel(tableName);
-    const resource = React.useMemo(
-      () =>
-        typeof model === 'object' ? new model.Resource({ id }) : undefined,
-      [model, id]
-    );
+    const newRecordId = parseInt(notification.payload.new_record_id);
```
maxxxxxdlp, line 166: could use use `Number.parseInt` in these two places here instead of `parseInt`? thanks

eslint: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-number-properties.md

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
```diff
@@ -164,22 +163,22 @@ export const notificationRenderers: IR<
...
-            <FormattedResource asLink={true} resource={resource} />
-          </div>
-        </>
-      )
+      <>
+        {mergingText.mergingHasSucceeded()}
+        <div className="flex items-center gap-2">
+          <TableIcon label name={tableName} />
+          <p>{`${tableName} #${newRecordId}`}</p>
+          <Link.NewTab
```
maxxxxxdlp, line 173: while this solves the problem here, a similar issue could happen in other places.
as a cleaner solution, I suggest undoing this change, and instead fixing `<FormattedResource />` to handle 404 gracefully, instead of displaying a 404 page

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/DataModel/legacyTypes.ts
```diff
@@ -171,7 +171,10 @@ export type SpecifyResource<SCHEMA extends AnySchema> = {
     readonly errorOnAlreadySaving?: boolean;
   }): Promise<SpecifyResource<SCHEMA>>;
   destroy(): Promise<void>;
-  fetch(): Promise<SpecifyResource<SCHEMA>>;
+  fetch(options?: {
+    success?: (model: SpecifyResource<SCHEMA>, resoonse: unknown) => void;
+    error?: (model: SpecifyResource<SCHEMA>, response: unknown) => void;
+  }): Promise<SpecifyResource<SCHEMA>>;
```
maxxxxxdlp, line 177: hmm, I don't see these being used in this PR
Also, I don't know how well these would work since these callbacks are handled by the backbone.
You might get by by using `hijackBackboneAjax` instead - see it's usages

melton-jason: Yes, I was looking into possible solutions for this and came across those callbacks mentioned in the Backbone docs. 
I had figured it would be helpful to include them, even if they were not used anywhere (to save someone the trouble of looking through the Backbone docs).

However, I would prefer using hijackBackboneAjax to handle such cases anyways, to remain consistent and have more control!

specifyweb/frontend/js_src/lib/components/Header/NotificationRenderers.tsx
melton-jason: Fixed in 408a413afc0698e170fd71cbc9e578c3a441a9af

maxxxxxdlp (approved):

bronwyncombs (approved):
Looks good! No 404 when notifications are opened. Links to merged agents no longer go stale and are replaced with newest agent link. 

https://github.com/specify/specify7/assets/135047322/c244ea60-54cc-4705-b191-b0289b31093a

After merge with the name changed:

https://github.com/specify/specify7/assets/135047322/056704b5-35f8-4fcf-8bc4-d992a0f72f90

melton-jason (approved):

PR #3911, opened 2023-08-11
Fix progress functionality for recursive merges

acwhite211:
Fix error so that the total is updated during a recursive merge

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
@@ -12,6 +12,7 @@ import { Label } from '../Atoms/Form';
 import { Progress } from '../Atoms';
 import { commonText } from '../../localization/common';
 import { icons } from '../Atoms/Icons';
+import { dialogIcons } from '../Atoms/Icons';
 
 const statusLocalization = {
   FAILED: mergingText.mergeFailed(),
@@ -96,13 +97,11 @@ export function Status({
       header={statusLocalization[state.status]}
       onClose={undefined}
       icon={
-        state.status === 'PENDING' ? (
-          icons.cog
-        ) : state.status === 'MERGING' ? (
-          <span className="text-red-500">{icons.exclamationCircle}</span>
-        ) : (
-          <span className="text-green-500">{icons.checkCircle}</span>
-        )
+        state.status === 'PENDING'
+          ? icons.cog
+          : state.status === 'MERGING'
+          ? dialogIcons.error
+          : dialogIcons.success
       }
     >
       <Label.Block aria-atomic aria-live="polite" className="gap-2">
--- a/specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
@@ -12,6 +12,7 @@ import { Label } from '../Atoms/Form';
 import { Progress } from '../Atoms';
 import { RemainingLoadingTime } from '../WorkBench/RemainingLoadingTime';
 import { commonText } from '../../localization/common';
+import { data } from 'jquery';
 
 const statusLocalization = {
   FAILED: mergingText.mergeFailed(),
@@ -66,6 +67,15 @@ export function Status({
 
   const loading = React.useContext(LoadingContext);
 
+  const percentage = Math.round((state.current / state.total) * 100);
+  console.log(
+    state.total,
+    state.current,
+    state.current / state.total,
+    (state.current / state.total) * 100,
+    percentage
+  );
+
   return (
     <Dialog
       buttons={
@@ -99,10 +109,7 @@ export function Status({
           {state.status === 'MERGING' && (
             <>
               <Progress max={state.total} value={state.current} />
-              <RemainingLoadingTime
-                current={state.current}
-                total={state.total}
-              />
+              <p>{percentage}%</p>
             </>
           )}
         </div>
--- a/specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
@@ -143,9 +143,7 @@ export function Status({
         </div>
       </Label.Block>
       {state.status === 'FAILED' ? (
-        <>
-          <p>{mergingText.mergingWentWrong()}</p>
-        </>
+        <p>{mergingText.mergingWentWrong()}</p>
       ) : null}
     </Dialog>
   );
--- a/specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
@@ -115,9 +115,7 @@ export function Status({
             {commonText.cancel()}
           </Button.Danger>
         ) : (
-          <Button.Danger onClick={handleClose}>
-            {commonText.close()}
-          </Button.Danger>
+          <Button.Info onClick={handleClose}>{commonText.close()}</Button.Info>
         )
       }
       className={{ container: dialogClassNames.narrowContainer }}
--- a/specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
@@ -71,9 +71,9 @@ export function Status({
     <Dialog
       buttons={
         state.status === 'SUCCESS' ? (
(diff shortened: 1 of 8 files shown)


CarolineDenis (approved):

carlosmbe (changes requested):
It works better than before. No longer negative. However, time remaining only escalates as time goes on. Here is a small snippet 

https://github.com/specify/specify7/assets/53784701/c541277d-5ab4-4ddd-9c5c-0a1c388ac849

specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
```diff
@@ -99,10 +100,7 @@ export function Status({
           {state.status === 'MERGING' && (
             <>
               <Progress max={state.total} value={state.current} />
-              <RemainingLoadingTime
-                current={state.current}
-                total={state.total}
-              />
+              <p>{percentage}%</p>
```
maxxxxxdlp, line 103: hmm, I don't like this change.
the video Carlos sent looked like a bug - if so, it needs to be fixed rather than the feature removed

CarolineDenis: Changing the remaining time by a percentage is what as been decided during the Sp meeting with the all team

carlosmbe (changes requested):
Nicely done.

Just a few aesthetic related things. Wouldn't call them road blocks. I lean towards polish. 

Firstly, change the highest number from 100% to 99% As right now it pauses for a moment when at 100. I believe having it stop at 99% would be ideal as the user can assume that it hit 100% when the merge success message shows up.

Secondly, can we change the red exclamation mark on top to something else. The tester deep inside me senses danger when they see a big red !!!!!

Also, at `Merge Successful`, we should make the close button Green not red.

Thanks for the work. 

https://github.com/specify/specify7/assets/53784701/d274ae38-30fb-4529-92cb-c0ee5767fa10

CarolineDenis:
> Nicely done.
> 
> Just a few aesthetic related things. Wouldn't call them road blocks. I lean towards polish.
> 
> Firstly, change the highest number from 100% to 99% As right now it pauses for a moment when at 100. I believe having it stop at 99% would be ideal as the user can assume that it hit 100% when the merge success message shows up.
> 
> Secondly, can we change the red exclamation mark on top to something else. The tester deep inside me senses danger when they see a big red !!!!!
> 
> Also, at `Merge Successful`, we should make the close button Green not red.
> 
> Thanks for the work.
> 
>  Screen.Recording.2023-08-18.at.11.01.38.AM.mp4

For: > Secondly, can we change the red exclamation mark on top to something else. The tester deep inside me senses danger when they see a big red !!!!!
@carlosmbe, any suggestion?

specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
melton-jason: The thought process behind the change is that the time remaining is often sporadic by nature when it comes to merging. 
 The current and total in the requests have a tendency to "jump around" a lot. More specifically, whenever the merging process needs to merge records recursively, the 'total' will need to be updated to reflect the number of added columns, which may cause the `time remaining` function to fluctuate. 

The thought process was that using a percentage "handles" the visual turbulence better from a user standpoint, since it is not _heavily_ changing the progress UI based on a drastic change in the current and total. 
In other words, a percentage is hopefully smoother. 
(Not my own personal thoughts here, just what I picked up as the reasoning behind the change)

> the video Carlos sent looked like a bug - if so, it needs to be fixed rather than the feature removed

While it may be tricky to have the time remaining function handle results like how it might get them form agent merging, I agree this is a bug and should probably be looked into/fixed

specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
```diff
@@ -94,6 +95,15 @@ export function Status({
       dimensionsKey="merging-progress"
       header={statusLocalization[state.status]}
       onClose={undefined}
+      icon={
+        state.status === 'PENDING' ? (
+          icons.cog
+        ) : state.status === 'MERGING' ? (
+          <span className="text-red-500">{icons.exclamationCircle}</span>
+        ) : (
+          <span className="text-green-500">{icons.checkCircle}</span>
```
melton-jason, line 104: Not the biggest fan of hardcoding the colors here. There are considerations with light/dark mode, and accessibility concerns to those that are color blind. 
If the colors are important, can you use a `dialogIcon` then, to stay consistent with the rest of Specify?
https://github.com/specify/specify7/blob/2a259ca22b439ba71b9313705cb2a253ac8edd6e/specifyweb/frontend/js_src/lib/components/Atoms/Icons.tsx#L111-L124

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
maxxxxxdlp: makes sense.

this is why the progress bars in any software with good UX are mostly fake.
they just gradually move across, and get slower the closer they reach to the end, and get faster the more actual progress is completed - until finally when everything is actually complete, instead of disappearing right away (which would break the illusion), the scroll bar reaches the end in a span of 1-2 seconds - giving immense satisfaction to the user

macOS update progress bar might be the best example of this
though, windows have also gotten much better with their progress bars since windows xp days

specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
maxxxxxdlp: @melton-jason great suggestion!

the only thing I can say in defense is that the `-500` colors are the only colors that work fine both in light and dark mode (though fine does not mean ideal)

melton-jason (approved):

carlosmbe (approved):

specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
```diff
@@ -40,13 +43,18 @@ export function Status({
...
         // eslint-disable-next-line @typescript-eslint/naming-convention
         headers: { Accept: 'application/json' },
       })
         .then(
           ({
-            data: { taskstatus: taskStatus, taskprogress: taskProgress },
+            data: {
+              taskstatus: taskStatus,
+              taskprogress: taskProgress,
+              response: response,
```
maxxxxxdlp, line 56:
```suggestion
              response,
```

specifyweb/specify/views.py
```diff
@@ -965,7 +966,8 @@ def merging_status(request, merge_id: int) -> http.HttpResponse:
     status = {
         'taskstatus': task_status,
         'taskprogress': task_progress,
-        'taskid': merge_id
+        'taskid': merge_id,
+        'response': merge.response
```
maxxxxxdlp, line 970: (optional) does `response` always contain either error message or null? in such case, consider renaming it

maxxxxxdlp (approved):

specifyweb/specify/views.py
CarolineDenis: it's either a string or an empty string

bronwyncombs (approved):
Functioning well thank you

https://github.com/specify/specify7/assets/135047322/c6e4d570-868a-4150-a941-5dc7ed98ba85

specifyweb/frontend/js_src/lib/components/Merging/Status.tsx
```diff
@@ -67,10 +78,29 @@ export function Status({
...
     <Dialog
       buttons={
-        state.status === 'MERGING' ? (
+        state.status === 'FAILED' ? (
+          <>
+            <Button.Info
+              onClick={(): void =>
+                void downloadFile(
+                  `Merging ${mergingId} Crash Report - ${new Date().toJSON()}.txt`,
+                  responseError
```
maxxxxxdlp, line 92: (optional) replace `responseError` with `produceStackTrace(responseError)` so that the file you would receive from the user contains much more useful information (including, which records the user was trying to merge)

maxxxxxdlp (approved):

specifyweb/specify/views.py
maxxxxxdlp: well, and when it's a string, it is an error string, right?
so wouldn't `errorMessage` be more appropriate?

also, try to not use empty strings in API responses - it should be explicitly `null` or the error message.

melton-jason (approved):

PR #3914, opened 2023-08-14
Add a max width for the side menu in security

### LINKED ISSUE #3905: Improve Security and Accounts on narrow screens
### Current:
image

### Proposed:
image

Something as simple as forcing it to take up only 33% of the screen until it stacks would be excellent.

CarolineDenis:
Fixes #3905


CarolineDenis:
https://github.com/specify/specify7/assets/108160931/882cbdc4-c99c-4c54-8fbf-b49b6a00fe77

melton-jason (commented):
Not the greatest fan of these breakpoints. I feel like they promote the use of [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)). 
(It doesn't help that [breakpoints themselves](https://tailwindcss.com/docs/responsive-design) are magic numbers of common pixel widths). 

Regardless, I can not deny that this does look better on narrow screens when the screen is "crowded". 
However, when the screen is not "crowded", there seems to be a significant amount of jittering near the breakpoint. 
Below are videos for how this is currently in `v7.9-dev` as well as the behavior for this current branch. 
@specify/ux-testing Let me know what your thoughts are on this. 

`v7.9-dev`:

https://github.com/specify/specify7/assets/64045831/3477bb4c-19a8-4854-abd1-d107034055d3

https://github.com/specify/specify7/assets/64045831/f00463cc-8d17-484e-8b00-248bd4f7239f

This branch (`issue-3905`):

https://github.com/specify/specify7/assets/64045831/5fb3eeb7-0294-4d34-82eb-8847760cfb89


https://github.com/specify/specify7/assets/64045831/2255171a-cbca-4408-a880-d34e163176fb

carlosmbe:
Based on the videos Jason provided. I prefer how this branch handles the Tight situations but prefer how the normal situation was handled previously. Is there a way to have both?

grantfitzsimmons (approved):
Looks good to me. I prefer this change in both "normal" and "tight" situations.
The magic number discussion is interesting– but I think the solution works fine. I'll leave it to you @melton-jason to suggest revisions if you see fit.

bronwyncombs (approved):
👍 

https://github.com/specify/specify7/assets/135047322/fcc20e2b-4252-4d3a-a11b-41b6af9954d4

PR #3916, opened 2023-08-15
Allow to resize text area on y axis only

### LINKED ISSUE #3873: Merge dialog: Stretching a large `textarea` field into the other records section leaves it stuck
https://github.com/specify/specify7/assets/37256050/76af545c-7833-4590-998e-575be47c2fcc

Expanding a text box outside of its specific column leaves it stuck under other text boxes. You cannot click or shrink it back to its original size.

It should be restricted to the width of its column.

CarolineDenis:
Fixes #3873


maxxxxxdlp (changes requested):
you added a `className` prop, but it's only supported by the text area - that is unexpected behavior to anyone not too familiar with the code.
thus, either make it be supported by all fields, or rename the prop to make it clear that it's for text area only (`textAreaClassName`).

better yet, props that are specific to just a single field type (i.e textarea), should be passed in `fieldDefinition`. everything outside field definition should be supported by all fields

Anoter solution, is to just add this class name to the div that is around `<FormField>` in CompareField.tsx: `[&_textarea]:resize-y`. @melton-jason thoughts on which is better?

The last solution is the simplest, but could have unintended side effects (if `<textarea />` would be rendered by anything else in the future, it would be affected too - probably not a bit deal as is easy to debug)

melton-jason:
> better yet, props that are specific to just a single field type (i.e textarea), should be passed in fieldDefinition. everything outside field definition should be supported by all fields
> 
> Anoter solution, is to just add this class name to the div that is around <FormField> in CompareField.tsx: [&_textarea]:resize-y. @melton-jason thoughts on which is better?

I agree that the first solution is the most modular and allows the most flexibility for future developers and components. 
If we wanted to future-proof this change against potential bugs and save someone the hassle of debugging (even if it is simple to debug) this is probably the solution we should do. 

But given that we have no immediate foreseeable other use cases for `<textarea>`, I would not mind the simple solution for now, as long as we make it clear to future developers/us about this problem (by opening an Issue, leaving a comment, etc.) so it's that much easier to resolve if it _does_ become a problem or someone wants to implement the first solution

bronwyncombs (approved):
Looks good! 


https://github.com/specify/specify7/assets/135047322/899d521d-099d-44aa-8a63-7e7462a0b1e2

PR #3917, opened 2023-08-15
Restore slider to fictitious record set when adding new records

### LINKED ISSUE #3706: Update the index in the slider when adding a new record in loan interaction form
When adding a new prep in the loan interaction form the record selector displays the last opened resource instead of the last added resource. 

1. Go to 'interaction'
2. Click on 'loan'
3. Choose a record set with prep and apply
4. Scroll to loan preparation subform 
5. Change the type in the meta to subform if not already done
6. Click on the + button to add a new prep 
7. See 1/2 in the slider instead of 2/2 

**Expected behavior**
The slider should be 2/2 and we should be on the last added resource

**Additional info from Maxx**
Some possible ways to do that

Add a callback i.e onAdded={} that would be called by InteractionDialog when it adds new records. the callback could be called right before new records are added (thus you could use collection.models.length as new index) or right after (in which case InteractionDialog will pass the new records, and you could do collection.models.length - addedRecords.length to get new index)
OR record the number of records in the collection before the interaction dialog is opened in a ref, and once the interaction dialog is closed, check if number of records changes and get the new index
OR, set an event listener for new items being added, and change the index when that happens - though this one would be tricky as if 4 records are added all at once, the event index would be called 4 times, thus, subview will end up displaying the last record

### LINKED ISSUE #3903: Fictitious record set header appears too late
**Describe the bug**
When you use the Add or Clone button to add a record, the form doesn't appear to have stacked records until the second record is saved. 

## v7.9:

https://github.com/specify/specify7/assets/135047322/be8cb12b-431d-4fcb-bb97-dfb0207322fa

## v7.8.13:

https://github.com/specify/specify7/assets/135047322/296257b6-a411-413f-86e0-68efb5e82e11

 - macOS, Chrome, `aafc_aac_3_8_23`

melton-jason:
Fixes #3903 
Fixes #3706 

- - - 

Alongside testing the issue, please do general testing of record sets! 
This includes things like:
- Fictitious Recordsets in DataEntry
- Fictitious Recordsets via Browse in Forms from the Query Builder
- Actual Recordsets

Particularly test cases involving adding/removing records.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
@@ -38,7 +38,7 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
   isDependent,
   mode,
   canRemove = true,
-  totalCount = ids.length,
+  totalCount = ids.length + (typeof newResource === 'object' ? 1 : 0),
   isLoading: isExternalLoading = false,
   isInRecordSet = false,
   onClose: handleClose,
@@ -98,7 +98,7 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
   );
   const index =
     typeof newResource === 'object'
-      ? totalCount
+      ? totalCount - 1
       : Math.min(rawIndex, totalCount - 1);
 
   const currentResource = newResource ?? records[index];


maxxxxxdlp (approved):

CarolineDenis (approved):

CarolineDenis:
We can test if it fixes this one too: https://github.com/specify/specify7/issues/3706
It seems like it does

bronwyncombs (approved):
Working as expected! Thanks! 

Fictitious Record Set via Data Entry:

https://github.com/specify/specify7/assets/135047322/c728e7a6-aaac-47bf-b910-6a15c99b2035

True Record Set:

https://github.com/specify/specify7/assets/135047322/918a2bdc-1656-450d-9a4f-8a509182f8a5

Preparation slider issue also corrected:

https://github.com/specify/specify7/assets/135047322/c9511f70-aeba-4227-8a2e-d4a8fcccc518

Browse in Forms via Query Builder also tested and working as expected.

Issue #2598, opened 2022-12-13
Abort fetch requests when component is removed


maxxxxxdlp:
When a component is removed before a network request is resolved, or an old request is replaced with a new one, the old request should be canceled. This would reduce wasted network bandwidth and reduce load on the back-end server.

Here is how to tell fetch to abort the request - https://developer.mozilla.org/en-US/docs/Web/API/AbortController/signal
Potentially, that could be integrated into the `useAsyncState` hook since that's where most network requests are made from on the front-end

PR #3926, opened 2023-08-18
Remove taxa stats, and kill requests for add stat dialog

realVinayak:
Removes the taxa stats - so any future stats page won't have "Taxa represented" category.

Also kills the not needed network requests when user closes add stats dialog, but some requests were queued.

TEST 1:
1. The current stat pages (created before this) should work fine - they should just have no items in the category

TEST 2
1. Open networks tab for testing second change. 
2. Use a big database (herb will do). Remove all stats (by deleting categories).
3. Click on 'Add' in Edit Mode. 
4. You should see network requests being made in networks tab. Close the add stats dialog. In v7.9-dev, some requests would still be made (it would say ephemeral). In this branch, they won't.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Statistics/AddStatDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Statistics/AddStatDialog.tsx
@@ -23,6 +23,7 @@ import type {
   StatFormatterSpec,
   StatLayout,
 } from './types';
+import { cleanThrottledPromises } from '../../utils/ajax/throttledPromise';
 
 export function AddStatDialog({
   defaultStatsAddLeft,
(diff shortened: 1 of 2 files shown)


melton-jason (approved):
Looks good, from a code standpoint. :+1:

On the initial render, handle loading the stats and then "kill" the remaining promises.

realVinayak:
> On the initial render, handle loading the stats and then "kill" the remaining promises.

Yep, that's what it is doing right now. It can't kill the request already sent, but the ones queued after that (since stats sends request in a throttled manner).  I added killing queued promises for stats page in general (if you navigate away from stats, it won't fetch anything), but thought this place would also benefit from it (imagine a tree with a lot of ranks) in the add stats dialog.

maxxxxxdlp:
Aborting not needed requests and making back-end respect the abort signal would be a useful performance feature (https://github.com/specify/specify7/issues/2598)

It would come in use more often than you might think. I.e imagine going between records in a record set - you might skip though them quickly, yet several requests for each query combo box are sent

maxxxxxdlp (approved):

PR #3927, opened 2023-08-18
Add new className to avoid shifting in browse in form

### LINKED ISSUE #3784: Query Browse in Forms dialog changes size as you navigate
**Describe the bug**

The dialog shrinks and grows as you click through. Makes it hard to go through at a decent pace since you have to move your mouse between forms. Not in `v7.8.13`.
 
**To Reproduce**
1. Query on Accession table
2. Click on `Browse in Forms`
3. Browse pages

**Expected behavior**

Dialog should stay uniform size as you browse

**Screenshots**

https://github.com/specify/specify7/assets/135047322/a6242a1e-90cb-4e57-8c75-8e71e630f931

**Desktop:**
- MacOS, Chrome, v7.9-dev

**Database Name:** 
`fwri_6_1_23`

CarolineDenis:
Fixes #3784 

+ add new className to slider to decrease shifting when passing from number with 1 digit to number with 2 digits 

Notes: used classNameProp to apply the new className only for certain dialogs


specifyweb/frontend/js_src/lib/components/FormSliders/Slider.tsx
```diff
@@ -45,7 +45,7 @@ export function Slider({
       >
         {icons.chevronLeft}
       </Button.Small>
-      <div className="grid grid-cols-[1fr_auto_1fr] items-center gap-1 font-bold">
+      <div className="grid min-w-[theme(spacing.20)] grid-cols-[1fr_auto_1fr] items-center gap-1 font-bold">
```
maxxxxxdlp, line 48: hmm, the slider is not supposed to have its width change

it has `grid-cols-[1fr_auto_1fr]` - meaning the number of the left and number on the right must be same width.

does that not work?

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
```diff
@@ -239,6 +239,7 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
             : undefined
         }
         onSaved={(): void => handleSaved(resource!)}
+        classNameProp="w-[inherit]"
```
maxxxxxdlp, line 242: rename to `className` here to match how such prop is called in other places
internally in ResourceView, you can rename it to avoid conflict:
```
className: classNameProp
```

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/FormSliders/Slider.tsx
maxxxxxdlp: if you look at the video in the original issue (#3784) you will see that the width change is likely caused by the textarea value being longer or shorter - if that's indeed the case, there are styles that can be applied that would forbid textarea to grow it's parent

BUT, fixing it for just text areas won't help - this bug can happen for many different reasons, and the reason depends on the form layout and whether record has subforms - a more generic solution is required - that is why @CarolineDenis you worked on the feature that was going to forbid dialogs from shrinking after expanding (to help minimize the layout shifts).

another solution you did was making dialogs remember their size when resized - if you manually resize the dialog, the bug in the original issue (#3784) likely won't happen

another generic solution is introducing keyboard shortcuts (see https://github.com/specify/specify7/issues/1746), which would reduce the need for reaching for the record slider at the top

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
maxxxxxdlp: is this class name added here to prevent the dialog from remembering size?
not sure how that fixes the issue, and also not sure why it was added just in this case, rather than in all the cases where ResourceView displays as a dialog

specifyweb/frontend/js_src/lib/components/FormSliders/Slider.tsx
CarolineDenis: > hmm, the slider is not supposed to have its width change
> 
> it has `grid-cols-[1fr_auto_1fr]` - meaning the number of the left and number on the right must be same width.
> 
> does that not work?

Without min-w-[theme(spacing.20)]: 

https://github.com/specify/specify7/assets/108160931/23dff548-7af4-4bbf-8cdc-059d8c0f8de9

With it: 

https://github.com/specify/specify7/assets/108160931/e8ca9468-2da3-412a-adf9-a11a80170723

There is 2 shiftings, one in the slider and one in the full container.

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
CarolineDenis: <<rather than in all the cases where ResourceView displays as a dialog>> 
Because I didn't want it to be apply to the ResourceView 'Create Record Set' by example. Here I can apply it only for the ones I need the w-inherit

CarolineDenis:
With w-inherit: 

https://github.com/specify/specify7/assets/108160931/ab924c36-76a5-4bc6-b3a3-46342194c002


https://github.com/specify/specify7/assets/108160931/dd3fc285-7f16-48fe-b295-60e5f5623f14

Without: 

https://github.com/specify/specify7/assets/108160931/6a42d3c7-8963-4d85-ad4b-0dae00b5df4c

CarolineDenis:
Without min-w-[theme(spacing.20)]: 

https://github.com/specify/specify7/assets/108160931/23dff548-7af4-4bbf-8cdc-059d8c0f8de9

With it: 

https://github.com/specify/specify7/assets/108160931/e8ca9468-2da3-412a-adf9-a11a80170723

There is 2 shiftings, one in the slider and one in the full container.

specifyweb/frontend/js_src/lib/components/FormSliders/Slider.tsx
maxxxxxdlp: I am sorry to say, but this is not a solution, but only an ugly hack
`min-w-[theme(spacing.20)]`

Turns out it was just a bug - the code was in place for making the input always be equal width to the total count, thus preventing the layout shift - see my commit with the fix

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
```diff
@@ -314,7 +316,7 @@ export function ResourceView<SCHEMA extends AnySchema>({
         )
       }
       className={{
-        container: `${dialogClassNames.normalContainer} ${
+        container: `${dialogClassNames.normalContainer} ${classNameProp} ${
```
maxxxxxdlp, line 319: `classNameProp` could be undefined - and then it will literally add a `"undefined"` class name - eslnit can prevent this bug

maxxxxxdlp: in the video you sent of Accession dialog, is the dialog manually resized?

With `w-[inherit]`, the dialog is full width for me always (because the parent div that the dialog would inherit the width from has full width) - the effect is the same as if I had `w-full` - thus, can you try out `w-full` instead of `w-[inherit]`

moreover, consider the variable `isFullHeight` in this file - consider renaming it to `isFullSize` and when true, adding both `h-full` and `w-full` to the dialog class names - that would remove the need for a separate classNameProp, and will fix this bug in more places

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
CarolineDenis: 'in the video you sent of Accession dialog, is the dialog manually resized?' 
No it's not

specifyweb/frontend/js_src/lib/components/FormSliders/Slider.tsx
```diff
@@ -83,7 +83,7 @@ export function Slider({
           />
         </label>
         <span>/</span>
-        <span>{count}</span>
+        <span className="text-center">{count}</span>
```
maxxxxxdlp, line 86: (optional) can you double-check if this is still needed? compare if it looks better with it or without when the total count is < 9 and when total count is >99

maxxxxxdlp (approved):

bronwyncombs (approved):
Working as expected! 🎊 

https://github.com/specify/specify7/assets/135047322/a70017fc-469f-4a6c-a4a8-0df9c65ce7a6

PR #3935, opened 2023-08-21
Display checkbox 'show conflicting field' only when necessary

### LINKED ISSUE #3879: Unnecessary show conflicting fields checkbox in merging subview dialogs
**Describe the bug**
A "Show conflicting fields only" checkbox appears on the record-editing sub-dialogs from large buttons in Merge Records dialog. It is for the main Merge Records dialog and shouldn't be on these also.

**Screenshots**

https://github.com/specify/specify7/assets/135047322/8e556318-e27c-4245-958e-0f905ac12da8

**Coming from a button for a field that was not conflicting, it exits the dialog:**

https://github.com/specify/specify7/assets/135047322/a4a14a80-e074-4e54-94d5-1ff562a24de8

**Desktop:**
 - macOS, Chrome, `v7.9-dev`

**Database Name:** `archireef`

CarolineDenis:
Fixes #3879


carlosmbe:
Should I UX Test this or is it a draft?

CarolineDenis:
> Should I UX Test this or is it a draft?

we can wait first for Jason review. I will tag UX-UI once ready

specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
```diff
@@ -147,6 +147,10 @@ function MergeDialog({
     ...[mergedRecords, ...children].map((children) => children.length)
   );
 
+  const childrenLength = resources
+    .map((record) => f.maybe(record, getChildren))
+    .filter((children) => children && children.length > 0).length;
```
melton-jason, line 152: Instead of doing the implicit conversion to a boolean here, can we explicitly check for undefined? 
Something like `typeof children !== undefined && children.length > 0`

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
maxxxxxdlp: also, you don't care about the number of children - you care if there is more than one column that has at least 1 child - i.e if the first column has two children, the checkbox should still not be displayed, it's only if there are children in more than column that there is something to compare to, and thus the "show conflicting fields only" checkbox makes sense.

specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
```diff
@@ -147,6 +147,10 @@ function MergeDialog({
     ...[mergedRecords, ...children].map((children) => children.length)
   );
 
+  const childrenLength = resources
+    .map((record) => f.maybe(record, getChildren))
```
maxxxxxdlp, line 151: why are you getting children here again if the code already did that just a few lines up?
```
  const [children, setChildren] = useChildren(
    mergedRecords,
    getChildren,
    resources,
    relationship
  );
```

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
CarolineDenis: children is not undefined when 0 records. Agree I need to change to > 1 not > 0

specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
CarolineDenis: not the same

specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
```diff
@@ -147,6 +147,10 @@ function MergeDialog({
     ...[mergedRecords, ...children].map((children) => children.length)
   );
 
+  const childrenLength = children.filter(
+    (child) => child && child.length > 1
+  ).length;
```
maxxxxxdlp, line 152: shouldn't this be the following instead?

```tsx
  const childrenLength = children.filter(
    (child) => child && child.length > 0
  ).length;
```
and then:
```tsx
childrenLength > 1 &
```
(notice 1 instead of 0)

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
CarolineDenis: yes it should. my bad

melton-jason (approved):

maxxxxxdlp (approved):

bronwyncombs (dismissed):
What is the purpose of the checkbox on the record merging sub-dialogs? Can we remove it from those altogether? It only appears on the dialogs that have a record as this PR describes, but even then, I am confused by the what the necessity of the checkbox is since selecting it only affects the primary merging dialog in the background 

https://github.com/specify/specify7/assets/135047322/e34f613c-be38-4ef1-9522-f7f52fcc68b3

I can see the benefit in having this tool if it applied to the data shown on the current dialog, but it does not at this time.

maxxxxxdlp:
The checkbox is useful and makes a difference if you have more than 1 resource in the row - in your video, one subview dialog had no records, and another had only one record in a row.
You can't show "conflicting fields" with just 1 or 0 record because then there are no conflicts - you need at least 2 records

specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
```diff
@@ -177,8 +173,6 @@ function MergeDialog({
     <MergeDialogContainer
       buttons={
         <>
-          {childrenLength > 1 && <ToggleMergeView />}
```
maxxxxxdlp: You can make the check for when the button appears smarter, but certainly don't remove the check

https://github.com/specify/specify7/pull/3935#issuecomment-1743948761

Didn't you already fix something similar recently, was it for the main dialog, not the sub-view?

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Merging/CompareSubView.tsx
CarolineDenis: It was this one, and we remembered that we kept the check box at some point but nobody could remember the reason ^^

CarolineDenis:
> The checkbox is useful and makes a difference if you have more than 1 resource in the row - in your video, one subview dialog had no records, and another had only one record in a row.
> You can't show "conflicting fields" with just 1 or 0 record because then there are no conflicts - you need at least 2 records

I couldn't find any cases where several records were on the same row in a sub view merging dialog. Do you have an example?

maxxxxxdlp:
Thanks for reminding me about this behavior.
When it comes to merging sub-views, I made the algorithm very conservative - assume that every child record is not a duplicate unless it's absolutely identical to the child record in the other resource.
That is why most times in practice there would only be one record per row.
I realized that it would be nearly impossible to come up with good heuristics for detecting duplicate child records in a way that is generic to any table without involving some sort of machine learning, so left it to the user to decide how they want to merge records - that is why you have "up" and "down" buttons next to each resource in the subview dialog, so that you can manually say that some records are duplicate.

Thus, I would say the checkbox could still be useful, but given that you have the same checkbox on the main dialog, also having it on the child dialog is not required.

maxxxxxdlp (approved):

pashiav (approved):
"Show conflicting fields only" checkbox no longer appears on the record-editing sub-dialogs. Looks good!

https://github.com/specify/specify7/assets/118556884/b4ac0c54-fc5b-4edf-87a6-c0a36dc6acb5

carlosmbe (approved):
It looks good. Would like @bronwyncombs to take a look as well but it seems fine to me.

specifysoftware:

PR #3936, opened 2023-08-21
Add warning dialog when no prep available

### LINKED ISSUE #2881: Improve the "Create New Loan" dialog

CarolineDenis:
Fixes #2881


specifyweb/frontend/js_src/lib/localization/interactions.ts
```diff
@@ -90,6 +90,9 @@ export const interactionsText = createDictionary({
     'uk-ua': 'Без препаратів',
     'de-ch': 'Ohne Präparate',
   },
+  continueWithoutPreparations: {
+    'en-us': 'Continue without preparations',
```
maxxxxxdlp, line 94: this is too long for a button label, but not sure if there is a briefer way that is just as clear

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/localization/interactions.ts
CarolineDenis: Agreed, but we decided as a team to go for better communication with 'Continue without preparations' in this case.

bronwyncombs (changes requested):
@grantfitzsimmons and I believe that the behavior should be one of two things: 

1. Give a warning in the dialog like so: 
Screenshot 2023-09-27 at 2 32 31 PM

In which case the ignored preparations would not appear in the next dialog. 

`OR`

2. Leave out the warning component and opt to just display all preparations in the next dialog since it does indicate they are unavailable there.
---
Here is the current behavior on this branch: 

There is no warning dialog for catalog numbers without preparations. Instead, it will just remove the line for the object without preps in the next dialog. 

In the recording below, you can see that Cat # 000042852 was entered, but no warning was given before removing it from the Preparations dialog. 

https://github.com/specify/specify7/assets/135047322/937421dc-e67f-4697-8d8d-c73402f0a491

If no objects have preps available, it will prompt you with this dialog, which is good! 

https://github.com/specify/specify7/assets/135047322/6a74c184-bdae-4c54-b0f7-f5d8f2b9d5df

If an object has more than one preparation, and one of the preps has none available, the unavailable prep will still display on the next dialog without warning. 

(This is essentially what option 2 describes) 

https://github.com/specify/specify7/assets/135047322/840e0404-2472-45fb-a084-17b4eee40d12

CarolineDenis:
@bronwyncombs @carlosmbe please test this in https://github.com/specify/specify7/pull/4030#pullrequestreview-1647535932

CarolineDenis:
Fixed in https://github.com/specify/specify7/pull/4030

PR #3942, opened 2023-08-24
Allow having tree stats use direct rank names (and climb tree)

realVinayak:
Allows defining a query like 'geography.Country'. Looks for name in the {TREE}treedefitem table. 

The previous code allowed just to modify the start values in the last query field, but these changes allow them to be generic. 

TO TEST:
1. Go to stats. All the previous versions of stats should still work - the query would be different in the query builder. New ranks will be visible in the add stats dialog (all ranks above Earth / Life are shown). 
2. The following query should be present (if rank is continent): 
image

Change 2:
Removes distinct from locality stat


realVinayak:
The query that fetches ranks is done only one time. Since tree ranks are fetched outside of stats already, those could also be used instead of this. 

But stats like type status utilizes this too (it fetches distinct type status) so I thought it would be simple enough to just have the taxa ones utilize this in the stats spec.

Alternatively, won't be hard to use the tree tank promise outside of stats

specifyweb/frontend/js_src/lib/components/Statistics/StatsSpec.tsx
```diff
@@ -606,7 +618,7 @@ const statSpecToItems = (
 
 function generateBackEndSpec(statsSpec: StatsSpec): RA<{
   readonly responseKey: string;
-  readonly querySpec: QuerySpec | undefined;
+  readonly querySpec: ((dynamicResult: string) => QuerySpec) | undefined;
```
maxxxxxdlp, line 621: careful, there may be an infinite re-render loop or at least a performance consideration from this - if before it was relying on the object being, but now the object would be different each time

maybe not an issue - you know your code best, so double check

maxxxxxdlp (approved):
I might have missed the context behind some of the things as I am not familiar with the code much (and you created a beast with the stats page), but looking good, and I like some of the refactoring you did

specifyweb/frontend/js_src/lib/components/Statistics/StatsSpec.tsx
realVinayak: Not infinite re-render loops - the QuerySpec aren sent down to children components (and not to the parents). Also there are no useEffects that use this as a dependency. The performance problem is a real problem though. 

Fixed in  https://github.com/specify/specify7/pull/3942/commits/8d9c44df904202e253a97abbffe77d73fffa219b. The logic there is that queries which we generate for the user (like dynamic ones) can be cached - it already caches the other default ones (defined in stats spec) and the custom stats (coming from layout directly) (until user modifies them). When users do modify a dynamic stat (which they can), we don't care about getting a stale value from this ref (which we will), since it is automatically converted to custom stat with latest query specs.

realVinayak: > beast with the stats page

It took me a while to re-understand how it works :)

maxxxxxdlp (approved):

melton-jason (approved):
It would take me too long to fully understand how Statistics works to fully consider these changes, but from what I am gathering from this little section, the changes seem fine.

bronwyncombs (approved):
Geog rank stats appeared in the choose stats dialog to be added to a new category. The queries for each rank appear correct. Upon clicking a stat on `herb_rbge_9_3_23`, the query results were returned promptly, but count was slightly delayed. 
**Geo represented:**
Screenshot 2023-10-25 at 10 11 32 AM
**Localities without Distinct:**
Screenshot 2023-10-25 at 10 12 05 AM

Also tested on `fwri`

grantfitzsimmons:
@realVinayak Seems there are some merge conflicts

realVinayak:
Also need to test performance on AWS

realVinayak:
Can this be tested another time by a ux-reviewer before merge? @specify/ux-testing

pashiav (approved):
Looks good!
v7 9-dev-count-indirect-continent

specifysoftware:

Issue #2602, opened 2022-12-13
Replace custom listboxes with a library


maxxxxxdlp:
Sp7 WorkBench Mapper and Query Builder use in-house list boxes.
They work quite well, and have good enough accessibility, but rewriting the code to use a third party solution like https://headlessui.com/react/listbox would improve accessibility even further, while adding more features for free and reducing tech dept

maxxxxxdlp:
I looked into this futher.
HeadlessUI's list box provides a way to control whether list box is visible or not, or even a way to leave it always visible (useful for things like mapping viewer and for controlling keyboard navigation between list boxes)

However, when decide to control whether listbox is visible, HeadlessUI's stops handling all keyboard events - quite sad. It provides features like search, page up/page down, home/end on top of the arrow keys. Since none of those are available when listbox is controlled, the value of transitioning to a different library is lower

PR #3945, opened 2023-08-24
Allow query line drop down to flow outside its container

### LINKED ISSUE #3941: Query field drop-down mapper is confined to scroll in the mapped fields height
**Describe the bug**
Using the drop-down field mapper to change an already-mapped field is now confined to scrolling in just the height of the line of the mapped field.

**Screenshots**

https://github.com/specify/specify7/assets/135047322/3cdfc78e-e5fe-4642-a31b-60f7d7b617c7

**Desktop:**
 - macOS, Chrome, v7.9-dev, `herb_rbge_7_14_23`

**Additional context**
Came from PR #3877

CarolineDenis:
Fixes #3941

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
@@ -286,7 +286,7 @@ export function QueryLine({
         {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
         <div
           className={`
-          flex flex-1 flex-wrap gap-2 print:gap-1
+          flex flex-1 flex-wrap gap-2 overflow-auto print:gap-1 sm:overflow-visible
           ${field.filters.length > 1 ? 'items-baseline' : 'items-center'}
           ${isFocused ? 'rounded bg-gray-300 dark:bg-neutral-700' : ''}
           ${isBasic ? 'contents' : ''}
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Line.tsx
@@ -286,7 +286,7 @@ export function QueryLine({
         {/* eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions */}
         <div
           className={`
-          flex flex-1 flex-wrap gap-2 overflow-auto print:gap-1
+          flex flex-1 flex-wrap gap-2 print:gap-1
           ${field.filters.length > 1 ? 'items-baseline' : 'items-center'}
           ${isFocused ? 'rounded bg-gray-300 dark:bg-neutral-700' : ''}
           ${isBasic ? 'contents' : ''}


CarolineDenis:
please test in various size of windows

carlosmbe:
Looks promising. Going to request @bronwyncombs to take a look to as they opened the issue.
Also, is it a technical limitation that we can not change the mapping in Basic View?

maxxxxxdlp (commented):
looks good, but can you look back at the commit/pull request that introduced that `overflow-auto` to see why it was added? it might have been there for good reason

i.e it might have been there for narrow screens

melton-jason:
> Blank space between mapper and results is fixed!
> 
> Part of the problem: Firefox allows the window to get smaller than our longer field names (i.e. Collecting Event Attachments, Reference Work Attachments) which causes the staggered button appearance. I'm not sure how we would want work around that.
> 
>  Screen.Recording.2023-08-15.at.9.51.50.AM.mov


More specifically, it was added to address the problem on https://github.com/specify/specify7/pull/3877#pullrequestreview-1578734136.

The cause of #3941 happens with `overflow-scroll` and `overflow-auto`. 
Maybe a more constrained overflow may work? 
Would `overflow-y-visible` work?

CarolineDenis:
> looks good, but can you look back at the commit/pull request that introduced that `overflow-auto` to see why it was added? it might have been there for good reason
> 
> i.e it might have been there for narrow screens

Yes I added it in the past but couldn't reproduce  a case when it was needed. Hence the ask to test in various screen sizes. (Added at first for the shift of the lines in small narrow screen)

CarolineDenis:
> > Blank space between mapper and results is fixed!
> > Part of the problem: Firefox allows the window to get smaller than our longer field names (i.e. Collecting Event Attachments, Reference Work Attachments) which causes the staggered button appearance. I'm not sure how we would want work around that.
> > Screen.Recording.2023-08-15.at.9.51.50.AM.mov
> 
> More specifically, it was added to address the problem on [#3877 (review)](https://github.com/specify/specify7/pull/3877#pullrequestreview-1578734136).
> 
> The cause of #3941 happens with `overflow-scroll` and `overflow-auto`. Maybe a more constrained overflow may work?
> 
> Would `overflow-y-visible` work?
> 
> https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y It should allow the drop down to display outside of the container

Overflow-y-visible allows the drop down to display outside the container but brings back the shift of the line: 
Screenshot 2023-08-25 at 7 03 54 AM

CarolineDenis:
with overflow-auto: 
- no shifting of the lines in narrow screen 
- drop down inside the div instead of outside 
without overflow-auto: 
- shifting of the lines in narrow screen 
- drop down outside the div like expected 

Not ideal but why not add overflow-auto only for small screens? 
That would result in: 
sm: no shifting - dropdown inside 
md and >: no shifting - dropdown outside 

@grantfitzsimmons @melton-jason

grantfitzsimmons (approved):

bronwyncombs (approved):
drop down scroll only on very small screens is working. tested firefox, chrome, safari

https://github.com/specify/specify7/assets/135047322/59643d9b-a422-412b-a2b7-1b70e4f44a02

melton-jason (approved):
I am not the biggest fan of this solution. 
Although I can not come up with a better solution without some better knowledge of css. 

This will probably have to do for now

maxxxxxdlp:
related issue that can resolve this bug once and for all - https://github.com/specify/specify7/issues/2602
while also reducing how much code we have to maintain

PR #3946, opened 2023-08-24
Replace separators in strings to avoid confusion with export separator

### LINKED ISSUE #3895: Records with commas break Selected Create CSV
**Describe the bug**
The Issue was first reported [here](https://discourse.specifysoftware.org/t/strange-behaviour-with-create-csv-from-selected-query-results/1249) on discourse.

The cause is exporting formatted Localities that have contain commas in their values. The "CSV" sees these are independent values and puts them each in a separate column. 

This should be reproducible anywhere where records have commas in their name.

**To Reproduce**
1. Go build a new Locality Query
2. Add formatted and anything else
3. Export a few records and compare it to the complete everything export

**Beautiful Pictures**
### Query Results
Screenshot 2023-08-03 at 9 41 01 AM

### Selected Export
Screenshot 2023-08-03 at 9 42 02 AM

### Full Export
Screenshot 2023-08-03 at 9 42 22 AM

**Desktop:**
 - OS: Mac OS
 - Browser: Chrome
 - Specify 7 Version: 7.9

**Database Name:** aafc-aac 

**Reported By**
AAFC

CarolineDenis:
Fixes #3895


specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -86,7 +86,15 @@ export function QueryExportButtons({
     );
 
     const joinedSelected = selectedResults?.map((subArray) =>
-      subArray?.slice(1).join(separator)
+      subArray
+        ?.map((item) => {
+          if (typeof item === 'string') {
+            return item?.replace(new RegExp(separator, 'g'), '-');
```
maxxxxxdlp, line 92: hmm, I think there is no need for regex.
simply use `.replaceAll(separator, '-') instead

maxxxxxdlp: if you already checked for `typeof item === 'string'`, why have a question mark here?

really `?.` should be sufficient - you shouldn't need the `if()` or the other return, or the `{` `}`

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
CarolineDenis: I added the 'if' for: "Property 'replaceAll' does not exist on type 'number'."

maxxxxxdlp: ah, I see now what this code is trying to do.
replacing occurrences of separators with a different character is not correct - that corrupts the data. Fortunately, a better solution exists. See:
https://stackoverflow.com/questions/4617935/is-there-a-way-to-include-commas-in-csv-columns-without-breaking-the-formatting

maxxxxxdlp: even better, you don't need to do this on your own, because it's bug prone (there are other edge cases to consider - unicode, encodings, new lines...)
that is why we use a library `csv-stringify` for that
the `downloadDataSet` function could probably be modified/refactored to suit your needs

maxxxxxdlp (changes requested):
see https://github.com/specify/specify7/pull/3946#discussion_r1306466107

maxxxxxdlp (approved):
looks good!
good job refactoring the code to match the needs

carlosmbe (changes requested):
Slay. It works great with formatted Agents Queries. However, it's not working well with formatted Locality Queries.

Screenshot 2023-09-25 at 3 32 05 PM
Screenshot 2023-09-25 at 3 32 17 PM

CarolineDenis:
> Slay. It works great with formatted Agents Queries. However, it's not working well with formatted Locality Queries.
> 
> Screenshot 2023-09-25 at 3 32 05 PM Screenshot 2023-09-25 at 3 32 17 PM
Screenshot 2023-09-26 at 8 44 40 AM
@carlosmbe, the data starting in column 2 are phantom fields that are being added to the export file by mistake. I know it's confusing. There's already an open issue for that. From your screenshot, we can see that the formatted column hasn't been split where there is a comma, which is the expected behavior here.

bronwyncombs (approved):
Formatted fields aren't broken up and appear in CSV as they do in query builder results.

Screenshot 2023-10-02 at 9 29 53 AM
Screenshot 2023-10-02 at 9 35 29 AM

maxxxxxdlp:
@CarolineDenis (sanity check) I might be missing context but why do some columns on this screenshot don't have column headings?
and why are some columns completely blank?

image

CarolineDenis:
> @CarolineDenis (sanity check) I might be missing context but why do some columns on this screenshot don't have column headings? and why are some columns completely blank?
> 
> image
> 
> If those are the phantom fields, they should be excluded from the CSV export

Yes you are right, it's the phantom. an other issue is open for that. Vinny is working on it

PR #3950, opened 2023-08-25
Fix refresh stopping other items from loading

realVinayak:
As part of performance improvements, the requests were stopped when user navigates away from stats page. The same logic was incorrectly used for refresh. So, clicking refresh multiple times, cleans up all the requests from the previous click - which caused the items still waiting in queue to never get loaded.

### DIFF
--- a/specifyweb/frontend/js_src/lib/utils/ajax/throttledPromise.ts
+++ b/specifyweb/frontend/js_src/lib/utils/ajax/throttledPromise.ts
@@ -10,8 +10,8 @@ const currentRequestsGenerator = <T>(): WritableArray<PromiseWithSpec<T>> => [];
 
 let maybeFulfilled: WritableArray<KillablePromise<any>> = [];
 
-export const cleanThrottledPromises = (killPromise = true): void => {
-  if (killPromise) maybeFulfilled.forEach((promise) => promise.killPromise());
+export const cleanThrottledPromises = (killPromises = true): void => {
+  if (killPromises) maybeFulfilled.forEach((promise) => promise.killPromise());
   maybeFulfilled = [];
   /*
    * Since the kill promise is already supplied, the promises will resolve within a work loop.
--- a/specifyweb/frontend/js_src/lib/utils/ajax/throttledPromise.ts
+++ b/specifyweb/frontend/js_src/lib/utils/ajax/throttledPromise.ts
@@ -10,8 +10,8 @@ const currentRequestsGenerator = <T>(): WritableArray<PromiseWithSpec<T>> => [];
 
 let maybeFulfilled: WritableArray<KillablePromise<any>> = [];
 
-export const cleanThrottledPromises = (): void => {
-  maybeFulfilled.forEach((promise) => promise.killPromise());
+export const cleanThrottledPromises = (killPromise = true): void => {
+  if (killPromise) maybeFulfilled.forEach((promise) => promise.killPromise());
   maybeFulfilled = [];
   /*
    * Since the kill promise is already supplied, the promises will resolve within a work loop.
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/utils/ajax/throttledPromise.ts
```diff
@@ -10,8 +10,8 @@ const currentRequestsGenerator = <T>(): WritableArray<PromiseWithSpec<T>> => [];
 
 let maybeFulfilled: WritableArray<KillablePromise<any>> = [];
 
-export const cleanThrottledPromises = (): void => {
-  maybeFulfilled.forEach((promise) => promise.killPromise());
+export const cleanThrottledPromises = (killPromise = true): void => {
```
melton-jason, line 13: Maybe `killPromises` might be a better name for the parameter name? 
We are not killing one promise, we are potentially killing multiple.

melton-jason (approved):
SInce there is now a parameter involved with `cleanThrottledPrimses`, I think it would be safer to wrap the cleanup functions of React useEffects in an anonymous function.

https://github.com/specify/specify7/blob/7e3bf8afd37f22db9dc37dcba94f0b707994d27c/specifyweb/frontend/js_src/lib/components/Statistics/index.tsx#L300

Would become 
```ts
return () => cleanThrottledPromises(); 
```
Not sure if React will ever put a value in the parameter when calling the cleanup function, but this is the safest bet. 
Also need to do this for https://github.com/specify/specify7/blob/7e3bf8afd37f22db9dc37dcba94f0b707994d27c/specifyweb/frontend/js_src/lib/components/Statistics/AddStatDialog.tsx#L58

specifyweb/frontend/js_src/lib/utils/ajax/throttledPromise.ts
realVinayak: You do kill a single promise when you do promise.killPromise(). It's just that you do it for every single one of them.

realVinayak:
https://github.com/specify/specify7/pull/3950#pullrequestreview-1598295129

That change shouldn't be necessary. React would just call the function, and not pass any arguments. How will it know to do that?

specifyweb/frontend/js_src/lib/utils/ajax/throttledPromise.ts
melton-jason: Sure, but some to a consumer of the function that distinction is not made clear until they look at the implementation. 
I think `killPromises` makes the intention behind the parameter immediately clear: you put in a truthy value, then multiple promises will be killed.

Ultimately, this is just semantics. The function implementation should only be a click (or hover) away from most developers :shrug:

melton-jason:
> [#3950 (review)](https://github.com/specify/specify7/pull/3950#pullrequestreview-1598295129)
> 
> That change shouldn't be necessary. React would just call the function, and not pass any arguments. How will it know to do that?

If you're sure that is the case, then no changes needed! :smile: 
I _assumed_ this was the case anyways, but I am not 100% sure so I erred on the side of caution.

realVinayak:
https://github.com/specify/specify7/pull/3950#issuecomment-1695717577

No need to take shots in the dark when you can just test it: https://playcode.io/1575606. Nothing provided as args

CarolineDenis (approved):

carlosmbe:
Can Testing Instructions be added @realVinayak 
Otherwise, trying the initial issue seems fixed. I just don't know what else this could have touched hence my request for instructions.
Thanks.
OwO

CarolineDenis (approved):

realVinayak:
@carlosmbe @specify/ux-testing 
To TEST:
1. Clicking refresh multiple times shouldn't stop items from loading
2. Loads stats page. Before all the items get the time to load, close it. Go back to stats page, those items should load automatically,

bronwyncombs (approved):
Stats page loads even after spamming refresh button. Will also load like normal after leaving the page and then going back while loading in.

grantfitzsimmons:
@realVinayak Merge when ready.

realVinayak:
waiting for @carlosmbe approval

carlosmbe (approved):

PR #3958, opened 2023-08-30
Don't pass empty string to uniqueIn when scope undefined

### LINKED ISSUE #3953: Extra "accessing unknown field" warnings in v7.9, as compared to production
There are some extra warnings when navigating to specify. The first image is v7.9-dev, and the second is production. Need to figure out why that warning appears, to make sure this isn't a code smell.
image
image

melton-jason:
Fixes #3953 
As discovered by @CarolineDenis, the issue was that I was incorrectly casting an undefined `scope` value to an empty string when it is undefined and did not implement any logic to handle that case. 
Thus, when Specify tried to get the field with value ` `, it caused the console warning  and returned undefined. 

#3953 should have only occurred in cases where a field should be unique at a database level:
  - Collectingevent uniqueIdentifier
  - CollectionObject uniqueIdentifier
  - Institution name
  - Locality uniqueIdentifier
  - Permit permitNumber
  - SpAppResourceData spAppResource
  - Specifyuser name
 
 TO TEST:
Navigate to a form which has one of the above uniqueness rules and change a value in the mentioned field. There should be no warning in the console

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
@@ -167,7 +167,9 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
           scope = uniqueRule.field;
         }
         return this.uniqueIn(
-          ((scope ?? '') as string).toLowerCase(),
+          typeof scope === 'string'
+            ? scope.toLowerCase()
+            : (scope as undefined),
           fieldNames
         );
       }


CarolineDenis (approved):

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -167,7 +167,9 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
           scope = uniqueRule.field;
         }
         return this.uniqueIn(
-          ((scope ?? '') as string).toLowerCase(),
+          typeof scope === 'string'
+            ? scope.toLowerCase()
+            : (scope as undefined),
```
maxxxxxdlp, line 172:
```suggestion
          scope?.toLowerCase()
```

maxxxxxdlp (changes requested):

melton-jason:
Implemented in #3962

maxxxxxdlp:
for the future: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

melton-jason:
I am aware of how to change the base branch of a PR :sweat_smile: 
However, changing it directly to `v7.9-dev` would _also_ include every commit in `fix-busrule-tests`, and there is still some work to be done before `fix-busrule-tests` can be considered resolved. 

Ultimately, I figured this might be the best way of handling this mistake. 
Either way, a PR would need to be closed and one opened. 

Definitely a learning experience, regardless.

PR #3963, opened 2023-09-01
Don't clone fields unique to database

### LINKED ISSUE #3960: Permit Number Being Cloned
Permit Number should not be carried over since it's a unique field. Saving the permit leads to a business rule error.

https://github.com/specify/specify7/assets/53784701/bdbb2e9e-57f5-4625-9801-e1de82ed3705

[Specify 7 Crash Report - 2023-08-31T20_06_20.320Z.txt](https://github.com/specify/specify7/files/12489909/Specify.7.Crash.Report.-.2023-08-31T20_06_20.320Z.txt)

melton-jason:
Fixes #3960 

The following fields should not be carried over/cloned: 
> * Collectingevent uniqueIdentifier
> * CollectionObject uniqueIdentifier
> * Institution name
> * Locality uniqueIdentifier
> * Permit permitNumber
> * SpAppResourceData spAppResource
> * Specifyuser name

(Also, there was another issue in which global uniqueness rules were not working on the frontend. That fix should be included in #3962)

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
@@ -270,7 +270,8 @@ export const getUniqueFields = (model: SpecifyModel): RA<string> =>
     ...Object.entries(businessRuleDefs[model.name]?.uniqueIn ?? {})
       .filter(
         ([_fieldName, uniquenessRules]) =>
-          uniquenessRules in schema.domainLevelIds
+          uniquenessRules in schema.domainLevelIds ||
+          uniquenessRules === undefined
       )
       .map(([fieldName]) => model.strictGetField(fieldName).name),
     /*
--- a/specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
@@ -268,11 +268,6 @@ const uniqueFields = [
 export const getUniqueFields = (model: SpecifyModel): RA<string> =>
   f.unique([
     ...Object.entries(businessRuleDefs[model.name]?.uniqueIn ?? {})
-      .filter(
-        ([_fieldName, uniquenessRules]) =>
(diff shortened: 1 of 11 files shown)


specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
```diff
@@ -270,7 +270,8 @@ export const getUniqueFields = (model: SpecifyModel): RA<string> =>
     ...Object.entries(businessRuleDefs[model.name]?.uniqueIn ?? {})
       .filter(
         ([_fieldName, uniquenessRules]) =>
-          uniquenessRules in schema.domainLevelIds
+          uniquenessRules in schema.domainLevelIds ||
+          uniquenessRules === undefined
```
maxxxxxdlp, line 274: (I seen https://github.com/specify/specify7/issues/3960#issuecomment-1702819456)

can you double-check that this doesn't mark absolutely every field as unique, and rather only fields that have uniqueness constraints of some sort?

also, if you have a moment to spare, can you update/add the test cases?

CarolineDenis (approved):

grantfitzsimmons (changes requested):
When cloning accessions, I still get this error when there is an existing agent:
https://github.com/specify/specify7/issues/3967

Is this in the scope of this PR?

melton-jason:
> When cloning accessions, I still get this error when there is an existing agent: #3967
> 
> Is this in the scope of this PR?

That issue (#3967) has been around since at _least_ `v7.8.10` (and possibly earlier, I just stopped checking previous versions).
I do not think there is enough overlap in underlying cause to justify that solution being included in this PR. 

_However_, it still could be included in the `v7.9` release in it's own PR if it is important enough (at the cost of further delaying the release)

grantfitzsimmons (approved):

grantfitzsimmons:
> I do not think there is enough overlap in underlying cause to justify that solution being included in this PR.

I agree. Let's fix at a later time.

realVinayak (approved):

PR #3969, opened 2023-09-01
Create a tree split widget 

### LINKED ISSUE #3600: Split Tree View
> It would be lovely to see the split tree view that was present in Specify 6 in Specify 7. The was where you could have two windows open, stacked one above the other on the same screen and scroll to different parts of the tree. This would be particularly useful for synonymy and merging nodes in our taxon tree which is rather large!

**Requested By:** Sally King on [Speciforum](https://discourse.specifysoftware.org/t/split-tree-view-in-specify-7/1197)

- _Liked by Liath Appleton at UT Austin_

CarolineDenis:
Fixes #3600


specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -386,24 +405,59 @@ export function TreeViewWrapper(): JSX.Element | null {
...
+            isSplit={[isSplit, setIsSplit]}
+            isFirst={isFirst}
+          />
+        </IsTreeSplitContext.Provider>
+      );
+    } else {
+      return null;
+    }
+  };
+
```
CarolineDenis, line 437: @maxxxxxdlp , 
What would be your suggestion to have a re-render of the first TreeView when doing an edit in the second and vice-versa? At the moment a page reload is necessary

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
maxxxxxdlp, line 1: (optional)
idea:
instead of rendering two `<TreeView />` when split, how about having a single `<TreeView />` render two trees when split?

this way, the header would only be rendered once, and so you would avoid all the complication of rendering button labels vs button icons

while you can see two trees at once, you can only really interact with one tree at a time - so if you interacted with the tree on the left last, the buttons at the top will work for that tree - and if you clicked on the right tree, that tree get's focus style and buttons now work for it

that would mean extracting the table in TreeView into a separate component - probably a good idea anyway as that component is too big

thoughts?

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
maxxxxxdlp: since each tree row fetches it's children, rather than having all fetched children in the `<TreeView>` component, you have some choices:
1. Refactor the code so that the "rows" state for all rows is stored in "TreeRow"
2. Set event listeners and listen for them. remember you set an event listener for queries being deleted, so that the query list dialog can remove them from the list? this would be similar but on a larger scale
    - an event is emited when tree item is deleted - all rows listen for that event and check if one of their children is deleted, and if so, update the state accordingly
    - an event is emited when tree item is created - all rows listen for that event and check if they are the parent of the new child, and if so, update the state accordingly
3. Make it so that the two TreeRows that represent the same tree item share the same "rows" state - i.e instead of `const [rows, setRows] = React.useState<RA<Row> | undefined>(undefined);`, do `const [rows, setRows] = useRows(row.nodeId);` - and that hooks checks if it was already called with a node with this id, and if so, returns the same "rows" state rather than creating a new one

My recommendation is for option 2 or 3
2 is the simplest, and would benefit other places too (setting events and listening to them would be useful for other places in specify - i.e it already benefited the query list dialog by letting it delete removed queries - it could benefit other places in the future too)
3 is the best in terms of performance - this way two split trees won't have to fetch the same children again, as they are sharing the children

feel free to ask Jason or Vinny or me for help on either of these solutions or ask for more implementation details

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -386,24 +405,59 @@ export function TreeViewWrapper(): JSX.Element | null {
...
-        />
-      ) : null}
+      <>
+        {isSplit ? (
+          <div className="h-full w-full">
+            <Splitter
+              key="test"
+              position={'horizontal'}
+              primaryPaneMinWidth={0}
+              primaryPaneWidth="30%"
```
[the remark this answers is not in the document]
CarolineDenis: Yes I had a button to switch from vertical to horizontal and vice versa, hence the min/max width and min/max height. But after discussion with Grant we decided to not implement this and therefore I removed the feature but forgot to erase the props.

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
CarolineDenis: I like the idea I can try to implement that.

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
CarolineDenis: with the refactoring of Tree to a separate component the updates is done now

CarolineDenis: although now, with the refactoring of Tree to a separate component the update is functioning.

specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
```diff
@@ -77,7 +77,7 @@ export function TreeViewActions<SCHEMA extends AnyTree>({
         <li className="contents">
           {typeof focusedRow === 'object' ? (
             isRoot ? (
-              <Button.Small onClick={undefined}>
+              <Button.Small onClick={undefined} title={queryText.query()}>
```
maxxxxxdlp, line 80: now that the button contains text again, there is no need for a title (especially since it's redundant)

I said earlier that title and aria-label is needed because the button was displaying only an icon without text

maxxxxxdlp: same applies to all other button

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
```diff
@@ -0,0 +1,254 @@
...
+    async (nodeId: number | 'null', rankId: number): Promise<Stats> =>
+      rankId >= statsThreshold
+        ? fetchStats(`${baseUrl}/${nodeId}/stats/`)
+        : Promise.resolve({}),
+    [baseUrl, statsThreshold]
+  );
+
+  return (
+    <div
+      className={`
```
maxxxxxdlp, line 99: please reformat this when you get a chance

also, if you are sick of doing that by hand, feel free to consider using a library like https://www.npmjs.com/package/classnames OR finding an esling plugin/prettier plugin that will do the formating of multi-line class names for you

maxxxxxdlp: the Tailwind ESLint plugin we are using is not official because official one didn't exist back then

now it does - https://www.npmjs.com/package/eslint-plugin-tailwindcss

it might handle this case better - will need to try at some point (not for this PR)

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
```diff
@@ -0,0 +1,254 @@
...
+        );
+      }}
+    >
+      <div role="none rowgroup">
+        <div role="none row">
+          {treeDefinitionItems.map((rank, index, { length }) => {
+            const rankName = rank.title || rank.name;
+            return (
+              <div
+                className={`
```
maxxxxxdlp, line 137: same

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -150,49 +118,65 @@ function TreeView<SCHEMA extends AnyTree>({
...
-          onClick={(): void => {
-            setFocusPath([0]);
-            setConformation([]);
-          }}
-        >
-          {commonText.collapseAll()}
-        </Button.Small>
-        <span className="-ml-2 flex-1" />
+      <header
+        className={`flex items-center gap-2 overflow-x-auto sm:flex-wrap sm:overflow-x-visible`}
```
maxxxxxdlp, line 146: (optiaonl) could simply a bit here by using just `"` instead of {`

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -150,49 +118,65 @@ function TreeView<SCHEMA extends AnyTree>({
...
+            disabled={conformation.length === 0}
+            onClick={(): void => {
+              setFocusPath([0]);
+              setConformation([]);
+            }}
+          >
+            {commonText.collapseAll()}
+          </Button.Small>
+          <Button.Small onClick={() => setIsSplit(!isSplit)}>
+            {isSplit ? treeText.join() : treeText.split()}
```
maxxxxxdlp, line 176: (optional)

rename "Split" to "Split View" to be clearer to the user?

could also add `aria-pressed={isSplit}` - if you do that, then you won't need the "Join" string (could just display "Split View" always

maxxxxxdlp (approved):
looking good for the most part. I will do a deeper review when you say it's ready

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -386,24 +405,59 @@ export function TreeViewWrapper(): JSX.Element | null {
       : undefined;
 
   if (treeName === undefined || !isTreeModel(treeName)) return <NotFoundView />;
+
+  const [isSplit, setIsSplit] = React.useState(false);
```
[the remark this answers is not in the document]
CarolineDenis: It was decided to keep the same behavior as 6, therefore to not cache it. But we could add the feature later if we have requests.

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -168,16 +194,28 @@ export function TreeRow({
...
               style={{
                 color:
                   typeof row.acceptedId === 'number' ? synonymColor : undefined,
               }}
-              onClick={({ metaKey, shiftKey }): void =>
-                metaKey || shiftKey ? handleFocusNode([]) : handleToggle(false)
-              }
+              onClick={({ metaKey, shiftKey }): void => {
+                metaKey || shiftKey ? handleFocusNode([]) : handleToggle(false);
+                isFirst
```
maxxxxxdlp, line 211: 1. Rather than do this on every node (and have to pass down the setLastTreeClicked prop, why not just set the event listener in Tree or TreeView component?
2. Instead of "click", you should probably be listening for focus - click is mouse only, where as focus is both mouse and keyboard
3. there lastTreeClicked shouldn't really ever be undefined - just set it to `"first"` by default - will simplify things a bit as no need to check for undefined anymore

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -105,26 +116,41 @@ export function TreeRow({
...
   function handleToggle(focusChild: boolean): void {
-    if (row.children === 0) {
-      handleFocusNode([]);
-      // If children are later added, the node would expand because of this line
-      handleChangeConformation([]);
-    } else if (Array.isArray(conformation)) {
-      previousConformation.current = conformation;
-      handleChangeConformation(undefined);
-      handleFocusNode([]);
+    if (isFirst) {
```
maxxxxxdlp, line 119: lots of duplication of complicated code here - bad

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -63,7 +69,12 @@ export function TreeRow({
   );
 
   React.useEffect(() => {
-    if (Array.isArray(focusPath) && focusPath.length === 0) setFocusedRow(row);
+    if (
+      Array.isArray(focusPath) &&
+      focusPath.length === 0 &&
+      lastTreeClicked === 'first'
```
maxxxxxdlp, line 75: this doesn't look right
1. only the first tree can set focused row? 2nd tree can't? that doesn't seem right
2. even if that was desired, instead of checking for that here, the check should be inside the setFocusedRow function, not here
3. but really, both trees should be able to have independent focused row states, and calling setFocusedRow updates correct state - no checks should be needed here

maxxxxxdlp: ideally, the tree shouldn't even know if it's first or 2nd or if it's focused or not - the difference in behavior would be facilitated by providing different callbacks, not by providing magic flags like "isFirst"

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -53,6 +56,9 @@ export function TreeRow({
   readonly setFocusedRow: (row: Row) => void;
   readonly synonymColor: string;
   readonly treeName: string;
+  readonly isFirst: boolean;
+  readonly isSecondFocused: GetSet<number | undefined>;
```
maxxxxxdlp, line 60: doesn't look right. variables starting with is are of boolean type, not number - rename. `secondTreeFocusedNode`? but really, I don't understand why you need that here

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -168,16 +194,28 @@ export function TreeRow({
                     ? 'outline outline-1 outline-blue-500'
                     : ''
                 }
+                ${
```
maxxxxxdlp, line 197: you should modify the ternary above rather than add a new ternary here - the code above already checks if tree is focused or is in action - you should just modify that code

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -168,16 +194,28 @@ export function TreeRow({
...
+                  isSecondFocused === row.nodeId
+                    ? 'outline outline-1 outline-green-500'
```
maxxxxxdlp, line 199: does this match sp6 behavior? if not, then I don't see a reason to do it this way - instead of introducing yet another color that user has to learn and remember, why not just use a dimmed blur when not focused or gray color. OR thiker outline in active tree

that also has a benefit of better a11y (for people with color blindness)

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -118,28 +98,17 @@ function TreeView<SCHEMA extends AnyTree>({
...
-  );
-
   const [rows, setRows] = useAsyncState<RA<Row>>(
     React.useCallback(async () => getRows('null'), [getRows]),
     true
   );
-  const id = useId('tree-view');
 
   // FEATURE: synchronize focus path with the URL
   const [focusPath = [], setFocusPath] = useCachedState(
```
maxxxxxdlp, line 107: the focus paths should be independent for both trees
as far as caching, store the most recently focused tree's focused path in this cache

<details>
<summary>Solution</summary>
two react states, and a useEffect that updates local storage on change to focused tree's focus path
</details>

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -150,49 +119,67 @@ function TreeView<SCHEMA extends AnyTree>({
...
-          disabled={conformation.length === 0}
-          onClick={(): void => {
-            setFocusPath([0]);
-            setConformation([]);
-          }}
-        >
-          {commonText.collapseAll()}
-        </Button.Small>
-        <span className="-ml-2 flex-1" />
+        <>
```
maxxxxxdlp, line 148: `<>` not needed

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -210,167 +197,26 @@ function TreeView<SCHEMA extends AnyTree>({
...
-                    <EditTreeRank rank={rank} />
-                  ) : undefined}
-                </div>
-              );
-            })}
-          </div>
+      {isSplit ? (
+        <div className="h-full w-full">
+          <Splitter
+            key="test"
```
maxxxxxdlp, line 203: `key` shouldn't be needed

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -210,167 +197,26 @@ function TreeView<SCHEMA extends AnyTree>({
...
+            position={'horizontal'}
```
maxxxxxdlp, line 204: hmm, I though you were implementing a vertical split, not horizontal.
horizontal split doesn't much sense, unless you have a square monitor (on specify open in half screen)
trees when expanded are very tall, not very wide

@grantfitzsimmons thougs?

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -150,49 +119,67 @@ function TreeView<SCHEMA extends AnyTree>({
...
-  );
-  const [synonymColor] = userPreferences.use(
-    'treeEditor',
-    treeToPref[tableName],
-    'synonymColor'
-  );
+  const [isSplit, setIsSplit] = React.useState(false);
+
+  const treeContainer = (isFirst: boolean) => {
+    return rows === undefined ? null : (
```
maxxxxxdlp, line 125: remove `{` and `return`

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
```diff
@@ -0,0 +1,267 @@
...
+        ? fetchStats(`${baseUrl}/${nodeId}/stats/`)
+        : Promise.resolve({}),
+    [baseUrl, statsThreshold]
+  );
+
+  const [isSecondFocused, setIsSecondFocused] = React.useState<
+    number | undefined
+  >(undefined);
+
+  const [lastTreeClicked, setLastTreeClicked] = React.useState<
```
maxxxxxdlp, line 103: you are setting this, but never reading as far as I can see.
is it just unfinished code or a bug?

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
grantfitzsimmons: Can we ask the #sp7-7-ui-ux channel about this? It's interesting– I figured we would recreate the 6 tree split interface but it was conceived in a different era.

CarolineDenis:
https://github.com/specify/specify7/assets/108160931/f828f6b7-dfde-463e-8c0a-a94a13816158

https://github.com/specify/specify7/assets/108160931/48eb7cc8-0550-4474-94a4-d522ec0ee83b

https://github.com/specify/specify7/assets/108160931/a3b9f022-aa03-4b83-8eb2-64354adcf66a

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
maxxxxxdlp: ah, I didn't remember what sp6 looked like, but just assumed that vertical split makes more sense given the common dimensions of monitors, and the shape of the tree

(i.e having windows side by side is common, or having one or two sidebars in a webpage is common)

also ask ChatGPT

but feel free to ignore me: I am not a biologist, and not the one who will be using this feature

maxxxxxdlp: edit:
the videos Caroline sent (https://github.com/specify/specify7/pull/3969#issuecomment-1718282639) indeed work better with horizontal - but those videos contain test data where a state has only a few counties

in real trees, genus often contains a lot of species, or subspecies contains many species, and (I am assuming) it's more valuable to be able to see neighboring species within a given genus, than neighboring genera - thus the data is taller rather than wider

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -104,14 +104,31 @@ function TreeView<SCHEMA extends AnyTree>({
...
+  );
+
+  const [lastFocusedTree, setLastFocusedTree] = React.useState<boolean>(true);
+
+  React.useEffect(() => {
+    if (lastFocusedTree) setFocusedRow(focusedRow1);
+    else {
+      setFocusedRow(focusedRow2);
+    }
+  }, [focusedRow1, focusedRow2]);
```
maxxxxxdlp, line 131: This doesn't need to be a useEffect. Simplify:

```
const focusedRow = lastFocusedTree ? focusedRow1 : focusedRow2;
```

and then focusedRow is not a state, but just a simple variable

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -104,14 +104,31 @@ function TreeView<SCHEMA extends AnyTree>({
   );
 
   // FEATURE: synchronize focus path with the URL
-  const [focusPath = [], setFocusPath] = useCachedState(
+  const [_focusedPath = [], setFocusPath] = useCachedState(
     'tree',
     `focusPath${tableName}`
   );
+  const [focusPath1 = [], setFocusPath1] = React.useState<RA<number>>();
```
maxxxxxdlp, line 111: (optional) create a hook that will return all the states you need in a single object, and then just call that hook twice.

i.e:
```ts
const leftTreeStates = useTreeStates();
const rightTreeStates = useTreeStates();
```

or even:
```ts
const treeStates = {
  left: useTreeStates(),
  right: useTreeStates(),
};

const [lastFocusedTree, setLastFocusedTree] = React.useState<'left' | 'right'>('left');
const state = treeStates[lastFocusedTree];
```

now treeStates.left always stores the states specific to the left tree, treeStates.right always stores the states specific to the right tree, and state always denotes the currently focused tree

maxxxxxdlp (approved):
code looks good overall, and I like the videos you sent

when you think you are mostly done making changes, please tell me and I will do a deeper look at the code to make sure everything is perfect

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -118,50 +98,63 @@ function TreeView<SCHEMA extends AnyTree>({
...
   // FEATURE: synchronize focus path with the URL
-  const [focusPath = [], setFocusPath] = useCachedState(
+  const [_focusedPath = [], setFocusPath] = useCachedState(
     'tree',
     `focusPath${tableName}`
   );
-  const [focusedRow, setFocusedRow] = React.useState<Row | undefined>(
-    undefined
-  );
+  const [focusPathTop = [], setFocusPathTop] = React.useState<RA<number>>();
```
maxxxxxdlp, line 111: the default value for these should be coming from the `_focusedPath`

in that case consider renaming `_focusedPath` to cachedFocusPath or defaultFocusPath or previousFocusPath...

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -118,50 +98,63 @@ function TreeView<SCHEMA extends AnyTree>({
...
+  const [focusPathBottom = [], setFocusPathBottom] =
```
maxxxxxdlp, line 112: focus path top and bottom should probably be moved into useTreeStates too

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -118,50 +98,63 @@ function TreeView<SCHEMA extends AnyTree>({
...
+    React.useState<RA<number>>();
+
+  const treeStates = {
+    top: useTreeStates(),
+    bottom: useTreeStates(),
+  };
+
+  const [lastFocusedTree, setLastFocusedTree] = React.useState<
+    'top' | 'bottom'
```
maxxxxxdlp, line 121: (option) I know I suggested using `'top' | 'bottom'`, but given that you have both vertical and horizontal, perhaps now it would be better to use `'first' | 'second'`?

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -118,50 +98,63 @@ function TreeView<SCHEMA extends AnyTree>({
...
-    treeToPref[tableName],
-    'treeAccentColor'
-  );
-  const [synonymColor] = userPreferences.use(
-    'treeEditor',
-    treeToPref[tableName],
-    'synonymColor'
-  );
+  const [isSplit, setIsSplit] = React.useState(false);
+  const [horizontal, setHorizontal] = React.useState(true);
```
maxxxxxdlp, line 131: (optional) useCachedState for both of these? or at least for `horizontal`

maxxxxxdlp, line 131: also, could you please rename `horizontal` to `isHorizontal`?

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -192,11 +185,22 @@ function TreeView<SCHEMA extends AnyTree>({
         >
           {commonText.collapseAll()}
         </Button.Small>
+        <Button.Small
+          aria-pressed={isSplit}
+          onClick={() => setIsSplit(!isSplit)}
+        >
+          {treeText.splitView()}
+        </Button.Small>
```
maxxxxxdlp, line 193: (optional) this is up to you and Grant and others, but you may want to consider having just one button for this, and have that button go between 3 states: no split view, horizontal split view, and vertical split view

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
maxxxxxdlp: the hook's return type could look something like this:

```ts
RR<
  'top' | 'bottom',
  {
    readonly focusedRow: GetOrSet<Row | undefined>;
    readonly focusPath: GetOrSet<RA<number>>;
  }
>
```

and then later in the file where you have:

```ts
          isTop
            ? [focusPathTop, setFocusPathTop]
            : [focusPathBottom, setFocusPathBottom]
```

replace by:
```ts
   treeStates[type].focusPath
```

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -118,50 +98,63 @@ function TreeView<SCHEMA extends AnyTree>({
...
     `focusPath${tableName}`
   );
-  const [focusedRow, setFocusedRow] = React.useState<Row | undefined>(
-    undefined
-  );
+  const [focusPathTop = [], setFocusPathTop] = React.useState<RA<number>>();
+  const [focusPathBottom = [], setFocusPathBottom] =
+    React.useState<RA<number>>();
+
+  const treeStates = {
```
maxxxxxdlp, line 115: (optional) in this file everything is about trees, so could rename `treeStates` to `states` to be a bit less redunant

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
maxxxxxdlp: also, whenever setFocusPathTop is called, setFocusPath should be called too (so that local storage is updated with new path


<p>

 to accommodate that, move the useCachedState hook inside the useTreeStates

</p>
</details> 

<p>

 to accommodate that, move the useCachedState hook inside the useTreeStates, then that hook instead of returning setFocusPathTop, will return a React.useCallback, which in turn will call both setFocusPathTop and setFocusPath (note variable names would be different in that case - I am using the variable names you have here for now - i.e "Top" won't make sense inside the useTreeStates)

</p>
</details>

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -118,50 +98,63 @@ function TreeView<SCHEMA extends AnyTree>({
...
-  );
-  const [synonymColor] = userPreferences.use(
-    'treeEditor',
-    treeToPref[tableName],
-    'synonymColor'
-  );
+  const [isSplit, setIsSplit] = React.useState(false);
+  const [horizontal, setHorizontal] = React.useState(true);
+
+  const treeContainer = (isTop: boolean) =>
```
maxxxxxdlp, line 133: this is a great idea! you are deduplicating a lot of code by creating this inline function

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -118,50 +98,63 @@ function TreeView<SCHEMA extends AnyTree>({
...
+            : [focusPathBottom, setFocusPathBottom]
+        }
+        focusRef={toolbarButtonRef}
+        getRows={getRows}
+        isEditingRanks={isEditingRanks}
+        ranks={rankIds}
+        rows={rows}
+        searchBoxRef={searchBoxRef}
+        setFocusedRow={
+          isTop ? treeStates.top.setFocusedRow : treeStates.bottom.setFocusedRow
```
maxxxxxdlp, line 151: this function should be called with `type` instead of isTop

that would simplify this expression, and the setLastFocusedTree below, and the focusPath above (given you also apply the changes to useTreeStates)

maxxxxxdlp: then, consider creating a typescript type `type TreeType = 'first' | 'second';` since at that point that type would be used in several places

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -210,167 +214,28 @@ function TreeView<SCHEMA extends AnyTree>({
...
-              );
-            })}
-          </div>
+      {isSplit ? (
+        <div className="h-full w-full">
+          <Splitter
+            position={horizontal ? 'horizontal' : 'vertical'}
+            primaryPaneHeight="40%"
+            primaryPaneMaxHeight="80%"
+            primaryPaneMinHeight={1}
```
maxxxxxdlp, line 223: I don't know the library so this might not be an issue, but it's a bit weird that primaryPaneMinHeight is 1, but primaryPaneMinWidth is 0

does 0 and 1 mean 0px and 1px or something else? and shouldn't both of these be the same value?

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -210,167 +214,28 @@ function TreeView<SCHEMA extends AnyTree>({
...
+            primaryPaneMaxWidth="80%"
+            primaryPaneWidth="50%"
```
maxxxxxdlp, line 225: (optional) (feel free to open a separate issue for this) remember primaryPaneHeight and primaryPaneWidth between sessions?

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -175,9 +177,11 @@ export function TreeRow({
                 color:
                   typeof row.acceptedId === 'number' ? synonymColor : undefined,
               }}
-              onClick={({ metaKey, shiftKey }): void =>
-                metaKey || shiftKey ? handleFocusNode([]) : handleToggle(false)
-              }
+              onClick={({ metaKey, shiftKey }): void => {
+                metaKey || shiftKey
+                  ? handleFocusNode([])
+                  : (handleToggle(false), setLastFocusedTree());
```
maxxxxxdlp, line 183: this is hacky (avoid comma expressions as they can create bug prone and unreadable code), and buggy - setLastFocusedTree is only called on mouse click, not onKeyDown - why not call `setLastFocusedTree` inside of handleToggle instead?

maxxxxxdlp: but EVEN BETTER, why not call setLastFocusedTree() when `setFocusPath` is called in `const treeContainer = ...` in `index.tsx`?

that way, this file won't need to be changed at all - to keep code cleaner and decoupled this file should not know anything about the fact that multiple trees can be displayed as that is above it - this file is only concerned with displaying a single tree row, and calling some callbacks on interactions

maxxxxxdlp (changes requested):
overall quite good

you learned a lot over the last 9 months!!!

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -192,18 +182,18 @@ function TreeView<SCHEMA extends AnyTree>({
...
         <span className="-ml-2 flex-1" />
         <ErrorBoundary dismissible>
           <TreeViewActions<SCHEMA>
             actionRow={actionRow}
-            focusedRow={treeStates[lastFocusedTree].focusedRow}
+            focusedRow={states[lastFocusedTree].focusedRow}
             focusRef={toolbarButtonRef}
             ranks={rankIds}
-            setFocusPath={setFocusPath}
+            setFocusPath={states[lastFocusedTree].setFocusPath}
```
maxxxxxdlp, line 196: is there a reason why this isn't using `setFocusPathAndCachePath`?

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -274,12 +264,40 @@ export function TreeViewWrapper(): JSX.Element | null {
   );
 }
 
-function useTreeStates() {
+function useStates<SCHEMA extends AnyTree>(
+  tableName: SCHEMA['tableName']
+): {
+  readonly focusedRow: Row | undefined;
```
maxxxxxdlp, line 270: I was going to use a GetOrSet<> type here. Thus, return type would be this:

```ts
{
  readonly focusedRow: GetOrSet<Row | undefined>;
  readonly focusPath: GetOrSet<RA<number>>;
}
```

and the actual return statement like this:

```ts
return {
  focusRow: React.useState(undefined),
  focusPath: [focusPath, setFocusAndCachePath],
};
```

That should be a bit simpler and cleaner

There is no reason to expose both setFocusPath and setFocusAndCachePath to the outside - only setFocusAndCachePath should be exposed

---

though?

specifyweb/frontend/js_src/lib/utils/cache/definitions.ts
```diff
@@ -75,6 +75,10 @@ export type CacheDefinitions = {
   } & {
     readonly /** Open nodes in a given tree */
     [key in `conformations${AnyTree['tableName']}`]: Conformations;
+  } & {
+    readonly isSplit: boolean;
+  } & {
+    readonly isHorizontal: boolean;
```
maxxxxxdlp, line 81:
```suggestion
  } & {
    readonly isSplit: boolean;
    readonly isHorizontal: boolean;
```

maxxxxxdlp: you can't mix the `{ [key in ...]: ... }` and `{ key: ...}` styles in the same type at the same time, which is why I had `&` in there before

but you can have two `{ key: ...}` `{ key: ...}` in the same object without the need for `&`

maxxxxxdlp (approved):
👍
🎆

grantfitzsimmons (changes requested):
image

Encountered here, could not recreate on `production`

I am seeing opaque backgrounds behind the rank names but not a full background occupying the whole column width like in production

image
image


So the rank name backgrounds should be full width instead of occupying just the space behind the rank name

grantfitzsimmons (changes requested):
When switching between vertical and horizontal split, it should reset to be in the center of the screen rather than maintaining the proportion of the split from the previous view.

grantfitzsimmons (changes requested):
https://github.com/specify/specify7/assets/37256050/cbc9293d-cc46-4e30-aaab-ed309b1b70d9

**1.** When expanding nodes, the tree can go off the page, which leads to a second scroll bar appearing. The tree viewer should be contained within the view area when the window is first opened.

**2.** The option to change the split form horizontal to vertical should always be visible, but disabled, to insinuate to the user that this only useful when in split view.

**3.** The "collapse all" button is dangerously close to the split view button:
image

**4.** The tree area viewer has shadows that are cut off and the rounding looks bad :-(
image

**5.** The gap between the two views is inconsistent, and the space is not matched:
image

Sorry if this is a little nit-picky, just wanted to surface all of my design & function concerns 😂 just for the record

grantfitzsimmons:
Screenshot 2023-09-29 at 1 20 59 PM

Can we remove the gap? @realVinayak and @bronwyncombs are not sold on it, and neither am i 😢

CarolineDenis:
> Screenshot 2023-09-29 at 1 20 59 PM
> Can we remove the gap? @realVinayak and @bronwyncombs are not sold on it, and neither am i 😢

:weary: ok :sweat_smile:

realVinayak:
Can the active node be highlighted? @grantfitzsimmons what do you think?

image

CarolineDenis:
> Can the active node be highlighted? @grantfitzsimmons what do you think?
> 
> image

which the current implementation it is not possible to have a different color for each tree for the borders of the focused node

realVinayak (approved):

grantfitzsimmons (approved):
:+1:

specifyweb/frontend/js_src/lib/components/Forms/DeleteButton.tsx
```diff
@@ -77,16 +79,27 @@ export function DeleteButton<SCHEMA extends AnySchema>({
...
-        }}
-      >
-        {isBlocked ? icons.exclamation : undefined}
-        {commonText.delete()}
-      </ButtonComponent>
+      {isIcon ? (
+        <Button.Icon
+          title={isBlocked ? formsText.deleteBlocked() : commonText.delete()}
+          icon="trash"
+          onClick={(): void => {
```
maxxxxxdlp, line 86: (optional) can you extract the callback into a common function instead of duplicating it two times?

specifyweb/frontend/js_src/lib/components/Forms/DeleteButton.tsx
```diff
@@ -77,16 +79,27 @@ export function DeleteButton<SCHEMA extends AnySchema>({
...
-          handleOpen();
-          setDeferred(false);
-        }}
-      >
-        {isBlocked ? icons.exclamation : undefined}
-        {commonText.delete()}
-      </ButtonComponent>
+      {isIcon ? (
+        <Button.Icon
+          title={isBlocked ? formsText.deleteBlocked() : commonText.delete()}
```
maxxxxxdlp, line 84: add aria-label too (whenever interactive element does not have a text content or an associated label (in case of input field), it must have aria-label)
in this case, the aria-label can be the same as the title

maxxxxxdlp: ah forgot that I made it automatic. see this:
https://github.com/specify/specify7/blob/b5d64d14e659c3e8c50d0200f80954725cc114f4/specifyweb/frontend/js_src/lib/components/Atoms/Button.tsx#L134

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -147,6 +148,12 @@ function TreeView<SCHEMA extends AnyTree>({
       />
     );
 
+  // Used to force dimensions of panes to go back to default between orientations changes
+  const [splitterKey, setSplitterKey] = React.useState(1);
```
maxxxxxdlp, line 152: smart

maxxxxxdlp (approved):
all good!

grantfitzsimmons (changes requested):
On narrow screens, the slider is made incredibly small

https://github.com/specify/specify7/assets/37256050/7e4c9514-ca1a-48a0-ac33-652ad54b7115

maxxxxxdlp:
should it fallback to horizontal on narrow screens? or hide the split view?

grantfitzsimmons:
@maxxxxxdlp Split view should probably not be supported on narrow screens.

CarolineDenis:
@maxxxxxdlp , do you have a better solution than the magic number '670' for narrow screen checker to disable split option?

maxxxxxdlp:
@CarolineDenis (optional) you can define the threathhold in rem instead of px if you want

( in specifyweb/frontend/js_src/lib/components/Atoms/index.tsx there is a variable called `oneRem` which corresponds to `1rem` in CSS)

though the value of 640px is good too since it matches the threthhold value in tailwind

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -126,6 +126,17 @@ function TreeView<SCHEMA extends AnyTree>({
...
+  const [screenWidth, setScreenWidth] = React.useState(window.innerWidth);
+  React.useEffect(() => {
+    const handleResize = () => {
+      setScreenWidth(window.innerWidth);
+    };
+
+    window.addEventListener('resize', handleResize);
+    return () => {
+      window.removeEventListener('resize', handleResize);
+    };
```
maxxxxxdlp, line 138: can use the `listen` function to simplify this
search for usages

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -179,12 +190,13 @@ function TreeView<SCHEMA extends AnyTree>({
         />
         <Button.Icon
           aria-pressed={isSplit}
+          disabled={screenWidth <= 640}
```
maxxxxxdlp, line 193: create a variable `canSplit` and use that, rather than duplicating the magic number 3 times

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -179,12 +190,13 @@ function TreeView<SCHEMA extends AnyTree>({
...
           icon="template"
           title={treeText.splitView()}
           onClick={() => setIsSplit(!isSplit)}
         />
         <Button.Icon
-          disabled={!isSplit}
+          disabled={!isSplit || screenWidth <= 640}
```
maxxxxxdlp, line 199: rename `isSplit` to `rawIsSplit` and create variable `const isSplit = rawIsSplit && canSplit;` and use that variable in these places

maxxxxxdlp (changes requested):

grantfitzsimmons (approved):
:+1:

maxxxxxdlp (dismissed):
The changes you made look good, but I noticed one new problem (sorry for being late):

Your useEffect listen for screen width change, and in response to that the `canSplit` variable may get changed.

The problem is that the screen width may change a lot (imagine you are resizing the window), thus leading to the `screenWidth` state changing a lot, thus leading to a lot of re-renders, thus bad performance (or at least inefficient). And yet, it may be that despite all those changes to screenWith, the value of `canSplit` will still remain the same, thus all of those re-renders were useless.

So to optimize your code, instead of having `screenWidth` state, have a `canSplit` state, and set that state to true or false inside of useEffect in response to resize

Basically, the logic is the same, but by having `canSplit` be your state instead of `screenWidth`, you avoid all the useless re-renders of the component, as the component would only get updated on screen change only when `canSplit` changes

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -122,9 +122,8 @@ function TreeView<SCHEMA extends AnyTree>({
   const toolbarButtonRef = React.useRef<HTMLAnchorElement | null>(null);
   const [isEditingRanks, _, __, handleToggleEditingRanks] = useBooleanState();
 
-  const [screenWidth, setScreenWidth] = React.useState(window.innerWidth);
   const [rawIsSplit = false, setRawIsSplit] = useCachedState('tree', 'isSplit');
-  const canSplit = screenWidth >= 640;
+  const [canSplit, setCanSplit] = React.useState(window.innerWidth >= 640);
```
melton-jason, line 126: (optional) 

Can you extract `640` into its own variable? 
something like `const minimumSplitWidth = 640` and use that everywhere we use 640. 

Even better if you put it outside of the JSX React function at the top level of the module and use the standard global variable naming convention in Specify:
Something like
`const SPLIT_MIN_SCREEN_WIDTH = 640`

See below for an example:

https://github.com/specify/specify7/blob/b63632a7eb9d247b879023d7e19ca4fba07a567c/specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx#L11-L13

melton-jason (approved):

melton-jason (approved):

specifyweb/frontend/js_src/package.json
```diff
@@ -51,6 +51,7 @@
     "leaflet-gesture-handling": "^1.2.2",
     "leaflet.featuregroup.subgroup": "^1.0.2",
     "leaflet.markercluster": "^1.5.3",
+    "m-react-splitters": "^1.2.0",
```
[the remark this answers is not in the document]
maxpatiiuk: fyi, there is https://www.npmjs.com/package/react-resizable-panels
it has proper keyboard support
it's also used in react devtools (it's developed by the person who developed react devtools)
no rush to switch, but good to keep in mind

PR #3972, opened 2023-09-04
Implements minimal IIIF on the front-end

realVinayak:
I hosted Cantaloupe Image Server (IIIF compliant - https://cantaloupe-project.github.io/) that processes requests. 

The request is based IIIF specification - `{scheme}://{server}{/prefix}/{identifier}/{region}/{size}/{rotation}/{quality}.{format}`, and Cantaloupe returns the corresponding image (https://test.specifysystems.org:8090/iiif/2/21c44ff1-9843-4343-9445-587e3b20f4db.tif/0,0,2224,3000/278,/0/default.jpg). I configured it to connect to asset server on the test panel (it can be configured in multiple other ways). 

I then used a prebuilt IIIF viewer https://mejackreed.github.io/Leaflet-IIIF/examples/?url=https://test.specifysystems.org:8090/iiif/3/4814c409-e3c5-4177-9425-112480c83ca6.tif/info.json. 

Based on `attachmentlocation`, I fetch if the corresponding image has info.json (served by Cantaloupe), and which versions are present. Then, the options as following are shown: 
image
and then as 
image
when clicking on a version (rendered as IFrame)

This is a very minimal implementation, for upcoming meeting with Barcelona about IIIF - we don't know all the use cases, so this serves as a place to start.


realVinayak:
Relevant running internal documentation: https://docs.google.com/document/d/1zr_3_zLxRkot-9AQ1xlVDVN23iR8y940ucKZLUepfgY/edit?pli=1

specifyweb/frontend/js_src/lib/components/Attachments/Dialog.tsx
```diff
@@ -16,6 +16,7 @@ import type { SpecifyModel } from '../DataModel/specifyModel';
 import type { Attachment } from '../DataModel/types';
 import { SaveButton } from '../Forms/Save';
 import { Dialog } from '../Molecules/Dialog';
+import { getIIIFUrlFromVersion, IIIFViewer, useIIIFSpec } from './IIIFViewer';
```
maxxxxxdlp, line 19: can you please rename all usages to be `Iiif` instead of `IIIF` or `IiiF`

see https://medium.com/@thomasreggi/enforcing-strict-camelcase-5332c53a7484

tldr: if we always use strict camel case, then there is no ambiguity of whether an acronym should be capitalized or not, and there is consistency (you already used two different variations in this PR

specifyweb/frontend/js_src/lib/components/Attachments/IIIFViewer.tsx
```diff
@@ -0,0 +1,71 @@
...
+export const getIIIFUrlFromVersion = (
+  version: number,
+  location: string
+): string => `${BASE_IIIF_URL}iiif/${version}/${location}/info.json`;
+
+// eslint-disable-next-line @typescript-eslint/naming-convention
+async function validateIIIF(
+  attachmentLocation: string,
+  // eslint-disable-next-line @typescript-eslint/no-magic-numbers,@typescript-eslint/naming-convention
+  IIIF_supported = [1, 2, 3]
```
maxxxxxdlp, line 21: no underscores. use camel case

specifyweb/frontend/js_src/lib/components/Attachments/IIIFViewer.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+export function IIIFViewer({
+  baseUrl,
+  title,
+}: {
+  readonly baseUrl: string;
+  readonly title: string;
+}): JSX.Element {
+  return (
+    <iframe
```
maxxxxxdlp, line 62: security. see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#:~:text=to%20insecure%20origins.-,sandbox,-Controls%20the%20restrictions

specifyweb/frontend/js_src/lib/components/Attachments/IIIFViewer.tsx
```diff
@@ -0,0 +1,71 @@
...
+      src={`https://mejackreed.github.io/Leaflet-IIIF/examples/?url=${baseUrl}`}
+      style={{
+        width: '800px',
```
maxxxxxdlp, line 65: don't hardcode size?

specifyweb/frontend/js_src/lib/utils/ajax/definitions.ts
```diff
@@ -18,6 +18,7 @@ export const Http = {
   UNAVAILABLE: 503,
   GATEWAY_TIMEOUT: 504,
   INSUFFICIENT_STORAGE: 507,
+  NOT_IMPLEMENTED: 501,
```
maxxxxxdlp, line 21: can you sort these so that HTTPs codes are in ascending order?

maxpatiiuk (commented):
code logic looking good
some style issues, but not bad for quick demo code

specifyweb/frontend/js_src/lib/components/Attachments/Dialog.tsx
realVinayak: fixed

realVinayak:
What the current implementation does:
1. Implements a IIIF image server on top of the specify's asset server. So, the images saved in the asset server can be fetched as IIIF requests.
2. It assumes that IIIF image server has access to the file system of asset server (so, at least a mount). However, Cantaloupe IIIF server used serves many other configuration (including S3 buckets)
3. A separate port is assigned to IIIF server. I made nginx and docker-compose.yml changes, which are not included in this PR. But, if I'm remembering correctly, I just made the Cantaloupe open to host at a port. An alternative implementation could be using nginx (which will be needed to support SSL)

TODO:
1. Remove dependence on [external](https://github.com/specify/specify7/blob/2635ca4b8d09c06dbfa80d01ef389b0f81de2096/specifyweb/frontend/js_src/lib/components/Attachments/IiifViewer.tsx#L69) website for _displaying images and making network requests._ Recommended to use a library which will be more optimized than in-house implementation (fetching and displaying can be optimized using tiling, and there are edge cases to it, for example some TIF can be tiled, some not)

PR #3975, opened 2023-09-06
Fix localization tests failing

realVinayak:
Supposedly should have passed entire test suite, but jest test suites are still failing

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
@@ -26,7 +26,7 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
 
   private readonly rules: BusinessRuleDefs<AnySchema | SCHEMA> | undefined;
 
-  // eslint-disable-next-line functional/prefer-readonly-type 
+  // eslint-disable-next-line functional/prefer-readonly-type
   public pendingPromise: Promise<BusinessRuleResult | undefined> =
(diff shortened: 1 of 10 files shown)


realVinayak:
Can we merge this?

melton-jason:
I would prefer being more explicit about the rules being disabled. 
For instance, replacing the readonly rules with  `// eslint-disable-next-line functional/prefer-readonly-type`

melton-jason (approved):
See my comment https://github.com/specify/specify7/pull/3975#issuecomment-1708615367

acwhite211 (approved):

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -266,7 +266,7 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
 
     const invalidResponse: BusinessRuleResult<SCHEMA> = {
       valid: false,
-      /* eslint-disable-next-line */
+      /* eslint-disable-next-line array-func/prefer-includes */
       reason: fieldInfo.some((field) => field === undefined)
```
maxxxxxdlp, line 278: hmm, why was this ESLint rule disabled? seems like a valid suggestion

`fieldInfo.includes(undefined)`

maxxxxxdlp: ups, stale comment. this is what happens when you review one commit at a time

PR #4006, opened 2023-09-13
Display a count of 0 when the prep count has not been defined instead of NaN

### LINKED ISSUE #3881: Loan/Gift preparations without count display as NaN on preparation selecting dialog
**Describe the bug**
If a preparation doesn't have a count entered, it will show as NaN in the Preparations dialog for loans and gifts. It is also falsely linked. 

**To Reproduce**
1. Create a CO, add preparation but leave count blank
3. Create New Loan or Gift and enter Cat # for the CO record you made 
4. See error in Preparations selection dialog

**Expected behavior**
It should probably default to 1 in the available column and 0 for unavailable. Currently, you can add a preparation to the loan or gift without a count. Count isn't a required field, so this makes sense, but I'm not sure if we want that. 
**Screenshots**

https://github.com/specify/specify7/assets/135047322/6eceb988-134e-4db0-957e-0ea9e5fb138e

**Desktop:**
 - macOS, Chrome, v7.9-dev

**Database Name:** archireef

CarolineDenis:
Fixes #3881


maxxxxxdlp (changes requested):
hmm, I didn't look into the code too closely, but NaN is not expected result - it implies that there might be some wrong calculation going on somewhere (either division by 0, or arithmetic on non-numeric data)

please double check where NaN is coming from and correct the issue there

i.e before the calculation that could result in a NaN, add a check to guard against such case,  - if such case is detected, then set 0 instead of doing the calculation

tldr: the code change you made is correct, but it is better to prevent NaN from being created rather than changing how NaN is being displayed

small thing, but also prefer using Number.isNaN over isNaN (isNaN is legacy and has some weird behavior that has been fixed in Number.isNaN)

CarolineDenis:
> hmm, I didn't look into the code too closely, but NaN is not expected result - it implies that there might be some wrong calculation going on somewhere (either division by 0, or arithmetic on non-numeric data)
> 
> please double check where NaN is coming from and correct the issue there
> 
> i.e before the calculation that could result in a NaN, add a check to guard against such case, - if such case is detected, then set 0 instead of doing the calculation
> 
> tldr: the code change you made is correct, but it is better to prevent NaN from being created rather than changing how NaN is being displayed
> 
> small thing, but also prefer using Number.isNaN over isNaN (isNaN is legacy and has some weird behavior that has been fixed in Number.isNaN)

ok will do that. The NaN comes from the fact that the count has not been defined (the field is just empty). If the user sets it to 0 we would see 0.

maxxxxxdlp (changes requested):
Better but not quite there. I probably wasn't clear, so let me clarify - NaN should never be created - yes, back-end returns `null`, but front-end converting that to NaN is not right.

The bug occurs here:

Screenshot 2023-09-14 at 18 23 24

See how line 194 uses Number.parseInt? that results in a NaN, instead, it should use `f.parseInt(prepData[10] ?? undefined) ?? 0` like previous lines

---

for why I (and the JS community in general) are about trying to avoid NaNs as much as possible see any of the following:

- https://stackoverflow.com/questions/43580225/comparison-with-nan-doesnt-work-the-way-i-expect
- https://medium.com/codex/javascripts-nan-is-dumbass-661a3a8e8d1a
- https://ntgard.medium.com/why-nan-nan-3d41af988d30
- https://blog.webdevsimplified.com/2020-10/javascript-nan/
- https://levelup.gitconnected.com/dont-fall-into-the-nan-trap-d2fbc5536e50

maxxxxxdlp (approved):

bronwyncombs (approved):
working!

https://github.com/specify/specify7/assets/135047322/3f435f05-1c10-4b80-806b-ef463a593e44


https://github.com/specify/specify7/assets/135047322/7532ce1b-4dda-45a7-a94d-303e2f77cd51

Issue #4005, opened 2023-09-13
Counts are ambiguous in delete blockers and agent merging references


grantfitzsimmons:
**Describe the bug**
Counts shown in the deletion blocker dialog and agent merging reference section do not show what the references refer to.

**To Reproduce**
1. Go to an agent with many references (https://dbfordocskufish-v79-dev.test.specifysystems.org/specify/view/agent/1514/)
2. Try to delete the agent
3. See issue

**Screenshots**
<img width="484" alt="image" src="https://github.com/specify/specify7/assets/37256050/09b123e1-bbf3-4053-ac03-0eab4c2905de">

**Expected behavior**
Instead of showing **Collection Object** 4 times with indiscernible counts, it should show what the counts are for:
<img width="642" alt="image" src="https://github.com/specify/specify7/assets/37256050/df5439b9-4b12-4625-bea5-cf6d1fac79a5">

`v7.9-dev`, KU Fish database copy

realVinayak:
I'm not sure if that'd be clear to the user, or more confusing. I've implemented the described solution in https://github.com/specify/specify7/pull/4008. I still think it is unclear (and confusing) to the user. Although, the previous method in production was not much better, and just showed one relationship LOL. So, maybe this is fine.

PR #4008, opened 2023-09-14
Add related model label in delete blockers

realVinayak:
Fixes https://github.com/specify/specify7/issues/4005

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
@@ -9,7 +9,14 @@ import { commonText } from '../../localization/common';
 import { interactionsText } from '../../localization/interactions';
 import { queryText } from '../../localization/query';
 import { f } from '../../utils/functools';
-import { filterArray, type GetOrSet, type GetSet, type IR, type R, type RA } from '../../utils/types';
+import {
+  filterArray,
+  type GetOrSet,
+  type GetSet,
+  type IR,
+  type R,
+  type RA,
+} from '../../utils/types';
 import { removeKey } from '../../utils/utils';
 import { Container, H3 } from '../Atoms';
 import { Button } from '../Atoms/Button';
@@ -202,7 +209,7 @@ export function QueryResults(props: Props): JSX.Element {
         Array.isArray(results) &&
         Array.isArray(loadedResults) &&
         results.length > 0 &&
-        typeof fetchResults === 'function'? (
+        typeof fetchResults === 'function' ? (
           <>
             {hasPermission('/record/replace', 'update') &&
               hasTablePermission(model.name, 'update') && (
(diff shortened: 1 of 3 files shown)


realVinayak:
See https://github.com/specify/specify7/issues/4005#issuecomment-1718495574

realVinayak:
image

grantfitzsimmons:
@realVinayak Can these be separate just like Collection Object Attribute?

realVinayak:
Like it is in production? Based on code, @maxxxxxdlp made a deliberate decision to show these dependent resources as part of parent resource (so collection object instead of collection object attachment).

grantfitzsimmons:
No, dependent resources should be shown as children collapsed under the parent, but the table icon should reflect the related table (Cit) or (Att) rather than showing the full name printed out?

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -393,7 +400,8 @@ export function useFetchQueryResults({
   const resultsRef = React.useRef(results);
   const handleSetResults = React.useCallback(
     (results: RA<QueryResultRow | undefined> | undefined) => {
-      const filteredResults = results !== undefined ? filterArray(results) : undefined
+      const filteredResults =
+        results !== undefined ? filterArray(results) : undefined;
```
maxxxxxdlp, line 404: no negated conditions - swap branches instead
or even better, just use `f.maybe(results, filterArray);`

maxxxxxdlp (approved):
looks good
hope this file wasn't confusing to work with

(when I was writing this, I kept being confused between directRelationship.model/relatedModel and parentRelationship.model/relatedModel as well as relationship label VS table name)

realVinayak:
> hope this file wasn't confusing to work with

LOL very! Looking at the code, I knew it was gonna be convoluted logic given so many branches. Took me some solid time on understanding boundaries. I think the logic could be improved though, as the parent resources should be available via othersidename and relationship name, and looking for `isdependent` fields, while also including string parsing. Only reason I was able to make my way through this was that I worked on https://github.com/specify/specify7/pull/3752

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
realVinayak: the one time it was obvious to use f.maybe, and I didn't :)

realVinayak:
image
image

realVinayak:
To test:
Make sure deletion blockers appear correctly when tried to delete a resource with references. Do it for forms, and for agent merging dialog.

grantfitzsimmons (changes requested):
image

https://dbfordocskufish-v79-dev.test.specifysystems.org/specify/view/agent/1028/
https://dbfordocskufish-issue-4005.test.specifysystems.org/specify/view/agent/1028/

grantfitzsimmons (approved):
Wow! 😍 

image

Screenshot 2023-09-14 at 6 12 35 PM

image

maxxxxxdlp (approved):

melton-jason (approved):

CarolineDenis (approved):

bronwyncombs (approved):
👍 
Screenshot 2023-09-15 at 9 23 35 AM

PR #4016, opened 2023-09-17
fix(Input.Number): Allow for floats

maxpatiiuk:
Input.Number allowed entering floating point numbers, but didn't allow to save them (because they would get converted to integers) - this PR fixes that


CarolineDenis (approved):

melton-jason (approved):

maxxxxxdlp:
Looking at the usages of Input.Number, the impact of this bug should have been minimal given that `step="1"` is the default for Input.Number, thus resulting in validation error if non-numeric value is put in.

Forms would have been one place where this would have had an impact, but fortunately we were using Input.Generic in there for numeric fields so the bug didn't manifest itself

maxxxxxdlp:
After looking at usages of Input.Number again, I realized that the fix in this PR would have actually broken things.

Here is why:
- All the usages of Input.Number before were expecting an integer, thus this bug didn't manifest
- After this change, the usages would still be expecting an integer and would print out a validation error if non-integer is inputted HOWEVER, they won't prevent putting a float value at least temporary (before the change, even temporary putting float wasn't allowed). Which means, for example, in the `<Slider>` component, when user puts an index of a record to go to, they could put 3.5, which would break things. Other usages may be broken too.

Thus, I fixed it in a better way:
- Reverted the change in this PR
- Renamed Input.Number to Input.Integer and added a runtime check to ensure Input.Integer is used correctly when expecting an integer value
- Added Input.Float for the case in the future when we would want to receive a float from the user (again we receive floats from the user in only a few places so far, and those places use Input.Generic at the moment)

CarolineDenis (approved):

maxpatiiuk:
To test:
- test that numeric inputs in several different places accept only integers, not floating point values (i.e test record index field, and inputs in the preparation return dialog)

emenslin (changes requested):
I tried this in an integer field and was not able to save, it looks the exact same as previous versions. 
Screenshot 2024-01-03 113213

maxpatiiuk:
Sorry, I wasn't clear

Not being able to enter floating point number into an integer field is expected behavior, and wasn't changed by this pull request

Collecting Trip Attribute form has some files that allow floating points numbers - I tested this and was able to enter floating points in there, but only integers in other places

PR #4020, opened 2023-09-18
Unit tests for useTriggerState

CarolineDenis:


specifyweb/frontend/js_src/lib/hooks/__tests__/useTriggerState.test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  act(() => {
+    result.current[1](updatedValue);
+  });
+
+  expect(result.current[0]).toBe(updatedValue);
+});
+
+async function fakeAsyncFunction() {
+  await new Promise((resolve) => setTimeout(resolve, 100));
```
maxxxxxdlp, line 51: (optional) could create a util for this in ./lib/tests/utils.ts

```ts
export const sleep = async (time:number): Promise<void> =>
  new Promise((resolve) => setTimeout(resolve, time));
```

maxxxxxdlp (changes requested):
you very thoroughly tested the get state/set state feature

a few suggestions:

# 1.

Given that you are just learning and getting practice, being this thorough is good, but in the future, you can simplify a bit - mainly, your 2nd test is redundant as anything that could cause test 2 to fail, would likely also cause test 4 to fail, so you could merge several test cases together.

Being this thorough is helpful when:
- You are testing complex logic -because then if both test 2 and test 4 fails, you can look first at the simpler test 2 to try to fix that, and chances are it would also fix the test 4
- When the size of the individual test is large, and you want to split the test up to make it more readable
But in this case, the logic you are testing is simple, and the tests for it are small.

While in the ideal world, all tests would be thorough always, in practice there are at least 3 limitations:
- programmer's time to write tests (you want to be efficient and be able to write tests for more things, especially given how many tests we are lacking),
- the time to execute tests (each additional test means tests take longer to run, which means the programmer has to wait longer when running `npm test`). This is especially true for tests that take a lot of time to execute (generally, anything that involves rendering a react component, or making a mocked network request, or timeouts/async)
- tests, like any other code, need to be maintained and updated over time (so needless tests means more code maintenance overhead). And if the behavior of the code you are testing changes intentionally (i.e a new feature), having more tests means more tests to update.

None of what I said is a big issue in this case and you don't have to make any changes here, but just keep these tradeoffs and considerations in mind when writing future tests.

# 2.

Much more importantly, the hook you are testing does 2 things - returns result of React.useState, and changes the value of the state in response to default value change.

All of the tests you have here testing the result returned from React.useState. But, useState is a react hook - react has a TON of tests already, so you don't need to test React.useState. As a rule, people don't test the library's code as it's the library's responsibility to test the code (and your responsibility to pick a library that tests their code). You just need to test that the way you integrated with the library is correct - i.e make sure you didn't forget to pass the default value to the useState, and didn't forget to return state and setState from the hook - for testing that just a single test or two would have sufficed in this case.

The 2nd thing the hook does is change the state in response to defaultValue change - that is the very reason the hook exists (if not for that, it is equivalent to useState), and yet that isn't covered by any of your test cases.

The library docs have a nice example on how to update the arguments you passed to the hook - https://react-hooks-testing-library.com/usage/basic-hooks#props

Can you please add test cases that test that behavior?

specifyweb/frontend/js_src/lib/hooks/__tests__/useTriggerState.test.ts
maxxxxxdlp: but really this test case is not needed:
- it's somewhat redundant with the others (besides the async part, but that is testing React.useState, not our functionally)
- it adds 100ms to the execution time of the tests - bad, just 1ms should have sufficed, unless you have a good reason to do otherwise

maxxxxxdlp (approved):

CarolineDenis:
@melton-jason @maxxxxxdlp this will not be tested by UX-UI team right?

melton-jason (approved):
I'm glad you're writing more frontend tests! :partying_face: 
These tests look good!

And there is no need for UX testing here. 
Only other developers need to look through and approve/suggest changes to this PR.

PR #4021, opened 2023-09-18
Unit test for useBooleanState hook

CarolineDenis:


specifyweb/frontend/js_src/lib/hooks/__tests__/useBooleanState.test.ts
```diff
@@ -0,0 +1,167 @@
...
+  expect(result.current[0]).toBe(false);
+});
+
+test('Verify state does not change when setter is not called', () => {
+  const { result } = renderHook(() => useBooleanState(true));
+
+  expect(result.current[0]).toBe(true);
+});
+
+test.each([
```
maxxxxxdlp, line 16: I had no idea this was allowed!
that's so cool - reduces the need for `theories()`

maxxxxxdlp (approved):
The tests look good and very thorough!
though could you add at least 1 test case for making sure that when default value provided to useBooleanState changes, then the state value also changes to that value? (basically, making sure that useBooleanState integrates with useTriggerState correctly)

melton-jason (approved):
Very nice!

maxxxxxdlp (approved):
🎉

Issue #4017, opened 2023-09-17
Merging two agents failing


grantfitzsimmons:
https://fwri6123-v79-load-manual-ref.test.specifysystems.org/specify/overlay/merge/Agent/?records=1%2C2

https://github.com/specify/specify7/assets/37256050/467fceb2-d23e-4feb-ac3f-efd8bfcc8f3b

[Merging 9835f5fb-7de7-4dfd-b1bb-03974517d8ab Crash Report - 2023-09-17T18 10 04.309Z.txt](https://github.com/specify/specify7/files/12643406/Merging.9835f5fb-7de7-4dfd-b1bb-03974517d8ab.Crash.Report.-.2023-09-17T18.10.04.309Z.txt)

```
[...]

PR #4022, opened 2023-09-18
Hot fix for prevent deletion of asset server files

realVinayak:
There is a business rule that deletes files from the asset server. In case of agent merging, it deletes the files, but doesn't add it back - leading to loss of asset server files. 

The solution is to just set attachment location of the object to delete to null. This would prevent deletion from the asset server. If an error occurs, everything is rollbacked, so we don't care what we put in.

### DIFF
--- a/specifyweb/businessrules/attachment_rules.py
+++ b/specifyweb/businessrules/attachment_rules.py
@@ -42,7 +42,8 @@ def attachment_save(attachment):
 @orm_signal_handler('post_delete', 'Attachment')
 def attachment_deletion(attachment):
     from specifyweb.attachment_gw.views import delete_attachment_file
-    delete_attachment_file(attachment.attachmentlocation)
+    if attachment.attachmentlocation is not None:
+        delete_attachment_file(attachment.attachmentlocation)
 
 def get_attachee(jointable_inst):
     main_table_name = JOINTABLE_NAME_RE.match(jointable_inst.__class__.__name__).group(1)
--- a/specifyweb/specify/views.py
+++ b/specifyweb/specify/views.py
@@ -27,7 +27,6 @@ from specifyweb.celery_tasks import LogErrorsTask, app
 from . import api, models as spmodels
 from .build_models import orderings
 from .specify_jar import specify_jar
-
 from celery.utils.log import get_task_logger # type: ignore
 logger = get_task_logger(__name__)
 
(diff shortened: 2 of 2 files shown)


realVinayak:
Doing this, preserves the attachment location since the front-end sends data for dependents.

realVinayak:
To test:
1. Make an new agent (agent 001)
2. Make an new agent again (a bit later than first). Add an attachment to this agent.
3. Merge agent from 1 and 2. If attachment server is configured, the merge will be successful but opening attachment in agent 1 won't work (1 will be chosen as base because created earlier). In this branch, the attachment should appear.

realVinayak:
https://github.com/specify/specify7/issues/4017 should not occur. It won't try deleting the files now.

CarolineDenis (approved):

specifyweb/businessrules/attachment_rules.py
```diff
@@ -42,7 +42,8 @@ def attachment_save(attachment):
 @orm_signal_handler('post_delete', 'Attachment')
 def attachment_deletion(attachment):
     from specifyweb.attachment_gw.views import delete_attachment_file
-    delete_attachment_file(attachment.attachmentlocation)
+    if attachment.attachmentlocation is not None:
+        delete_attachment_file(attachment.attachmentlocation)
```
melton-jason, line 46: The only concern with these changes is if Specify was able to delete (or try and delete) an attachment file when its `attachmentlocation` is None. 
Of course, this does not make sense logically, but it is the only implication of this change to consider. 
I have gone through the code briskly and am pretty sure this could not be the case before. 

https://github.com/specify/specify7/blob/80b2cd818408672e0b0f61d19125f5bbc5ab7ce0/specifyweb/attachment_gw/views.py#L165-L176

Ultimately, it comes down to how the Asset Server handles null values in the body of the `filedelete` endpoint.

specifyweb/specify/views.py
```diff
@@ -550,6 +558,7 @@ def record_merge_fx(model_name: str, old_model_ids: List[int], new_model_id: int
                 if table_name in dependant_table_names:
                     # Note: need to handle case where deletion throws error because it is referenced my other records
                     try:
+                        clean_fields_pre_delete(obj)
                         obj.delete()
```
melton-jason, line 562: > Doing this, preserves the attachment location since the front-end sends data for dependents.

https://github.com/specify/specify7/pull/4022#issuecomment-1723921603

Would it be worth adding a note of this behavior here? 
While it is known to us, this might be initially confusing to future developers who stumble upon this code (assuming the code remains untouched). 
They might stumble upon this in the future and think it is a potential bug without making the connection that the frontend sends/overrides the data for dependent resources.

specifyweb/businessrules/attachment_rules.py
melton-jason: Actually, a None attachmentlocation would throw an error in Specify currently anyways, stating something along the lines of: "NoneType object has no attribute `encode`"

https://github.com/specify/specify7/blob/80b2cd818408672e0b0f61d19125f5bbc5ab7ce0/specifyweb/attachment_gw/views.py#L183-L185

(I'll leave this as a comment anyways, for documentation purposes)

melton-jason (approved):

specifyweb/businessrules/attachment_rules.py
realVinayak: Yep, that's why I used attachmentlocation!

realVinayak:
Closing this as changes are superseded by https://github.com/specify/specify7/pull/4029

PR #3613, opened 2023-06-12
Begin adding recursive CTE to tree queries

realVinayak:
In tree viewer, tree stats refer to the count of collection object associated with tree nodes sharing a parent ParentID. Backend currently does this by making a flat join "depth" number of times. So, starting from the parent, it will make "depth" joins. A problem with this approach is since children are many-to-one to parent, the rows in all the tables (except the last one) could get duplicated - hence, we need to have distinct determination / collectionobject for getting counts. This leads to a major performance problem. I've tested these queries, and they sometimes don't even finish in 30 minutes if tree gets too large or co/det table gets too big. Using a recursive-cte approach fixes this by making a join only when it is needed and queries finish in less than a second. 

For example, a query that looks like this using flat join : 

```
SELECT 
  taxon_1.`TaxonID` AS `taxon_1_TaxonID`, 
  count(
    DISTINCT determination_1.`DeterminationID`
  ) AS count_1, 
  count(
    DISTINCT determination_1.`DeterminationID`
  ) + count(
    DISTINCT determination_2.`DeterminationID`
  ) + count(
    DISTINCT determination_3.`DeterminationID`
  ) + count(
    DISTINCT determination_4.`DeterminationID`
  ) + count(
    DISTINCT determination_5.`DeterminationID`
  ) + count(
    DISTINCT determination_6.`DeterminationID`
  ) + count(
    DISTINCT determination_7.`DeterminationID`
  ) + count(
    DISTINCT determination_8.`DeterminationID`
  ) + count(
    DISTINCT determination_9.`DeterminationID`
  ) + count(
    DISTINCT determination_10.`DeterminationID`
  ) AS anon_1 
FROM 
  taxon AS taxon_1 
  LEFT OUTER JOIN taxon AS taxon_2 ON taxon_2.`ParentID` = taxon_1.`TaxonID` 
  LEFT OUTER JOIN taxon AS taxon_3 ON taxon_3.`ParentID` = taxon_2.`TaxonID` 
  LEFT OUTER JOIN taxon AS taxon_4 ON taxon_4.`ParentID` = taxon_3.`TaxonID` 
  LEFT OUTER JOIN taxon AS taxon_5 ON taxon_5.`ParentID` = taxon_4.`TaxonID` 
  LEFT OUTER JOIN taxon AS taxon_6 ON taxon_6.`ParentID` = taxon_5.`TaxonID` 
[...]

PR #4023, opened 2023-09-18
Add ability to show only collection applicable nodes in tree

### LINKED ISSUE #2051: Add the ability to only show tree nodes applicable to the selected collection
> Add the ability to only show tree nodes applicable to the selected collection

**Requested By:** **[CSIRO](https://app.asana.com/0/1201972579963460/1202844554923815/f)**

CarolineDenis:
Fixes #2051


specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -30,6 +30,7 @@ export function TreeRow({
   setFocusedRow,
   synonymColor,
   treeName,
+  showApplicableNodes,
```
maxxxxxdlp, line 33: the name `showApplicableNodes` is not very clear as to what it does.

how about `hideLeafNodes` or `hideEmptyNodes` or `hideChildlessNodes`?

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -141,7 +143,10 @@ export function TreeRow({
     []
   );
 
-  return (
+  const hasNoChildrenNodes =
+    nodeStats?.directCount === 0 && nodeStats.childCount === 0;
+
+  return showApplicableNodes && hasNoChildrenNodes ? null : (
```
maxxxxxdlp, line 149: nodeStats are loaded async, right?

can you double check that the nodeStats are always loaded by the time this condition is evaluated for child rows? i.e does it cause nodes to show a loading indicator while node stats are fetching, and then disappear from view once server responds with `.chldCount === 0`?

if that is the case, that is a rather bad UX - it would perhaps be better to gray out those nodes instead

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
maxxxxxdlp, line 1: Can you also modify tree search to exclude nodes without children?
Or is such information not available when the search widget is doing query? Or would that make search slower?
If so, ignore this

maxxxxxdlp (approved):
Overal good, but in the future I would discourage you from making changes to the same component at the same time in different PRs (this and split tree) - that causes merge conflicts, which are not fan to resolve

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
CarolineDenis: @maxxxxxdlp , indeed the nodeStats can be undefined at first when loading. I thought adding a condition that if nodeStat is undefined && hideEmptyNodes is true => add the class 'opacity-50 pointer-events-none'. The problem is that the nodes for which we never display the counts will always have a nodeStat undefined. What would be your go to in this case?

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -168,6 +173,7 @@ export function TreeRow({
                     ? 'outline outline-1 outline-blue-500'
                     : ''
                 }
+                ${hideEmptyNodes && isLoadingStats ? 'opacity-50' : ''}
```
maxxxxxdlp, line 176: this solution is good by me, if user-facing people approve. Can you ask Grant or others?

maxxxxxdlp (approved):

CarolineDenis:
@grantfitzsimmons , when you have time could you take a look at this PR on the test panel please?

bronwyncombs (dismissed):
With the checkbox selected, the tree still shows nodes that don't have associated objects.

https://github.com/specify/specify7/assets/135047322/a1dbd96d-1d31-4b50-ad28-d798bee603b6

[URL](https://fwri6123-issue-2051.test.specifysystems.org/specify/tree/taxon/?conformation=%7E%7E1%7E12643-%7E2671%7E2672%7E2673---%7E12679%7E12680--%7E12478%7E12454%7E12455%7E12456-%7E12460%7E12461%7E12462------%7E2%7E3%7E52070%7E6982%7E6983---%7E2657%7E12532%7E3375%7E3376--%7E12535%7E12583%7E12584%7E12585%7E12639-%7E12596%7E12597-%7E12674%7E12675---%7E12618%7E12619%7E12620---%7E12586%7E12587%7E12588%7E12589----%7E12634%7E12635%7E12666-%7E12636%7E12637---%7E12654%7E12655-%7E12660%7E12661--------%7E3378%7E48521%7E48522%7E48523---%7E52972%7E55476%7E3379--%7E52973%7E52974-%7E52977%7E52978-----------)

realVinayak:
@bronwyncombs they are different queries, that's why the result are different. the tree count in tree viewer check for preferred taxon. The query builder (the one you opened in new tab), checks for taxon.

realVinayak:
Can this also be tested on some bigger trees like herb_rbge?

realVinayak:
Based on https://github.com/specify/specify7/issues/2051, looks like collections want to hide phylums (which is a very high rank). A bit concerned about performance since the query will be more expensive for it. Would need to have performance testing for it. @grantfitzsimmons do you know any other institutions which requested this, on which we can test?

If it becomes a problem, 
1. Merge changes with https://github.com/specify/specify7/pull/3613/
2. Or make a faster query. In this case, this is possible. Because you don't care how many nodes a particular node, you can modify the stats end-point to use an `exists()` clause to the sql generated. This would be faster (likely a LOT) since MySQL is smart enough to do a semi-join https://dev.mysql.com/doc/refman/8.0/en/semijoins.html

grantfitzsimmons:
**Dev Discussion:** Make this run queries on the entire tree once Vinny's count query rework is implemented.

acwhite211 (approved):

grantfitzsimmons:
@CarolineDenis Can this be in 7.9.3?

grantfitzsimmons:
>2. In the trees, a button for quick filtering of only the nodes with any associated collection object (as indicated by the number in parentheses near the nodes).

**Requested By:** Eyal at The Hebrew University of Jerusalem

emenslin (approved):
Works as expected, looks good. 

https://github.com/specify/specify7/assets/150940840/ab434142-8832-4de6-8686-fd362fe46e9b

grantfitzsimmons (changes requested):
>Conclusion: Make this run queries on the entire tree once Vinny's count query rework is implemented.

We still haven't implemented this yet, right? In @emenslin's video, counts aren't showing for every node in the tree. I still see this as a requirement (as discussed in our group weeks ago).

grantfitzsimmons (approved):
After meeting with @CarolineDenis today, I feel the current implementation is satisfactory. While there may be a request in the future to run the count for all nodes, the behavior is understandable and resolving it is easy. I need to make a note in the documentation before release to clarify behavior to the users.

⚡ Nice work!! Thanks for talking with me about it @CarolineDenis 😄

PR #4025, opened 2023-09-18
Don't trigger save/unload protect when only clicking in input field

### LINKED ISSUE #3798: Clicking on a field in a query triggers user to save/unload protect
https://github.com/specify/specify7/assets/37256050/cc31d12c-ee20-45be-9c3d-597af56282e9

**Reported By:** @tlammer 

Very small issue

CarolineDenis:
Fixes #3798


specifyweb/frontend/js_src/lib/components/QueryBuilder/FieldFilter.tsx
```diff
@@ -161,7 +161,7 @@ export function QueryInputField({
             .map(({ parsed }) => (parsed as number | string).toString())
             .join(', ');
 
-      handleChange?.(parsed);
+      if (parsed) handleChange?.(parsed);
```
maxxxxxdlp, line 164: this seems like it introduces a bug
if I had a search query in the field before, and now I remove it, the search query would still be present

maxxxxxdlp (changes requested):
this is not quite right
it doesn't fix the bug (if I had a value in the field before, the click on the field would still wrongly trigger a saveRequired), and it now introduces a bug (can't clear the value from the field)

instead, you should be able to use `currentValue !== parsed` as your condition

but even better, would be to get rid of `saveRequired` state entirely, and instead compute dynamically whether changes have been made (store original query fields as a json in a ref, and on change to fields, convert new fields to json string, and compare that to original string). I believe the reason I didn't implement that is due to concern for performance if query contains a lot of fields (converting all the fields to json on every keystore could be bad)

but, I didn't test if performance would actually suffer, and there are ways around it (i.e using the throttle() util)

specifyweb/frontend/js_src/lib/components/QueryBuilder/FieldFilter.tsx
CarolineDenis: @maxxxxxdlp Do you mean: 
create stateFieldsJSONRef in FieldFilter.tsx 
like so: 
`  const stateFieldsJSONRef = React.useRef('');
  const stateFieldsJSON = JSON.stringify(state.fields);
  stateFieldsJSONRef.current = stateFieldsJSON;`
  
  And have a check in the useEffect: 
  `React.useEffect(() => {
    handleChange?.({
      fields: unParseQueryFields(state.baseTableName, state.fields),
      isDistinct: query.selectDistinct,
    });
    if (JSON.stringify(state.fields) !== stateFieldsJSONRef.current)
      saveRequired = true;
  }, [state, query.selectDistinct]);`
  ?

maxxxxxdlp: generally yes

except the assignment to ref should only happen once, not on every render.
and use `Json` instead of `JSON` when naming (to follow strict camel case)

<details>
<summary>Code (look at this to compare after you tried implementing the solution)</summary>

  In Wrapped.tsx we currently have this:
  ```tsx
    const [state, dispatch] = React.useReducer(reducer, pendingState);
    React.useEffect(() => {
      dispatch({
        type: 'ResetStateAction',
        state: buildInitialState(),
      });
    }, [buildInitialState]);
  ```

  I would replace it with this:

  ```tsx
    const [state, dispatch] = React.useReducer(reducer, pendingState);
    const initialFields = React.useRef<string>('');
    React.useEffect(() => {
      const initialState = buildInitialState();
      dispatch({
        type: 'ResetStateAction',
        state: initialState,
      });
      initialFields.current = JSON.stringify(initialState.fields);
    }, [buildInitialState]);
  
    const saveRequired = React.useMemo(()=>
      initialFields.current !== JSON.stringify(state.fields)
    ,[state.fields]);
  ```

  Then can remove saveRequired from the state.

  But, this could cause performance problems (that JSON.stringify would be executed on every keystroke when typing a query field value - which could be slow if there are a lot of query fields and computer is not powerful, thus it can be optimized further by using throttle

  So you can replace this part:
  
  ```tsx
      const saveRequired = React.useMemo(()=>
        initialFields.current !== JSON.stringify(state.fields)
      ,[state.fields]);
  ```
  
  with this part:
  
  ```tsx
    const [saveRequried, setSaveRequired] = React.useState(false);
    const checkForChanges =
      throttle(() => {
        setSaveRequired(
          initialFields.current !== JSON.stringify(state.fields)
        );
      }, 200);
    React.useEffect(checkForChanges, [state.fields]); // or can make the call to checkForChanges() be part of the existing useEffect that calls handleChange, like in your example
  ```
  
  (optionally, if you are interested, you can read about how to analyze page performance to be able to inspect what optimization is necessary and when - https://developer.chrome.com/docs/devtools/performance/)

</details>

CarolineDenis:
@maxxxxxdlp, it seems for me that this new code is more complicated to understand and to read than the only 3 lines we had before: 
BEFORE: 
`const isResourceModified = useIsModified(queryResource);
const saveRequired = isResourceModified || state.saveRequired;
if (currentValue !== parsed) handleChange?.(parsed);`
NOW
`const initialFields = React.useRef<string>('');`
`initialFields.current = JSON.stringify(initialState.fields);`
`const [saveRequired, setSaveRequired] = React.useState(false);`
`const checkForChanges = throttle(() => {`
 `const initialFieldsObject = JSON.parse(initialFields.current);`
  `const stateFieldsObject = JSON.parse(JSON.stringify(state.fields));`
  `setSaveRequired(`
     `JSON.stringify(initialFieldsObject) !== JSON.stringify(stateFieldsObject)`
  `);`
 `}, 200);`

 `React.useEffect(checkForChanges, [state.fields]);`
 What do you think?

maxxxxxdlp:
yes it may be more complicated - that is why it wasn't done before.

But:
-  it creates a better UX - if user makes a change and undoes it, the more complicated code will work correctly, but the simpler code won't
- it prevents bugs - a saveRequired flag that is switched on in parallel with some action is quite bug prone - as you have seen from the bug encountered here
  - another example is the forms - we constantly have bugs with forms falsely claiming there are unsaved changes. quite a few of such bugs are because we have a boolean `.changed` flag on a resource, rather than computing whether it's actually changed.

As far as how much code it adds, I can argue here. Consider all of the places in the reducers.ts where saveRequired is set to true or false - all of those can be removed. Then the places that initialize the state and that define the state type - you can remove saveRequired from those too - so at the end it might not add any more lines than it removed.

And of course there is performance to consider - as you have seen this can be done quite simply using useMemo, but for performance reasons, we probably should go with a slightly more complicated version.

Not forcing you to go either way, just saying there are trade-offs.

FYI: here is how to add code blocks in markdown: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks. That would also add syntax highlighting

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -149,8 +155,17 @@ export function QueryBuilder({
     'showHiddenFields'
   );
 
-  const isResourceModified = useIsModified(queryResource);
-  const saveRequired = isResourceModified || state.saveRequired;
+  const [saveRequired, setSaveRequired] = React.useState(false);
+  const checkForChanges = throttle(() => {
+    const initialFieldsObject = JSON.parse(initialFields.current);
```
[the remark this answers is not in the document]
CarolineDenis: because that will always return true otherwise

CarolineDenis:
> yes it may be more complicated - that is why it wasn't done before.
> 
> But:
> 
> * it creates a better UX - if user makes a change and undoes it, the more complicated code will work correctly, but the simpler code won't
> * it prevents bugs - a saveRequired flag that is switched on in parallel with some action is quite bug prone - as you have seen from the bug encountered here
>   
>   * another example is the forms - we constantly have bugs with forms falsely claiming there are unsaved changes. quite a few of such bugs are because we have a boolean `.changed` flag on a resource, rather than computing whether it's actually changed.
> 
> As far as how much code it adds, I can argue here. Consider all of the places in the reducers.ts where saveRequired is set to true or false - all of those can be removed. Then the places that initialize the state and that define the state type - you can remove saveRequired from those too - so at the end it might not add any more lines than it removed.
> 
> And of course there is performance to consider - as you have seen this can be done quite simply using useMemo, but for performance reasons, we probably should go with a slightly more complicated version.
> 
> Not forcing you to go either way, just saying there are trade-offs.
> 
> edit: your code snippet is too complicated - can be simplified - see comment below
> 
> FYI: here is how to add code blocks in markdown: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks. That would also add syntax highlighting

The issue I am facing is that when I load for the first time a saved query is enable the button 'Save Query'. Than disable if I click on a field then click outside. 
The button 'Save Query' is disable like it should on load when I put a break point on setSaveRequired in the checkForChanges(). The function runs before every is loaded. What would be an ideal solution?

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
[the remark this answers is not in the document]
maxxxxxdlp: then you have a bug somewhere as that's not supposed to be the case
if you are comparing different objects, then yeah, it would always be different.
but if you converted each to string before (using JSON.stringify() then it should compare correctly

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -149,8 +155,17 @@ export function QueryBuilder({
...
+    const stateFieldsObject = JSON.parse(JSON.stringify(state.fields));
+    setSaveRequired(
+      JSON.stringify(initialFieldsObject) !== JSON.stringify(stateFieldsObject)
+    );
+  }, 200);
+
+  React.useEffect(checkForChanges, [state.fields]);
+
```
maxxxxxdlp: can you move all of this code to bear near line 135?
it's because this code is related to `initialFields`, so it's best to keep related code closer to each other - easier to read the code that way

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -121,12 +121,18 @@ export function QueryBuilder({
...
   );
   const [state, dispatch] = React.useReducer(reducer, pendingState);
+
+  const initialFields = React.useRef<string>('');
+
   React.useEffect(() => {
+    const initialState = buildInitialState();
     dispatch({
       type: 'ResetStateAction',
       state: buildInitialState(),
```
maxxxxxdlp, line 130: don't call `buildInitialState()` here again  reuse the `initialState` variable you defined above

maxxxxxdlp:
Seems like the bug happens because we forgot to wrap `checkForChanges` function in `React.useMemo` - thus we are creating a new function on each render (and don't forget to provide proper dependency arrays)

Though, you may notice that the save button blinks when page loads - it shows as save required initially - this is because of pendingState (Which is the initial value for the state in query builder). Thus inside of checkForChanges, also add the `state !== pendingState` condition.

There might be a cleaner solution, but these two changes seem to fix it for me

CarolineDenis:
> Seems like the bug happens because we forgot to wrap `checkForChanges` function in `React.useMemo` - thus we are creating a new function on each render (and don't forget to provide proper dependency arrays)
> 
> Though, you may notice that the save button blinks when page loads - it shows as save required initially - this is because of pendingState (Which is the initial value for the state in query builder). Thus inside of checkForChanges, also add the `state !== pendingState` condition.
> 
> There might be a cleaner solution, but these two changes seem to fix it for me

:+1: works now

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -121,12 +120,34 @@ export function QueryBuilder({
...
+
+  const checkForChanges = React.useMemo(
+    () =>
+      throttle(() => {
+        if (state === pendingState) setSaveRequired(false);
+        else
+          setSaveRequired(
+            initialFields.current !== JSON.stringify(state.fields)
+          );
+      }, 200),
```
maxxxxxdlp, line 145:
```suggestion
      throttle(() =>
        setSaveRequired(
          state !== pendingState &&
          initialFields.current !== JSON.stringify(state.fields)
        )
      , 200),
```

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -310,7 +329,10 @@ export function QueryBuilder({
             saveRequired={saveRequired}
             state={state}
             unsetUnloadProtect={unsetUnloadProtect}
-            onSaved={(): void => dispatch({ type: 'SavedQueryAction' })}
+            onSaved={(): void => {
+              setSaveRequired(false);
```
maxxxxxdlp, line 333: need to add `initialFields.current = JSON.stringify(state.fields);` here

maxxxxxdlp (changes requested):

maxxxxxdlp (approved):

bronwyncombs (approved):
Working as expected! 


https://github.com/specify/specify7/assets/135047322/653cfb6b-bed8-4539-9045-d4dd28459f03

PR #4029, opened 2023-09-20
Drop dependents on clones, and fix ordering on the backend, and asset server fix

realVinayak:
Testing instructions coming soon.


realVinayak:
Testing instructions:

Attachments:
1. Make an agent with an attachment.
2. Make a second agent with or without attachment.
3. Merge first two agents together. All the attachments post upload should be valid, and be seen.

Agent Specialty
1. Make an agent. Add specialities to it.
2. Make a second agent, and add more specialties to it.
3. Merge two agents together. v7.9 should crash. This branch shouldn't.

Agent Addresses:
1. Merge agents with a lot of addresses together, and try choosing the addresses (and modifying them) from the agent merging dialog (will appear towards the bottom left of the dialog)

General testing of agent merging
In the preview of the record, try adding or modifying the fields present - since some of the code was refactored in this branch. The final record should have all the modifications done via preview.
Do the above for special fields such as addresses, and specialties too.

UX Testing
The changes recommended in the specify meeting have been made.
a. Use 'Linked Records' instead of 'References to this record' or 'Load References'
b. Make the 'Linked Records' on the left always at the top.
c. Make the spacing between the button to load references, and the preview buttons dynamic based on whether ANY reference was loaded. Additionally, if user loads references for any record, all the spaces would be expanded.  I added a bit of space even in the default one (nothing loaded) because it looked awkward with no spacing.

specifyweb/frontend/js_src/lib/components/Merging/autoMerge.ts
```diff
@@ -73,26 +74,60 @@ const sortResources = (
...
+            /*
+             * Don't preserve dependents if targetId is not defined. This will happen if autoMerge gets called recursively for -to-one
+             * resources, but those resources also have dependent resources.
+             * TODO: Handle this case better
+             */
+            targetId === undefined
+              ? undefined
+              : parentChildValues
+                  .find(([parentId]) => parentId === targetId)
+                  ?.at(1);
```
melton-jason, line 112: This value is being calculated for every stringified resource in the `uniqueDependentsCombined` Array, but it does not depend on `resource`. 

Can this be moved before `resourcesToReturn` is initialized, and thus before the `uniqueDependentsCombined.map((resource) => {...})`?

specifyweb/specify/views.py
```diff
@@ -464,6 +464,52 @@ def resolve_record_merge_response(start_function, silent=True):
...
+        except FieldDoesNotExistError:
+            continue
+        ordering_fields = ordering_tables.get(relationship.relatedModelName.lower(), None)
+        if (ordering_fields is None or
+                # Can this ever happen?
+                not isinstance(records, list)):
+            continue
+        order_fields_data = set([tuple([record.get(ordering_field, None)
+                              for ordering_field in ordering_fields])
+                                for record in records])
```
melton-jason, line 493: While probably not something we need to consider immediately, at least for Django we _do_ also include `rankid` in the list of ordering fields (if applicable) outside of the global `orderings` dictonary

https://github.com/specify/specify7/blob/e572df6fdcb365e5537f50b326c0ebc10035715c/specifyweb/specify/build_models.py#L56-L59

melton-jason (changes requested):
There are a lot of changes here to go through. 
These changes do make the code a little more complicated, so they will need to be thoroughly considered. 

However, the logic appears sound after a once-over. 
Some small things I noticed and other miscellaneous refactorings that could be done

specifyweb/frontend/js_src/lib/components/Merging/autoMerge.ts
realVinayak: Good catch!

specifyweb/frontend/js_src/lib/components/Merging/Usages.tsx
```diff
@@ -16,32 +16,56 @@ export function UsagesSection({
 }: {
   readonly resources: RA<SpecifyResource<AnySchema>>;
 }): JSX.Element {
+  const [blockerLoaded, setBlockerLoaded] = React.useState<boolean>(false);
+  const handleBlockersLoaded = React.useCallback(
+    () => setBlockerLoaded(true),
+    [setBlockerLoaded]
+  );
```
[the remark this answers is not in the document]
realVinayak: Hmm I could. But this is simple enough to not have that. I don't use useBooleanState unless I need toggle :)

specifyweb/specify/views.py
realVinayak: Thanks for catching this!

specifyweb/frontend/js_src/lib/components/Merging/Header.tsx
```diff
@@ -132,16 +132,18 @@ function SummaryLines({
 export function MergeRow({
   header,
   children,
+  className = '',
```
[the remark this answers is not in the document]
realVinayak: I think we do that a lot. Search `className?` in the codebase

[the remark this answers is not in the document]
melton-jason: Well, I guess we _do_ do this a lot :laughing: 
It occurs as an optional prop in 13 places and as a required prop (in functions which are exported) in 2 places. 

There are some benefits to this approach: it allows the consumer of the function to customize the behavior of the component it is rendering 
Although, I can imagine this can _also_ be a hindrance if used to _either_ an extensive or limited degree:
From a UX consistency point of view and code maintenance view, respectively.  

I would be more inclined to accept this if there were more of a use for the className prop outside of the one place, but this _also_ seems to be a common occurrence in the cases where className is an optional prop... 

Anyways, this change is fine!
This comment is more subjective than objective, and should not hold up merging this when necessary

melton-jason (approved):

CarolineDenis (approved):

specifyweb/frontend/js_src/lib/components/Merging/Header.tsx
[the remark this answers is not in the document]
realVinayak: What would be a better way?

specifyweb/frontend/js_src/lib/components/Merging/Usages.tsx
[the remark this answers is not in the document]
maxxxxxdlp: yes toggle callback is useful feature of useBooleanState
the other useful feature of useBooleanState is that when the default value you pass to it changes, the state is also changed (thanks to useTriggerState underneath)
but the reason I use useBooleanState most of the time is because it makes code more readable (in my opinion)
instead of having to do `onClick={(): void => setIsOpen(false) }`, I can just do `onClick={handleClose}`

though, useBooleanState would be slower than useState (as useBooleanState involves 3 useCallback calls, and one useLayoutEffect ) - though I never benchmarked so don't know if it makes a difference outside of extreme cases

still the benefits are not too strong, so I don't require people to use useBooleanState (but encourage when applicable)

specifyweb/frontend/js_src/lib/components/Merging/Header.tsx
```diff
@@ -132,16 +132,18 @@ function SummaryLines({
...
+  readonly className?: string;
   readonly children: React.ReactNode;
 }): JSX.Element {
   return (
     <tr>
       <th
         className={`
           sticky left-0 text-left ${mergeCellBackground()} z-[10]
-        `}
+        ${className}`}
```
maxxxxxdlp, line 146: can this be formatted nicely, pretty please? 🥺

specifyweb/frontend/js_src/lib/components/Merging/Usages.tsx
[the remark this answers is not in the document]
maxxxxxdlp: although, I am looking at this code again, and useBooleanState would totally make things simpler.

just think, you don't event need this:
```ts
  const [blockerLoaded, setBlockerLoaded] = useBooleanState();
	  const handleBlockersLoaded = React.useCallback(
	    () => setBlockerLoaded(),
	    [setBlockerLoaded]
	  );
```

because you can just do:
```ts
  const [blockerLoaded, setBlockerLoaded] = useBooleanState();
	  const handleBlockersLoaded = setBlockerLoaded;
```

and thus:

```ts
  const [blockerLoaded, handleBlockersLoaded] = useBooleanState();
```

tada 🎉

specifyweb/frontend/js_src/lib/components/Merging/Usages.tsx
```diff
@@ -16,32 +16,56 @@ export function UsagesSection({
...
+      [loadBlockers, resource, handleBlockersLoaded]
     ),
     false
   );
   return (
-    <td className="h-[theme(spacing.40)] flex-col !items-start overflow-auto">
+    <td
+      className={`${
+        blockersLoaded ? 'h-[theme(spacing.40)]' : 'h-[theme(spacing.14)]'
+      } flex-col !items-start overflow-auto`}
```
maxxxxxdlp, line 67: you can make this more readable by formatting it like this:

```suggestion
      className={`
        flex-col !items-start overflow-auto
        ${blockersLoaded ? 'h-[theme(spacing.40)]' : 'h-[theme(spacing.14)]'}
      `}
```

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -187,14 +187,36 @@ function Merging({
...
+  const [needUpdate, setNeedUpdate] = React.useState(false);
+
+  const rawSpecifyResources = React.useMemo(
+    () => records.map(deserializeResource),
+    [records, needUpdate]
+  );
+
+  const sortedResources = React.useMemo(
+    () =>
+      /*
```
maxxxxxdlp, line 203: this comment might need to be moved or updated.
you should either move it to be right above `(resource) => resource.get('timestampCreated')` and then add a separate comment for `(resource) => resource.get('specifyUser'),` (probably better way)
or modify it to also mention why `(resource) => resource.get('specifyUser'),` is there

sorting by agents that have user first (and if both have user, than user with lowest id first) because agents that have users would presumably have much longer history/more connected records. though you might want to double check that `resource.get('specifyUser')` does not emit warnings if 'specifyUser' fields is not present in the resource (if merging something other than agents in the future)

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/Merging/Usages.tsx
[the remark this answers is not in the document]
realVinayak: fixed

specifyweb/frontend/js_src/lib/components/Merging/Header.tsx
realVinayak: Fixed. Looked fine to me just fine before though ;)

carlosmbe:
So I tested this and functionality wise. Awesome work Biggy Vinny. Good S..Stuff. I have one question though. I hear there was a meeting discussing UI Changes.

Regarding ` Use 'Linked Records' instead of 'References to this record' or 'Load References' `

I believe the term `Linked Records` sounds more intimidating versus "`Show Linked Records`" or the previous "`Load References`". I can easily picture a situation where a collection manager new to Specify or freshly trained is afraid to click that button just because they don't know what it does and it sounds scary. That's definitely how I felt before reading the instructions. 

The other UI Changes look fine though

CarolineDenis:
> So I tested this and functionality wise. Awesome work Biggy Vinny. Good S..Stuff. I have one question though. I hear there was a meeting discussing UI Changes.
> 
> Regarding `Use 'Linked Records' instead of 'References to this record' or 'Load References'`
> 
> I believe the term `Linked Records` sounds more intimidating versus "`Show Linked Records`" or the previous "`Load References`". I can easily picture a situation where a collection manager new to Specify or freshly trained is afraid to click that button just because they don't know what it does and it sounds scary. That's definitely how I felt before reading the instructions.
> 
> The other UI Changes look fine though

Agree with you, `Show Linked Records` is more descriptive to what the button does.

realVinayak:
Agreed, but it was supposed to be more intimidating. Users should be scared of doing agent merging, and for good reason - it is not easily recoverable.

PR #4031, opened 2023-09-20
Merge new record attributes from the POST request

melton-jason:
The record merging endpoint requires the attributes of the "canonical" final record to be passed in the body of the POST request. 
After the references have been updated in the database (and the old duplicate records deleted), these attributes would be used to populate the record in a PUT request through the main Specify api. 

However, an issue arises if the value of an attribute or field references an agent who was deleted by the merging process. 
This PR handles that case by doing a "merge" process on the attribute data before it is passed to the Specify API. 
Any references to models who were deleted in the merged should now be replaced with the id of the canonical record. 
All other references should be unaffected. 

(Before this is merged, I would like to clean up the code a little and provide comments. As well as write a unit test)

**_TESTING INSTRUCTIONS_**:
In the Agent Merging dialog window, change the `createdbyagent` or `modifiedbyagent` field to be an agent who is involved in the current merging process. 

Video demonstrating the process:

[Screencast from 09-20-2023 04:41:13 PM.webm](https://github.com/specify/specify7/assets/64045831/16c999f7-e3e6-4fd4-8871-48b9ee631af5)

I have prepared a query on the `herbrbge9323` database on the test panel which returns two agent who would throw an error in `v7.9-dev`, but should merge successfully on this branch. 

https://herbrbge9323-v79-merge-record-data.test.specifysystems.org/specify/query/380/

### DIFF
--- a/specifyweb/specify/views.py
+++ b/specifyweb/specify/views.py
@@ -509,6 +509,32 @@ def fix_orderings(base_model: Table, new_record_data):
                     resources.insert(0, record)
             new_record_data[field_name] = resources
 
+def fix_record_data(new_record_data, model_name, old_model_ids, new_model_id):
+    final = {}
+    for key, value in new_record_data.items():
+
+        if isinstance(value, str) and model_name.lower() in value:
+            if "=" in value:
+                url = value.split("=")
+                if int(url[-1]) in old_model_ids:
+                    url[-1] = str(new_model_id)
+                final[key] = "=".join(url)
+                continue
+
+            url = value.split("/")
+            if int(url[-2]) in old_model_ids:
+                    url[-2] = str(new_model_id)
+            final[key] = "/".join(url)
+
+        elif isinstance(value, list):
+            final[key] = [fix_record_data(dependent_data, model_name, old_model_ids, new_model_id) for dependent_data in value]
+        else:
+            final[key] = value
+
+    return final
+
+
+
 
 @transaction.atomic
 def record_merge_fx(model_name: str, old_model_ids: List[int], new_model_id: int,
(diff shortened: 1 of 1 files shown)


maxxxxxdlp (approved):
so many edge cases to consider when merging 🫨 much more complicated then I though at the very begginning

specifyweb/specify/views.py
```diff
@@ -509,6 +509,32 @@ def fix_orderings(base_model: Table, new_record_data):
...
+    final = {}
+    for key, value in new_record_data.items():
+
+        if isinstance(value, str) and model_name.lower() in value:
+            if "=" in value:
+                url = value.split("=")
+                if int(url[-1]) in old_model_ids:
+                    url[-1] = str(new_model_id)
+                final[key] = "=".join(url)
+                continue
```
realVinayak, line 522: This isn't needed. Backend ignores to-many when putting.

Relevant lines https://github.com/specify/specify7/blob/038b7e52c6e72ad27850fa068021723dd1630208/specifyweb/specify/api.py#L552-L557

specifyweb/specify/views.py
```diff
@@ -509,6 +509,32 @@ def fix_orderings(base_model: Table, new_record_data):
                     resources.insert(0, record)
             new_record_data[field_name] = resources
 
+def fix_record_data(new_record_data, model_name, old_model_ids, new_model_id):
+    final = {}
+    for key, value in new_record_data.items():
+
+        if isinstance(value, str) and model_name.lower() in value:
```
realVinayak, line 516: wouldn't this break if value is a random text field and someone put in `agent` text?

realVinayak (changes requested):

realVinayak:
Can you see similar changes done in https://github.com/specify/specify7/pull/4036

PR #4033, opened 2023-09-21
Fix specifyuser not being taken as main agent

realVinayak:
>       Array.from(rawSpecifyResources).sort(
>         multiSortFunction(
>           (resource) => resource.get('specifyUser'),
>           true,
>           (resource) => resource.get('timestampCreated')
>         )
>       )

The problem in above code is that `null < 'aaa'`  returns false.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Merging/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Merging/index.tsx
@@ -206,7 +206,7 @@ function Merging({
        */
       Array.from(rawSpecifyResources).sort(
         multiSortFunction(
-          (resource) => resource.get('specifyUser'),
+          (resource) => resource.get('specifyUser') ?? '',
           true,
           (resource) => resource.get('timestampCreated')
         )


realVinayak:
To test:
1. Find two agents such that one agent was created before, but doesn't have a specify user. And the second agent was created after, but has a specify user. An example of this is `Martin Pullan` in herb-rbge

maxxxxxdlp (approved):

maxxxxxdlp (changes requested):
It works correctly it seems:

```js
[null,'a'].sort(multiSortFunction((x)=>x,true))
// ["a", null]
```

the resource that has an agent should come first. if all resources don't have agents, oldest resource should come first (which is the resource with lowest value of timestampCreated)

maxxxxxdlp:
I mean, this PR doesn't break it, but I don't think it was broken before either

maxxxxxdlp:
I would advise to extract that logic into a function and add tests for that function

maxxxxxdlp (approved):
ups, saw your Slack message.

I see now, it doesn't always work correctly:
```
[null,'a',null].sort(multiSortFunction((x)=>x,true))
// [null,'a',null]
```

PR #4059, opened 2023-10-02
Create a ref for since argument adding condition

CarolineDenis:


specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
```diff
@@ -10,7 +10,7 @@ import type { GenericNotification } from './NotificationRenderers';
 
 const INITIAL_INTERVAL = 5000;
 const INTERVAL_MULTIPLIER = 1.1;
-let addSinceToNotification = true;
+// let addSinceToNotification = true;
```
CarolineDenis, line 13: from @maxxxxxdlp 
this is a wrong location. it should be as a ref inside of useNotificationsFetch()
the notifications state is initialized to undefined when the hook is first called - so this boolean flag should be initialized at the same time

the Notifications component may be re-rendered completely (maybe when user changes the menu position)

CarolineDenis: line 85:  lastFetchedTimestamp = startFetchTimestamp;
this should probably also be stored in a ref - right now lastFetchedTimestamp value is lost on useEffect re-executions (happens every time the user opens/closes the notifications component)

CarolineDenis:
NOTES: this doesn't work. Work in progress

CarolineDenis:
@maxxxxxdlp could you give me more explanations for the use of ref you would have like to see for the `lastFetchedTimeStamp` and for `addSinceToNotification ` please?

specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
```diff
@@ -29,6 +29,9 @@ export function useNotificationsFetch({
   const [notifications, setNotifications] = React.useState<
     RA<GenericNotification> | undefined
   >(undefined);
+
+  const addSinceToNotificationRef = React.useRef(true);
```
maxxxxxdlp, line 33: don't you need to set it to false by default?
i.e don't add "since" on first fetch so that it fetched all?

specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
```diff
@@ -71,7 +74,7 @@ export function useNotificationsFetch({
           })
         )
         .then(({ data: newNotifications }) => {
-          addSinceToNotification = false;
+          addSinceToNotificationRef.current = false;
```
maxxxxxdlp, line 77: shouldn't this be "true" instead? i.e add "since" after first fetch?

maxxxxxdlp (changes requested):
the variable name seems to cause some confusion. Consider renaming to "isFirstFetch" to be less confusing.

maxxxxxdlp:
See longer explanation in https://github.com/specify/specify7/pull/3346#discussion_r1343355895 and https://github.com/specify/specify7/pull/3346#discussion_r1343350724

specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
```diff
@@ -29,6 +29,9 @@ export function useNotificationsFetch({
   const [notifications, setNotifications] = React.useState<
     RA<GenericNotification> | undefined
   >(undefined);
+
+  const addSinceToNotificationRef = React.useRef(true);
+
   React.useEffect(() => {
     let pullInterval = INITIAL_INTERVAL;
     let lastFetchedTimestamp: Date | undefined;
```
maxxxxxdlp, line 37: replace this with `const lastFetchDateRef = React.useRef<undefined | Date>(undefined);` and put it outside the useEffect.

and actually, then you don't need addSinceToNotificationRef at all, right? if lastFetchDateRef.current is undefined, then you know that you are in a first fetch, and so "since" needs to be added to the URL
will simplify your code

specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
```diff
@@ -29,17 +28,20 @@ export function useNotificationsFetch({
...
     const doFetch = (since = new Date()): void => {
       const startFetchTimestamp = new Date();
 
-      const url = addSinceToNotification
-        ? getSinceUrl(`/notifications/messages/`, since)
-        : `/notifications/messages/`;
+      const url =
+        lastFetchDateRef.current === undefined
+          ? `/notifications/messages/`
+          : getSinceUrl(`/notifications/messages/`, since);
```
maxxxxxdlp, line 44: sorry, I didn't look closer at getSinceUrl before.
it doesn't make sense for `getSinceUrl` function to receive the baseUrl argument - the base URL should always stay the same.
it should only receive the `since` argument

---

And change the type of `since` argument to `Date | undefined`, so that you can do something like this in here:
```tsx
const url = getSinceUrl(lastFetchDateRef.current === undefined ? undefined : since)
```

benefit - no duplication of URL

maxxxxxdlp (approved):
👍

maxxxxxdlp (approved):
💯

grantfitzsimmons:
@CarolineDenis Can you provide testing instructions for @specify/ux-testing ?

CarolineDenis:
For Testing: 

- notifications are fetched
- new notifications are correctly appended to the existing results and sorted
- can delete notifications

acwhite211 (approved):

PR #4070, opened 2023-10-05
Remove console.log for future dates not supported

CarolineDenis:
Fixes #4051

Issue #4051, opened 2023-09-29
`Future dates are not supported` being reported repeadedly 


grantfitzsimmons:
!!! Not about the underlying issue causing this warning, that should be solved in https://github.com/specify/specify7/issues/4050

When viewing the notifications dialog, there is a seemingly never ending list of notifications saying "Future dates are not supported". When viewing the console, the browser slows down dramatically.

https://github.com/specify/specify7/assets/37256050/a337744d-e51e-48ab-b05d-85078b31a58b

https://fwri6123-edge.test.specifysystems.org/specify/

`production`, macOS, Firefox

https://github.com/specify/specify7/assets/37256050/bb15e584-748e-4c59-a85f-923b79e6b6bd

maxxxxxdlp:
Yeah, the logging of those should be disabled (that is in https://github.com/specify/specify7/issues/641#issuecomment-1326964987)

When I added those (that was 2 or more years ago), I assumed https://github.com/specify/specify7/issues/641 would get fixed soon

The other thing we can do (and modern logging systems do), is that they don't report the same warning more than once (that could be as simple as having a boolean flag, and flipping it after first time the warning was printed, and not printing the warning again (that is what react does) - or if we need something fancier, there are logging libraries that do that (But probably overkill)

CarolineDenis:
we want to take of the console.log here for now? 
https://github.com/specify/specify7/blob/4d007df018f61b003128d74bcdbf808083e35886/specifyweb/frontend/js_src/lib/components/internationalization.tsx#L177-L188

@maxxxxxdlp @grantfitzsimmons

maxxxxxdlp:
Yeah, let's just remove it.

Hopefully, https://github.com/specify/specify7/issues/641 can be fixed someday

PR #4071, opened 2023-10-05
Fix infinite logs of future dates not supported

### LINKED ISSUE #4060: Going to notifications triggers lots of console errors in edge
To reproduce:
1. Go to notifications (do any action which will create it)
2. See the console gets flooded with console errors (Future dates are not supported)

Likely infinite re-render issue.

Not in 7.9, so introduced recently.

Needs to be fixed before v7.9.1 gets released.

image

CarolineDenis:
Fixes #4060


specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -34,8 +34,9 @@ export function DateElement({
     function updateRelativeDate() {
       if (date) {
         const now = new Date();
-        const timeDifference =
-          (now.getTime() - dateObject.getTime()) / MILLISECONDS;
+        const timeDifference = Math.abs(
```
maxxxxxdlp, line 37: uh, that's a tricky one. good job debugging this

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/Atoms/Internationalization.ts
```diff
@@ -153,8 +153,7 @@ export function getRelativeDate(date: Readonly<Date>): string {
      * create a data set and see its date of creation be 5 hours into the
      * future
      */
-    // Throw new Error('Future dates are not supported');
-    console.error('Future dates are not supported');
+    // Throw new Error('Future dates are not supported'), removed console.log in issue #4051;
```
melton-jason, line 156: (very optional)

```suggestion
    // Throw new Error('Future dates are not supported'), removed console.log in #4071;
```

or if #4070 is being used as the PR for fixing #4051:

```suggestion
    // Throw new Error('Future dates are not supported'), removed console.log in #4051;
```

It may be **_very slightly_** more helpful for future developers to have a link to the Pull Request over the issue. The Pull Request generally should include more information which may be immediately more helpful such as code changes, what was causing the problem, testing instructions, potential problems with code changes, etc. 
It _may_ save future developers a click if they do not need the issue for context (which is arguably rare)

grantfitzsimmons (approved):
No more console errors. All seems to be working well from a user point of view.

melton-jason (approved):
Besides the fact that this PR also fixes #4060, what is the difference between this PR and #4070?

I see #4070 has `production` as the base branch, while this has `v7.9.1-prerelease` as the base branch. 
Was the intention to merge them separately so production has the changes before  `v7.9.1-prerelease` is out, and/or is it to assess and test the impact of 2e2caa8 in a specific testing/developing environment before the changes are merged into production?

PR #4074, opened 2023-10-06
Fix infinite loop when entering index to browse in form

### LINKED ISSUE #4065: Query builder Browse in Forms results numbering inconsistent
**Describe the bug**
The record numbering in the results from a query builder query when viewed in Browse In Forms is inconsistent

**To Reproduce**
1. Use the "CE observation records" query in the voucher collection of the KU_Fish_Tissue database when logged in as abentley
2. Run the query and then select "browse in Forms" from the header menu bar
3. Try and navigate to record 130 by typing in the number in the numbering toggle in the top right.  You will get a never-ending spinning orange wheel and never be directed to the correct record.
4. Re-run the query and now use the slider to bring up more records than are on the first page of the paginated result set.  Then select Browse in Forms and you will be directed to the following record:

image

However, this is not the 130th record

6. If you use the arrow keys to navigate one by one to the 130th record you get this record:

image

**Expected behavior**
Numbering should be consistent no matter which method is used to navigate to through the records while selecting a number in the top right should work without having to scroll.

**Desktop:**
 - OS: Windows 10
 - Browser: Chrome
 - Specify 7 Version: 7.9

**Database Name:** KU_Fish_Tissue

**Reported By**
KUBI

CarolineDenis:
Fixes #4065


maxxxxxdlp:
> this fixes it but I cannot understand the underlying issue

@CarolineDenis can you use `git blame`/`git log` or GitHub UI to find out who added that filterArray call that this PR is removing and why it was added?
Would be surprising/interesting

CarolineDenis:
> > this fixes it but I cannot understand the underlying issue
> 
> @CarolineDenis can you use `git blame`/`git log` or GitHub UI to find out who added that filterArray call that this PR is removing and why it was added? Would be surprising/interesting

oh dear.. :sweat_smile:

specifyweb/frontend/js_src/lib/components/QueryBuilder/ToForms.tsx
```diff
@@ -34,10 +35,16 @@ export function QueryToForms({
       ? (results[index]![queryIdField] as number)
       : Array.from(selectedRows)[index];
 
+  const filteredResults = f.maybe(results, filterArray);
```
maxxxxxdlp, line 38: 1. you don't need the actual results, you just want to know if any results are not undefined, right? in that case, think about how you could use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some in here to improve performance and maybe even simplify code a bit
2. you could use useMemo

maxxxxxdlp (changes requested):

maxxxxxdlp:
@specify/ux-testing When testing this, besides https://github.com/specify/specify7/issues/4065, please test that the fix for #3829 still works correctly

---

Also, I am not sure if mentioning the team pings everyone in that team or no one. If it's everyone, sorry about that, I will stop doing it.

specifyweb/frontend/js_src/lib/components/QueryBuilder/ToForms.tsx
```diff
@@ -34,10 +34,19 @@ export function QueryToForms({
       ? (results[index]![queryIdField] as number)
       : Array.from(selectedRows)[index];
 
+  const containUndefinedResults = React.useMemo(
+    () => results.includes(undefined),
+    [results]
+  );
+
```
maxxxxxdlp, line 41: the code looks good (and good job using .includes - even better than .some if that works for you), but I am confused on what you are trying to achieve here.

the code in present form disables the "Browse in Forms" if any result is not fetched yet - is that what you are trying to do?
i.e if query has 1000 results, the "Browse in Forms" will remain disabled until the user scrolls to the very bottom of the query results table (and wait for it to load) - only then would "containUndefinedResults" become false - surely this is not what you are going for

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/ToForms.tsx
CarolineDenis: no only when the first 80 are loaded

CarolineDenis: and it allows to not have the browse in form when there is no results

maxxxxxdlp: see me break it:

1. go to the last result
2. the "Browse in Forms" becomes disabled


https://github.com/specify/specify7/assets/40512816/e4286034-4574-48aa-acca-9dc1650003bd

maxxxxxdlp: When query returned to results, the `results` array looks like this:

Screenshot 2023-10-11 at 19 23 02

take a look at why `undefined` is there, and make it not be there (and check if that change is safe to do - i.e double check it doesn't break anything else) - then there won't be need for `containsUndefined`

maxxxxxdlp (changes requested):
see https://github.com/specify/specify7/pull/4074#discussion_r1355970754

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -40,17 +24,18 @@ import {
...
 import { sortTypes } from './helpers';
+import { useFetchQueryResults } from './hooks';
 import { QueryResultsTable } from './ResultsTable';
 import { QueryToForms } from './ToForms';
 import { QueryToMap } from './ToMap';
 
 export type QueryResultRow = RA<number | string | null>;
 
-type Props = {
+export type Props = {
```
maxxxxxdlp, line 38: if you make something globally available, make the name more verbose (`QueryResultsProps`) to prevent name collisions (confuses auto-importer in the editor)

specifyweb/frontend/js_src/lib/components/QueryBuilder/ToForms.tsx
```diff
@@ -34,18 +34,11 @@ export function QueryToForms({
...
-  );
-
   return (
     <>
       <Button.Small
         disabled={
-          results.length === 0 ||
-          totalCount === undefined ||
-          containUndefinedResults
+          results.length === 0 || totalCount === undefined || totalCount === 0
```
maxxxxxdlp, line 41: `results.length === 0` and `totalCount === 0` seems redundant
is having just `totalCount === 0` sufficient?

maxxxxxdlp (approved):

pashiav (approved):
Searching by typing in record number results in:
issue-4065-search
Navigating with arrow key to record:
issue-4065-arrow-keys

Results in the same record. Looks good!

bronwyncombs (approved):
Browse in forms record numbering is consistent when navigating with arrows and by entering desired number. Tested with original case of CT partial date query on `KU_Fish` and on `herb_rbge`.

Fix for rendering of query results when fields are hidden is still working as well:

Screenshot 2023-11-06 at 2 18 21 PM

specifysoftware:

Issue #2395, opened 2022-11-03
WorkBench validation can require create access


grantfitzsimmons:
> **Workbench - validation of new field on a table you don't have "Create" access to**
>
> We plan to use the transfer option for workbench uploads so that "Data Entry" users can create the upload, and then pass it to a "Curation Manager" to approve and actually do the upload. Our Data Entry role does not have the "Create" ability for the taxon tree - so when they try to validate they get an error (see attached).
> Is this a bug, or is there a roles/permissions setting that would fix this issue?
> 
> ![image](https://user-images.githubusercontent.com/37256050/199795004-e17e6324-a1f7-472a-b39f-00e88ccfab4e.png)
> 
> Attempting to add a value to a pick list:
> 
> ![image](https://user-images.githubusercontent.com/37256050/199795029-575b9031-8e30-4ff2-9301-1efb91c33647.png)

This should not be required. Validation should not require create access.

**Reported By:** CSIRO ([Asana link](https://app.asana.com/0/1201972579963460/1203290426639052/f))

grantfitzsimmons:
Corinna at CSIRO has reiterated the importance of this feature for their workflow. They have requested that we increase this item in the priorities internally.

>We are closely working with the two collections that have gone live late last year and are 'pushing' the use of the Workbench for data uploads. Quite a number of the staff members who will prepare the csv files to be uploaded will have lower-level roles, so this validation error is turning into a bit of a problem for our internal workflows.

@acwhite211 Do you think we can make this change for them sometime soon?

grantfitzsimmons:
This is the most irritating issue of the three– we need to prioritize this over the others.

grantfitzsimmons:
At least make it not error!

grantfitzsimmons:
@acwhite211 Can you fix this before 7.9.4?

realVinayak:
Idea: Skip all table permissions if validating. Since rows are rollbacked, the data entered by a lower permission user would be rollbacked.

PR #4090, opened 2023-10-08
Skip create permission check if validating

realVinayak:
Allows users with just validate permission to validate a dataset without create permission to tables.  

See https://github.com/specify/specify7/issues/2395#issuecomment-1749720717


grantfitzsimmons (approved):
Tested as `wbuser` on a copy of RBGE's database. Looks like all is working as expected! Very exciting development. 

The role configured for this user is read-only for _everything_, with additional abilities granted for creating queries, record sets, and WorkBench datasets.  No write permission for any table, and this is blocked as one would expect.

https://herbrbge9323-issue-2395.test.specifysystems.org/

[Video here demonstrating it](https://drive.google.com/file/d/1FabYBJTjWS1zc5zBccFgaMFnr-O0N-Sd/view?usp=sharing)

realVinayak:
Testing instructions

1. Use a specify user without exclusive right access to tables, but with the permission to validate a dataset.
2. Use a prior dataset or make a new one which will make new records in the database (important). For example, add a random species, or a collectionobject with a high catalog number. Make sure that records in this dataset don't already exist, so that workbench will create records.
3. This branch should allow validating, even though create permission to tables isn't given. Production should throw an error saying NoMatchingRuleException(.... create)

maxxxxxdlp (approved):
Thanks for fixing this!

specifyweb/workbench/upload/auditor.py
```diff
@@ -12,11 +12,12 @@
...
+    skip_create_permission_check: bool = False
     def insert(self, inserted_obj: Any, agent: Union[int, Any], parent_record: Optional[Any]) -> None:
+
         if agent is None:
             logger.warn('WB inserting %s with no createdbyagent. Skipping permissions check.', inserted_obj)
-        else:
+        elif not self.skip_create_permission_check:
             if isinstance(agent, int):
                 # TODO: Optimize this potential w/ just memoization
                 agent_obj = Agent.objects.get(id=agent)
```
melton-jason, line 23: (optional) 
While we're here, we _could_ do some refactoring of this code. 
I would extract this validation logic into it's own function. 
Perhaps something like 
```py
def validate(self, agent: Union[int, Any], inserted_obj: Any) -> None:
        if agent is None:
            logger.warn('WB inserting %s with no createdbyagent. Skipping permissions check.', inserted_obj)
            return
        
        if self.skip_create_permission_check: return

        if isinstance(agent, int):
            # TODO: Optimize this potential w/ just memoization
            agent_obj = Agent.objects.get(id=agent)
            check_table_permissions(self.collection, agent_obj, inserted_obj, "create")
```

> this code is functionally identical to how it was before. Not sure why the `check_table_permissions(...)` was outside of the if statement which defines `agent_obj` before. If `agent` is not an int, pretty sure it would throw an error stating `agent_obj` was used before it was initialized. If this is intentional, we should throw a more explicit and descriptive error. 

and then `insert(...)` can be simplified to 

```py
def insert(self, inserted_obj: Any, agent: Union[int, Any], parent_record: Optional[Any]) -> None:
        self.validate(agent, inserted_obj)

        if self.audit_log is not None:
            self.audit_log.insert(inserted_obj, agent, parent_record)
```

melton-jason (approved):

specifyweb/workbench/upload/auditor.py
grantfitzsimmons: This can be opened as a new issue if you'd like!

specifysoftware:

PR #4095, opened 2023-10-09
Add a basic view btn to the report dialog

### LINKED ISSUE #4094: Add a basic view button in the report dialog

CarolineDenis:
Fixes #4094

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Reports/Parameters.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Reports/Parameters.tsx
@@ -88,8 +88,6 @@ export function QueryParametersDialog({
       }
       dimensionsKey="ReportParameters"
       header={query.name ?? reportsText.reports()}
-      icon={<span className="text-blue-500">{icons.documentReport}</span>}
-      onClose={handleClose}
       headerButtons={
         <>
           <span className="-ml-2 flex-1" />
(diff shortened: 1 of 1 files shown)


CarolineDenis:
Screenshot 2023-10-09 at 9 09 10 AM

grantfitzsimmons (changes requested):
Is there any way to make it so that Specify remembers the chosen view for a specific label/report?

For example, I can choose a report, then click on **Basic View**:

image

After I close out of the dialog or run a report, the next time I open it it's already in detailed view, but I want it to remember I want 'Basic View' everytime.

What do you think? Functionally, this PR does work! Only comment is that the wrapping issue on small screens still makes it a bit tough to use, but that's the same story in the query builder for the basic view.

https://github.com/specify/specify7/assets/37256050/561b640e-2a1b-4e19-aa29-12477c478d93

maxxxxxdlp (approved):
this looks good

as far as accomplishing what Grant asked for, it's not tricky, and we do that already in a few places
basically, normally when you use useCachedState, you hardcode some category name and key name, but in this case, you would have hardcoded category name, but key name would be dynamic (probably report/label id)

Look at definitions at any of these, and see their usages in the code to see examples:

https://github.com/specify/specify7/blob/345aeae52f907195330acc59671367a6d5fe2162/specifyweb/frontend/js_src/lib/utils/cache/definitions.ts#L70-L88

Would be happy to provide more information if you need help

maxxxxxdlp:
edit:
I should clarify that Grant asked that it would remember the "chosen view for a specific label/report", not for all reports (and query builder) together

up to you and Grant whether you implement that, but I think that would be cool to have, and not too complicated (and we already do similar in other places)

maxxxxxdlp (approved):
oh, all good. sorry, I must have gotten confused when reading the code

carlosmbe (approved):

PR #4099, opened 2023-10-09
Disable attachment gallery button when at least 1 resource is not uploaded

### LINKED ISSUE #4080: Disable record attachment gallery view button until files added
**Describe the bug**

The record gallery view button is enabled even before a file is added to the record causing an infinite loading dialog on click.

**Screenshots**

https://github.com/specify/specify7/assets/135047322/823bd59d-8fba-4d6c-bb1c-42d21de1e657

https://github.com/specify/specify7/assets/135047322/93e65372-3bb6-43a3-b01d-025176c4e53d

**Desktop:**
 - OS: Chrome
 - Browser: macOS Sonoma
 - Specify 7 Version: v7.9.1

CarolineDenis:
Fixes #4080

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
@@ -48,7 +48,7 @@ export function AttachmentsCollection({
 
   React.useEffect(() => {
     const savedAttachments = attachments.filter(
-      (attachment) => typeof attachment.id === 'number'
+      (attachment) => attachment.attachmentLocation !== null
     );
     setIsAttachmmentsSaved(savedAttachments.length === attachments.length);
   }, attachments);
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
@@ -56,9 +56,9 @@ export function AttachmentsCollection({
   return attachments.length > 0 ? (
     <>
       <Button.Small
+        disabled={!isAttachmmentsSaved}
         title={attachmentsText.attachments()}
         onClick={handleOpenAttachments}
-        disabled={!isAttachmmentsSaved}
       >
         {icons.gallery}
       </Button.Small>
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
@@ -44,11 +44,21 @@ export function AttachmentsCollection({
     })
   );
 
+  const [isAttachmmentsSaved, setIsAttachmmentsSaved] = React.useState(false);
+
+  React.useEffect(() => {
+    const savedAttachments = attachments.filter(
+      (attachment) => typeof attachment.id === 'number'
+    );
+    setIsAttachmmentsSaved(savedAttachments.length === attachments.length);
+  }, attachments);
+
   return attachments.length > 0 ? (
     <>
       <Button.Small
         title={attachmentsText.attachments()}
         onClick={handleOpenAttachments}
+        disabled={!isAttachmmentsSaved}
       >
         {icons.gallery}
       </Button.Small>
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
```diff
@@ -44,9 +44,19 @@ export function AttachmentsCollection({
...
 
+  const [isAttachmmentsSaved, setIsAttachmmentsSaved] = React.useState(false);
+
+  React.useEffect(() => {
+    const savedAttachments = attachments.filter(
+      (attachment) => attachment.attachmentLocation !== null
+    );
+    setIsAttachmmentsSaved(savedAttachments.length === attachments.length);
+  }, attachments);
+
```
maxxxxxdlp, line 55: this should be a useMemo

specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
```diff
@@ -44,9 +44,19 @@ export function AttachmentsCollection({
...
   );
 
+  const [isAttachmmentsSaved, setIsAttachmmentsSaved] = React.useState(false);
+
+  React.useEffect(() => {
+    const savedAttachments = attachments.filter(
+      (attachment) => attachment.attachmentLocation !== null
+    );
+    setIsAttachmmentsSaved(savedAttachments.length === attachments.length);
+  }, attachments);
```
maxxxxxdlp, line 54: (line 54) this looks like a bug.
while yes attachments is an array, and useEffect (or useMemo) accept an array, there are two problems here:
1. This will make react check every attachment in the array to see if it changed - that is redundant because in react normally on any change to item in an array, entire array is changed, so it's sufficient to just check if array is the same - no need to check individual array items (that's the benefit of not mutating things)
2. The length of the dependency array should not change - if it does, react doesn't like it and can emit warnings

thus, you should use `[attachments]` instead of `attachments`

maxxxxxdlp: but, I also just noticed that `attachments` is an array that is created from scratch on each render - so it will cause useEffect to be called every time - no point in useEffecy/useMemo then

specifyweb/frontend/js_src/lib/components/FormSliders/AttachmentsCollection.tsx
```diff
@@ -44,9 +44,19 @@ export function AttachmentsCollection({
     })
   );
 
+  const [isAttachmmentsSaved, setIsAttachmmentsSaved] = React.useState(false);
+
+  React.useEffect(() => {
+    const savedAttachments = attachments.filter(
+      (attachment) => attachment.attachmentLocation !== null
+    );
+    setIsAttachmmentsSaved(savedAttachments.length === attachments.length);
```
maxxxxxdlp, line 53: this can be simplified if you use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

take some time to look at all the array methods there are and learn to use them - arrays are very common in JavaScript, so this will make your life easier
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#array_methods_and_empty_slots

maxxxxxdlp (changes requested):
besides the code changes, I also have a question about the expected behavior:

(optional) rather than "Disable attachment gallery button when at least 1 resource is not uploaded",
would it make sense to have "Enable attachment gallery button when at least 1 resource is uploaded"?

maxxxxxdlp (approved):

maxxxxxdlp:
> No because if you click the + button and drag a picture in then click again +, you will have a picture on the first and nothing on the second. When you click on the gallery icon it will open a dialog where you will see the first picture and a loading for the second empty one. In my opinion it was making more sense to do it like this.

Sounds good. If you want, you can fix it so that it doesn't display a loading (just doesn't display anything) - or you can leave it as is - up to you

bronwyncombs (approved):
Looks good!

https://github.com/specify/specify7/assets/135047322/6d3ecf7a-896b-43c4-a5ed-11dbf3294233

PR #4101, opened 2023-10-11
Add a focus synchronizer in tree split

### LINKED ISSUE #4100: Split tree mode does not have a 'sync' button
> One thing that is still missing in split trees in 7 that is in 6 is the ability the sync the two trees to the same place.  It is useful when you want to navigate to similar positions in both trees once you have been navigating around in one of them.

**Requested By:** @acbentley

CarolineDenis:
Fixes #4100


grantfitzsimmons (changes requested):
image

Instead of the sync button being to the right of the synonymy button, it should be to the right of the horizontal/vertical switch button. 

It only applies to the split view tree, so having it disabled most of the time next to the regular action buttons is not clear. 

Oh, and it works! 🎉 

It would be best like this:

image

CarolineDenis:
> image
> Instead of the sync button being to the right of the synonymy button, it should be to the right of the horizontal/vertical switch button.
> 
> It only applies to the split view tree, so having it disabled most of the time next to the regular action buttons is not clear.
> 
> Oh, and it works! 🎉
> 
> It would be best like this:
> 
> image

Yes this is a work in progress

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -211,6 +211,14 @@ function TreeView<SCHEMA extends AnyTree>({
             if (!isHorizontal) resetDimensions();
           }}
         />
+        <Button.Icon
+          disabled={!isSplit}
+          icon="synchronize"
+          title={treeText.synchronize()}
+          onClick={() => {
+            states.second.focusPath[1](states.first.focusPath[0]);
```
maxxxxxdlp, line 219: (optional)
instead of always setting 2nd tree to where 1st tree is, would it be better to set the least recently tree's focus to where the most recently used tree's focus is?

whichever tree is focused at the moment, can assume that it's the one that was used most recently

would that be better UX?

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
CarolineDenis: mhhh might be... I just wanted to make it consistent. Always the same behavior, like that user understands directly wath is happening.

carlosmbe:
Could UX Testing instructions be provided. 

I'm experiencing some interesting things such as not being able to collapse a node without the clicking the sync button. Not sure if that's a bug or a feature

CarolineDenis:
> Could UX Testing instructions be provided.
> 
> I'm experiencing some interesting things such as not being able to collapse a node without the clicking the sync button. Not sure if that's a bug or a feature

collapse all should actually be disabled when tree is split

carlosmbe (approved):

grantfitzsimmons (approved):

PR #4106, opened 2023-10-12
Add Tool buttons to app ressources in full screen

CarolineDenis:


specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
```diff
@@ -267,117 +379,6 @@ export function AppResourceEditor({
...
-                })
-              );
-
-              return false;
-            }}
-          />
-        ) : undefined}
-      </>
-    ),
+    footer: footer,
```
maxxxxxdlp, line 382: if `isFullScreen`, then you might want to return undefined instead of footer here to avoid duplicate footer (and then change types and usages accordingly)

specifyweb/frontend/js_src/lib/components/AppResources/Tabs.tsx
```diff
@@ -41,7 +39,7 @@ export function AppResourcesTab({
   resource,
   directory,
   data,
-  isFullScreen: [isFullScreen, handleChangeFullScreen],
+  isFullScreen: [isFullScreen, _],
```
maxxxxxdlp, line 42: does this component ever need `handleChangeFullScreen`? if not, then change the type to remove it from the type (replace `GetSet<boolean>` with `boolean` and then replace usages accordingly)

maxxxxxdlp (approved):

maxxxxxdlp:
code looks good, though someone should test the behavior to make sure it's good:
- make sure footer elements are displayed when in full screen or not in full screen when editing an app resource
- open schema config, and try to edit a formatter or a weblink - should open an app resource editor in a dialog with footer correctly displayed

grantfitzsimmons (approved):
https://github.com/specify/specify7/assets/37256050/36d62d0f-1fb4-4c42-af19-fa8268d13c68

Works as expected, but is there any way to make it so that it doesn't reopen and instead remains after saving? See that after saving it closes briefly and reopens.

CarolineDenis:
> Screen.Recording.2023-11-01.at.10.54.01.AM.mov 
> 
> Works as expected, but is there any way to make it so that it doesn't reopen and instead remains after saving? See that after saving it closes briefly and reopens.

It doesn't really closes it navigates to another page. 

When editing: specify/resources/view-set/18/Accession/testIssue1
After saving: /specify/resources/view-set/18/

acwhite211 (approved):

PR #4107, opened 2023-10-12
 Display count next to table name

### LINKED ISSUE #4073: `xml-editor`: Show count of form definitions next to the table name, not on the right
With the current behavior, it's unclear which table the count corresponds to without the user running their finger across the screen to make sure it is pointing to the correct line. Placing the counts directly to the right of the table name makes it immediately clear.

## Desired Behavior:
_See some of https://github.com/specify/specify7/issues/4072 here as well_
image

## Current Behavior:
image#4072

CarolineDenis:
Fixes #4073


maxxxxxdlp (approved):
good idea

I put them on the right initially because I wanted all counts to be aligned with each other - and since during development my screen was very narrow always (because of devtools open on the side) it looks good - but yeah, would totally look off on a wide screen

pashiav (approved):
Looks good!
issue-4073

carlosmbe:
It looks good. Just curious if there's a way to verify and test the counts. Otherwise, I'd approve @CarolineDenis

bronwyncombs (approved):
Count is accurate and placement looks good 👍 

https://github.com/specify/specify7/assets/135047322/a88d12d1-6b9b-481d-a09e-928d4104a4fc

PR #4108, opened 2023-10-12
Display only present tables + button add new tables

### LINKED ISSUE #4072: `xml-editor`: Only show tables that have form definitions in the current form definition by default
image

Showing **all** non-advanced tables slows down the form editing process quite a bit. It would make more sense to show all tables that currently have form definitions associated with them already.

You can make it so that the "non-advanced" forms are at the top and separated from the advanced ones if this distinction is useful. 

## Idea

image

You would then click **Add** and choose from a dialog of tables which form you want to create.

_Note: Gray button is just for the mock-up, this would look better with a green plus and Add text_

CarolineDenis:
Fixes #4072


maxxxxxdlp (commented):
string change looks good, but if I read the issue correctly (https://github.com/specify/specify7/issues/4072), Grant asked for it to show by default only tables that have forms defined (and all tables when the checkbox is checked)
I might be missing it but I don't see that in this PR

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
```diff
@@ -127,13 +139,34 @@ export function TableList({
...
-      </Label.Inline>
-    </>
+    );
+  };
+
+  return (
+    <div className="flex flex-col items-start gap-2">
+      {tableList(presentFilteredTables)}
+      <Button.Save onClick={handleOpen}>{commonText.add()}</Button.Save>
+      {isOpen && (
```
maxxxxxdlp, line 147: sorry, looks like you made it more complicated than what I meant
There should be a checkbox just like there was before. if you click the checkbox, the list above the checkbox excludes or includes the tables that have 0 forms - that's it - no need for dialogs

if you need an example of how such checkbox works, look at the base table selection dialog in the workbench mapper

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
CarolineDenis: I followed Grant's requests :) 
I will create another version with the check box request and post 2 videos

CarolineDenis:
@grantfitzsimmons request with add button 

https://github.com/specify/specify7/assets/108160931/e065ae23-4514-47d5-b609-1b658cf14c77

@maxxxxxdlp suggestion with check box 

https://github.com/specify/specify7/assets/108160931/24643e14-c4b7-4ccf-96fb-b692da672aad

CarolineDenis:
> I prefer the `Add` button, but I am biased.
> 
> @carlosmbe @bronwyncombs @maxxxxxdlp Can you take your pick between these two?

The check box option has less clicks
The add button is more clear

So 50/50

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
```diff
@@ -85,20 +85,28 @@ export function TableList({
...
+      Object.values(sortedTables).filter((table) => {
+        const childrenResult = children?.(table);
+        return typeof childrenResult === 'string';
+      }),
+    [sortedTables, children]
+  );
+
+  const tablesToDisplay = React.useMemo(
+    () => (showHiddenTables ? sortedTables : presentFilteredTables),
+    [showHiddenTables, sortedTables, presentFilteredTables]
```
maxxxxxdlp, line 103: (optional) this can be combined into a single useMemo - would be less code

performance difference won't be noticeable either way

maxxxxxdlp (approved):

maxxxxxdlp:
> Can you take your pick between these two?

I like the checkbox variant because:
- The code with the checkbox is much simpler than the code with the dialog
- It is consistent with how this functions in other places in specify (workbench mapper base table selection dialog, schema config dialog, app resource resource/field formatters editor, and now form editor)

I don't like the dialog because:
- The "Add" button is not clear to me - as a user I would think it would either add a new form definition (thus I have to select a table first), or permanently add a new table to the list (like in the data entry/interactions/query list dialogs)
- The "Show Advanced Tables" checkbox in the dialog is also getting a bit confusing - because we have 3 lists now - list of tables with forms, list of common tables, and list of all tables - seems too much.

But we are building software for the user, and if you say it's better for the user, let's go with it.

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
CarolineDenis: initially we used a filter prop for 'advance tables', we do not use it here anymore. Is there any advantages of keep that? At the moment it just displays all tables that already have a resource and when checking the check box all the other ones

carlosmbe (changes requested):
I would prefer an all tables checkbox. Also I can not scroll down on this branch to see additional tables.

There are more tables. I can see an icon cut off.

https://github.com/specify/specify7/assets/53784701/eb6ebde6-c834-4d21-a6c2-237cf1564325

emenslin (approved):
I like the checkbox and I think it makes what it does clear and simple to the user.

acwhite211 (approved):

Issue #4110, opened 2023-10-13
Improve the `docker-compose.yml` setup


grantfitzsimmons:
Instead of repeating values twice that only need to be established once, we should consider using a `.env` file.

Right now, you need to establish this environment **twice**, once for the worker and once for the service.

```yml
    environment:
      - DATABASE_HOST=mariadb
      - DATABASE_PORT=3306
      - DATABASE_NAME=specify
      - MASTER_NAME=master
      - MASTER_PASSWORD=master
      - SECRET_KEY=change this to some unique random string
      - ASSET_SERVER_URL=http://host.docker.internal/web_asset_store.xml
      - ASSET_SERVER_KEY=your asset server access key
      - REPORT_RUNNER_HOST=report-runner
      - REPORT_RUNNER_PORT=8080
      - CELERY_BROKER_URL=redis://redis/0
      - CELERY_RESULT_BACKEND=redis://redis/1
      - LOG_LEVEL=WARNING
      - SP7_DEBUG=true
```

We could use a `.env` file to set these things once for both to prevent unnecessary issues.

maxxxxxdlp:
This issue is a duplicate of https://github.com/specify/docker-compositions/issues/20
Also, can you take a look at the changes in https://github.com/specify/docker-compositions/pull/21?

grantfitzsimmons:
@maxxxxxdlp They are different repositories with different goals, I think both need to happen, but implementation is different (development v. production) :smile:

I am reviewing your PR and implementing many of your ideas on this repo as well. Great work!

PR #4111, opened 2023-10-13
Use `.env` for docker deployment

grantfitzsimmons:
Solves #4110 

This is just a proof of concept to improve deploying Specify 7


maxxxxxdlp:
Can you take a look at the changes in https://github.com/specify/docker-compositions/pull/21?
It is complete, just needs your review and testing - that pull request has been open for 4 months and no one looked at it(((

acwhite211:
Looks good

acwhite211 (approved):

PhatWheZ:
You guys not considered using env_file instead of environment. This would reduce the amount of lines in the docker-compose.yml file and work exactly how you are looking to implement this change

grantfitzsimmons:
Thank you for the advice @PhatWheZ! :smile:

.env
```diff
@@ -0,0 +1,36 @@
...
+MYSQL_ROOT_PASSWORD=password
+DATABASE_NAME=specify
+# When running Specify 7 for the first time or during updates that 
+# require migrations, ensure that the MASTER_NAME and MASTER_PASSWORD 
+# are set to the root username and password. This will ensure proper 
+# execution of Django migrations during the initial setup. 
+# After launching Specify and verifying the update is complete, you can 
+# safely replace these credentials with the master SQL user name and password.
+MASTER_NAME=root
+MASTER_PASSWORD=password
```
melton-jason, line 12: Did you want to use the default password for the database container? 
It is currently not `password`, so this will throw a rather unhelpful message for most users when building Specify 7:
`django.db.utils.OperationalError: (1045, "Access denied for user 'root'@'172.18.0.10' (using password: YES)")`

If the error is intentional and we do not want access to be granted until they modify this file, it would be helpful to either update the documentation (on either the Speciforum or on the Installation Instructions on GitHub... or both), or provide a more helpful error message if possible .

melton-jason (changes requested):

.env
grantfitzsimmons: The default password for the database container depends on what you configure it to be in the `.env` file. Did you try to deploy this for testing?

The `MYSQL_ROOT_PASSWORD` variable sets the root password for the container in the `.env` file. See my deployment without any modification:

```zsh
❯ docker exec -it 5c915398e224 /bin/bash                                                             at 06:54:10 PM
root@5c915398e224:/# mysql -uroot -ppassword
Welcome to the MariaDB monitor.  Commands end with ; or \g.
Your MariaDB connection id is 7
Server version: 10.11.5-MariaDB-1:10.11.5+maria~ubu2204 mariadb.org binary distribution
```

grantfitzsimmons: Exactly, I switched it back to the original password, and I got the same error as you, in the other direction.
```
django.db.utils.OperationalError: (1045, "Access denied for user 'master'@'172.19.0.10' (using password: YES)")
```

So are you advocating for using `master` instead for backwards compatibility? As someone who switches between databases often, I prefer to use `root` (or another high level user) in testing/development so that I can run all migrations without needing to modify the configuration outside of changing the DB name.

melton-jason: > The default password for the database container depends on what you configure it to be in the .env file. Did you try to deploy this for testing?

Interesting. Yes, you are correct. 
Turns out `docker container prune`, `docker image prune -a`, `docker volume prune -a`, and even `docker system prune -a` was not removing the old, dangling, or unused containers/images/volumes on my system.

I had to manually, individually remove the stopped containers,  images, and volumes to completely run Specify "from scratch". 
Even then, on the first time the containers were built Django threw an `OperationalError: Unknown MySQL server host 'mariadb' `. Stopping the containers and restarting them fixed this problem (most likely caused because the order the containers were build was "incorrect"?)

But yes, the `mariadb` does use the default password as specified in the `.env` file :partying_face: 

> So are you advocating for using master instead for backwards compatibility? As someone who switches between databases often, I prefer to use root

I also prefer to use `root` in such cases.

melton-jason (approved):
This is fine now! 
My earlier request for changes was largely in part related to an issue I had with docker, not with these changes.

> > The default password for the database container depends on what you configure it to be in the .env file. Did you try to deploy this for testing?
> 
> Interesting. Yes, you are correct. Turns out `docker container prune`, `docker image prune -a`, `docker volume prune -a`, and even `docker system prune -a` was not removing the old, dangling, or unused containers/images/volumes on my system.
> 
> I had to manually, individually remove the stopped containers, images, and volumes to completely run Specify "from scratch". Even then, on the first time the containers were built Django threw an `OperationalError: Unknown MySQL server host 'mariadb'`. Stopping the containers and restarting them fixed this problem (most likely caused because the order the containers were build was "incorrect"?)
> 
> But yes, the `mariadb` does use the default password as specified in the `.env` file 🥳

- https://github.com/specify/specify7/pull/4111#discussion_r1360530389

CarolineDenis (approved):

.env
maxxxxxdlp: @melton-jason If you want to go thermonuclear on docker, here is the command for that:

Screenshot 2023-10-16 at 17 34 30

it will delete volumes too

melton-jason: Yes, thank you! 
I will be careful about wielding such power over Docker. :laughing:

.env
maxxxxxdlp, line 1: doesn't matter for development much, but just a thing to note for the production when we start using `.env` files there too:
by having one large `.env` file, each container receives env variables from all containers.
which means, if some attacker were to get access to for example report runner (because it's running old java and old dependencies that hasn't been updated in a while), the would gain database credentials, master user credentials, and asset server credentials

the solution of course is to use separate `.env` files, but that is less convenient. don't know how important this is (maybe either of you should read online for security recommendations), but just something to keep in mind

maxxxxxdlp: noticed that `.env` file is not given to the report runner - good

also, ChatGPT says env secrets exist - https://docs.docker.com/engine/swarm/secrets/ (though probably too much overhead for us)

maxxxxxdlp (approved):

realVinayak (approved):

specifysoftware:

https://discourse.specifysoftware.org/t/specify7-in-docker-keeps-on-failing-on-port-8080/1361/19

maxxxxxdlp:
@grantfitzsimmons could you please update https://github.com/specify/specify7/wiki/Docker-Workflow-for-Development with the instructions on how to get the dev environment running now?

do I make a copy of the .env file and then do some edits?
or do I change my MySQL password to "password" from "root"?

melton-jason:
> The default password for the database container depends on what you configure it to be in the .env file. Did you try to deploy this for testing?

https://github.com/specify/specify7/pull/4111#discussion_r1359982764

@maxxxxxdlp if you are using a prior development installation of Specify through docker before these changed, you will probably have to change the `MYSQL_ROOT_PASSWORD` to whichever password the container expects. 
Previously, this was `root` so if you do not wish to change the root password in the container, you should be able to change `MYSQL_ROOT_PASSWORD` to root. 

Additionally, the same needs to be considered for the `MASTER_PASSWORD` variable. 
For the previous defaults of the containerized database, this was `master`. However, if that is _also_ giving you issues (or simply for convenience while developing), you can  replace the master creditionals in the .env to the root username/password (which you should do anywas when running migrations)

PhatWheZ:
@maxxxxxdlp you will need to simply set `.env` in your Git ignore section so it does not show up when pushing.
Double check that `.env` is being seen in your container.

example:
```
  specify7-saiab:
    restart: unless-stopped
    image: specifyconsortium/specify7-service:v7.9.0
    container_name: specify7-saiab
    init: true
    volumes:
      - "specify6803:/opt/Specify:ro"
      - "static-files-saiab:/volumes/static-files"
      - "/etc/timezone:/etc/timezone:ro"
    env_file:
      - ./environment/specify-saiab.env
```

maxxxxxdlp:
Thanks! I guess that's one of the ways to do it

You reminded me of a similar solution I used in the past - see the ["Initial State" section in the specify7-test-panel](https://github.com/specify/specify7-test-panel/#initial-state) repository - you can have some file be part of a repository, and yet have git ignore the changes to that file

specifysoftware:

PR #4114, opened 2023-10-13
Improve formatter shema config UI  …

### LINKED ISSUE #3565: `xml-editor`: Add definition is not enabled by default for a new formatter
**Describe the bug**
When making a new formatter, the 'Add definition' is not enabled by default. It suggests that 'Condition Field' needs to mapped. However, 'Condition Field' can be temporarily mapped, and 'Add definition' is then enabled. Then, fields can be added, and then 'Condition Field' can be unmapped. Since no error occurs, it looks like 'Condition Field' is allowed to be unmapped (which I think is the current behavior)

**To Reproduce**
1. Go to the visual editor for defining Formatters.
2. Click on 'Collecting Trip Attribute' table for as an example.
3. Add definition is disabled.
4. Add a Condition field temporarily, add fields and then remove the Condition field. This should be allowed be default.

**Expected behavior**
Add definition should be enabled by default, and not require a Condition field definition.

image

CarolineDenis:
Fixes #3565


CarolineDenis:
https://github.com/specify/specify7/assets/108160931/0774a72d-bea8-48f4-9ca8-43e0bf0f0c86

CarolineDenis:
Screenshot 2023-10-18 at 8 54 08 AM

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -169,78 +237,103 @@ function Fields({
   readonly fields: GetSet<Formatter['definition']['fields'][number]['fields']>;
   readonly onDelete: (() => void) | undefined;
 }): JSX.Element {
+  const [displayFormatter, setDisplayFormatter] = React.useState(false);
```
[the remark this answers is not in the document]
maxxxxxdlp: if any field has a non-default formatter, should the default value here be set to true then?

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -169,78 +237,103 @@ function Fields({
...
+    <>
+      {fields.length === 0 ? null : (
+        <table
+          /*
+           * REFACTOR: replace min-w-[35rem] with a container query that replaces
+           *   table layout with list layout
+           */
+          className={`
+            grid-table min-w-[35rem]
+            gap-y-2 gap-x-4
```
maxxxxxdlp, line 251: (optional) try to have the gap in both directions be the same unless there is a reason to be otherwise
(if anything, if I do decide to make gaps uneven, the gap between rows is usually larger than the gap between columns as data within the row are related to each other, while data between rows is not)

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -169,78 +237,103 @@ function Fields({
...
+                ? 'grid-cols-[min-content_max-content_auto_min-content]'
+                : 'grid-cols-[min-content_1fr_min-content]'
+            }
+          `}
+        >
+          <thead>
+            <tr>
+              <th>{resourcesText.separator()}</th>
+              <th>{schemaText.field()}</th>
+              {displayFormatter && <th>{schemaText.fieldFormat()}</th>}
```
maxxxxxdlp, line 263: could you please rename this column name to "Custom Field Format" or something like that?

specifyweb/frontend/js_src/lib/localization/resources.ts
```diff
@@ -450,4 +447,13 @@ export const resourcesText = createDictionary({
...
     'en-us': 'You need to save this form before you edit another one',
   },
+  defaultTableFormatter: {
+    'en-us': 'Default Table Formatter',
+  },
+  conditionalFormatter: {
+    'en-us': 'Conditional Formatter',
+  },
+  showFieldFormatters: {
+    'en-us': 'Show Field Formatters',
```
maxxxxxdlp, line 457: While I like the fact that the non-default formatter option is now not always displayed to make UI simpler for common cases, I feel like we could better describe the role of this checkbox than "Show Field Formatters"

maybe "Customize Field Formatters" or "Use non-default field formatters" or "Use Custom Field Formatters" (because that's effectively what it's doing - the default value Formatter will be used unless you customize it)

specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
```diff
@@ -251,7 +251,9 @@ export function ResourceMapping({
 
   return (
     <Ul
-      className="flex flex-wrap gap-2"
+      className={`
```
maxxxxxdlp, line 254: can you test how these changes work if there is more than one box displayed? (i.e pick a -to-one relationship field, then the 2nd box would be displayed)

specifyweb/frontend/js_src/lib/localization/resources.ts
maxxxxxdlp: ALSO, I don't remember this too fwell, but doesn't this UI also allow you to pick which table formatter you want to use, rather than just what field formatter you want to use?
i.e if in the "Field" column you select a -to-one relationship?

then also if you select a -to-many relationship, does the UI offer selecting which aggregator to use?

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -74,40 +102,64 @@ function Definitions({
...
+    if (formatter.definition.fields.length === 0) {
+      handleChange([
+        ...formatter.definition.fields,
+        {
+          value: undefined,
+          fields: [],
+        },
+      ]);
+    }
+  }, [formatter.definition.fields, handleChange]);
```
maxxxxxdlp, line 142: (optional micro-optimization):

```suggestion
  const needFormatter = formatter.definition.fields.length === 0;
  React.useEffect(() => {
    if (needFormatter) {
      handleChange([
        ...formatter.definition.fields,
        {
          value: undefined,
          fields: [],
        },
      ]);
    }
  }, [needFormatter, handleChange]);
```

This way the useEffect is only called when there is need to, rather than being called on every keystroke if user is typing.

don't think it would make any noticeable performance difference in this case, but in places like the query builder or workbench it might - so don't need to change anything here, just keep this in mind for the future

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -74,40 +102,64 @@ function Definitions({
...
+
   return table === undefined ? null : (
     <div className="flex flex-col gap-4 divide-y divide-gray-500 [&>*]:pt-4">
-      {trimmedFields.map(({ value, fields }, index) => {
-        const handleChanged = (
-          field: Formatter['definition']['fields'][number]
-        ): void =>
-          handleChange(replaceItem(formatter.definition.fields, index, field));
```
maxxxxxdlp, line 97: (optional)
while it might have looked a bit weird to define the function inside a .map(), I feel like the code was simpler that way because you didn't need to explicitly pass the index argument.

defining functions inside of .map() shouldn't feel weird because you do that all the time (the first argument to .map() it an arrow function too after all), and each callback handler is an arrow function too

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -116,11 +168,28 @@ function Definitions({
...
+                  trimmedFields.length < 2
+                    ? undefined
+                    : (): void =>
+                        handleChange(
+                          removeItem(formatter.definition.fields, index)
+                        )
+                }
+              />
+            </>
+          )}
```
maxxxxxdlp, line 187: a lot of code duplication here. could you please extract this into a variable and reuse it?

but really, even that shouldn't be needed. could be simplified to look like this:

```tsx
.map((...)=>{
  <div ...>
    {hasCondition && <Label.Block>...</Label.Block>}
    <Fields ... />
  </div>
})
```

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -169,78 +237,103 @@ function Fields({
...
-                {resourcesText.deleteDefinition()}
-              </Button.Danger>
-            )}
-          </td>
-          <td />
-        </tr>
-      </tbody>
-    </table>
+    <>
+      {fields.length === 0 ? null : (
```
maxxxxxdlp, line 243: it's considered bad practice/code smell when the component decides for itself if it should be rendered or not - such power should lie in it's parent (i.e if the parent decided than the child should be displayed, than the child should be displayed - otherwise, if both parent decides that, and child decides that, there could be a bug where parent wants to display the child, but child decided not to show up).
could you please move this check into the parent? (i.e `{fields.length > 0 && <Fields ... />`)

of course there are exceptions to all rules - the exception to this one is when the child component is called in multiple places with the same condition, or if condition is not available outside of the child (i.e child is fetching some data, and while loading doesn't render anything to the screen) - neither of those apply here

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -169,78 +237,103 @@ function Fields({
...
+        <table
+          /*
+           * REFACTOR: replace min-w-[35rem] with a container query that replaces
+           *   table layout with list layout
```
maxxxxxdlp, line 247: if you are curious about this and want to explore more, here is the tailwind plugin for container queries: https://github.com/tailwindlabs/tailwindcss-container-queries

it's an official plugin from the developers of tailwind and they talked how it would be integrated into the core tailwind in the future.

container queries are cool and useful, especially for complicated layouts like those in sp7 - they can improve adaptive design and aid in code reuse - if you learn them now, you will find how to apply them in other places
(i.e with container queries, the page could display differently based on whether the side menu is collapsed or not as that affects available space - without container queries you can only listen for browser window size change, not the container size change, which can be not enough)

maxxxxxdlp (changes requested):
I like the changes, and code is looking pretty good!
most comments are optional suggestions/small improvements

maxxxxxdlp:
Also, please test your layout more with cases when "Field" column displays more than one box - when you have a relationship selected

specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
```diff
@@ -116,16 +116,6 @@ export function FormatterWrapper(): JSX.Element {
...
-          <Label.Block>
-            {resourcesText.title()}
-            <Input.Text
-              isReadOnly={isReadOnly}
-              value={getSet[0].title}
-              onValueChange={(title): void =>
-                getSet[1]({ ...getSet[0], title })
-              }
-            />
-          </Label.Block>
```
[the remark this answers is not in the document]
CarolineDenis: Where would we see the title by example?

specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
CarolineDenis: Screenshot 2023-10-19 at 2 07 11 PM
Screenshot 2023-10-19 at 2 05 50 PM
Screenshot 2023-10-19 at 2 05 28 PM
Screenshot 2023-10-19 at 2 05 22 PM

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
CarolineDenis: the field.length is just for the table not for the rest (Add button etc) 
If I move it to the parent than we are missing those

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
CarolineDenis: it was to access it in my use effect

specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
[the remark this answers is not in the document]
maxxxxxdlp: In Schema Config when you are picking a formatter
And in web link editor when you are picking a formatter

Both title and name field is also present in web link, field formatters, aggregators, pick lists, tree ranks and other places

And name must be unique, where as title doesn't have to be
name and title serve different use cases that's why they are in separate fields

maxxxxxdlp (approved):

pashiav (changes requested):
When selecting `Conditional Formatter` => any `Optional Fields`, the size of the pop up increases dramatically. This makes the text in the selected option not as smooth to read for the user as it is centered aligned and much further from the headings. 
Could we fix the change in popup size when selecting an option in `Conditional Formatter` or left-align the text?

https://github.com/specify/specify7/assets/118556884/55bd4bb8-5ac3-4d89-a36b-bf0e19b58f6b


--
Add Field works without a Condition field. Functionality looks good!

https://github.com/specify/specify7/assets/118556884/96ea39c7-fbd5-4134-9c09-825ffb796252

CarolineDenis:
@pashiav, thank you for the suggestion, it will be fixed once this is merged in xml-editor as the align-left fixes (https://github.com/specify/specify7/pull/4174) are merged with it already.

pashiav (approved):

maxxxxxdlp:
@CarolineDenis I think besides the text alignment issue, the other problem here is that the Conditional formatter field takes up the entire available horizontal space rather than just what is needed for it's content. if you want, this can be fixed simplify by wrapping the element in a `<div>` (currently this select element is direct child of a display:flex, which forces all children to be full width - by adding an intermediate `<div>`, while `<div>` would be full width, the select element won't be forced to be wider than it needs to be anymore

emenslin (approved):
Aside from the issues with the popup that others mentioned it functions as expected.

PR #4124, opened 2023-10-17
Fix form definition being cached

realVinayak:
Clears the form definition in the memory object after HEAD requests are made to the back-end, and are fulfilled.


CarolineDenis (approved):

specifyweb/frontend/js_src/lib/components/FormParse/index.ts
```diff
@@ -85,6 +86,10 @@ export const getViewSetApiUrl = (viewName: string): string =>
         : undefined,
   });
 
+export const clearViewLocal = (viewName: string) => {
+  views = removeKey(views, viewName);
```
maxxxxxdlp, line 90: (optional) `views[viewName] = undefined;` - then you don't have to change `views` from `let` to `const`

maxxxxxdlp (approved):
ah, right, forgot it was cached there too

even though the form is cached on the back-end too, the caching is needed there (or at least was needed at one point) because otherwise when you go between records in a subview/record selector, the brief time it took to make a request and receive a cached response from browser was still enough to cause issues (either layout blink, or layout shift, don't remember)
ideally of course, as much caching as possible would be delegated to the browser or to some library

grantfitzsimmons (approved):

PR #4135, opened 2023-10-18
Allow the user to switch between labels and localized field

### LINKED ISSUE #4127: `xml-editor`: Add checkbox to enable/disable localized field labels while editing forms
image

This would be useful to make it clear what fields are being modified. Ideally, this would not require a full page reload since it should be toggle-able.

CarolineDenis:
Fixes #4127


specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
```diff
@@ -148,6 +150,7 @@ export function FormEditorWrapper(): JSX.Element {
             ? icons.switchVertical
             : icons.switchHorizontal}
         </Button.Small>
+        <UseLabelsSchema />
```
maxxxxxdlp, line 153: (optional) could you put a `<span className="flex-1 -ml-4" />` here to put the checkbox on the side? does that look better or not? does it lead to better UX (i.e it may look weird to have button to close to the checkbox)

specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
```diff
@@ -176,6 +179,25 @@ export function FormEditorWrapper(): JSX.Element {
   );
 }
 
+function UseLabelsSchema(): JSX.Element {
```
maxxxxxdlp, line 182: could you reuse this component with the Form Meta menu? or does the logic differ a bit?

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
CarolineDenis: the logic differs so I had to create a new one.

specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
CarolineDenis: We tried and it was decided to keep the check box on the left next to the switch button :)

pashiav (approved):
Looks good! 

https://github.com/specify/specify7/assets/118556884/56f5629f-d51f-45ba-af6b-e0086d28bcec

bronwyncombs (approved):
Placement in the UI looks good and is working as expected. There's the same refresh call that happens when done via the FormMeta menu.

https://github.com/specify/specify7/assets/135047322/a3bbd6f0-053d-4a39-98f3-044f9fc10526

grantfitzsimmons:
Do we _have_ to reload the entire page? Can we not make it specific to the form?

maxxxxxdlp:
> Do we have to reload the entire page? Can we not make it specific to the form?

for now the form system and specify is working under the assumption that schema localization does not change, and thus it does not listen anywhere for schema localization changing.
there is also no support at the moment to change it for one part of the front-end (i.e form preview), without affecting other places.
and the way disabling schema localization works is that it just refuses to load the schema localization - which is something you can only do when the page is loading, not later on.

while it would be possible to refactor code to react to schema localization dynamically, that is not in place now, so until that refactor is done this would have to remain in place.
while it shouldn't be too hard to implement support for live schema update in forms (disregarding all the other places where schema localization is used as it is not needed for this feature), the implementation of it that works just in form preview, and implementation of it that works for regular form too is not going to be very different in difficulty

PR #4143, opened 2023-10-20
Trigger update value sooner in the render cycle

realVinayak:
This was trickier to debug

In the `xml-editor`, the parser https://github.com/specify/specify7/blob/93e46207474e2c12d5fe170bddc0a16ca683e9ef/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx#L62 gets defined in the initial render. 
<br/>

So, the initial render in the Field.tsx, the fields which are required (as defined in the parser) get their `required=true` prop https://github.com/specify/specify7/blob/f311809ee78c8564dc618ad87684b74295cc6301/specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx#L156. However, the value of the input is not set till updateValue is called https://github.com/specify/specify7/blob/a4a66e63a8ec27c82c60bca0c01bc0e1410b0725/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx#L208-L218
<br/>

So, Validity of all required fields becomes invalid initially. This can be confirmed by putting the following anywhere before the return in useResourceValue.tsx.
```
console.log(
    'Value missing for ',
    field?.name,
    inputRef.current?.validity.valueMissing
  );
```
The result is:
image

Thus, the level field and name field become invalid. However, the clearing of level field happens when name field reports its validity, which triggers a focus event on name field, and a blur event on the level. field https://github.com/specify/specify7/blob/0778c338d4a8781d65fc80397e28187e03ea2298/specifyweb/frontend/js_src/lib/hooks/useValidation.tsx#L105

The above can be confirmed by setting a breakpoint at https://github.com/specify/specify7/blob/a4a66e63a8ec27c82c60bca0c01bc0e1410b0725/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx#L189. 
image
Notice that the target is the level field, and that callstack also has useValidation file. Going inside useValidation file, this is shown 
image
Notice that this input is the name field (which is also invalid right now, confirmed by console)
```
> input.validity.valueMissing
true
> input.id
'form-SpViewSetObj-3-name'
```

Going back to the scope of onBlur event, and inspecting the event.target (which is `input`), I see the following
```
> input.id
'form-SpViewSetObj-3-level'
> input.value
''
```

Now, setting a breakpoint at https://github.com/specify/specify7/blob/a4a66e63a8ec27c82c60bca0c01bc0e1410b0725/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx#L144, I see 
image
Hence, the level is set as empty string in the backbone resource.


realVinayak:
A false track of the origin of the bug is that setValidation (in useValidation.tsx) doesn't trigger reportValidity() if empty error message and focus in production. However, in xml-editor, it always trigger reportValidity. 

Production: https://github.com/specify/specify7/blob/f311809ee78c8564dc618ad87684b74295cc6301/specifyweb/frontend/js_src/lib/hooks/useValidation.tsx#L84-L85
XML-Editor
https://github.com/specify/specify7/blob/a4a66e63a8ec27c82c60bca0c01bc0e1410b0725/specifyweb/frontend/js_src/lib/hooks/useValidation.tsx#L99-L105
@maxxxxxdlp was that intentional?

Making the behaviour similar to production also fixes the issue, as onBlur doesn't get called, hence level is not set to empty string.

However, this is still incorrect fix. This is because replacing the useValidation hook file in production with the useValidation hook file in xml-editor DOES NOT clear the field. This indicates that bug is not in the useValidation hook. (Note: some non-breaking things needed to be changed to properly replace the hook between branches, in case someone wants to replicate)

Additionally, it shouldn't be assumed that onBlur won't be triggered because of non-specify code at any point during the rendering

maxxxxxdlp:
On the xml-editor branch I did refactoring of useValidation (and useResourceValue) hooks to simplify code, and to remove some edge cases that I though don't need to be handled (either because they won't happen, or because if they do happen the impact won't be bad). I belive I also discovered some bugs that weren't yet reported in production, so I fixed those.

But I definitely could have been wrong on either of these points when doing the refactor.

> A false track of the origin of the bug is that setValidation (in useValidation.tsx) doesn't trigger reportValidity() if empty error message and focus in production. However, in xml-editor, it always trigger reportValidity.

However, in this case if I remember correctly (and mind you this was half a year ago), this was the reasoning:

While the validation message generated by specify might be empty, it does not mean the field has no validation errors - if field has "required" attribute, and no value, then the browser will still display the validation error. If there is no validation error of any kind, then .reportValidity(); won't do anything.

And in cases when that behavior is not desirable, the "type" argument exists in the setValidation function - it is "auto" by default, meaning validation won't be reported if there is no custom error message - in your case apparently, the "type" is set to "focus" when the function is called (which would mean show the error message, even if the error came from browser, not from specify -- well of course specify was the one that told the browser in which condition to display the error, i.e when filed is required so you could say it comes from specify in both cases)

maxxxxxdlp (approved):

CarolineDenis (approved):

realVinayak:
@specify/ux-testing: Can this be tested soon? 

UX Testing instructions

1. Go to edit any form definition in app resources. Click on the pencil icon next to the name in the header 
image
2. In the xml-editor branch, the first field (level) will be cleared out. Open the same resource in a different (non-xml-editor) branch, and the level field should be present.
3. This branch should not clear the level field

The same clearance can be noticed in any form which have more than one required fields. For example, taxon. Open a taxon form, and enter the parent, name and select the rank. In the xml-editor, going back to the saved required will clear the name. However, this branch should not clear the field.

pashiav (approved):
Having the value left-aligned would be better for ux, but functionality is good! 
issue-4142

maxxxxxdlp:
> Having the value left-aligned would be better for ux, but functionality is good!

the numeric inputs are right aligned (as inspired by Excel and other software that includes lots of numbers) because this way you can compare the size of numbers easier when two numeric fields are on top of each other.

for example, compare this:
```
1233521
123
```

and

```
1233521
        123
```

even though both number start with `123`, the 2nd example makes it much easier to see that the `123` in the first number is of much greater magnitude than `123` in the 2nd example (this difference is even more important when length of numbers differs only by 1 digit, thus potentially causing user errors)

but, your criticism is valid in that when an input field is so wide, having some values right aligned is not great (this was intended for narrow fields like on a collection object form)

a few things to note
1. user can already disable this behavior in user settings
2. we could probably improve the UI by adding a small gap (0.25rem) between the numeric input value and the up/down arrow buttons - I though we were alreadying doing this, but maybe it stopped working.
3. maybe we shouldn't be right-aligning inputs when there is only one column? don't know

@bronwyncombs @realVinayak feel free to create a separate issue out of this

PR #4144, opened 2023-10-20
Assign new title to view and view def

### LINKED ISSUE #4134: `xml-editor`: Creating a new form definition based on a default with a custom name does not assign the new view name to the corresponding viewdef
# Steps to Recreate
1. Go to a form (https://nbmmnb5523-xml-editor.test.specifysystems.org/specify/resources/view-set/2/)
2. Select the table "Preparation" (https://nbmmnb5523-xml-editor.test.specifysystems.org/specify/resources/view-set/2/Preparation)
3. Create a new form definition by clicking **Create**
4. Select an existing one (for example, use the `Preparation` form from the `common.views.xml` seen here:
    image
5. Preview and click **Select**
6. Give it a name (for example, `GitHub`)
     image
7. Now see this form:
     image
   * I have set the `<desc>` to "Here's a GitHub example, just to show the issue" so that we can find this later
   * The heading shows "Preparation (2)" instead of `GitHub` as we named it before
       image
8. Click **Save**.
9. Navigate back to "Preparation" (https://nbmmnb5523-xml-editor.test.specifysystems.org/specify/resources/view-set/2/Preparation)
10. See that `GitHub` exists as a form
     image
11. Click on `GitHub`, see this error:
   > Visual editor is not available for this resource
   image

12. Error seen. Now to find out why that is...

# Actual Issue
After following the steps above, instead of creating a `<view>` and `<viewdef>` pair that are correctly connected, two disparate pieces were made incorrectly.

Switching to the 'XML Editor' view instead of the visual editor, we can debug this.

> For future reference: [fish_views_bug.xml.zip](https://github.com/specify/specify7/files/13033510/fish_views_bug.xml.zip)

## Incorrectly created `view`

We can see that the `<view>` for `GitHub` was created, but with a major flaw:

```xml
<view name="GitHub" class="edu.ku.brc.specify.datamodel.Preparation" busrules="edu.ku.brc.specify.datamodel.busrules.PreparationBusRules">
	<desc>Subform within the Preparation form.</desc>
	<altviews>
		<altview name="Preparation Table View" viewdef="Preparation Table (2)" mode="view"/>
		<altview name="Preparation Table Edit" viewdef="Preparation Table (2)" mode="edit"/>
		<altview name="Preparation View" viewdef="Preparation (2)" mode="view"/>
		<altview name="Preparation Edit" viewdef="Preparation (2)" mode="edit" default="true"/>
	</altviews>
</view>
```

## Missing `viewdef`s

The `viewdef`s being referenced are incorrect. There are no `viewdef` sections named `Preparation Table (2)` or `Preparation (2)` respectively! **This is not good.**

Remember my comment earlier?
>I have set the `<desc>` to "Here's a GitHub example, just to show the issue" so that we can find this later

This allowed me to search and find that there was _indeed_ a `viewdef` created, but it was not created properly. Here's a snippet of that `viewdef`:

```xml
<viewdef type="form" name="Preparation" class="edu.ku.brc.specify.datamodel.Preparation" gettable="edu.ku.brc.af.ui.forms.DataGetterForObj" settable="edu.ku.brc.af.ui.forms.DataSetterForObj" useresourcelabels="true">
	<desc>Here's a GitHub example, just to show the issue.</desc>
	<enableRules/>
	<columnDef>100px,2px,133px,5px,91px,2px,210px,5px,87px,2px,123px,2px,70px,p:g</columnDef>
	<columnDef os="lnx">115px,2px,150px,5px,110px,2px,223px,5px,100px,2px,126px,2px,75px,p:g</columnDef>
	<columnDef os="mac">130px,2px,180px,5px,130px,2px,255px,5px,100px,2px,174px,2px,70px,p:g</columnDef>
	<columnDef os="exp">p,2px,p:g(2),5px:g,p,2px,p:g,5px:g,p,2px,p:g,10px,p:g</columnDef>
	<rowDef auto="true" cell="p" sep="2dlu"/>
```

Notice that the name is not `GitHub`, rather it is `Preparation`, which is problematic for two reasons:
1. There is already a `viewdef` with the name `Preparation` existing in this form definition file
2. We lost the `viewdef` we added since it doesn't have the identifier we intended to assign to it

Now to fix this`viewdef` problem, we need to do some find & replace. In the `viewdef` section, we need to change it from this:
```xml
<viewdef type="form" name="Preparation"...
```
to this:
```xml
<viewdef type="form" name="GitHub"...
```

---

## Not creating the `{Name} Table` `viewdef`

There is also not a `Preparation Table (2)` `viewdef`, just the one that already exists for Preparation:
See the default one here:

```xml
		<viewdef type="formtable" name="Preparation Table" class="edu.ku.brc.specify.datamodel.Preparation" gettable="edu.ku.brc.af.ui.forms.DataGetterForObj" settable="edu.ku.brc.af.ui.forms.DataSetterForObj">
			<desc>Preparation subform table for Collection Object form.</desc>
			<definition>Preparation</definition>
		</viewdef>
```

We would need to copy this and modify it as well if we want to reference it. This should be created automatically, **and it is not.**

**Here's what it should look like:**

```xml
		<viewdef type="formtable" name="GitHub Table" class="edu.ku.brc.specify.datamodel.Preparation" gettable="edu.ku.brc.af.ui.forms.DataGetterForObj" settable="edu.ku.brc.af.ui.forms.DataSetterForObj">
			<desc>Preparation subform table for Collection Object form.</desc>
			<definition>GitHub</definition>
		</viewdef>
```

## How to fix the `view`
 
In the `view` section that was created, we need to change it from this:
```xml
<view name="GitHub" class="edu.ku.brc.specify.datamodel.Preparation" busrules="edu.ku.brc.specify.datamodel.busrules.PreparationBusRules">
	<desc>Subform within the Preparation form.</desc>
	<altviews>
		<altview name="Preparation Table View" viewdef="Preparation Table (2)" mode="view"/>
		<altview name="Preparation Table Edit" viewdef="Preparation Table (2)" mode="edit"/>
		<altview name="Preparation View" viewdef="Preparation (2)" mode="view"/>
		<altview name="Preparation Edit" viewdef="Preparation (2)" mode="edit" default="true"/>
	</altviews>
</view>
```
to this:
```xml
<view name="GitHub" class="edu.ku.brc.specify.datamodel.Preparation" busrules="edu.ku.brc.specify.datamodel.busrules.PreparationBusRules">
	<desc>Subform within the Preparation form.</desc>
	<altviews>
		<altview name="Preparation Table View" viewdef="GitHub Table" mode="view"/>
		<altview name="Preparation Table Edit" viewdef="GitHub Table" mode="edit"/>
		<altview name="Preparation View" viewdef="GitHub" mode="view"/>
		<altview name="Preparation Edit" viewdef="GitHub" mode="edit" default="true"/>
	</altviews>
</view>
```

Replacing the `Preparation Table View` and `Edit` `viewdef` with "GitHub Table" (form fixed in the (Not creating the `{Name} Table` `viewdef`) section above) and the `Preparation View` and `Edit` `viewdef` with "GitHub", also fixed above.

**Video:**
https://github.com/specify/specify7/assets/37256050/26f28214-4523-45f8-a28f-d8f27ee7e4f7

**Expected behavior**
This should be magical!

Met with @CarolineDenis and demonstrated this, unfortunately did not record.

CarolineDenis:
Fixes #4134

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
+++ b/specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
@@ -266,6 +266,40 @@ function createViewFromTemplate(
   const nameMapper = Object.fromEntries(
     originalNames.map((name) => [name, getUniqueDefinitionName(name, viewSets)])
   );
+
+  const uniqueName = getUniqueDefinitionName(name, viewSets);
+
+  const updatedAltViews = view.altViews.altViews.map((view) => ({
+    ...view,
+    viewDef: view.viewDef ? `${view.viewDef} ${uniqueName}` : uniqueName,
+    name: view.name ? `${view.name} ${uniqueName}` : uniqueName,
+  }));
+
+  const updatedView = {
+    ...view,
+    name: uniqueName,
+    altViews: { ...view.altViews, altViews: updatedAltViews },
+  };
+
+  const updatedViewDefinitions = viewDefinitions.map((viewDef) => ({
+    ...viewDef,
+    name: uniqueName,
+    raw: {
+      ...viewDef.raw,
+      attributes: {
+        ...viewDef.raw.attributes,
+        name: `${viewDef.raw.attributes.name} ${uniqueName}`,
+      },
+    },
+  }));
+
+  console.log(
+    'updatedView',
+    updatedView,
+    'updatedViewDefinitions',
+    updatedViewDefinitions
+  );
+
   return {
     ...viewSets,
     views: [
@@ -274,27 +308,15 @@ function createViewFromTemplate(
         ...view,
         name,
         altViews: {
-          ...view.altViews,
-          altViews: view.altViews.altViews.map((altView) => ({
+          ...updatedView.altViews,
+          altViews: updatedView.altViews.altViews.map((altView) => ({
             ...altView,
-            viewDef:
-              typeof altView.viewDef === 'string'
-                ? nameMapper[altView.viewDef] ?? altView.viewDef
-                : altView.viewDef,
+            viewDef: uniqueName,
           })),
         },
       },
     ],
-    viewDefs: [
-      ...viewSets.viewDefs,
-      ...viewDefinitions.map((definition) => ({
-        ...definition,
-        name:
-          typeof definition.name === 'string'
-            ? nameMapper[definition.name] ?? definition.name
-            : definition.name,
-      })),
-    ],
+    viewDefs: [...viewSets.viewDefs, ...updatedViewDefinitions],
   };
 }
 


CarolineDenis:
work in progress

maxxxxxdlp:
The tests for this PR are failing because they need to be updated (see https://github.com/specify/specify7/actions/runs/6616192263/job/17969977482)

@CarolineDenis can you update the tests?
@melton-jason or @realVinayak could you look at why github reported that tests completed successfully despite the fact that they failed?

could be due to it running out of memory: "Reached heap limit Allocation failed - JavaScript heap out of memory"

still, super concerning that we might be shipping bugs to production that could have been caught by tests

specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
```diff
@@ -263,38 +263,56 @@ function createViewFromTemplate(
   const originalNames = filterArray(
     view.altViews.altViews.map(({ viewDef }) => viewDef)
   );
-  const nameMapper = Object.fromEntries(
-    originalNames.map((name) => [name, getUniqueDefinitionName(name, viewSets)])
-  );
+
+  const uniqueName = getUniqueDefinitionName(name, viewSets);
```
maxxxxxdlp, line 267: the name coming from the user was supposed to only be used for the `view` name (as that is what they would specify in the `<subview view="some name here">`
all the other places specify should generate a name of it's own, without relying on what the user provided

maxxxxxdlp (changes requested):
There are problems with the new implementation - mainly, the name provided by the user shouldn't be used everwhere in my opinion.
Instead, it should just take previous name, make it unique, and use that

Can you check if the issues in https://github.com/specify/specify7/issues/4134 are caused by some subtle bugs in here, or does it really require rewriting this logic like you done in this pr?

If you want, I can look at the code deeper and spend time debugging this to see how it's supposed to work

realVinayak:
@maxxxxxdlp opened a new issue for github incorrectly setting status code (https://github.com/specify/specify7/issues/4159)

specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
```diff
@@ -263,38 +263,56 @@ function createViewFromTemplate(
...
+                  ...subChild,
+                  string: `${viewDef.raw.attributes.name} ${uniqueName}`,
+                };
+              }
+              return subChild;
+            }),
+          };
+        }
+        return child;
+      }),
```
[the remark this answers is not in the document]
CarolineDenis: It's the best way I found with typescript

specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
```diff
@@ -263,38 +263,56 @@ function createViewFromTemplate(
...
-  const nameMapper = Object.fromEntries(
-    originalNames.map((name) => [name, getUniqueDefinitionName(name, viewSets)])
-  );
+
+  const uniqueName = getUniqueDefinitionName(name, viewSets);
+
+  const updatedAltViews = view.altViews.altViews.map((altView, index) => ({
+    ...altView,
+    viewDef: `${originalNames[index]} ${uniqueName}` as LocalizedString,
+    name: `${originalNames[index]} ${uniqueName}` as LocalizedString,
```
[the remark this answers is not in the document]
CarolineDenis: yes it was just to show the links at first, once the behavior was good the idea was to just keep unique name

specifyweb/frontend/js_src/lib/components/FormEditor/__tests__/createView.test.ts
```diff
@@ -489,7 +489,7 @@ test('Add new view definition based on existing', () =>
         name: 'CollectionObjectAttachment Table (2)',
         raw: `<viewdef>
   <desc>ObjectAttachment grid view.</desc>
-  <definition>ObjectAttachment Form</definition>
+  <definition>CollectionObjectAttachment Table (2)</definition>
```
maxxxxxdlp, line 492: hmm, this change doesn't look right
see https://sp7demofish.specifycloud.org/context/view.json?name=CollectionObjectAttachment - it is set to "ObjectAttachment Form" in there at the moment

maxxxxxdlp: note that the attachment table may be special.

could you check on some attachment, and non-attachment form that has formtable to see what the expected value for `<definition>` is?
then compare what `createViewFromTemplate` would output before this change and after the change and see which is correct

specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
```diff
@@ -263,38 +263,68 @@ function createViewFromTemplate(
...
+    altViews: { ...view.altViews, altViews: updatedAltViews },
+  };
+
+  const updatedViewDefinitions = viewDefinitions.map((definition) => ({
+    ...definition,
+    name: definition.name ? nameMapper[definition.name] : definition.name!,
+    raw: {
+      ...definition.raw,
+      attributes: {
+        ...definition.raw.attributes,
```
maxxxxxdlp, line 297: was the bug caused by the fact that the `name` from line 298 was used instead of the name from line 293?
do you need to specify the `name` in both places or only one of the two?

note that `raw` key exists here as a way to defer parsing the viewdef (so that all viewdefs are not parsed when opening a view set file - viewdef is then only parsed when it's actually opened for editing)

see how raw XML string is gotten back from it here:
https://github.com/specify/specify7/blob/0cbeff429e84f653e83d0ef49ab5e9675d2c4660/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx#L198-L213
and see how attributes are cleared away in there (so that they are not editable in the UI)

but then see how the attributes are added back in here after user made edits to the view def itself:
https://github.com/specify/specify7/blob/0cbeff429e84f653e83d0ef49ab5e9675d2c4660/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx#L159-L170

I think the actual bug causing https://github.com/specify/specify7/issues/4134 might be that the raw.attributes.name is overriding the name
at the same time, I belive raw.attributes is not needed and should probably be set to empty object to avoid such issues (Because it is empty in all the other places in this file. e.x in https://github.com/specify/specify7/blob/0cbeff429e84f653e83d0ef49ab5e9675d2c4660/specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts#L132-L141).

Thus, could you do the following changes:
1. set attributes to empty object here: https://github.com/specify/specify7/blob/0cbeff429e84f653e83d0ef49ab5e9675d2c4660/specifyweb/frontend/js_src/lib/components/FormEditor/spec.ts#L275
2. that would remove the need to set it to empty object here: https://github.com/specify/specify7/blob/0cbeff429e84f653e83d0ef49ab5e9675d2c4660/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx#L198-L213
3. and remove the need to set it back here: https://github.com/specify/specify7/blob/0cbeff429e84f653e83d0ef49ab5e9675d2c4660/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx#L159-L170
4. and remove the need to set raw.attributes in this file (createView.ts)

specifyweb/frontend/js_src/lib/components/FormEditor/__tests__/createView.test.ts
maxxxxxdlp: okay. I think I know what's happening now.

The way `createViewFromTemplate` worked before wasn't correct - I forgot to make the viewdef name in the `<definition>` unique

And the way it works after you change isn't quite correct either - if you look at the viewdef after your code changes:
```js
        name: 'CollectionObjectAttachment Table (2)',
        raw: `<viewdef>
  <desc>ObjectAttachment grid view.</desc>
  <definition>CollectionObjectAttachment Table (2)</definition>
```
it is saying that when the form finds viewdef named "CollectionObjectAttachment Table (2)" (coming from the name attribute), it should use the "CollectionObjectAttachment Table (2)" viewdef (coming from the definition) - which would create a cycle

instead, it should be taking the `<definition>` value that was there before, and making that unique
thus, you need to change `definition.name` in:
https://github.com/specify/specify7/pull/4144/files#diff-95191ed3f14da84d43cc55789dfd7fb11c121532e0fc5615b08b1ede1e58f61cR312-R314
to `subChild.string`. i.e so that it looks like this:
```ts
                 string: nameMapper[subChild.string as LocalizedString] ??  subChild.string,
```

maxxxxxdlp (changes requested):
thanks for looking into this code and trying to figure it out!

I think you got this bug almost fully figured out now

specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
```diff
@@ -263,38 +263,68 @@ function createViewFromTemplate(
...
+  const updatedAltViews = view.altViews.altViews.map((altView) => ({
+    ...altView,
+    viewDef:
+      typeof altView.viewDef === 'string'
+        ? nameMapper[altView.viewDef] ?? altView.viewDef
+        : altView.viewDef,
+    name:
+      typeof altView.viewDef === 'string'
+        ? nameMapper[altView.viewDef] ?? altView.viewDef
+        : altView.viewDef!,
```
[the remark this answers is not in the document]
CarolineDenis: but wouldn't hurt to use the same unique name to identify the right section in the xml file?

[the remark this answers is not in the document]
maxxxxxdlp: just extra needless code. doesn't matter much so pick either way

specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
```diff
@@ -263,38 +263,63 @@ function createViewFromTemplate(
...
-          })),
-        },
-      },
-    ],
-    viewDefs: [
-      ...viewSets.viewDefs,
-      ...viewDefinitions.map((definition) => ({
-        ...definition,
+
+  const uniqueName = getUniqueDefinitionName(name, viewSets);
```
maxxxxxdlp, line 271: read line 18 in this file - the name would already be unique (see this component - https://github.com/specify/specify7/blob/a66e7a8681d2833757ad33dc502d636944be7d09/specifyweb/frontend/js_src/lib/components/FormEditor/Create.tsx#L201#)

feel free to copy-paste that comment onto line 256 if that would be helpful for the future

specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
```diff
@@ -263,38 +263,63 @@ function createViewFromTemplate(
...
+      typeof altView.viewDef === 'string'
+        ? nameMapper[altView.viewDef] ?? altView.viewDef
+        : altView.viewDef,
+  }));
+
+  const updatedView = {
+    ...view,
+    name: uniqueName,
+    altViews: { ...view.altViews, altViews: updatedAltViews },
+  };
```
maxxxxxdlp, line 285: nothing to change here, but just a FYI on why I often create giant pyramids of nested objects rather than break them down into separate variables:
- if I need readability, separate variables are usually better
- but, by nesting it all together in one large object, I get much better autocomplete and typescript error messages that highlight the exact place that caused the error (because typescript knows what is the expected type - in your case, you could make a typo on line 273-279, but the error would be highlighted all the way on line 321)
- if I want best of both worlds, I have to explicitly specify the variable type when creating a variable (on line 273 and 281 and 287) - that's what you get automatically when you define the object in the function `return` - typescript knows what the correct type should be

so it's a trade off. just one more thing to think about

cc @realVinayak @melton-jason

specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
maxxxxxdlp: because your first code snippet is not correct. To make attributes empty, you should have had something like this:
```ts
    raw: syncer<SimpleXmlNode, XmlNode>(
      (node) => ({
        ...defined(getOriginalSyncerInput(node), ''),
        attributes: {},
      }),
      toSimpleXmlNode
    ),
```

In your case, you first call defined() (which makes sure input is never undefined, and if it happens to be undefined, throw an error), and then use `??`, which only makes sense if you expect the value to the left to be undefined - but by calling `defined()` you ensure it's never undefined, so the code after `??` is never executed

can you try again with this change please?

(and if you get tired of this, I can take over when you want)

maxxxxxdlp (changes requested):
if you run the tests for this PR locally, are they still failing, or are they failing only on the github action?

CarolineDenis:
> if you run the tests for this PR locally, are they still failing, or are they failing only on the github action?

they pass, it's just on GitHub

specifyweb/frontend/js_src/lib/components/FormEditor/createView.ts
maxxxxxdlp: I pushed a commit to this PR that implements the changes I suggested in https://github.com/specify/specify7/pull/4144/files#r1372511231 to simplify code a bit and prevent the bug with name being overridden from happening in the future.

In the process I also remembered why the contents of the `<definition>` tag wasn't replaced - the thing is that it shouldn't always be replaced. If `<definition>` is equal to `ObjectAttachment Form`, and your current view name is `CollectionObjectAttachment`, then even after cloning, the `<definition>` should still point to `ObjectAttachment Form`.

But if `<definition>` points to `CollectionObjectAttachment Form`, then it needs to be replaced with `CollectionObjectAttachment Form (2)` - so that is the 2nd bug in here - I also pushed a commit with the fix -please take a look and test the new behavior

maxxxxxdlp (approved):

grantfitzsimmons:
> There are problems with the new implementation - mainly, the name provided by the user shouldn't be used everwhere in my opinion. Instead, it should just take previous name, make it unique, and use that

I disagree with this. I wrote up a _huge_ reply and GitHub lost it (😢), so I will keep this short.

The user sees the `<viewdef>` name when editing in the visual editor, and right now it's totally automatically made (`Preparation (2)`, `Preparation (3)`, etcetera). This is confusing, since the user assigned it a name, and now for some reason this is the title showing at the top of the editor.

The user also has to manually enter the `<viewdef>` name when assigning an alternative `<definition>`, which is an example of a place where the user needs to type in the name of another `<viewdef>`.

If we can make it unique (which @CarolineDenis already is), we should allow the user to name the `<viewdef>` and `<view>`. I don't think these should be auto-generated until we have a complete user interface for customizing forms without ever needing to edit XML.

CarolineDenis:
current behavior: 
Screenshot 2023-10-31 at 1 27 14 PM

CarolineDenis:
Suggestion after today's meeting: 
All <viewdef> names should be generated with the view name the user is defining.

CarolineDenis:
`<view name="testIssue1" class="edu.ku.brc.specify.datamodel.Accession" busrules="edu.ku.brc.specify.datamodel.busrules.AccessionBusRules">
<desc>The Accession form as seen in the Deaccession form.</desc>
			<altviews>
				<altview name="AccessionDeaccession View" viewdef="AccessionDeaccession" mode="view"/>
				<altview name="AccessionDeaccession Table View" viewdef="AccessionDeaccession Table" mode="view" default="true"/>
			</altviews>
		</view>`
		
		`<viewdef type="form" name="AccessionDeaccession" class="edu.ku.brc.specify.datamodel.Accession" gettable="edu.ku.brc.af.ui.forms.DataGetterForObj" settable="edu.ku.brc.af.ui.forms.DataSetterForObj">
			<desc>The Accession subform for the deaccession form.</desc>
			<enableRules/>
			<columnDef>p,0px,190px,5px,130px,2px,141px,5px,110px,2px,140px,0px,15px,p:g</columnDef>
			<columnDef os="lnx">135px,2px,200px,5px,135px,2px,156px,5px,125px,2px,150px,0px,15px,p:g</columnDef>
			<columnDef os="mac">130px,2px,220px,5px,153px,2px,189px,5px,133px,2px,189px,0px,15px,p:g</columnDef>
			<columnDef os="exp">p,0px,p:g,5px:g,p,2px,p:g(2),5px:g,p,2px,p:g(2),0px,p,p:g</columnDef>
			<rowDef auto="true" cell="p" sep="1px"/>
			<rows>
				<row>
					<cell type="field" id="1" name="accessionNumber" uitype="formattedtext"/>
					<cell type="label" labelfor="3"/>
					<cell type="field" id="3" name="type" uitype="combobox"/>
					<cell type="field" id="2" name="status" uitype="combobox"/>
				</row>
				<row>
					<cell type="field" id="5" name="dateAccessioned" uitype="formattedtext" uifieldformatter="Date" default="today"/>
				</row>
			</rows>
		</viewdef>`
		
		`<viewdef type="formtable" name="AccessionDeaccession Table" class="edu.ku.brc.specify.datamodel.Accession" gettable="edu.ku.brc.af.ui.forms.DataGetterForObj" settable="edu.ku.brc.af.ui.forms.DataSetterForObj">
			<desc>The Accession subform for the deaccession form.</desc>
			<definition>AccessionDeaccession</definition>
		</viewdef>`
		
		**Should be:** 
		`<view name="testIssue1" class="edu.ku.brc.specify.datamodel.Accession" busrules="edu.ku.brc.specify.datamodel.busrules.AccessionBusRules">
<desc>The Accession form as seen in the Deaccession form.</desc>
			<altviews>
				<altview name="AccessionDeaccession View" viewdef="testIssue1" mode="view"/>
				<altview name="AccessionDeaccession Table View" viewdef="testIssue1n Table" mode="view" default="true"/>
			</altviews>
		</view>`
		
		`<viewdef type="form" name="testIssue1" class="edu.ku.brc.specify.datamodel.Accession" gettable="edu.ku.brc.af.ui.forms.DataGetterForObj" settable="edu.ku.brc.af.ui.forms.DataSetterForObj">
			<desc>The Accession subform for the deaccession form.</desc>
			<enableRules/>
			<columnDef>p,0px,190px,5px,130px,2px,141px,5px,110px,2px,140px,0px,15px,p:g</columnDef>
			<columnDef os="lnx">135px,2px,200px,5px,135px,2px,156px,5px,125px,2px,150px,0px,15px,p:g</columnDef>
			<columnDef os="mac">130px,2px,220px,5px,153px,2px,189px,5px,133px,2px,189px,0px,15px,p:g</columnDef>
			<columnDef os="exp">p,0px,p:g,5px:g,p,2px,p:g(2),5px:g,p,2px,p:g(2),0px,p,p:g</columnDef>
			<rowDef auto="true" cell="p" sep="1px"/>
			<rows>
				<row>
					<cell type="field" id="1" name="accessionNumber" uitype="formattedtext"/>
					<cell type="label" labelfor="3"/>
					<cell type="field" id="3" name="type" uitype="combobox"/>
					<cell type="field" id="2" name="status" uitype="combobox"/>
				</row>
				<row>
					<cell type="field" id="5" name="dateAccessioned" uitype="formattedtext" uifieldformatter="Date" default="today"/>
				</row>
			</rows>
		</viewdef>`
		
		`<viewdef type="formtable" name="testIssue1 Table" class="edu.ku.brc.specify.datamodel.Accession" gettable="edu.ku.brc.af.ui.forms.DataGetterForObj" settable="edu.ku.brc.af.ui.forms.DataSetterForObj">
			<desc>The Accession subform for the deaccession form.</desc>
			<definition>testIssue1</definition>
		</viewdef>`

CarolineDenis:
would be

maxxxxxdlp:
> The user sees the `<viewdef>` name when editing in the visual editor, and right now it's totally automatically made (Preparation (2), Preparation (3), etcetera). This is confusing, since the user assigned it a name, and now for some reason this is the title showing at the top of the editor.

That seems like a bug of title coming from a wrong place - needs to be fixed

> The user also has to manually enter the `<viewdef>` name when assigning an alternative `<definition>`, which is an example of a place where the user needs to type in the name of another `<viewdef>`.

I was hoping that the user never has to touch `<definition>` anymore or even know that they existed - it's an ugly artifact of sp6. See explanation in https://github.com/specify/specify7/issues/4179#issuecomment-1788320467

> If we can make it unique (which @CarolineDenis already is), we should allow the user to name the <viewdef> and <view>. I don't think these should be auto-generated until we have a complete user interface for customizing forms without ever needing to edit XML.

If you are saying there is still a valid use case for people to edit the viewset file directly often rather than the individual view definition, then sure. But I was under impression (or merely delusionally hoping) that even the current implementation of the UI removed the need for editing the entire viewset file - and the next gen UI will remove to even edit the view def XML

> current behavior: 
> Screenshot 2023-10-31 at 1 27 14 PM

with the exception of not being as user friendly, it appears that this behavior is fully correct now and resolves all the bugs pointed out in https://github.com/specify/specify7/issues/4134. is that the case?

---

TLDR: I am ok to with making it work the way it looks in https://github.com/specify/specify7/pull/4144#issuecomment-1788026975 if we agree that there is still a need to edit the viewset XML. I was hoping that wasn't the case

CarolineDenis:
"That seems like a bug of title coming from a wrong place - needs to be fixed"

The "title name" is made from viewdefiniton.name and since SP generate that name base on the view and not the user input we have the title showing the viewdef name and not the view name

specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
```diff
@@ -83,7 +83,7 @@ export function FormEditorWrapper(): JSX.Element {
           className={`${className.headerPrimary} flex items-center gap-2 text-xl`}
         >
           <TableIcon label name={table.name} />
-          {viewDefinition.name}
+          {view.name}
```
maxxxxxdlp, line 86: good change. sorry for the oversight

specifyweb/frontend/js_src/lib/components/FormEditor/spec.ts
```diff
@@ -272,7 +272,10 @@ const viewDefSpec = f.store(() =>
      * formDefinitionSpec() later on
      */
     raw: syncer<SimpleXmlNode, XmlNode>(
-      (node) => defined(getOriginalSyncerInput(node), ''),
+      (node) => ({
+        ...defined(getOriginalSyncerInput(node), ''),
```
melton-jason, line 276: If `getOriginalSyncerInput(node)` is undefined, would this not throw an Error with no message? 

Consider the definition of `defined`: 
https://github.com/specify/specify7/blob/48c1f00d576a03e25927847723c17be340944f80/specifyweb/frontend/js_src/lib/utils/types.ts#L37-L41

`'' ?? 'other string'` will return `''`, and `''` is probably not the most immediately helpful error message.  

Of course, ignore this if this is intentional.

melton-jason (approved):
There is a lot of discussion here, so grasping the _full_ context of the changes is a little difficult without a heavy time investment. 

That being said, from what I have gathered thus far, the changes here seem to address the underlying issues, and the typescript is valid. 

The remaining concerns, it seems, stem from the `<definition>` attribute and that a direct modification of the viewset file may still be necessary.
How pressing is addressing this issue? 

Given that this Pull Request has resolved the issues presented in #4134, I would be under the impression of not withholding its merge before a satisfactory solution to the `<definition>` problem is found.

pashiav (approved):
Looks good! No error and has correct title.

https://github.com/specify/specify7/assets/118556884/4f675e16-9d7e-4181-86a7-3af0110189a8

PR #4119, opened 2023-10-16
Fix warnings blocking save

realVinayak:

PR #4155, opened 2023-10-24
Allow several children for xml node when condition present

### LINKED ISSUE #3641: Conditional forms are not working
This could be a case of misconfiguration, but I am trying to create an example geology form using conditional forms. See the snippet of XML below (this is on `localhost`, so I apologize for not having a URL):

https://github.com/specify/specify7/assets/37256050/71abc9a1-7c28-4153-8779-68fa730faa07

```xml
                <row>
                <cell type="panel" coldef="100px,2px,76px" colspan="6">
                  <rows>
                    <!-- This view definition will be displayed in Specify 6, or if all other conditions don't match -->
                    <row>
                      <cell type="label" label="Default view definition." />
                    </row>
                  </rows>
                  <rows condition="text1=1">
                    <row>
                      <cell type="label" label="You chose 1!" />
                    </row>
                  </rows>
                  <rows condition="text1=Mineral" coldef="13px,14px,15px">
                    <row>
                      <cell type="label" label="You chose Mineral!" />
                    </row>
                  </rows>
                </cell>
                </row>
```

I have a pick list of options in `text1` on the Collection Object, with a Title of `Mineral` and Value of `1`. I wanted to see which value would be used by the conditional form system, but neither have any effect. 

**Version:**
`v7.8.12`, macOS, Chrome

**Here's the full file:** [Geology Views.xml.zip](https://github.com/specify/specify7/files/11779687/Geology.Views.xml.zip)

CarolineDenis:
Fixes #3641

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Syncer/syncers.ts
+++ b/specifyweb/frontend/js_src/lib/components/Syncer/syncers.ts
@@ -238,14 +238,23 @@ export const syncers = {
         const currentChildren =
           children[tagName] ?? children[tagName.toLowerCase()] ?? [];
         const child = currentChildren[0];
+
+        let hasCondition = false
+        currentChildren.forEach((child) => {
+            if (child.attributes.condition !== undefined && child.attributes.condition.length > 0) {
+              hasCondition = true;
+            }
+          });
+
         if (child === undefined && mode === 'required')
           console.error(`Unable to find <${tagName} /> child`);
 
         pushContext({ type: 'Child', tagName });
 
-        if (currentChildren.length > 1)
+        if (currentChildren.length > 1 && !hasCondition)
           console.warn(`Expected to find at most one <${tagName} /> child`);
         return child;
+
       },
       (child) => ({
         type: 'SimpleXmlNode',
--- a/specifyweb/frontend/js_src/lib/components/Syncer/syncers.ts
+++ b/specifyweb/frontend/js_src/lib/components/Syncer/syncers.ts
@@ -239,12 +239,15 @@ export const syncers = {
           children[tagName] ?? children[tagName.toLowerCase()] ?? [];
         const child = currentChildren[0];
 
-        let hasCondition = false
+        let hasCondition = false;
         currentChildren.forEach((child) => {
-            if (child.attributes.condition !== undefined && child.attributes.condition.length > 0) {
-              hasCondition = true;
-            }
-          });
+          if (
+            child.attributes.condition !== undefined &&
+            child.attributes.condition.length > 0
+          ) {
+            hasCondition = true;
+          }
+        });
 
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/Syncer/syncers.ts
```diff
@@ -238,12 +238,23 @@ export const syncers = {
...
+
+        let hasCondition = false;
+        currentChildren.forEach((child) => {
+          if (
+            child.attributes.condition !== undefined &&
+            child.attributes.condition.length > 0
+          ) {
+            hasCondition = true;
+          }
+        });
```
maxxxxxdlp, line 250: refactor to use .some
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

please look over js array methods and apply them more - they will make you feel more powerful!

maxxxxxdlp, line 250: this code assumes specific structure from the input code
(it assumes that the child has a condition attribute and assigns some semantic meaning to it)
this function is used in many places - you can't just change it like this - it will affect all the other places where it's used

Instead, if you need to support multiple children, then modify the rule for the form definition to call syncer.xmlChildren instead of syncer.xmlChild

maxxxxxdlp (changes requested):

maxxxxxdlp:
@CarolineDenis am I missing something, or is all that this pr does is disable a warning?
can you clarify how did that warning cause this feature to not work?
I would have though the warning being emitted or not would have no impact on the behavior

maxxxxxdlp:
This PR is linked to the issue titled "Conditional forms are not working"
Are conditional forms not working just because of a validation warning?
And is the XML spec for the conditional form incorrect? If so, it needs to be corrected (the spec, not the definition of a syncer function. see https://github.com/specify/specify7/pull/4155#discussion_r1371002502)

But if the issue is resolved by https://github.com/specify/specify7/pull/4119 then seems like this PR can be closed unmerged

CarolineDenis:
> This PR is linked to the issue titled "Conditional forms are not working" Are conditional forms not working just because of a validation warning? And is the XML spec for the conditional form incorrect? If so, it needs to be corrected (the spec, not the definition of a syncer function. see [#4155 (comment)](https://github.com/specify/specify7/pull/4155#discussion_r1371002502))
> 
> But if the issue is resolved by #4119 then seems like this PR can be closed unmerged

Yes it is working :+1:

maxxxxxdlp:
ok, then let's have someone else independently test it (may @grantfitzsimmons or other product person) and close this PR

grantfitzsimmons:
Looks like that has been tested. Are we safe to close this PR? @CarolineDenis

PR #4158, opened 2023-10-24
Do not show warning for hashtable

### LINKED ISSUE #3708: `xml-editor`: Unknown table: Hashtable
image

From `herb_rbge`'s forms. Not sure if this is unknown due to an oversight.

```xml
        <viewdef
            type="form"
            name="Collection Object Search"
            class="java.util.Hashtable"
            gettable="edu.ku.brc.af.ui.forms.DataGetterForHashMap"
            settable="edu.ku.brc.af.ui.forms.DataSetterForHashMap">
            
            <desc><![CDATA[The Collection Object form.]]></desc>
            <enableRules/>
            
            <columnDef>120px,2px,p,2px,120px,2px,p,2px,60px,2px,p,p:g</columnDef>
            <rowDef auto="true" cell="p" sep="2dlu"/>
            
            <rows>
                <row>
                    <cell type="field" id="1" name="catalogNumber" uitype="text"/>
                    <cell type="field" id="2" name="collection_collectionName" uitype="text"/>
                </row>
                <row>
                    <cell type="label" labelfor="3"/>
                    <cell type="field" id="3" name="accession_number"  uitype="formattedtext" uifieldformatter="AccessionNumber"/>
                    <cell type="label" labelfor="4"/>
                    <cell type="field" id="4" name="cataloger" uitype="text"/>
                    <cell type="field" id="5" uitype="plugin" name="this" initialize="name=PartialDateUI;df=catalogedDate;tp=catalogedDatePrecision" uifieldformatter="Date"/>
                </row>
            </rows>
        </viewdef>
```

https://herbrbge33123-issue-3657.test.specifysystems.org/specify/resources/view-set/1/

CarolineDenis:
Fixes #3708

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Syncer/syncers.ts
+++ b/specifyweb/frontend/js_src/lib/components/Syncer/syncers.ts
@@ -148,7 +148,11 @@ export const syncers = {
       (className) => {
         const tableName = f.maybe(className, parseJavaClassName);
         const table = getTable(tableName ?? className ?? '');
-        if (table === undefined && tableName !== 'ObjectAttachmentIFace')
+        if (
+          table === undefined &&
+          tableName !== 'ObjectAttachmentIFace' &&
+          tableName !== 'Hashtable'
+        )
           console[strict ? 'error' : 'warn'](
             `Unknown table${
               (className ?? '').length === 0


CarolineDenis:
@maxxxxxdlp 
- Did you had `tableName !== 'ObjectAttachmentIFace` in here: https://github.com/specify/specify7/blob/ef66dde63689fe650611f7be0a875e5a0e4948e6/specifyweb/frontend/js_src/lib/components/Syncer/syncers.ts#L151
because it's a legacy table from 6 not used in 7? 
- Does 'Hashtable' has its place here too? 
- Do you happened to have a list of all the tables that should be included in this check? 
Thank you :)

maxxxxxdlp:
Now that https://github.com/specify/specify7/pull/4119 is fixed, I am not sure this PR is needed (though probably won't cause problems)

> Did you had tableName !== 'ObjectAttachmentIFace in here...because it's a legacy table from 6 not used in 7?

It is not used directly in sp6 or sp7. It's a bit of a fictitious name in sp6 standing for any attachment table (at least it seems to me - don't know how sp6 is handing it). So I wanted to make specify 7 treat it as fictitious table too.

> Does 'Hashtable' has its place here too?

Hashtable is only used in the form definitions for the search forms, isn't it? (at least that's what I seem to remember). The form editor UI does not indent to support editing search forms, at least not in the way search forms are defined right now - instead we would either emphasize using query builder, or create a new simpler UI for configuring search forms (or reuse the form editor UI, but maybe store search forms a bit differently)
The reason for that is that there are A LOT of weirdnesses and special exceptions in how sp6 search forms are defined
- There are a lot of form definition XML attributes that sp6 has just for search forms - those have some meaning in sp6, but not supported in sp7
- Sp6 allows you to reference a -to-many table as if it is a -to-one table when in the search form - sp7 doesn't support that (i.e in sp6, you can say name="addresses.address" in the Agent form definition - try out opening the Agent search form in sp7 and you will see that all the address fields are read-only because sp7 does not support this)
- Many other weird things
That is why I was not looking to support search forms in current form in the form edititor, and instead think of a new solution for them

> Do you happened to have a list of all the tables that should be included in this check?

This should be it. But if you wan to be sure, take a look at this:

https://github.com/specify/specify_tools/tree/main/xml-inspector

It's a small script I developed when reverse engineering how sp6 treats forms when I was writing the syncer rules for form definitions.
Those scripts told me all the attributes ever supported by sp6 and all the possible values for those attributes.
If you want to work on form editor code more in the future, I would say looking at https://github.com/specify/specify_tools/tree/main/xml-inspector code and playing around with it a bit would be helpful

maxxxxxdlp (approved):
see https://github.com/specify/specify7/pull/4158#issuecomment-1778359541

I would not do this change if I were you, but am fine if you want to do it

CarolineDenis:
Fixed by https://github.com/specify/specify7/pull/4119

PR #4163, opened 2023-10-26
Add a pref to have sidebar bg light in light mode

### LINKED ISSUE #4157: Add preference for customizing sidebar color
> Since upgrading to 7.9 I have received several requests/queries about the colour of the sidebar. Several people in the team would like to have the bar light again, rather than dark.
> From looking at the changes it seems that this now not something that can be adjusted and it is dark regardless of the user preference for light/dark theme. Is there any way we can have the side bar a lighter colour again? I’ve had a look in the user preferences, but couldn’t see an option.

**Requested By:** Robyn and others at RBGE

CarolineDenis:
Fixes #4157

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Header/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/index.tsx
@@ -69,6 +69,7 @@ export function Header({
     'lightSideBarBackground'
   );
   const isDarkMode = useDarkMode();
+  const isMenuLight = isSideBarLight === 'matchThemeColor' && !isDarkMode;
   // Top menu is only available as collapsed
   const isCollapsed = rawIsCollapsed || isHorizontal || forceCollapse;
 
@@ -111,7 +112,7 @@ export function Header({
             : 'dark:border-t'
         }
         ${
-          isSideBarLight === 'matchThemeColor' && !isDarkMode
+          isMenuLight
             ? 'bg-gray-100 shadow-md shadow-gray-400'
             : 'border-neutral-700 bg-neutral-800'
         }
--- a/specifyweb/frontend/js_src/lib/components/Header/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/index.tsx
@@ -111,7 +111,7 @@ export function Header({
             : 'dark:border-t'
         }
         ${
-          isSideBarLight && !isDarkMode
+          isSideBarLight === 'matchThemeColor' && !isDarkMode
             ? 'bg-gray-100 shadow-md shadow-gray-400'
             : 'border-neutral-700 bg-neutral-800'
         }
@@ -214,19 +214,11 @@ export function MenuButton({
   const getClassName = (isActive: boolean): string => `
     p-[1.4vh]
     ${
-      isActive && isDarkMode
+      isActive
         ? 'bg-brand-300 text-white'
-        : !isActive && isDarkMode
+        : isDarkMode || isSideBarLight === 'dark'
         ? 'text-white'
-        : isActive && !isDarkMode && isSideBarLight
-        ? 'bg-brand-300 text-white'
-        : !isActive && !isDarkMode && isSideBarLight
-        ? 'text-gray-700'
-        : !isActive && !isDarkMode && !isSideBarLight
-        ? 'text-white'
-        : isActive && !isDarkMode && !isSideBarLight
-        ? 'bg-brand-300 text-white'
-        : 'text-white'
+        : 'text-gray-700'
     }
     ${className.ariaHandled}
     ${extraProps?.className ?? ''}
--- a/specifyweb/frontend/js_src/lib/components/Header/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/index.tsx
@@ -212,12 +212,13 @@ export function MenuButton({
     'lightSideBarBackground'
   );
   const isDarkMode = useDarkMode();
+  const isSideBarDark = isDarkMode || isSideBarLight === 'dark';
   const getClassName = (isActive: boolean): string => `
     p-[1.4vh]
     ${
       isActive
         ? 'bg-brand-300 text-white'
-        : isDarkMode || isSideBarLight === 'dark'
+        : isSideBarDark
         ? 'text-white'
         : 'text-gray-700'
     }
--- a/specifyweb/frontend/js_src/lib/components/Header/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/index.tsx
@@ -20,6 +20,7 @@ import { schema } from '../DataModel/schema';
 import { userInformation } from '../InitialContext/userInformation';
 import { titleDelay, titlePosition } from '../Molecules/Tooltips';
 import { Notifications } from '../Notifications/Notifications';
+import { useDarkMode } from '../Preferences/Hooks';
 import { userPreferences } from '../Preferences/userPreferences';
 import { ActiveLink } from '../Router/ActiveLink';
 import { Logo } from './Logo';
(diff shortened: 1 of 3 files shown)


specifyweb/frontend/js_src/lib/components/Header/index.tsx
```diff
@@ -194,9 +205,29 @@ export function MenuButton({
...
+        : isActive && !isDarkMode && isSideBarLight
+        ? 'bg-brand-300 text-white'
+        : !isActive && !isDarkMode && isSideBarLight
+        ? 'text-gray-700'
+        : !isActive && !isDarkMode && !isSideBarLight
+        ? 'text-white'
+        : isActive && !isDarkMode && !isSideBarLight
+        ? 'bg-brand-300 text-white'
+        : 'text-white'
+    }
```
maxxxxxdlp, line 230: this seems very overcomplicated.
your code is equivalent to this:
```ts
  isActive
    ? 'bg-brand-300 text-white'
    : isDarkMode || !isSideBarLight
    ? 'text-white'
    : 'text-gray-700';
```

(you can use a website like https://www.boolean-algebra.com/ or https://web.stanford.edu/class/cs103/tools/truth-table-tool/ to see how to simplify boolean expressions)

specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
```diff
@@ -215,6 +215,13 @@ export const userPreferenceDefinitions = {
             renderer: ColorPickerPreferenceItem,
             container: 'label',
           }),
+          lightSideBarBackground: defineItem({
+            title: preferencesText.lightSideBarBackground(),
+            requiresReload: false,
+            visible: isLightMode,
+            defaultValue: false,
+            type: 'java.lang.Boolean',
+          }),
```
maxxxxxdlp, line 224: instead of making this a boolean, could we have it be one of the two values:
- "Dark"
- "Match theme color"

Benefits:
- If it is not a boolean, but a multiple choice, it's much easier to extend it in the future with additional types
- You don't have to handle the case when theme color is dark but side bar color is light (unless we want that to be a feature)

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Header/index.tsx
CarolineDenis: so so much better

specifyweb/frontend/js_src/lib/components/Header/index.tsx
```diff
@@ -104,6 +110,11 @@ export function Header({
             ? 'dark:border-l'
             : 'dark:border-t'
         }
+        ${
+          isSideBarLight === 'matchThemeColor' && !isDarkMode
```
maxxxxxdlp, line 114: could you create a variable like `isMenuDark` OR `isMenuLight` and use it in here?
that way the code would be more-self descriptive (takes a moment to realize that `isSideBarLight === 'matchThemeColor' && !isDarkMode` basically means light mode)

specifyweb/frontend/js_src/lib/components/Header/index.tsx
```diff
@@ -104,6 +110,11 @@ export function Header({
...
+            ? 'bg-gray-100 shadow-md shadow-gray-400'
```
maxxxxxdlp, line 115: (optional) could you double check that these light mode styles are the ones that we had for menu in light mode before it got permanently switched to dark mode?
if they are not the same, make sure that was intentional. if they are the same, then great

specifyweb/frontend/js_src/lib/components/Header/index.tsx
```diff
@@ -194,9 +205,21 @@ export function MenuButton({
   readonly props?: TagProps<'a'> & TagProps<'button'>;
 }): JSX.Element | null {
   const [position] = userPreferences.use('header', 'appearance', 'position');
+  const [isSideBarLight] = userPreferences.use(
+    'general',
+    'appearance',
+    'lightSideBarBackground'
+  );
+  const isDarkMode = useDarkMode();
```
maxxxxxdlp, line 213: (optional) create a hook that returns `readonly [isDarkMode: boolean, isMenuDarkMode: boolean]` and use it in two places

maxxxxxdlp (approved):
Logic looks good, though left some refactoring suggestions

specifyweb/frontend/js_src/lib/components/Header/index.tsx
CarolineDenis: I took the style from how it was before :+1:

specifyweb/frontend/js_src/lib/components/Header/index.tsx
```diff
@@ -194,9 +206,21 @@ export function MenuButton({
...
   const getClassName = (isActive: boolean): string => `
     p-[1.4vh]
-    ${isActive ? 'bg-brand-300 !text-white' : 'text-white'}
+    ${
+      isActive
+        ? 'bg-brand-300 text-white'
+        : isDarkMode || isSideBarLight === 'dark'
```
maxxxxxdlp, line 220: just like above you created `isMenuLight` variable, could you do the same here?

(optional) (or create a hook like I suggested in https://github.com/specify/specify7/pull/4163/files#r1375147892 to have it consistent, and not duplicate)

CarolineDenis: this one was clear for me without having to create a variable...

maxxxxxdlp: oh, I didn't notice that the expressions are slightly different. sorry about that.
then yeah, separate variable is not necessary, but having it does not hurt

acwhite211 (approved):

carlosmbe (dismissed):
Has a bug where the preference disappears in Dark Mode

https://github.com/specify/specify7/assets/53784701/d4c21e23-200b-49ca-8ecc-818e5148778d

I also feel like having the option in between two colour pickers breaks the design language and flow. Could we put it on top or below the other colour pickers?

Now that I think about it. You can probably make it a check box under the Rounded Corners option and call it "Always Dark Navigation Bar" or something like that.

I think it would look pretty good under Rounder Corners since it's a pretty long piece of text. Then again, only one way to find out.

Other than that, it seems functionally fine. Nice work. SLAYYYYYY. :D 

Screenshot 2023-11-02 at 3 53 30 PM

CarolineDenis:
> Has a bug where the preference disappears in Dark Mode
> 
>  Screen.Recording.2023-11-02.at.3.51.42.PM.mov 
> 
> I also feel like having the option in between two colour pickers breaks the design language and flow. Could we put it on top or below the other colour pickers?
> 
> Now that I think about it. You can probably make it a check box under the Rounded Corners option and call it "Always Dark Navigation Bar" or something like that.
> 
> I think it would look pretty good under Rounder Corners since it's a pretty long piece of text. Then again, only one way to find out.
> 
> Other than that, it seems functionally fine. Nice work. SLAYYYYYY. :D
> 
> Screenshot 2023-11-02 at 3 53 30 PM

@carlosmbe, thanks for the testing! 
Yes it's normal that it disappear in dark mode as it is a pref for light mode only. 
I can put it above the background color, no prob. 
It's a drop down menu to have the possibility to extend the list later on.

maxxxxxdlp (approved):

pashiav (approved):
Looks good!

https://github.com/specify/specify7/assets/118556884/c68f08f7-b00e-4998-83c7-1c4e6545e37d

CarolineDenis:
@pashiav thank you for catching that!

specifysoftware:

PR #4164, opened 2023-10-26
Display uniqueness requirement for name and disable save if not met

### LINKED ISSUE #4150: `xml-editor`: Formatter `Name` uniqueness requirement not shown in the dialog
**Describe the bug**
The warning for name uniqueness appears behind the dialog, obscuring it from the user.
It also doesn't prevent saving...

Encountered on:
https://kufish51623-xml-editor.test.specifysystems.org/specify/resources/app-resource/9/formatter/Accession/0

**To Reproduce**
1. Create a new formatter
2. Change the `Name` to match an existing formatter
3. See warning `Formatter names must be unique` under the current dialog

**Expected behavior**
The warning should appear in the dialog so you can resolve it before closing it.

**Screenshots**
https://github.com/specify/specify7/assets/37256050/3d9dc5b3-d7f1-49c0-95ed-d7f39537db7e

CarolineDenis:
Fixes #4150

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
@@ -48,7 +48,7 @@ export function XmlEditorShell<
   const item = items[index];
   const uniqueNames = f.unique(items.map(({ name }) => name));
   const hasDuplicates = uniqueNames.length !== items.length;
-  const { validationRef } = useValidation(
+  const { validationRef, setValidation } = useValidation(
     hasDuplicates ? resourcesText.duplicateFormatters() : ''
   );
 
@@ -103,7 +103,15 @@ export function XmlEditorShell<
             isReadOnly={isReadOnly}
             required
             value={item.name}
-            onValueChange={(name): void => setItem({ ...item, name })}
+            onValueChange={(name): void => {
+              setItem({ ...item, name });
+              setValidation(
+                uniqueNames.includes(name)
+                  ? resourcesText.duplicateFormatters()
+                  : '',
+                'focus'
+              );
+            }}
           />
         </Label.Block>
         {children({ items: allItems, item: getSet })}
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
@@ -48,8 +48,9 @@ export function XmlEditorShell<
   const item = items[index];
   const uniqueNames = f.unique(items.map(({ name }) => name));
   const hasDuplicates = uniqueNames.length !== items.length;
-  const { validationRef, setValidation } = useValidation(
-    hasDuplicates ? resourcesText.duplicateFormatters() : ''
+  const { validationRef } = useValidation(
+    hasDuplicates ? resourcesText.duplicateFormatters() : '',
+    'focus'
   );
 
   const setItem = (newItem: ITEM): void =>
@@ -101,15 +102,7 @@ export function XmlEditorShell<
             isReadOnly={isReadOnly}
             required
             value={item.name}
-            onValueChange={(name): void => {
-              setItem({ ...item, name });
-              setValidation(
-                uniqueNames.includes(name)
-                  ? resourcesText.duplicateFormatters()
-                  : '',
-                'focus'
-              );
-            }}
+            onValueChange={(name): void => setItem({ ...item, name })}
           />
         </Label.Block>
         {children({ items: allItems, item: getSet })}
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
@@ -60,6 +60,7 @@ export function XmlEditorShell<
   const isReadOnly = React.useContext(ReadOnlyContext);
   const navigate = useNavigate();
   const handleClose = (): void => navigate(resolveRelative('../'));
+
   const id = useId('item');
   return item === undefined ? (
     <Dialog
@@ -92,7 +93,7 @@ export function XmlEditorShell<
           item.name,
       })}
       icon={icons.variable}
-      onClose={handleClose}
+      onClose={undefined}
     >
       <Form id={id('form')} onSubmit={handleClose}>
         <Label.Block>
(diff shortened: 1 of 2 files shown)


CarolineDenis:
Do we want to keep the warning also on the list behind the dialog or now that we have it in the form we're good? 
@grantfitzsimmons

specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
```diff
@@ -45,6 +46,8 @@ export function XmlEditorShell<
   const [items, setItems] = allItems;
   const index = f.parseInt(rawIndex)!;
   const item = items[index];
+  const uniqueNames = f.unique(items.map(({ name }) => name));
+  const hasDuplicates = uniqueNames.length !== items.length;
```
maxxxxxdlp, line 50: smart

maxxxxxdlp (changes requested):
this is a good first try, but can you try to improve it:
could you try using the useValidation hook to create a validation error on the `name` field when there are duplicates?

benefits:
- the error would be connected to the field that caused it - making it even easier to understand how to fix it
- if user tries to submit the form, the invalid field will get focus again and error message will be displayed shown again automatically
- better screen reader/accessibility support

--

as far as having the validation behind this dialog, some thoughts:
- take a look if something in the code would go terribly wrong if name is not unique (i.e terribly wrong means a crash, or loss of data; not terribly wrong is formatter that is used is not the one user was expecting)
- we are adding the UI for editing this, but in the past people might have hand-edited these files, so there could be mistakes that we have to catch an present as soon as possible
- on the other hand, it would be annoying to have to fix all of the mistakes before you can continue on with your work just because some test formatter you are no longer using has a duplicate name
- having the same check in both places is redundant (more code, and also a bit distracting when the error appears in both places)

given that UI is not letting user create new mistakes, that alone is good enough and forcing them to fix previous mistakes may be too forcefull.

grantfitzsimmons:
> Do we want to keep the warning also on the list behind the dialog or now that we have it in the form we're good?

If the error appears in the dialog, we can remove it from behind. Our goal is to prevent the user from closing the dialog until the issue is resolved, right? In that case, we might want to not let the user close the dialog until it's resolved.

image

specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
```diff
@@ -75,7 +81,9 @@ export function XmlEditorShell<
             {commonText.delete()}
           </Button.Danger>
           <span className="-ml-2 flex-1" />
-          <Submit.Info form={id('form')}>{commonText.close()}</Submit.Info>
+          <Submit.Info disabled={hasDuplicates} form={id('form')}>
```
maxxxxxdlp, line 84: please remove `disabled={hasDuplicates}`
now that you set validation for Input.Text, making sure the submit button is NOT disabled is important.
Because, if some field on the form is invalid, then clicking the submit button will focus the invalid field - but not if the submit button is disabled

specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
```diff
@@ -91,10 +99,19 @@ export function XmlEditorShell<
...
             value={item.name}
-            onValueChange={(name): void => setItem({ ...item, name })}
+            onValueChange={(name): void => {
+              setItem({ ...item, name });
+              setValidation(
+                uniqueNames.includes(name)
+                  ? resourcesText.duplicateFormatters()
+                  : '',
+                'focus'
+              );
```
maxxxxxdlp, line 113: is this needed? 

isn't this code you added higher up already doing this?
```ts
hasDuplicates ? resourcesText.duplicateFormatters() : ''
```

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
CarolineDenis: yes you're write I might have to added it because I had my 'save' button disabled

maxxxxxdlp (commented):
looks great!

isn't it cool how few lines of code it takes to build a fully-accessible and user-friendly validation, that is also quite robust?

grantfitzsimmons (changes requested):
You can bypass this by clicking on the space outside of the dialog

https://github.com/specify/specify7/assets/37256050/9085a4f1-9552-4de6-837b-f1beecb3a9cc

maxxxxxdlp (approved):
@grantfitzsimmons 

> You can bypass this by clicking on the space outside of the dialog

I think forbidding closing the dialog is worse UX than accidentally creating two formatters with the same name and thus seeing wrong one being used on the form

What if user wants to briefly look at something behind the dialog? (i.e some other formatter, or a notification they have just received)

The role of validation in this dialog in my opinion is not bulletproof data integrity, but guiding the user towards right choices (guiding, not holding under the gun)

---

but I am not a user so feel free to pick either way

CarolineDenis:
> @grantfitzsimmons
> 
> > You can bypass this by clicking on the space outside of the dialog
> 
> I think forbidding closing the dialog is worse UX than accidentally creating two formatters with the same name and thus seeing wrong one being used on the form
> 
> What if user wants to briefly look at something behind the dialog? (i.e some other formatter, or a notification they have just received)
> 
> The role of validation in this dialog in my opinion is not bulletproof data integrity, but guiding the user towards right choices (guiding, not holding under the gun)
> 
> but I am not a user so feel free to pick either way

The user could drag the dialog on the side otherwise?

grantfitzsimmons:
@maxxxxxdlp If we disallow saving with duplicate view names, that would be acceptable. Otherwise, it's preferable to require them to resolve the issue before dismissing it.

CarolineDenis:
> @maxxxxxdlp If we disallow saving with duplicate view names, that would be acceptable. Otherwise, it's preferable to require them to resolve the issue before dismissing it.

I suggest we keep it like this then and if the user needs to see what's behind, he would move the dialog

maxxxxxdlp:
sounds good, just pointing out that the ability to bypass validation might be a feature and not a bug

emenslin (approved):
This works and you can no longer close the popup if it has the same name either by using the close button or by clicking out of it. Can still move the popup though if the user did need to see something behind it.

acwhite211 (approved):

PR #4171, opened 2023-10-27
Display summary of warnings at opening in App Resources

### LINKED ISSUE #4168: `xml-editor`: See a summary of warnings
When there are a lot of warnings in an XML file, I need to now scroll and find them one-by-one (after #4119 was merged). There is no place where they are all displayed anymore.

> I would very much like to see a summary of these warnings.

Maybe this means they show at the bottom again? Maybe a button to see them all in a dialog? It would be nice to see.

CarolineDenis:
Fixes #4168


CarolineDenis:
TODO: 
- add icon for warning in button
- add icon for errors in button 
- add title in button 
- assemble 'line from to' string in one localized string 
- add number of warnings in button

specifyweb/frontend/js_src/lib/components/AppResources/TabDefinitions.tsx
```diff
@@ -97,29 +105,40 @@ const generateEditor = (xmlSpec: (() => BaseSpec<SimpleXmlNode>) | undefined) =>
...
+            value={data ?? ''}
+            /*
+             * FEATURE: provide supported attributes for autocomplete
+             *   https://codemirror.net/examples/autocompletion/
+             *   https://github.com/codemirror/lang-xml#api-reference
+             */
+            onChange={handleChange}
+          />
+        </div>
+        {showWarning &&
```
maxxxxxdlp, line 134: I could see issues with this:
- when you are editing, warnings appear or disappear, causing the editor container height to change - does that happen? is that an issue?
- if there are a lot of warnings, this `<p>` becomes very tall and makes the codemirror container too small - you need to add some sort of max-height in here
- this has poor accessibility. better would be to put this in a ul with each error being a li and add `aria-live="polite"` to `ul` (read about what that does)

specifyweb/frontend/js_src/lib/components/AppResources/TabDefinitions.tsx
```diff
@@ -97,29 +105,40 @@ const generateEditor = (xmlSpec: (() => BaseSpec<SimpleXmlNode>) | undefined) =>
...
+          warnings.map((warning, index) => (
```
maxxxxxdlp, line 135: see optional suggestion in https://github.com/specify/specify7/issues/4168#issuecomment-1783666322

maxxxxxdlp (changes requested):
see https://github.com/specify/specify7/issues/4168#issuecomment-1783668588

CarolineDenis:
Decisions: 
- Display warning lintPanel at loading 
- Add instructions in documentation for short cut shift + cmd + m to display warning panel later on

specifyweb/frontend/js_src/lib/components/AppResources/TabDefinitions.tsx
```diff
@@ -96,6 +97,24 @@ const generateEditor = (xmlSpec: (() => BaseSpec<SimpleXmlNode>) | undefined) =>
       [stateRestored]
     );
     const isReadOnly = React.useContext(ReadOnlyContext);
+
+    const displayWarningPanel = React.useMemo(
+      () => () => {
+        f.maybe(codeMirrorRef.current?.view, openLintPanel);
```
maxxxxxdlp, line 103: can you double-check that this behaves correctly:
- when there are no errors/warnings (Does not do anything and does not steal focus from the editor)
- when there are errors/warnings

specifyweb/frontend/js_src/lib/components/AppResources/TabDefinitions.tsx
```diff
@@ -96,6 +97,24 @@ const generateEditor = (xmlSpec: (() => BaseSpec<SimpleXmlNode>) | undefined) =>
...
+      },
+      [codeMirrorRef]
+    );
+
+    React.useEffect(() => {
+      const timeoutId = setTimeout(() => {
```
maxxxxxdlp, line 109: is setTimeout necessary? try to avoid it whenever possible as setTimeout is a hacky solution
i.e could you instead display the warnings panel when the first lint error occurs? (which may happen on page load, or later caused by user edit)

maxxxxxdlp (changes requested):
Almost there

specifyweb/frontend/js_src/lib/components/AppResources/TabDefinitions.tsx
CarolineDenis: it does steal focus unfortunately... and if no warnings/errors it's displayed with "No Diagnostic"

specifyweb/frontend/js_src/lib/components/AppResources/TabDefinitions.tsx
CarolineDenis: @maxxxxxdlp,  I change the logic in my new commit but I cannot figure out why now the diagnostic is always empty even if the `warnings` array shows the proper warnings...

specifyweb/frontend/js_src/lib/components/AppResources/EditorComponents.tsx
```diff
@@ -201,15 +201,26 @@ export function useCodeMirrorExtensions(
       [appResource.specifyTable]
     )
   );
+
+  const [warnings, setWarnings] = React.useState<RA<Diagnostic>>([]);
+
   React.useEffect(() => {
-    const handleLinted = (results: RA<Diagnostic>): void =>
+    const handleLinted = (results: RA<Diagnostic>): void => {
+      setWarnings(
```
maxxxxxdlp, line 209: not sure why, but the issue is caused by this setWarnings call (or rather, the re-render it causes)
possible, calling setWarnings triggers re-render at an inopportune time, and causes AppResourceTextEditor to hold a stale reference to code mirror, leading to no lint errors being visible (and if you look in the editor state, the are 0 errors there)

note that we are not using Codemirror directly, but using a react wrapper library - maybe it has a bug (though I can reproduce the issue on the latest version too)

I found a solution that works - calling openLintPanel directly in this file, rather than setting state - see the commit I pushed

Can you test if it behaves the same way in production (i.e react rendering twice may be causing this to happen)

maxxxxxdlp (approved):
FYI: we already had the code in here for displaying the diagnostics panel if user tries to save and there are errors: https://github.com/specify/specify7/blob/68523859b4667c69ae9e5ffab475ab9605f8d06b/specifyweb/frontend/js_src/lib/components/AppResources/TabDefinitions.tsx#L77-L80

specifyweb/frontend/js_src/lib/components/AppResources/TabDefinitions.tsx
```diff
@@ -96,6 +101,20 @@ const generateEditor = (xmlSpec: (() => BaseSpec<SimpleXmlNode>) | undefined) =>
       [stateRestored]
     );
     const isReadOnly = React.useContext(ReadOnlyContext);
+
+    const displayWarningPanel = React.useMemo(
+      () => () => {
```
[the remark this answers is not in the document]
CarolineDenis: @maxxxxxdlp , because we are memoizing a function and not a value?

specifyweb/frontend/js_src/lib/components/AppResources/EditorComponents.tsx
CarolineDenis: @maxxxxxdlp, it does not work. Neither on local. Why did you completely take off the warnings? I than cannot have that value in mu conditional in my useEffect to call displayWarningPanel

CarolineDenis: the new push has the same behavior has how it is in xml-editor. But, we want the panel to open when opening the editor if warnings are present. Hence the call to openLintPanel in TabDefinition.

maxxxxxdlp (approved):

maxxxxxdlp:
ups, I just noticed that I forgot to publish my comments yesterday. probably got distracted. sorry about that
github should be sending out reminders about things like that

specifyweb/frontend/js_src/lib/components/AppResources/EditorComponents.tsx
CarolineDenis: works perfectly. Thank you!!!!

pashiav (approved):
https://github.com/specify/specify7/assets/118556884/b65ad020-3be1-4a09-b8a2-8e805062f21f

Looks good! Displays warnings and navigates to them on click.

melton-jason (approved):
Timing issues strike again! 
Utilizing `setTimeout` is a little unfortunate, because as often stated it is a bit 'hacky' if there is a separate underlying issue causing the problem, but sometimes it is necessary. 
Perhaps the _real_ problem is the asynchronous nature of the code/browser! :laughing: 

- - - 

> queueMicrotask is a bit like setTimeout(()=>..., 0);, but a bit better in this case and less crude
> see more docs about it - https://developer.mozilla.org/en-US/docs/Web/API/queueMicrotask

https://github.com/specify/specify7/pull/4171#discussion_r1379530746

I was not aware of queueMicrotask! 
We might be able to utilize/leverage such a function in other places in the codebase. 
Though I do wonder why it was not working in this case...

maxxxxxdlp:
> Though I do wonder why it was not working in this case...

I spent 20 minutes reading the source code of the CodeMirror 6, but couldn't figure it out. Yes, this is unfortunate, but don't know if investing more time into this is waranted

PR #4175, opened 2023-10-30
Warn many-to-many relationship not being supported instead of error

### LINKED ISSUE #4167: `xml-editor`: Projects or many-to-many items on the form should be a warning, not an error
image

Projects not being supported should not prevent saving.

CarolineDenis:
Fixes #4167


specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
```diff
@@ -361,7 +361,7 @@ const subViewSpec = (
           }
           if (field?.type === 'many-to-many') {
             // ResourceApi does not support .rget() on a many-to-many
-            console.error('Many-to-many relationships are not supported');
+            console.warn('Many-to-many relationships are not supported');
             return undefined;
```
maxxxxxdlp, line 365: by returning undefined here, the field will be lost if user makes edits to the form and tries to save again
it wasn't an issue before because console.error prevent saving, but it is an issue now.
thus, one solution would be:
- remove the return statement (so that the `return fields;` statement below it is used)
- test the <Subview> component to make sure it behaves like expected (depends on what we expect - either doesn't render anything, or console error, but we probably don't want it to crash)

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
CarolineDenis: @maxxxxxdlp, what do you mean the field will be lost? 
If I add 
`<cell type="subview" viewname="Project" id="proj" name="projects" initialize="btn=true"/>`
I will have a warning saying many-to-many red are not supported. I can save. 
When reopening the form I can make any other changes and save. 
When reopening my changes are visible and the line with the many-to-many warning is still present

maxxxxxdlp: Looked into why it is still there. Here is the answer:
The Form Editor at the moment uses the viewdef as is, not the one it gets after running it though the syncer:
https://github.com/specify/specify7/blob/d0dcc1bac05cee5589d7666f44fe5497c8dcd550/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx#L157-L158

The syncer spec is used in here:

https://github.com/specify/specify7/blob/d0dcc1bac05cee5589d7666f44fe5497c8dcd550/specifyweb/frontend/js_src/lib/components/AppResources/EditorComponents.tsx#L220

Thus, you still get the validation errors, but the value is not **YET** affected by the syncer

However, the visual form editor v2 (the one that will not include an XML editor, but will include UI for all things) will be using syncer result, thus the way you handle the many-to-many here will make an effect (but we might forget by then that this change here could cause issues).

Thus my recommendation would be to remove `return undefined;` - thus will return the field
Since the result of the syncer is not used yet this won't have any impact yet, but will prevent a possible bug in the future

---

Also a clarification:
Result of syncer on individual viewdef is not used, but the view/altview portion is fully using syncer result, not just the validation part, since we have a UI that takes care of creating altviews and etc.

The way viewdef currently avoids syncer is a bit hacky. Besides the fact that we don't yet have UI for this, the other reason "raw" XML is used rather than syncer parse is because I was afraid of the performance impact of parsing all viewdefs when opening an XML editor for a large viewset file. As a consequence, when editing viewset file, viewdef errors are not shown (besides XML syntax errors) - not a big deal since you see errors when editing individual viewdef

https://github.com/specify/specify7/blob/d0dcc1bac05cee5589d7666f44fe5497c8dcd550/specifyweb/frontend/js_src/lib/components/FormEditor/spec.ts#L268-L277

maxxxxxdlp (approved):

pashiav (approved):
Looks good! Warning instead of error.
issue-4167

PR #4183, opened 2023-11-01
Add viewsetName to category when not a collection

### LINKED ISSUE #4176: `xml-editor`: Forms from different disciplines shown under the same header when creating new form based on existing one
Screenshot 2023-10-31 at 9 11 06 AM

When creating a new form based on an existing one, all forms are shown under the same discipline even thought they are distinct. This can be seen in the `nbm_mnb` database where the two preparation forms come from Ichthyology and Herpetology, but they are shown under the same header.

CarolineDenis:
Fixes #4176

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormEditor/fetchAllViews.ts
+++ b/specifyweb/frontend/js_src/lib/components/FormEditor/fetchAllViews.ts
@@ -78,7 +78,8 @@ const augmentDatabaseView = (
       ? userInformation.availableCollections.find(
           ({ id }) => id === view.collectionId
         )?.collectionName
-      : undefined) ?? camelToHuman(`${view.viewsetLevel} ${view.viewsetName}`),
+      : undefined) ??
+    camelToHuman(`${view.viewsetLevel} (${view.viewsetName})`),
   editUrl:
     view.viewsetId === null
       ? undefined
--- a/specifyweb/frontend/js_src/lib/components/FormEditor/fetchAllViews.ts
+++ b/specifyweb/frontend/js_src/lib/components/FormEditor/fetchAllViews.ts
@@ -78,7 +78,7 @@ const augmentDatabaseView = (
       ? userInformation.availableCollections.find(
           ({ id }) => id === view.collectionId
         )?.collectionName
-      : undefined) ?? camelToHuman(view.viewsetLevel),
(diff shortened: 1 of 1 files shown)


CarolineDenis:
Before
Screenshot 2023-11-01 at 12 20 31 PM

After
Screenshot 2023-11-01 at 12 20 21 PM

specifyweb/frontend/js_src/lib/components/FormEditor/fetchAllViews.ts
```diff
@@ -78,7 +78,7 @@ const augmentDatabaseView = (
       ? userInformation.availableCollections.find(
           ({ id }) => id === view.collectionId
         )?.collectionName
-      : undefined) ?? camelToHuman(view.viewsetLevel),
+      : undefined) ?? camelToHuman(`${view.viewsetLevel} ${view.viewsetName}`),
```
maxxxxxdlp, line 81: (optional) could we put some punctuation between level and name?

i.e
`${view.viewsetLevel} > ${view.viewsetName}`
or the nicer
`${view.viewsetName} (${view.viewsetLevel})`

maxxxxxdlp (approved):

grantfitzsimmons (changes requested):
image

Instead of "Discipline Fish Views", can we make it accurate to the discipline the form definition comes from? 

In this case, it should say "Ichthyology Discipline" where `Ichthyology` comes from the `discipline.Name` field so that it is based on the actual name assigned by the user. 

Some people use "Botany" as the discipline template but manage pathogens or lichens in it instead. This would make the name more intuitive for users.

CarolineDenis:
DECISION: 
Only display "Herpetology View" and loose Discipline.

grantfitzsimmons (approved):
💯 🚀

maxxxxxdlp:
before merging this, @CarolineDenis can you test how dropping the viewsetLevel affect view sets from other levels?
if you are not sure, then proceed to only drop viewsetLevel when viewsetLevel is discipline

CarolineDenis:
@maxxxxxdlp , @grantfitzsimmons , note that the discipline has been added as a title of the list in https://github.com/specify/specify7/pull/4190

pashiav (approved):
Looks good!
issue-4176

PR #4184, opened 2023-11-01
Allow to delete only one view

### LINKED ISSUE #4180: `xml-editor`: Deleting a definition deletes **all** `<viewdef>`s!
Deleting a definition deletes **all** `<viewdef>`s!!!

https://github.com/specify/specify7/assets/37256050/705de67e-6b06-4ca6-af79-5b229be8a29f

CarolineDenis:
Fixes #4180

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { useOutletContext } from 'react-router';
 import { useNavigate, useParams } from 'react-router-dom';
+import type { LocalizedString } from 'typesafe-i18n';
 import _ from 'underscore';
 
 import { useBooleanState } from '../../hooks/useBooleanState';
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
```diff
@@ -113,7 +113,7 @@ export function FormEditorWrapper(): JSX.Element {
               const newViewDefs = viewSets.viewDefs.filter(
                 (viewDefinition) =>
                   currentUsedViewDefinitions.has(viewDefinition.name) &&
-                  !updatedUsedViewDefinitions.has(viewDefinition.name)
+                  updatedUsedViewDefinitions.has(viewDefinition.name)
               );
 
               setViewSets(
```
maxxxxxdlp, line 119: seems like I got confused here - the comment above is correct, but I forgot that .filter() in this case is not deleting view definitions, but deciding which definitions should remain (i.e not get deleted), thus the condition should have been reversed:

> delete all views that were not used before and are used now

turns into

> preserve all views that either weren't used before, or are used now

your code change would make it delete all view definitions that aren't used now OR weren't used before - I would say that's potentially unsafe and maybe unexpected to the user (why deleting this view def caused other view defs to disappear). this is a reasonable behavior for a new system, but here we are adding a UI to a very old system, will all kinds of messy data

for all we know, some views are marked as unused only because the user made a typo in the name somewhere
OR maybe instead of commenting out a view, the user just renamed it to an unused name

whatever it is, we probably don't want to delete all unused, which is what the code after your code change does

maxxxxxdlp (changes requested):

CarolineDenis:
@maxxxxxdlp I made the code simpler, let me know what you think

maxxxxxdlp (changes requested):
while the code is simpler, it is not equivalent.

you are making an assumption that viewDef can only be used by one altView (the altView s belonging to the view you are deleting)
But, we are dealing here with forms that people have been hand editing for a decade - we shouldn't make assumptions if we don't need to. I have no idea what files users might be having (but I do know that many of them contain bugs, typos, and real issues, given by how many of the form-related bug requests on github have been caused by improper XML)

if it were to add a lot of complexity to not make this potentially unsafe assumption, then fine. But in here it's just 15 more lines, so I would say it's better to keep the more comprehensive, safer variant

Thougs?

CarolineDenis:
@maxxxxxdlp
I agree with you. I wrote this code having the current behavior in mind. I didn't take into consideration the manual alterations a user could have been done in the past.

maxxxxxdlp (approved):

melton-jason (approved):
This works!
The cases this does not cover: like having another view/view definition with the same name as the altview of the view/ view definition to be deleted, would already cause problems otherwise.

pashiav (approved):
Looks good! Deletes the correct view and all other views work after deleting one.


https://github.com/specify/specify7/assets/118556884/7061ff89-a8d8-4df3-b200-4b3812d89e3c

PR #4190, opened 2023-11-02
 Organize view form list by discipline

### LINKED ISSUE #4177: `xml-editor`: Creating a new form definition based on a default form is unintuitive
image

I am in a fish collection, I want to create a new form based on the fish collection form definition.
I am greeted with a giant list of all discipline's forms, and it is really unclear which I should choose.

**Expected Behavior**
I should see my discipline's forms show up first (Fish)

Here are my options at this time:
```
Fish
Fish > Fishbase
Fish > Manager > Fish
Fish > Guest > Fish
```
This is very strange, as a user, since I do not have a concept of what these things mean.

Maybe we could represent this under a header for each discipline, shown like this:

> **Ichthyology**
> * Fish Default
> * Manager User
> * Guest User
> * Fishbase
>
> **Herpetology**
> * Herpetology Default
> * Manager User
> * Guest User

CarolineDenis:
Fixes #4177


CarolineDenis:
How it looks now: 
Screenshot 2023-11-02 at 2 10 37 PM

CarolineDenis:
I couldn't find an easy way to do the same for the views coming from 6. I added a collectionId to the view in augmentDiskView() that we could use to call the API discipline url related to that collection id and have the discipline name some how... 
I'm not sure this worst the time as the aim is to get rid of this disk files. 
Screenshot 2023-11-02 at 2 10 42 PM

specifyweb/frontend/js_src/lib/components/FormEditor/index.tsx
```diff
@@ -53,9 +58,12 @@ export function FormEditorWrapper(): JSX.Element {
   );
 
   const [changed, setChanged] = React.useState<ReadonlySet<string>>(new Set());
+  const disciplines =
```
realVinayak, line 61: I added disciplines here because we don't need the rest of content from AppResourcesOutlet. If you need additional context, maybe pass all of them down.

specifyweb/frontend/js_src/lib/components/FormEditor/Create.tsx
```diff
@@ -112,35 +112,56 @@ function ListViews({
   readonly views: AllTableViews['database'] | AllTableViews['disk'];
   readonly onSelect: (view: ViewDefinition) => void;
 }): JSX.Element {
-  const grouped = React.useMemo(
-    () => group(views.map((view) => [view.category, view] as const)),
-    [views]
-  );
+  const grouped = React.useMemo(() => {
+    const byDiscipline = group(
```
maxxxxxdlp, line 116: btw, soon enough it would be safe to use the newly-added Object.groupBy for this rather than our custom group - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy
probably too early to do that today (without transpiling) as safari has no support yet)

specifyweb/frontend/js_src/lib/components/FormEditor/fetchAllViews.ts
```diff
@@ -78,7 +82,14 @@ const augmentDatabaseView = (
...
         )?.collectionName
-      : undefined) ?? camelToHuman(view.viewsetLevel),
+      : undefined) ?? camelToHuman(view.viewsetName),
+  disciplineId: strictParseResourceUrl(
+    defined(
+      userInformation.availableCollections.find(
+        ({ id }) => id === view.collectionId
+      )?.discipline
+    )
+  ).at(1) as number,
```
maxxxxxdlp, line 92: instead of `as number`, could you do `!`?
i.e whenever possible pick the least powerful type assertion so that you are asserting less, and letting TypeScript type-check more

maxxxxxdlp: or you could replace `.at(1)!` with [1].
`.at()` is useful when:
- you need negative indexes `-1` (last item)
- you want typescript to force you to double check that item is not undefined

but in this case, you know that strictParseResourceUrl always returns a tuple of size 2, so getting element at index `[1]` is safe, so you don't need to use .at()

specifyweb/frontend/js_src/lib/components/FormEditor/Create.tsx
```diff
@@ -112,35 +112,56 @@ function ListViews({
...
-                      className={className.dataEntryEdit}
-                      href={view.editUrl}
-                      icon="pencil"
-                      title={commonText.edit()}
-                    />
-                  )}
-                </li>
+              {categories.map(([category, views], index) => (
+                <div className="flex flex-col gap-2" key={index}>
+                  <h4 className={className.headerGray}>{category}</h4>
```
maxxxxxdlp, line 146: this can't be h4 because you used h4 on the previous level - it must be one level higher

maxxxxxdlp, line 146: now that there are so many nested levels, semantically it might be better to represent this using nested ul and li, rather than div+h3,4,5

i.e you have nested lists, not sections with subsections, thus ul+li makes more sense than headers

specifyweb/frontend/js_src/lib/components/FormEditor/fetchAllViews.ts
```diff
@@ -91,6 +102,8 @@ const augmentDiskView = (view: ViewDefinition): PresentableViewDefinition => ({
     typeof view.viewsetFile === 'string'
       ? filePathToHuman(view.viewsetFile)
       : camelToHuman(view.viewsetLevel),
+  disciplineId: undefined,
```
maxxxxxdlp, line 105: I might be misunderstanding, but what is the trouble with getting the disciplineId in here just like you did it in the place above?

i.e
```ts
  disciplineId: strictParseResourceUrl(
    defined(
      userInformation.availableCollections.find(
        ({ id }) => id === view.collectionId
      )?.discipline
    )
  ).at(1) as number
```

> I'm not sure this worst the time as the aim is to get rid of this disk files.

or do you also need to do some react code changes to accommodate this (in addition to the changes you already made)?

specifyweb/frontend/js_src/lib/components/FormEditor/fetchAllViews.ts
```diff
@@ -78,7 +82,14 @@ const augmentDatabaseView = (
       ? userInformation.availableCollections.find(
           ({ id }) => id === view.collectionId
         )?.collectionName
-      : undefined) ?? camelToHuman(view.viewsetLevel),
+      : undefined) ?? camelToHuman(view.viewsetName),
+  disciplineId: strictParseResourceUrl(
+    defined(
+      userInformation.availableCollections.find(
+        ({ id }) => id === view.collectionId
+      )?.discipline
```
maxxxxxdlp, line 90: can you double check the python code to see if it does the check for read-permissions or not?
it is possible that back-end might return a view from a collection the user does not have access to, which would cause this to throw an error

specifyweb/frontend/js_src/lib/components/FormParse/index.ts
```diff
@@ -67,6 +67,7 @@ export type ViewDefinition = {
   readonly viewsetSource: string;
   readonly viewsetId: number | null;
   readonly viewsetFile: string | null;
+  readonly collectionId: number;
```
maxxxxxdlp, line 70: was `collectionid` already part of ViewDefinition but just not included in the type?
otherwise, I don't see the code where you added collectionId to ViewDefinition when fetching from the back-end

specifyweb/frontend/js_src/lib/components/FormEditor/fetchAllViews.ts
```diff
@@ -78,7 +82,14 @@ const augmentDatabaseView = (
       ? userInformation.availableCollections.find(
```
maxxxxxdlp, line 82: githuv doesn't let me leave comments on lines that weren't modified for some weird reason, but regarding lines 77 and 78:
since ViewDefinition now includes collectionId, you don't need `& { readonly collectionId: number }` in here anymore

maxxxxxdlp: and could you also do a quick check to see if we have `& { readonly collectionId: number }` anywhere else where it doesn't need to be anymore?

maxxxxxdlp (approved):
code logic is valid. just leaving some refactoring comments

specifyweb/frontend/js_src/lib/components/FormEditor/Create.tsx
```diff
@@ -132,18 +132,18 @@ function ListViews({
   const { disciplines } = useOutletContext<FormEditorOutlet>();
 
   return (
-    <div className="flex flex-col gap-4">
+    <Ul className="flex flex-col gap-4">
       <h3 className={className.headerPrimary}>{header}</h3>
```
maxxxxxdlp, line 136: this is not a correct markup:. headers can't be inside of `<ul>`

a correct markup may look like this:
```html
<ul>
  <li>
    <h2>Section Title</h2>
    <ul>
      <li>
        <h3>Subsection Title</h3>
        <ul>
          <li>
            <h4>Subsubsection Title</h4>
            <!-- Further content here -->
          </li>
        </ul>
      </li>
    </ul>
  </li>
</ul>
```

note, that you might also want to convert div on line 74 into a `<Ul>`, and then the element on line 135 into a `<li>`

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/FormEditor/Create.tsx
maxxxxxdlp: not fully resolved yet.

see the comment I left above:
> note, that you might also want to convert div on line 74 into a `<Ul>`, and then the element on line 135 into a `<li>`

in current implementation you have two `<ul>`, with only one `<li>` in each - why not just replace the div on line 74 and then you won't have to create two separate `<ul>` (simpler, and better semantics)

maxxxxxdlp (changes requested):
see https://github.com/specify/specify7/pull/4190#discussion_r1390486958

specifyweb/frontend/js_src/lib/components/FormEditor/Create.tsx
maxxxxxdlp: sorry, the latest commit is quite wrong

1. you converted from `div` to `Ul` correctly. BUT:
2. changing all `Ul` to `li` is not correct. you can't have `li` inside of `li` without any `<ul>` in between. Instead of this incorrect change, you should have just removed the `Ul` in `ListViews` as we now have `Ul` in `CreateFormDefinition`, and then move over the class names `Ul` had over to the root `li` in `ListViews`

I made the necessary changes in 6e0b18b

So in summary, the structure of the DOM tree should be something like this:
```html
<ul>
  <li>
    <h3>Some header</h3>
    <ul>
      <li>
        <h4>Some subheader</h4>
         <ul>
          <li>
            <h5>Some sub-sub header</h5>
            ...
          </li>
          <li>
            ...
          </li>
        </ul>
      </li>
      <li>
        ...
      </li>
    </ul>
  </li>
  <li>
    ...
  </li>
  <!-- you can have as many li as you want -->
</ul>
```

(note that Ul is pretty much just ul so you can treat them as the same for purposes of this sample)

(you can also see a similar dom structure in this code sample: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul#ordered_list_inside_unordered_list, except they don't wrap headers in `<h2>`, and use `ol` for nested list)

The code is now semantically correct. Can you double-check that the styling is still correct, and if not adjust the class names as necessary?

also, please tell if I should have explained things in some other way to reduce the confusion in the future

maxxxxxdlp (approved):
see https://github.com/specify/specify7/pull/4190#discussion_r1391822591

grantfitzsimmons (approved):
This looks great! So nice to see, especially on databases with many collections under a discipline.

PR #4192, opened 2023-11-06
Add background color behind query result

### LINKED ISSUE #4170: `xml-editor`:  Background behind the query results is missing
image

This serves as a visual indication that the results are separated from the query. It can already be difficult to tell when one section ends and another begins.

image

CarolineDenis:
Fixes #4170

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Tabs.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Tabs.tsx
@@ -2,7 +2,6 @@ import { Tab } from '@headlessui/react';
 import React from 'react';
 import type { LocalizedString } from 'typesafe-i18n';
 
-import { commonText } from '../../localization/common';
 import { resourcesText } from '../../localization/resources';
 import { f } from '../../utils/functools';
 import type { GetSet, IR, RA, RR } from '../../utils/types';
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
@@ -195,7 +195,7 @@ export function QueryResults(props: Props): JSX.Element {
   );
 
   return (
-    <Container.Base className="w-full bg-[color:var(--form-background)]">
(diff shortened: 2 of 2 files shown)


melton-jason (approved):
Do you know which component's style was overriding `bg-[color:var(--form-background)]` to make the important/specificity `!` Tailwind operator necessary? 

Both production and xml-editor have the same className for `Container.Base`: 

https://github.com/specify/specify7/blob/e630090617870dde0edb3cb76b8b2e117fa42da2/specifyweb/frontend/js_src/lib/components/Atoms/index.tsx#L39

https://github.com/specify/specify7/blob/f4a963b57523a8644895b4304e80771ef9707978/specifyweb/frontend/js_src/lib/components/Atoms/className.ts#L21-L22

And both had the same Container.Base definitions/classNames in this file: 
https://github.com/specify/specify7/blob/e630090617870dde0edb3cb76b8b2e117fa42da2/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx#L198

I would assume the `bg-[color:var(--form-foreground)]` in the definition of `containerBase` in the className.ts is somehow overriding the Query Results in xml-Editor? 

Regardless, this change is probably fine for now. I doubt there would be a "Specificity Clash" in regards to setting the background of the Query Results, though it is still a case to be considered

CarolineDenis:
@melton-jason 
No I couldn't find the exact reason...

pashiav (approved):
Looks good!
issue-4170

maxxxxxdlp:
> No I couldn't find the exact reason...

Screenshot 2023-11-12 at 14 23 56

`<Container.Base>` uses the "--form-foreground" color:

```ts
 const containerBase = `${containerBaseUnstyled} bg-[color:var(--form-foreground)] 
   rounded p-4 shadow-gray-400 shadow-2xl`; 
```

where as the change in this pull request makes it use `--form-background` in this one place (background vs foreground)

@CarolineDenis can you investigate what change caused this issue? could be that I broke something on xml-editor; or maybe this bug was introduced during merging of xml-editor and production

grantfitzsimmons (approved):
Looks and works as expected.

PR #4195, opened 2023-11-06
Display a warning if ther's no available preps

### LINKED ISSUE #4075: Warn when entering prep by Cat Num for unavailable preparations
The dialog's warning for when a catalog number entered is unavailable is based on if there is a Collection Object record for it, and/or if that record has preparation(s) added. 

My request is that we add a warning/notification for catalog numbers entered where the CO does have preparations but they are unavailable to add to a new Disposal/Gift/Loan. This would require looking at the Available column to make sure it is at least 1 to determine eligibility. 

In this video, the dialog removed 0000001 and 0000002 and kept 0000003 and 0000004. 

https://github.com/specify/specify7/assets/135047322/01dace05-2f9c-4149-b94c-e29dfd022fd8

0000001 and 00000002 were removed because there are no records with that catalog number in this database: 

Screenshot 2023-10-06 at 11 16 23 AM

In the following dialog, we find that there still aren't any available preparations to be added to this new loan so the user would have to start over with the cancel button or continue without preparations: 

Screenshot 2023-10-06 at 11 20 22 AM

CarolineDenis:
Fixes #4075


specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
```diff
@@ -86,6 +86,12 @@ export function InteractionDialog({
           readonly entries: RA<PreparationData>;
         }
       >
+    | State<
+        'UnavailableState',
```
maxxxxxdlp, line 90: just a fyi: I used to have the name of each state end with `*State` to clearly differentiate them, but am not doing that anymore as it is redundant pretty often (it's clear that state name is being referred to when you see state.type in the code; adding `*State` is only adding verbosity)

not changes needed here; just a FYI for the future

specifyweb/frontend/js_src/lib/localization/interactions.ts
```diff
@@ -47,6 +47,9 @@ export const interactionsText = createDictionary({
     'ru-ru': 'Не обнаружено препаратов для следующих номеров по каталогу:',
     'uk-ua': 'Не знайдено препаратів для наступних каталожних номерів:',
   },
+  preparationsNotAvailableFor: {
+    'en-us': 'No preparations are available for the following Catalog Numbers:',
```
maxxxxxdlp, line 51: I see that this string is based on the `preparationsNotFoundFor`, so they both have the same issue
- try hard not to use database field names/table names in UI localization as database field names/table are edited by collection managers, so this would create inconsistency/confusion.

Refactor to either pass field name as an argument (see how it's done in other places in the code - you did the refactor for this), or reword the sentence to not have to include the field name (i.e use `...for the following records:`

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/localization/interactions.ts
```diff
@@ -47,6 +47,9 @@ export const interactionsText = createDictionary({
     'ru-ru': 'Не обнаружено препаратов для следующих номеров по каталогу:',
     'uk-ua': 'Не знайдено препаратів для наступних каталожних номерів:',
   },
+  preparationsNotAvailableFor: {
```
maxxxxxdlp, line 50: looks good. can you do the same change for preparationsNotFoundFor?

maxxxxxdlp (approved):

bronwyncombs (changes requested):
Still doesn't catch if the count of available preps is 0. In other words, if the number of unavailable preps is equal to the total count for a given prep under the cat num entered, the user should be made aware that no loan preparations can be added for that catalog number.

https://github.com/specify/specify7/assets/135047322/37ea5f95-046d-402b-9de1-45451821bceb

Edge case consideration:

For catalog numbers with multiple preparations, there could be variation in the availability status per preparation. We could indicate partial unavailability for that cat number in the response and then leave unavailable preps out of subsequent dialog?

CarolineDenis:
Strange because it works on other db 
Screenshot 2023-11-21 at 9 26 09 AM
Screenshot 2023-11-21 at 9 26 22 AM

CarolineDenis:
I see, 
the issue was that if a record didn't have preparations we had the "No preparations were found for the following" and nothing else even if one record would have been a part of "No preparations are available..." 
I change the code to display both warnings if both present or just the relevant one. 
Screenshot 2023-11-21 at 10 05 03 AM
Screenshot 2023-11-21 at 10 05 15 AM
Screenshot 2023-11-21 at 10 05 56 AM

emenslin (approved):
This looks good. I think having the distinction between the preparation being available versus not found will make things more clear to the user.

pashiav (approved):
Lets user know if selected records have no preparations. Looks good! 

https://github.com/specify/specify7/assets/118556884/cb12a1ae-e82f-4891-9ec2-c3997bf28799

PR #4196, opened 2023-11-07
Set new resource to the right scope level

### LINKED ISSUE #3956: Creating a new resource at the 'Global' level sets the user group as 'prefs'
Probably related to #3918, #3828, and #3780

When creating a new app resource at the 'Global' level, it is currently added to the User Group `Prefs` instead of `Common` as it was before!

### Correct Behavior:
`v7.8.0`:
image

### Incorrect Behavior:
`v7.8.13`:
image

This should be resolved and investigated ASAP.

CarolineDenis:
Fixes #3956

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
@@ -15,9 +15,7 @@ import { Ul } from '../Atoms';
 import { Button } from '../Atoms/Button';
 import { Link } from '../Atoms/Link';
 import { addMissingFields } from '../DataModel/addMissingFields';
-import type { SerializedResource } from '../DataModel/helperTypes';
 import { deserializeResource } from '../DataModel/serializers';
-import type { SpAppResourceDir } from '../DataModel/types';
 import { filePathToHuman } from '../FormEditor/fetchAllViews';
 import {
   spAppResourceView,
@@ -255,7 +253,7 @@ function EditAppResource({
   mimeType,
   templateFile,
 }: {
-  readonly directory: SerializedResource<SpAppResourceDir>;
+  readonly directory: ScopedAppResourceDir;
   readonly name: string;
   readonly type: AppResourceType;
   readonly mimeType: string | undefined;
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
@@ -241,6 +241,13 @@ function ViewSetTemplates({
   );
 }
 
+const scopeLevel = {
+  user: 0,
+  collection: 1,
+  discipline: 2,
+  global: 3,
+};
+
 function EditAppResource({
   directory,
   name,
@@ -258,8 +265,7 @@ function EditAppResource({
     () =>
       deserializeResource(
         addMissingFields(type.tableName as 'SpAppResource', {
-          // I don't think this field is used anywhere
-          level: 0,
+          level: scopeLevel[directory.scope as keyof typeof scopeLevel],
           mimeType,
           name: name.trim(),
           specifyUser: userInformation.resource_uri,
--- a/specifyweb/frontend/js_src/lib/components/DataModel/types.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/types.ts
@@ -4956,7 +4956,6 @@ export type SpAppResourceDir = {
     readonly timestampModified: string | null;
     readonly userType: string | null;
     readonly version: number | null;
-    readonly scope: string;
   };
   readonly toOneDependent: RR<never, never>;
   readonly toOneIndependent: {
--- a/specifyweb/frontend/js_src/lib/components/DataModel/types.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/types.ts
@@ -4956,6 +4956,7 @@ export type SpAppResourceDir = {
     readonly timestampModified: string | null;
     readonly userType: string | null;
     readonly version: number | null;
+    readonly scope: string;
   };
   readonly toOneDependent: RR<never, never>;
   readonly toOneIndependent: {
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Tabs.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Tabs.tsx
@@ -2,7 +2,6 @@ import { Tab } from '@headlessui/react';
 import React from 'react';
 import type { LocalizedString } from 'typesafe-i18n';
 
-import { commonText } from '../../localization/common';
 import { resourcesText } from '../../localization/resources';
 import { f } from '../../utils/functools';
 import type { GetSet, IR, RA, RR } from '../../utils/types';


CarolineDenis:
Screenshot 2023-11-07 at 11 08 37 AM
Screenshot 2023-11-07 at 11 08 05 AM

CarolineDenis:
For testers: 
When testing this, test https://github.com/orgs/specify/projects/5/views/2?pane=issue&itemId=36052543 as well. 
Thanks!

specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
```diff
@@ -258,8 +265,7 @@ function EditAppResource({
     () =>
       deserializeResource(
         addMissingFields(type.tableName as 'SpAppResource', {
-          // I don't think this field is used anywhere
-          level: 0,
+          level: scopeLevel[directory.scope as keyof typeof scopeLevel],
```
maxxxxxdlp, line 268: does this place need to be refactored too?
https://github.com/specify/specify7/blob/issue-4170/specifyweb/frontend/js_src/lib/components/AppResources/EditorWrapper.tsx#L62C11-L62C17

specifyweb/frontend/js_src/lib/components/DataModel/types.ts
```diff
@@ -4956,6 +4956,7 @@ export type SpAppResourceDir = {
     readonly timestampModified: string | null;
     readonly userType: string | null;
     readonly version: number | null;
+    readonly scope: string;
```
maxxxxxdlp, line 4959: I don't see the `scope` field in the response from http://localhost/api/specify/spappresourcedir/?limit=0. can you double check why this was added?

maxxxxxdlp: oh I see what you are doing.
the `scope` field is added on the front-end only.
this file defines types for data we receive from and send to back-end only!

what is more, this file is auto-generated - manual changes should be minimized. see:
https://github.com/specify/specify7/blob/issue-4170/specifyweb/frontend/js_src/lib/components/DataModel/types.ts#L1-L24

perhaps you meant to use the ScopedAppResourceDir type instead of modifying this type?

specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
maxxxxxdlp: Ben and I were pretty sure this field wasn't used by sp6 and sp7 anywhere. Can you show what it is used by now?

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
CarolineDenis: it's used to organized the views by scope

specifyweb/frontend/js_src/lib/components/DataModel/types.ts
maxxxxxdlp: sorry for the misunderstanding. let me clarify this further:

'scope' field is not coming from the back-end (database/python)
it is added by the front-end
[types.ts](https://github.com/specify/specify7/blob/issue-4170/specifyweb/frontend/js_src/lib/components/DataModel/types.ts#L1-L24) file is only for fields coming from the back-end
'scope' is added by the front-end
so 'scope' field does not belong in the types.ts file.

specifically, see this part - https://github.com/specify/specify7/blob/6e0b18b3b4612787d1de74f01780bb56dc92fd11/specifyweb/frontend/js_src/lib/components/AppResources/hooks.ts#L41-L45

we fetch the SpAppResourceDir from the back-end
and then we add the scope field.

the types.ts file is responsible for telling TypeScript what to expect to receive from the back-end API (i.e what `fetchCollection`/`fetchResource`/`new table.CollectionObject.Resource({id: someId}).fetch()` would return)
because of the code change you did, typescript now things that `fetchCollection` would include the scope field - IT DOES NOT. The scope field is added later by the front-end.

Does this part make sense?

That is why you should not be modifying the [types.ts](https://github.com/specify/specify7/blob/issue-4170/specifyweb/frontend/js_src/lib/components/DataModel/types.ts#L1-L24) file. It is auto-generated from the database schema. see the comment at the top of that file for more information:

https://github.com/specify/specify7/blob/issue-4170/specifyweb/frontend/js_src/lib/components/DataModel/types.ts#L1-L24

Now, if you check the place on the front-end where the scope field is added (the [useAppResources](https://github.com/specify/specify7/blob/issue-4177/specifyweb/frontend/js_src/lib/components/AppResources/hooks.ts#L34-L66) hook), you will see that the return type of the hook uses this type for directories:

https://github.com/specify/specify7/blob/issue-4177/specifyweb/frontend/js_src/lib/components/AppResources/hooks.ts#L26

See how it uses ScopedAppResourceDir rather than AppResourceDir?

Which is defined as:

https://github.com/specify/specify7/blob/issue-4177/specifyweb/frontend/js_src/lib/components/AppResources/types.tsx#L39C13-L41

SO what you had to do in this case is to change the argument type of the EditAppResource component to expect the ScopedAppResourceDir type rather than AppResourceDir if you need to use the `scope` field (and resolve the upstream typescript errors resulting from that if in places where EditAppResource is called, it receives  AppResourceDir rather than ScopedAppResourceDir)

specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
maxxxxxdlp: Can you please link to where in the code it is used???

maxxxxxdlp (changes requested):
see https://github.com/specify/specify7/pull/4196/files#r1391832335 and https://github.com/specify/specify7/pull/4196/files#r1391832468

specifyweb/frontend/js_src/lib/components/DataModel/types.ts
CarolineDenis: Thank you for details :+1:

specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
CarolineDenis: https://github.com/specify/specify7/blob/7caa59814a7a81564aff541f1be0a29cc95aeecb/specifyweb/frontend/js_src/lib/components/DataModel/scoping.ts#L75

I am actually not 100% sure, I found the solution when using dev tools and seeing that level was always assigned to the same number but determined where it was stored in the 'menu tree'

maxxxxxdlp: I don't see any references to the `level` variable in either of these two files. Your code change is setting the `level` variable. Unless something is looking at the value of that variable, this change is needless.

Can you link to the exact line of code where the value of the `level` variable is read?
If not, can you double check that the bug does occur as soon as you revert this change, and is fixed as soon as you re-apply the change?

maxxxxxdlp: FYI: The location of the app resource in the app resource tree is determined solely by which spAppResourceDir the app resource belongs too. `level` field was added long time ago by sp6, but it's very redundant and so was never used AFAIK.
So when debugging the problem, look at what spAppResourceDir is attached to this app resource, and if it is what you expect

CarolineDenis: there is an issue with global level which I understood wrong. My fixes in this branch bring nothing. I will revert and try from scratch to find the problem...

PR #4198, opened 2023-11-07
Differenciate list of tables for interactions and data entry

### LINKED ISSUE #4197: The list of tables for interactions and for data entry is the same.
Add a new table to the list of table in data entry dialog, see that it will be then added to the interactions list table. 
Same is true when deleting. 
Same is true when starting in interactions.

CarolineDenis:
Fixes #4197


maxxxxxdlp (approved):
good catch!

grantfitzsimmons (approved):
Works as expected!

pashiav (approved):
Looks good! Adding and deleting tables in a list only affects its own list.

https://github.com/specify/specify7/assets/118556884/007c6747-5d2b-4d96-a4fd-4f75b62a7ddf

PR #4199, opened 2023-11-08
Add an option for view button next to queryComboBox

### LINKED ISSUE #3707: Enable adding a View (👁️) button to a query combo box
## Mock-up

image

This is using the existing eye icon and behavior when viewing a record in read-only mode

## Rational

Theresa and I met with Willem of SAIAB after he raised the following concerns:

> I want to raise the issue of over-writing of linked records. If I’m not mistaken, it was in 2009 (Specify 6) that the application stopped making a copy of a linked record if it was edited, and started to allow the record to be edited.
> 
> I have seen some terrible cases of corruption of the taxon and geography trees, and over-writing of agent records. I used to think that this was related to education level or job grade or country development level. But now I think it’s just a human thing.
> 
> I think Specify should revert back to cloning a linked record that gets edited. This was a good idea to preserve data integrity. The claim that it causes duplication is perhaps true, but this duplication is by far the lesser evil.
> 
> I’m not really concerned about my database. I can train our people and do frequent checks. But I think this is a reputational issue for Specify, and I predict that if left unattended it could result in some really bad press down the line. I actually regret not being more assertive when I first detected the change. I think I should have made a stronger case to keep the cloning feature.

Training can not be sufficient for some staff, and explaining how a relational database system works can be tricky.

The implications of editing a linked record are difficult to understand for the layman. 

## Recommended Solution

When clicking on the 👁️ (eye), it should display the linked record in a read-only view. When the user clicks on the link out button next to the record name, it should bring them to the normal form which they can edit.

The idea is that linked records accessed using the 👁️ (eye) icon should not be editable when displayed in a dialog. 

This should be implemented as https://github.com/specify/specify7/pull/3669 is, where a new attribute is added to the form definition/configuration that allows an administrator to replace the plus button with an eye icon. Users should still be able to clone, add, or search even if they cannot edit the linked record from the form.

CarolineDenis:
Fixes #3707

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
@@ -226,7 +226,7 @@ export function QueryComboBox({
       : undefined;
 
   const loading = React.useContext(LoadingContext);
-  const handleOpenRelated = (): void =>
+  const handleOpenRelated = (isReadOnly: boolean): void =>
     state.type === 'ViewResourceState' || state.type === 'AccessDeniedState'
       ? setState({ type: 'MainState' })
       : typeof relatedCollectionId === 'number' &&
@@ -241,7 +241,7 @@ export function QueryComboBox({
             })
           )
         )
-      : setState({ type: 'ViewResourceState', isReadOnly: false });
+      : setState({ type: 'ViewResourceState', isReadOnly });
 
   const subViewRelationship = React.useContext(SubViewContext)?.relationship;
   const pendingValueRef = React.useRef('');
@@ -355,7 +355,7 @@ export function QueryComboBox({
 
   const isReadOnly = React.useContext(ReadOnlyContext);
 
-  const viewButton = (isViewState: boolean): JSX.Element => (
+  const viewButton = (
     <DataEntry.View
       aria-pressed={state.type === 'ViewResourceState'}
       className="ml-1"
@@ -363,14 +363,9 @@ export function QueryComboBox({
         formatted?.resource === undefined ||
         collectionRelationships === undefined
       }
-      onClick={(): void => {
-        handleOpenRelated();
-        if (isViewState)
-          setState({ type: 'ViewResourceState', isReadOnly: true });
-      }}
+      onClick={(): void => handleOpenRelated(true)}
     />
   );
-
   return (
     <div className="flex w-full min-w-[theme(spacing.40)] items-center sm:min-w-[unset]">
       <AutoComplete<string>
@@ -428,7 +423,7 @@ export function QueryComboBox({
         {formType === 'formTable' ? undefined : isReadOnly ? (
           formatted?.resource === undefined ||
           hasTablePermission(formatted.resource.specifyTable.name, 'read') ? (
-            viewButton(false)
+            viewButton
           ) : undefined
         ) : (
           <>
@@ -439,7 +434,7 @@ export function QueryComboBox({
                   formatted?.resource === undefined ||
                   collectionRelationships === undefined
                 }
-                onClick={handleOpenRelated}
+                onClick={(): void => handleOpenRelated(false)}
               />
             )}
             {canAdd && hasNewButton ? (
@@ -535,7 +530,7 @@ export function QueryComboBox({
                 formatted.resource.specifyTable.name,
                 'create'
               ))
-              ? viewButton(true)
+              ? viewButton
               : undefined}
           </>
         )}
@@ -561,7 +556,6 @@ export function QueryComboBox({
             resource={formatted.resource}
             onAdd={undefined}
             onClose={(): void => {
-              setState({ type: 'ViewResourceState', isReadOnly: false });
               setState({ type: 'MainState' });
             }}
             onDeleted={(): void => {
--- a/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
@@ -4,6 +4,7 @@ import type { State } from 'typesafe-reducer';
 
 import { useAsyncState } from '../../hooks/useAsyncState';
 import { useResourceValue } from '../../hooks/useResourceValue';
+import { useTriggerState } from '../../hooks/useTriggerState';
 import { commonText } from '../../localization/common';
 import { userText } from '../../localization/user';
 import { f } from '../../utils/functools';
@@ -65,6 +66,7 @@ export function QueryComboBox({
   hasNewButton = true,
   hasSearchButton = true,
   hasEditButton = true,
+  hasViewButton = false,
   typeSearch: initialTypeSearch,
   forceCollection,
   searchView,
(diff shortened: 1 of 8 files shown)


CarolineDenis:
For testers: 
- open a record set, the query combo boxes should not have the 👁️ icon next to them
- edit the view definition of the form
- add to the QCBX, by example for cataloguer, the parameters `viewBtn=true` in 'initialize' like so: `<cell type="field" id="4" name="cataloger" uitype="querycbx" initialize="name=Agent;title=Catalog Agent;viewBtn=true" isrequired="true"/>`
- save
- disable cache and reload the form page
- the 👁️ icon should be present next to the cataloguer QCBX
- click on the 👁️ icon, the cataloguer record should open in read-only mode
Screenshot 2023-11-08 at 9 19 41 AM

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
```diff
@@ -518,6 +522,15 @@ export function QueryComboBox({
                 }
               />
             )}
+            {hasViewButton && (
```
maxxxxxdlp, line 525: there is already a `<DataEntry.View>` on line 415 - could you combine this button with it by updating the condition?

maxxxxxdlp, line 525: also, handle the case when user has no edit permission, but hasEditButton and hasViewButton is true - in that case your code displays two view buttons

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
```diff
@@ -352,6 +355,7 @@ export function QueryComboBox({
     hasTablePermission(field.relatedTable.name, 'create');
 
   const isReadOnly = React.useContext(ReadOnlyContext);
+  const [temporaryReadOnly, setTemporaryReadOnly] = useTriggerState(isReadOnly);
```
maxxxxxdlp, line 358: the value of this is only used when state.type === 'ViewResourceState', so why not modify the state's ViewResourceState state to add a prop like `isReadOnly: boolean`?
that fixes the issue with temporaryReadOnly existing even when it's not needed (possible source of bugs)
additionally, if two states are updated at the same time, then they should be merged into one state to avoid bugs where two states get out of sync/one state is updated before the other possibly causing UI to work incorrectly
(@realVinayak had this issue several times on the stats page)

maxxxxxdlp (changes requested):

maxxxxxdlp:
Also, don't forget to update the documentation in https://github.com/specify/specify7/wiki/Form-System with this change

similarly, you need to change https://github.com/specify/specify7/blob/issue-4170/specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts#L788C7-L788C20 for form editor to see this change
(once Form Editor 2.0 is done, the form parsing code will be thrown out as it will use form spec and XML syncer library directly - but until then, both places need to be updated)

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
```diff
@@ -534,24 +551,29 @@ export function QueryComboBox({
...
-        />
+        <ReadOnlyContext.Provider value={state.isReadOnly}>
+          <ResourceView
+            dialog="nonModal"
+            isDependent={field.isDependent()}
+            isSubForm={false}
+            resource={formatted.resource}
+            onAdd={undefined}
+            onClose={(): void => {
+              setState({ type: 'ViewResourceState', isReadOnly: false });
```
maxxxxxdlp, line 562: this is not right. what's the point of calling setState twice?
the 2nd call would just override the 1st call!
and the resource.set... line that was here before is still needed - without it the parent resource would still be pointing to a child resource, which no longer exists

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
maxxxxxdlp: sorry if I wasn't clear, but you misunderstood what I meant.
I meant to remove the `<DataEntry.View>` on line 525 because there already is a `<DataEntry.View>` on line 415.
instead change condition on line 412+make necessary refactors so that you only have `<DataEntry.View>` in the code once, but it is displayed in all the cases when necessary (and `<DataEntry.Edit>` is still displayed when necessary. this also avoids the possible bug with `<DataEntry.View>` being displayed twice

ANOTHER solution that avoids duplication is to create a variable that like `const viewButton = <DataEntry.View ... />` and then use it in two places - in either case, code duplication is avoided.

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
maxxxxxdlp: ups was looking at a wrong line. the 2nd part of my comment no longer applies (I crossed it out). the first part is still the case and must be fixed.
> this is not right. what's the point of calling setState twice?
the 2nd call would just override the 1st call!

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
```diff
@@ -352,6 +354,23 @@ export function QueryComboBox({
...
+      aria-pressed={state.type === 'ViewResourceState'}
+      className="ml-1"
+      disabled={
+        formatted?.resource === undefined ||
+        collectionRelationships === undefined
+      }
+      onClick={(): void => {
+        handleOpenRelated();
+        if (isViewState)
+          setState({ type: 'ViewResourceState', isReadOnly: true });
```
maxxxxxdlp, line 369: handleOpenRelated() takes care of calling setState
in that case why do you call setState again on line 369?

perhaps you meant to add an argument for handleOpenRelated that would determine whether it would call setState with isReadOnly:true or false?

maxxxxxdlp: after this change, both occurrences of `<DataEntry.View>` would be identically, so you don't need a function for this, but can assign the jsx expression to the variable directly and reuse that variable

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
```diff
@@ -352,6 +354,18 @@ export function QueryComboBox({
...
+    <DataEntry.View
+      aria-pressed={state.type === 'ViewResourceState'}
+      className="ml-1"
+      disabled={
+        formatted?.resource === undefined ||
+        collectionRelationships === undefined
+      }
+      onClick={(): void => handleOpenRelated(true)}
+    />
+  );
```
maxxxxxdlp, line 368:
```suggestion
  const viewButton = formatted?.resource === undefined ||
          hasTablePermission(formatted.resource.specifyTable.name, 'read') ? (
    <DataEntry.View
      aria-pressed={state.type === 'ViewResourceState'}
      className="ml-1"
      disabled={
        formatted?.resource === undefined ||
        collectionRelationships === undefined
      }
      onClick={(): void => handleOpenRelated(true)}
    />
  ) : undefined;
```

then you don't need to duplicate the condition in both places

maxxxxxdlp (approved):
looks good, but I suggested one nice refactor that would reduce code duplication

bronwyncombs (approved):
Adding "viewBtn=true" to initialize statement for qcbx field enables view button when browsing a record set. Form view is read-only and view button will be disabled if field is empty. Looks good!

https://github.com/specify/specify7/assets/135047322/74378244-cfaf-45f0-87e8-55f0a0a0fccc

PR #4200, opened 2023-11-08
Remove validation errors on form loading

### LINKED ISSUE #3700: Fields are shown as invalid immediately when forms load
This is on `xml-editor`:

https://github.com/specify/specify7/assets/37256050/4e056b23-e5fd-4d20-b727-c791dad3cbba

Though I have seen glimpses of this in production. This may be a bigger issue.

This is inconsistent and annyoing as in almost every instance refreshing results in the error not being recreatable. I know that @bronwyncombs and @melton-jason have seen this in the wild, and I'm sure we will see it again.

CarolineDenis:
Fixes #3700

### DIFF
--- a/specifyweb/frontend/js_src/lib/hooks/useValidation.tsx
+++ b/specifyweb/frontend/js_src/lib/hooks/useValidation.tsx
@@ -115,7 +115,7 @@ export function useValidation<T extends Input = Input>(
     validationRef: React.useCallback(
       (input): void => {
         inputRef.current = input;
-        setValidation(validationMessageRef.current, 'focus');
+        setValidation(validationMessageRef.current, 'auto');
       },
       [setValidation]
     ),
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/hooks/useValidation.tsx
```diff
@@ -115,7 +115,7 @@ export function useValidation<T extends Input = Input>(
     validationRef: React.useCallback(
       (input): void => {
         inputRef.current = input;
-        setValidation(validationMessageRef.current, 'focus');
+        setValidation(validationMessageRef.current, 'auto');
```
maxxxxxdlp, line 118: while this change here may fix the issue, it also changes the behavior in a lot of places (useValidation is used in a lot of places)

thus, you need to test the impact of this change in all the places where validationRef is used

it is possible that this change in behavior is desirable in all cases - but we need to check it, not just assume it is so

CarolineDenis: tested.

maxxxxxdlp (approved):

grantfitzsimmons (approved):
Yes!!! So much better.

pashiav (approved):
Looks good!

https://github.com/specify/specify7/assets/118556884/f0ddbd01-d72b-4deb-9f26-137a0e9e245b

PR #4201, opened 2023-11-08
Speedup GitHub Action dockerize job

acwhite211:
Remove qemu to speedup GitHub Action dockerize job.  Also, to further speedup development, only build for both platforms on commits to the production branch or tagged commits.  If you want to build an arm version of a development branch, then create a new branch ending with "arm" to get an arm64 build of the branch.  So, for example the branch name "blah-arm" will build an arm64 image and be push to DockerHub.


CarolineDenis (approved):

.github/workflows/docker.yml
```diff
@@ -48,9 +48,18 @@ jobs:
...
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
+          
+          # Determine the platforms based on the branch
+          PLATFORMS="linux/amd64"
+          REF_NAME=$(echo "${GITHUB_REF#refs/*/}")
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            PLATFORMS="linux/amd64,linux/arm64"
+          elif [ "$REF_NAME" = "production" ]; then
```
maxxxxxdlp, line 57: instead of hardcoding `production`, could we get the default branch name from github?
see https://stackoverflow.com/a/68414395/8584605

.github/workflows/docker.yml
```diff
@@ -48,9 +48,18 @@ jobs:
...
           echo "created=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" >> $GITHUB_OUTPUT
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
+          
+          # Determine the platforms based on the branch
+          PLATFORMS="linux/amd64"
+          REF_NAME=$(echo "${GITHUB_REF#refs/*/}")
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            PLATFORMS="linux/amd64,linux/arm64"
```
maxxxxxdlp, line 56: instead of having two if/else with the same body, could use the bash's OR syntax:
https://unix.stackexchange.com/a/47585/324980

maxxxxxdlp (approved):

grantfitzsimmons (approved):
See Max's comments, looks good to me

.github/workflows/docker.yml
```diff
@@ -52,9 +54,7 @@ jobs:
           # Determine the platforms based on the branch
           PLATFORMS="linux/amd64"
           REF_NAME=$(echo "${GITHUB_REF#refs/*/}")
-          if [[ $GITHUB_REF == refs/tags/* ]]; then
-            PLATFORMS="linux/amd64,linux/arm64"
-          elif [ "$REF_NAME" = "production" ]; then
+          if [[ $GITHUB_REF == refs/tags/* || "$REF_NAME" = "$DEFAULT_BRANCH" ]]; then
```
maxpatiiuk, line 57: what you have is fine, but this might work too, and would be a bit simpler:

```suggestion
          if [[ $GITHUB_REF == refs/tags/* || "$REF_NAME" = "${{ github.event.repository.default_branch }}" ]]; then
```

Issue #1460, opened 2022-05-09
Default form options


grantfitzsimmons:
**Describe the bug**
Users should have a choice between several form display options:

- White background, white enabled fields, gray disabled fields (System default)
- Off-white background, off-white enabled fields, gray disabled fields
- Dark background, dark enabled fields, black disabled fields

grantfitzsimmons:
Started to be fixed on this branch https://github.com/specify/specify7/issues/1232

grantfitzsimmons:
Will start on another branch and bring into agent merging

CarolineDenis:
@grantfitzsimmons do you mean add new pref to choose from this 3 options?

maxxxxxdlp:
hmm, I am struggling to come up with how this would be represented in the preferences.
would this be a separate theme picker from the main theme picker,
or would the main theme picker get a third option (off-white)?

maybe it would be best to just add a checkbox like "Accent form fields" or "Contrast form fields" or etc - if checked, fields are light gray in light mode and dark gray in dark mode - if not checked, fields match background color - in either case, disabled fields will be the opposite

grantfitzsimmons:
<img width="453" alt="image" src="https://github.com/specify/specify7/assets/37256050/3db095c8-aac3-4f4e-b567-8763fbd10490">

grantfitzsimmons:
My dream

PR #4202, opened 2023-11-08
Allow to specify boldness in xml editor

### LINKED ISSUE #3095: Make field label bold in form definition
It would be nice to be able to make text or a field's label bold in the form definition

CarolineDenis:
Fixes #3095

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormCells/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/index.tsx
@@ -41,6 +41,7 @@ const cellRenderers: {
     readonly formType: FormType;
     readonly align: typeof cellAlign[number];
     readonly verticalAlign: typeof cellVerticalAlign[number];
+    readonly isBold: boolean;
   }) => JSX.Element | null;
 } = {
   Field({
@@ -65,10 +66,16 @@ const cellRenderers: {
       />
     );
   },
-  Label({ cellData: { text, labelForCellId, title }, formatId, align }) {
+  Label({
+    cellData: { text, labelForCellId, title },
+    formatId,
+    align,
+    isBold,
+  }) {
     const style: React.CSSProperties = {
       textAlign:
         align === 'right' ? 'right' : align === 'center' ? 'center' : undefined,
+      fontSize: isBold ? 'bold' : undefined,
     };
     return typeof text === 'string' &&
       text.length === 0 ? null : typeof labelForCellId === 'string' ? (
@@ -321,6 +328,7 @@ export function FormCell({
   formType,
   align,
   verticalAlign,
+  isBold,
 }: {
   readonly resource: SpecifyResource<AnySchema>;
   readonly cellData: ValueOf<CellTypes>;
@@ -329,6 +337,7 @@ export function FormCell({
   readonly formType: FormType;
   readonly align: typeof cellAlign[number];
   readonly verticalAlign: typeof cellVerticalAlign[number];
+  readonly isBold: boolean;
 }): JSX.Element {
   const Render = cellRenderers[cellData.type] as typeof cellRenderers['Field'];
   return (
@@ -338,6 +347,7 @@ export function FormCell({
       formatId={formatId}
       formType={formType}
       id={id}
+      isBold={isBold}
       resource={resource}
       verticalAlign={verticalAlign}
     />
--- a/specifyweb/frontend/js_src/lib/components/FormParse/cells.ts
+++ b/specifyweb/frontend/js_src/lib/components/FormParse/cells.ts
@@ -281,6 +281,7 @@ export type FormCellDefinition = ValueOf<CellTypes> & {
   readonly visible: boolean;
   readonly ariaLabel: LocalizedString | undefined;
   readonly verticalAlign: typeof cellVerticalAlign[number];
+  readonly isBold: boolean;
 };
 
 const cellTypeTranslation: IR<keyof CellTypes> = {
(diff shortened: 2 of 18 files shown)


CarolineDenis:
Screenshot 2023-11-09 at 7 28 51 AM
Screenshot 2023-11-09 at 7 29 49 AM

specifyweb/frontend/js_src/lib/components/FormParse/cells.ts
```diff
@@ -355,6 +356,10 @@ export function parseFormCell(
       getBooleanAttribute(cellNode, 'invisible') !== true ||
       parsedCell === processCellType.Unsupported,
     ...parsedCell({ cell: cellNode, table, getProperty }),
+    isBold:
+      getBooleanAttribute(cellNode, 'isBold') === true ||
+      parsedCell === processCellType.Unsupported,
+    ...parsedCell({ cell: cellNode, table, getProperty }),
```
maxxxxxdlp, line 362: parsedCell is not supposed to be call twice - after your change it's called twice

specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
```diff
@@ -203,6 +204,7 @@ const processFieldType: {
       step: f.parseFloat(getProperty('step')),
       minLength: f.parseInt(getProperty('minLength')),
       maxLength: f.parseInt(getProperty('maxLength')),
+      isBold: getProperty('isBold')?.toLowerCase() === 'true',
```
maxxxxxdlp, line 207: why do you read `isBold` here if you already read `isBold` on the Cell level in https://github.com/specify/specify7/blob/issue-3095/specifyweb/frontend/js_src/lib/components/FormParse/cells.ts#L359-L361

specifyweb/frontend/js_src/lib/components/FormCells/index.tsx
```diff
@@ -65,10 +66,16 @@ const cellRenderers: {
...
+  Label({
+    cellData: { text, labelForCellId, title },
+    formatId,
+    align,
+    isBold,
+  }) {
     const style: React.CSSProperties = {
       textAlign:
         align === 'right' ? 'right' : align === 'center' ? 'center' : undefined,
+      fontSize: isBold ? 'bold' : undefined,
```
maxxxxxdlp, line 78: `fontSize` is not correct here. did you mean `fontWeight`?
and are you trying to set that style here if you are already doing that in `DataEntry.Cell`?

specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
maxxxxxdlp, line 207: also, if you look at sp6's behavior, if I remember correctly, the default value for whether form is bold is not always false - it might depend on whether field is required, or field type or etc, don't remember
not saying we have to make it work exactly the same way in sp7, but saying that we should at least discuss that

maxpatiiuk (changes requested):
Besides the issues with the code changes in this pull request, I would push back on making non-essential form parsing changes/additions for now. Reasons:
- Until form editor 2.0 is implemented (in the UI, not XML editing), to add new form attribute, we have to update form parsing, form parsing tests, form spec, form spec tests, and the wiki page - after form editor 2.0, we would only edit form spec+it's tests (and optionally provide description in the UI if not already clear from the name). Thus adding new form attributes would become way easier
- It's easy to add new form attribute. It's hard to maintain it in the future - just think of it, whatever attribute you add right now, would have to be supported in the future forever OR we would be forced to make breaking changes.
  - For example, imagine we want to also add italics, underline, strikeout, font weight, font family, font size, background color and etc. Do we add a separate attribute for each on label? And what if user also want this customization on the Separator, and SubView header and buttons and fields and etc? do we add this attribute in every place (which is what sp6 kind of did - horrible inconsistency resulted).
    - Perharps a better solution would be to add these attributes on the field or cell level, not on the label level. And perhaps we would want to prefix them with "style-", or combine them into a single CSS string, or do whatever else we might find convenient once we have 7-10 style-related attributes rather than random 1 or 2.
  - OR imagine we would move in the future to using a form system implemented by a third-party library, rather than by us - if they don't support bold text, we are out of luck (or even if they do support it, we have to do more work to migrate) - thus if this feature is not providing enough value to justify the future cost, it sholdn't be added (it's easy to add something now, but not as easy to design it in a way that it doesn't cause trouble in the future - sp6 made a lot of mistakes like that)
- Right now, there are very few form styling options available - which makes all forms look consistent, thus preserving specify brand identity and reducing maintenance effort for those who are editing forms. Also, whenever you add options, you allow users to mess things up (either break things and email us for support, or cause color contrast issues or etc). Plus, as it stands now, we are free to change form styling, without worrying that it will clash with the form styling changes users made - allowing users to customize styling will make this way harder or break user forms

Thus my suggestion is to:
- Get XML Editor out
- Work on Form Editor 2.0 (I will help with that)
- Then take time to think what the future of the form system may look like, plan things out, and then move in that direction decisively
  - for example, Grant suggested offering a few form presents rather than allowing arbitrary customizability (https://github.com/specify/specify7/issues/1460)
  - Grant also proposed making the "Simplified form definition" a bit more customizable (let collection managers reoder fields and hide/show fields), and add an option to set how many columns to render on the screen rather than just 1 - which would reduce number of customization options by a lot compared to current XML, but in return simplify form creating a lot AND simplify form editor UI implementation AND make forms usable on mobile devices/narrow screens
  - very good starting point for the brainstorming is this list of issues - https://github.com/specify/specify7/issues?q=is%3Aopen%20is%3Aissue%20project%3Aspecify%2Fspecify7%2F21

cc @grantfitzsimmons @CarolineDenis @melton-jason thoughts?

Issue #4172, opened 2023-10-30
Warning text for "no restrictions mode" in the query builder is misleading


grantfitzsimmons:
If you want to enable this in the UI at all times, you can check "No restrictions mode" in User Preferences under Query Builder. The warning here is misleading as it is reused from the "Data Mapper" no restrictions mode, so you can safely ignore it. Building a query should not lead to data loss or database corruption.

<img width="675" alt="image" src="https://github.com/specify/specify7/assets/37256050/3b25afb7-ec0c-4f5d-8543-bb86aca81958">

The reason it is showing this absurd text is because it is reusing the warning for the no restrictions mode in the WorkBench section:

<img width="680" alt="image" src="https://github.com/specify/specify7/assets/37256050/c4eaaab4-c9c4-4fee-80f7-39f1ad0095a7">

In that case, it is justified.

**Desired Behavior:**

Just remove the warning. There is no value in it, and having a "no restrictions mode" for query builder is not problematic for advanced users.

maxxxxxdlp:
The reason I added it is that in no-restrictions mode you can query system tables
Then, you can open those systems tables using "Browser in Forms"
From there, it's easy to make changes or hit Delete

If No restrictions mode is not enabled, then non-advanced users can't get to those tables (while advanced users can still change the URL to query any table they want)

Though, now that the permission system is here, that is less fo an issue. @grantfitzsimmons if you think it is not useful, then I am ok with removing it

PR #4203, opened 2023-11-09
Issue-4172 query builder preference warning text

carlosmbe:
Fixes Issue #4172. In this PR, I deleted the red text that was below the Query Builder's No Resreicitions Mode preference. To UX Test, make sure that "`no restctions mode`" does in fact still work and that there is no red text underneath the option in User Prefrences.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
@@ -1427,10 +1427,6 @@ export const userPreferenceDefinitions = {
             description: (
               <span>
                 {preferencesText.noRestrictionsModeQueryDescription()}
-                <br />
-                <span className="text-red-500">
-                  {preferencesText.noRestrictionsModeWarning()}
-                </span>
               </span>
             ),
             requiresReload: false,


carlosmbe:
# Before
Screenshot 2023-11-09 at 3 51 13 PM

# After

Screenshot 2023-11-09 at 3 41 50 PM

CarolineDenis (approved):

pashiav (approved):
Looks good!
issue-4172

melton-jason (approved):
:partying_face:

maxxxxxdlp (commented):
Before this pull request is merged, could someone address the questions I raised in https://github.com/specify/specify7/issues/4172#issuecomment-1786286824?

TLDR: I think this warning is useful and shouldn't be removed, but I am not a user-facing person so decide what you think is best, just hear out the reasons for why this was added in the first place

carlosmbe:
@grantfitzsimmons 
This is the comment Max is referring to:
> The reason I added it is that in no-restrictions mode you can query system tables
Then, you can open those systems tables using "Browser in Forms"
From there, it's easy to make changes or hit Delete

> If No restrictions mode is not enabled, then non-advanced users can't get to those tables (while advanced users can still change the URL to query any table they want)

> Though, now that the permission system is here, that is less fo an issue. @grantfitzsimmons if you think it is not useful, then I am ok with removing it

grantfitzsimmons:
>TLDR: I think this warning is useful and shouldn't be removed, but I am not a user-facing person so decide what you think is best, just hear out the reasons for why this was added in the first place

I think the dangerous action is deleting the records in those tables, not the ability to query them itself. If the user doesn't have permission to delete records from those tables, there is no inherent risk. 

Maybe the text could warn against modifying records in 'advanced' tables unless the user understands the consequences.

maxxxxxdlp:
> I think the dangerous action is deleting the records in those tables, not the ability to query them itself. If the user doesn't have permission to delete records from those tables, there is no inherent risk.

maxxxxxdlp:
> I think the dangerous action is deleting the records in those tables, not the ability to query them itself. If the user doesn't have permission to delete records from those tables, there is no inherent risk.

I guess it depends then on how much we are relying on users there to know what they are doing before they are given the permission to delete those tables. The main collection person would have access to everything. Other collection admins would have lots of access too.

The warning on that checkbox is guarding against the case you have person with lots of power - without this query builder pref, the person with lots of power but little specify experience won't have a way to access system tables in the UI, thus there is only so much they can do to break things accidentally. with this checkbox, you don't need to know the API or do URL editing to be able to delete from various tables - way lower bar.

anyway, just my 2c. feel free to soften the warning or remove it

Issue #4153, opened 2023-10-24
`xml-editor`: Workbench cell counts not displaying and not navigable


grantfitzsimmons:
<img width="1800" alt="image" src="https://github.com/specify/specify7/assets/37256050/fc275377-f16b-4795-92ea-337aaca51a11">

Counts should appear like this:
<img width="551" alt="image" src="https://github.com/specify/specify7/assets/37256050/de75cd69-a370-4611-8851-f74f0f5b7a34">

**Left:** https://nbmmnb5523-v79.test.specifysystems.org/specify/workbench/61/
**Right:** https://nbmmnb5523-xml-editor-2.test.specifysystems.org/specify/workbench/61/

Same database, `spadmin` user in the `Fish Voucher Collection`

PR #4205, opened 2023-11-13
Fix update cell count not running

realVinayak:
Fixes https://github.com/specify/specify7/issues/4153


CarolineDenis (approved):

melton-jason (approved):
How many times are we going to fix this issue? :smile: 

Nice adjustments: removing the `thisArgument` from `throttle`. It definitely looks cleaner this way. 

However, this may make it easier for a similar problem to crop up in the future regarding `throttle`: someone forgetting to bind a function or method to use the proper context. 

Given that this is the result of a 'feature' of Javascript, this cleaner approach is probably fine. 
Let's hope well all keep `this` in mind going further.

specifyweb/frontend/js_src/lib/components/WorkBench/CellMeta.ts
```diff
@@ -61,7 +61,7 @@ export class WbCellMeta {
 
   public constructor(private readonly wbView: WbView) {
     this.updateCellInfoStats = throttle(
-      () => this.updateCellInfoStats(),
+      () => this.updateCellInfoStats.bind(this),
```
maxxxxxdlp, line 64: thanks for catching this. one cool thing I recently learned about TypeScript, is that if you replace your property definition from this style:

```ts
  private name(): type {}
```

to this style:

```ts
  private name = (): type => {};
```

then TypeScript takes adds this to the constructor() automatically for you:

```ts
this.name = this.name.bind(this);
```

which means you don't have to add .bind() manually anywhere anymore

(but of course an even better solution is to not use classes  😋 as that's the only way to avoid all this-related bugs and all the other unholy things that come with OOP style)

cc @melton-jason

maxxxxxdlp (approved):

pashiav (changes requested):
Cell counts aren't showing up.
https://nbmmnb5523-issue-4153.test.specifysystems.org/specify/workbench/61
issue-4153

realVinayak:
@pashiav @specify/ux-testing can this be re-tested?

pashiav (commented):
**Modified**, **New Cells**, and **Error Cells** cell counts work!
issue-4153-cell-count

**Modified cells test:**

https://github.com/specify/specify7/assets/118556884/5e396b74-5f60-407f-a43d-ed331e662616




--
Not sure if related to to this issue, but **Search** doesn't work. I cannot properly test the **Search Results** cell count. 
In the screen recording, first I show that search doesn't return anything in `issue-4153` then I show how it looks like on `v79`.

https://github.com/specify/specify7/assets/118556884/70f60929-57e6-4b09-970d-b0c34dfa1c88

pashiav (approved):
Looks good!

https://github.com/specify/specify7/assets/118556884/5f2fe695-165e-4c58-8692-7f31693def47

PR #4208, opened 2023-11-13
Move record merging code into its own file [Issue-3923]

acwhite211:
Move record merge functionality into it's own file, but keep the view related functions in the view.py file


CarolineDenis (approved):

maxxxxxdlp (approved):
(optional) should we also move the django view definitions for record merging from views.py into record_merging.py (or some other file)?
or it is a django code style to have all views for one namespace in one file?

specifyweb/specify/views.py
melton-jason, line 1: Could the following be removed from `specify/views.py`? 
They are never used in `views.py` and are declared/used in `specify/record_merging.py`. 

https://github.com/specify/specify7/blob/964d4ec3d96871a72acb2ab1f6a5abd01367316a/specifyweb/specify/views.py#L414-L434

specifyweb/specify/views.py
```diff
@@ -5,29 +5,25 @@
 import json
 import mimetypes
 from functools import wraps
-from itertools import groupby
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Union
```
melton-jason, line 8: (optional)
Looks like `Any`, `Dict`, and `List` types are never used in this file. 
Could they potentially be removed?

specifyweb/specify/views.py
```diff
@@ -5,29 +5,25 @@
...
 from uuid import uuid4
 import traceback
```
melton-jason, line 10: (optional)
The `traceback` module is now never used in this file.
Could this import be removed?

melton-jason (changes requested):

melton-jason (approved):

PR #4210, opened 2023-11-13
 Keep white text on hovering side menu items

### LINKED ISSUE #4207: `xml-editor`: Hovering over a menu item when enabled completely hides the text
XML Editor:
xml_editor

V7.9.1
v7 9 1

CarolineDenis:
Fixes #4207

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Header/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/index.tsx
@@ -221,7 +221,7 @@ export function MenuButton({
     p-[1.4vh]
     ${
       isActive
-        ? 'bg-brand-300 text-white'
+        ? 'bg-brand-300 !text-white'
         : isSideBarDark
         ? 'text-white'
         : 'text-gray-700'


specifyweb/frontend/js_src/lib/components/Header/index.tsx
```diff
@@ -221,7 +221,7 @@ export function MenuButton({
     p-[1.4vh]
     ${
       isActive
-        ? 'bg-brand-300 text-white'
+        ? 'bg-brand-300 !text-white'
```
maxxxxxdlp, line 224: do you also need to add `!` on line 226 or is that good as is?

maxxxxxdlp, line 224: (optionally) find out what causes this to become an issue now because it wasn't an issue before

CarolineDenis: it's good at it is

CarolineDenis: I couldn't find the reason :confused:

carlosmbe (approved):
Looks Good and works well

realVinayak:
Actually, this seems like a bug in the production, rather than a bug in xml-editor. Not in v7.9.1, so got introduced recently.

image

grantfitzsimmons:
@CarolineDenis Can we cherry pick this commit into `v7.9.2`?

maxxxxxdlp:
yes, that's what https://github.com/specify/specify7/pull/4229 does

PR #4213, opened 2023-11-15
Sync conditional field in formatters

### LINKED ISSUE #3566: `xml-editor`: Changes are not synced if Condition field is unmapped in formatters
**Describe the bug**
If there is a definition for formatter with switch fields with condition field unmapped, then it looks like changes are not synced to XML if condition field is unmapped and then saved.

**To Reproduce**
1. Go to visual editor of formatters, and choose any formatter. If there is no Condition field, map a condition field and make multiple definitions.
2. Click on save and record the XML generated.
3. Unmap the Condition field, and every definition should disappear (as desired)
4. Click on save, and the record the XML again. The changes are not reflected in the XML. Additionally, the frontend seems to remember that Condition field is unmapped (and thus it does not reflect the actual XML, where Condition field is still mapped.).

**Expected behavior**
The changes should be synced.

**Screenshots**
The first two pics show the visual and xml representation of formatter before the condition field is unmapped.
before unmapping condition field visual
before unmapping Condition field
</br>
</br>
The second two pics show the visual and xml representation of formatter after the condition field is unmapped. Notice the discrepancy between XML and visual editor
after unmapping condition field visual
after unmapping condition field

CarolineDenis:
Fixes #3566

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataEntryTables/fetch.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataEntryTables/fetch.ts
@@ -18,7 +18,6 @@ export const fetchLegacyForms = f.store(
   async (): Promise<RA<SpecifyTable>> =>
     Promise.all([
       ajax<Element>(url, {
-         
         headers: { Accept: 'text/xml' },
       }),
       fetchSchema,
--- a/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/addMissingFields.test.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/addMissingFields.test.ts
@@ -4,7 +4,7 @@ import { addMissingFields } from '../addMissingFields';
 import type { AnySchema, SerializedResource } from '../helperTypes';
 import { getResourceApiUrl } from '../resource';
 import type { Agent } from '../types';
-import {localized} from '../../../utils/types';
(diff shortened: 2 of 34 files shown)


CarolineDenis:
Screenshot 2023-11-15 at 12 19 48 PM
Screenshot 2023-11-15 at 12 20 00 PM
Screenshot 2023-11-15 at 12 20 11 PM
Screenshot 2023-11-15 at 12 20 28 PM

maxxxxxdlp (changes requested):
This pull request has a ton of unrelated changes - how am I supposed to review it? 👽
Did you open a pull request from a wrong branch?

Also, to fix #3566 (which I don't think needs to be fixed, but ok), there are two ways:
- in react when user removes the condition, remove the other definitions if there are any - not an ideal option because the user may have just temporary changed the condition, without meaning to have all of the definitions deleted
- in the syncer spec for the formatters, when formatters are converted back to XML, just check if there is no condition, then remove all but first field

but again, I don't think we should really be fixing this.
see https://github.com/specify/specify7/issues/3566#issuecomment-1769757463
and the whole point of editors for XML is to avoid having users look at XML - this "bug" is only visible if you look at XML, which users should not have to do anymore

CarolineDenis:
@maxxxxxdlp , agreed. Thank you for you input

Issue #2859, opened 2023-01-20
`BusinessRuleException` when trying to create a CO record with an existing catalog number


grantfitzsimmons:
https://umichmollusks20221222-edge.test.specifysystems.org/specify/view/collectionobject/new/

```
```

```
('CollectionObject must have unique catalognumber in collection', {'table': 'CollectionObject', 'localizationKey': 'childFieldNotUnique', 'fieldName': 'catalognumber', 'fieldData': ('catalognumber', '000031300'), 'parentField': 'collection', 'parentData': 'collection: id=4', 'conflicting': [218354]})
```

grantfitzsimmons:
@melton-jason was unable to recreate the issue after I created a new CO with that catalog number. It looks like the database will need to be restored to the previous copy I have on the vault.

grantfitzsimmons:
**Reported By:** Paul Larson at FWRI

> I’ve been working my way through a record set and getting occasional error messages after trying to save a change record. It’s happened 5 or 6 times now, out of 165 edited records. Usually I can come back to the record set and try again and it works the second time. 

This is on `v7.8.5`

[Specify 7 Crash Report - 2023-02-01T17_56_35.727Z (1).txt](https://github.com/specify/specify7/files/10560054/Specify.7.Crash.Report.-.2023-02-01T17_56_35.727Z.1.txt)

melton-jason:
> @melton-jason was unable to recreate the issue after I created a new CO with that catalog number. It looks like the database will need to be restored to the previous copy I have on the vault.

I believe this is a frontend issue, not a backend one. The error is thrown because they already have a CollectionObject with that catalognumber (if this is not the case, please correct me). The issue is a bug in the frontend uniqueness rules.

I was unable to recreate the issue specifically because of the way Specify handled formatting numeric catalognumbers for CollectionObjects:

Consider the numeric catalog number format for the collection: `#########`:

Here is a the process Specify takes in an example workflow:

1. I enter `31300` into the Catalog Number field
2. The frontend asks the backend if `31300` is a unique catalog number
3. The backend then (correctly) tells the front end that there is not a CollectionObject with catalog number `31300`
4. The frontend then formats the catalognumber to the required format `000031300`
5. The frontend does not run the check on catalognumber again

To consistently reproduce:
1. Go to https://umichmollusks20221222-edge.test.specifysystems.org/specify/view/collectionobject/new/
2. Enter `1` in the CatalogNumber field
[...]

PR #4214, opened 2023-11-15
Do not set businessrule saveblockers sequentially

melton-jason:
There is currently a concerning bug on the `xml-editor` branch where the save blockers for business rules are not being shown. 

On  `xml-editor`: 

https://github.com/specify/specify7/assets/64045831/fbb8104b-4bfa-422a-a3fa-8206ce4f204d

### Dev Notes

The order of business rules whenever a field is changed is the following: 
- fieldChecks from the table's [business rule definitions](https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts)
- uniqueness rules
- tree-related business rules (if applicable)

https://github.com/specify/specify7/blob/1c85f365add6795ae841f892cfe9f6d2c30802d7/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts#L114-L123

Each of these checks return a BusinessRuleResult. 
The problem here is that the businessrule saveblockers were being set one at a time, by iterating over the `checks` array:

https://github.com/specify/specify7/blob/1c85f365add6795ae841f892cfe9f6d2c30802d7/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts#L148-L152

This does not seem to be the intended way of setting save blockers, which expects an array of strings corresponding to every save blocker message which each get converted to a `blocker`. 

The following snippet is the most relevant from setSaveBlockers()
 https://github.com/specify/specify7/blob/1c85f365add6795ae841f892cfe9f6d2c30802d7/specifyweb/frontend/js_src/lib/components/DataModel/saveBlockers.tsx#L108-L111

Consider the case in the video provided above. 
`checks` would become an array with the following:
- CollectionObject has no catalogNumber fieldCheck -> so this is valid
- There _is_ another CO with the same catalognumber -> invalid
- CO is not a tree resource -> valid

When the uniqueness rule save blocker is set, it then gets to the tree business rule checks which is valid. 
Due to the way setSaveBlockers() is implemented, the uniqueness rule save blockers become 'overridden' and filtered out, leaving only the valid tree business rule check. 

@maxxxxxdlp is this behavior intentional? Practically, it would imply only one 'type' of saveBlockers would be shown at a time. 
For example consider the case where there is both a parsing error and a business rule violation for a field. The error shown to the user will be the one _last_ set. 
This could lead to another bug as well: if parsing saveBlockers are applied first followed by businessrules, consider the case where the parsing is invalid but businessrules is valid (not that likely in the wild, however). 
The field would theoretically be OK to save because the parsing blockers would be filtered out. 

- - - 

### Testing Instructions
- Make sure the three types of uniqueness rules are properly indicated to the user 
  - Examples are provided in https://github.com/specify/specify7/pull/3851#issue-1824735764
  - A list of every uniqueness rule is provided in https://github.com/specify/specify7/pull/3996#issue-1888914474 or on the [Speciforum](https://discourse.specifysoftware.org/t/business-rules-and-logic-in-specify-7/1231)


maxxxxxdlp:
Looking at this function from before business rules were converted from .js to. ts, it looked like this:
https://github.com/specify/specify7/blob/135b342ad40eeaa68c192be198fd66251dd2c521/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.js#L117-L128

and here is commit (by me) that introduced this faulty code:
https://github.com/specify/specify7/commit/e41380df025a197cf285802b3fbb29f133665544#diff-afde7250d8079dd81ae9657ed120cf46a1f5147a88ac908f3d20840d79d88ee2R111-R123

can't deny, code looked simpler after the change, but oh well 😊

it's also possible that save blockers setting semantics was changed at some point which affected business rules

> is this behavior intentional? 
so safe to say that this is not intentional. but also safe to say that this bug has not been shipped to production yet 👏

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -138,23 +138,22 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
...
-          );
-        }
-        if (result.isValid) result.action?.();
-        return result;
-      })
-    );
+    const field = this.resource.specifyTable.strictGetField(fieldName);
+
+    if (!specialFields.has(fieldName)) {
+      const saveBlockerMessages = filterArray(
```
maxxxxxdlp, line 146: not related to this PR, but though you might find interesting:
I have stopped using the `filterArray` in my projects now, because I discovered a better version of it at Esri:
instead of `filterArray(someArray.map(...))`, we do `someArray.map(...).filter(isDefined)`

where isDefined looks like:
```ts
const isDefined = <T,>(value:T | undefined): value is T => value !== undefined;
```

pros:
- prettier has an easier job formatting your code to look better because .filter(isDefined) call is just chained at the end, rather than having to wrap your entire .map() in a filterArray()
- besides `isDefined`, you can come up with other helpful type assertion utils (i.e isTruthy)

feel free to do a refactor of specify7 code at some point to replace filterArray() with .filter(isDefined)

maxxxxxdlp (approved):

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
melton-jason: Good refactoring suggesting!
I would agree that would be a very nice improvement. :clap:

maxxxxxdlp: cc @CarolineDenis @realVinayak

pashiav (approved):
Tested for unique identifier, combination, and value. Looks good!

https://github.com/specify/specify7/assets/118556884/57da1985-2bc2-46b2-a590-91c0475f3c5b


busrule-unique-collector

melton-jason:
Can we get at least one more approval on this PR before merging?

CarolineDenis:
@melton-jason , I tested the following: 
V existing agent 
V unique agent for accession 
X unique cat number for CO (see screen recording below) 

https://github.com/specify/specify7/assets/108160931/2aa5ee53-be42-4161-bddf-cd50f6d14600

melton-jason:
> @melton-jason , ... X unique cat number for CO (see screen recording below)
> 
>  Screen.Recording.2023-11-22.at.7.14.41.AM.mov

@CarolineDenis 
:grimacing: 
This is a known issue which might not be related to the changes here. 
See https://github.com/specify/specify7/issues/2859#issuecomment-1468868462

To get around the bug, you would need to make a change to the catalognumber field after it has been formatted, or entering the formatted catalognumber.  
An easy test, for example, would be removing a character at the end of the catalognumber and readding it.

CarolineDenis (approved):

PR #4220, opened 2023-11-17
Remove gallery icon from fictious record set

### LINKED ISSUE #4211: `xml-editor`: Attachment gallery button should not display unless there are attachments
image

Added in https://github.com/specify/specify7/pull/3363

I believe it appears out of place and lacks usefulness without attachments. It does not align with the form meta menu button and lacks functionality without attachments already uploaded.

CarolineDenis:
Fixes #4211

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
@@ -222,9 +222,9 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
                   className={`flex-1 ${dialog === false ? '-ml-2' : '-ml-4'}`}
                 />
               )}
-              {hasAttachments && (
+              {hasAttachments && title !== undefined ? (
                 <RecordSetAttachments records={records} onFetch={handleFetch} />
-              )}
+              ) : undefined}
               {specifyNetworkBadge}
             </div>
             {totalCount > 1 && <div>{slider}</div>}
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
```diff
@@ -222,9 +222,9 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
                   className={`flex-1 ${dialog === false ? '-ml-2' : '-ml-4'}`}
                 />
               )}
-              {hasAttachments && (
+              {hasAttachments && title !== undefined ? (
```
maxxxxxdlp, line 225: you can make the code more self-documenting by defining a variable like `const isNewRecordSet = title === undefined` and then using it here.

and while you are at it, improve your condition by adding isInRecordSet to it:

```ts
const isNewRecordSet = isInRecordSet && title === undefined;
```

maxxxxxdlp (commented):
I am ok with this solution if @grantfitzsimmons is ok

maxxxxxdlp (approved):

CarolineDenis:
@pashiav 
- go to Data Entry 
- verify that the gallery icon is not present in the header
- save the form and add a new one
- save the second one
- verify the record set icon is now present but not the gallery icon
- click on record set icon to create a record set 
- verify that now the gallery icon is present 
- open any record set and verify gallery icon is present
- NOTES: gallery icon is disabled if there is no attachments

bronwyncombs (approved):
Gallery icon is not shown on data entry form headers or fictitious record set headers. 
Screenshot 2023-11-22 at 10 51 36 AM
Screenshot 2023-11-22 at 10 52 39 AM

In a record set, gallery icon is now shown: 
Screenshot 2023-11-22 at 10 53 28 AM

You get this dialog if no attachments are added upon clicking the icon: 
Screenshot 2023-11-22 at 10 53 57 AM

I know that disabling the icon in this case was discussed. I don't have a preference one way or the other really but am requesting review from @grantfitzsimmons to follow up

emenslin (approved):
Galley icon does not show up unless it is an actual record set. Even if there are no attachments the gallery icon is there, as Bronwyn said, I don't care if it is there or not, but it depends on what Grant thinks.

PR #4221, opened 2023-11-20
Relaod page after unmapped fields deletion in queries

### LINKED ISSUE #4216: Saved Queries - Unmapped Fields Not Immediately Removed After Clicking "Remove"
**Describe the bug**
In the Query, after saving a query with unmapped fields, a popup appears asking whether to "remove them." However, upon clicking "Remove," the unmapped fields persist in the view until the page is refreshed. I talked to Caroline about it and this behavior is a bug. The expectation is for the unmapped fields to disappear immediately after clicking "Remove."

**To Reproduce**

1. In the Query Builder, save a query with unmapped fields.
2. A popup will appear asking if you want to "remove them."
3. Click "Remove" in the popup.
4. Observe that the unmapped fields are still visible in the view.
5. Refresh the page to see that the fields are then removed.

**Expected behavior**
The expected behavior is for the unmapped fields to be immediately removed from the view upon clicking "Remove," without the need for refreshing the page.

**Beautiful Pictures**

https://github.com/specify/specify7/assets/53784701/f877e816-e3ca-4cbb-8f8a-8c2e755ba133

**Desktop:**
 - OS: Mac
 - Browser: Chrome

CarolineDenis:
Fixes #4216

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Components.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Components.tsx
@@ -173,7 +173,10 @@ export function QueryButton({
       dialogMessage={queryText.queryDeleteIncompleteDescription()}
       disabled={disabled}
       showConfirmation={showConfirmation}
-      onConfirm={handleClick}
+      onConfirm={(): void => {
+        handleClick();
+        globalThis.location.reload();
+      }}
     >
       {children}
     </ButtonWithConfirmation>


specifyweb/frontend/js_src/lib/components/QueryBuilder/Components.tsx
```diff
@@ -173,7 +173,10 @@ export function QueryButton({
       dialogMessage={queryText.queryDeleteIncompleteDescription()}
       disabled={disabled}
       showConfirmation={showConfirmation}
-      onConfirm={handleClick}
+      onConfirm={(): void => {
+        handleClick();
+        globalThis.location.reload();
```
maxxxxxdlp, line 178: the component you are modifying (QueryButton) is used in a lot of places - you certainly don't want to cause page reload in all those places.
please remember to check for usages when modifying some component so that you can see all the places it could be affecting!

maxxxxxdlp, line 178: also, reloading the page is an ugly hack, not a solution
this looks like an issue with state not being updated? or updated but change not propagated? that is what should be fixed. forcing the user to reload the page as a workaround for this is a terrible UX (and note not all specify instances load as fast as the do on your expensive laptop, so we need to avoid reloading needleslly)

maxxxxxdlp (changes requested):
this looks like a workaround for the bug

PR #4223, opened 2023-11-20
Add toString to use trim function

### LINKED ISSUE #4067: `TypeError: value.trim is not a function` when uploading a spreadsheet in the WorkBench
image

**Steps to Recreate:**
1. Upload this spreadsheet as a new data set for the WorkBench
2. See error

[**Link to Spreadsheet**](https://docs.google.com/spreadsheets/d/1vQpWDXa1PPWABGqX3tzJlj5IJELkCszD/edit?usp=drive_link&ouid=115297369899192276851&rtpof=true&sd=true) 🖨️ 

**Crash report from `localhost`:**
[Specify 7 Crash Report - 2023-10-04T21 14 44.858Z.txt](https://github.com/specify/specify7/files/12809102/Specify.7.Crash.Report.-.2023-10-04T21.14.44.858Z.txt)

This crash happens on any database, not specific to Edinburgh when using this spreadsheet.
Firefox, both on `v7.9.0` and `v7.8.13`. Seems to be specific to this spreadsheet.

Creating a new spreadsheet with the same contents still results in a crash.

**Reported By:** RBGE

CarolineDenis:
Fixes #4067

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/WbImport/helpers.ts
+++ b/specifyweb/frontend/js_src/lib/components/WbImport/helpers.ts
@@ -61,9 +61,9 @@ export const getMaxDataSetLength = (): number | undefined =>
   );
 
 export function extractHeader(
-  data: RA<RA<string>>,
+  data: RA<RA<number | string>>,
   hasHeader: boolean
-): { readonly rows: RA<RA<string>>; readonly header: RA<string> } {
+): { readonly rows: RA<RA<number | string>>; readonly header: RA<string> } {
   const header = hasHeader
     ? uniquifyHeaders(data[0].map((value) => f.trim(value.toString())))
     : Array.from(data[0], (_, index) =>
--- a/specifyweb/frontend/js_src/lib/components/WbImport/helpers.ts
+++ b/specifyweb/frontend/js_src/lib/components/WbImport/helpers.ts
@@ -65,7 +65,7 @@ export function extractHeader(
   hasHeader: boolean
 ): { readonly rows: RA<RA<string>>; readonly header: RA<string> } {
   const header = hasHeader
-    ? uniquifyHeaders(data[0].map(f.trim))
+    ? uniquifyHeaders(data[0].map((value) => f.trim(value.toString())))
     : Array.from(data[0], (_, index) =>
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/WbImport/helpers.ts
```diff
@@ -65,7 +65,7 @@ export function extractHeader(
   hasHeader: boolean
 ): { readonly rows: RA<RA<string>>; readonly header: RA<string> } {
   const header = hasHeader
-    ? uniquifyHeaders(data[0].map(f.trim))
+    ? uniquifyHeaders(data[0].map((value) => f.trim(value.toString())))
```
maxxxxxdlp, line 68: what was the type of the value on which the code tried to call .trim() and failed?

was it undefined, or null or number? whatever it was, the typescript type shows that the type would always be string - so at very least you should modify the typescript type on line 64
but you should go one step further and see what caused some non-string values to be there in the first place

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/WbImport/helpers.ts
CarolineDenis: it was number

maxxxxxdlp (approved):
code looks good, but could you please take a few minutes to research that the places using this function are ok with receiving a number (i.e in one place, data is sent to the back-end - how does back-end treat receiving a number instead of string)

carlosmbe (approved):
Tested with the spreadsheet Grant mentions in the issue and importing the file now works on multiple databases

pashiav (approved):
Looks good! Imports spreadsheet without error.

https://github.com/specify/specify7/assets/118556884/3a069d73-6489-4d4a-8cb0-f14a0cc2a03d

melton-jason (approved):
Assuming `data` is only ever of type `RA<RA<number | string>>`, then this should work. 
I am curious if the innermost type of `data` in this case could ever be something like null and/or undefined. 
If this can not ever be the case, then these changes should be enough to account for a number or string.

PR #4224, opened 2023-11-20
Add a bulk resolve and bulk return to return loan

### LINKED ISSUE #4049: Bulk return and resolve for loanprepreturn popup
>It would be very handy to have the same "bulk" option that was added to the addition of loan preparations to a loan added to the loan preparation return popup of a loan:
>
>image
>
>For our case we have tissue loans where every object needs to be resolved but not returned - this is not something we can do with the feature to return loan preps via a query

**Requested By:** Dan Baker at CSIRO ([Asana](https://app.asana.com/0/1201972579963460/1205536478752065/f))

CarolineDenis:
Fixes #4049


specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
```diff
@@ -104,6 +104,36 @@ function PreparationReturn({
     ({ resolve, unresolved }) => resolve < unresolved
   );
 
+  const [bulkReturn, setBulkReturn] = React.useState(0);
+  const [bulkResolve, setBulkResolve] = React.useState(0);
+  const quantities = preparations.map(
+    (preparation) => preparation.get('quantity') ?? 0
+  );
+  const maxPrep = Math.max(...quantities);
```
maxxxxxdlp, line 112: can quantities.length ever be 0 here? if so, use f.max instead of Math.max as then TypeScript will force you to handle this edge case

specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
```diff
@@ -205,6 +235,36 @@ function PreparationReturn({
           resource={loanReturnPreparation.current}
           viewDefinition={loanReturnPrepForm()}
         />
+        <div className="flex justify-end gap-2">
```
maxxxxxdlp, line 238: since fields are inline, not on top of each other, I would suggest adding more gap to this `<div>` to better differentiate the fields (i.e change gap-2 to gap-4 or gap-8)

also, just because the person who sent you the screenshot had these fields right-aligned, doesn't mean you should implement them that way - decide what do you think looks the best or produces best UX

specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
```diff
@@ -104,6 +104,36 @@ function PreparationReturn({
...
+        const unresolved =
+          quantities[index] -
+          (preparations[index].get('quantityResolved') ?? 0);
+        const resolvedCount = Math.min(newCount, unresolved);
+
+        return {
+          ...returnItem,
+          resolve: type === 'resolve' ? resolvedCount : resolve,
+          returns: type === 'returns' ? resolvedCount : returns,
+        };
```
maxxxxxdlp, line 132: this logic is almost correct, but not fully there:

see how if user modified returned amount, the resolved amount can get changed too:
https://github.com/specify/specify7/blob/xml-editor/specifyweb/frontend/js_src/lib/components/Interactions/PrepReturnRow.tsx#L109-L112

similarly, if user modified the resolved amount, the returned amount may change too - you need to add the same logic here (ideally you won't copy paste this logic between PrepReturnRow and here but would define a common small utility and reuse - even though this logic is quite concise, it's still a bit tricky, so better not have it defined in two places as that could lead to bugs like this)

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
maxxxxxdlp: sure.

first, definitions:
- unresolved - total number of items that aren't accounted for yet - thus the maximum number of items we could account for in this transaction
- resolved - the number of items we want to account for. this of course has to be lower than unresolved count
- returned - the number of items we physically received back from the loan. this is not always the same as resolved count. i.e if a specimen was destoyed/lost/consumed, then it is resolved, but not returned. return count has to always be less or equal to the resolved count.

in your function at the moment, if I try to increase the return count, it may create a situation where return count is larger than resolved, which should not be allowed - in such cases, the resolved count should be increased too, to match the return count.
similarly, if your function is called to lower the resolved amount, it needs to make sure that returned amount is lowered to match resolved amount.


this is what the max and min in here is for:
https://github.com/specify/specify7/blob/xml-editor/specifyweb/frontend/js_src/lib/components/Interactions/PrepReturnRow.tsx#L109-L112
https://github.com/specify/specify7/blob/xml-editor/specifyweb/frontend/js_src/lib/components/Interactions/PrepReturnRow.tsx#L129-L132
those are making sure to enforce this constraint when user is modifying the returned/resolved count

now with two new fields you added in this PR, you do not have similar constraint enforcement yet. so my comment is that you need to handle these cases.
and since this logic is somewhat tricky and confusing, it would be best if you could define this logic in one place (some sort of function - or two functions, one for when resolved amount is changed, and another for when returned amount changed), and then use that function both when user modified a single row directly, and when user modified the field responsible for all rows.

maxxxxxdlp (changes requested):

specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
CarolineDenis: ok thank you very much. Will try that :+1:

specifyweb/frontend/js_src/lib/components/Interactions/PrepReturnRow.tsx
```diff
@@ -175,3 +175,63 @@ export function PrepReturnRow({
     </>
   );
 }
+
+export function handleReturnChanged({
```
maxxxxxdlp, line 179: handle* naming convention is usually used for functions that react to change by updating the sate, and those functions usually don't return anything. in your case, the [function is pure](https://en.wikipedia.org/wiki/Pure_function), and it does return a value

perhaps a better name would be `updateReturnCount` and `updateResolveCount`

specifyweb/frontend/js_src/lib/components/Interactions/PrepReturnRow.tsx
```diff
@@ -175,3 +175,63 @@ export function PrepReturnRow({
...
+  newReturn,
+  resolve,
+  unresolved,
+  remarks,
+}: {
+  readonly newReturn: number;
+  readonly resolve: number;
+  readonly unresolved: number;
+  readonly remarks: string;
+}): {
```
maxxxxxdlp, line 189: don't duplicate the type definition. use PrepReturnRowState instead

maxxxxxdlp (changes requested):

maxpatiiuk (approved):

emenslin (approved):
Looks good and works as expected. 
Screenshot 2023-12-13 152950

acwhite211 (approved):

PR #4230, opened 2023-11-22
Fixes tree using stale focus row when switching between trees

realVinayak:
Consider the below case. 
image
The current active node is `mnianceum` in the first tree. If I click on any other node in the bottom tree, say `lastii` and then click on `niveum` in the first tree again, below function will get triggered with nodeId of `mnianceum`. 
https://github.com/specify/specify7/blob/f00f425e6a089bc3ff2d111505a753deff630a02/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L225-L238

This is because https://github.com/specify/specify7/blob/f00f425e6a089bc3ff2d111505a753deff630a02/specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx#L65-L67 has not been called yet, which is supposed to switch focus node to `niveum`.

Refactored to store focusedRow as a separate state, which is used directly (instead of maintaining two separate focused rows, which caused an implicit change if tree has not finished focusing yet, and then again after focus)


CarolineDenis (approved):

maxxxxxdlp (approved):
good catch!

realVinayak:
@specify/ux-testing @specify/dev-testing

**General**
Test all tree actions. Especially merging across regions (selecting a node from bottom/right and merging/moving/synonymizing into top/left or vice-versa.
**Specific Instructions**
1. Select a node from top tree.
2. Click on merge icon.
3. Select a target from the bottom tree, and merge.
4. Click on any other node in the top tree. Notice that error is thrown (saying no Taxon matches the query) in `xml-editor`. This branch doesn't throw that error.

pashiav (approved):
Tested Merge, Move, and Synonymize/Desynoymize. All are functional.

**_Specific Instructions_** - doesn't throw an error when clicking on other nodes after merging. Looks good! 


https://github.com/specify/specify7/assets/118556884/65ba0c08-e6c6-478e-82de-a2c00c6a71ea

Issue #4235, opened 2023-11-27
`ParseError` upon logging in when a user has an invalid `ExpressSearchConfig`


grantfitzsimmons:
```
Error occurred fetching from https://sandbox-rbge.specifycloud.org/context/available_related_searches.json

ParseError
not well-formed (invalid token): line 1, column 1
Content
"None"
```
[Specify 7 Crash Report - 2023-11-27T18_16_21.485Z.txt](https://github.com/specify/specify7/files/13478880/Specify.7.Crash.Report.-.2023-11-27T18_16_21.485Z.txt)

User `rcubey` at Edinburgh is reporting an issue upon logging in. Immediately after logging in, this error occurs.

It appears that their `ExpressSearchConfig` is invalid:

<img width="819" alt="image" src="https://github.com/specify/specify7/assets/37256050/bc5394aa-4666-4bd6-9ce1-5fd92ad33e81">
[Broken_ExpressSearchConfig.xml.zip](https://github.com/specify/specify7/files/13478850/Broken_ExpressSearchConfig.xml.zip)

Here's it fixed, after fixing the formatting, and now the error does not occur:
<img width="959" alt="image" src="https://github.com/specify/specify7/assets/37256050/52688c66-cd64-4250-a6bb-698fb276707f">
[Fixed_ExpressSearchConfig.xml.zip](https://github.com/specify/specify7/files/13478847/Fixed_ExpressSearchConfig.xml.zip)

Normally I would not write this as a bug, as the file is not structured properly, but this error is only happening in `v7.9.1` and not `v7.9.0`. If no other reports come out, feel free to close this, but it's strange/potentially an issue that the error was not being reported before.

It should fall back and use the default if the file is corrupt, not throw a warning each time.

**Reported By:** RBGE

melton-jason:
It may be the case the file being misformed/corrupted is a result of an agent merging bug (see https://github.com/specify/specify7/pull/4237#issue-2017195096). 

Would #4237 resolve this issue, or is the intended fix to handle the case when an AppResource is misformed in such a manner?

emenslin:
fixed

PR #4237, opened 2023-11-29
Restrict update fields to just relationships in duplicates

realVinayak:
This is needed because of discrepancy in how SpAppresourceData's data field is stored in the database. It is actually a binary field in MySQL (BLOB), but treated as a TextField. Due to this, when saving a SpAppResourceData's data field, it gets double encoded. That is, a string like `b"some"` gets converted to `b'b"some"'`.

This is actually an issue anywhere objects from SpAppresourceData is stored. However, other usages don't preserve the data field (they are overwritten from the user provided data, which happens to be decoded - as a string). It just becomes apparent in agent merging because we replace the createdby and modifiedby fields, but django updates all the fields when calling `.save()`, so the data field in SpAppResourceData gets corrupted. This PR gets around encoding issue by just updating the relationship field (instead of all the fields)


melton-jason (approved):
Nice change!

Since the changes on this branch occurred before #4208 was merged into production, I have updated this branch and moved the change over to `record_merging.py`

CarolineDenis (approved):

acwhite211 (approved):

melton-jason:
@realVinayak While the changes here look solid, I would imagine this might warrant an approval from the UX testing team before merging, correct?
If so, would you recommend any testing instructions? 

Perhaps the following may be enough to minimally recreate the issue:
-  With a specifyuser assigned to an agent, modify and save any Spresource
- Merge the agent associated with the specifyuser into another agent
- Ensure the Spresource is not malformed (as in https://github.com/specify/specify7/issues/4235#issue-2012852262)

realVinayak:
@melton-jason It would be worth adding a backend unit test for this, rather than UX. However, that is complicated because Django doesn't treat `data` field in `spappresourcedata` as `BinaryField` but instead as `TextField`. This also means that the models developed during testing (ie in the test database) also have `TextField` instead of `BinaryField`, which means the backend test for this won't be useful. 

We discussed fixing this discrepancy in the first place (treating it as BinaryField). I tested that, and it seems to work. Still discussing a good way to fix this

grantfitzsimmons (approved):
Here's the video of me performing the merge on a copy of the `mcnb` database:
[Google Drive link](https://drive.google.com/open?id=1w8xUm5rNDXd0KLwCCTnqqpEihIuDQOdM&usp=drive_fs)

Here's the outcome where I could verify that the app resources are corrupted on `edge` (approximately `7.9.3` at this time) and not on `restrict-update-field`.  

https://github.com/specify/specify7/assets/37256050/3abfab5f-9b2a-4a3a-a3b1-38a728677ee1

**Steps Taken:**
1. Started with this backup of the [mcnb database](https://drive.google.com/open?id=1JISw5CYHsrhrskIQKZDTe63zurCuIT9f&usp=drive_fs)
2. Created two instances with two databases using both `edge` and `restrict-update-field`
3. Built a query to find agents identified by `ID` `13330, 12822, 13386`
4. Merged these agents (`13330` is the final ID for the new agent)
5. Examine app resources and see that the issue has been resolved with this PR

specifysoftware:

PR #4240, opened 2023-11-30
perf: Improve Home Page load performance

maxpatiiuk:
Inspired by an accent post series [Speeding up the JavaScript ecosystem](https://marvinh.dev/blog/speeding-up-javascript-ecosystem/), I used the website recommended by that author (https://www.speedscope.app/), to analyze the performance traces from when specify login page and home page loads.

In the process, I noticed that considerate time was spent loading files which are not needed on the home page (Form system, Query Builder, Agent Merging, Reports, Leaflet, Codemirror, Attachments, ...).

Main causes:
- a file imports a small util from a related file, but the related file in turns imports a lot of different large files - solution is to move the util into a separate file
- a file imports react component, which in turns imports a giant react component (like ResourceView), which in turn had thousands of indirect dependencies - solution is to load such react components async using React.Lazy

at the end, this reduced the javascript execution time on the home page by 100ms on an m1 MacBook (on a development build of code)


maxxxxxdlp:
I will look at the failing tests tomorrow evening, unless someone wants to look at it before than

realVinayak (approved):

specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
```diff
@@ -72,7 +73,7 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
       <AnyComponent {...allProps} />
       {isOpen && (
         <IsNotReadOnly.Provider value>
-          <ResourceView
+          <LazyResourceView
```
melton-jason, line 76: Currently, this is the only use of LazyResourceView. 
Can the function be used anywhere else to improve performance or was this the only place where it can be used or was a significant bottleneck?

melton-jason (approved):

specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
maxpatiiuk: in other places where ResourceView is used, most of ResourceView's dependencies would have to be loaded anyway.

i.e if ResourceView is called from inside of ResoruceView, then ResourceView is already loaded, and so using LazyResourceView in that place would add (tiny) overhead.

I guess the query "View in forms" is one place where this might help a bit, but probably won't be noticeable.

---

on the other hand, the ResourceLink was called from the Notifications rendered, which is from the Header, which is called from everywhere - meaning every single webpage in specify was loading ResourceView and all of it's dependencies (attachments, all form cells and plugins, etc...)
and so all that extra code on the home page adds extra time to 1. network fetching (though not a big issue for an app like specify as js bundles are going to be cached, and aren't updated often) 2. browser needs time to parse all that JS (100ms as seen in my benchmarking - not too much, but that's 100ms on every page load for every user, so it adds up), 3. more ram usage since more functions are declared/objects created needlessly (this one might be more significant for performance on weaker systems)

Issue #4245, opened 2023-12-01
Prevent scrolling from altering numeric input fields


grantfitzsimmons:
**Describe the bug**
Clicking on a numeric field and scrolling/zooming result in the number changing. While this behavior is fine for other web forms, when you have numeric fields with precise measurement data, it is easy to accidentally change something when scrolling away.

We should disable this feature as the probability of accidental data loss is high. I accidentally edited some fields during a demo – not a good look.

**To Reproduce**
1. Click on a numeric field
2. Scroll while your mouse is hovering over the field
3. See numbers are changing.

**Expected behavior**
Scrolling should not change numeric values.

**Screenshots**

https://github.com/specify/specify7/assets/37256050/86fe7116-0da6-4c65-9a63-0a028e84fa8b

Brave, macOS

PR #4249, opened 2023-12-02
feat(Atoms): prevent numeric input from changing on scroll

maxpatiiuk:
Fixes https://github.com/specify/specify7/issues/4245

To test:
Make sure that https://github.com/specify/specify7/issues/4245 does not happen for numeric inputs in the following two places:
- numeric field that is part of the form (like the one you see the video in https://github.com/specify/specify7/issues/4245)
- numeric field that occurs outside the form (i.e the input field that shows the index of current item in the sub view)

(those input fields look the same, but are different in the code)


CarolineDenis (approved):

melton-jason (approved):
If a numeric field is required and has no value, scrolling over it will still insert a value. 
Barring that (if that is even an issue at all), this is fine!  

@specify/ux-testing 
When testing, please also provide your thoughts regarding focusing the field when scrolling over it!

grantfitzsimmons (approved):

Issue #4250, opened 2023-12-03
Parsing tree remote preferences doesn't strip whitespace characters


realVinayak:
In the barcelona's database, the 
`sp7.allow_adding_child_to_synonymized_parent.Taxon=true` is actually `sp7.allow_adding_child_to_synonymized_parent.Taxon=true\r`. But backend checks strictly for 'true' when enforcing this preference, leading to an error.

PR #4251, opened 2023-12-03
Strip white-space characters when enforcing business rule

realVinayak:
See https://github.com/specify/specify7/issues/4250

Testing instructions
1. Make sure children can be added to a synonymized node on mcnb databases on the test panel


CarolineDenis (approved):

maxpatiiuk (approved):
can we just move on to institution prefs as fast as possible?
@realVinayak are there any major steps remaining before that migration can be done? it seems like all the major pieces are completed (UI, fetching logic, hooks, typescript types) - now just write the migration script and that's it

until we have that, fixing the old system might be a waste of time.
not to mention all the cool and requested global prefs we could add once we have this

not to mention, the new system would be separate from sp6

emenslin (approved):
I was able to add children to a synonymized node, looks good.

PR #4254, opened 2023-12-04
Remove subviews from search dialogs

### LINKED ISSUE #4226: `xml-editor`: Forms being used for searching should not allow attachments or records to be created
image

Currently, when you try to search for records to preview how an aggregator/format will look, it displays the whole form (unless a search dialog form is defined). For instance, look at the `Preparation` aggregation editor here:
https://saiabjan1023-xml-editor.test.specifysystems.org/specify/resources/app-resource/233/formatter/Preparation/43

You can upload attachments here and they will end up on the asset server. Everything should be read-only in this dialog, as it's meant to act as a place to search for records.

CarolineDenis:
Fixes #4226

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
@@ -93,10 +93,6 @@ export function IntegratedRecordSelector({
 
   const isAttachmentTable = tables.Collection.name.includes('Attachment');
 
-  const isInSearchDialog = React.useContext(SearchDialogContext);
-  const isAttachmentPlugin = viewName === 'ObjectAttachment';
-  const isAttachmentInSearch = isInSearchDialog && isAttachmentPlugin;
-
   return formType === 'formTable' ? (
     <ReadOnlyContext.Provider value={isReadOnly}>
       <FormTableCollection
@@ -118,7 +114,7 @@ export function IntegratedRecordSelector({
       />
     </ReadOnlyContext.Provider>
   ) : (
-    <ReadOnlyContext.Provider value={isAttachmentInSearch || isReadOnly}>
+    <ReadOnlyContext.Provider value={isReadOnly}>
       <RecordSelectorFromCollection
         collection={collection}
         defaultIndex={isToOne ? 0 : index}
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -93,6 +93,10 @@ export function IntegratedRecordSelector({
 
   const isAttachmentTable = tables.Collection.name.includes('Attachment');
 
+  const isInSearchDialog = React.useContext(SearchDialogContext);
+  const isAttachmentPlugin = viewName === 'ObjectAttachment';
+  const isAttachmentInSearch = isInSearchDialog && isAttachmentPlugin;
```
maxpatiiuk, line 98: while #4226 only mentioned attachments, I belive this restriction should be applied to all records.
in fact, SubViews should probably not even be rendered on the search forms. @grantfitzsimmons what do you think?

the way sp6 solved this, is that it allows entering -to-many fields on the search form (like addresses.city) as if it is a -to-one for the purposes of searching (this can be seen on the default search form). I did not implement the same in sp7 as that is ugly (a special edge case), and makes validating forms harder - instead query builder should be encouraged for more complicated searches

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
CarolineDenis: DECISION: 
- subview not displayed on search form

grantfitzsimmons: I've thought through this more, I agree.

maxpatiiuk (approved):

emenslin (approved):
Form is now read only and can no longer add attachments
Screenshot 2023-12-13 155948

pashiav (approved):
Looks good! Add and Remove functionalities are disabled.

https://github.com/specify/specify7/assets/118556884/dd0d2319-bd23-4933-965b-c2e2271c4ec6

PR #4257, opened 2023-12-05
Add an Add and Delete tree ranks option

### LINKED ISSUE #598: Add support for customizing tree definitions
There is no way to edit the tree definitions in 7.  
Especially problematic when you need to add/remove a tree level.

acwhite211:
Fixes #598 

### Checklist

### Testing instructions
- Open a tree 
- Click on the plus icon in the header, located between the pencil and the first rank (Planet | Earth | Root | ...) 
- Choose a parent rank 
- Click continue 
- Define the proprieties of your new rank 
- save 

- [x] verify that the new node has been added into the tree ranks 

- Open a tree 
- Click on the pencil icon in the header, located at the front of the headers, before the first rank (Planet | Earth | Root | ...) 
- Click on a pencil next to a tree rank 
- Click delete 
- Confirm deletion 

- [x] verify that the new node has been deleted from the tree ranks


specifyweb/specify/tree_views.py
```diff
@@ -342,4 +343,361 @@ def perm_target(tree):
...
+                "name": "newRankTitle",
+                "in": "formData",
+                "required": False,
+                "schema": {
+                    "type": "string"
+                },
+                "description": "The title of the rank to add (defaults to the name)"
+            },
+            {
+                "name": "useDefaultRankIDs",
```
maxpatiiuk, line 468: what is the use case for the front-end to pass this parameter? seems like your logic for determining new rank id is robust enough already without requiring additional input from the front-end

specifyweb/specify/tree_views.py
```diff
@@ -342,4 +343,361 @@ def perm_target(tree):
...
+                "name": "targetRankName",
+                "in": "formData",
+                "required": True,
+                "schema": {
+                    "type": "string"
+                },
+                "description": "The name of the parent rank to add the new rank to (use 'root' to add to the front)"
+            },
+            {
+                "name": "treeType",
```
maxpatiiuk, line 449: could you call this treeName instead (and for the other endpoint too).
that is the name front-end uses (rather than treeType)
(unless treeType is more common on the back-end)

specifyweb/specify/tree_views.py
```diff
@@ -342,4 +343,361 @@ def perm_target(tree):
...
+        tree_def_model_name = str.join(tree_type, 'treedef').lower().title()
+        tree_def_item_model_name = str.join(tree_type, 'treedefitem').lower().title()
+        tree_def_model = getattr(spmodels, tree_def_model_name)
+        tree_def_item_model = getattr(spmodels, tree_def_item_model_name)
+
+        # Make sure no nodes are present in the rank before deleting rank
+        rank = tree_def_item_model.objects.get(name=rank_name)
+        nodes_in_rank = tree_model.objects.filter(rank_id=rank.rank_id)
+        if nodes_in_rank is None or nodes_in_rank.count() > 0:
+            raise Exception("The Rank is not empty, cannot delete!")
```
maxpatiiuk, line 690: should this be something like `raise HttpBadRequest(...)` or something like that so that the front-end could catch this specific case and handle it gracefully (by proving a localized friendly error message to the user)

specifyweb/specify/tree_views.py
```diff
@@ -342,4 +343,361 @@ def perm_target(tree):
...
+
+        # Regenerate full names
+        tree_extras.set_fullnames(tree_def_model, null_only=False, node_number_range=None)
+
+        logger.info(f"Added new tree rank: {new_rank_name} with ID: {new_rank_id}")
+        return HttpResponse("Success")
+    
+    except Exception as e:
+        logger.error(f"Error in adding tree rank: {str(e)}")
+        return HttpResponseServerError(f"Failed: {str(e)}")
```
maxpatiiuk, line 619: do we normally wrap all api endpoints in try/catch?
doesn't HttpResponseServerError (or something like that) already get returned to the front-end if there is an uncaught error?
thus, what's the benefit of explicitly catching and re-throwing the error?

maxpatiiuk (changes requested):
excited to finally have this coming to sp7!
just hoping there isn't some dark magic/special logic going on in sp6 that we may be missing here

specifyweb/specify/tree_views.py
acwhite211: I could go either way with this.  I wanted to be able to handle the case where the client wants to use a default name, but not the default rank id.  It's also not a required field, it's just there to provide clarity to the fact that there are default ranks to the client.

specifyweb/specify/test_trees.py
```diff
@@ -208,4 +211,238 @@ def test_counts_correctness(self):
             for parent_id, correct in correct_results.items()
         ]
 
+class AddDeleteRanksTest(ApiTests):
+    def setUp(self) -> None:
+        super().setUp()
```
maxpatiiuk, line 216: if you don't define this method, won't it be inherited from the parent automatically, so this function seems useless? or am I misunderstanding how python works (I am not experienced with it)

specifyweb/specify/tree_views.py
```diff
@@ -343,4 +351,392 @@ def perm_target(tree):
...
+                        "type": "object",
+                        "description": "The request body.",
+                        "properties": {
+                            "newRankName": {
+                                "type": "string",
+                                "description": "The name of the new rank to add."
+                            },
+                            "parentRankName": {
+                                "type": "string",
+                                "description": "The name of the parent rank to add the new rank to (use 'root' to add to the front)."
```
maxpatiiuk, line 472: instead of 'root' magic string, can you let users pass `null`?

specifyweb/specify/tree_views.py
```diff
@@ -343,4 +351,392 @@ def perm_target(tree):
...
+                            },
+                            "treeName": {
+                                "type": "string",
+                                "description": "The name of tree."
+                            },
+                            "treeID": {
+                                "type": "integer",
+                                "description": "The ID of the tree (defaults to the first tree)."
```
maxpatiiuk, line 480: what do you mean defaults to the first tree? this argument should be required

specifyweb/specify/tree_views.py
```diff
@@ -343,4 +351,392 @@ def perm_target(tree):
...
+                            },
+                            "newRankTitle": {
+                                "type": "string",
+                                "description": "The title of the rank to add (defaults to the name)."
+                            },
+                            "useDefaultRankIDs": {
```
maxpatiiuk, line 486: this should just be true always, shouldn't it?

specifyweb/specify/tree_views.py
```diff
@@ -343,4 +351,392 @@ def perm_target(tree):
...
+                "application/json": {
+                    "schema": {
+                        "type": "object",
+                        "description": "The request body.",
+                        "properties": {
+                            "newRankName": {
+                                "type": "string",
+                                "description": "The name of the new rank to add."
+                            },
+                            "parentRankName": {
```
maxpatiiuk, line 470: (optional) use parentRankId instead?

specifyweb/specify/tree_views.py
```diff
@@ -343,4 +351,392 @@ def perm_target(tree):
...
+                                "type": "string",
+                                "description": "The name of the parent rank to add the new rank to (use 'root' to add to the front)."
+                            },
+                            "treeName": {
```
maxpatiiuk, line 474: why does this argument exists if treeId was provided?

specifyweb/specify/tree_views.py
```diff
@@ -343,4 +351,392 @@ def perm_target(tree):
...
+        "requestBody": {
+            "required": True,
+            "description": "Add rank to a tree.",
+            "content": {
+                "application/json": {
+                    "schema": {
+                        "type": "object",
+                        "description": "The request body.",
+                        "properties": {
+                            "newRankName": {
```
maxpatiiuk, line 466: instead of coming up with new arguments, could we just let front-end pass the entire <TreeName>TreeDefItem resource to this endpoint?

OR EVEN BETTER, just have front-end call the regular /api/specify/<TreeName>TreeDefItem endpoint, and then only call this endpoint with an id of the created rank id and the desired parent rank - that way, you don't duplicate code as a lot of the work would be handled by the /api/specify/<TreeName>TreeDefItem endpoint.

at which point, pretty much the only thing this endpoint would be doing is setting corect rank id and regenerating full_names - which is pretty much exactly what a `move_tree_rank` endpoint would look like, so we just got a feature for free, while reducing code duplication!

specifyweb/specify/tree_views.py
```diff
@@ -343,4 +351,392 @@ def perm_target(tree):
...
+            "200": {"description": "Success"},
+            "400": {"description": "Cannot delete tree rank"},
+            "500": {"description": "Server Error"}
+        }
+    },
+})
+@login_maybe_required
+@require_POST
+@transaction.atomic
+def delete_tree_rank(request, tree) -> HttpResponse:
```
maxpatiiuk, line 688: does this endpoint have to exist?
could you just use the `@orm_signal_handler('pre_delete', 'TaxonTreeDefItem')` (and post_delete, for each tree) to automatically clean up before and after tree rank deletion?

specifyweb/specify/tree_views.py
maxpatiiuk: oh, I just realized that a `parent` relationship exists in the TaxonTreeDefItem table - so this endpoint shouldn't exist at all, and the work should happen in the pre_save and post_save `orm_signal_handler`

the goal is not to add API endpoints with gazillion options, but to make it "just work" with the smallest API possible, to make it easier for anyone using the API

also, having these actions happen automatically in pre/post save/delete hook creates a "pit of success", where even if you don't know that special cleanup logic is needed for tree ranks, or are editing tree ranks directly though a Specify Form, rather than inside the Tree View, things would still "just work"

maxpatiiuk (changes requested):
back-end endpoints are overcomplicated

maxpatiiuk:
Could someone take a brief look at what magic sp6 might be doing when it adds/deletes a tree rank just to make sure we are not missing any business logic here?

specifyweb/specify/tree_views.py
```diff
@@ -343,4 +351,392 @@ def perm_target(tree):
...
+            "200": {"description": "Success"},
+            "400": {"description": "Cannot add tree rank with those parameters"},
+            "500": {"description": "Server Error"}
+        }
+    },
+})
+@login_maybe_required
+@require_POST
+@transaction.atomic
+def add_tree_rank(request, tree) -> HttpResponse:
```
maxpatiiuk, line 507: Could someone take a brief look at what magic sp6 might be doing when it adds a tree rank just to make sure we are not missing any business logic here?

specifyweb/specify/tree_views.py
maxpatiiuk, line 688: Could someone take a brief look at what magic sp6 might be doing when it deletes a tree rank just to make sure we are not missing any business logic here?

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
```diff
@@ -172,10 +197,50 @@ export function Tree<SCHEMA extends AnyTree>({
...
+                    ) : undefined}
+                    <Button.Icon
+                      icon="plus"
+                      title={treeText.addNewRank()}
+                      onClick={() => setIsAddingRank(true)}
+                    />
+                  </>
+                )}
+                {isAddingRank && (
+                  <Dialog
```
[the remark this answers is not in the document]
CarolineDenis: not really, the parent is selected when adding the new node, by example, you click on the + next to Continent, continent is the parent

specifyweb/specify/tree_views.py
CarolineDenis: how do you get the treeId in the frontend?

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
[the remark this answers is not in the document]
maxpatiiuk: ah, looking good - sorry, I misread the code

specifyweb/specify/tree_views.py
maxpatiiuk: two ways:
- the parent of the `Tree` component (TreeView) has a `treeDefinition` prop - from it, you can get tree id using `treeDefinition.id` (so need to pass that down as a prop to `Tree` component
- the tree component has an array from treeDefinitionItems. From there, you can get tree id by getting the id from the treeDef relationship: `const treeId = strictIdFromUrl(treeDefinitionItems[0].treeDef);`

CarolineDenis: thank you :)

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -0,0 +1,158 @@
...
+          onClose={() => setIsAddingParentRank(false)}
+        >
+          <Select
+            aria-label={treeText.chooseParentRank()}
+            className="w-full min-w-[theme(spacing.40)]"
+            value={parentRank}
+            onChange={({ target }): void => setParentRank(target.value || '')}
+          >
+            <option disabled value="">
+              {treeText.chooseParentRank()}
```
maxpatiiuk, line 129: use `<Label.Block>` instead of this hacky label)

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -0,0 +1,158 @@
...
+              >
+                {interactionsText.continue()}
+              </Button.Save>
+            </>
+          }
+          header={treeText.chooseParentRank()}
+          onClose={() => setIsAddingParentRank(false)}
+        >
+          <Select
+            aria-label={treeText.chooseParentRank()}
```
maxpatiiuk, line 123: aria-label won't be needed once `<Label.Block>` is used

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -0,0 +1,158 @@
...
+      />
+      {isAddingParentRank && (
+        <Dialog
+          buttons={
+            <>
+              <Button.DialogClose>{commonText.cancel()}</Button.DialogClose>
+              <Button.Save
+                onClick={() => {
+                  setIsAddingRank(true);
+                  setIsAddingParentRank(false);
```
maxpatiiuk, line 112: since these two states are mutually exclusive, why not create a state like this instead?
```ts
const [state, setState] = React.useState<'initial' | 'parent' | 'edit'>('initial');
```

specifyweb/businessrules/rules/tree_rules.py
```diff
@@ -17,9 +23,57 @@ def cannot_delete_root_treedefitem(sender, obj):
...
 
 @orm_signal_handler('pre_save')
 def set_is_accepted_if_prefereed(sender, obj):
     if hasattr(obj, 'isaccepted'):
         obj.isaccepted = obj.accepted_id == None
+
+def post_tree_rank_initiation(tree_def_item_model, new_rank):
+    tree_def = new_rank.treedef
+    parent_rank = new_rank.parent
+    new_rank_id = new_rank.rankid
```
maxpatiiuk, line 41: I think instead of trusting the rank id from the front-end here, that's where you should have your smart logic that determines new rank id
OR, we should move the rank id determination logic to the front-end

specifyweb/specify/tree_views.py
maxpatiiuk, line 1: Now that you have business rules (and they look good), there shouldn't be any changes in this file, right?
there is no need for separate special endpoints anymore as the front-end will just create or delete the resource like normal, and pre/post business rules will take care of the rest

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -0,0 +1,158 @@
...
+            treeResource as SpecifyResource<GeographyTreeDefItem>
+          ).get('isInFullName'),
+          fullNameSeparator: (
+            treeResource as SpecifyResource<GeographyTreeDefItem>
+          ).get('fullNameSeparator'),
+        },
+      }).then(() => globalThis.location.reload())
+    );
+    setIsAddingRank(false);
+  }
```
maxpatiiuk, line 95: none of this should be necessary - you should just let ResourceView save the resource like it would normally do, and the back-end business rules should take care of the rest
if there is a back-end business rule violation, back-end will throw, but the front-end should be able to handle that gracefully already thanks to the changes Jason did

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
maxpatiiuk, line 1: You should remove all *TreeDefItem tables from this list to let front-end create new resources of this type

https://github.com/specify/specify7/blob/6590bb8698361307f86773a45da782ef33753e75/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx#L39-L47

maxpatiiuk, line 1: could you also verify that you are able to delete a tree rank by clicking on the pencil icon next to tree rank, and from there on the delete button?
and then, the state should be updated to remove the deleted tree rank (or if that is tricky, simply reload the page)

maxpatiiuk, line 1: you should remove all of this to let front-end edit these fields:

https://github.com/specify/specify7/blob/6590bb8698361307f86773a45da782ef33753e75/specifyweb/frontend/js_src/lib/components/DataModel/schemaExtras.ts#L28-L50

maxpatiiuk, line 1: @acwhite211 back-end should also have a business rule that would rebuild the full names after a change to one of those fields (if you are able to detect what fields changed - otherwise, front-end will do it)

maxpatiiuk (changes requested):
getting better - good work

acwhite211:
I'm transferring the code from the api endpoints into the business rules.  Still got some more code cleanup to do.

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -28,128 +27,79 @@ export function AddRank<SCHEMA extends AnyTree>({
...
+      {state === 'parent' && (
         <Dialog
           buttons={
             <>
               <Button.DialogClose>{commonText.cancel()}</Button.DialogClose>
               <Button.Save
                 onClick={() => {
-                  setIsAddingRank(true);
-                  setIsAddingParentRank(false);
+                  const resourceParent = React.useMemo(
```
maxpatiiuk, line 55: remember rules of hooks 🚨🚨🚨

https://legacy.reactjs.org/docs/hooks-rules.html

maxpatiiuk: what's the point of this useMemo in here at all?

As a reminder, `useMemo` is to prevent re-execution of some expression during re-renders of the same component.
For react to know clearly what was the value of that hook during the previous render, all hooks must be called in the same order, and not inside a condition/callback/async

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -28,128 +27,79 @@ export function AddRank<SCHEMA extends AnyTree>({
...
-                {rank.name}
-              </option>
-            ))}
-          </Select>
+          <Label.Block className="gap-2">
+            {treeText.chooseParentRank()}
+            <Select
+              className="w-full min-w-[theme(spacing.40)]"
+              // Value={parentRank}
+              value={JSON.stringify(parentRank)}
```
maxpatiiuk, line 75: use parentRank.id instead

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -28,128 +27,79 @@ export function AddRank<SCHEMA extends AnyTree>({
...
+              onChange={({ target }): void => {
+                setParentRank(JSON.parse(target.value));
+              }}
+            >
+              {treeDefinitionItems.map((rank, index) => (
+                <option key={index} value={JSON.stringify(rank)}>
```
maxpatiiuk, line 81: use parentRank.id instead

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -28,128 +27,79 @@ export function AddRank<SCHEMA extends AnyTree>({
...
           onAdd={undefined}
-          onClose={(): void => setIsAddingRank(false)}
+          onClose={(): void => setState('initial')}
           onDeleted={undefined}
           onSaved={undefined}
           onSaving={() => {
-            addRank();
-            return false;
+            setState('initial');
+            globalThis.location.reload();
```
maxpatiiuk, line 102: onSaving is for intercepting the save - you should let the save go though like normal
which means, this logic has to be in onSaved rather than onSaving

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -28,128 +27,79 @@ export function AddRank<SCHEMA extends AnyTree>({
...
           title={treeText.addNewRank()}
           onAdd={undefined}
-          onClose={(): void => setIsAddingRank(false)}
+          onClose={(): void => setState('initial')}
           onDeleted={undefined}
           onSaved={undefined}
           onSaving={() => {
-            addRank();
-            return false;
+            setState('initial');
```
maxpatiiuk, line 101: > setState('initial');

likely useless line since you are reloading the page right afterward anyway

maxpatiiuk (changes requested):

acwhite211:
I pushed a commit with the unit tests, which have been changed to use the resource api, working now.  I have a lot of code cleanup to do before completing this PR.

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -96,11 +93,10 @@ export function AddRank<SCHEMA extends AnyTree>({
           onAdd={undefined}
           onClose={(): void => setState('initial')}
           onDeleted={undefined}
-          onSaved={undefined}
-          onSaving={() => {
-            setState('initial');
+          onSaved={() => {
             globalThis.location.reload();
           }}
+          onSaving={undefined}
```
maxpatiiuk, line 99: this prop is optional, so feel free to remove this line entirely

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
```diff
@@ -151,12 +152,18 @@ export function Tree<SCHEMA extends AnyTree>({
...
-                    onClick={handleToggleEditingRanks}
-                  />
+                  <>
+                    <Button.Icon
+                      aria-pressed={isEditingRanks}
+                      icon="pencil"
+                      title={treeText.editRanks()}
+                      onClick={handleToggleEditingRanks}
+                    />
+                    <AddRank
```
maxpatiiuk, line 162: `AddRank` should only be displayed if isEditingRanks

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
maxpatiiuk, line 1: sorry I didn't notice it earlier, but the `handleToggleEditingRanks` prop in the Tree component should be renamed to `onToggleEditingRanks` to be consistent with how we name callbacks in all the other places

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
maxpatiiuk, line 162: before displaying `AddRank`, you should check for `create` permission for the `treeDefinitionItems[0]._tableName` table

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -0,0 +1,104 @@
...
+            {treeText.chooseParentRank()}
+            <Select
+              className="w-full min-w-[theme(spacing.40)]"
+              value={parentRank}
+              onChange={({ target }): void => {
+                setParentRank(target.value);
+              }}
+            >
+              {treeDefinitionItems.map((rank, index) => (
+                <option key={index} value={rank.id}>
```
maxpatiiuk, line 78: (optional)
This is looking good!
optionally you could refactor this to use `resource_uri` instead of `id` as then you wouldn't need to do this in the Button.Save:

```ts
                  const resourceParent = getResourceApiUrl(
                    treeDefinitionItems[0]._tableName,
                    parentRank
                  );
```

acwhite211:
Everything seems to be working now.  Might make a few small code changes before marking this PR as ready for review.

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
```diff
@@ -48,7 +49,7 @@ export function Tree<SCHEMA extends AnyTree>({
   searchBoxRef,
   baseUrl,
   setLastFocusedTree,
-  handleToggleEditingRanks,
+  handleToggleEditingRanks: onToggleEditingRanks,
```
maxpatiiuk, line 52: no 🤦
this is the opposite of how we do it)

do you remember how we have it as `onToggleEditingRanks: handleToggleEditingRanks` everywhere?
looks like you swapped it accidentally here

specifyweb/specify/test_trees.py
acwhite211: This is needed when extending an inherited function.  So, the code from the inherited function runs, and then the rest of the code in this function runs.  If we wanted to overwrite the inherited function completely, this line would not be included.  If we just wanted to use the inherited function, then this function would not be needed.

specifyweb/specify/tree_views.py
acwhite211: So, in just about every database I have looked at, the Geographytreedef, Geologictimeperiodtreedef, Lithostrattreedef, Storagetreedefitem, and Taxontreedef all have just one record (one tree).  So, defaulting to the first tree, when the parameter is not given, seemed to be the usual case.  Now, with using the resource API, the front-end can set this value if needed.

acwhite211: I added treeDef on the front-end, so it will be passed in the tree rank resource API call.

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -0,0 +1,103 @@
...
+    treeDefinitionItems[0].resource_uri
+  );
+
+  const treeDef = treeDefinitionItems[0].treeDef;
+
+  const treeResource = React.useMemo(() => {
+    const resource = new tables[treeDefinitionItems[0]._tableName].Resource();
+    resource.set('treeDef', treeDef);
+    return resource;
+  }, [tableName, treeDef]);
```
maxpatiiuk, line 41: `tableName` dependency is not used inside of this useMemo

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
```diff
@@ -0,0 +1,103 @@
...
+                }}
+              >
+                {interactionsText.continue()}
+              </Button.Save>
+            </>
+          }
+          header={treeText.chooseParentRank()}
+          onClose={() => setState('initial')}
+        >
+          <Label.Block className="gap-2">
```
maxpatiiuk, line 68: (optional) use `<Form>` and `<Submit.Save>` here - that way, the Enter key would work automatically to submit the form

specifyweb/specify/build_models.py
```diff
@@ -63,9 +64,18 @@ def save(self, *args, **kwargs):
             return super(model, self).save(*args, **kwargs)
         except AbortSave:
             return
+        
+    def avant_delete(self):
```
maxpatiiuk, line 68: @melton-jason could you take a look at this?

maxpatiiuk, line 68: also, I'm not sure what does the name of this function mean
is this like `pre_delete`? or `pre_pre_delete`? 😊

specifyweb/specify/tree_extras.py
```diff
@@ -659,4 +659,21 @@ def renumber_tree(table):
...
+    Sptasksemaphore.objects.filter(taskname__in=tasknames).update(islocked=False)
+
+def is_instance_of_tree_def_item(obj):
+    from specifyweb.specify.models import (
+        Geographytreedefitem,
+        Geologictimeperiodtreedefitem,
+        Lithostrattreedefitem,
+        Storagetreedefitem,
+        Taxontreedefitem,
+    )
```
maxpatiiuk, line 671: not familiar with python, but is it ok performance-wise to import this inside of is_instance_of_tree_def_item?
is this function in the hot-path? would this cause performance issues?

or can we not import these in the global scope as they are not defined yet?

specifyweb/specify/tree_ranks.py
```diff
@@ -0,0 +1,277 @@
...
+def set_rank_id(new_rank):
+    """
+    Sets the new rank to the specified tree when adding a new rank.
+    Expects at least the name and parent of the rank to be set.
+    All the other parameters are optional.
+    """
+    # Get parameter values from data
+    tree = new_rank.specify_model.name.replace("TreeDefItem", "").lower()
+    new_rank_name = getattr(new_rank, 'name', None)
+    parent_rank_name = getattr(new_rank.parent, 'name', 'root') if getattr(new_rank, 'parent', None) else 'root'
```
maxpatiiuk, line 150: sad that this is using `name` rather than `rankid` or even better, `id`
oh well, not much we can do without a schema change((

maxpatiiuk, line 150: can we make sure that editing the parent rank name does not break the linking of the child rank?
do we need to have a business rule for this?

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
CarolineDenis: mhhh kook good to know

specifyweb/specify/build_models.py
acwhite211: lol, ya, I thought `pre_pre_delete` sounded dumb, so I instead used the french word "avant" meaning before, as to distinct itself from `pre_delete`.  I wrote a pretty lengthy description just to clarify its use.  I'll take any ideas for another name.

It is unfortunate that it needed to be done this way.  The problem was that Django enforces database constraints prior to the `pre_delete` signal.  So, Django returned errors before the `pre_delete` function could run.  Adding this new handler function was the simplest and most canonical way to incorporate it into the existing business rules code.  At least with this solution, we can easily extend the functionality to any other time this problem comes up.

https://github.com/specify/specify7/blob/3a548c3125d4917d26958256d89b340d752bbc67/specifyweb/specify/build_models.py#L68-L73

specifyweb/specify/tree_extras.py
acwhite211: This was done in order to prevent a circular import error.

Luckily it should be fine performance-wise, Python efficiently handles imports by reusing already imported modules instead of reloading them, so placing import statements inside functions should not significantly impact performance.

But, I think I might go ahead and just change it to `import specifyweb.specify.models as spmodels` instead.

specifyweb/specify/build_models.py
acwhite211: decided on going with `pre_constraints_delete` instead

maxpatiiuk (approved):
all good!
(besides https://github.com/specify/specify7/pull/4257#discussion_r1544051679)

acwhite211:
This PR is looking good now.  We'll wait to test and merge this PR until after 7.9.4 and 7.9.5 are out.

specifyweb/specify/tree_ranks.py
```diff
@@ -278,3 +279,41 @@ def set_rank_id(new_rank):
 
     # Set the new rank id
     new_rank.rankid = new_rank_id
+
+class RankOperation(Enum):
+    CREATED = 'created'
+    DELETED = 'deleted'
+    UPDATED = 'updated'
+
+def verify_rank_parent_chain_integrity(rank, rank_operation: RankOperation):
```
maxpatiiuk, line 288: this looks like a good check to guard against https://github.com/specify/specify7/pull/4257#discussion_r1544051679

but my question still remains:
- is such check necessary? i.e since rank is referring to it's parent by ID, does database enforce this integrity already, or do we have to do that manually using the verify_rank_parent_chain_integrity function. i.e could you use our back-end API as it stands right now (without verify_rank_parent_chain_integrity) to intentionally break the tree def integrity? if not, then this check is needless
- if this check is needed, in what cases is it needed? does any tree rank update cause issues? if so, we must handle the rank rename logic. or is it only needed on database migrations?

specifyweb/specify/tree_ranks.py
```diff
@@ -0,0 +1,319 @@
...
+            old_root_rank.save()
+
+    # Regenerate full names
+    tree_extras.set_fullnames(tree_def, null_only=False, node_number_range=None)
+
+def pre_tree_rank_deletion(tree_def_item_model, rank):
+    tree_def = rank.treedef
+
+    # Make sure no nodes are present in the rank before deleting rank
+    if tree_def_item_model.objects.filter(parent=rank).count() > 1:
```
realVinayak, line 123: nitpick: use exists()

emenslin (changes requested):
### Testing instructions
- Open a tree 
- Click on the plus icon in the header, located between the pencil and the first rank (Planet | Earth | Root | ...) 
- Choose a parent rank 
- Click continue 
- Define the proprieties of your new rank 
- save 

- [x] verify that the new node has been added into the tree ranks 

- Open a tree 
- Click on the pencil icon in the header, located at the front of the headers, before the first rank (Planet | Earth | Root | ...) 
- Click on a pencil next to a tree rank 
- Click delete 
- Confirm deletion 

- [x] verify that the new node has been deleted from the tree ranks 

Looks good, however I think there should be some sort of deletion blocker when you try to delete a rank that has nodes in it instead of giving an error. 

Current behavior:

https://github.com/specify/specify7/assets/150940840/bc3a31e3-293f-4a90-a9ed-db883d2d8db6

Deletion blocker could just look something like this 
Screenshot 2024-05-06 151820

acwhite211:
Ya, that error should be occurring, because there is a taxon in the 'Test' Rank.  The Rank needs to be empty of taxon before being deleted.  We should add a check on the front-end to make sure this is the case before calling the delete api.  Ya, that example for the delete blocker pop-up would be a good idea in this instance.

I push a commit to add an end point that would help out with this `/api/specify_tree/<tree>/<rankid>/tree_rank_item_count/`.  It just returns the number off tree items belonging to a rank.  This is a simple solution for the front-end to check.  We could also just form the query on the front-end, either works.

specifyweb/specify/tree_ranks.py
```diff
@@ -0,0 +1,345 @@
...
+        if new_rank_id:
+            return
+
+    # Determine the new rank id parameters
+    new_rank_id = getattr(new_rank, 'rankid', None)
+    tree_def = tree_def_model.objects.get(id=tree_id)
+    try:
+        tree_def = tree_def_model.objects.get(name=tree_name)
+    except tree_def_model.DoesNotExist:
+        pass
```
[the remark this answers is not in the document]
realVinayak: in fact, it is the exact query as before, without any change. get rid of this,

specifyweb/specify/tree_ranks.py
```diff
@@ -0,0 +1,345 @@
...
+        for child_rank in child_ranks:
+            child_rank.parent = new_rank
+            child_rank.save()
+
+    # Update the old root rank to point to the new root rank if the new rank is the new root rank
+    if new_rank.parent is None:
+        old_root_rank = tree_def_item_model.objects.exclude(id=new_rank.id).filter(treedef=tree_def, parent=None).first()
+        if old_root_rank is not None:
+            old_root_rank.parent = new_rank
+            old_root_rank.save()
```
realVinayak, line 140: ha, what use cases does this has? the only case where this branch is taken is if you are adding the same rankid as a previous rank, and the ranks both have the same parent as None. 

instead, simplify things by doing an exclude(id=new_rank.id) instead of exclude(rankid=new_rank_id), above and then getting rid of the lower branch.

specifyweb/specify/test_trees.py
realVinayak: hmm but you're not extending the inhereted function. read-up on what these methods do.

realVinayak (changes requested):
Getting there. critical of all the signal stuff though.

specifyweb/specify/test_trees.py
realVinayak: See this unit test:

https://github.com/specify/specify7/blob/df8a781f9611fb4ab2ffe79cc66b01529b4408bf/specifyweb/stored_queries/tests.py#L91-L109

specifyweb/businessrules/rules/tree_rules.py
```diff
@@ -2,13 +2,28 @@
 
 from specifyweb.businessrules.orm_signal_handler import orm_signal_handler
 from specifyweb.businessrules.exceptions import TreeBusinessRuleException
+from specifyweb.specify.tree_extras import is_instance_of_tree_def_item
+from specifyweb.specify.tree_ranks import *
 
 logger = logging.getLogger(__name__)
 
+@orm_signal_handler('pre_save')
```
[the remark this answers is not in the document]
melton-jason: +1 on using an inheritance-based approach for instances of treedefitem! 
Will be even better when #4764 is implemented and IDEs can easily take you to the method definition. 

Although, if we ever intend on deleting treedefitems via `queryset.delete()` or through a cascade, then we should probably maintain `pre_delete`/`post_delete` signals: 

> Overridden model methods are not called on bulk operations
>
> Note that the delete() method for an object is not necessarily called when deleting objects in bulk using a QuerySet or as a result of a cascading delete. To ensure customized delete logic gets executed, you can use pre_delete and/or post_delete signals.

https://docs.djangoproject.com/en/3.2/topics/db/models/#overriding-model-methods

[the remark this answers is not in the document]
acwhite211: Ya, I should of mentioned this in the comments.  My plan is use an inheritance approach after the #4358 issue is merged.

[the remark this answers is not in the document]
realVinayak: wait, what's the difference in doing it right now, using inheritance, and doing it later, via inheritance, again? if you just use inheritance now, you can subclass later treeitems once https://github.com/specify/specify7/issues/4358 has been merged. BUT, you can also do it now.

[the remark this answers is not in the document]
realVinayak: that is, my point is that there is no difference between doing it now, and once schema independence has been achieved.

specifyweb/specify/tree_ranks.py
```diff
@@ -0,0 +1,345 @@
...
+
+from specifyweb.businessrules.exceptions import TreeBusinessRuleException
+from . import tree_extras
+from . import models as spmodels
+from sys import maxsize
+
+import logging
+logger = logging.getLogger(__name__)
+
+TAXON_RANKS = {
```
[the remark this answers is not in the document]
acwhite211: "Shouting", LOL, THIS_IS_A_CONSTANT https://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations.  If it gets used in other files, I'll move it to it's own file, otherwise it's good here.  Specify6 is implemented similarly https://github.com/specify/specify6/blob/945d28f00288d2d64979848fb2fb63fda34bdd1b/src/edu/ku/brc/specify/datamodel/TaxonTreeDef.java#L48

specifyweb/specify/views.py
```diff
@@ -91,6 +92,18 @@ def raise_error(request):
...
+    if not is_instance_of_tree_def_item(obj):
+        return
+    # Filter out blocker that is the child tree rank of tree rank being deleted.
+    # This is handled by the business rules when deleting a tree rank, the parent is set to the grandparent.
+    delete_blockers[:] = list(
+        filter(
+            lambda db: db['field'] != 'parent' and db['table'] == type(obj).__name__, 
+            delete_blockers
+        )
+    )
```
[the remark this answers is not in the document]
acwhite211: Ya, I agree that it should be done on the front-end.  I just put this in as an alternative solution for Caroline.  Not going to keep this code after we implement the front-end solution.  See my earlier comment:

 > Ya, that error should be occurring, because there is a taxon in the 'Test' Rank. The Rank needs to be empty of taxon before being deleted. We should add a check on the front-end to make sure this is the case before calling the delete api. Ya, that example for the delete blocker pop-up would be a good idea in this instance.

specifyweb/businessrules/rules/tree_rules.py
```diff
@@ -17,9 +32,26 @@ def cannot_delete_root_treedefitem(sender, obj):
...
                      "id": obj.id
                  }})
+        elif is_tree_rank_empty(sender, obj):
+            raise TreeBusinessRuleException(
+                "cannot delete tree rank containing items",
+                {"tree": obj.__class__.__name__,
+                 "localizationKey": 'deletingNonEmptyTreeRank',
+                 "node": {
+                     "id": obj.id
+                 }})
```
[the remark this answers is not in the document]
acwhite211: Ya, it should be `not is_tree_rank_empty`.  This is part of the last commit I quickly made for the rank delete blocker.  This code will get removed shortly with a front-end solution.

specifyweb/specify/tree_ranks.py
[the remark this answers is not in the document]
realVinayak: maybe be it was unclear from me. the variable is fine as all caps. don't see value in keeping rank names all caps. minor nitpick, so wouldn't block.

specifyweb/specify/build_models.py
```diff
@@ -97,6 +107,12 @@ def protect(collector, field, sub_objs, using):
     'Spappresourcedata.spappresource': models.CASCADE,
     'Spappresourcedata.spviewsetobj': models.CASCADE,
     'Spreport.appresource': models.CASCADE,
+
+    'Geographytreedefitem.parent': models.DO_NOTHING,
+    'Geologictimeperiodtreedefitem.parent': models.DO_NOTHING,
+    'Lithostrattreedefitem.parent': models.DO_NOTHING,
+    'Storagetreedefitem.parent': models.DO_NOTHING,
+    'Taxontreedefitem.parent': models.DO_NOTHING,
```
realVinayak, line 115: are these necessary here? seems excessive (bc some other code could naively perform a delete and django wouldn't say no). Since it doesn't affect the database itself, this doesn't cause a problem. 

however, apparently the delete blockers do not report parent relationship bc of this (even after removing the filtering logic). Would a cleaner way be to remove these, and have filtering logic on the frontend?

realVinayak: On second thought, this seems fine, bc we just handle it.

emenslin (approved):
### Testing instructions
- Open a tree 
- Click on the plus icon in the header, located between the pencil and the first rank (Planet | Earth | Root | ...) 
- Choose a parent rank 
- Click continue 
- Define the proprieties of your new rank 
- save 

- [x] verify that the new node has been added into the tree ranks 

- Open a tree 
- Click on the pencil icon in the header, located at the front of the headers, before the first rank (Planet | Earth | Root | ...) 
- Click on a pencil next to a tree rank 
- Click delete 
- Confirm deletion 

- [x] verify that the new node has been deleted from the tree ranks 

--- 
Deletion Blocker looks great!

https://github.com/specify/specify7/assets/150940840/78cf6868-849e-4ccd-a1eb-a021e746cbfe

Areyes42 (approved):
### Testing instructions

-   Open a tree
    
-   Click on the plus icon in the header, located between the pencil and the first rank (Planet | Earth | Root | ...)
    
-   Choose a parent rank
    
-   Click continue
    
-   Define the proprieties of your new rank
    
-   save
    
-  [x]  verify that the new node has been added into the tree ranks
    
-   Open a tree
    
-   Click on the pencil icon in the header, located at the front of the headers, before the first rank (Planet | Earth | Root | ...)
    
-   Click on a pencil next to a tree rank
    
-   Click delete
    
-   Confirm deletion
    
-   [x] verify that the new node has been deleted from the tree ranks
    

* * *
Adding and deleting works as expected and the deletion blocker shows up, great job!

Screenshot 2024-05-16 at 4 00 43 PM

specifyweb/specify/tree_ranks.py
```diff
@@ -0,0 +1,314 @@
...
+def set_rank_id(new_rank):
+    """
+    Sets the new rank to the specified tree when adding a new rank.
+    Expects at least the name, parent, and tree_def of the rank to be set.
+    All the other parameters are optional.
+    """
+    # Get parameter values from data
+    tree = new_rank.specify_model.name.replace("TreeDefItem", "").lower()
+    new_rank_name = getattr(new_rank, 'name', None)
+    parent_rank_name = getattr(new_rank.parent, 'name', 'root') if getattr(new_rank, 'parent', None) else 'root'
```
realVinayak, line 159: you need to be clear on what you are getting in attr.
```
  new_rank_name = getattr(new_rank, 'name', None)
  parent_rank_name = getattr(new_rank.parent, 'name', 'root') if getattr(new_rank, 'parent', None) else 'root'
```
you don't want that. you want
```
    new_rank_name = new_rank.name
    parent_rank_name = new_rank.parent.name if new_rank.parent else 'root'
```
think why. we know these fields must exist, don't overcomplicate it.

specifyweb/specify/tree_ranks.py
```diff
@@ -0,0 +1,345 @@
...
+        )
+
+        if is_default_rank_id_unused and is_placement_valid:
+            new_rank_id = default_rank_id
+
+    # Set the new rank id if a default rank id is not available
+    if new_rank_id is None:
+        # If this is the first rank, set the rank id to the default increment
+        if is_tree_def_items_empty:
+            new_rank_id = rank_increment
```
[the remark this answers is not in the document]
acwhite211: So that there will be room to create smaller ranks if needed.  Using a rank ID of zero should not be done unless one is 100% that they will never create a lower rank.

specifyweb/businessrules/rules/tree_rules.py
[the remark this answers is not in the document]
acwhite211: I have planned a follow-up issue to the Django schema issue that will do a lot of inheritance and other optimizations to the new models code.  I think it will be cleaner to do those changes in that PR.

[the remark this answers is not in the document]
realVinayak: I don't believe it will be "cleaner". Can you expand on some specifics on why it will be? Curious about them. I think it'll be exactly same.

Trees work. they've been working for 10 years using inheritance, without schema independence, in a lot more aggressive environments like workbench. Anyways, won't block this PR bc of it.

specifyweb/specify/test_trees.py
acwhite211: This was just leftover code from refactoring that needed to be removed.

specifyweb/frontend/js_src/lib/components/TreeView/AddRank.tsx
acwhite211: There is a python Django library "django-model-utils" I've been looking at the detects the specific fields that have been changed.  Will look into it in the future, it would be a nice tool to have in the back-end.

specifyweb/businessrules/rules/tree_rules.py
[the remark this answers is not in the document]
acwhite211: I went ahead and implemented an abstract TreeRank class with the save and delete logic override.  I'm keeping the pre_constraints_delete system around because it might be useful in the future.

PR #4264, opened 2023-12-08
Update attachment galery on record changes

### LINKED ISSUE #4246: Browsing Forms Attachment Viewer doesn't recognize saved image
**Describe the bug**
From query results or record sets, the new attachment viewer icon in the browsing header doesn't show a newly saved image.

https://github.com/specify/specify7/assets/135047322/5fa8413d-62e6-4e49-8840-58f49c3f7892

**To Reproduce**
1. Go to query results or record set without attachments and get to the "browse" view
2. Add an attachment to one of the records
3. Save
4. Click the attachment viewer icon in header
5. See error

**Expected behavior**
Once an image is saved to the record, the record set attachment viewer should recognize and display those images 

**Desktop:**
 - OS: macOS Sonoma
 - Browser: Chrome
 - Specify 7 Version: edge

**Database Name:** `AAFC_AAC_3_8_23`

CarolineDenis:
Fixes #4246

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
@@ -184,6 +184,15 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
 
   const hasAttachments = tablesWithAttachments().includes(model);
 
+  const [trackChanges, setTrackChanges] = React.useState(0);
+  const hasChanged =
+    resource?.changed !== undefined &&
+    Object.keys(resource?.changed).length > 0;
+  React.useEffect(() => {
+    if (hasChanged)
+      setTrackChanges((previousTrackChanges) => previousTrackChanges + 1);
+  }, [hasChanged]);
+
   return (
     <>
       <ResourceView
@@ -224,7 +233,11 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
                 />
               )}
               {hasAttachments && (
-                <RecordSetAttachments records={records} onFetch={handleFetch} />
+                <RecordSetAttachments
+                  records={records}
+                  trackChanges={trackChanges.current}
+                  onFetch={handleFetch}
+                />
               )}
               {specifyNetworkBadge}
             </div>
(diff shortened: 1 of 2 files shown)


CarolineDenis:
NOTES: 

- we might wait to fix this once we are moving away from Backbone

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
```diff
@@ -184,6 +184,35 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
...
+
+  React.useEffect(() => {
+    if (attachmentObject !== undefined && attachmentObject.length > 0) {
+      setAttachmentIsPopulated(true);
+    }
+
+    if (hasChanged) {
+      setTrackChanges((previousTrackChanges) => previousTrackChanges + 1);
+    }
+  }, [hasChanged, attachmentObject]);
```
maxpatiiuk, line 214:
```suggestion
const attachmentsCollection = resource?.getDependentResource(
  'collectionobjectattachments'
);
const trackChanges = attachmentsCollection.length;
```

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
maxpatiiuk: this might or might not work. if it won't, then you have to add a useEffect that would add the 'change' listener on the attachmentsCollection and inside even handler update the value of trackChanges (trackChanges would be converted to a useState)

CarolineDenis: unfortunately it doesn't work. I also tried with a useEffect but the changes to the attachments doesn't trigger the resource dependencies in the use effect, 
I also tried to create a ref with the length of attachmentsCollection and check if the ref is equal to the new ref and depending on that updating trackChanges but the issue is that the changes appear as soon as we click the + button to add an attachment instead of when the user save the attachment.

maxpatiiuk: yeah, two reasons the above didn't work:
- instead of hardcoding collectionobjectattachments, should use the <tableName>attachments
- since backbone resource changes don't trigger re-render, the code does not get re-executed when needed, thus you see stale state

solution would be:
- wait for migration away from backbone to make this all simple
- or add a useEffect that would be listening for collection to be created (because it may not exist right away, especially if you are opening the form for a new resource - http://localhost/specify/view/accession/new/), and then listen for changes to that collection, and on changes set the trackChanges state

PR #4266, opened 2023-12-08
Scope tree def items in query builder

realVinayak:
Why wasn't this done before? oh well

To test:
1. Use a database that has multiple trees for every tree table.
2. Make a query in the query builder using TREEdefitem as base table, and choose any field. Choose 'name' for testing.
3. Check that production gives all the tree ranks. For example, you will find "Life" rank twice in production. Verify that only tree ranks in your scope will be visible.
For example, first is production, second is this branch
image
image

Note:
When testing in context of statistics, remove the taxa represented category and the 'error' stats. You shouldn't be able to add those items back when going to Add Stats Dialog

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Statistics/hooks.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Statistics/hooks.tsx
@@ -278,7 +278,6 @@ export function resolveStatsSpec(
     return {
       type: 'QueryStat',
       querySpec,
-      pathToValue: item.pathToValue,
     };
   }
   if (statSpecItem.spec.type === 'QueryStat')
@@ -293,21 +292,7 @@ export function useResolvedStatSpec(
   item: CustomStat | DefaultStat,
   formatterSpec: StatFormatterSpec
 ): BackEndStatResolve | QueryBuilderStat | undefined {
-  const rawStatsSpec = React.useMemo(
-    () => resolveStatsSpec(item, formatterSpec),
-    [item]
-  );
-  const statsSpecRef =
-    React.useRef<ReturnType<typeof useResolvedStatSpec>>(undefined);
-  if (rawStatsSpec !== undefined) {
-    if (rawStatsSpec.pathToValue === undefined) {
-      statsSpecRef.current = undefined;
-    } else {
-      statsSpecRef.current ??= rawStatsSpec;
-    }
-  }
-
-  return statsSpecRef.current ?? rawStatsSpec;
+  return React.useMemo(() => resolveStatsSpec(item, formatterSpec), [item]);
 }
 
 /**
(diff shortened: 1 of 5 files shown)


melton-jason (approved):
> Why wasn't this done before? 

Great question! :sweat_smile:

CarolineDenis (approved):

bronwyncombs (changes requested):
Can still add error stats back from dialog: 

https://github.com/specify/specify7/assets/135047322/d2fa03c7-996e-40ff-b528-1bcd33e187c2


Edge:
Screenshot 2023-12-08 at 3 44 53 PM
This branch:
Screenshot 2023-12-08 at 3 44 36 PM

melton-jason:
The reason the ERROR stats are present in the dialog is because they still exist in the `CollectionPreferences` AppResource. 
More specifically, these statistics were generated sometime between the inclusion of tree statistics and before this pull request, so they still contain the ERROR stats. 

Thus for newly generating statistics the erorr in https://github.com/specify/specify7/pull/4266#pullrequestreview-1773221485 should not occur. 

This should not affect many users as the Tree statistics have only ever been on `production` and tested internally: those on `v7.9.1` and previous releases will be unaffected. 

@specify/dev-testing, while there is likely a way to hardcode ignoring these ERROR tree stats it may be more prudent to recommend those affected to either manually remove those stats from  `CollectionPreferences` (if changes have been made to the stats) or completely reset Statistics (If stats are close to how they were by default by using the Reset button on the stats page or deleting the AppResource)

- - - 

However, there is another concern. 
If a user's statistics page was generated before the inclusion of tree stats (i.e., in `v7.9.1`), then a default`Taxa Represented` page will not be added. The stats will exist as defaults in the dialog and have to be manually added if the user wishes to see them. 
(Alternatively, the user could regenerate the statistics page at the expense of any modifications that had been made)

@specify/ux-testing, what are you thoughts on this?

melton-jason:
From an internal discussion regarding https://github.com/specify/specify7/pull/4266#issuecomment-1850669241

- > If a user's statistics page was generated before the inclusion of tree stats (i.e., in v7.9.1), then a defaultTaxa Represented page will not be added. The stats will exist as defaults in the dialog and have to be manually added if the user wishes to see them.
(Alternatively, the user could regenerate the statistics page at the expense of any modifications that had been made)

  - We knew this would be an issue when we agreed to delay including tree stats. We decided that users could easily add them from the menu if they wanted
  
- - - 
  
- > while there is likely a way to hardcode ignoring these ERROR tree stats it may be more prudent to recommend those affected to either manually remove those stats from CollectionPreferences (if changes have been made to the stats) or completely reset Statistics (If stats are close to how they were by default by using the Reset button on the stats page or deleting the AppResource)

  - These recommendations still stand. For now, if there are any affected users they should either manually remove the statistics from the CollectionPreferences or reset/regenerate Statistics.

melton-jason (approved):

PR #4271, opened 2023-12-09
fix(Syncer): make findNodePosition handle / in attribute names

### LINKED ISSUE #3678: Some XML Editor errors appear on a wrong line
This error is supposed to be shown down below next to one of the `class="..."` attributes, not on top of the file:

Screenshot 2023-04-25 at 12 17 48

The code for positioning the errors does this when it fails to find the place where to put the error message.

Note, this error may be connected to the presence of `<?xml  ...` in this file (though, it should have been able to handle that)

Note, I have tests for error placement, so if you fix this bug, remember to add the test case to prevent it in the future

maxpatiiuk:
Fixes #3678

Debugging process:
- findNodePosition is called with the raw XML string and a path to an error, and it's job is to find the position in the string at which to show the error underline:
   https://github.com/specify/specify7/blob/xml-editor/specifyweb/frontend/js_src/lib/components/Syncer/findNodePosition.ts#L10
- it does that by stripping the XML declaration if present, and then calls findElement, which reduces over the path, progressively getting closer and closer to the position at which the error occurs - until it reaches the end of the path, at which point it would return the resulting position
- when iterating over the path, I noticed that the findChildPosition function returned unexpected results:  
Screenshot 2023-12-08 at 21 19 43
  it's the responsibility in this case was to find the start and end index of the 17th "<format>" element in the XML file, but it returned position 0 - incorrect

  I realize that the file is doing pretty complicated things, and thus more comments explaining what's happening is a good idea - fixed that

- at this point I narrowed down the problem to findChildPosition, and so do debugging inside of it - reading the code to understand what it does, and putting different breakpoints.

  findChildPosition calls xmlStringTraverse (indirectly).

  after debugging the xmlStringTraverse I found the problem:

  the regex that is used to match opening/closing XML tag fails if XML tag contains an attribute with `/` in it:

  https://github.com/specify/specify7/blob/xml-editor/specifyweb/frontend/js_src/lib/components/Syncer/findNodePosition.ts#L184

   Screenshot 2023-12-08 at 21 50 28

- Now add a smallest test case that reproduces the issue:

  ```ts
  theories(findChild, [
    // ...
    { in: ['<b></b><b attribute="/"></b><b></b></a>', 'b', 1], out: 7 },
  ]);
  ```

- Then use https://regex101.com/ to construct a better regex that matches correctly:

  ```ts
  const reTag = /<!--|<!\[CDATA\[|<[\s\S]*?>/gu;
  ```

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Syncer/findNodePosition.ts
+++ b/specifyweb/frontend/js_src/lib/components/Syncer/findNodePosition.ts
@@ -210,7 +210,7 @@ export function xmlStringTraverse<T>(
   return undefined;
 }
 
-const reTag = /<!--|<!\[CDATA\[|<\/?[^/>]+\/?>/gu;
+const reTag = /<!--|<!\[CDATA\[|<[\s\S]*?>/gu;
 
 export const exportsForTests = {
   findChild,
--- a/specifyweb/frontend/js_src/lib/components/Syncer/findNodePosition.ts
+++ b/specifyweb/frontend/js_src/lib/components/Syncer/findNodePosition.ts
@@ -7,7 +7,13 @@ type Position = {
   readonly to: number;
 };
 
+/**
+ * "findNodePosition" is called with the raw XML string and a path to an
+ * error, and it's job is to find the position in the string at which to
+ * show the error underline
+ */
 export function findNodePosition(xml: string, path: RA<LogPathPart>): Position {
+  // Strip xml declaration if present
   const declaration = xml.startsWith('<?') ? xml.indexOf('>') + 1 : 0;
   const element = findElement(xml.slice(declaration), path);
   return {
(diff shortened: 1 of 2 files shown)


CarolineDenis (approved):

specifyweb/frontend/js_src/lib/components/Syncer/findNodePosition.ts
```diff
@@ -91,21 +107,18 @@ function findChildPosition(
...
 /** Find the next nth instance of a direct child with a given tagName */
 function findChild(
   xml: string,
   tagName: string,
-  index: number
+  targetIndex: number
 ): number | undefined {
-  let currentIndex = 0;
+  let index = 0;
+  debugger;
```
melton-jason, line 119: Looks like you forgot to remove this debugger statement

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/Syncer/findNodePosition.ts
maxpatiiuk: good catch!
btw, i am often finding `debugger;` nicer than setting breakpoints in the dev tools. reasons:
- dev tools breakpoints may shift when you edit code - `debugger;` remain in place you want them in
- easier to write conditional breakpoints `if(someCondition) debugger;` - you can do conditional in dev tools, but you don't get typescript validation or an error message if you make a syntax error
  - if code was minified, the variable names in source map don't match the executed javascript - not an issue if your conditional breakpoint is defined in the code

but of course, you can then forget to remove the debugger statement - thus I could add an eslint rule for this

PR #4273, opened 2023-12-10
Populate rankid with next enforced rank when adding tree node

### LINKED ISSUE #3597: Auto-populate rank field when adding new tree node
Auto-fill the next lowest rank when adding a child node to a tree. This should be in the required rank field 

Could allow the option to choose from the pick list while still auto-populating with the next most logical option
Could add auto-population as a boolean in preferences
Could allow for user override settings for certain ranks in preferences

**Additional context**
Was available in 6 and is being requested

melton-jason:
Fixes #3597

### Testing Instructions
- Click on an existing tree node on any tree
- Use the `Add Child` tree action (displayed as a :heavy_plus_sign:)
- The rank (definitionitem) picklist should automatically populate with the next enforced rank in the tree

If is it unknown which ranks are enforced, perform a query on `{TREE}treedefitem` where {TREE} is the name of the tree (i.e. taxontreedefitem) and add the `isenforced` field to the query. 

Untill #4266 is merged, if there is more than one discipline, it would be helpful to narrow the results to tree items only in the desired discipline. 
This can be accomplished through `treedef -> discipline` when {TREE}treedefitem is the base table.


CarolineDenis (approved):

specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
```diff
@@ -40,8 +40,8 @@ const fetchPossibleRanks = async (
       // Remove ranks after enforced rank
       return (enforcedIndex === 0 ? ranks : ranks.slice(0, enforcedIndex)).map(
         (rank) => ({
-          value: rank.id.toString(),
-          title: rank.title || rank.name,
+          value: rank.resource_uri,
+          title: rank.title ?? rank.name,
```
maxpatiiuk, line 44: you might want to still use `rank.title || rank.name` here instead of `rank.title ?? rank.name` because .title could be empty string

if eslint complains, could do it explicitly: `rank.title.length === 0 ? rank.name : rank.title`

maxpatiiuk (changes requested):
thanks for fixing this!

specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
maxpatiiuk: related though:
`||` is a useful syntax shortcut, but may lead to bugs if not used carefully

thus, should we continue to not allow it, or relax the eslint rule?

melton-jason: Fixed in a1716250469e0f442d56ca31dbfa82b9397f9be5

As far as allowing `||` or not is concerned, I would personally prefer to be explicit most of the time. It is probably better to have slightly more code that is more robust over shorter code which may potentially be buggy. 
Also, incentivizing the use of `||` may encourage more developers to "cut corners" and use the shortcut carelessly.

Although, the frequency of bugs caused by `||` i'm sure is relatively low, and sometimes the behavior of the implicit type cast _is_ what is desired, so there would be reason beyond a strict rule to not use the shortcut.

Regardless, it's clear both approaches have their pros and cons so I don't have a strong opinion either way.

maxpatiiuk: in our team at Esri, they found more luck in encouraging people to write tests than enforcing people to use certain code style (though tests are still not always written) because people are too opinionated about the code style OR don't care enough about using consistent code style (some people care about shipping features more than having code look a certain way - both sides have their merits).

my personal opinion is that writing code is more fun than writing tests, so if you can use strict ESLint config and strict typing (with TypeScript) to reduce the need for writing tests (while also enforcing higher-quality code), then that's great.

plus, having TypeScript and ESLint catch and report issues before PR is created is great as it reduces PR review time. not to mention many benefits of consistency in the codebase

I would like to spend some time in the coming months improving our ESLint config, 
after 7.9.4 is out, I would go about fixing all remaining ESLint errors, and enabling 0 ESLint error enforcement in Github action.
Important context about this:
- I would disable/loosen ESLint rules that in the past 2 years have been more of a hindrance/annoyance/noise than actual help
  - example of noise is when we have a linting rule discouraging creating classes, having a linting rule discouraging the use of `this` is mostly redundant
  - some rules are designed to prevent bugs, that aren't likely to happen thanks to typescript (i.e the rule requiring that each .then() returns something), thus are redundant/hindrance
  - while I liked the "writing-good-comments" eslint rules in theory, they weren't too helpful in practice. I.e resolving the rule about "no passive voice in comments" often required making comments more verbose (or we just ignored the errors suggested by this rule)
- I would look for more useful rules to enable
- I would ask for more input and feedback before imposing the rules on others

@melton-jason @CarolineDenis @realVinayak thoughts about all of this?

maxpatiiuk (approved):

emenslin (approved):
Rank is enforced, looks good!

https://github.com/specify/specify7/assets/150940840/eebc9470-bfad-4026-b86b-bc772cf08061

specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
melton-jason: > my personal opinion is that writing code is more fun than writing tests, so if you can use strict ESLint config and strict typing (with TypeScript) to reduce the need for writing tests (while also enforcing higher-quality code), then that's great.

I agree in this case. 
Perhaps in an ideal world, there would be much more unit tests written to cover all possible cases, but this is not very practical both on time and developer interest.
Rules which are directly enforced and thus difficult to ignore will hopefully cover cases of carelessness, ignorance/inexperience, or the occasional honest slip up. 
(However, ESLint rules and type checking alone are definitely not a stand-in for unit tests)

These are also great points, which I second:
> plus, having TypeScript and ESLint catch and report issues before PR is created is great as it reduces PR review time. not to mention many benefits of consistency in the codebase

However as with Typescript's type checking, there are times when ESLint rules can get in the way. The examples you proposed could probably be disabled or at very east loosened, and other rules could be looked at as well. 
It would probably be worth the time to go over them, at least.

maxpatiiuk: > However as with Typescript's type checking, there are times when ESLint rules can get in the way. The examples you proposed could probably be disabled or at very east loosened, and other rules could be looked at as well.
It would probably be worth the time to go over them, at least.

This is what I and doing at Esri this time. I am tasked with creating a strict, but smart ESLint rule set that is geared toward discovering as many bugs as possible, while being least annoying, and not slowing down editor/tests much. Then, I would also provide justification for each rule and do discussions with wider team at Esri. Eventually, this may become the standard rule set for all TypeScript devs at Esri.
Of course, the rule set I compile together can benefit the Specify rule set too

specifysoftware:

Issue #4225, opened 2023-11-21
`xml-editor`: Date fields not working as expected


grantfitzsimmons:
**Steps to Recreate:**
1. Go to a form with a date field (make sure accessible date picker is turned on in preferences)
2. Try to type in a date ending in `2021`
3. See that it is not possible

Arc Browser, macOS Sonoma

## How it works on `xml-editor`:

https://github.com/specify/specify7/assets/37256050/bd0b529f-6e11-469e-a241-95cca379e341

## How it should work (`v7.9.1`):

https://github.com/specify/specify7/assets/37256050/fadd6498-c1ad-40f5-9c9b-5afeacd70342

CarolineDenis:
NOTES: 
When selecting precision 'year' there is an issue as well.

PR #4276, opened 2023-12-11
refactor(PartialDateUi): split the component and add tests

maxpatiiuk:
### TLDR:
- Fixes https://github.com/specify/specify7/issues/4225
- Refactors PartialDateUi to break it into small utility functions, hooks and sub-components - much more readable now
- Adds tests to PartialDateUi and each of it's sub-parts
- Improves sizing of PartialDateUi:
  - if there is enough space, precision select takes as much space as it needs, and input takes the rest of available space
  - if there is not enough space, precision select shrinks at twice the rate of the input field as it's less important

### Lesson learned:
The main cause of bugs in the previous implementation of PartialDateUI was the gigantic useEffect.
The useEffect then would set the validation message, formatted input value, database date value, and precision - too many responsibilities.
To follow "react good practices", every variable that useEffect used was mentioned in it's dependency array.
This means if any dependency changed (often "resource" when going though a sub view), a lot of things would happen in the useEffect - leading to unexpected bugs.
SImilarly, a lot of things are happening in that useEffect on page load - often leading to save button triggering unsaved changes warning.

[React docs on this](https://react.dev/learn/synchronizing-with-effects#what-are-effects-and-how-are-they-different-from-events)

### The main thing this refactor did:
- Convert as many useEffects to callbacks as possible, and then only call those callbacks when needed (i.e when user explicitly changed the date)
- Extract the remaining useEffects into hooks, and add tests for those hooks
- Split the remaining code into separate, smaller components


specifyweb/specify/api.py
```diff
@@ -345,8 +345,24 @@ def cleanData(model, data: Dict[str, Any], agent) -> Dict[str, Any]:
             logger.warn('field "%s" does not exist in %s', field_name, model)
         else:
             cleaned[field_name] = data[field_name]
+
```
maxpatiiuk, line 348: @melton-jason 

I shifted the responsibility for standardizing the date precision to the back-end - that is where it belongs in the first place, and it is simpler (on the front-end, if resource has precision but no date, can't unset precision right away because maybe user was going to enter the date later...)

question is - is this the best place to have this code?
i.e I could see one case where it would fail (not sure if we care about that case) - if the body of the API request includes the date field, but not the precision field, than the `if field_name.endswith('precision'):` would never run

specifyweb/frontend/js_src/lib/components/DataModel/specifyTable.ts
```diff
@@ -324,9 +324,10 @@ export class SpecifyTable<SCHEMA extends AnySchema = AnySchema> {
     if (typeof unparsedName !== 'string') throw new Error('Invalid field name');
 
     const splitName = unparsedName.toLowerCase().trim().split('.');
-    let fields = filterArray([
-      this.fields.find((field) => field.name.toLowerCase() === splitName[0]),
-    ]);
+    const exactMatch = this.fields.find(
```
maxpatiiuk, line 327: micro-optimization to avoid filterArray here
(I run a profiler and it showed a small slow down and garbage collector on this line, so decided to optimize it as it's a hot path)

specifyweb/frontend/js_src/lib/components/FormPlugins/PartialDateUi.tsx
maxpatiiuk, line 1: look at how much smaller this file is!!!

The PartialDateUi component is now just 53 lines! (and with tests!)

specifyweb/frontend/js_src/lib/components/FormPlugins/__tests__/useMoment.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+  test('sets default value on resource if missing', async () => {
+    const resource = new tables.CollectionObject.Resource(undefined, {
+      noBusinessRules: true,
+    });
+    const date = '2020-01-01';
+    const defaultValue = new Date(date);
+
+    renderHook(() => useMoment(resource, dateFieldName, defaultValue));
+
+    await waitFor(() => {
```
maxpatiiuk, line 59: await waitFor(()=>...); is super useful!
not sure about the performance impact of it though...

specifyweb/frontend/js_src/lib/utils/parser/definitions.ts
```diff
@@ -298,8 +298,11 @@ export function mergeParsers(base: Parser, extra: Parser): Parser {
         'required',
         base?.required === true || extra?.required === true ? true : undefined,
       ],
-      ...concat
-        .map((key) => [key, [...(base[key] ?? []), ...(extra[key] ?? [])]])
+      ...uniqueConcat
```
maxpatiiuk, line 301: fix bug when `base` and `extra` are the same parser, resulting in formatters being concatenated, thus applied twice, thus crashing

maxpatiiuk (commented):
@specify/ux-testing:
Some suggestions on what areas to focus on when testing:
- test with a date field that has precision (i.e Collection Object catalogedDate with catalogedDatePrecision)
  - make sure can save resource with month-year or year precision, and reopen the resource and precision is remembered correctly
  - test adding the same field to the form twice, and making sure the changes are synchronized between them (should get synchronized once you are finished editing and the date picker lost focus)
- test with timestampCreated and timestampModified fields
- test with disabling the "Use accessible date picker" options in the preferences
  - try putting invalid date and see how that is handled (should see validation error)
- test the date picker when inside a -to-many sub-view and you switch between resources, the date picker correctly shows the date and date precision for each resource. similarly when date picker is inside a record set
- try switching precision for a full date, and see how date is updated

CarolineDenis (approved):

specifyweb/specify/api.py
melton-jason: I would agree that standardization of data should probably occur on the backend (the backend needs a little more love anyways). 

This is a fine spot for the code, I think. 
The only other place I might consider is https://github.com/specify/specify7/blob/22bda472777c22c28a660890cd84f7991169f2e4/specifyweb/specify/api.py#L408-L420
This already handles converting the SQL timestamp objects to strings (by way of https://github.com/specify/specify7/blob/22bda472777c22c28a660890cd84f7991169f2e4/specifyweb/specify/api.py#L670-L674)

However, `cleanData` also works, and is probably more intuitive at the moment. 
Would it be worth adding an assert or at the very least a console.warn to handle the case? It would (hopefully) only impact those directly using the API.

emenslin (approved):
I tested this in the ways suggested and did not run into any issues, looks good.

specifyweb/frontend/js_src/lib/components/FormPlugins/useMoment.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  return [
+    moment,
+    React.useCallback(
+      (moment) => {
+        if (moment?.isValid() !== false)
+          resource?.bulkSet({
+            [dateFieldName]: moment?.format(databaseDateFormat) ?? null,
+          });
+        setMoment(moment);
```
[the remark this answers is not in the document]
melton-jason: On a side note, is there any harm in changing `if (moment?.isValid() !== false)` to  `if (moment?.isValid() === true)`? 
It would be slightly easier to reason about and doesn't change the functionality of the following `setMoment(moment)` which is always being called.

melton-jason (approved):

grantfitzsimmons (approved):
❤️ 

I tested both precise and non-precise dates to ensure everything is functioning correctly. The precision is being properly set, and data entry through copy & paste and typing is working well. Date formats from other regions also appear to be functioning correctly.

specifyweb/frontend/js_src/lib/components/FormPlugins/useMoment.tsx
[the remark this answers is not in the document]
maxpatiiuk: > Why was bulkSet used here over set?

great question, and something I should have made clearer in the code

in present form `.set()` is trying to be type safe with field names and argument names
when I was adding that, I was considering the cases when you are dealing with a specific known table, and a hardcoded field name - in those cases this type safety is very useful
but I failed to consider the generic cases like this - the PartialDateUi could be used in any table, and field name is dynamic too, so typing fails

`.bulkSet()` does not have any of this type-safety logic, so that's why I used it here (to avoid a typescript error)
in other places in the code we used an `as never` assertion, which is probably worse than using not-type-checked `.bulkSet()`

we could add a `.genericSet()` that would be exactly like `.set()` but untyped, but since it would be equivalent to `.bulkSet()`, the only benefit is code clarity
but since we are looking to refactor the backbone code, I didn't want to do such changes now in this PR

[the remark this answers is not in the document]
maxpatiiuk: > On a side note, is there any harm in changing if (moment?.isValid() !== false) to if (moment?.isValid() === true)?

Not equivalent.
The current condition is true if moment is valid OR moment is undefined (in which case the field value is set to null)
Your condition is only true if moment is defined AND valid

being explicit like `if(moment === undefined || moment.isValid())` is probably a better idea for readability - thanks for the suggestion, will refactor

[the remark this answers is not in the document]
melton-jason: Ah yes, you are correct my apologies. 
At the moment I was assuming that if moment was undefined, then it is a falsy value and thus the conditions would be equivalent (because `(moment?.isValid() !== false)` would return `false` in that case). 
Somehow slipped my mind we were doing an explicit check (!==) rather than implicitly converting the type (!=) in the comparison :facepalm:

PR #4282, opened 2023-12-11
Define proper user type for global resource

### LINKED ISSUE #3956: Creating a new resource at the 'Global' level sets the user group as 'prefs'

CarolineDenis:
Fixes #3956

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/tree.ts
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/tree.ts
@@ -70,12 +70,11 @@ function getGlobalAllResources(resources: AppResources): {
 
   const directoriesNeeded = ['Global Prefs', 'Prefs', 'Common'];
   if (globalDirectories.length < directoriesNeeded.length) {
-    const directoriesPresent: readonly (string | null)[] = [];
-    globalDirectories.forEach((directory) =>
-      directoriesPresent.push(directory.userType)
+    const directoriesPresent = new Set(
+      globalDirectories.map((directory) => directory.userType)
     );
     const missingUserType = directoriesNeeded.find(
-      (userType) => !directoriesPresent.includes(userType)
+      (userType) => !directoriesPresent.has(userType)
     );
 
     if (missingUserType !== undefined) {
--- a/specifyweb/frontend/js_src/lib/components/AppResources/tree.ts
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/tree.ts
@@ -67,13 +67,26 @@ function getGlobalAllResources(resources: AppResources): {
   const globalDirectories = resources.directories.filter(
     (directory) => directory.scope === 'global'
   );
-  if (globalDirectories.length === 0)
-    globalDirectories.push({
-      ...addMissingFields('SpAppResourceDir', {
-        userType: 'Common',
-      }),
-      scope: 'global',
-    });
+
+  const directoriesNeeded = ['Global Prefs', 'Prefs', 'Common'];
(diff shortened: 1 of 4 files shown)


CarolineDenis:
NOTES: 
- don't forget to uncomment spAppResourceDir in webOnlyViews 
- add a way to only choose from Preferences or Global Preferences when "Other Properties Resource" is chosen?

CarolineDenis:
@maxpatiiuk , 
Can we create two new keys one for Preferences and one for Global Preferences that would have a name? We would then use those to create them instead of otherPropertiesResource. 
https://github.com/specify/specify7/blob/84e56043528e3fa4c765d31da0f186882de86116/specifyweb/frontend/js_src/lib/components/AppResources/types.tsx#L199

specifyweb/frontend/js_src/lib/components/AppResources/tree.ts
```diff
@@ -67,13 +67,25 @@ function getGlobalAllResources(resources: AppResources): {
...
-  if (globalDirectories.length === 0)
-    globalDirectories.push({
-      ...addMissingFields('SpAppResourceDir', {
-        userType: 'Common',
-      }),
-      scope: 'global',
-    });
+
+  const directoriesNeeded = ['Global Prefs', 'Prefs', 'Common'];
+  if (globalDirectories.length < directoriesNeeded.length) {
```
maxpatiiuk, line 72: this condition should be removed - this assumes that there are only 3 possible user types at a global level - not an assumption I am willing to make given that these directories are created by sp6 - who knows what it does

specifyweb/frontend/js_src/lib/components/AppResources/tree.ts
```diff
@@ -67,13 +67,25 @@ function getGlobalAllResources(resources: AppResources): {
...
+    const directoriesPresent = new Set(
+      globalDirectories.map((directory) => directory.userType)
+    );
+    const missingUserType = directoriesNeeded.find(
+      (userType) => !directoriesPresent.has(userType)
+    );
```
maxpatiiuk, line 78: this code assumes that at most 1 user type is missing - can you rewrite to add any user type that is missing, not just one?

specifyweb/frontend/js_src/lib/components/AppResources/types.tsx
```diff
@@ -203,6 +203,20 @@ export const appResourceSubTypes = ensure<IR<AppResourceSubType>>()({
...
+    icon: icons.viewList,
+    label: preferencesText.preferences(),
+  },
+  globalPreferences: {
+    mimeType: 'text/x-java-properties',
+    name: 'Global Preferences',
+    documentationUrl: undefined,
+    icon: icons.viewList,
+    label: resourcesText.globalPreferences(),
+  },
```
maxpatiiuk, line 219: we are not encouraging users to create these app resources (they should be created by sp6 in every database). In fact, we should deprecate these app resources as soon as Institutional Preferences are added), thus we should not be exposing explicit options for these legacy resources in the UI.

maxpatiiuk (changes requested):

maxpatiiuk:
see also https://github.com/specify/specify7/issues/3956#issuecomment-1853251752 for proposed solutions

PR #4283, opened 2023-12-13
fix(Preferences): correctly render dynamic titles

maxpatiiuk:
Bug caused by incorrect merge conflict resolution in Preferences/index.tsx (thus function calls disappeared), followed by incorrect TypeScript error resolution (thus type assertion was added)

### DIFF
--- a/specifyweb/frontend/js_src/eslint.config.js
+++ b/specifyweb/frontend/js_src/eslint.config.js
@@ -52,22 +52,6 @@ module.exports = [
           },
         },
       ],
-      'jest/require-hook': [
-        'warn',
-        {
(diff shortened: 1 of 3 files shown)


maxpatiiuk:
FYI: dynamic titles were added to reduce need for this ugliness:

https://github.com/specify/specify7/blob/xml-editor/specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx#L1936-L1945

if not for cyclical import, that entire block at the end of UserDefinitions.tsx can be refactored into dynamic titles

CarolineDenis (approved):

PR #4290, opened 2023-12-15
fix(FormEditor): form cache not being cleared

### LINKED ISSUE #1977: Bust cached app resource on changes
If a user edited an app resource, need to tell the browser to bust the cached app resource.

If this is done on the front-end, it can be done for the current user only. If it is done on the back-end (though `ETag` or some other solution), this would clear the cache for all users

---

From @grantfitzsimmons:

> When updating the schema configuration, it automatically clears your cache. 
>
> The same behavior should happen when updating a views file! Bonus points if it can also preserve the current URL.

maxpatiiuk:
Because of stale state, the cache for the form that was edited most recently wasn't getting cleared.

Changed to ref, since using state here didn't make sense in the first place (state isn't used in render). I used state because at the time I followed idea that we should use state whenever possible, and ref only when must, because state would be cleaner and less bug prone. However, today I belive that each has it's place - use state when you need to use the value in the render or dependency array - use ref otherwise

Fixes #1977


CarolineDenis (approved):

pashiav:
Instructions for testers? @maxpatiiuk So far the cache doesn't seem to be clearing on Chrome.

maxpatiiuk:
1. Open a form definition in app resources and make an edit
2. Without reloading the page, open that form and see that changes are visible
3. Reload the page and see that changes are visible (this is different code pathway than 2)
4. If 2 or 3 does not work, open incognito browser tab, and make sure changes are visible (if not, then there is a mistake in step 1 - perhaps you edited a form that is overwritten at a different level)

emenslin (approved):
I just tested this and I was able to see changes without having to reload the page so it seems like it is working.

Issue #4294, opened 2023-12-15
Error switching base tables in queries 


emenslin:
**Describe the bug**
When switching between base tables when querying without saving an error message pops up.
[Specify 7 Crash Report - 2023-12-15T14_32_58.308Z (1).txt](https://github.com/specify/specify7/files/13686449/Specify.7.Crash.Report.-.2023-12-15T14_32_58.308Z.1.txt)

**To Reproduce**
1. Create a new query using any tree table
2. Add anything to the query 
3. Don't save and add a new query under a different table 

**Expected behavior**
I would expect there to be no error and you should be able to switch between queries with no problems.

**Desktop:**
 - OS: Windows 11
 - Browser: Chrome
 - Specify 7 Version: edge

**Database Name:** lsumzmammals (6.8.03)

melton-jason:
> **Describe the bug** If you create a query based on any of the Trees:
> 
> 1. Taxon
> 2. Storage
> 3. Geography
> 4. Lithostrat
> 5. Chronostrat
> 
> Then go and create another new query with any other table as a base. It's going to through an OperationalError
> 
> **Expected behavior** No error [Specify 7 Crash Report - 2023-11-24T22_08_09.151Z.txt](https://github.com/specify/specify7/files/13462329/Specify.7.Crash.Report.-.2023-11-24T22_08_09.151Z.txt)
> 
> **Beautiful Screen Recording**
> 
>  Screen.Recording.2023-11-24.at.4.07.31.PM.mp4 
> **Desktop:**
> 
> * OS: Mac OsX
> * Browser: Chrome
> * Specify 7 Version: Any

From @carlosmbe in https://github.com/specify/specify7/issues/4231#issue-2010253721

Issue #4231, opened 2023-11-24
Query Builder - Tree Table to Anything Else OperationalError


carlosmbe:
**Describe the bug**
If you create a query based on any of the Trees:

1. Taxon
2. Storage
3. Geography
4. Lithostrat
5. Chronostrat

Then go and create another new query with any other table as a base. It's going to through an OperationalError

**Expected behavior**
No error
[Specify 7 Crash Report - 2023-11-24T22_08_09.151Z.txt](https://github.com/specify/specify7/files/13462329/Specify.7.Crash.Report.-.2023-11-24T22_08_09.151Z.txt)

**Beautiful Screen Recording**

https://github.com/specify/specify7/assets/53784701/86e67a69-b87d-4fed-95a5-4c6f8a37acf3

**Desktop:**
 - OS: Mac OsX
 - Browser: Chrome
 - Specify 7 Version: Any

melton-jason:
Closing this as a duplicate of #4294 as that issue is more more general and covers all tables, not just tree tables

PR #4298, opened 2023-12-16
Add proper key props to query builder wrappers

realVinayak:
To test
1, Make sure going between queries in query builder doesn't cause a crash
2. To test for https://github.com/specify/specify7/issues/4294, 

> Create a new query using any tree table
> Add anything to the query
> Don't save and add a new query under a different table

Make sure the crash doesn't happen.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/index.tsx
@@ -23,7 +23,7 @@ import { QueryBuilder } from './Wrapped';
 function useQueryRecordSet(): SpecifyResource<RecordSet> | false | undefined {
   const [recordsetid = ''] = useSearchParameter('recordsetid');
   const [recordSet] = useAsyncState<SpecifyResource<RecordSet> | false>(
-    React.useCallback(() => {
+    React.useCallback(async () => {
       if (!hasToolPermission('recordSets', 'read')) return false;
       const recordSetId = f.parseInt(recordsetid);
       if (recordSetId === undefined) return false;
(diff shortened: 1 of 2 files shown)


realVinayak:
@emenslin Can you test this branch and check if https://github.com/specify/specify7/issues/4294 still occurs?

melton-jason:
If further testing instructions are desired/required: see https://github.com/specify/specify7/issues/4231#issue-2010253721

CarolineDenis (approved):

emenslin (approved):
Tested querying with all of the trees and there is no longer an error, looks good now!

melton-jason (approved):

carlosmbe (approved):

specifyweb/frontend/js_src/lib/components/QueryBuilder/index.tsx
```diff
@@ -84,7 +84,7 @@ function QueryById({
   const recordSet = useQueryRecordSet();
 
   return query === undefined || recordSet === undefined ? null : (
-    <QueryBuilderWrapper query={query} recordSet={recordSet} />
+    <QueryBuilderWrapper key={queryId} query={query} recordSet={recordSet} />
```
maxpatiiuk, line 87: solution looks good and thanks for fixing this.

ideally, our component should be able to handle any prop changes without crashing - if you have time, and it doesn't look to completed, it would be better to modify the component to not crash, rather than nuke the component on pro pchange

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/QueryBuilder/index.tsx
realVinayak: It was @melton-jason's idea for the key prop!

>  "don't change query on the fly" is like an implicit contract QueryBuilderWrapper puts out - this implicit contract is not guaranteed by TypeScript or tests

Can be done via ESLint? Exhaustive deps?

maxpatiiuk: add a required prop in the QueryBuilderWrapper component called "yesIAddedTheKeyProp"? 🤣

realVinayak: crash during runtime if not. can't let anyone get away

Issue #3141, opened 2023-03-09
Label showing author as `null`


grantfitzsimmons:
## **Problem:**

The author is not displaying (displaying as `null`) on CSIRO's labels in Specify 7 whereas it does display correctly in Specify 6 and in query results. It looks like there is a bug in the way 7 handles reports.

#### The problem:

Works in 6, not in 7

![image](https://user-images.githubusercontent.com/37256050/224113114-4c13212e-1b59-4269-9892-714a834722d4.png)

<img width="1008" alt="image" src="https://user-images.githubusercontent.com/37256050/224111304-cfcc6297-8daa-49fa-bc78-ffe9965ec65b.png">

<img width="603" alt="image" src="https://user-images.githubusercontent.com/37256050/224112330-281ef557-495a-419c-8df1-6eaa86a179e3.png">

## **Links:**

The report and query are available on the test panel's `coldfish` database.

https://coldfish-edge.test.specifysystems.org/specify/resources/app-resource/173/ <- Here is the JRXML definition

Here's the specific snippet:

```xml
<textFieldExpression   class="java.lang.String"><![CDATA[$F{1,9-determinations,4.taxon.author}]]></textFieldExpression>
```

Expression in iReport:
<img width="350" alt="image" src="https://user-images.githubusercontent.com/37256050/224111959-46c8b7b1-c90b-4ef5-b159-18ed720ccd2e.png">

It is displayed correctly in the query results:

<img width="2211" alt="image" src="https://user-images.githubusercontent.com/37256050/224111873-3d8f6c7d-412a-4e2f-ace0-d21b8631e0e7.png">

**Associated Query:**
https://coldfish-edge.test.specifysystems.org/specify/query/195/
[Report_Adult Label TYPE.json.zip](https://github.com/specify/specify7/files/10934565/Report_Adult.Label.TYPE.json.zip)

**SCC Vault:** [Specify 6 Report Export](https://drive.google.com/open?id=1DVAtd-6aO1x9filFKxiOPeub8fXhL5qA&authuser=grant%40specifysoftware.org&usp=drive_fs)

---

**Report Runner Output:**

[Sp7_Output.pdf](https://github.com/specify/specify7/files/10934532/Sp7_Output.pdf)
[Sp6_Output.pdf](https://github.com/specify/specify7/files/10934531/Sp6_Output.pdf)

---

**Reported By:** Corinna P at CSIRO ([Asana](https://app.asana.com/0/1201972579963460/1204133778858353/f))

[...]

PR #4299, opened 2023-12-16
Allow consistent behavior for Reports/Labels and improve Report preferences

### LINKED ISSUE #3570: Inconsistent Label Generation for same Collection Object
> Some more strange label report behavior ... .
> 
> The label report should be excluding labels that don’t have the current determination, and should exclude labels for preparations with sizes other than the selected ones (we have Large, Small, and Mini labels for different size jars).
> 
> We create tissue labels in Excel and there is no tissue label report in Specify (that value is just a placeholder in the table) so  it is odd that it is generating “small” labels for those.
- FWRI

**Describe the bug**
There seems to be an inconsistent generation of labels among all of the different ways one _can_ generate a label.  
More specifically, among the 'Generate Label' button on the form ('Generate Label on Save' checkbox produces the same behavior), and manually identifying the Collection Object (in this case via catalog number) in the report in the Reports Dialog. 

I have prepared a simple [test Collection Object record ](https://fwri2023-v7811.test.specifysystems.org/specify/view/collectionobject/453150/) to demonstrate this issue. 

I will be using the [`Small Labels` JRXML App Resource](https://fwri2023-v7811.test.specifysystems.org/specify/resources/app-resource/112/) as a Label throughout this example. 

Note that the Report Query is only supposed to fetch records that have a `Small` Label Size for their Preparations

fwri_report_query_line

Here are the preparations on the test Collection Object record. 
fwri_test_preps

When the Collection Object is saved (when the 'Generate Label on Save' checkbox is checked), or when the Generate Label button is pressed on the form and the `Small Labels` Label selected, the following Label is generated:

fwri_test_label_button

As shown in the Report Query, this Label should NOT have any information for the Tissue preparation. 
In fact, Specify will generate a Label for _ALL_ Preparations on the Collection Object, regardless of its data.

If the Catalog Number (000165434) is manually entered into the `Small Labels` Report Query, then the correct and expected Label is generated:
fwri_test_manual_report

**Possible Explanation**
It appears that somehow the Query might be `undefined` when the Label is run. I think that instead of a Query, the Label is using all of the available data on the Collection Object as a substitute for a Query. 

This would make sense, given the Label Definition. Here is a relevant section.
```xml
<textFieldExpression   class="java.lang.String"><![CDATA[$F{1,63-preparations.preparation.countAmt}==null?"":"Count: "+$F{1,63-preparations.preparation.countAmt}+" Pres. "+$F{1,63-preparations.preparation.text2}]]></textFieldExpression>
```
Nowhere in the logic of the Label does it check for the Label Size of the Preparation (`preparation.remarks` in the database), as this is supposed to be filtered out when the Query is executed. 

For future and easy reference for the Developer tasked to this issue, the `Generate Label` button is rendered via the following function. 
https://github.com/specify/specify7/blob/de940f090b68553f063b40fe15be22077b5099b9/specifyweb/frontend/js_src/lib/components/FormCommands/index.tsx#L20-L52

**Desktop:**
Replicated on:
 - OS: Nix
 - Browser: Chrome
 - Specify 7 Version: 7.8.11

**Reported By**
Florida Fish and Wildlife Research Institute (FWRI)

melton-jason:
Fixes #3570 

The underlying cause of this issue is a preference which cleared all query filters when running a report or label from a Form. This behavior was first implemented in https://github.com/specify/specify7/commit/199c11697ed0fb8cafee436888a4f9f259537cbf and later made into a preference in https://github.com/specify/specify7/commit/e1faeb2d0f200ec3084e5ae16bfa1c0b7c8efaf4. 

This preference has a **_significant_** impact to the behavior of reports. For example, if the query (with a CollectionObject base table) for the report had the fields `determinations -> isCurrent -> Equals -> True` and `preparations -> preptype -> name -> Equals -> SomePrepType`, these would be passed to the report runner as `determinations -> isCurrent -> Equals -> Any` and `preparations -> preptype -> name -> Equals -> Any` (respectively). 

@specify/ux-testing, what is the benefit for this behavior? Why was it implemented in such a way in the first place? I severely doubt this is wanted by users and is definitely not expected.

Thankfully for those currently struggling with #3570, the remedy of disabling the preference should suffice for most cases without having to update.

In this PR, I have set the default value of the preference to False, provided a better description for the preference, and defined a new preference which would display the query filters and not auto-run the Report or Label from the Form. 

### Testing Instructions 
- Identify a label/report which has an associated Query with any filters (such as the [Small Labels](https://fwri6123-issue-3570.test.specifysystems.org/specify/resources/app-resource/112/) which has a Query field which only returns COs which have "Small"  in the remarks field-localized as Label size- on preparation)
- With the `Reset query filters` preference disabled, locate a record which would normally not return any results with the associated Query (If using the Small Labels report from the previous instruction, an example of such a record is [CO with ID 136663](https://fwri6123-issue-3570.test.specifysystems.org/specify/view/collectionobject/136663/)) 
- Ensure that no results are returned. 
- Modify the record so it would normally be returned by the Query (in the previous example, this would be adding a preparation which has a Small Label size, or modifying already existing preparations so they have a Small Label size)
- Generate the report/label with the modified record and confirm a correct/expected output. 

- - - 

- Ensure the new preference `Show query filters when running a Report from a Form` behaves correctly (once the report/label is selected, the Query Parameters dialog should be displayed and the query should be able to be temporarily modified before being passed to the report runner. Note the query should have an additional field at the bottom of the Query with an ID field; this is expected).

- An inconsistency in reports/labels was also demonstrated in https://github.com/specify/specify7/issues/3141#issuecomment-1569495107. Can this be tested as well?


CarolineDenis (approved):

maxpatiiuk:
> what is the benefit for this behavior? Why was it implemented in such a way in the first place? I severely doubt this is wanted by users and is definitely not expected.

sp6 has ability to say on a per-field basis whether when you run a report based on a query, the filter condition for that field is preserved or emptied.

i.e for a field like status, you would want to preserve, but for a field like data, you would want to wipe the value so that you can enter it anew when generating the report

(not having to erase the value manually is saving keystrokes when running the report often - but there may be other use cases for this)

if I remember correctly, this is stored in an "always" field, and is represented as a checkbox in the query builder at the end of each field - it takes up space all the time, but is not used often, so I definitely didn't want to implement it like that in sp7.
some sort of better UI was needed. but, I wanted to get at least something done quickly, so I added a use pref that allows globally setting or unseting all filters when running a query in a report

maxpatiiuk:
Auto-running the query without prompting user for parameters was designed as a shortcut for when the user needs to run many reports.

the way it works (or supposed to):
a query would have a list of fields it wants to output, and some filtes
those filters would either be reset (as determined by this preference) or the filters would match the current record anyway
the query is run, but scoped to return only a single record - the record you have open in a form
thus, there is no need for prompting user to enter any parameters - saves keystrokes (which is important as this action is performed many times per day by some users)

but Grant would be able to tell much more about what use cases and workflows we are trying to cover with these tools

pashiav (approved):
Looks good! Labels are only generated for the corresponding label query results selected in Reports. 
issue-3570-co
issue-3570-label


Selecting a label query where there are no results for the CO displays _"The report query returned no results."_ to the user and no label is generated.
Screen Shot 2023-12-21 at 2 52 56 PM

CarolineDenis:
Confirm that https://github.com/specify/specify7/issues/3141 is fixed by this PR.

grantfitzsimmons (approved):
Works as expected. Great job, @melton-jason!
I saw Caroline test this as well and it looks great.

emenslin:
> @emenslin Can you please test to see if this PR resolves an issue reported by RBGE?

@grantfitzsimmons Yes this issue was resolved. All reports now come out with the the same result with one page for the sample number. 
[RBGE report test.pdf](https://github.com/specify/specify7/files/14103735/RBGE.report.test.pdf)

grantfitzsimmons:
@melton-jason Would you like to merge?

PR #4300, opened 2023-12-17
refactor(Workbench): delete wbView.js

maxpatiiuk:
wbView.js has been converted to TypeScript and split into many smaller files. wbView.js was deleted

However, during the resolution of merge conflicts, this file has been accidentally "resurrected".

This PR deletes that file again, as well as makes a few tiny refactors (done in the process of verifying if wbView.js has any changes that aren't yet applied in the workbench typescript files)


realVinayak (approved):
Looks good! However, can we merge this towards the end of release of xml-editor? There were some bugs introduced in refactoring, and it was really helpful to refer at the js code to see what the previous code was (using just the IDE rather than switching branches everytime, I guess you could pull and add it to .gitignore). See: https://github.com/specify/specify7/pull/4205.

CarolineDenis (approved):

maxpatiiuk:
@realVinayak whenever I need that, I just open the file on github - https://github.com/specify/specify7/blob/41f92effcb5d7c2a109b521a16c030f8f062ac45/specifyweb/frontend/js_src/lib/components/WorkBench/wbView.js

Also, when one branch is far ahead of production (like xml-editor), I clone specify7 twice (and have npm install and docker compose setup in both directories) - one is using production and another xml-editor
this allows me to open them simultaneously, or switch between adding new features on branch and fixing bugs on production without having to git stash my ongoing changes

does that work for you?

realVinayak:
It does, thanks!

Issue #2617, opened 2022-12-13
Test all pages with 400% zoom


maxxxxxdlp:
Report any overflow issues

One obvious problem is that menu takes up half the screen - thus this needs to be tested after https://github.com/specify/specify7/issues/1462 is implemented

PR #4310, opened 2023-12-20
Confine the app resource editor to the width of screen when line wrap…

### LINKED ISSUE #4212: When 'Line wrap' is disabled, the app resource editor doesn't stay confined to the view width
image

https://github.com/specify/specify7/assets/37256050/f9fedd4b-9355-463d-ae8f-876690eff7de

`v7.9.1`
[Specify 7 System Information - 2023-11-15T19_58_38.734Z.txt](https://github.com/specify/specify7/files/13369101/Specify.7.System.Information.-.2023-11-15T19_58_38.734Z.txt)

CarolineDenis:
… disabled

Fixes #4212

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
@@ -222,7 +222,7 @@ export function AppResourceEditor({
   return typeof resourceData === 'object' ? (
     <Container.Base
       className={`flex-1 overflow-auto
-        ${lineWrap === true ? 'sm:overflow-visible' : ''}
+        ${lineWrap ? 'sm:overflow-visible' : ''}
       `}
     >
       <DataEntry.Header className="flex-wrap">
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
@@ -28,7 +28,6 @@ import { DeleteButton } from '../Forms/DeleteButton';
 import { SaveButton } from '../Forms/Save';
 import { AppTitle } from '../Molecules/AppTitle';
 import { hasToolPermission } from '../Permissions/helpers';
-import { userPreferences } from '../Preferences/userPreferences';
 import { isAppResourceSubType } from './Create';
 import {
   AppResourceDownload,
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
@@ -63,12 +63,6 @@ export function AppResourceEditor({
     directory: ScopedAppResourceDir
   ) => void;
 }): JSX.Element | null {
-  const [lineWrap] = userPreferences.use(
-    'appResources',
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
```diff
@@ -213,7 +220,11 @@ export function AppResourceEditor({
   );
 
   return typeof resourceData === 'object' ? (
-    <Container.Base className="flex-1 overflow-auto sm:overflow-visible">
```
grantfitzsimmons, line 216: What purpose did this serve previously? I'm trying to comprehend why we would retain it rather than checking the user preferences for determining the appropriate class to use. 

@maxpatiiuk ?

grantfitzsimmons: When `sm:overflow-visible` is not applied issue #4212 does not occur

maxpatiiuk: Appears to have been added by @CarolineDenis in https://github.com/specify/specify7/commit/1078bccd1 as part of improving app resource layout on small screens (probably as part of https://github.com/specify/specify7/issues/2617)

specifyweb/frontend/js_src/lib/components/AppResources/Editor.tsx
```diff
@@ -213,7 +220,11 @@ export function AppResourceEditor({
...
+    <Container.Base
+      className={`flex-1 overflow-auto
+        ${lineWrap ? 'sm:overflow-visible' : ''}
```
maxpatiiuk, line 225: I don't think this has to be conditional based on user preference
Whenever possible, try to have the code path be as common as possible for all users regardless of preferences - that way there are fewer things that could break and fewer things that need testing

So here, try to come up with styles that work both when line wrap is disabled, and when screen is narrow

maxpatiiuk: really, the overflow style should be set on the text editor container directly, not the page container (i.e if editor has a long line, and horizontal scrollbar appears, the header should not be scrolling off - header should stay in place, and only the editor should scroll)

you may still have overflow-auto on the page container as a fallback, but setting it on the text editor container is probably a better idea

maxpatiiuk (changes requested):

CarolineDenis:
When removing sm:overflow-visible 

Line wrap: 
Screenshot 2023-12-21 at 9 01 47 AM
Screenshot 2023-12-21 at 9 01 30 AM
Screenshot 2023-12-21 at 9 01 09 AM

No wrap: 
Screenshot 2023-12-21 at 9 14 03 AM
Screenshot 2023-12-21 at 9 14 19 AM
Screenshot 2023-12-21 at 9 14 38 AM

grantfitzsimmons (approved):
Looks good to me. Not being conditional makes more sense as far as I can tell.

emenslin (approved):
Does not cut off anymore, looks good. 
Screenshot 2023-12-21 143603

maxpatiiuk (approved):

specifysoftware:

PR #4312, opened 2023-12-20
Display tables or advanced tables in schema config

### LINKED ISSUE #4307: `xml-editor`: Schema config table list appears incorrectly
**Describe the bug**
The list of tables in the Schema Config tool do not display, and when you view them all, the scroll bar is immediately to the right of the list, not the along the side of the window.

**To Reproduce**
1. Go to Schema Config in the User Tools menu
2. See issue

https://github.com/specify/specify7/assets/37256050/805de557-ec0d-4a36-aea0-8de929c3aadd

https://kufishdemo-xml-editor.test.specifysystems.org/specify/schema-config/en/

Arc on macOS

CarolineDenis:
Fixes #4307


specifyweb/frontend/js_src/lib/localization/wbPlan.ts
```diff
@@ -418,6 +418,9 @@ export const wbPlanText = createDictionary({
     'uk-ua': 'Показати додаткові таблиці',
     'de-ch': 'Erweiterte Tabellen anzeigen',
   },
+  showAdvancedTables: {
```
maxpatiiuk, line 421: we already have `wbText.showAllTables()`. is this slightly different string added intentionally?
can we change them both to use the same label?

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
```diff
@@ -75,18 +75,33 @@ export function TableList({
...
+        : allTables;
+
+    return filteredTables.sort(sortFunction(({ name }) => name));
+  }, [filter, showHiddenTables]);
+
   const sortedTables = React.useMemo(
     () =>
       Object.values(genericTables)
-        .filter((table) => (filter ? filter(table) : !table.isSystem))
+        .filter((table) => (filter ? filter(false, table) : !table.isSystem))
```
maxpatiiuk, line 104: hmm, looks like the edits I did in this PR are lost - https://github.com/specify/specify7/pull/4242/files#diff-c703adeb4352c7832286317011ba686112a5f56f7c0ca20f3a493fc2561197d4R86-R97

do you know why?

that change was important for improving typescript type checking performance by several seconds

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
```diff
@@ -75,18 +75,33 @@ export function TableList({
 }: {
   readonly cacheKey: CacheKey;
   readonly getAction: (table: SpecifyTable) => string | (() => void);
-  readonly filter?: (table: SpecifyTable) => boolean;
+  readonly filter?: (showHiddenTables: boolean, table: SpecifyTable) => boolean;
```
maxpatiiuk, line 78: just a friendly reminder:
the TableList component is used in multiple places and on changes you have to consider impact in app places

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
```diff
@@ -75,18 +75,33 @@ export function TableList({
...
+      (showHiddenTables ? undefined : ({ isSystem }): boolean => !isSystem);
+
+    const filteredTables =
+      typeof filterFunction === 'function'
+        ? allTables.filter(filterFunction)
+        : allTables;
+
+    return filteredTables.sort(sortFunction(({ name }) => name));
+  }, [filter, showHiddenTables]);
+
```
maxpatiiuk, line 100: why do all of this here?
couldn't this be much simpler - in `sortedTables`, instead of always passing false to `filter` function, just pass showHiddenTables

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
```diff
@@ -102,10 +117,13 @@ export function TableList({
     return showHiddenTables ? sortedTables : presentFilteredTables;
   }, [showHiddenTables, sortedTables, children]);
 
+  const tablesToMapOver =
+    cacheKey === 'appResources' ? tablesToDisplay : advancedTables;
```
maxpatiiuk, line 121: this is bad practice and should be avoided if possible
(because needless separate code paths should be avoided if possible, and because cacheKey, as it's name implies, it only used for local storage caching, not for display logic)
you shouldn't need this once the suggestion I added for sortedTables is implemented

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
CarolineDenis: not really sortedTables and advancedTables do not serve the same purpose. In one case we only want the tables that already have resources in the other we only want the tables that are not advanced

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
CarolineDenis: because your changes were applied to the previous code, before we had the feature to only tables with app resources

CarolineDenis: because they were applied to an old code modified with the new feature 'only show tables that have an app resource'

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
CarolineDenis: Would you have another suggestion than using cacheKey to differentiate the fact that we are in one component or the other?

maxpatiiuk: how does this not combine the sortedTables and advancedTables?
```
  const sortedTables = React.useMemo(
    () =>
      allTables
        .filter((table) =>
          filter
            ? filter(showHiddenTables, table)
            : showHiddenTables || !table.isSystem
        )
        .sort(sortFunction(({ name }) => name)),
    [filter, allTables, showHiddenTables]
  );
```

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
maxpatiiuk: sorry, not understanding this.
can you link to the code commit that removed them?
that PR was a fixing a performance issue - are you saying a user-facing feature remove a performance optimization?

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
CarolineDenis: For app resources list of table, when the check box is not checked, this is what we want to see: 
Screenshot 2023-12-22 at 10 57 53 AM
With the code above, this is what is displayed: 
Screenshot 2023-12-22 at 10 59 51 AM

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
CarolineDenis: the changes you made are on `const advancedTables = ...` which are kept in this PR. The `const sortedTables = ...` and `const tablesToDisplay = ...` are from a new code for the feature 'only display tables that have resources ([here](https://github.com/specify/specify7/pull/4108/files)) '.

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
CarolineDenis: we still need a check to display sortedTables for schema config and another to display tablesToDisplay for app resources.

maxpatiiuk (approved):
looks good, but this could be made a bit simpler
look at the changes I did in https://github.com/specify/specify7/pull/4312/commits/33b3c73fe0fabe23c06b632020dbcfc4aa8d8584

emenslin (changes requested):
It doesn't seem like the changes have applied as I do not see any difference.

https://github.com/specify/specify7/assets/150940840/956f1c80-2093-4c9f-a485-32039d7a678a

emenslin (changes requested):
So it does automatically open the tables however the scroll bar is still to the right of the list and not along the edge of the window.

https://github.com/specify/specify7/assets/150940840/bdff92ba-dedb-4cbc-9abd-ac9730ccd2c3

emenslin (approved):
Looks good now, all tables open and scroll bar is on the right side! 
Screenshot 2024-01-02 111304

maxpatiiuk:
@CarolineDenis careful, when you force-pushed to this branch, you overwrote the commit I made

for the future, you can use `git push --force-with-lease` instead of `git push --force` - it's a safer variant (that would make git refuse to overwrite if remote contains some commit that local git is not aware of)

CarolineDenis:
@maxpatiiuk, the force push was intended as the code in [25afbdd](https://github.com/specify/specify7/pull/4312/commits/25afbdd57be71a05d1724cc19b2be27360079a6e) doesn't work for the feature we need.

maxpatiiuk:
It appears that there was a bug in my code
I pushed a tiny fix. Could you check if it works now?

no need to delete commits. reminds me of https://github.com/munificent/vigil - a language that deletes your code if you have a bug

CarolineDenis:
@maxpatiiuk works great :+1: thank you!

pashiav (approved):

PR #4320, opened 2023-12-22
fix(AppResource): handle cases when linter throws exceptions

### LINKED ISSUE #4209: `xml-editor`: XML-editor allows saving even if there are errors in the XML definition
Steps to replicate:
1. Open a data object formatter definition in xml-editor. 
2. Modify `<field>SOME_FIELD</field>` of any formatter definition to be incorrect. For example, set it to  `<field>SOME_FIELD.</field>` (notice the period after "SOME_FIELD" . 
3. Notice that `Save` button is not disabled. In fact, clicking on Save actually saves the incorrect definition.
4. Next time opening that formatter definition leads to specify crashing, so user cannot fix the issue in any case.

While the actual issue is that save button is not disabled if there is an error, I think that it is a bit too restrictive to entirely crashing if there is a faulty xml-definition (for example, the ones edited in specify 6 could have a typo in the `<field>SOME_FIELD</field>` definition. 

The faulty xml: 
[faulty role xml.txt](https://github.com/specify/specify7/files/13340966/faulty.role.xml.txt)

The faulty lines:
```
    <format
        name="AccessionAgent"
        title="AccessionAgent"
        class="edu.ku.brc.specify.datamodel.AccessionAgent"
        default="true"
        >
        <switch single="true">
            <fields>
                <field formatter="Agent">agent</field>
                <field sep=" - ">role.</field>
            </fields>
        </switch>
    </format>
```

The crash report that occurs the moment the definition is opened: 
[Specify 7 Crash Report - 2023-11-13T20_32_43.280Z.txt](https://github.com/specify/specify7/files/13341001/Specify.7.Crash.Report.-.2023-11-13T20_32_43.280Z.txt)

maxpatiiuk:
Fixes #4209

1. Don't throw an exception if invalid field name is entred
2. If linting fails in the app resource editor, show an error in the editor rather that nothing

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/codeMirrorLinters.ts
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/codeMirrorLinters.ts
@@ -23,7 +23,20 @@ export const createLinter =
     handleChange: (results: RA<Diagnostic>, view: EditorView) => void
   ): Extension =>
     linter((view) => {
-      const results = handler(view);
+      let results: RA<Diagnostic>;
+      try {
+        results = handler(view);
+      } catch (error) {
+        console.error(error);
+        results = [
+          {
+            from: 0,
+            to: 0,
+            severity: 'error',
+            message: (error as Error).message,
+          },
+        ];
+      }
       handleChange(results, view);
       return results;
     });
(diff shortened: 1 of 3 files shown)


CarolineDenis (approved):

emenslin (approved):
Error now shows up and you are unable to save, looks good. 
Screenshot 2023-12-22 092527

PR #4329, opened 2023-12-26
Rename record merging policy to /record/merge

### LINKED ISSUE #4063: Add record merging permission as a policy
> Is it possible to set permissions so that certain users are prevented from using the merging tool, while retaining full access to the agents table?
> 
> As admin, it would be nice to be able to easily control it via the tool dropdown.

The issue is just because someone has permission to read, create, and update all tables **record by record** they shouldn't necessarily have the ability to perform bulk updates across the database.

There should be an independent permission for merging in the tool menu or query builder

image

**Requested By:** Cris at Gothenburg museum of natural history

melton-jason:
Fixes #4063

### DIFF
--- a/specifyweb/notifications/migrations/0004_rename_merge_policy.py
+++ b/specifyweb/notifications/migrations/0004_rename_merge_policy.py
@@ -1,0 +1,24 @@
+from django.db import migrations
+
+def initialize(apps, schema_editor):
+    UserPolicy = apps.get_model('permissions', 'UserPolicy')
+    LibraryRolePolicy = apps.get_model('permissions', 'LibraryRolePolicy')
+    RolePolicy = apps.get_model('permissions', 'RolePolicy')
+
+    user_policies = UserPolicy.objects.filter(resource='/record/replace')
+    role_policies = RolePolicy.objects.filter(resource='/record/replace')
+    library_role_policies = LibraryRolePolicy.objects.filter(resource='/record/replace')
+
+    user_policies.update(resource="/record/merge")
+    role_policies.update(resource="/record/merge")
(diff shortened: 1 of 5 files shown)


realVinayak (approved):
Tested by making merge policies in previous version of specify, they get correctly renamed!

specifyweb/notifications/migrations/0004_rename_merge_policy.py
maxpatiiuk, line 1: 👏 great job for not forgetting to do this

maxpatiiuk (approved):
looks good - thanks for making the changes
see also my comment in https://github.com/specify/specify7/issues/4063#issuecomment-1871674519

pashiav (approved):
Renamed "Replace" to "Merge". Looks good!
issue-4063

melton-jason:
> 1. I agree that it's redundant to have both update and delete
> 2. I see value in collection admins being able to allow someone to edit a record, without trusting them to bulk-edit those records (which is what record merging is essentially doing). Following this logic, to allow the user to merge Agents, they should have Agent delete, Agent update permissions and /record/replace permission. Is this what front-end and back-end are currently enforcing? if not, we should consider changing it

https://github.com/specify/specify7/issues/4063#issuecomment-1871674519

@specify/ux-testing, @specify/dev-testing 
I think these points are valid: the update and delete record merging permissions should be merged into one permission because they both are associated with the same action and are redundant with the Table update/delete permissions. 

Is this reasonable? 

Regardless, given that this PR already has three approvals, we can likely merge this into `xml-editor` and open a new PR if that is to be implemented.

PR #4330, opened 2023-12-27
Allow handling multiples in scoping

### LINKED ISSUE #3564: `xml-editor`: Hierarchy Exception when defining formatters and aggregators for some tables.
**Describe the bug**
The backend raises a Hierarchy Exception when defining formatters and aggregators for some tables. Need to do more digging into what the cause is. 

**To Reproduce**
1. Go to the visual editor for defining formatters and aggregators.
2. Go to either Formatters menu or the Aggregators menu.
3. Click on table 'Appraisal' and then click on 'Add'

**Expected behavior**
It shouldn't crash and allow to make formatters / aggregators

**Screenshots**
image

Also: HierarchyException when choosing a table on the web link form

Describe the bug
When choosing a table for a web link, a HierarchyException occurs

HierarchyException at /api/specify/agentidentifier/ queryset model Agentidentifier has no hierarchy field 
To Reproduce

Go to a 'WebLinks' app resource
Edit a web link by choosing one from the list (or create a new one)
Choose the 'Agent Identifier' table as a base
See error appear
Screen Recording
https://github.com/specify/specify7/assets/37256050/bae92ea3-323e-4c3e-a0f9-0e9930b0fc3f

[Specify 7 Crash Report - 2023-12-05T15_21_29.112Z.txt](https://github.com/specify/specify7/files/13561952/Specify.7.Crash.Report.-.2023-12-05T15_21_29.112Z.txt)

Also: Creating a new formatter for AccessionCitation throws an error

image
https://kufish51623-issue-3565.test.specifysystems.org/specify/resources/app-resource/9/formatter/AccessionCitation/58

[Specify 7 Crash Report - 2023-10-24T01_18_54.997Z.txt](https://github.com/specify/specify7/files/13096897/Specify.7.Crash.Report.-.2023-10-24T01_18_54.997Z.txt)

Encountered when testing https://github.com/specify/specify7/pull/4114

realVinayak:
Fixes #3564 

For something like Appraisal, which has multiple collection objects, the query should be `collectionobjects__collection=`, which wasn't handled currently.


melton-jason:
This would have the following consequences if `domainfilter=true` is provided as a query parameter: 
- Accession scoped to `collectionobjects__collection`  
  - should be division
 
- Appraisal scoped to `collectionobjects__collection`

- Collectingevent scoped to `collectionobjects__collection` 
  - should be discipline
 
- Fieldnotebookpage scoped to `collectionobjects__collection`
  - should be discipline?  
  
- Paleocontext scoped to `collectionobjects__collection`
  - Previously was being scoped to discipline, but should vary from database to database depending on where paleocontext is scoped, correct? 

- Discipline scoped to `{collections: <Collection Instance>}`
  - should be division

- Division scoped to discipline
  - should be Institution (although this scoping also occurs currently in `xml-editor`) 

- Address scoped to division 

- Institution scoped to division
  - Should not be scoped (or at least should be 'globally' scoped)
  
There may be more cases I am missing, but these are the ones I can find. 

In addition, there are still some models which would throw the HierarchyException. These are models which do not have any 'hiearchy' fields (models like AccessionCitation, SpAppResource, Disposal, etc.). 
While this may be desired  (as `strict` is provided as True by default), in the case of #3564 can we remove the `domainfilter` from the query? I am not too familiar with the `xml-editor` formatter/aggregator editor, but isn't the query is being sent to fetch records to use to preview the Formatter/Aggregator. Is it really necessary here to ensure the resources are in the proper scope? 

https://github.com/specify/specify7/blob/e2d245febde1791a82f3c9ef421fbc761a422fd5/specifyweb/frontend/js_src/lib/components/Formatters/Preview.tsx#L35-L49

realVinayak:
Ah, good catch. Yeah, I thought of just removing the domain filter for this case, but wanted to see how much effort it will take to solve the root problem.

Of the cases you mentioned, in theory, the only ones are a trouble are if the scope gets set to a lower one (like discipline set to collection or institution scoped to divisions), this is because collection/collectionobject lies at the lowest scope possible so if you have (Fieldnotebookpage scoped to `collectionobjects__collection`), it automatically implies (Fieldnotebookpage scoped to `discipline`). 

Will make this a draft PR and see if I can come up with a better way

Issue #4308, opened 2023-12-20
`xml-editor`: Clicking edit or add for a pick list causes the page to reload before the user can take action


grantfitzsimmons:
**Describe the bug**
After loading the Schema Config, if I want to add or edit a pick list from the 'field format' menu, it first reloads the page and I have to click edit or add once again after it finishes.

**To Reproduce**
1. Go to a table in the Schema Config
2. Select a field
3. Try to add a pick list
4. See page refresh
5. Try to add a pick list again
6. See that it works now

**Expected behavior**
It should let you add or edit a pick list immediately, not require a reload.

**Screenshots**

https://github.com/specify/specify7/assets/37256050/6848fc13-ad2b-4f32-a489-e4af56f254f4

**Desktop:**
Arc on macOS

maxpatiiuk:
might be related to https://github.com/specify/specify7/pull/3441 and https://github.com/specify/specify7/issues/3528

realVinayak:
Also happens when you go to a tree and click on "add child" or "edit".

Noticed that this issue doesn't happen without `LazyResourceView` introduced in https://github.com/specify/specify7/pull/4240.

PR #4331, opened 2023-12-27
Handle ReactNodes in Async

realVinayak:
React.lazy needs to be handled in an immediately parented `<Suspense/>`.

`LazyResourceView` was added in https://github.com/specify/specify7/pull/4240, but the `<Suspense/>` wasn't added, leading to remount of the parents.

 We already had logic for doing so for `React.FunctionElement`, this PR extends it to `React.ReactNode`. 

Testing Instructions:

**Tree**
1. In `xml-editor`, go to a tree, and click on a node. Click on 'Add Child' or 'Edit' icon on the top right. Note that the entire tree gets refreshed.
2. Repeat the above step in this branch, and refresh should not happen.

**Picklist** (from https://github.com/specify/specify7/issues/4308)
In `xml-editor`, to reproduce the bug:
1. Go to a table in the Schema Config
3. Select a field
4. Try to add a pick list
5. See page refresh
In this branch, the page would not get refreshed

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
@@ -11,7 +11,6 @@ import { getResourceViewUrl } from '../DataModel/resource';
 import { deserializeResource } from '../DataModel/serializers';
 import { LazyResourceView } from '../Forms/LazyResourceView';
 import type { ResourceView } from '../Forms/ResourceView';
-import { LoadingScreen } from './Dialog';
 
 /**
  * Context created to set a resource not
@@ -74,7 +73,6 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
       <AnyComponent {...allProps} />
       {isOpen && (
         <IsNotReadOnly.Provider value>
-          <React.Suspense fallback={<LoadingScreen />}>
             <LazyResourceView
               dialog="modal"
               onAdd={undefined}
@@ -88,7 +86,6 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
               resource={resource}
               onClose={handleClosed}
             />
-          </React.Suspense>
         </IsNotReadOnly.Provider>
       )}
     </>
--- a/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
@@ -11,6 +11,7 @@ import { getResourceViewUrl } from '../DataModel/resource';
 import { deserializeResource } from '../DataModel/serializers';
 import { LazyResourceView } from '../Forms/LazyResourceView';
 import type { ResourceView } from '../Forms/ResourceView';
+import { LoadingScreen } from './Dialog';
 
 /**
  * Context created to set a resource not
@@ -73,19 +74,21 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
       <AnyComponent {...allProps} />
       {isOpen && (
         <IsNotReadOnly.Provider value>
-          <LazyResourceView
-            dialog="modal"
-            onAdd={undefined}
-            onSaved={(): void => {
-              resourceView.onSaved?.();
-              handleClose();
-            }}
-            {...resourceView}
-            isDependent={false}
-            isSubForm={false}
-            resource={resource}
-            onClose={handleClosed}
-          />
+          <React.Suspense fallback={<LoadingScreen />}>
+            <LazyResourceView
+              dialog="modal"
+              onAdd={undefined}
+              onSaved={(): void => {
+                resourceView.onSaved?.();
+                handleClose();
+              }}
+              {...resourceView}
+              isDependent={false}
+              isSubForm={false}
+              resource={resource}
+              onClose={handleClosed}
+            />
+          </React.Suspense>
         </IsNotReadOnly.Provider>
       )}
     </>
--- a/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
@@ -73,19 +73,19 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
       <AnyComponent {...allProps} />
       {isOpen && (
         <IsNotReadOnly.Provider value>
-            <LazyResourceView
-              dialog="modal"
-              onAdd={undefined}
-              onSaved={(): void => {
-                resourceView.onSaved?.();
-                handleClose();
-              }}
-              {...resourceView}
-              isDependent={false}
-              isSubForm={false}
-              resource={resource}
-              onClose={handleClosed}
-            />
+          <LazyResourceView
+            dialog="modal"
+            onAdd={undefined}
+            onSaved={(): void => {
+              resourceView.onSaved?.();
+              handleClose();
+            }}
+            {...resourceView}
+            isDependent={false}
+            isSubForm={false}
+            resource={resource}
+            onClose={handleClosed}
+          />
         </IsNotReadOnly.Provider>
       )}
     </>
--- a/specifyweb/frontend/js_src/lib/components/Router/ReactLazy.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Router/ReactLazy.tsx
@@ -20,7 +20,7 @@ export function ReactLazy<PROPS extends IR<unknown>>(
   function Wrapped(props: PROPS & React.PropsWithRef<PROPS>): JSX.Element {
     return (
       <React.Suspense fallback={<LoadingScreen />}>
-        {<Lazy {...props} />}
+        <Lazy {...props} />
       </React.Suspense>
     );
   }
--- a/specifyweb/frontend/js_src/lib/components/Router/ReactLazy.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Router/ReactLazy.tsx
@@ -17,7 +17,7 @@ function FakeAsync<PROPS extends IR<unknown>>({
   );
 }
 
-export function LazyAsync<PROPS extends IR<unknown>>(
+export function ReactLazy<PROPS extends IR<unknown>>(
   componentPromise: () => Promise<React.FunctionComponent<PROPS>>
 ) {
   const lazy = React.lazy(async () =>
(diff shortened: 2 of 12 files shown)


specifyweb/frontend/js_src/lib/components/Router/RouterUtils.tsx
```diff
@@ -126,18 +126,20 @@ export const toReactRoutes = (
...
   title,
+  ...rest
 }: {
-  readonly Element: React.FunctionComponent;
-  readonly title: LocalizedString | undefined;
-}): JSX.Element {
+  readonly title?: LocalizedString | undefined;
+} & (
+  | { readonly Element: React.FunctionComponent }
+  | { readonly Node: React.ReactNode }
```
maxpatiiuk, line 135: why not just `Element: React.FunctionComponent | React.ReactNode`?

and then in the code do `typeof Element === 'function'` to differentiate

specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
```diff
@@ -73,18 +74,22 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
...
-            onSaved={(): void => {
-              resourceView.onSaved?.();
-              handleClose();
-            }}
-            {...resourceView}
-            isDependent={false}
-            isSubForm={false}
-            resource={resource}
-            onClose={handleClosed}
+          <Async
```
maxpatiiuk, line 77: (optional) since you are not using Async's set title feature, and it wasn't adapted to handle receiving JSX, I see less value in using Async here instead of `<React.Suspense fallback={<LoadingScreen />}>`

up to you

maxpatiiuk (approved):
thanks for catching this - I misunderstood how Suspense worked.
I wrongfully assumed that it just won't display anything in the split second while the component is being loaded

(optional) btw, consider creating a wrapper around React.Lazy that would do two things:
- Provide a LoadingScreen suspense
- Would not require returning `({default: ComponentName})` (React.Lazy expects default export 🤦 - default exports are evil, won't go into reasons why here)

and then refactor the usages of React.Lazy in the codebase to use it
then, if we use that wrapper, that would make bugs like this less likely

specifyweb/frontend/js_src/lib/components/Router/RouterUtils.tsx
realVinayak: Hmm both of them are of type objects. I experimented with putting in above function definition
```
  console.log(
    Object.keys(rest)[0],
    typeof ('Element' in rest ? rest.Element : rest.Node)
  );
```
and got 'object' for both of them (Element, and Node). I suspect React handles conversion for us when actually calling it from lazy.

specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
realVinayak: Yeah, I think it took "sharing code" too far here (see https://github.com/specify/specify7/pull/4331/commits/adbd22b8a809c729af1b866f9dfe7350f3b892e3), where it was just Suspense and I used Async afterwards

realVinayak: Will revert to that commit, doesn't seem worth here

maxpatiiuk (approved):

maxpatiiuk:
also, the PR that added React.Lazy in this place, added it in two more places. see:

https://github.com/specify/specify7/pull/4240/files#diff-5764850b98cd53c725244fa87e99d2b8e09579461068f57e832325ee5652120bR82

and

https://github.com/specify/specify7/pull/4240/files#diff-0d80cdc583b8c203e6a7c081feda065cb2796bd9a9e105dabad93f4219737724R7

do those need to be wrapped in React.Suspense too?

(optional) consider creating a wrapper I mentioned in https://github.com/specify/specify7/pull/4331#pullrequestreview-1798551656 and using that instead of React.Suspense to avoid such bugs

CarolineDenis (approved):

emenslin (approved):
Looks good, page no longer gets refreshed in either instance.

realVinayak:
> (optional) consider creating a wrapper I mentioned in https://github.com/specify/specify7/pull/4331#pullrequestreview-1798551656 and using that instead of React.Suspense to avoid such bugs

@maxpatiiuk made a wrapper for it. Do you see a better way to type it?

specifyweb/frontend/js_src/lib/components/Router/RouterUtils.tsx
```diff
@@ -126,19 +125,20 @@ export const toReactRoutes = (
...
 export function Async({
-  Element,
+  componentPromise,
   title,
 }: {
-  readonly Element: React.FunctionComponent;
+  readonly componentPromise: () => Promise<React.FunctionComponent>;
   readonly title: LocalizedString | undefined;
 }): JSX.Element {
+  const Lazy = LazyAsync(componentPromise);
```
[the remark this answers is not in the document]
maxpatiiuk: I would suggest you just get rid of Async, and inline the call to LazyAsync above in this file

the other usage of Async (taxonTiles) can be refactored to use `ReactLazy`

reasoning for this is that I would rather not have React.lazy call create a new component on every URL route - it should be done once. By doing this work inside of a component (Async), the work is happening for every route (even if you use useMemo)

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/Router/RouterUtils.tsx
```diff
@@ -80,12 +81,12 @@ export const toReactRoutes = (
 
     const resolvedElement =
```
maxpatiiuk, line 82: I noticed a bug here (was preseent before)
If rawElement is a component, not a function, then `title` is ignored
(line 95 may need to be modified to include title too)
in that case, factor our AppTitle ternary into a common variable to not duplicate it

specifyweb/frontend/js_src/lib/components/Router/RouterUtils.tsx
```diff
@@ -126,19 +127,20 @@ export const toReactRoutes = (
  * when any component is being loaded.
  */
 export function Async({
```
maxpatiiuk, line 129: this function can be deleted now?

specifyweb/frontend/js_src/lib/components/Router/ReactLazy.tsx
```diff
@@ -0,0 +1,32 @@
...
+  readonly props: PROPS;
+}) {
+  return (
+    <React.Suspense fallback={<LoadingScreen />}>
+      <Element {...props} />
+    </React.Suspense>
+  );
+}
+
+export function ReactLazy<PROPS extends IR<unknown>>(
```
maxpatiiuk, line 20: can you add this comment to ReactLazy?

```ts
/**
 * Using this allows Webpack to split code bundles.
 * React Suspense takes care of rendering a loading screen if component is
 * being fetched.
 * Having a separate Suspense for each async component rather than a one main
 * suspense on the top level prevents all components from being un-rendered
 * when any component is being loaded.
 */
```

maxpatiiuk: (comment coppied from the comment Async function had)

and lol, reading the comment I see that I mentioned the exact thing that caused #4308 to happen 🤣
> * Having a separate Suspense for each async component rather than a one main
> * suspense on the top level prevents all components from being un-rendered
> * when any component is being loaded.

my react skills are getting rusty🤣 (we don't use React at esri - but we have a Frankenstein called Esriact - less said about it, the better)

specifyweb/frontend/js_src/lib/components/Router/ReactLazy.tsx
```diff
@@ -0,0 +1,32 @@
...
+  componentPromise: () => Promise<React.FunctionComponent<PROPS>>
+) {
+  const lazy = React.lazy(async () =>
+    componentPromise().then((module) => ({ default: module }))
+  );
+  function Wrapped(props: React.PropsWithRef<PROPS>): JSX.Element {
+    return (
+      <FakeAsync<React.PropsWithRef<PROPS>> Element={lazy} props={props} />
```
maxpatiiuk, line 28: FakeAsync is not needed anymore? you can just inline it's definition in here

specifyweb/frontend/js_src/lib/components/Router/ReactLazy.tsx
maxpatiiuk, line 20: (nitpick) for the future, try to have the main function at the top of the file (ReactLazy), and other variables and functions called by it (FakeAsync) below it in the order they appear in the main function (ReactLazy)

this way code can be read from top to bottom, in and order that is closer to order in which it executes (rather than jumping from bottom to top and back)

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/Router/RouterUtils.tsx
realVinayak: fixed

specifyweb/frontend/js_src/lib/components/Router/ReactLazy.tsx
realVinayak: Had to adjust types, but did it

specifyweb/frontend/js_src/lib/components/Router/ReactLazy.tsx
realVinayak: thanks for reminding that!

realVinayak: I thought that React would be smart and try to find the first parent Suspense (which would have not caused https://github.com/specify/specify7/issues/4308) , but in my experimentation it doesn't do that, so React.lazy needs to be immediately parented by the Suspense!

emenslin (approved):
Looks good, no refreshing with the trees or in schema config.

maxpatiiuk (approved):
thanks!

specifyweb/frontend/js_src/lib/components/Router/ReactLazy.tsx
```diff
@@ -0,0 +1,28 @@
...
+  );
+  function Wrapped(props: PROPS & React.PropsWithRef<PROPS>): JSX.Element {
+    return (
+      <React.Suspense fallback={<LoadingScreen />}>
+        <Lazy {...props} />
+      </React.Suspense>
+    );
+  }
+  return Wrapped;
+}
```
maxpatiiuk, line 28: lol, someone made a library just for this - https://github.com/gregberge/loadable-components
and the library gets 100k downloads/week

PR #4332, opened 2023-12-27
Ignore usertype when searching for Common dir

realVinayak:
Implements second solution in
https://github.com/specify/specify7/issues/3956#issuecomment-1853251752

Note: This will cause unexpected behavior if resources with same name appears under multiple usertypes at Global level ("Common"). I don't know if that case happens or not.

@grantfitzsimmons Have you ever seen that?

### DIFF
--- a/specifyweb/context/app_resource.py
+++ b/specifyweb/context/app_resource.py
@@ -30,10 +30,6 @@ disciplines = ElementTree.parse(disc_file)
 discipline_dirs = dict( (disc.attrib['name'], disc.attrib.get('folder', disc.attrib['name']))
     for disc in disciplines.findall('discipline') )
 
-# Defines which fields to ignore when looking up
-class Ignore(object):
-    pass
-
 # get_app_resource is the main interface provided by this module
 def get_app_resource(collection, user, resource_name):
     """Fetch the named app resource in the context of the given user and collection.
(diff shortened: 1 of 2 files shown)


specifyweb/context/app_resource.py
```diff
@@ -125,46 +125,53 @@ def get_app_resource_from_db(collection, user, level, resource_name):
...
         if user is None: return []
-        filters['specifyuser'] = user
-        # when filtering by user, usertype is implied
-        # also if a user's type is changed, what would happen?
-        del filters['usertype']
+        raw_filters['specifyuser'] = user
+
+    filters = {key: value
+               for (key, value) in raw_filters.items()
+               if not isinstance(value, Ignore)}
```
maxpatiiuk, line 174: don't know about Python, but in JavaScript, instanceof is relatively expensive, and this function is in the hot path (called several times on every page load)

could we do something like this instead?
```py
ignore = {}

# and later:

if value != ignore

```

specifyweb/export/feed.py
```diff
@@ -20,14 +20,8 @@ class MissingFeedResource(Exception):
...
-    try:
-        return Spappresourcedata.objects.get(
-            spappresource__name="ExportFeed",
-            spappresource__spappresourcedir__usertype="Common",
-            spappresource__spappresourcedir__discipline=None,
-        ).data
-    except Spappresourcedata.DoesNotExist:
-        return None
+    from_db = get_app_resource_from_db(None, None, 'Common', 'ExportFeed')
+    return None if from_db is None else from_db[0]
```
maxpatiiuk, line 24: great refactoring - thanks

maxpatiiuk (approved):

specifyweb/context/app_resource.py
realVinayak: Didn't know js did that!

Fixed in https://github.com/specify/specify7/commit/268ae03e44b0e2923cfb9c340ccdda4c9535ce26

maxpatiiuk (approved):
👏

CarolineDenis (approved):

CarolineDenis (approved):

realVinayak:
Testing instructions:

These are specific to `insectcollection_umn` database on the test panel.

1. Create ExportFeed resource under the Global Resources, so at the top level. It is fine if the ExportFeed is empty:
image

2. Run the feed by clicking on "Create DwC Archive" in User Tools
image

3. This branch should "run" the feed (you will not get actual results till you set the feed correctly). However, `xml-editor` should give a notification with the error message saying "MissingFeedResource" somewhere in the crash report.

emenslin (approved):
Looks good, did not get the notification with the error message.

grantfitzsimmons:
> Note: This will cause unexpected behavior if resources with same name appears under multiple usertypes at Global level ("Common"). I don't know if that case happens or not.
> 
> @grantfitzsimmons Have you ever seen that?

I can't think of situations like this at the global level, but I would not endorse it.

PR #4333, opened 2023-12-27
Set default value before calling update

realVinayak:
The issue was the default value was being set after updateValue was being called (since LayoutEffect gets called before Effect). So, even `setValue` never received the default value (even though the resource had the default value). 

That is, the underlying object had `#########` as the catalog number, but state only gets updated by updateValue which didn't have the default value at the time of calling.

Funny enough this issue does not happen in React's StrictMode, just because since the hook to set default value gets called after the second hook, So, during the next cycle (React calls all effect twice in strict mode), the updateValue had the default value (because of mutation), so it updated the state correctly. It was fun to see everything working locally but not on the test panel (since production mode doesn't do StrictMode)

### DIFF
--- a/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
+++ b/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
@@ -164,7 +164,7 @@ export function useResourceValue<
   );
 
   // Set default value
-  React.useEffect(() => {
+  React.useLayoutEffect(() => {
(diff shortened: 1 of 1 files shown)


maxpatiiuk (approved):
good job debugging this one!

funny: one could say useResouceValue is just for getting and setting the value for a single field. who knew that takes 227 lines of complicated code (in addition to all the mess in resoruceApi.js)

pashiav (approved):
Looks good!
bf-af-4321

Issue #4286, opened 2023-12-14
Migrate off backbone.js and jQuery


maxpatiiuk:
We talked about this a lot, but didn't have an explicit issue for this.

Backbone.js is used for representing database resources on the front-end. That library also uses jQuery. That's the only place where jQuery is still used.

Backbone.js is not very compatible with the way react works (it's not reactive at all - making relatively simple things very complicated due to need for manually setting up listeners in useEffect - see https://github.com/specify/specify7/pull/4264 for an example), is not very type-safe (may return null or undefined, rather than always undefined)

Closely related to https://github.com/specify/specify7/issues/2621 (and see https://github.com/specify/specify7/issues/2621#issuecomment-1478779191)

This is going to be a somewhat large project because:
- Resources are being used in a lot of places
- Data integrity is the most important thing. Need to be sure not to introduce bugs. Need to write good tests
- There is also a concept of SerializedResource that was a temporary stop-gap measure and should be merged with the final solution

We could write our own library for representing resources on the front-end, but it would be best to do at least some research into available options before doing so (surely some new nice thing has became available in the last 15 years since backbone.js)

maxpatiiuk:
What I am dreaming of:
1. A central store that holds all fetched objects in the front-end
2. React components receive immutable references to those objects (either table name + id, or table name with temporary guid for new objects*)

then, inside the component, it can call a hook like useResourceField or useResource, that would in turn resolve a single field value from the object, or entire object - and as a result, re-render on that field change, or any change
in non-reactive code (business rules), you would still be able to resolve the reference to an object to the mutable object itself

*so the above can be implemented even with backbone remaining in place (in fact, we have a useResource that kind of does that already), but there are other issues with backbone, so bandaids like this are not the best long-term solution

*on further thought, probably all references should be table name + front-end only GUID, so that if new resource becomes saved, the previous reference still holds true

At the same time, I would ideally like to have a separate typescript type for saved resource (or reference to a resource), new resource, or a type that accepts both - this would add type safety and would clearly show whether react component expects to receive a saved resource, or if it doesn't care either way

Additional benefit of central store for all objects is the fact that we can load the same object again in the code, and not worry about caching as load would get the object from the central store
also means that any changes to the object would be centralized between all the components that use that object, avoiding out of sync bugs AND worse, potential data loss when there are two instances of the same database object

So now just to find a library that can do the above, or as close as we can, or else implement our own

maxpatiiuk:
[...]

PR #4339, opened 2024-01-01
Migrate off backbone.js and jQuery

maxpatiiuk:
Begun work on #4286. Read comments in that issue for architecture details and justification - feedback is welcome

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Resource/resourceApi.ts
+++ b/specifyweb/frontend/js_src/lib/components/Resource/resourceApi.ts
@@ -1,0 +1,105 @@
+import { eventListener } from '../../utils/events';
+import { addMissingFields } from '../DataModel/addMissingFields';
+import { SerializedResource } from '../DataModel/helperTypes';
+import { fetchResource as internalFetchResource } from '../DataModel/resource';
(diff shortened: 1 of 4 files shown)


specifyweb/frontend/js_src/lib/components/Resource/resourceApi.ts
```diff
@@ -0,0 +1,140 @@
...
+  const savedResource = savedResourceStore[tableName]?.[id]?.deref();
+  if (typeof savedResource === 'object') return savedResource;
+
+  // Simplified:
+  const data = await internalFetchResource(tableName, id);
+  const ref: SavedResourceRef<TABLE_NAME> = {
+    table: tables[tableName],
+    id,
+  };
+  const resource: SavedResource<TABLE_NAME> = { ...ref, ref, data };
```
maxpatiiuk, line 72: ref includes table and id properties
it is useful to have these properties on the resource
but, it may also be useful to be able to quickly get from resource back to resource ref, so the `ref` property exists
if it turns out to be redundant/needless, we could remove it (in favor of doing `{ table: resource.table, id: resource.id }` - which could be a helper function `resourceToRef`)

note, resource refs are not globally unique - but, their contents (table name + id pair) resolve to one resource only

PR #4340, opened 2024-01-02
Set lastFocusedRow to undefined when refreshing in tree

realVinayak:
Needed because `lastFocusedRow` doesn't get updated immediately (only gets updated when the focused node is rendered). This causes an issue where rows get updated (and TreeViewActions) are mounted again, but `lastFocusedRow` still refers to the node from the previous action before refresh (which will be the deleted node in case of deletion). Since `lastFocusedRow` isn't updated yet, businessrules get run for the deleted node (which doesn't exist, so 404 error)

Testing instructions
1. Go to a tree.
2. Click on a node, and delete it using trash icon or  by clicking on pencil icon and then clicking on delete button.
3. The `xml-editor` should throw 404 not found, this branch should not.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
@@ -281,7 +281,10 @@ function TreeView<SCHEMA extends AnyTree>({
             onRefresh={(): void => {
               // Force re-load
               setRows(undefined);
-              globalThis.setTimeout(() => setRows(rows), 0);
+              globalThis.setTimeout(() => {
+                setLastFocusedRow(undefined);
+                setRows(rows);
+              }, 0);
             }}
           />
         </ErrorBoundary>
--- a/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
@@ -123,8 +123,8 @@ export function TreeViewActions<SCHEMA extends AnyTree>({
             nodeId={focusedRow?.nodeId}
             tableName={tableName}
             onDeleted={() => {
+              handleRefresh();
               setFocusPath(focusPath?.slice(0, -1));
-              setTimeout(handleRefresh, 0);
             }}
           />
         </li>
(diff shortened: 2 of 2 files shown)


specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -281,7 +281,10 @@ function TreeView<SCHEMA extends AnyTree>({
             onRefresh={(): void => {
               // Force re-load
               setRows(undefined);
-              globalThis.setTimeout(() => setRows(rows), 0);
+              globalThis.setTimeout(() => {
+                setLastFocusedRow(undefined);
```
maxpatiiuk, line 285: this is only needed if we deleted the focused row right?
in all other cases, this would needlessly cause the focused node to be lost?

edit: looks like the focusedRow is only used by TreeViewActions. so it would cause "Edit" and "Query" button to stop working after you made an edit, until you focus another node

can you confirm if that's what happening? if so, we might want to resolve it. otherwise all good

resolution:
- add an argument to onRefresh like `cause: 'edit' | 'delete'` or something like that

realVinayak: Ah, it still works since `focusPath` sets `lastFocusedNode`. That is, even if `lastFocusedNode` is undefined, when the node at `focusPath` gets mounted, it will set the `lastFocusedNode` to itself. Since `focusPath` is still a valid value during other tree operations (synonymize, move, merge), `lastFocusedNode` is not lost (that's a lie, it gets lost immediately after setting to undefined, but it eventually gets set to node at end of `focusPath`)
https://github.com/specify/specify7/blob/e2d9b08a6a61b91723f06b5192494a4ad1948697/specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx#L65-L68

realVinayak: ^confirmed via testing

emenslin (approved):
Looks good, no longer get the 404 error after deleting.

CarolineDenis (approved):

maxpatiiuk (approved):
👍

PR #4343, opened 2024-01-02
Add web link preview

### LINKED ISSUE #4335: `xml-editor`: Add weblink's text next to the icon
Currently, to see the link, I'd have to rely on browser making a little popup on the bottom left corner on highlighting the weblink. I think it is worth making the link visible right next to the button.

@specify/ux-testing @CarolineDenis @grantfitzsimmons thoughts?

Screenshot (321)

CarolineDenis:
Fixes #4335

### DIFF
--- a/specifyweb/frontend/js_src/css/main.css
+++ b/specifyweb/frontend/js_src/css/main.css
@@ -240,10 +240,6 @@
   optgroup {
     @apply font-bold;
   }
-
-  .text-transform-none {
-    text-transform: none;
-  }
 }
 
 /**
--- a/specifyweb/frontend/js_src/css/main.css
+++ b/specifyweb/frontend/js_src/css/main.css
@@ -240,6 +240,10 @@
   optgroup {
     @apply font-bold;
   }
+
+  .text-transform-none {
+    text-transform: none;
+  }
 }
 
 /**
--- a/specifyweb/frontend/js_src/lib/components/WebLinks/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/WebLinks/index.tsx
@@ -155,7 +155,7 @@ export function WebLinkField({
             }}
           >
             {image}
-            {localized(url ?? '')}
+            <div className="text-transform-none">{localized(url ?? '')}</div>
           </Component>
           {showPrompt && (
             <PromptDialog
--- a/specifyweb/frontend/js_src/lib/components/WebLinks/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/WebLinks/index.tsx
@@ -6,7 +6,6 @@ import { useId } from '../../hooks/useId';
 import { commonText } from '../../localization/common';
 import { getAppResourceUrl, isExternalUrl } from '../../utils/ajax/helpers';
 import type { GetSet, IR, RA } from '../../utils/types';
-import { localized } from '../../utils/types';
 import { caseInsensitiveHash } from '../../utils/utils';
 import { Ul } from '../Atoms';
 import { Button } from '../Atoms/Button';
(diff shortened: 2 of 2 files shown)


CarolineDenis:
Screenshot 2024-01-02 at 9 21 34 AM

specifyweb/frontend/js_src/css/main.css
```diff
@@ -240,6 +240,10 @@
   optgroup {
     @apply font-bold;
   }
+
+  .text-transform-none {
```
maxpatiiuk, line 244: you can use the existing tailwind className - [`normal-case`](https://tailwindcss.com/docs/text-transform#:~:text=%3A%20capitalize%3B-,normal%2Dcase,-text%2Dtransform%3A)

specifyweb/frontend/js_src/lib/components/WebLinks/index.tsx
```diff
@@ -154,6 +155,7 @@ export function WebLinkField({
             }}
           >
             {image}
+            <div className="text-transform-none">{localized(url ?? '')}</div>
```
maxpatiiuk, line 158: this might be equivalent, and simpler:

```suggestion
            <div className="text-transform-none">{url}</div>
```

maxpatiiuk: fyi: LocalizedString is enforced for direct children of Button, but not for indirect, which is why you don't need to call it here anymore

specifyweb/frontend/js_src/lib/components/WebLinks/index.tsx
```diff
@@ -154,6 +155,7 @@ export function WebLinkField({
...
           </Component>
```
maxpatiiuk, line 159: I was [suggesting](https://github.com/specify/specify7/issues/4335#issuecomment-1871684164) we put the URL outside the button, not inside:

Screenshot 2024-01-02 at 18 36 35

Reasons:
- URL can be selected without triggering the button
- The button looks the way it would look like on the form
- Better for accessibility (having a URL as button label is going to sound bad in the screen reader)

maxpatiiuk: Also, I just realized that this component is the same on that is used on the form - we don't want the weblink on the form to print the URL

So, you should add something like this in the component to check if we are in the web link editor:

```ts
  const isInEditor = React.useContext(WebLinksContext) !== undefined;
```

and then only print the URL preview if we are in the weblink editor

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/WebLinks/index.tsx
```diff
@@ -139,22 +143,27 @@ export function WebLinkField({
...
+                  definition.parts.some(({ type }) => type === 'PromptField')
+                ) {
+                  event.preventDefault();
+                  setShowPrompt(true);
+                }
+              }}
+            >
+              {image}
+            </Component>
+            {isInEditor ? url : null}
```
maxpatiiuk, line 165: you could just wrap this url in div, not need to wrap Component too, is there?

```tsx
{isInEditor && <div className="flex items-center gap-2">{url}</div>}
```

emenslin (approved):
Looks good, link is now visible next to the button. 
Screenshot 2024-01-04 104053

pashiav (approved):
Links are displayed. Looks good!
issue-4335

specifyweb/frontend/js_src/lib/components/WebLinks/index.tsx
```diff
@@ -155,6 +159,7 @@ export function WebLinkField({
           >
             {image}
           </Component>
+          {isInEditor && <div className="flex items-center gap-2">{url}</div>}
```
maxpatiiuk, line 162: gap is no longer needed here

maxpatiiuk (approved):
❤️

PR #4344, opened 2024-01-02
Change use localized field label checkbox to button

### LINKED ISSUE #4341: 'Use localized field labels' checkbox in App Resource should be a button
From @maxpatiiuk comment: 

> Given that the "Use localized field labels" checkbox is reloading the page, it would be more semantically correct to have that as a button.
> I know that it's reused with the Form Meta menu, so we could convert it to a button there too.
> If it's a button, the label would have to be different depending on on/off state

CarolineDenis:
Fixes #4341

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
@@ -198,8 +198,8 @@ function UseLabelsSchema(): JSX.Element {
   return (
     <Button.Secondary onClick={update}>
       {useFieldLabels
-        ? formsText.localizedLabels()
-        : formsText.databaseLabels()}
+        ? formsText.useDatabaseLabels()
+        : formsText.useLocalizedLabels()}
     </Button.Secondary>
   );
 }
--- a/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
@@ -198,8 +198,8 @@ function UseLabelsSchema(): JSX.Element {
   return (
     <Button.Secondary onClick={update}>
       {useFieldLabels
-        ? formsText.useDatabaseLabels()
-        : formsText.useLocalizedLabels()}
+        ? formsText.showDataModelLabels()
+        : formsText.showLocalizedLabels()}
     </Button.Secondary>
   );
 }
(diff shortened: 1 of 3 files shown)


specifyweb/frontend/js_src/lib/localization/forms.ts
```diff
@@ -986,6 +986,12 @@ export const formsText = createDictionary({
     'uk-ua': 'Використовуйте локалізовані назви полів',
     'de-ch': 'Verwenden Sie lokalisierte Feldbezeichnungen',
   },
+  localizedLabels: {
+    'en-us': 'Localized Labels',
```
maxpatiiuk, line 990: since this is used as a button label, can you reword it to be something like `Display localized field labels` and `Display database field names` 

(could change `Display` to `Use` if you want)

maxpatiiuk: also, if `useFieldLabels` is no longer used, then remove that string

specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
```diff
@@ -197,10 +196,11 @@ function UseLabelsSchema(): JSX.Element {
   };
 
   return (
-    <Label.Inline>
-      <Input.Checkbox checked={useFieldLabels} onValueChange={update} />
-      {formsText.useFieldLabels()}
-    </Label.Inline>
+    <Button.Secondary onClick={update}>
+      {useFieldLabels
```
maxpatiiuk, line 200: Shouldn't these labels be swapped?

i.e if we currently use localized field labels, then the button should say `Use Database Field Names` so that the user knows what action to expect when they click the button

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/FormEditor/Editor.tsx
CarolineDenis: yes indeed!

CarolineDenis:
I'm not convinced of the button in the meta menu dialog as the reload only happens once the user closes the dialog.

emenslin (approved):
Looks good, I think a button here makes a lot more sense. 
Screenshot 2024-01-04 102436

pashiav (approved):
Looks good!

https://github.com/specify/specify7/assets/118556884/f5c9f42a-9e22-49bc-8575-189bd9c5080b

specifyweb/frontend/js_src/lib/localization/forms.ts
```diff
@@ -986,6 +986,12 @@ export const formsText = createDictionary({
     'uk-ua': 'Використовуйте локалізовані назви полів',
     'de-ch': 'Verwenden Sie lokalisierte Feldbezeichnungen',
   },
+  showLocalizedLabels: {
```
maxpatiiuk, line 989: (optional) to better match the naming of the `useFieldLabels` option, could rename this to `showFieldLabels` and `showDataModelLabels`

maxpatiiuk (approved):

PR #4345, opened 2024-01-02
Add SpecifyUser to Query Combo Box in Export Feed

realVinayak:
Added a `typeSearch` to allow searching. Users will have to configure a formatter for SpecifyUser to see anything other than  `<Formatter not defined>`

Testing instructions
0. Make sure you have a formatter defined for SpecifyUser table. Add one if not.
1. Go to ExportFeed resource in xml-editor.
2. Go to `Run as user` or `Send completion notification to user` field (which is query combo box)
3. Make sure you are able to search for different users in those combo boxes
image

It is rare that databases will have a formatter for specifyUser (but that's easy to add with xml-editor, so I think it is fine). Until https://github.com/specify/specify7/pull/4312 is merged, SpecifyUser won't be visible in table list for visual editor. So, the following can be added to record object formatters to add a basic formatter for SpecifyUser
```
<format
    name="SpecifyUser"
    title="SpecifyUser"
    class="edu.ku.brc.specify.datamodel.SpecifyUser"
    default="true"
    >
    <switch single="true">
        <fields>
            <field>name</field>
        </fields>
    </switch>
</format>

```

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/ExportFeed/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/ExportFeed/Editor.tsx
@@ -30,6 +30,8 @@ import { LoadingScreen } from '../Molecules/Dialog';
 import { QueryComboBox } from '../QueryComboBox';
 import { dwcaAppResourceFilter, PickAppResource } from './Dwca';
 import type { ExportFeedDefinition } from './spec';
+import { TypeSearch } from '../QueryComboBox/spec';
+import { QueryFieldSpec } from '../QueryBuilder/fieldSpec';
 
 export function ExportFeedEditor({
   definition: [definition, setDefinition],
@@ -340,6 +342,16 @@ function ResourcePicker({
   );
 }
 
+const specifyUserTypeSearch: TypeSearch = {
+  displayFields: undefined,
+  table: tables.SpecifyUser,
+  searchFields: [[getField(tables.SpecifyUser, 'name')]],
+  name: localized('SpecifyUserName'),
+  title: getField(tables.SpecifyUser, 'name').label,
+  formatter: undefined,
+  format: undefined,
+};
+
 function UserPicker({
   id: [id, setId],
   isRequired,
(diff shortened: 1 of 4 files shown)


specifyweb/frontend/js_src/lib/components/ExportFeed/Editor.tsx
```diff
@@ -340,6 +341,16 @@ function ResourcePicker({
   );
 }
 
+const specifyUserTypeSearch: TypeSearch = {
+  displayFields: undefined,
+  table: tables.SpecifyUser,
+  searchFields: [[getField(tables.SpecifyUser, 'name')]],
```
maxpatiiuk, line 347: this is not safe - you should not be accessing `tables` in global scope as it's only defined after the datamodel is fetched

it might work right now if this file does not get imported before datamode is fetched, but it may randomly stop working in the future

solutions:
1. use f.store() around entire object - this is what we done in other places - read the comment above f.store() as well as look at some usages to see how that fixes it
2. OR move this object into useMemo inside of UserPicker (but 1st solution is probably better)

maxpatiiuk (changes requested):
now that you identified the solution, I can see what might have caused it:

https://github.com/specify/specify7/blob/e2d9b08a6a61b91723f06b5192494a4ad1948697/specifyweb/frontend/js_src/lib/components/QueryComboBox/useTypeSearch.tsx#L37-L38

that code says that if type search wasn't explicitly passed, it would try to find a type search for the specify user table

probably I had a type search for specify user defined in my database when I was testing this, but your database doesn't?

if that is the case, your solution is good - this kind of stuff shouldn't be database dependent

and even better, if typesearch doesn't exist, we might be able to define one on the fly

specifyweb/frontend/js_src/lib/components/ExportFeed/Editor.tsx
realVinayak: Fixed!

maxpatiiuk (approved):

emenslin (approved):
I did have to add the formater but it looks good and you can search for users. 
Screenshot 2024-01-03 085547

CarolineDenis (approved):

PR #4346, opened 2024-01-02
Prompt user to define record set name when creating one

### LINKED ISSUE #4304: Users are not prompted to create a name for a new record set after using the "create record set" button
**Describe the bug**
Users are not prompted to create a name for a new record set after using the "create record set" button. This leads to a new record set being created without a name. **This name should be required**.

**To Reproduce**
1. Create several records in a table
2. Click on the 'new record set' button in the top right
3. See a new record set is created, but no name is given to it
4. See the record sets have no name in the record set dialog

**Screen Recording**
https://github.com/specify/specify7/assets/37256050/d3dda18f-5d2a-474a-a3aa-5c63a10cf58f

**Expected behavior**
When clicking on the 'new record set' button, the user should be asked to give the record set a name. Record sets **should not exist** without a name.

image

`v7.9.2`

CarolineDenis:
Fixes #4304


specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -320,6 +322,10 @@ function RecordSet<SCHEMA extends AnySchema>({
       )
     ).then(f.void);
 
+  const isTitlePresent = recordSet.get('name') === '' && !currentRecord.isNew();
```
maxpatiiuk, line 325: the variable is called `isTitlePresent`, but you are checking for `=== ''`? a bug, or poor variable naming?

maxpatiiuk: Wouldn't it be better to instead open the dialog when user tries to create a record set?

https://github.com/specify/specify7/blob/79fc28bcf32ea8326b1f8c613afeab3c8613e4b8/specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx#L336

(so instead of creating the record set, ask user for the name first)

and then, in the onSaved callback from ResourceView, call the createNewRecordSet function

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -320,6 +322,8 @@ function RecordSet<SCHEMA extends AnySchema>({
       )
     ).then(f.void);
 
+  const [openDialogForTitle, setOpenDialogForTitle] = React.useState(false);
```
maxpatiiuk, line 325: (optional) can use useBooleanState if you wish

maxpatiiuk (approved):
👍

bronwyncombs (approved):
Creating a record set from data entry and query results prompts to name upon adding 👍   
Screenshot 2024-01-04 at 12 05 31 PM
Screenshot 2024-01-04 at 12 04 34 PM

emenslin (approved):
Looks good, user is now forced to name record set.
Screenshot 2024-01-04 121434

Issue #4347, opened 2024-01-03
`xml-editor`: Cloned resource dialog disappears in tree viewer


realVinayak:
1. Go to any taxa tree.
2. Click on a node, and then click on Edit button (Pencil icon)
3. Click on Clone. 
4. Notice that the dialog has disappeared.

PR #4354, opened 2024-01-03
Use ResourceView instead of ResourceLink in TreeViewer

realVinayak:
Debugging issues for Tree viewer for a decent time, I realized that ResourceLink has been causing issues specifically in TreeViewer as seen in https://github.com/specify/specify7/pull/4230 and https://github.com/specify/specify7/pull/4340, and https://github.com/specify/specify7/issues/4347.

For the first two PRs, I've been able to find ways still using  `ResourceLink`. However for https://github.com/specify/specify7/issues/4347, it is tricky since `setResource` is used to handle cloned resources. 
https://github.com/specify/specify7/blob/b365affc238a1e1dcd7a7a5f39f67d74050998fb/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L251
Since the state is passed to `ResourceLink`, it gets changed whenever a cloned resource is created. This closes the dialog:
https://github.com/specify/specify7/blob/b365affc238a1e1dcd7a7a5f39f67d74050998fb/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx#L44

So, to solve the problem, we could either
1. Add  a boolean state in EditRecordDialog  (used in tree, parent of ResourceLink) to keep ResourceLink open, and pass that prop down to ResourceLink.
2. Make a new state that stores the cloned resource, and render ResourceView for this state.
3. Use ResourceView from the beginning.

Currently, I've implemented the third one. @maxpatiiuk ideas on above/new solutions? I would rather keep ResourceLink to just do one thing: show a link, and the resource, and not worry about parent component for anything so didn't do the first solution. The second solution seems half-way to the third one so didn't go with second one.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
@@ -19,10 +19,12 @@ import { genericTables } from '../DataModel/tables';
 import { DeleteButton } from '../Forms/DeleteButton';
 import { getPref } from '../InitialContext/remotePrefs';
 import { Dialog } from '../Molecules/Dialog';
-import { ResourceLink } from '../Molecules/ResourceLink';
+import { IsNotReadOnly } from '../Molecules/ResourceLink';
 import { hasPermission, hasTablePermission } from '../Permissions/helpers';
 import type { Row } from './helpers';
 import { checkMoveViolatesEnforced } from './helpers';
+import { useBooleanState } from '../../hooks/useBooleanState';
+import { LazyResourceView } from '../Forms/LazyResourceView';
 
 type Action = 'add' | 'desynonymize' | 'edit' | 'merge' | 'move' | 'synonymize';
 
@@ -222,10 +224,13 @@ function EditRecordDialog<SCHEMA extends AnyTree>({
   readonly isRoot: boolean;
   readonly onRefresh: () => void;
 }): JSX.Element | null {
+  const [isOpen, _, handleClose, handleToggle] = useBooleanState();
+
   const [resource, setResource] = useLiveState<
     SpecifyResource<AnySchema> | undefined
   >(
     React.useCallback(() => {
+      if (!isOpen) return undefined;
       const table = genericTables[tableName] as SpecifyTable<AnyTree>;
       const parentNode = new table.Resource({ id: nodeId });
       let node = parentNode;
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
```diff
@@ -234,25 +239,33 @@ function EditRecordDialog<SCHEMA extends AnyTree>({
...
-      resource={resource}
-      resourceView={{
-        dialog: 'nonModal',
-        onAdd: isRoot ? undefined : setResource,
-        onDeleted: handleRefresh,
-        onSaved: handleRefresh,
-      }}
-    />
+    <>
+      <Button.Icon
```
maxpatiiuk, line 247: can you turn this into Link.Icon like it was before?
and then in onClick do `(event): void => { event.preventDefault(); handleToggle(); }`

this way, users can still right click on the link to open it in new tab, but on regular click it acts like a button

maxpatiiuk (changes requested):
could you instead add a prop like `autoClose={false}` to ResourceLink?

that would require less code, and may come in handy in other places too

specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
realVinayak: Thought about it, but then you have another bug: User clicks on edit. Clicks on Clone (resource state gets set to clone resource). Clicks on Close button. Resource is still the cloned resource. So, the next time they click on Edit it will show cloned resource (until they click somewhere else and then back on current node)

This bug doesn't happen in production because dependency of callback in live state includes  `isOpen`

realVinayak:
@maxpatiiuk regarding autoClose prop

Thought about it, but then you have another bug: User clicks on edit. Clicks on Clone (resource state gets set to clone resource). Clicks on Close button. Resource is still the cloned resource. So, the next time they click on Edit it will show cloned resource (until they click somewhere else and then back on current node)

This bug doesn't happen in production because dependency of callback in live state includes  `isOpen`

maxpatiiuk:
> Thought about it, but then you have another bug: User clicks on edit. Clicks on Clone (resource state gets set to clone resource). Clicks on Close button

If user did clone, then they either save or discard their changes, right?
If they save, the tree refreshes, and so we have no such issue
If they didn't save, opening Edit again, kind of makes sense that they would see the clonned resource as they didn't finish it

Not saying it's ideal, but I think it's better than duplicating logic (as duplication causes inconsistencies and bugs to prop up, like https://github.com/specify/specify7/pull/4354/files#r1441180147)

CarolineDenis:
> could you instead add a prop like `autoClose={false}` to ResourceLink?
> 
> that would require less code, and may come in handy in other places too

@maxpatiiuk, 

I made a PR with a prop autoClose. Let me know if that was what you had in mind. 
https://github.com/specify/specify7/pull/4390

CarolineDenis:
Fixed in https://github.com/specify/specify7/pull/4390

PR #4355, opened 2024-01-03
Move side bar pref to ui + rename

### LINKED ISSUE #4305: Remove redundancy in the sidebar background preference
image

The caption for this preference is:
`Light sidebar background in light mode`

There are two items to choose from:
```
Dark
Match theme color
```

This is not communicated well and is redundant. It would

### **Solution:**
The caption should be:
`Sidebar theme`

The items to choose from should be:
```
Light (in light mode)
Dark
```

This should also probably be under 'Theme' in the User Interface area, not under Appearance. This is essentially an extension of that, so we should group them together logically.

CarolineDenis:
Fixes #4305

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Header/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/index.tsx
@@ -66,7 +66,7 @@ export function Header({
   const isHorizontal = position === 'top' || position === 'bottom';
   const [isSideBarLight] = userPreferences.use('general', 'ui', 'sidebarTheme');
   const isDarkMode = useDarkMode();
-  const isMenuLight = isSideBarLight === 'matchThemeColor' && !isDarkMode;
+  const isMenuLight = isSideBarLight === 'light' && !isDarkMode;
   // Top menu is only available as collapsed
   const isCollapsed = rawIsCollapsed || isHorizontal || forceCollapse;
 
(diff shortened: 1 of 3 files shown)


CarolineDenis:
Screenshot 2024-01-03 at 1 33 37 PM

maxpatiiuk (approved):
@grantfitzsimmons can we mention in the release notes that if they had this pref set, they would have to set it again since it has been moved?

or is it important enough for us to provide a small migration for this?

emenslin (changes requested):
The sidebar theme button is not there, I checked under ui and appearance and I did not see it. 
Screenshot 2024-01-04 100709
Screenshot 2024-01-04 100721

CarolineDenis:
@emenslin, it's because it's available only for light mode

emenslin (approved):
Looks good now!
Screenshot 2024-01-04 111948

bronwyncombs (approved):
Working as expected, looks good!

Screenshot 2024-01-04 at 11 47 08 AM

grantfitzsimmons:
> can we mention in the release notes that if they had this pref set, they would have to set it again since it has been moved?

I think we'll just mention this in the release notes. I don't want to do a migration for it, and only one user that I know of has configured this since it's a very new feature.

PR #4357, opened 2024-01-04
fix(TableList): make lists of tables consistent

### LINKED ISSUE #4342: Inconsistent Behavior with Tables Hidden in Schema and New Query/Possible Query lists
**Describe the bug**

**From Zoe at CSIRO:** 

 > Steps to re-create what I've found (Sp v7.8.13 ANWC UAT)
 > 1.       From Queries module, I select New Query. As expected I’m given a list of tables to choose from to start my new query.
 > image
 > 
 > 2.       I want to refine the list and remove ‘Appraisal’, as I never build an Appraisal query. I select the ‘pencil’ edit icon next to 'New Query' heading and open the ‘Configure visible query tables’ pop up.
 > image
 > 3.      As you can see above, ‘Appraisal’ isn’t in the ‘Selected Tables’ list. It’s not in the Possible Tables list on the right either. So how do I remove it?
 > 4.       I also looked in Sp6 to see if that is where the configuration was coming from, but the configured list there is different to the list presented in Sp7, Appraisal does not appear at all. 
> image
> 5.      Appraisal also doesn’t appear when I ‘Configure Query Creators’ in Sp6. 
> image
> 
> 6. Nor does it appear in the lists if I hit the green plus + icons
> image
> 
> So I’m a bit flummoxed.

**Desktop:**
version: v7.8.13 ANWC UAT

**Additional context**

Could not be recreated on the test panel (DB: Balearicis | v7.8.13) or on production. Still investigating the cause.

maxpatiiuk:
Make list of tables in different places consistent with each other
Also refactor code to make inconsistencies in the future less likely

Fixes #4342

Test this only after #4312 is merged
Testing instructions:

- Queries:
  1. Open "Queries" dialog
  2. Click "New"
  3. See list of queries
  4. Click on the pencil icon at the top of the dialog
  5. See list of tables in the first column - should be the same list as in step 3
  6. Should be able to edit the list (move some tables from Selected to Possible or back), and see that change be reflected in the "New Query" dialog
  7. Tables you don't have read access to should be hidden in the "Selected Tables" and "Possible Tables" column
- Workbench -> Data Mapper -> Base table selection dialog
  1. Dialog should contain only 20 tables when "Show advanced tables" is not checked
  2. When clicking on "Show advanced tables", all unhidden, non system tables that you have read access to are visible
- Schema Config
  1. Should show all non sysytem, unhidden tables that you have read access to
  2. When clicking on "Show All Tables", should also show hidden and system tables

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Permissions/helpers.ts
+++ b/specifyweb/frontend/js_src/lib/components/Permissions/helpers.ts
@@ -27,7 +27,7 @@ export function hasTablePermission(
   action: typeof tableActions[number],
   collectionId = schema.domainLevelIds.collection
 ): boolean {
-  const isReadOnly = getCache('forms', 'readOnlyMode');
+  const isReadOnly = getCache('forms', 'readOnlyMode') ?? false;
   if (isReadOnly && action !== 'read') return false;
   if (
     getTablePermissions()[collectionId][tableNameToResourceName(tableName)][
--- a/specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
+++ b/specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
@@ -6,6 +6,7 @@ import { commonText } from '../../localization/common';
 import { schemaText } from '../../localization/schema';
 import { wbPlanText } from '../../localization/wbPlan';
 import type { CacheDefinitions } from '../../utils/cache/definitions';
+import type { RA } from '../../utils/types';
 import { localized } from '../../utils/types';
 import { sortFunction } from '../../utils/utils';
 import { Ul } from '../Atoms';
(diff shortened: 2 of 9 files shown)


CarolineDenis:
@maxpatiiuk, should xml-editor be merged in this before starting testing? https://github.com/specify/specify7/pull/4312 has been merged

maxpatiiuk:
thanks for reminder. done

CarolineDenis (approved):

emenslin (approved):
Looks good, everything seems to be working as expected. 

https://github.com/specify/specify7/assets/150940840/949cad31-5ce5-46d2-827b-d2b9f7745984

carlosmbe (approved):
Passes All Testing Instructions.

Compared it against 7.9, while hiding and unhiding tables to make double sure. 

Works fine. 

Nice.
giphy

PR #4373, opened 2024-01-08
Add an option to delete new formatter def with conditional formatter

### LINKED ISSUE #4368: `xml-editor`: You can't remove an additional definition without adding a field to it
**Describe the bug**
https://kufishdemo-xml-editor.test.specifysystems.org/specify/resources/app-resource/9/formatter/CollectionObject/17

When you are creating a "Conditional Formatter", you can add multiple definitions. Once you add a definition, it cannot be removed until an item is added to that definition. Removing the item now deletes both the item and the definition.

There should be a way to delete the definition before adding an item.

**To Reproduce**
1. Create a conditional format
2. Add a new definition
3. See it cannot be deleted
4. Add an item to that definition
5. Delete the item
6. See the definition is deleted

**Expected behavior**
You should be able to remove a definition without having to add an item

**Screenshots**

https://github.com/specify/specify7/assets/37256050/b83d31d6-64e1-47fc-87f8-ef3f8606df8c

**Desktop:**
macOS, Arc

CarolineDenis:
Fixes #4368 

Testing instructions: 

- open record formatters
- create a new formatter or open an existing one 
- check 'Conditional Formatter'
- select the field 
- click 'Add Definition' 
=> verify the newly created definition can be deleted without having to add a field by clicking on "Delete Definition".

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
@@ -179,7 +179,7 @@ function Definitions({
             ]}
             table={table}
           />
-          <div className="flex">
+          <div className="inline-flex">
             {index === 0 ? null : (
               <Button.Danger
                 onClick={(): void =>
@@ -189,7 +189,6 @@ function Definitions({
                 {resourcesText.deleteDefinition()}
               </Button.Danger>
             )}
-            <span className="-ml-2 flex-1" />
           </div>
         </div>
       ))}
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
@@ -178,13 +178,19 @@ function Definitions({
               (fields): void => handleChanged({ value, fields }, index),
             ]}
             table={table}
-            onDelete={
-              trimmedFields.length < 2
-                ? undefined
-                : (): void =>
-                    handleChange(removeItem(formatter.definition.fields, index))
-            }
           />
+          <div className="flex">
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -178,13 +178,19 @@ function Definitions({
...
+            {index === 0 ? null : (
+              <Button.Danger
+                onClick={(): void =>
+                  handleChange(removeItem(formatter.definition.fields, index))
+                }
+              >
+                {resourcesText.deleteDefinition()}
+              </Button.Danger>
+            )}
+            <span className="-ml-2 flex-1" />
```
maxpatiiuk, line 192: is this line necessary?

oh, is that to prevent button from being full-width?
if so, just replace `<div className="flex">` with `<div className="inline-flex">`

maxpatiiuk (approved):
👍

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
CarolineDenis: yes indeed it was for that reason

pashiav (approved):
Looks good! Definitions can be deleted without having to add an item
issue-4368

grantfitzsimmons (approved):
It works well! I just wanted to ask about this possibility (UX)

Would it be worth doing this:
Screenshot 2024-01-09 at 2 45 26 PM


Instead of this?
Screenshot 2024-01-09 at 2 45 35 PM

This might be irrelevant depending on what @CarolineDenis does to resolve the other UI request I've made about the table format editor, especially with conditional formats

PR #4376, opened 2024-01-09
Reset save blockers when field's parse result is valid

### LINKED ISSUE #4361: Cannot create new taxon
**Describe the bug**
When adding a new taxon within the determinations table it autofill's the rank but then says it cannot save because the  "field is required"

**To Reproduce**
1. Go to data entry and create a new collection object 
2. Add a determination and add new taxon
3. Add a parent 

**Expected behavior**
This should be able to save without any error.

https://github.com/specify/specify7/assets/150940840/203030a9-9d21-4295-8e7f-9580b36cc434

**Desktop:**
 - OS: Windows 11
 - Browser: Chrome
 - Specify 7 Version: xml editor 

**Database Name:** lsumz mammals

melton-jason:
Fixes #4361

Looks like this functionality was removed as a part of https://github.com/specify/specify7/commit/6bdeba0106e69cbd41c2018cfaec1a7a839e2297 and never reintroduced. 

### Checklist

- [ ] Add automated tests

### Testing instructions
Alongside recreating the issue from https://github.com/specify/specify7/issues/4361#issue-2066207840, general testing on forms would also be appreciated. Specifically test cases where a Save Blocker is triggered (such as a missing required field, a uniqueness rule validation, etc.)  and then make the input in the field valid again and ensure the save blocker is removed.

### DIFF
--- a/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
+++ b/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
@@ -132,7 +132,11 @@ export function useResourceValue<
       );
       if (field === undefined) return;
 
-      if (!parseResults.isValid) setBlockers([parseResults.reason]);
+      if (parseResults.isValid) {
+        setBlockers([]);
+      } else {
+        setBlockers([parseResults.reason]);
+      }
       ignoreChangeRef.current = true;
       /*
        * If value changed as a result of being formatted, don't trigger
--- a/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
+++ b/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
@@ -132,11 +132,10 @@ export function useResourceValue<
       );
       if (field === undefined) return;
 
-      if (parseResults.isValid) {
-        setBlockers([]);
-      } else {
-        setBlockers([parseResults.reason]);
-      }
+      // This assumes that there are no field blockers set by anything else
+      if (parseResults.isValid) setBlockers([]);
+      else setBlockers([parseResults.reason]);
+
       ignoreChangeRef.current = true;
       /*
        * If value changed as a result of being formatted, don't trigger


CarolineDenis (approved):

carlosmbe (commented):
The issue looks fixed. Going to do some general form testing on this branch and then approve.

nice-alishaan

carlosmbe (approved):

specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
```diff
@@ -132,7 +132,11 @@ export function useResourceValue<
       );
       if (field === undefined) return;
 
-      if (!parseResults.isValid) setBlockers([parseResults.reason]);
+      if (parseResults.isValid) {
+        setBlockers([]);
+      } else {
+        setBlockers([parseResults.reason]);
+      }
```
maxpatiiuk, line 139: The way it was in https://github.com/specify/specify7/commit/6bdeba0106e69cbd41c2018cfaec1a7a839e2297#diff-0d189e32be0ec55985090bda95c99a9ae7d87706461a2b9667242015a904f916R134-R135:

```suggestion
      // This assumes that there are no field blockers set by anything else
      if (parseResults.isValid) setBlockers([]);
      else setBlockers([parseResults.reason]);
```

The comment about "This assumes that there are no field blockers set by anything else" is also a good point.
Do not know if it causes issues anywhere yet, but it might at some point.
Before the rewrite to TypeScript in 2021, each save blocker had a unique key, thus making such overwriting less likely. I removed the keys during TypeScript rewrite to make things simpler, but we might want to add them back at some point

maxpatiiuk (approved):

maxpatiiuk:
Hmm, the commit you linked (https://github.com/specify/specify7/commit/6bdeba0106e69cbd41c2018cfaec1a7a839e2297#diff-0d189e32be0ec55985090bda95c99a9ae7d87706461a2b9667242015a904f916R134-R135) actually added those lines.

They were removed in https://github.com/specify/specify7/commit/4b0cc359f3ff4ae67eb7453b113c19daeba5e13f#diff-0d189e32be0ec55985090bda95c99a9ae7d87706461a2b9667242015a904f916L137-L139

emenslin (approved):
Looks good, no longer running into the original issue or any issues with save blocker.

PR #4381, opened 2024-01-09
Add the possibibility to collapse conditional formatters

### LINKED ISSUE #4370: `xml-editor`: Improve Table Format Editor interface
The 'Table Format Editor' displays the same information in both applications, just in different ways. I believe the Specify 6 implementation has a more intuitive user interface and experience when configuring these.

It is super easy to use and allows you to quickly switch between a table view for easily viewing what format will be used for each field value and a single table format which just lets you pick from a list of fields.

https://github.com/specify/specify7/assets/37256050/8cb519e6-958d-49eb-860d-0c55bb503771

Conditional or "switch" table formats are not exceedingly common but are used in cases like agents and reference work. The interface for these can be tricky, but I believe Specify. 6 does a good job.

### Specify 6

It is very intuitive. To start, it is clear for the user that they need to pick a type between two options. a "Single table format" or "Display depends on value of field:" and an easy way to pick which field that is. 

image

I can simply see that there are several possible values for that field, and the format is clear to see based on those values:

| Field Value | Table Format |
|----------|----------|
| 0   | `lastName`  |
| 1   | `lastName` `middleInitial` `lastName`  |
| 2   | `lastName`  |
| 3   | `lastName`  |

### Specify 7

There is a very long list of every possibility separated by a full-width line containing the field value that results in the configured format. 

Besides the Christmas-themed formatting, it is clear it could use the grid view that Specify 6 has above. Imagine that, and just like in Specify 6, you could click on a line and open another dialog with the fields and separators contained within.

https://github.com/specify/specify7/assets/37256050/84a5b0e7-4c97-4cb6-b3ae-41b0de87d3bd

image

### Specify 7
Screenshot 2024-01-08 at 10 41 41 AM

### Specify 6
image

## TL;DR

Specify 6 has a great table format editor that we should take inspiration from and match more closely in the future.

CarolineDenis:
Fixes #4370 

### Checklist

### Testing instructions

- Open the formatter UI
- Verify that checking conditional formatter field box displays a new UI for conditionals
- Verify that unchecking conditional formatter displays only the separator + field component 
- test adding new field 
- test deleting field 
- test adding a conditional value
- test deleting conditional with icon button
- test deleting conditional with Delete Definition button
- test collapsible function when conditional checked


CarolineDenis:
https://github.com/specify/specify7/assets/108160931/9830b6c4-0a59-4f3f-9e60-4576774bce7b


https://github.com/specify/specify7/assets/108160931/1346c39a-7968-4065-b369-584ba341c1e9

CarolineDenis:
Screenshot 2024-01-10 at 10 50 54 AM
Screenshot 2024-01-10 at 10 51 02 AM
Screenshot 2024-01-10 at 10 51 13 AM
Screenshot 2024-01-10 at 10 51 22 AM

CarolineDenis:
https://github.com/specify/specify7/assets/108160931/5d863da0-908d-485c-b342-93025e702dee

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -145,51 +147,142 @@ function Definitions({
...
+                        value: value.length === 0 ? undefined : value,
+                        fields,
+                      },
+                      index
+                    )
+                  }
+                />
+                {trimmedFields.length > 0 && showConditionalField[index] ? (
+                  <Button.Danger
+                    className="w-[40%]"
```
maxpatiiuk, line 187: the red button seems too large

Screenshot 2024-01-10 at 19 16 49

can we remove className?

```suggestion
```

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -145,51 +147,142 @@ function Definitions({
...
 
+  const [showConditionalField, setShowConditionalField] = React.useState(
+    trimmedFields.map(() => false)
+  );
+
   return table === undefined ? null : (
     <div className="flex flex-col gap-4 divide-y divide-gray-500 [&>*]:pt-4">
       {trimmedFields.map(({ value, fields }, index) => (
-        <div className="flex flex-col gap-2" key={index}>
+        <div
```
maxpatiiuk, line 157: turn each row into a component, and then have `isCollapsed` state inside of it instead of `showConditionalField[index]`

a general idea - instead of looping over the logic for each child in the parent, the parent should just render the children, and each child handles it's own state and logic as much as possible - this simplifies the code usually and avoids bugs.

for example, you currently have a bug:
- if a row is deleted, you forget to also delete from showConditionalField
- also, if formatter.definition where to be modified by some other component, your showConditionalField would become stale

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -145,51 +147,142 @@ function Definitions({
...
-              />
-              <span>
-                {index === 0
-                  ? resourcesText.elseConditionDescription()
-                  : resourcesText.conditionDescription()}
-              </span>
+              {showConditionalField[index]
+                ? resourcesText.conditionFieldValue()
+                : null}
+              <div className="flex items-center gap-2">
```
maxpatiiuk, line 170: is `items-center` needed here??

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -145,51 +147,142 @@ function Definitions({
...
-                : (): void =>
-                    handleChange(removeItem(formatter.definition.fields, index))
-            }
-          />
+          {showConditionalField[index] ||
+          !hasCondition ? null : fields.length === 0 ? (
+            <Button.Small
+              onClick={(): void => {
+                const newConditionalField = Array.from(showConditionalField);
+                newConditionalField[index] = !newConditionalField[index];
```
maxpatiiuk, line 212: fyi, could have used `replaceItem` util here

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -145,51 +147,142 @@ function Definitions({
...
+                    ],
+                  },
+                  index
+                );
+              }}
+            >
+              {resourcesText.addField()}
+            </Button.Small>
+          ) : (
+            <Label.Inline>
```
maxpatiiuk, line 234: what is this Label for?
label should have an associated input field

maxpatiiuk (changes requested):
Love how much more compact it is in the UI, and the fact that it does not require too much code to implement it

still, the code could use some improvements

maxpatiiuk:
also, remove "!!! work in progress not ready for review" from your PR description

CarolineDenis:
> also, remove "!!! work in progress not ready for review" from your PR description

Yes that will be removed and replaced by testing instructions once it is ready for review :)

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
CarolineDenis: If we remove the class name it looks like this:
Screenshot 2024-01-11 at 10 00 42 AM

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
CarolineDenis: yes 
Screenshot 2024-01-11 at 10 03 21 AM
VS
Screenshot 2024-01-11 at 10 03 30 AM

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
CarolineDenis: when using replaceItem I have a typescript error `Argument of type 'RA<boolean>' is not assignable to parameter of type 'SetStateAction<boolean[]>'.
  The type 'readonly boolean[]' is 'readonly' and cannot be assigned to the mutable type 'boolean[]'.`

CarolineDenis: Same thing for the `onRemoveField() `

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
CarolineDenis: I gave it a try let me know how it is

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -145,51 +149,37 @@ function Definitions({
...
+          key={index}
+          showConditionalField={[showConditionalField, setShowConditionalField]}
+          table={table}
+          trimmedFieldsLength={trimmedFields.length}
+          value={value}
+          onChange={handleChange}
+          onChanged={handleChanged}
+          onRemoveField={(removedIndex: number): void => {
+            const updatedShowConditionalField =
+              Array.from(showConditionalField);
```
CarolineDenis, line 174: I would like to use removeItem() but am facing typescript errors. 
`Argument of type 'RA<boolean>' is not assignable to parameter of type 'SetStateAction<WritableArray<boolean>>'.
  The type 'readonly boolean[]' is 'readonly' and cannot be assigned to the mutable type 'WritableArray<boolean>'.`

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -204,13 +194,164 @@ function Definitions({
...
             {resourcesText.addDefinition()}
-          </Button.Success>
+          </Button.Info>
         </div>
       ) : null}
     </div>
   );
 }
 
+function ConditionalFormatter({
```
CarolineDenis, line 204: @maxpatiiuk, 
Should I create a new file with the smaller component to reduce the length of this one?

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
maxpatiiuk: I see. Can you please add one of these class names instead of hardcoding the size:
- https://tailwindcss.com/docs/text-wrap
- https://tailwindcss.com/docs/word-break
- https://tailwindcss.com/docs/whitespace

alternatively, this might display correctly if you wrap either the Input or the Button in a `<div>`

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
maxpatiiuk: why are you using WritableArray here?
https://github.com/specify/specify7/pull/4381/files/b4b09659c6fb2b5af4d3718fa53dfe7160c0b04e#diff-0ebb7583f8406c8fc8b3d62b05833856ef677a547d0e626fcffd0691db2f833aR153

that is what is causing the issue

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
maxpatiiuk: see https://github.com/specify/specify7/pull/4381#discussion_r1449555484

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
maxpatiiuk: yes please 👍

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
maxpatiiuk: good job creating ConditionalFormatter

now, can you also remove showConditionalField?
and replace that with a simple boolean state that would be inside of ConditionalFormatter?
and that way, you won't have to manage showConditionalField at all - because if ConditionalFormatter is removed or added, its state would be removed or created automatically
see how that would be simpler, and less bug-prone?

or did you encounter problems implementing this?

maxpatiiuk (changes requested):

CarolineDenis:
@maxpatiiuk , 
Is it better to call the context in a component or better to pass isReadOnly as a prop from parent to children when already defined in the parent?

specifyweb/frontend/js_src/lib/components/Formatters/Formatter.tsx
```diff
@@ -224,24 +213,26 @@ function ConditionalFormatter({
...
   ) => void;
-  readonly showConditionalField: GetOrSet<WritableArray<boolean>>;
   readonly trimmedFieldsLength: number;
   readonly formatter: GetSet<Formatter>[0];
   readonly onChange: (fields: Formatter['definition']['fields']) => void;
   readonly table: SpecifyTable;
-  readonly onRemoveField: (index: number) => void;
 }): JSX.Element {
   const isReadOnly = React.useContext(ReadOnlyContext);
+  const [collapsedConditionalField, _, __, toggleConditionalField] =
```
maxpatiiuk: since you are inside a component called `ConditionalFormatter` there is no need for a name as verbose as `collapsedConditionalField`

`isCollapsed` would work great here!

a verbose name like `collapsedConditionalField` is needed if it was global variable that you export and could import from any other file - in that case the verbosity is required to give context of what the variable is about, and to reduce chance of collision with other exported variable names

maxpatiiuk (approved):

pashiav (changes requested):
When clicking on the arrow icon for expanding a Conditional Field, the icon moves to the bottom right of the field. Typically the expand/collapse icon is always fixed, so this may cause users a bit of confusion as to where the icon went or if they are allowed to collapse the field again after expanding it.




https://github.com/specify/specify7/assets/118556884/df6a7b35-5bc0-4203-bc48-f1cfc6429266

On hover for collapsing a field, the message "showConditionalField" is displayed. This could be misleading since the actual function is to hide the field. It might confuse users about the intended action.

4370-hover

Overall functionality looks good!

CarolineDenis:
@pashiav, 
Yes I am aware of this button moving to the bottom. The thing is, if we decide to keep it where it is once uncollapsed we would need to add a lot of complexity to the code. I don't know if it's worth it. 
@grantfitzsimmons any opinion?

grantfitzsimmons:
I'd like to review this PR as soon as I can before further comment, but I'm inclined to agree with Pashia's thoughts, even with the additional complexity.

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -0,0 +1,241 @@
...
+            }}
+          />
+        )}
+      </div>
+      <div className="flex">
+        {isExpanded ? <span className="-ml-2 flex-1" /> : null}
+        {hasCondition ? (
+          <Button.Icon
+            icon={isExpanded ? 'chevronUp' : 'chevronDown'}
+            title="showConditionalField"
```
maxpatiiuk, line 232: good catch @pashiav!

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -0,0 +1,241 @@
...
+  const [isExpanded, _, __, handleToggle] = useBooleanState(false);
+  return (
+    <div
+      className={`flex ${
+        isExpanded || !hasCondition ? 'flex-col' : 'items-center'
+      } gap-2`}
+      key={index}
+    >
+      {hasCondition && (
+        <Label.Block>
```
maxpatiiuk, line 136: The button is not supposed to be inside the label - labe should only contain the input field

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -0,0 +1,241 @@
...
+      </div>
+      <div className="flex">
+        {isExpanded ? <span className="-ml-2 flex-1" /> : null}
+        {hasCondition ? (
+          <Button.Icon
+            icon={isExpanded ? 'chevronUp' : 'chevronDown'}
+            title="showConditionalField"
+            onClick={(): void => {
+              handleToggle();
+            }}
```
maxpatiiuk, line 235:
```suggestion
            onClick={handleToggle}
```

maxpatiiuk (changes requested):

maxpatiiuk:
> Yes I am aware of this button moving to the bottom. The thing is, if we decide to keep it where it is once uncollapsed we would need to add a lot of complexity to the code. I don't know if it's worth it.

@CarolineDenis would it decrease complexity if you would keep the Delete icon as an icon rather than a button both when expanded and collapsed?

> would need to add a lot of complexity to the code

give it a try, and check if you can refactor a code to make it less complex - then I can review and check if there are futhre ways to simplify

CarolineDenis:
Screenshot 2024-01-25 at 7 25 17 AM
Screenshot 2024-01-25 at 7 25 29 AM

pashiav (approved):
Looks good! Though the icons are not fixed in one position on-click, users can still make sense of it. The decision to place the delete and expand/collapse icons right next to each other maintains cohesion in the ui, so I don't think there would be much confusion.

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -0,0 +1,246 @@
...
+        </Button.Small>
+      ) : (
+        fields.map((field, index) => (
+          <p key={index}>
+            {field.separator === undefined ? '' : field.separator}
+            {field.field === undefined ? '' : field.field[0].label}
+          </p>
+        ))
+      )}
+      {isExpanded || !hasCondition ? (
```
maxpatiiuk, line 213: you are using `isExpanded || !hasCondition` and `!isExpanded && hasCondition` a lot of times in this component - can you refactor by creating a variable for it and then using that variable and !variable?

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -0,0 +1,246 @@
...
+          table={table}
+        />
+      ) : null}
+      <span className="-ml-2 flex-1" />
+      <div className="inline-flex">
+        {trimmedFieldsLength === 1 || isExpanded ? null : (
+          <Button.Icon
+            icon="trash"
+            title={resourcesText.deleteDefinition()}
+            onClick={(): void => {
```
maxpatiiuk, line 228: create a `const handleDelete = ():void=>handleChange(removeItem(formatter.definition.fields, index));` function, and then use that as an onClick callback for the two delete buttons

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -0,0 +1,246 @@
...
+              handleChange(removeItem(formatter.definition.fields, index));
+            }}
+          />
+        )}
+      </div>
+      <div className="flex">
+        {isExpanded ? <span className="-ml-2 flex-1" /> : null}
+        {hasCondition && !isExpanded ? (
```
maxpatiiuk, line 236: this seems like a mistake - the span is visible if isExpanded, but button is visible if !isExpanded
can the span be removed as it seems useless here

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
maxpatiiuk: this doesn't look resolved - can you use a localized title, not a camelCased string?

maxpatiiuk (changes requested):

emenslin (approved):
Looks good, I like how it's more compact and intuitive now. 
Screenshot 2024-01-26 134442

grantfitzsimmons (changes requested):
1. It appears that the conditional field value entry field is not as tall as the button next to it:
image
image

2. The preview is misleading, as it adds spacing when it shouldn't, and this makes the user think there will be spaces included by default.

image

The `gap-2` class adds a small gap between each of the `<p>` elements. 

image

So instead of appearing like this (**mock up**):
image

It appears like this (current behavior):
image

https://herbrbge252024-issue4370.test.specifysystems.org/specify/resources/app-resource/24/formatter/Agent/4

grantfitzsimmons (changes requested):
image

image

Because of the way these are `<p>` elements instead of a continuous string, it wraps inappropriately when the dialog is narrow. Is there an easy way to resolve this or is this as tricky as I think?

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -197,12 +197,14 @@ function ConditionalFormatter({
...
       ) : (
-        fields.map((field, index) => (
-          <p key={index}>
-            {field.separator === undefined ? '' : field.separator}
-            {field.field === undefined ? '' : field.field[0].label}
-          </p>
-        ))
+        <div className="flex flex-wrap p-2">
+          {fields.map((field, index) => (
+            <p key={index}>
```
maxpatiiuk, line 202: why is this P even needed? can you use React.Fragment instead?
this p is causing too much trouble 🤣

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -124,21 +114,29 @@ function ConditionalFormatter({
...
       {hasCondition && (
         <Label.Block>
           {isExpanded ? resourcesText.conditionFieldValue() : null}
-          <div className="flex items-center gap-2">
-            <div className="flex-1">
+          <div className="flex items-center gap-2 pr-2">
+            <div className="h-full flex-1">
               <Input.Text
-                className="h-full"
+                className="h-full pr-2"
```
maxpatiiuk, line 139: why does this have pr-2?
this seems like it would be better accomplished by gap

i.e you can just increase `gap-2` if you need to
or, you can remove the "-ml-2" from `<span className="-ml-2" />` (remember that the "-ml-2" is there to "eat the gap", though normally we have it as `<span className="flex-1 -ml-2" />`)

specifyweb/frontend/js_src/lib/localization/resources.ts
```diff
@@ -448,4 +444,10 @@ export const resourcesText = createDictionary({
   customizeFieldFormatters: {
     'en-us': 'Customize Field Formatters',
   },
+  showConditionalField: {
+    'en-us': 'Show Conditional Field',
+  },
+  hideConditionalField: {
+    'en-us': 'Hide Conditional Field',
+  },
```
maxpatiiuk, line 452: collapse/expand would be better terms than hide/show

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
maxpatiiuk: and these are just strings anyway, right?
so you could just use do `.map().join()`, so no need for React.Fragment at all

maxpatiiuk (changes requested):

maxpatiiuk:
> an ideal solution would make just the field names scrollable while the separator and trash/expand icons remain in the same position

definitely agree that the current looks is sub-optimal
but I think the optimal solution is not a scroll bar (scroll bars should be avoided when they are not necessary), but just let it wrap to 2nd line or however many lines it needs (which would happily do once it's rendered as a single element, rather than a bunch of separate PARAGRAPHS - which is totally wrong from the a11y perspective, but okay)

grantfitzsimmons (changes requested):
https://github.com/specify/specify7/assets/37256050/3ea9c089-d732-4e76-ac80-61aa1834a514


I made the separator for everything (except the first term) be `    ` (4 spaces) to verify something I had noticed.

Space separators are not shown at all in the preview, so it looks like this:

` First NameMiddle NameLast NameTitleSuffixAbbreviationDate Of BirthDate Of Death`
when it should look like this:
`First Name    Middle Name    Last Name    Title    Suffix    Abbreviation    Date Of Birth    Date Of Death`

It displays correctly in the preview below.

grantfitzsimmons:
> but I think the optimal solution is not a scroll bar (scroll bars should be avoided when they are not necessary), but just let it wrap to 2nd line or however many lines it needs (which would happily do once it's rendered as a single element, rather than a bunch of separate PARAGRAPHS - which also is wrong from the a11y perspective)

Funny you say that– I changed my mind about this in a private conversation with @CarolineDenis. I agree! I hadn't thought through the UX enough.

grantfitzsimmons:
> but I think the optimal solution is not a scroll bar (scroll bars should be avoided when they are not necessary), but just let it wrap to 2nd line or however many lines it needs (which would happily do once it's rendered as a single element, rather than a bunch of separate PARAGRAPHS - which also is wrong from the a11y perspective)

Funny you say that– I changed my mind about this in a private conversation with @CarolineDenis. I agree! I hadn't thought through the UX enough.

maxpatiiuk:
good reminder - the container should have `whitespace-pre`, to tell browser to faithfully render everything as is
https://tailwindcss.com/docs/whitespace

CarolineDenis:
NOTES: 
- no scroll on field but go to second line when needed. 
- field spaces are represented as is 
- no p tags 
- no pr-2

Separator no space: 
Screenshot 2024-02-14 at 7 17 09 AM

Separator with space: 
Screenshot 2024-02-14 at 7 17 24 AM

No separator no space:
Screenshot 2024-02-14 at 7 17 51 AM

No separator with space: 
Screenshot 2024-02-14 at 7 18 03 AM

grantfitzsimmons (approved):
👍 🎸 !!!

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -136,7 +136,7 @@ function ConditionalFormatter({
           <div className="flex items-center gap-2 pr-2">
             <div className="h-full flex-1">
               <Input.Text
-                className="h-full pr-2"
```
maxpatiiuk: (optional) just a FYI that adding `fiex` to the parent of this Input would also fix the issue with inconsistent height

maxpatiiuk (approved):

PR #4390, opened 2024-01-11
Prevent cloning dialog to close itslef

### LINKED ISSUE #4347: `xml-editor`: Cloned resource dialog disappears in tree viewer

CarolineDenis:
Fixes #4347 

### Checklist

- [ ] Add automated tests

### Testing instructions

- open a tree 
- click on a node 
- click on edit button 
- click on clone 
- verify dialog doesn't close itself + verify cloning function works

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
@@ -25,7 +25,7 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
   props: rawProps,
   resourceView,
   children,
-  autoClose,
+  autoClose = false,
 }: {
   readonly resource: SpecifyResource<AnySchema> | undefined;
   readonly component: COMPONENT;
--- a/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
@@ -25,6 +25,7 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
   props: rawProps,
   resourceView,
   children,
+  autoClose,
 }: {
   readonly resource: SpecifyResource<AnySchema> | undefined;
   readonly component: COMPONENT;
@@ -39,9 +40,12 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
     >,
     'dialog' | 'onAdd' | 'onClose' | 'onSaved'
   >;
+  readonly autoClose: boolean;
 }): JSX.Element {
   const [isOpen, _, handleClose, handleToggle] = useBooleanState();
-  React.useEffect(handleClose, [resource]);
+  React.useEffect(() => {
+    if (autoClose) handleClose();
+  }, [resource]);
 
   function handleClosed(): void {
     handleClose();
--- a/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
@@ -40,11 +40,11 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
     >,
     'dialog' | 'onAdd' | 'onClose' | 'onSaved'
   >;
-  readonly autoClose: boolean;
+  readonly autoClose?: boolean;
 }): JSX.Element {
   const [isOpen, _, handleClose, handleToggle] = useBooleanState();
   React.useEffect(() => {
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
```diff
@@ -25,6 +25,7 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
   props: rawProps,
   resourceView,
   children,
+  autoClose,
```
maxpatiiuk, line 28:
```suggestion
  autoClose = false,
```

whenever possible, try to set the default value when prop is initialized, rather than when used.
this applies both to react components, and regular functions

benefits:
- some IDEs provide better integration and autocomplete when you do this (i.e when calling the function, they will show what the default value for an argument is)
  - similarly, someone looking at a component, will easily know what a default value for a prop is by looking at the component header (without having to search for it's usages in the code)
- variable is declared only once, but may be used multiple times - by setting default value, you avoid the need for specifying the default value/fallback place in every place where the variable is used (thus, saving code, making it easier to change the default value, and reducing possibly bugs when default values mismatch)

drawbacks:
- if default value is an object or other mutable value, declaring default value may recreate default value multiple times - in that case, you can define the default value outside the component, and use it as a variable (only applies to react components, not functions). see example solution - https://github.com/specify/specify7/pull/4357/files#diff-c703adeb4352c7832286317011ba686112a5f56f7c0ca20f3a493fc2561197d4R99-R107

specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
```diff
@@ -39,9 +40,12 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
     >,
     'dialog' | 'onAdd' | 'onClose' | 'onSaved'
   >;
+  readonly autoClose?: boolean;
 }): JSX.Element {
   const [isOpen, _, handleClose, handleToggle] = useBooleanState();
-  React.useEffect(handleClose, [resource]);
+  React.useEffect(() => {
+    if (autoClose === undefined || autoClose) handleClose();
```
maxpatiiuk, line 47: fyi: `autoClose === undefined || autoClose` is good, and `autoClose !== false` would be equivalent
both are fine
(and both won't be needed once you set the default value)

specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
```diff
@@ -39,9 +40,12 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
...
+  }, [resource]);
```
maxpatiiuk, line 48: your version is fine, but just showing that this variant is equivalent:
```suggestion
  React.useEffect(autoClose ? handleClose : f.void, [resource]);
```

or, if you want to placate the eslint rule that complains about the dependency array, could do this:

```suggestion
  React.useEffect(() => {
    // Close the dialog when resource changes
    void resource;
    if (autoClose === undefined || autoClose) handleClose();
  }, [handleClose, resource]);
```

(i.e we are needlessly referring the the `resource`, thus ESLint stops complaining about it being needlessly specified in the dependency array)

maxpatiiuk (changes requested):
please change to use default value

other comments are just for information - don't need to implement if you don't want to

emenslin (approved):
Looks good, dialog doesn't close and cloning works.

https://github.com/specify/specify7/assets/150940840/7e4fb651-a15d-43cf-92ad-efc957917924

pashiav (approved):
Looks good!

https://github.com/specify/specify7/assets/118556884/ccec1214-97ae-47da-a23c-48fbb6370a29

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/Molecules/ResourceLink.tsx
```diff
@@ -25,6 +25,7 @@ export function ResourceLink<COMPONENT extends typeof Link['Icon']>({
   props: rawProps,
   resourceView,
   children,
+  autoClose = false,
```
realVinayak, line 28: Should this be `autoClose=true` since the previous default behavior was dialog closing on resource changing? Now, the default behavior is dialog not closing on resource changing.

maxpatiiuk: oh, good catch!
@CarolineDenis could you open a small fixup pr for this?

Issue #4392, opened 2024-01-12
Report dialog shows 'Created By Agent' as heading while displaying Specify Users


grantfitzsimmons:
**Describe the bug**
While the heading says 'Created By Agent' in the Reports dialog, a list of users are shown instead.
Despite the `CreatedByAgentID` field existing in the `spreport` table, the more useful information is the user associated with it. 

**Expected behavior**
I recommend that we change the heading to say "Specify User" instead of "Created By Agent".
I also believe that the text should be the same color as the "Timestamp Created" text, as it is a shade lighter.

**Screenshots**
<img width="404" alt="Screenshot 2024-01-12 at 9 05 50 AM" src="https://github.com/specify/specify7/assets/37256050/bc046019-8ec0-482e-89fb-7dd535f4ee9a">

macOS, Arc, `xml-editor`

PR #4396, opened 2024-01-12
Improve Reports dialog by fixing heading text color and making edit arrows  orange

### LINKED ISSUE #4395: The edit arrows are black in the report dialog rather than orange
image

image

grantfitzsimmons:
Fixes #4395 and #4392

Screenshot 2024-01-12 at 12 38 48 PM

### Testing instructions

1. Open the 'Reports' dialog
2. Verify that the ✏️ icon is now orange instead of black
3. Verify that 'Created By Agent' now says 'Specify User'


grantfitzsimmons:
Don't be fooled by the passing tests, I was trying to figure out what the nonsensical sorting behavior is for 'Specify User'. I created a number of fake reports and assigned them to every user in the database. I intend to sort this by the username of each user, but when trying without further changes, the sorting seems non-sensical. Need to investigate this more before I do anything else.

**Descending Sort:**
Screenshot 2024-01-12 at 2 44 01 PM

**Ascending Sort:**
Screenshot 2024-01-12 at 2 44 30 PM

At least it looks better! Need to work on other things but plan to revisit this. In the meantime, if someone else wants to resolve the sorting behavior, please feel free.

specifyweb/frontend/js_src/lib/components/Reports/index.tsx
```diff
@@ -216,7 +217,17 @@ function ReportRow({
               />
             </Button.LikeLink>
           </th>
-          <th>{getField(tables.SpReport, 'createdByAgent').label}</th>
+          <th>
+          <Button.LikeLink
+              onClick={(): void => handleSort('specifyUser')}
+            >
```
maxpatiiuk, line 223:
```suggestion
          <Button.LikeLink onClick={(): void => handleSort('specifyUser')}>
```

specifyweb/frontend/js_src/lib/components/Reports/index.tsx
```diff
@@ -216,7 +217,17 @@ function ReportRow({
...
+            {getField(tables.SpReport, 'specifyUser').label}
+            <SortIndicator
+                fieldName="specifyUser"
+                sortConfig={sortConfig}
+              />
```
maxpatiiuk, line 228:
```suggestion
              {getField(tables.SpReport, 'specifyUser').label}
              <SortIndicator
                fieldName="specifyUser"
                sortConfig={sortConfig}
              />
```

maxpatiiuk (approved):
some formatting issues - not sure why prettier didn't resolve them
code looks good though

melton-jason (changes requested):
The nonsensical sorting is caused because Specify is sorting the users on their resource uri's, not their names. 

This is ultimately caused because of the following line: 

https://github.com/specify/specify7/blob/6872b0247f0241b3491dc626447491aec85971a5/specifyweb/frontend/js_src/lib/components/Reports/index.tsx#L191

When the sorting button is clicked and `handleSort('specifyUser')` is called, is changes the `sortConfig.sortField` to be `specifyUser`. 

In this line, we are getting the value of `specifyUser` on a `SerializedResource<SpAppResource>`, which you can directly see the structure of by either setting a breakpoint on the line, or by seeing what is returned from the API when it serialized and sends a SpAppResource object (the following link returns an array of spappresource objects: `http://localhost/api/specify/spappresource/`, assuming you have a local instance of Specify running). 

With either approach, the value of `specifyUser` is the url to the resource. For example:
```json
{
...
"specifyuser": "/api/specify/specifyuser/1/",
...
}
```
This is passed to `applySortConfig` in the definition of `useSortConfig` and actually sorts the array at the following line: 
https://github.com/specify/specify7/blob/6872b0247f0241b3491dc626447491aec85971a5/specifyweb/frontend/js_src/lib/components/Molecules/Sorting.tsx#L84

The names are properly displayed in the UI in the first place because they are being formatted by a `FormattedResourceUrl` component: 

https://github.com/specify/specify7/blob/6872b0247f0241b3491dc626447491aec85971a5/specifyweb/frontend/js_src/lib/components/Reports/index.tsx#L248-L252

This eventually fetches the resource and formats it as Specify would normally format the resource:
https://github.com/specify/specify7/blob/6872b0247f0241b3491dc626447491aec85971a5/specifyweb/frontend/js_src/lib/components/Formatters/formatters.ts#L64-L119

The solution here would be to not sort the `ReportEntry` objects on the value of the appResource's specifyUser when the sortField is set to specifyUser, and instead sort the values by the SpecifyUsers' names.

maxpatiiuk:
ah, good catch @melton-jason!
now I remembered why weren't yet letting users sort by a relationship in any other dialog - not supported yet

and given that users are fetched async, we can't easily allow sorting by that yet
(similarly, how you can't sort by record set length in the record sets dialog)

grantfitzsimmons:
After further discussion, we have concluded that it is not worth allocating development resources at this time to enable sorting by specify user. This PR now makes it visually consistent with other dialogs.

specifyweb/frontend/js_src/lib/components/Reports/index.tsx
```diff
@@ -216,7 +217,9 @@ function ReportRow({
               />
             </Button.LikeLink>
           </th>
-          <th>{getField(tables.SpReport, 'createdByAgent').label}</th>
+          <th className="link">
+            {getField(tables.SpReport, 'specifyUser').label}
+          </th>
```
melton-jason, line 222: Can the `className=link` be removed from this Table Header element? This makes the element appear interactive when hovered. 

```suggestion
          <th>{getField(tables.SpReport, 'specifyUser').label}</th>
```

**With className `link`**: 
image

**Without className `link`**: 
image

Or is this behavior intentional?

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/Reports/index.tsx
grantfitzsimmons: Intentional. Notice the heading is gray instead of white (or gray instead of black, in light mode). Rather than change styles globally I made this choice.

grantfitzsimmons: This heading text color inconsistency appears in the 'Size' heading for record sets as well.

maxpatiiuk: > Intentional. Notice the heading is gray instead of white (or gray instead of black, in light mode). Rather than change styles globally I made this choice.

sorry, I don't follow - can you explain?

if the heading is not clickable, then it makes sense that it should appear a bit more grayed out and not have a pointer mouse. or would in this case visual consistency matter more than UX?

also, if you want to still have `link` in there, could you please replace `className="link"` with `className={className.link}` - it won't make a visual difference, but it would make it much easier to find all usages of the "link" class name (for example when testing the impact of a change on all usages of a class name, this would be useful; similarly, when renaming a class name)

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/Reports/index.tsx
grantfitzsimmons: In this case, visual consistency is more important in my opinion. So in the spirit of compromise, I've added `pointer-events-none` to this heading as well so it doesn't appear to be clickable either!

emenslin (approved):
Feels a lot more cohesive now, looks good!
Screenshot 2024-01-30 155721

specifyweb/frontend/js_src/lib/components/Reports/index.tsx
```diff
@@ -216,7 +217,9 @@ function ReportRow({
               />
             </Button.LikeLink>
           </th>
-          <th>{getField(tables.SpReport, 'createdByAgent').label}</th>
+          <th className="link pointer-events-none">
```
maxpatiiuk, line 220: pointer-events-none sounds good.

as per https://github.com/specify/specify7/pull/4396#discussion_r1470463857, could you also do this refactor?

```suggestion
          <th className={`${className.link} pointer-events-none`}>
```

maxpatiiuk (approved):

melton-jason (approved):
Looks good! 
See @maxpatiiuk's comment in https://github.com/specify/specify7/pull/4396#discussion_r1472306768

PR #4397, opened 2024-01-12
Display resource mapping inline when no customize field formatter

### LINKED ISSUE #4380: `xml-editor`: Table Format field mappings stack inappropriately and are not separated
**Describe the bug**
Nested items in the table format editor appear in a large stack rather than taking up the available width in the dialog, resulting in confusing appearance.

Separators are also not visible between items in the table format, making it appear like they run into each other.

https://kufishdemo-xml-editor.test.specifysystems.org/specify/resources/app-resource/9/formatter/Loan/39

**To Reproduce**
1. Create or edit a table format
2. See that any nested fields are stacked
3. See visual/interface issues

Screenshot 2024-01-09 at 3 17 14 PM

**Expected behavior**
Separators should be placed between fields and mapped fields should take up the full width available. 

**Screenshots**

image

image

https://github.com/specify/specify7/assets/37256050/10bc984b-fe3b-4096-a5b7-506ae2a62563

CarolineDenis:
Fixes #4380 

### Checklist

### Testing instructions

- open app resource 
- open formatters
- choose a table 
- add nested fields
- verify they are next to each other and not stacked above each other when "Customize Field Formatters" in not checked.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
@@ -149,13 +149,11 @@ export function ResourceMapping({
   mapping: [mapping, setMapping],
   openIndex: [openIndex, setOpenIndex],
   isRequired = false,
-  displayedFormatter,
 }: {
   readonly table: SpecifyTable;
   readonly mapping: GetSet<RA<LiteralField | Relationship> | undefined>;
   readonly openIndex: GetSet<number | undefined>;
   readonly isRequired?: boolean;
-  readonly displayedFormatter?: boolean;
 }): JSX.Element {
   const sourcePath = React.useMemo(() => {
     const rawPath =
@@ -254,9 +252,7 @@ export function ResourceMapping({
 
   return (
     <Ul
-      className={
-        displayedFormatter === false ? 'flex w-[100%] flex-wrap gap-2' : ''
-      }
+      className="flex w-[100%] flex-wrap gap-2"
       onKeyDown={({ key }): void =>
         handleMappingLineKey({
           key,
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
@@ -67,7 +67,6 @@ export function FormattersPickList({
   return (
     <>
       <Input.Text
-        className="h-full"
         isReadOnly={isReadOnly}
         list={id('list')}
         placeholder={resourcesText.defaultInline()}
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
@@ -254,9 +254,9 @@ export function ResourceMapping({
 
   return (
     <Ul
-      className={`
-        ${mappingLineProps.length > 1 ? 'flex' : ''}
-        w-[100%] flex-wrap gap-2`}
+      className={
+        displayedFormatter === false ? 'flex w-[100%] flex-wrap gap-2' : ''
+      }
       onKeyDown={({ key }): void =>
         handleMappingLineKey({
           key,
@@ -269,21 +269,17 @@ export function ResourceMapping({
         })
       }
     >
-      <div
-        className={displayedFormatter === false ? 'flex flex-wrap gap-2' : ''}
-      >
-        {join(
-          mappingLineProps.map((mappingDetails, index) => (
-            <li className="contents" key={index}>
-              <MappingElement
-                {...mappingDetails}
-                validation={validation[index]}
-              />
-            </li>
-          )),
-          mappingElementDivider
-        )}
-      </div>
+      {join(
+        mappingLineProps.map((mappingDetails, index) => (
+          <li className="contents" key={index}>
+            <MappingElement
+              {...mappingDetails}
+              validation={validation[index]}
+            />
+          </li>
+        )),
+        mappingElementDivider
+      )}
     </Ul>
   );
 }
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
@@ -67,6 +67,7 @@ export function FormattersPickList({
   return (
     <>
       <Input.Text
+        className="h-full"
         isReadOnly={isReadOnly}
         list={id('list')}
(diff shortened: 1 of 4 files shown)


specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
```diff
@@ -266,7 +269,9 @@ export function ResourceMapping({
         })
       }
     >
-      <div>
+      <div
+        className={displayedFormatter === false ? 'flex flex-wrap gap-2' : ''}
```
maxpatiiuk, line 273: adding `flex flex-wrap gap-2` class name always seems fine - is this condition needed?

maxpatiiuk: also, this change along seems to fix the issue -why are any other changes in this PR necessary?

maxpatiiuk: but also, if you look at the code closer, the parent element of this div `Ul` already has the necessary class names (flex flex-wrap gap-2), and so this seems to suggest that the `<div>` was added here later and broke something

maxpatiiuk: and by looking at `git blame` for this file, you can see that the `<div>` was added in https://github.com/specify/specify7/commit/10ca4d8b5 as part of https://github.com/specify/specify7/pull/4114/files

maxpatiiuk: in https://github.com/specify/specify7/pull/4114#discussion_r1366110895 the screenshot shows that this was working correctly

maxpatiiuk: GitHub doesn't let me add a comment, but `${mappingLineProps.length > 1 ? 'flex' : ''}` seems not necessary - can simplify to always have `flex`

before:

Screenshot 2024-01-12 at 16 30 33

after:

Screenshot 2024-01-12 at 16 30 50

maxpatiiuk (commented):
https://github.com/specify/specify7/pull/4114 complicated code, and this PR is further complicating code (conditionally applying class names just adds more branches that need to be tested or else bug may slip though the cracks)

while it might "fix" it, complication is a no-way to go (in PartialDateUi, we were "fixing" it by adding more and more complication all the time, and it was breaking all the time)

instead, can you take a look at https://github.com/specify/specify7/pull/4114 again, to see why did that PR add an empty `<div>`?
and thus, if that div can be removed without breaking any usages?
that would simplify the code

maxpatiiuk (changes requested):
see https://github.com/specify/specify7/pull/4397#pullrequestreview-1819429066

CarolineDenis:
Here is a preview of the reasoning behind the condition for the className: 
`className={
        displayedFormatter === false ? 'flex w-[100%] flex-wrap gap-2' : ''
      }`
With the conditional: 
    
Screenshot 2024-01-16 at 7 55 11 AM

Without:   
Screenshot 2024-01-16 at 7 54 52 AM

maxpatiiuk:
Does this work better? (pushed the code)

https://github.com/specify/specify7/assets/40512816/1785a3d3-1682-470b-88f4-4c2ea6dff27c

We could also optionally remove the arrows (or condense them) as they take a lot of horizontal space

CarolineDenis:
@maxpatiiuk , 
The only issue I see is when there is only one field and no custom formatter there is empty space. See bellow: 
Screenshot 2024-01-17 at 7 26 51 AM

maxpatiiuk:
> Also, is the todo for this PR or can it wait for later?

TODOs are unrelated. Sorry for confusion)

> The only issue I see is when there is only one field and no custom formatter there is empty space. See bellow:

I am fine with the way that looks.
ANd if we change that to have the box be full width when it's the only box, then it may be a bit weird when you select a relationship in the first box, and suddenly the box shrinks a lot

---


But if you really want to chang that, you can replace this `contents` with `w-full` when there is only one box:

https://github.com/specify/specify7/blob/97356979b731c754e6d58e98176604650664b19f/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx#L270

CarolineDenis:
> I am fine with the way that looks.
I am fine with that too! Just wanted to mention it

maxpatiiuk:
unrelated to this PR, but something Interesting I noticed:

we have a `main.css` like this:

```css
  .grid-table
    :is(th, td, [role~='columnheader'], [role~='rowheader'], [role~='cell']) {
    @apply flex items-center gap-1;
  }
```

PostCSS polyfills that like this:

Screenshot 2024-01-18 at 19 23 54

(see the .does-not-exist class name added)

it is added automatically during compilation to increase specificity, to match the semantics of the `:is()` operator

[documentation for this](https://www.npmjs.com/package/@csstools/postcss-is-pseudo-class#:~:text=The%20default%20value%20is%20does%2Dnot%2Dexist)

cc @melton-jason

specifyweb/frontend/js_src/lib/components/Formatters/Fields.tsx
```diff
@@ -38,11 +38,11 @@ export function Fields({
            *   table layout with list layout
            */
           className={`
-            grid-table min-w-[35rem]
-            gap-y-4 gap-x-4
+            grid-table min-w-[35rem] gap-y-4 gap-x-4
+            [&_td]:items-stretch
```
maxpatiiuk, line 42: I didn't like that you had to add `h-full` in so many different places

reasoning:
`h-full` was needed because of the `grid-table` styling we have (`items-center`)
and given that `grid-table` is the cause of the issue, ideally, the solution for this would be somewhere in the grid table
where as by adding `h-full` in all those places, the cause of the problem and the solution to the problem are further away, making it less obvious what is the solution for

pushed a fix in https://github.com/specify/specify7/pull/4397/commits/3402921830119d428802b74b0d8d9a38305efa53

maxpatiiuk (approved):

CarolineDenis:
@maxpatiiuk , 
It looks like this now:
Screenshot 2024-01-19 at 6 37 27 AM

maxpatiiuk:
Ups, made a typo - pushed a fix

In the future, feel free to look at the changes I make in these commits and debug what could be causing the issue - would be helpful exercise

pashiav (approved):
Looks good!

issue-4380

emenslin (approved):
Looks good, stacks better now. 
Screenshot 2024-01-25 110610

PR #4402, opened 2024-01-14
ci(nodejs): check node.js version

maxpatiiuk:
Add a small dev-time only lib [check-engine](https://www.npmjs.com/package/check-engine) that ensures the Node.js version the developer has is in the range of versions supported by Specify 7.

The check happens whenever `npm install` is run
The supported version range is defined in here:
https://github.com/specify/specify7/blob/edf768a40dddf2fb93532550c7c8bb5b9cce8906/specifyweb/frontend/js_src/package.json#L14-L16

### Checklist

- [ ] Add relevant issue to release milestone. @grantfitzsimmons do we add dev-only issues to release milestones? or are internal issues no longer added to milestones? (I liked having internal milestone, but it doesn't seem like we are using that anymore)

### Testing instructions

A demo of how this works:
Screenshot 2024-01-14 at 09 14 49

feel free to try it out locally if you wish (I am using [nvm](https://github.com/nvm-sh/nvm) to have multiple node.js versions installed and switch between them easily)

### DIFF
--- a/specifyweb/frontend/js_src/package-lock.json
+++ b/specifyweb/frontend/js_src/package-lock.json
@@ -7,6 +7,7 @@
     "": {
       "name": "specify7-frontend",
       "version": "7.8.5",
+      "hasInstallScript": true,
       "license": "GPL-2.0",
       "dependencies": {
         "@codemirror/lang-json": "^6.0.0",
@@ -77,6 +78,7 @@
         "babel-jest": "^28.1.3",
         "babel-loader": "^8.2.5",
         "chalk": "^4.1.2",
+        "check-engine": "^1.12.0",
         "commander": "^9.4.1",
         "core-js": "^3.23.4",
         "css-loader": "^6.7.1",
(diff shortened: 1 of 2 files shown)


melton-jason (changes requested):
Hmmm, at least for me this functionality was already present without the addition of this dependency.
Likely since https://github.com/specify/specify7/commit/91ef1f938a7577dd87c777cc85c6689d457b9021

image

Even on production, I am unable to run `npm install` with an outdated version of node. 

image

From the npm store page it appears that the library might be able to produce a detailed list of dependencies which fall outside of the currently installed version of `node`: https://www.npmjs.com/package/check-engine?activeTab=readme

Is this the intended use case? 
Or rather, is this a discrepancy in OS or some other factor?

I am using the following:
OS: `Ubuntu 22.04.3 LTS`
NVM: 0.39.7

grantfitzsimmons:
> * @grantfitzsimmons do we add dev-only issues to release milestones? or are internal issues no longer added to milestones? (I liked having internal milestone, but it doesn't seem like we are using that anymore)

I've been operating under the assumption we'd include internal issues on the milestone since grouping them together is convenient and useful for future reference. I do think we should discuss using internal milestones again, but we'll need to talk about that at the development meeting!

maxpatiiuk:
Hmm, I remember after adding 91ef1f938a7577dd87c777cc85c6689d457b9021, trying it out, and it didn't catch the outdated node.js, so I wasn't sure what the `"engine"` field in package.json is for. Maybe I was testing it out wrong back then...
But indeed, it appears to be checked correctly now

Closing

Issue #1954, opened 2022-08-12
Forms shrinking on Safari


grantfitzsimmons:
<img width="1912" alt="image" src="https://user-images.githubusercontent.com/37256050/184411641-53e9f4d4-b113-4ca7-91d5-f8bd1e270069.png">

The forms are shrinking! On `wv_6_28_22`, go to a CO record. See it shrink!

Make the window smaller and refresh the page. Let it shrink, then make it full screen!

grantfitzsimmons:
<img width="1531" alt="image" src="https://user-images.githubusercontent.com/37256050/184413306-ce6c2f6a-639d-4eed-a38a-ae1e5750c813.png">

maxxxxxdlp:
@grantfitzsimmons said that this did not happen before. Could be a Safari bug 🤞
@grantfitzsimmons Can you check if this is happening in a different database? I.e, kufish?

grantfitzsimmons:
Only can replicate on `wv_6_28_22` on Safari

maxxxxxdlp:
https://user-images.githubusercontent.com/40512816/184926144-957a1aeb-2288-455a-aab5-5c6bf36a5175.mp4

grantfitzsimmons:
Reluctantly closing this issue. Please reopen if it is encountered!

grantfitzsimmons:
https://github.com/user-attachments/assets/46d05376-c9d0-4dbf-8578-b5919bd0b81c

PR #4405, opened 2024-01-14
fix: run w3 validation

maxpatiiuk:
Run w3 html validator (https://validator.w3.org/) on Specify 7 pages and resolved detected issues

### Checklist

### Testing instructions

I tested the impact of these refactoring changes.
Further testing could involve general testing - but that would be done pre-release anyway.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
@@ -302,6 +302,7 @@ export const Textarea = wrap<
   `${className.notTouchedInput} ${className.textArea}`,
   ({ onValueChange, isReadOnly, ...props }) => ({
     ...props,
+    type: undefined,
     ...withHandleBlur(props.onBlur),
     onChange(event): void {
       onValueChange?.(localized((event.target as HTMLTextAreaElement).value));
--- a/specifyweb/frontend/js_src/lib/components/Atoms/className.ts
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/className.ts
@@ -16,6 +16,7 @@ const secondaryButton = `${buttonClassName} hover:brightness-90 dark:hover:brigh
 dark:text-gray-100`;
 const secondaryLightButton = `${buttonClassName} hover:brightness-90 dark:hover:brightness-125 bg-[color:var(--secondary-light-button-color)] text-gray-800 shadow-sm
 dark:text-gray-100`;
+
 const containerBaseUnstyled = `flex flex-col gap-4 overflow-scroll
   overflow-x-auto [overflow-y:overlay] [scrollbar-gutter:auto]`;
 const containerBase = `${containerBaseUnstyled} bg-[color:var(--form-foreground)]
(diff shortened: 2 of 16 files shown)


specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
```diff
@@ -302,6 +302,7 @@ export const Textarea = wrap<
   `${className.notTouchedInput} ${className.textArea}`,
   ({ onValueChange, isReadOnly, ...props }) => ({
     ...props,
+    type: undefined,
```
maxpatiiuk, line 305: unset type attribute in case `props` accidentally contains it (happens if `{...getValidationAttributes(parser)}` was used in the code)

otherwise, typescript won't allow explicitly passing in `type` for textarea

specifyweb/frontend/js_src/lib/components/Atoms/className.ts
```diff
@@ -74,13 +75,13 @@ export const className = {
   containerFullGray: `${containerFull} bg-[color:var(--form-background)]`,
   containerBaseUnstyled,
   containerBase,
-  containerCenter: `${containerBase} max-w-[min(100%,var(--form-max-width))]
+  containerCenter: `${containerBase} max-w-[var(--form-max-width)]
```
maxpatiiuk, line 78: w3 was complaining about using `min()` in `max-width`

setting max-width to 100% does not make any impact (hopefully), thus it's redundant and can be simplified

(in my limited testing in chrome, this had no impact at different breakpoints with different `--form-max-width` values)

though, this could have been the cause of the infamous form shrinking bug that @grantfitzsimmons was encountering in the past (https://github.com/specify/specify7/issues/1954)

specifyweb/frontend/js_src/lib/components/Molecules/Dialog.tsx
```diff
@@ -476,7 +476,7 @@ export function Dialog({
       onRequestClose={handleClose}
     >
       {/* "p-4 -m-4" increases the handle size for easier dragging */}
-      <span
+      <div
```
maxpatiiuk, line 479: w3 does not like `<h3>` or `<div>` inside of `<span>`

specifyweb/frontend/js_src/lib/components/Preferences/Renderers.tsx
```diff
@@ -346,7 +346,7 @@ export function DefaultPreferenceItemRender({
   ) : parser?.type === 'text' ? (
     <Textarea
       forwardRef={validationRef}
-      {...(validationAttributes ?? { type: 'text' })}
+      {...validationAttributes}
```
maxpatiiuk, line 349: type should not be provided to Textarea

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -151,7 +151,7 @@ export function TreeRow({
             formatTreeStats(nodeStats, row.children === 0);
           return (
             <Button.LikeLink
-              aria-controls={id('children')}
+              aria-controls={displayChildren ? id('children') : undefined}
```
maxpatiiuk, line 154: from w3 and chatgpt it seems like `aria-controls` should not be set if the element it's pointing to does not exist (yet)

specifyweb/frontend/js_src/lib/components/WbPlanView/CustomSelectElement.tsx
```diff
@@ -692,7 +692,7 @@ export function CustomSelectElement({
 
   const listOfOptionsRef = React.useRef<HTMLElement>(null);
   const customSelectOptions = (Boolean(unmapOption) || groups) && (
-    <span
+    <div
```
maxpatiiuk, line 695: w3 does not like `<h3>` or `<div>` inside of `<span>`

specifyweb/frontend/js_src/lib/components/PickLists/index.tsx
```diff
@@ -168,7 +168,6 @@ export function PickListComboBox({
           disabled={isDisabled || isReadOnly}
           forwardRef={validationRef}
           name={name}
-          required={parser.required}
```
maxpatiiuk, line 171: this is needelss because of the `{...getValidationAttributes(parser)}` line above

seems to be here because refactor in
https://github.com/specify/specify7/commit/3ce1509d005ca338e6186816d48a807d80bbc09d#diff-6bae6d2d398090d7c917d80ff60ef98c9c15e765669841b5a2b150069acf8e53R178 didn't simplify this all the way

CarolineDenis (approved):

grantfitzsimmons:
NOTE: this is marked as a draft so that the test panel does not attempt to auto-deploy it.
see this issue for more information https://github.com/specify/specify7-test-panel/issues/110

PR #4353, opened 2024-01-03
Handle null item titles for tree rank picklists

melton-jason:
Inadvertently caused by the changes in #4273.

If the <TREE>defitem had a `NULL` title, the null title would be utilized as the picklist's title, leading to the picklist value (the resource uri of the treedefitem) to be displayed instead, as demonstrated in the following screenshot. 
Screenshot from 2024-01-03 14-49-57

The intended behavior is to use the treedefitem's name if the title is either null or an empty string. 
To prevent a similar mistake in the future, I have written automated tests to cover the functions defined in `TreeLevelPicklist.tsx`

### Testing Instructions
On a database with NULL treedefitem titles, following the testing instructions in https://github.com/specify/specify7/pull/4273#issue-2034443957

The following databases should have NULL titles for most treedefitems:
- coldFish
- gree
- dukebryo
- KUFish_5_16_23 
- KU_Fish_2023 
- KU_Fish_DEMO

(You can determine if the title for a treedefitem is NULL by running a query using `<TREE> Tree Definition Item` as a base table and adding a `title -> empty` field to the query)

PR #4406, opened 2024-01-14
tests: await async tree tests

maxpatiiuk:
When using `expect(...).resolves`, a promise is returned
I forgot to await that promise in a few tests
This meant that if test fails, the failure would be printed in the console, but the test would still be considered passing

We already have an ESLint rule that cathes this case, but, because there are too many errors highlighted, it got lost in the noise - something I definitely want to fix (as soon as xml-editor is out)

### Checklist

- [x] Add automated tests


melton-jason (approved):
Nice catch!

Those `treeLevelPicklist` tests were actually written not long ago, and by me in https://github.com/specify/specify7/pull/4353. 
Whoops. 

There are additionally some tests I found while reviewing other tests which can be simplified further: 

https://github.com/specify/specify7/blob/332069df197d376ed2f82b557c8598e66d12cd04/specifyweb/frontend/js_src/lib/utils/__tests__/fieldFormat.test.ts#L55-L61

CarolineDenis (approved):

PR #4414, opened 2024-01-17
Handle case when attachment relationship name is nonstandard

### LINKED ISSUE #4413: Cannot destructure property 'models' of 'undefined' as it is undefined when viewing DNASequence record
**Describe the bug**
When viewing any DNASequence, FieldNotebook, FieldNotebookPage, or FieldNotebookPageSet record on `v7.9.2`, the following error occurs. 

For example: https://ichthyology.specify.ku.edu/specify/view/dnasequence/30291/

Screenshot from 2024-01-16 22-04-53

[Specify 7 Crash Report - 2024-01-17T04_05_00.683Z.txt](https://github.com/specify/specify7/files/13958498/Specify.7.Crash.Report.-.2024-01-17T04_05_00.683Z.txt)

The following warning is outputted in the console: `accessing unknown field dnasequenceattachments in DNASequence value is undefined`

On `production`, the same error occurs for these tables, but only when the `attachments` icon is clicked

Screenshot from 2024-01-16 21-12-41

This bug was initially introduced in #3363, and was made somewhat less catastrophic (not immediately crashing when the form is loaded) by the changes in #4284

**Desktop:**
 - OS: Ubuntu 22.04.3 LTS
 - Browser: Chrome
 - Specify 7 Version:  7.9.2 and production

**Reported By**
@acbentley at the University of Kansas

melton-jason:
Fixes #4413 

The problem is that the following tables have a relationship to their attachment tables which is simply called `attachments` and not the 'expected' `tableNameAttachments`:
- DNASequence
- FieldNotebook
- FieldNotebookPage
- FieldNotebookPageSet

@realVinayak 
There are likely some places in `BulkAttachmentUpload` which could utilize this. 
Most notably: https://github.com/specify/specify7/blob/8907168c9785cebe8e82e6b787787412f95115de/specifyweb/frontend/js_src/lib/components/AttachmentsBulkImport/Upload.tsx#L396-L397
https://github.com/specify/specify7/blob/8907168c9785cebe8e82e6b787787412f95115de/specifyweb/frontend/js_src/lib/components/AttachmentsBulkImport/Upload.tsx#L434-L437
https://github.com/specify/specify7/blob/8907168c9785cebe8e82e6b787787412f95115de/specifyweb/frontend/js_src/lib/components/AttachmentsBulkImport/utils.ts#L332
While the affected tables are not currently supported for bulk uploading attachments with, I can imagine the tool be extended in the future to allow them. If this is likely, could this util be used in those places?

### Checklist

- [x] Add automated tests

### Testing instructions
Screenshot from 2024-01-16 21-12-41

For each of the following tables, ensure no error is thrown and attachments can be viewed when the `attachments` button is clicked (in the context of a viewing a single resource and in a recordset)
- DNASequence
- FieldNotebook
- FieldNotebookPage
- FieldNotebookPageSet

Additionally, ensure that the component still functions correctly for other tables with standard names for attachment relationships (CollectionObject, CollectingEvent, etc)


realVinayak:
@melton-jason it would be, once we support those!

specifyweb/frontend/js_src/lib/components/Attachments/__tests__/utils.test.ts
```diff
@@ -0,0 +1,32 @@
...
+
+requireContext();
+
+test('attachmentRelatedTables', () =>
+  expect(attachmentRelatedTables()).toMatchSnapshot());
+
+test('allTablesWithAttachments', () =>
+  expect(allTablesWithAttachments()).toMatchSnapshot());
+
+test('getAttachmentRelationship', () => {
```
maxpatiiuk, line 19: (optional) could use `theories` util or https://jestjs.io/docs/api#testeachtablename-fn-timeout here

specifyweb/frontend/js_src/lib/components/Attachments/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+ * - FieldNotebook
+ * - FieldNotebookPage
+ * - FieldNotebookPageSet
+ */
+export const getAttachmentRelationship = (
+  table: SpecifyModel
+): Relationship | undefined =>
+  table.name === 'Attachment'
+    ? undefined
+    : table.relationships.find((relationship) => {
```
maxpatiiuk, line 45: (optional optimization) could optimize for the common case by looking up the field by it's probable name directly on `table.field` dictionary, and only falling back to `.find()` if it's not found there

maxpatiiuk (approved):
Ah, I should have checked earlier if my assumption of `<tableName>Attachments` always holds true

---

I pushed a commit that adds one more usage of your util - can you review that?

grantfitzsimmons (approved):

specifyweb/frontend/js_src/lib/components/Attachments/__tests__/utils.test.ts
melton-jason: I had initially used `theories` and `test.each` here, but in both cases it seems the tests were being run in an environment without access to the schema. 
In other words, they were missing the context from the `requireContext()` call.

maxpatiiuk: ah I see
that is because you were using `schema.models.DNASequence` in the arguments array
the arguments array are created as soon as the file is opened - thus, before schema is loaded
where as the `test()` is only called after schema is loaded, thus accessing `schema.models.DNASequence` inside of it is not a problem

two solutions to that (the way you have it now is fine, just mentioning for your awareness):
- in the test file, create a small wrapper for getAttachmentRelationship - the wrapper would accept table name as a string, and resolve that to the table (since this resolving would only happen when the test is being run, the schema is loaded by then)
- modify `theories` to support passing a function that would be called by `theories` to get the arguments, rather than providing arguments right away. similar to have `overrideAjax` supports that - see this usage for example https://github.com/specify/specify7/blob/3402921830119d428802b74b0d8d9a38305efa53/specifyweb/frontend/js_src/lib/components/QueryBuilder/__tests__/fromTree.test.ts#L35

specifyweb/frontend/js_src/lib/components/Attachments/utils.ts
```diff
@@ -39,12 +39,18 @@ export const tablesWithAttachments = f.store(() =>
  */
 export const getAttachmentRelationship = (
   table: SpecifyModel
-): Relationship | undefined =>
-  table.name === 'Attachment'
-    ? undefined
-    : table.relationships.find((relationship) => {
+): Relationship | undefined => {
+  if (table.name === 'Attachment') return undefined;
+  const commonRelationship = table.field[`${table.name}Attachments`] as
```
maxpatiiuk, line 44: table.name is capitalized, but all fields are starting with lowercase, thus this will never match

so you would need something like `${table.name[0].toLowercase()}${table.name.slice(1)}Attachments`

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/Attachments/utils.ts
melton-jason: Fixed in https://github.com/specify/specify7/pull/4414/commits/2761272f9f3a2ead15f8e2a773714626409b908e

specifyweb/frontend/js_src/lib/components/Attachments/__tests__/utils.test.ts
melton-jason: Fixed in https://github.com/specify/specify7/pull/4414/commits/693113be83c3fe55ab27bc31c56394af42215215

maxpatiiuk (approved):

pashiav (approved):
Looks good! No longer receiving an error.

melton-jason:
Before this is merged, can this briefly be tested again? 
The changes were made before `xml-edtor` was merged into `production` and I have merged production back into this branch. 

No general testing should be required: only the fix for #4413 (see the testing instructions in https://github.com/specify/specify7/pull/4414#issue-2085440566)

emenslin (changes requested):
When you try to create a record set it throws an error. Not able to recreate in edge 

https://github.com/specify/specify7/assets/150940840/04c72b5c-2f2f-4a2b-ba36-b6e81d8f4441

melton-jason:
> When you try to create a record set it throws an error. Not able to recreate in edge
> 
>  LjnqsC8kLG.mp4

Fixed!

emenslin (approved):
Looks good! Attachment gallery not throwing error

Issue #4424, opened 2024-01-18
Create a 'Release checklist' issue template


grantfitzsimmons:
@maxpatiiuk asked if we should create a release checklist. I believe it's worth considering and potentially useful so we can ensure we complete all necessary tasks. This process is mostly in our heads at the moment, so formalizing it and making it public might be a good idea.

> It would have items on it like general testing, getting release milestone to 0, creating tag, updating specify cloud and etc
> 
> and it could also have post-release cleanup tasks like updating dev dependencies and merging PRs before they get stale
> 
> one way to do that is to create a "Release Checklist" issue template, and create an issue out of that template for each release (and maybe even pin it in the list of issues)
> alternatively, this could be a google doc, but I would prefer it to be on github (better integration with github's issues/PRs/milestons, and the fact that it is public)

CarolineDenis:
@grantfitzsimmons, can this be closed?

grantfitzsimmons:
Oh yes!

PR #4425, opened 2024-01-18
Add a "Release checklist" issue template

grantfitzsimmons:
Addresses #4424

> @maxpatiiuk asked if we should create a release checklist. I believe it's worth considering and potentially useful so we can ensure we complete all necessary tasks. This process is mostly in our heads at the moment, so formalizing it and making it public might be a good idea.
> 
> > It would have items on it like general testing, getting release milestone to 0, creating tag, updating specify cloud and etc
> > 
> > and it could also have post-release cleanup tasks like updating dev dependencies and merging PRs before they get stale
> > 
> > one way to do that is to create a "Release Checklist" issue template, and create an issue out of that template for each release (and maybe even pin it in the list of issues)
> > alternatively, this could be a google doc, but I would prefer it to be on github (better integration with github's issues/PRs/milestons, and the fact that it is public)


CarolineDenis (approved):

maxpatiiuk (approved):
thanks for creating this!

looks great, and I think this has the possibility of avoiding some headaches in the future

now, would you like to create an issue from this template for 7.9.4? (too late for 7.9.3 probably)

melton-jason (approved):

PR #4433, opened 2024-01-19
fix(Dialog): improve useFreezeDialogSize logic

### LINKED ISSUE #4385: `xml-editor`: Dialog sizing are experiencing issues allover Sp
**Describe the bug**
The new app resource name dialog should be appropriately sized when you create a new app resource. It has been quite jarring as I've used `xml-editor` more and more. 

Initially I thought it was because I resized it on a specific database but I have been able to recreate this on most databases. _Even if it typically remembers your dialog position/size,_ it should not in this case.

**To Reproduce**
1. Create a new app resource that can be given a name
2. See that the 'New App Resource' dialog is larger than it should be

**Expected behavior**

The behavior on `edge` (`v7.9.2`) is desired. See below:

https://github.com/specify/specify7/assets/37256050/e5c51f59-b022-4fe6-a89a-53cf80d0e227

**Screenshots:**
Screenshot 2024-01-10 at 12 59 11 PM

macOS, Arc

CarolineDenis:
See PR description in https://github.com/specify/specify7/pull/4433#issuecomment-1943078176

<details>
<summary>Old Description</summary>
Fixes #4385 

### Checklist

### Testing instructions

- open app resource
- click on add resource 
- click form definition 
- verify that the dialog to define the name of the def is smaller than before 

</details>

Testing instructions

Verify that https://github.com/specify/specify7/issues/4519 is fixed
Verify that https://github.com/specify/specify7/issues/4511 is fixed
Verify that opening GeoMap in full screen, and then reducing the browser window size does not cause dialog to overflow (should not have horizontal scroll bar)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -10,7 +10,9 @@ jobs:
     # "push" event is not triggered for forks, so need to listen for
     # pull_requests, but only for external ones, so as not to run the action
     # twice
-    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.fork == true
+    if:
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.head.repo.fork == true
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -27,6 +29,8 @@ jobs:
           elif [[ $GITHUB_REF == refs/tags/* ]]; then
             VERSION=${GITHUB_REF#refs/tags/}
           elif [[ $GITHUB_REF == refs/heads/* ]]; then
+            # Docker does not like / in tag names, so replace them with -
+            # See https://github.com/specify/specify7-test-panel/issues/110#issuecomment-1943045253
             VERSION=$(echo ${GITHUB_REF#refs/heads/} | sed -r 's#/+#-#g')
             if [ "${{ github.event.repository.default_branch }}" = "$VERSION" ]; then
               VERSION=edge
@@ -50,7 +54,7 @@ jobs:
           echo "version=${VERSION}" >> $GITHUB_OUTPUT
           echo "tags=${TAGS}" >> $GITHUB_OUTPUT
           echo "created=$(date -u +'%Y-%m-%dT%H:%M:%SZ')" >> $GITHUB_OUTPUT
-          
+
           # Determine the platforms based on the branch
           PLATFORMS="linux/amd64"
           REF_NAME=$(echo "${GITHUB_REF#refs/*/}")
@@ -60,7 +64,7 @@ jobs:
             PLATFORMS="linux/arm64"
           fi
           echo "platforms=${PLATFORMS}" >> $GITHUB_ENV
-      
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
 
--- a/specifyweb/frontend/js_src/lib/components/Atoms/Internationalization.ts
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/Internationalization.ts
@@ -48,7 +48,11 @@ declare namespace Intl {
     public constructor(
       locales?: RA<string> | string,
       options?: {
-        readonly unit: 'byte'; // TODO: Expand unit
+        /*
+         * Full list of possible units:
+         * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#unit_2
+         */
+        readonly unit: 'byte';
         readonly notation:
           | 'compact'
           | 'engineering'
--- a/specifyweb/frontend/js_src/lib/components/Molecules/Dialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/Dialog.tsx
@@ -14,10 +14,11 @@ import { useCachedState } from '../../hooks/useCachedState';
 import { useId } from '../../hooks/useId';
 import { listen } from '../../utils/events';
 import { f } from '../../utils/functools';
-import { KEY } from '../../utils/utils';
+import { KEY, removeKey } from '../../utils/utils';
 import { Button, DialogContext } from '../Atoms/Button';
 import { className, dialogIconTriggers } from '../Atoms/className';
 import { dialogIcons } from '../Atoms/Icons';
+import { SECOND } from '../Atoms/timeUnits';
 import { LoadingContext } from '../Core/Contexts';
 import {
   useHighContrast,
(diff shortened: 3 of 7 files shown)


CarolineDenis:
NOTES: 
This push is the best solution I found so far but I'm not entirely satisfied... 
Here is the before after the changes: 
BEFORE
Screenshot 2024-01-19 at 7 32 01 AM
AFTER
Screenshot 2024-01-19 at 7 32 43 AM

specifyweb/frontend/js_src/lib/components/SkeletonLoaders/Form.tsx
```diff
@@ -4,7 +4,7 @@ import { Skeleton } from './Skeleton';
 
 export function FormSkeleton() {
   return (
-    <Skeleton.Root className="flex-col">
+    <Skeleton.Root className="flex-col flex-wrap overflow-auto">
```
maxpatiiuk, line 7: does `flex-wrap` make any difference here?
I though it doesn't change anything in most cases unless you have `flex-row`

maxpatiiuk:
so the skeleton is or is not the cause of the problem #4385?

if it is the cause, then one of the solutions proposed in https://github.com/specify/specify7/issues/4385#issuecomment-1899756185 should resolve this issue properly (i.e could try out the timeout on the useFreezeDialogSize hook)
if skeleton is not the cause, then this change is unrelated?

specifyweb/frontend/js_src/lib/components/SkeletonLoaders/Form.tsx
CarolineDenis: No you're right. Doesn't seem like it

CarolineDenis:
@maxpatiiuk , 

So I tried the timeout and it doesn't seem to do anything. 
I changed the className and it's not conclusive. 

The only thing that I found that makes the difference and give us the right size (not a solution of course, just to demonstrate) is to change: 
`      dimensionsKey={viewName ?? resource?.specifyTable.view}
`
by 
`      dimensionsKey={resource?.specifyTable.view}
`
in ResourceView.tsx

CarolineDenis:
@maxpatiiuk, 
Yes it is reproducible in incognito mode with the className changes on the skeleton and with the timeout too...

maxpatiiuk:
Fixes #4385

Given that the original reason for adding the "prevent dialog size from shrinking" feature was to avoid dialog resizing when going between query builder results, and that place is now fixed (by using h-full w-full style), I would be okay with removing the "prevent dialog size from shrinking" feature if it would cause more trouble in the future

### Checklist

### Testing instructions

- Verify that https://github.com/specify/specify7/issues/4519 is fixed
- Verify that opening GeoMap in full screen, and then reducing the browser window size does not cause dialog to overflow (should not have horizontal scroll bar)

pashiav (changes requested):
- #4519 looks good!

- Verified that opening GeoMap in full screen, and then reducing the browser window size does not cause dialog to overflow!


https://github.com/specify/specify7/assets/118556884/dbac1f6f-59ab-4fc7-817c-bbef715df7a3



#4511 can still be reproduced. 

https://github.com/specify/specify7/assets/118556884/091211cd-da86-43b1-a10e-8d79e5e5c44e

maxpatiiuk:
regarding https://github.com/specify/specify7/issues/4511#issue-2127883881, I am actually thinking this should be considered a feature, and not a bug

@pashiav @grantfitzsimmons what do you think?

reasoning:

the idea of the "Prevent dialog size from reducing" feature is to reduce UI shifts, both because they are annoying, and because they can cause user to misclick (they meant to click in one place, but dialog resized, and now they clicked on a wrong button, or worse, outside the dialog, causing the dialog to close)

This was most useful in the "View in Forms" button in the query builder, but was implemented as a specify-wide feature as it could benefit other dialogs too.

case in point, if you look at the video in https://github.com/specify/specify7/issues/4511#issue-2127883881, see how because the dialog is resizing, when you click on the "+" button, you have to "catch" the button again every time you click it (that's why layout shifts are bad and should be avoided when possible) - in comparison, the Delete button is not running away from you - it stays in the same place even as the dialog gets to have less content - isn't that a useful UX feature?
we can easily see in the video in https://github.com/specify/specify7/issues/4511#issue-2127883881 that thanks to fewer layout shifts, you were able to remove records faster than you were able to add them (with exception of the last click that accidentally  opened the Form Meta dialog - this was caused by layout shift from the table row disappearing - something we should consider improving in some way)

Though if you would say that "it might look like a bug to a user" or "it makes the dialog take more space than it should", then I would propose we change to have this feature only enabled for modal dialogs (modal dialogs are dialogs that gray out the background when open), or have it only apply horizontally rather than vertically (prevent dialogs from becoming less wide, but not touching the height - though, this negates some of the UX benefits of this feature)

https://github.com/specify/specify7/issues/4511#issuecomment-1936642926 is definitely a bug though and should be fixed by this PR

grantfitzsimmons:
@maxpatiiuk 

I agree with you as long as we make sure the comment you mentioned is resolved and that skeleton loaders do not arbitrarily make dialogs bigger. As you mentioned, modal dialogs are the only ones that I want this to happen to. Not sure it would be necessary to restrict vertical expansion in all cases, so I'd like to scope this a bit.

> Though if you would say that "it might look like a bug to a user" or "it makes the dialog take more space than it should", then I would propose we change to have this feature only enabled for modal dialogs (modal dialogs are dialogs that gray out the background when open), or have it only apply horizontally rather than vertically (prevent dialogs from becoming less wide, but not touching the height - though, this negates some of the UX benefits of this feature)
> 
> https://github.com/specify/specify7/issues/4511#issuecomment-1936642926 is definitely a bug though and should be fixed by this PR

maxpatiiuk:
Sounds good. Updated the testing instructions in https://github.com/specify/specify7/pull/4433#issuecomment-1943078176

acwhite211 (approved):

CarolineDenis:
@maxpatiiuk, should we add an exception to cover this: https://github.com/specify/specify7/issues/4549

maxpatiiuk:
> Dialog shrinks after content is removed in queries and record set dialogs:

In this case, this is accidental. Even though the code says "don't allow shrinking for modal dialogs", in this case it seems like the dialog is temporary removed (replaced by a loading skeleton dialog), and then a new dailog with new query results are added (thus size is not preserved in between)

> should we add an exception to cover this: https://github.com/specify/specify7/issues/4549

options:
- make that dialog non-modal (given that the dialog freezing logic does not apply to modal dialogs)
- set max-size for the loading skeleton in this case
- pass `dimensionsKey={false}` to the Dialog WHILE the loading skeleton is visible

what solution would you like @CarolineDenis? and what would you be willing to implement 😁

PR #4435, opened 2024-01-19
fix(Cache-Buster): Handle case when method is not allowed

### LINKED ISSUE #4432: Clear browser cache dialog not being displayed
**Describe the bug**
When I click the clear browser cache button on edge the popup does not show up like it does on 7.9.2

**To Reproduce**
1. Click on clear browser cache in user tools 
2. See that there is no popup 

**Expected behavior**
You should get the popup saying that the cache was cleared. 

**Screenshots**
First I tested in edge and then I switched to 7.9.2 to show the original/expected behavior 

https://github.com/specify/specify7/assets/150940840/e7d0c633-c40e-4007-ad15-a31d7a3927d0

**Desktop:**
 - OS: Windows 11
 - Browser: Chrome
 - Specify 7 Version: edge

melton-jason:
Fixes #4432

### Checklist

### Testing instructions

See https://github.com/specify/specify7/issues/4432#issue-2090779004

### DIFF
--- a/specifyweb/businessrules/views.py
+++ b/specifyweb/businessrules/views.py
@@ -113,7 +113,7 @@ UniquenessRuleSchema = {
     }
 })
 @login_maybe_required
-@require_http_methods(['GET', 'PUT'])
+@require_http_methods(['HEAD', 'GET', 'PUT'])
 @transaction.atomic
 def uniqueness_rule(request, discipline_id):
     data = {}
--- a/specifyweb/frontend/js_src/lib/components/RouterCommands/CacheBuster.tsx
+++ b/specifyweb/frontend/js_src/lib/components/RouterCommands/CacheBuster.tsx
@@ -19,7 +19,7 @@ export const clearCache = async (): Promise<true> =>
         method: 'HEAD',
         cache: 'no-cache',
         errorMode: 'silent',
-        expectedErrors: [Http.NOT_FOUND],
+        expectedErrors: [Http.NOT_FOUND, Http.NOT_ALLOWED],
       }).then(() => console.log(`Cleaned cache from ${endpoint}`))
     )
   ).then(() => {


grantfitzsimmons (changes requested):
https://github.com/specify/specify7/assets/37256050/292f0aa5-95f2-496f-a3a0-ae83a4d64e58

I'm still getting a white screen rather than a confirmation dialog. 

Arc (Chromium Engine Version 120.0.6099.234), macOS

specifyweb/businessrules/views.py
```diff
@@ -113,7 +113,7 @@
     }
 })
 @login_maybe_required
-@require_http_methods(['GET', 'PUT'])
+@require_http_methods(['HEAD', 'GET', 'PUT'])
```
maxpatiiuk, line 116: (optional) could we create a wrapper for the `require_http_methods` decorator or some sort of middleware that makes every view that supports 'GET' also handle 'HEAD'?
there are no use cases for not supporting 'HEAD' when 'GET' is supported
having to specify it explicitly is only a source of bugs

maxpatiiuk: alternatively, we could make cache buster use GET instead of HEAD requests, but, HEAD are more efficient/faster for this purpose, and it's a good idea in general for API to support HEAD requests

maxpatiiuk (approved):

specifyweb/businessrules/views.py
melton-jason: Perhaps not the best solution, but I could not explicitly find a way in native Django to accomplish what we're looking for. 
I've written the two decorators used to state that `GET` is required with our desired functionality and used these decorators throughout the backend: https://github.com/specify/specify7/pull/4435/commits/3498c8e6122f1a5c1778231f51c3e52acfb65796

It seems most places like those decorators are defined in `specify/views.py`, although I doubt that would be the best place for them: I would expect a `views.py` to only contain the views for the urls, and little to nothing else. 
I'd advocate for moving similar functions, classes, etc., outside of the `views.py` and into another spot (i.e., `middleware/general.py` if used extensively throughout the backend). 

Though I can move these two decorators to `specify/views.py` to maintain consistency.

melton-jason:
> issue-4432.mp4 
> I'm still getting a white screen rather than a confirmation dialog.
> 
> Arc (Chromium Engine Version 120.0.6099.234), macOS

Fixed!

grantfitzsimmons (approved):

maxpatiiuk (approved):
❤️

melton-jason:
Yes, this can be merged once the merge conflicts are resolved.

specifyweb/middleware/general.py
```diff
@@ -1,26 +1,54 @@
...
+    if 'GET' in _request_method_list:
+        request_method_list.add('HEAD')
+
+    def callable(function: Callable):
+        @raw_require_http_methods(list(request_method_list))
+        def wrapped(*args, **kwargs):
+            return function(*args, **kwargs)
+        return wrapped
+
+    return callable
```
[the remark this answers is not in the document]
melton-jason: Fixed in https://github.com/specify/specify7/pull/4435/commits/62e3a2c283222dcf0c3e2886cc4d266ba0a63205

emenslin (approved):
Looks good, popup is now there. 

https://github.com/specify/specify7/assets/150940840/a757bb62-19ac-4d5c-a9b1-a50ffd9947a9

specifyweb/middleware/general.py
```diff
@@ -1,26 +1,46 @@
...
+
+def require_http_methods(_request_method_list: List[REQUEST_METHOD]):
+    request_method_list = set(_request_method_list)
+    if 'GET' in _request_method_list:
+        request_method_list.add('HEAD')
+
+    def callable(function: Callable):
+        return raw_require_http_methods(list(request_method_list))(function)
+        
+    return callable
```
realVinayak, line 26: Realized that this can be simplified more.

```suggestion
    return raw_require_http_methods(list(request_method_list))
```

melton-jason: fixed in https://github.com/specify/specify7/pull/4435/commits/e7843212e832270c0d5ef0b47356a08e09413cc4

realVinayak (approved):
much cleaner!

PR #4463, opened 2024-01-25
Move down the add button in uniqueness rule config

### LINKED ISSUE #4452: configure uniqueness rule dialog add button should move once clicked
**Describe the bug**

Since moving the `Add` button in line with the unique fields list, the sizing and alignment is weird when the red minus icons appear. 
Screenshot 2024-01-23 at 8 32 39 AM

**Expected behavior**

image

CarolineDenis:
Fixes #4452

### Checklist

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/SchemaConfig/UniquenessRuleRow.tsx
+++ b/specifyweb/frontend/js_src/lib/components/SchemaConfig/UniquenessRuleRow.tsx
@@ -246,39 +246,41 @@ function ModifyUniquenessRule({
         <H2>{invalidUniqueReason}</H2>
         <p>{schemaText.uniqueFields()}</p>
         {rule.fields.map((field, index) => (
-          <div className="inline-flex gap-2" key={index}>
-            <PickList
-              disabled={readOnly}
-              groups={{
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/SchemaConfig/UniquenessRuleRow.tsx
```diff
@@ -246,39 +246,41 @@ function ModifyUniquenessRule({
...
+                  onClick={(): void =>
+                    handleChanged({
+                      ...rule,
+                      fields: removeItem(rule.fields, index),
+                    })
+                  }
+                />
+              )}
+            </div>
             {rule.fields.length - 1 === index && (
```
maxpatiiuk, line 284: can this button be placed outside the `.map()` since it only needs to be rendered once?

maxpatiiuk: and that would also allow to remove the doubly-nested div here (only one div would be needed)

maxpatiiuk (changes requested):

maxpatiiuk (approved):

PR #4474, opened 2024-01-27
Change 'Transactions' dialog heading to 'Interactions', add chat icon to dialog, refactor dialog icons

### LINKED ISSUE #4473: Make 'Transactions' dialog consistent with other mentions of Interactions
At this time, the 'Show Loans'/'Show Interactions' buttons show a dialog that is titled 'Transactions', further confusing the user as to what the 'interactions' are. While these are 'transactions' it is more accurate to call these 'associated interactions' than 'transactions'.

grantfitzsimmons:
Fixes #4473

This makes things more consistent. This has been mentioned to me several times over the years and I figured it's time to align these together. Specify 6 titled this dialog 'Show Interactions' so this feels more in line with that.

### Checklist

- [ ] Add automated tests

**Previous Appearance (`v7.9.3`):**

image

**New Appearance:**

image

### Testing instructions

1. Open a Collection Object record that has Preparations associated with interactions
2. Click on the 'Show Loans' or 'Show Interactions' button
3. Verify the icon in the dialog matches the 'Interactions' menu item
4. Verify 'Transactions' is now 'Interactions'
5. Verify table icons appear next to the interactions subheadings
6. Verify all dialog icons appear in the correct color (red for danger, orange for warning, blue by default)

### Additional notes

The lack of the link out icon(Screenshot 2024-01-27 at 3 39 11 PM) seems to be a change made in `xml-editor`– https://github.com/specify/specify7/commit/d3de9c3a6573c66288b716ee54309e59f1142f20, it's not a part of this PR


specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
```diff
@@ -114,10 +116,12 @@ export function ShowLoansCommand({
   return typeof data === 'object' ? (
     <Dialog
       buttons={commonText.close()}
-      header={commonText.transactions()}
+      header={interactionsText.interactions()}
+      icon={<span className="text-blue-500">{icons.chat}</span>}
```
melton-jason, line 120: (Just an observation) 

There are a plethora of places in Specify which use `<span className="text-blue-500">{icons.someIcon}</span>` specifically for Dialog icons. 
It may be worth creating a utility function (perhaps in https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/components/Atoms/Icons.tsx) which would accept a JSX element (or a `keyof icons`) and returns a span with that className and the corresponding icon. 

We could then replace all instances where this dialog icon convention is used with a call to the utility function.

specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
```diff
@@ -114,10 +116,12 @@ export function ShowLoansCommand({
...
       onClose={handleClose}
     >
-      <H3>
+      <H3 className={`flex items-center gap-2`}>
```
melton-jason, line 123: The curly braces here are not needed: a simple string will do!
Curly braces are only needed in an element's className if an expression is needed to be evaluated. 
Commonly this is either concatenation or a string template literal (we prefer to use a string template literal). 
For an example where a string template literal is needed, see https://github.com/specify/specify7/blob/7324c28adc552ecb2aefd4c0533433001103cffc/specifyweb/frontend/js_src/lib/components/AppResources/Aside.tsx#L72-L77

```suggestion
      <H3 className="flex items-center gap-2">
```

If configured, ESLint would yell at you for this! 
Though this suggested change should occur automatically when the `Lint Code with ESLint + Prettier` commit is pushed.

specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
```diff
@@ -127,7 +131,8 @@ export function ShowLoansCommand({
         fieldName="loan"
         resources={data.openLoans ?? []}
       />
-      <H3>
+      <H3 className={`flex items-center gap-2`}>
```
melton-jason, line 134: See above

specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
```diff
@@ -137,7 +142,8 @@ export function ShowLoansCommand({
         fieldName="loan"
         resources={data.resolvedLoans ?? []}
       />
-      <H3>
+      <H3 className={`flex items-center gap-2`}>
```
melton-jason, line 145: See above

melton-jason (approved):

specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
grantfitzsimmons: That's what we pay ESLint to do!!! /s

specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
maxpatiiuk: good idea. @grantfitzsimmons can you tackle that?

though, while we have the dialog icon be blue in many places, it's not always the case - probably just accidental inconsistency

thus, would it make more sense to modify `<Dialog>` to accept an icon name in the `icon` prop, and it would make that icon blue automatically?

specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
maxpatiiuk: @grantfitzsimmons you marked these comments as resolved, but I don't see them resolved in the code - did you forget to push or are you still working on them?

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
grantfitzsimmons: Should occur with 'Lint Code with ESLint + Prettier', no?

specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
melton-jason: That may be more sensible, although _always_ making the dialog icon blue may not be the desired behavior in all cases. 
For instance, and most notably, when using a `dialogIcon`:

https://github.com/specify/specify7/blob/956a14bb75cb986a53f6f04988eca19fd66deae5/specifyweb/frontend/js_src/lib/components/Atoms/Icons.tsx#L134-L147

specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
melton-jason: Yes, this problem is auto-fixable so the `ESLint + Prettier` commit should automatically resolve these. 
Though worryingly, the commit never ran even though it theoretically _should of_  after https://github.com/specify/specify7/pull/4474/commits/9809a657f3ea000aacf6d5ffe7ca563dc5046802 was pushed: the frontend tests were ran and succeeded.

specifyweb/frontend/js_src/lib/components/FormCommands/ShowTransactions.tsx
grantfitzsimmons: @maxpatiiuk @melton-jason see my latest commits and let me know what you think

specifyweb/frontend/js_src/lib/components/Molecules/Dialog.tsx
```diff
@@ -485,9 +485,11 @@ export function Dialog({
...
       >
         <div className="flex items-center gap-2">
-          {typeof defaultIcon === 'object' && showIcon
-            ? defaultIcon
-            : dialogIcons[iconType]}
+          <span className="text-blue-500">
+            {typeof defaultIcon === 'object' && showIcon
+              ? defaultIcon
+              : dialogIcons[iconType]}
+          </span>
```
maxpatiiuk, line 492: after this change the code is always displaying a `<span>` even when it's empty, which causes the "gap-2" in the parent flex container to add some padding to the right of the header even when not needed

instead, define a variable like this:
```tsx
  const resolvedIcon = typeof defaultIcon === 'object' && showIcon
    ? defaultIcon
    : dialogIcons[iconType];
```

and then use it to only render span when needed:

```suggestion
          {resolvedIcon === undefined ? undefined : <span className="text-blue-500">{resolvedIcon}</span>}
```

maxpatiiuk: oh, I just realized that the only case when icon is not displayed is when you are in a Resource view dialog, or when user disabled the icon in settings - still, let's fix this bug

maxpatiiuk (changes requested):
ah that's smart - my idea was going to be to wrap the icon in `<span>` if I detect that user passed in an icon name (string) rather than icon jsx element (object)

but your solution of always wrapping the icon in span, but letting user of the component overwrite it if needed is much cleaner!

---

still, there is one small bug with your implementation - fix that, and this is ready to go

realVinayak (approved):

specifyweb/frontend/js_src/lib/components/Atoms/Icons.tsx
```diff
@@ -143,7 +143,7 @@ export const dialogIcons: RR<
     </span>
   ),
   success: <span className="text-green-500">{icons.checkCircle}</span>,
-  info: <span className="text-blue-500">{icons.informationCircle}</span>,
+  info: <span>{icons.informationCircle}</span>,
```
maxpatiiuk, line 146: you might not need this empty span anymore

```ts
info: icons.informationCircle,
```

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/Atoms/Icons.tsx
sharadsw: fixed

Areyes42 (approved):
### Testing instructions

1. [x] Open a Collection Object record that has Preparations associated with interactions
2. [x] Click on the 'Show Loans' or 'Show Interactions' button
3. [x] Verify the icon in the dialog matches the 'Interactions' menu item
4. [x] Verify 'Transactions' is now 'Interactions'
5. [x] Verify table icons appear next to the interactions subheadings
6. [x] Verify all dialog icons appear in the correct color (red for danger, orange for warning, blue by default)

***
Looks great! All dialog and icons in the interaction menu are updated and displayed properly.
Screenshot 2024-05-22 at 3 27 12 PM

lexiclevenger (approved):
- Open a Collection Object record that has Preparations associated with interactions
- Click on the 'Show Loans' or 'Show Interactions' button
- Verify the icon in the dialog matches the 'Interactions' menu item
- Verify 'Transactions' is now 'Interactions'
- Verify table icons appear next to the interactions subheadings
- Verify all dialog icons appear in the correct color (red for danger, orange for warning, blue by default)
---
Looks good!
Screenshot 2024-05-22 at 4 24 17 PM

Issue #1562, opened 2022-05-12
Data Sets should use the table icon they are mapped to


grantfitzsimmons:
![Screen Shot 2022-05-12 at 12 11 45 PM](https://user-images.githubusercontent.com/37256050/168131121-7a5875e7-aeca-4ac0-9917-e679b08c6b7d.png)

The WB icon does not indicate anything, especially because it is only visible in the Data Sets window.

maxxxxxdlp:
The limitation here is that detecting the data set base table would require fetching the upload plan for each data set, which would be rather inefficient.

Perhaps modifying the API response would be a better soluton

grantfitzsimmons:
![image](https://github.com/specify/specify7/assets/37256050/ba5a905e-9b87-42c3-b11d-32fc4c8513e1)
We would need to add the `uploadplan` when all the data sets are listed. If this is undefined, it should show `Workbench` instead

PR #4475, opened 2024-01-28
Add table icons to the WorkBench data sets dialog based on base table

grantfitzsimmons:
Solves #1562 

image

**Steps to test:**
1. Verify that all existing data sets use the base table icon rather than the "WB" icon in the WorkBench data sets dialog
2. Create a new data set in several base tables
3. Add many columns with complicated mappings
4. Add values across the spreadsheet
5. Save and upload
6. Open dialog and verify the base table icon is used and all looks correct
7. Upload an existing spreadsheet
8. Map, validate, and upload
9. Open dialog and verify the base table icon is used and all looks correct
10. Perform the testing steps 2-8 with the batch attachment uploader to ensure all is working as expected (no icons will show in that dialog)


specifyweb/frontend/js_src/lib/components/Toolbar/WbsDialog.tsx
```diff
@@ -241,7 +241,11 @@ export function DataSetsDialog({
                           : undefined
                       }
                     >
-                      <TableIcon label={false} name="Workbench" />
+                      {dataset.uploadplan ? (
+                        <TableIcon label name={dataset.uploadplan.baseTableName} />
+                      ) : (
+                        <TableIcon label name="Workbench" />
+                      )}
```
maxpatiiuk, line 248:
```suggestion
                      <TableIcon label name={dataset.uploadplan?.baseTableName ?? 'Workbench'} />
```

maxpatiiuk (changes requested):
Code looks good but I have a concern
The reason upload plan wasn't added by Ben to this endpoint is that it could be bad for performance (make the data sets dialog take more time to load, especially since we don't have any pagination or caching for this endpoint)

could you modify this PR to only add baseTableName field instead of the entire uploadPlan?

specifyweb/frontend/js_src/lib/components/Toolbar/WbsDialog.tsx
```diff
@@ -241,11 +241,7 @@ export function DataSetsDialog({
                           : undefined
                       }
                     >
-                      {dataset.uploadplan ? (
-                        <TableIcon label name={dataset.uploadplan.baseTableName} />
-                      ) : (
-                        <TableIcon label name="Workbench" />
-                      )}
+                      <TableIcon label name={dataset.uploadplan?.baseTableName ?? 'Workbench'} />
```
maxpatiiuk, line 244: now this looks good, but just a FYI for the future, that even without the fancy `?.` syntax, the original version you had could have been simplified by moving the `?` inside as much as possible so that the duplication is avoided

```tsx
<TableIcon label name={dataset.uploadplan ? dataset.uploadplan.baseTableName : 'Workbench'} />
```

grantfitzsimmons:
>The reason upload plan wasn't added by Ben to this endpoint is that it could be bad for performance (make the data sets dialog take more time to load, especially since we don't have any pagination or caching for this endpoint)

This is interesting– and absolutely something I considered. The upload plan is generally much smaller than the whole data set, but there are situations where it can become fairly complicated. I'll see if I can only add `baseTableName`

maxpatiiuk:
> This is interesting– and absolutely something I considered. The upload plan is generally much smaller than the whole data set, but there are situations where it can become fairly complicated. I'll see if I can only add baseTableName

No problem - the other thing you can do is benchmark what impact this change makes on a representative or worst case database

And also, adding pagination is something we have to do anyway at some point, so that would mitigate the concern quite a bit

maxpatiiuk:
<details>
<summary>spoiler about the cause of the TypeScript error you are getting in GitHub Actions</summary>
The typescript type you modified appears to be used not just in the WorkBench, but also in the Attachments Bulk Upload - you would likely need to create a separate TypeScript type that extend this one with one new field added, and then use this new TypeScript type in the WorkBench
</details>

specifyweb/frontend/js_src/lib/components/Toolbar/WbsDialog.tsx
```diff
@@ -241,7 +241,7 @@ export function DataSetsDialog({
                           : undefined
                       }
                     >
-                      <TableIcon label={false} name="Workbench" />
+                      <TableIcon label name={dataset.uploadplan?.baseTableName ?? 'Workbench'} />
```
maxpatiiuk, line 244: FYI: `label` is equivalent to `label={true}`
before it was `label={false}`

after this change, the table name would be announced by screen reader (and have a tooltip) - before it wasn't because it would have been "Workbench" for all, thus useless

sounds good to me, just mentioning this to make sure this was intentional change

maxpatiiuk (approved):
as per the performance test you did, I am fine with the way this PR is (or with only including baseTableName)

btw, if you are curious, take a look at GraphQL (as opposed to REST APIs that we use) - with GraphQL, the front-end defines what data it needs, and the back-end sends just that data

specifyweb/frontend/js_src/lib/components/Toolbar/WbsDialog.tsx
grantfitzsimmons: That was my thought as I realized it would be something screen reader users might want to hear/see.

maxpatiiuk (approved):
👍

realVinayak:
Nice! Maybe extend that to batch image upload too? You can also get what the table is from the uploadplan there!

emenslin (changes requested):
Everything workbench wise looks good, I like the new icons. The last part of the testing checklist says to  test the same things on batch attachment uploader however there weren't any changes made there.

realVinayak:
funny that we have so multiple names for batch upload
batch attachment upload
batch image upload
bulk uploader (in the code)

grantfitzsimmons:
> Everything workbench wise looks good, I like the new icons. The last part of the testing checklist says to test the same things on batch attachment uploader however there weren't any changes made there.

I should have clarified that I did not intend for the icons to be included in the batch attachment upload tool, I just wanted to ensure no functionality was lost/broken due to these changes!

realVinayak (approved):

grantfitzsimmons:
image

There you go– your dreams have come true @realVinayak!

grantfitzsimmons:
_I'll fix the tests soon– no merging until that's complete!_

emenslin (approved):
Looks good in batch attachment upload now as well! :) 
Screenshot 2024-02-01 111602

specifyweb/frontend/js_src/lib/components/AttachmentsBulkImport/Datasets.tsx
```diff
@@ -189,11 +192,18 @@ export function AttachmentsImportOverlay(): JSX.Element | null {
...
-                    className="overflow-x-auto font-bold"
+                    className="font-bold"
                     href={`/specify/attachments/import/${attachmentDataSet.id}`}
                   >
+                    {attachmentDataSet.uploadplan?.staticPathKey === undefined ? (
+                      'Workbench'
+                    ) : (
+                      <>
+                        <TableIcon label name={staticAttachmentImportPaths[attachmentDataSet.uploadplan.staticPathKey].baseTable} />
+                      </>
```
maxpatiiuk, line 205:
```suggestion
                      <TableIcon label name={staticAttachmentImportPaths[attachmentDataSet.uploadplan.staticPathKey].baseTable} />
```

maxpatiiuk (approved):

bronwyncombs (approved):
Excellent 🐸 🎩 

Screenshot 2024-02-13 at 8 26 13 AM
Screenshot 2024-02-13 at 8 25 37 AM

CarolineDenis (approved):

PR #4476, opened 2024-01-28
Use muted colors for taxon tiles

### LINKED ISSUE #1859: Use more muted colors for the taxon tiles
image

`testability` on the left, `7.7.0` on the right

This is subjective and my opinion. In light mode, the colors are very bold after the tiles were rewritten.

Please close if you believe I am wrong in this.

grantfitzsimmons:
Replaces the d3 scheme with a custom set of colors that matches the previous behavior

This has been requested a few times by users since our default taxon tiles colors changed.

Fixes #1859

## New Appearance

`issue-1859` on left, `v7.7` on right in advertisement screenshots

image
image

### Background Information

_This is simply an aesthetic change_

image

I tested all of the categorical color schemes out of the box included in d3 but none of them were satisfactory. I think Max chose the right one when it was updated, but the style that users have required

schemeSet2 (current)

schemeSet2 (current)

**Other default options:**

schemeAccent

schemeAccent

schemePaired

schemePaired

schemePastel2

schemePastel2

schemeSet3

schemeSet3

schemeTableau10

schemeTableau10

schemeSet1

schemeSet1

schemeCategory10

schemeCategory10

schemePastel1

schemePastel1

schemePastel2

schemePastel2

### Checklist

### Testing instructions

1. Configure the home screen to show Taxon Tiles
2. Verify the change appears in light mode
3. Verify the change appears in dark mode
4. Provide feedback on the color scheme in your review


specifyweb/frontend/js_src/lib/components/HomePage/taxonTileHelpers.ts
```diff
@@ -132,7 +125,20 @@ export function makeTreeMap(container: SVGElement, rawRoot: PairedNode) {
     .size([container.clientWidth, container.clientHeight])
     .round(true)(root);
 
-  const color = scaleOrdinal(schemeSet2);
+  const tileColors = [
```
maxpatiiuk, line 128: (optional) usually we define static arrays like this outside the function (idally right below or right above it)

reasons:
- it's a common convention in programming that config variables/options are at the top level of the file (top level does mean top of the file, but means outside any function/class)
- won't matter in this case, but by defining array inside the function, it's recreated again on each function call - when defined outside the function, it's only defined once (thus a bit more memory efficient) - doesn't matter in this case as the function is only called once, but still good to keep in mind

though, there is also a reason not move the array out of the function as then it's a bit father from where it's used in the code, thus making code a bit harder to follow

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/HomePage/taxonTileHelpers.ts
grantfitzsimmons: I think we should leave it inside the function since 
1. It is called once
2. Makes it easier to follow in the code
but I will keep this in mind going forward. thank you for sharing your feedback! 😄

emenslin (approved):
Looks good! I feel like the color schemes in both light and dark mode blend in a lot better and are less jarring now. I feel like light mode might be a little bright with the white borders, however I am normally in dark mode so that's probably just a me thing.

pashiav (approved):
It looks great in dark mode! In light mode, I feel that it is hard to see some of the small tiles because of the light colored border and the width of it. 

Aesthetically though, the colors look good! It matches the Specify color palette.

bronwyncombs:
Colors are pleasing and compliment the Specify brand well. 

Tiles are visible/distinct in both light and dark mode, although slightly more clear in light mode, IMO. 

`dark mode:`
Screenshot 2024-01-31 at 1 25 30 PM
`light mode:`
Screenshot 2024-01-31 at 1 23 15 PM

Congrats on a great idea turned successful PR @grantfitzsimmons!

PR #4480, opened 2024-01-30
Add 'Upload Plan' button to the Tools menu, allow importing/exporting data set mappings

### LINKED ISSUE #1363: Add ability to export/import data set mappings
This would be very helpful to have access to in the edit menu for the WB data sets.

grantfitzsimmons:
While this is not a full implementation of an import/export utility, this makes the upload plan dialog available to regular users

Fixes #1363

This adds an 'Upload Plan' button that lets the user see the full JSON upload plan. If the user shares this mapping, another user can import it by pasting it in the dialog and clicking 'Save'. This exposes the raw JSON plan that was previously only available in development (or by removing the class in dev tools). This is something that would greatly improve some of CSIRO's workflows.

image

### Checklist

### Testing instructions

1. Open a WorkBench data set
2. Create a mapping
3. Edit in grid view
4. Click 'Tools'
5. Click 'Upload Plan'
6. Verify it can be copied/updated and import into another data set

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/WorkBench/Template.tsx
+++ b/specifyweb/frontend/js_src/lib/components/WorkBench/Template.tsx
@@ -114,16 +114,6 @@ function WbView({
           {commonText.tools()}
         </Button.Small>
         <span className="-ml-1 flex-1" />
-        {/* This button is here for debugging only */}
-        <Button.Small
-          className={`
-            wb-show-plan
-            ${process.env.NODE_ENV === 'development' ? '' : 'hidden'}
-          `}
-          onClick={f.never}
-        >
-          [DEV] Show Plan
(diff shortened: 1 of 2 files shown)


maxpatiiuk (approved):
long overdue change - thanks for doing this
❤️

pashiav (approved):
As a user, the button label being "Upload Plan" makes me expect that clicking on it will only involve uploading a plan into this dataset. Maybe renaming it to something like "Edit Upload Plan" or "View or Upload Plan" would clarify its function better.

Verified copying/uploading to other data sets. Functionality looks good!

realVinayak (approved):

grantfitzsimmons:
image

There you go– your dreams have come true @realVinayak!

PR #2796, opened 2023-01-12
New library for XML<-->JSON two way sync / xml-editor

maxpatiiuk:
EDIT:
Because this PR is getting long, I created a milestone to track remaining issues: https://github.com/specify/specify7/milestone/49
If you discover an new issue, please add it to that milestone

---

Begins fixing https://github.com/specify/specify7/issues/1156
Begins fixing https://github.com/specify/specify7/issues/884
Fixes https://github.com/specify/specify7/issues/1961

Create a library that parses XML into JSON, and in the process validates it and makes it type-safe. In addition, on any changes to JSON, the original XML is automatically modified to reflect the changes. At the same time, unsupported attributes, unsupported child types, and comments are preserved in the original XML.

Then, use this library to implement UI for edting Data Object Formatters and Aggergators.
This paves the way for adding UIs for all other XML resources, including form definitions

### Formatters

Open `DataObjFormatters` app resource (if there are multiple in the database, open any of them - you can easily find all such files by disabling all but `Record Formatters` in the app resource filters).
If there is a syntax error in XML, and you go to "Visual Editor" tab, it should tell you what the error is.
Otherwise:
- Before making any edits, download the XML file. Test viewing/creating/deleting formatters/aggregators. After making the edits, download the XML file again. Reformat both files (i.e, using a tool like https://www.freeformatter.com/xml-formatter.html#before-output). Then, compare the files (i.e, using a tool like https://www.diffchecker.com/text-compare/). Make sure that the XML file was edited in a way you would expect (it should reflect the edits you made in the UI + if there were some required attributes missing, they should be added).
- After making edits in sp7, try making edits in sp6 (check sp6 can still read the formatters). Then, open it again in sp7 to see if things still work.
- On any edits, the preview section of the editor should update. You should also be able to select custom elements for preview (both in formatter and aggregator editors)
- Try making more than one formatter/aggregator a default - it should automatically unset "Is Default" from the other formatter
- Test putting comments in the XML file (or commenting out some nodes) and making edits - the comments should be preserved in the XML. Similarly, try adding unknown arguments to elements (i.e `abc="test"`) or unknown elements (`<test ><a /></test>`) - they should all be preserved upon edits
- Test opening/closing full-screen mode. Test using back/forward browser buttons
- The `CollectingEventDetail` formatter in the `CollectingEvent` table should not be fully editable - test that an error message is displayed (it's not editable in sp6 either - it's yet another sp6 special hacky thing). There are about default 4 formatters like this - just testing this one is good enough
[...]

PR #4482, opened 2024-01-31
Fix incorrect usage of function to define preferences

melton-jason:
Caused by #4299 merging into `production`

### Checklist

### Testing instructions
- ensure Specify is accessible and the `Show query filters when running a Report from a Form` preference behaves correctly (see the testing instructions outlined in https://github.com/specify/specify7/pull/4299#issue-2044959670)

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
@@ -1625,7 +1625,7 @@ export const userPreferenceDefinitions = {
             defaultValue: false,
             type: 'java.lang.Boolean',
           }),
-          queryParamtersFromForm: defineItem<boolean>({
+          queryParamtersFromForm: definePref<boolean>({
             title: preferencesText.queryParamtersFromForm(),
             requiresReload: false,
             visible: true,


pashiav (approved):
Login page works!

maxpatiiuk (commented):
hmm, wasn't this caught by TypeScript?
or it because the error only occurred after the branch was merged with upstream changes?

grantfitzsimmons:
>or it because the error only occurred after the branch was merged with upstream changes?

🥇

melton-jason:
Yes, this was not caught by Typescript because the preference was added in #4299 (hence the change was automerged into production and no merge conflicts were raised). 

More specifically, #4299 was created before the merge of `xml-editor` (#2796) into production, and thus #4299 was using a differently named function to define preferences after `xml-editor` was merged into `production`.
This problem would have been caught if `production`, after the merge of `xml-editor`, was merged back into #4299 before it was merged itself into `production`. 

This PR is a cautionary tale- and concrete example- about issues which could arise with automerging. To avoid this in the future we would need to ensure that branches about to be merged with production (or any other branch) are as up to date as possible so issues such as this do not arise.

realVinayak:
Maybe I am grossly misunderstanding, correct me!

> This problem would have been caught if production, after the merge of xml-editor, was merged back into https://github.com/specify/specify7/pull/4299 before it was merged itself into production.

Hmm are you sure it wasn't caught? See https://github.com/specify/specify7/actions/runs/7728517442/job/21072324139. Go to "Run JS test quite". 

> (hence the change was automerged into production and no merge conflicts were raised)

Why would that deter typescript checks?

maxpatiiuk:
> Why would that deter typescript checks? Since typescript is ran regardless on merges, mistakes like this should always be caught by typescript (ofc it is up to us to look at it)

@realVinayak this branch was based on an older version of xml-editor, back when there was a function called `defineItem`, and this branch used that - all valid, no typescript errors
then, the function was renamed to `definePref` on xml-editor - but this branch didn't receive the new xml-editor changes
then this branch is merged into production - and only at that point typescript error occurring because `defineItem` no longer exists

> This PR is a cautionary tale- and concrete example- about issues which could arise with automerging. To avoid this in the future we would need to ensure that branches about to be merged with production (or any other branch) are as up to date as possible so issues such as this do not arise.

I wouldn't be against using auto merge as it's convenient, and hasn't caused issues like this often in the past - and the fact that typescript catches it after merge, and sends an email to the PR author means the impact should be small

I guess the real lesson is to try to merge PRs quicker
(I personally am guilty of having some PRs open for so long that the changes in them have became quite obsolete by the time I'm ready to merge it 😁😢)

realVinayak:
> the fact that typescript catches it after merge, and sends an email to the PR author means the impact should be small

That's what I meant by catching it!

PR #4487, opened 2024-02-04
refactor(TypeScript): make JSON.parse() safer

maxpatiiuk:
JSON.parse() returns `any` by default
This means, TypeScript checking is completely disabled for all the places where the return of JSON.parse() is used - this could be dangerous
Instead, this PR:
- Changes the return type of `JSON.parse()` to be `unknown`
- Which in turn, requires explicit type cast in places where `JSON.parse` is used. A type cast is still much safer than `any`, as at least the usages of this value are checked against what's described in the type
- For greater security, to actually ensure that the json string from the back-end matches a pre-defined schema (and thus remove the need for type assertion), a library like `zod` can be used, but that might be overkill for now

[Great talk that goes into this deeper](https://portal.gitnation.org/contents/the-lies-we-tell-ourselves-using-typescript)

### Checklist

### Testing instructions

Type only changes. No code logic impact

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Preferences/BasePreferences.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Preferences/BasePreferences.tsx
@@ -88,7 +88,7 @@ export class BasePreferences<DEFINITIONS extends GenericPreferences> {
   /**
    * Fetch preferences from back-end and update local cache with fetched values
    */
-  async fetch(): Promise<ResourceWithData> {
+  public async fetch(): Promise<ResourceWithData> {
     const entryPoint = await contextUnlockedPromise;
     if (entryPoint === 'main') {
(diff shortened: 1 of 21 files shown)


specifyweb/frontend/js_src/lib/components/Preferences/BasePreferences.tsx
```diff
@@ -455,7 +459,7 @@ const fetchResourceData = async (
 const fetchDefaultResourceData = async (
   fetchUrl: string,
   defaultResourceName: string
-): Promise<ResourceWithData> =>
+): Promise<Partial<ResourceWithData>> =>
```
maxpatiiuk, line 462: not exactly a bug, but the presence of `any` masked an invalid type here (return type is `Partial<ResourceWithData>` not `ResourceWithData`

PR #4495, opened 2024-02-07
Implement delete all notifications API

### LINKED ISSUE #4494: Delete All Notifications API
Currently the front-end calls the delete notification API multiple times.  It would be better to just have an API endpoint for the the front-end to just call once.  This is not completely necessary, but is a good idea for a good first issue for new programmers learning the basics of the front-end and back-end.

The "/notifications/delete_all" URL will be created that will delete all the notifications in the Message table associated with the current user.  Return "OK", just the the current delete API.

On the front-end, change the code the loops through the notification messages and calls to the delete API.  Replace the code with a single call to the delete_all API.

Good Luck! 😁

pashiav:
Fixes #4494 

### Checklist

- [x] Add automated tests

### Testing instructions

1. Get notifications to pop up (ex: merge agents)
2. Click "Clear All" and verify that all notifications are cleared


melton-jason (changes requested):
Nice work! 

If you want to write an automated test for this endpoint, you can define a new method (Django would expect it to start with `test_`) on the `NotificationsTests` class, similar to how the `test_get_notification_with_param_since` method is defined:
https://github.com/specify/specify7/blob/2425eb5577a0da49ce19a7a6e89c4f5eaf82bfd5/specifyweb/notifications/tests.py#L11-L36

If you have the Specify running through Docker, you can run the following command to start tests: 
```bash
sudo docker exec -it specify7-specify7-1 bash -c "ve/bin/python3 manage.py test"
```
(assuming the folder containing the `docker-compose.yml` is called `specify7`)

if you specifically only want to run notification tests:
```bash
sudo docker exec -it specify7-specify7-1 bash -c "ve/bin/python3 manage.py test specifyweb.notifications.tests.NotificationsTests"
```

acwhite211 (approved):

specifyweb/frontend/js_src/lib/components/Notifications/Notifications.tsx
```diff
@@ -46,15 +46,20 @@ export function Notifications({
...
     if (notifications === undefined) return;
-    notifications.forEach((notification) => {
-      ping('/notifications/delete/', {
-        method: 'POST',
-        // eslint-disable-next-line @typescript-eslint/naming-convention
-        body: formData({ message_id: notification.messageId }),
-        errorMode: 'dismissible',
-      });
+
+    // Map over the notifications array and get the messageId from each notification
```
maxpatiiuk, line 50: there is no need to comment each line line with detailed on what's happening
generally, leave comments when:
- you need to explain why some code is needed rather than what it is doing
- when what it is doing is complicated/no obvious from looking at it (though in those cases, better naming the variables and refactoring code into smaller functions often removes the need for the comments)

maxpatiiuk (approved):
looks good! 💎

specifyweb/notifications/tests.py
```diff
@@ -34,4 +34,30 @@ def test_get_notification_with_param_since(self):
...
+        # Send a POST request to the delete_all endpoint with the IDs of the test messages
+        response = c.post('/notifications/delete_all/', {'message_ids': json.dumps([testMessage1.id, testMessage2.id])})
+
+        # Check if the response is OK
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.content.decode(), 'OK')
 
+        # Check if the messages have been deleted
+        self.assertFalse(Message.objects.filter(id=testMessage1.id).exists())
+        self.assertFalse(Message.objects.filter(id=testMessage2.id).exists())
```
melton-jason, line 63: (optional) 
If you want to only hit the database once after the API has been called, this can be simplified to 

```suggestion
        self.assertFalse(Message.objects.filter(id__in=[testMessage1.id, testMessage2.id]).exists())
```
If you wanted to make this approach even better, you could further extract `[testMessage1.id, testMessage2.id]` into a separate variable and use it in the `self.assertFalse` and when making the request to the endpoint https://github.com/specify/specify7/blob/360a1832cadf787ff37d15e98f4f671b9a23e6e5/specifyweb/notifications/tests.py#L55

(Extracting the ids into a variable would be helpful if this test would ever need to be changed. Instead of modifying every place which needs the ids, only the variable would need to be modified. Additionally, it makes the code more readable. Places which use the message id array don't need to know excactly which ids they are using, just that they are using the array of message ids, i.e., `messageIds` over `[testMessage1.id, testMessage2.id, ...]`)

melton-jason (approved):
Nice work! 
One very minor improvement could be made, although it is not a blocker for this PR. 

Now we need an approving review from @specify/ux-testing before merge

CarolineDenis:
@pashiav 
Be careful to verify that you have all needed approvals before merging a branch in production. This has not be tested by the UX-UI team

PR #4498, opened 2024-02-09
Robust Back-End Filter by Collection Function

### LINKED ISSUE #3564: `xml-editor`: Hierarchy Exception when defining formatters and aggregators for some tables.

acwhite211:
Fixes #3564 

This PR attempts to make the back-end's filter_by_collection more robust.
This solution introduces a set of dictionaries that explicitly show the relationship field path from any given table to a collection so that it may be filtered by a collection.
This is done in order to prevent convoluted logic, allow for simple understanding, be easily refactored/modified/verified, and can be copied to the front-end in the future with less chance of introducing differences and bugs.
I had been playing around more complex solutions using lots of logical rules, but found this more explicit way of implementation to be much more robust and verifiable.
There is also a parameter option in the function to use an explicit dictionary lookup for the filter query of a given table.

This PR has the initial intention of solving issue #3564.
The issue might be somewhat solved by setting the domain filter to false, which might be a good solution for now, but something like this might be a good idea for the future.

The unit test for this new function tests every specify table, so feel free to look at the paths set as "correct" for each table and make edits.

While there are still a handful of tables without any links to the collection table, and thus cannot be filtered by collection, this solution has a lot less tables than the previous solution that would raise a Hierarchy Exception.

The current code got a little more complicated than originally intended, more simplification can probably be achieved with another pass through.
Still some improvements to make, but don't have the time to do it right now.

Still working on finishing the unit tests for each table, as well as code clean up and abstractions, but wanted to make a draft PR now.


acwhite211:
I created the #4509 PR as a time sensitive fix.  This PR still needs some more work, so I'm going to keep it as a draft for now and work on it when I get a chance.

maxpatiiuk:
Hmm, there are a lot of hardcoded dictionaries in this PR - can some of that be computed dynamically?
i.e front-end has [this function](https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/components/DataModel/specifyTable.ts#L470-L538) that figures scoping for most tables

[see results of that function on every table](https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/__snapshots__/specifyTable.test.ts.snap#L1248) - nothing is hardcoded, and everything is computed automatically even if we update the datamodel

realVinayak:
> I went through each table and found all the possible relationship field paths that lead to the collection table, and then kept the reasonable ones. I went with a more static approach because after trying the fully dynamic solution, the logic got pretty convoluted and unreadable.

Hmm do you remember which cases are different than in the snapshot? Seems more reasonable to still use the dynamic solution, but only hardcode divergent cases.

PR #4516, opened 2024-02-10
fix(Formatters): correctly detect main table fields

### LINKED ISSUE #4512: When creating a record using a query combo box the entered value is deposited in the wrong field
**Describe the bug**
For example, when searching for a name of an agent, the user can 'Add' a new agent and the searched term is deposited in the `lastName` field of the form to make it easier to create a new agent. 

**Expected behavior**
On `edge`, it puts it in the `firstName` field, which is not desired.
On `v7.9.3`, it puts it in the `lastName` field, which is the desired behavior.

**To Reproduce**

1. Go to a data entry form
2. Type a value into an agent query combo box
3. Click "add"
4. See that it puts it in the `firstName` field instead of the `lastName` field

**Screen Recording**
`edge` left, `v7.9.3` right

https://github.com/specify/specify7/assets/37256050/4fbf374f-3b03-4991-8b20-73ac7e60b23d

maxpatiiuk:
Fixes #4512

### Checklist

- [x] Add automated tests

### Testing instructions

- [ ] Make sure #4512 is fixed
- [ ] Take a look at https://github.com/specify/specify7/blob/23450dd16261c26c6f8917544fde7dcf9e43e7b3/specifyweb/frontend/js_src/lib/components/Formatters/__tests__/__snapshots__/formatters.test.ts.snap#L2033 and see if the list of "main fields" for each table makes sense (the list is determined automatically based on set of rules)


specifyweb/frontend/js_src/lib/components/Formatters/formatters.ts
```diff
@@ -278,7 +278,7 @@ export const getMainTableFields = (tableName: keyof Tables): RA<LiteralField> =>
   genericTables[tableName].literalFields
     .filter(
       ({ type, isRequired, isHidden, isReadOnly }) =>
-        type === 'java.lang.String' && !isReadOnly && (isRequired || isHidden)
+        type === 'java.lang.String' && !isReadOnly && (isRequired || !isHidden)
```
maxpatiiuk, line 281: don't know how I managed to mess this up 🤦

grantfitzsimmons (approved):
### Testing instructions

- [X] Make sure #4512 is fixed
- [X] Take a look at https://github.com/specify/specify7/blob/23450dd16261c26c6f8917544fde7dcf9e43e7b3/specifyweb/frontend/js_src/lib/components/Formatters/__tests__/__snapshots__/formatters.test.ts.snap#L2033 and see if the list of "main fields" for each table makes sense (the list is determined automatically based on set of rules)

I can't say I completely understand the logic that determines this list, but it seems to be _more_ logical? Some don't make much sense to me (some tables have a series of generic text fields as the 'main fields') but it behaves as I'd expect in the application. I'd like another UX tester to verify the behavior is consistent across a number of QCBXes with production.

CarolineDenis (approved):

bronwyncombs (approved):
The issue of new agents filling firstName field has been resolved, the qcbx correctly fills the lastName field on the form again. 

> I'd like another UX tester to verify the behavior is consistent across a number of QCBXes with production.

I also don't fully understand this list, but things appear to be mostly correct and logical. 
I tested various parts of the application including Geography, Repository Agreement, Storage, etc.

maxpatiiuk:
For future reference, here is a human-readable translation of the algorithm that determines the main table fields:
1. For a given table, pick all string fields that are not read-only, AND that are either required or not hidden
2. Sort the given fields according to the following:
    - if field name includes `name`, put those first
    - else, if field is required
    - else, field that is not virtual, and does not have "read-only" or "hidden" front-end overrides (front-end overrides are a concept, mainly applicable to workbench that hides fields or modifies schema, on top of the settings in the schema configuration)
    - else, sort by name

The result is a sorted list of fields from most important to least.
Than, any place that needs X most important fields, can pick X top items from the resulting list.

In the case of query combo box, it picks the first "main field" that is specified in the searchfields attribute in the typesearches.xml
Other places might pick first 2-3 main fields (for example, front-end has capability to generate record formatter on the fly if it's missing)

PR #4528, opened 2024-02-13
Display query header inline when distinct selected

### LINKED ISSUE #4514: Query headings are stacked vertically
**Describe the bug**
Run a query and see that the column headings are stacked vertically

**To Reproduce**

1. Run this query (https://umherb11024-edge.test.specifysystems.org/specify/query/168/)
2. See problematic behavior

**Expected behavior**
Not that!

https://github.com/specify/specify7/assets/37256050/84d320fe-eb98-4ee7-9407-22771b9a3f6b

CarolineDenis:
Fixes #4514

### Checklist

### Testing instructions

run a query with "Distinct" selected, verify headers in the results table are in line
run a query with "Distinct" not selected, verify headers in the results table are in line

Screenshot 2024-02-13 at 8 09 27 AM

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
@@ -279,7 +279,7 @@ export function QueryResults(props: QueryResultsProps): JSX.Element {
         role="table"
         style={{
           gridTemplateColumns: [
-            ...new Array(metaColumns).fill('min-content'),
+            ...Array.from({ length: metaColumns }).fill('min-content'),
             ...Array.from({ length: visibleFieldSpecs.length }).fill('auto'),
           ].join(' '),
         }}
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
@@ -274,7 +274,12 @@ export function QueryResults(props: QueryResultsProps): JSX.Element {
           overflow-auto rounded
           ${tableClassName}
           ${showResults ? 'border-b border-gray-500' : ''}
-       `}
+          ${
+            hasIdField
+              ? 'grid-cols-[min-content_min-content_repeat(var(--columns),auto)]'
+              : 'grid-cols-[repeat(var(--columns),auto)]'
+          }
+        `}
         ref={scrollerRef}
         role="table"
         style={
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
@@ -181,6 +181,7 @@ export function QueryResults(props: QueryResultsProps): JSX.Element {
     'appearance',
     'showLineNumber'
   );
+  const metaColumns = (showLineNumber ? 1 : 0) + (hasIdField ? 2 : 0);
 
   return (
     <Container.Base className="w-full !bg-[color:var(--form-background)]">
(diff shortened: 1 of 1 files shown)


emenslin (approved):
Looks good, query is working normally.
Screenshot 2024-02-13 112804

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -274,7 +274,12 @@ export function QueryResults(props: QueryResultsProps): JSX.Element {
           overflow-auto rounded
           ${tableClassName}
           ${showResults ? 'border-b border-gray-500' : ''}
-       `}
+          ${
```
maxpatiiuk, line 277: not a correct solution (not commenting on whether it works or not)
can you spot the issue? look closely at some of the other classes AND props on this div

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
CarolineDenis: I forgot to remove the other class name that contained already the grid but I think the solution I have is the one that works. Any other suggestion?

maxpatiiuk: thanks for the clarification

a few comments:
- the code no longer handles the `showLineNumber` case (see the `--meta-columns` CSS variable declared a ew lines above, and that variable was used in the class name before) resulting in a bug in cases when that pref is set
- given that `--meta-columns` is no longer used, it should be removed

it seems like the real issue is that repeat() CSS function does not allow `0` as the first value - https://developer.mozilla.org/en-US/docs/Web/CSS/repeat#:~:text=It%20is%20specified%20with%20an%20integer%20value%20of%201%20or%20more
and when that value happens to be 0, it breaks the grid-cols, causing this issue

thus, I would instead suggest you to add a style like this (rather than a tailwind class name as it's getting too complex for tailwind):

```ts
const metaColumns = (showLineNumber ? 1 : 0) + (hasIdField ? 2 : 0);

// later:
style={{
  gridTemplateColumns: `${metaColumns === 0 ? '' : `repeat(${metaColumns}, min-content) `}repeat(${visibleFieldSpecs.length}, auto)`;
}}
```

or alternative version (using js fill rather than CSS repeat):

```ts
const metaColumns = (showLineNumber ? 1 : 0) + (hasIdField ? 2 : 0);
style={{
  gridTemplateColumns: [...Array.from({length:metaColumns}).fill('min-content'), ...Array.from(visibleFieldSpecs).fill('min-content')].join(', ');
}}
```

and then --columns and --meta-columns CSS variable should be removed

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -270,19 +271,18 @@ export function QueryResults(props: QueryResultsProps): JSX.Element {
...
         role="table"
-        style={
-          {
-            '--columns': visibleFieldSpecs.length,
-            '--meta-columns': (showLineNumber ? 1 : 0) + (hasIdField ? 2 : 0),
-          } as React.CSSProperties
-        }
+        style={{
+          gridTemplateColumns: [
+            ...new Array(metaColumns).fill('min-content'),
```
maxpatiiuk, line 282: don't use new Array please (we should have an eslint-rule against that)
it's bug prone - and Array.from was added to javascript to replace it - `Array.from({length: metaColumns})`
it's bug prone because `Array.from(2)` creates array with two items, but `Array.from(2, 1)` creates an `[2, 1]` array

bronwyncombs:
Good screenshot to compare this fix:

Screenshot 2024-02-19 at 12 35 33 PM

maxpatiiuk (approved):
this PR shows that some things a better solved with JS than CSS)

I am really surprised that `repeat()` has such a horrible behavior when 0 passed in - whu couldn't it just repeat 0 times??? anyway - thanks for fixing this

PR #4529, opened 2024-02-13
Remove unecessary skeleton loader from QueryListDialog

### LINKED ISSUE #4518: Statistics page shows extra queries dialog while loading queries
**To Reproduce**

1. Go to the stats page
2. Go to add a new query (click **Add**) in the bottom right corner of a category
3. See an extra queries dialog appear while the queries are loaded
4. Once they have loaded, see that it disappears

**Expected behavior**
There should only be one dialog

**Screenshots**
image

https://github.com/specify/specify7/assets/37256050/cd49431f-50cd-4726-9b25-271e465f789f

`edge` on left, `v7.9.3.1` on right

CarolineDenis:
Fixes #4518

### Checklist

### Testing instructions

Go to the stats page
Go to add a new query (click Add) in the bottom right corner of a category
See that there isn't an extra queries dialog appear while the queries are loaded

NOTES: 
Since the refactoring of QueryListDialog component the skeleton loader is not necessary anymore.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Toolbar/Query.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Toolbar/Query.tsx
@@ -31,6 +31,8 @@ import { QueryEditButton } from '../QueryBuilder/Edit';
 import { OverlayContext } from '../Router/Router';
 import { SafeOutlet } from '../Router/RouterUtils';
 import { QueryTablesWrapper } from './QueryTablesWrapper';
+import { DialogListSkeleton } from '../SkeletonLoaders/DialogList';
+import { f } from '../../utils/functools';
 
 export function QueriesOverlay(): JSX.Element {
   const handleClose = React.useContext(OverlayContext);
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/Toolbar/Query.tsx
```diff
@@ -125,101 +124,90 @@ export function QueryListDialog({
 
   const isReadOnly = React.useContext(ReadOnlyContext);
 
-  return data === undefined ? (
-    <Dialog
```
maxpatiiuk, line 129: isn't this still needed for the other places where QueryListDialog is used though?
so instead of removing it, could you instead move it into the `defaultChildren` function?

CarolineDenis: I added the skeleton in the default children but not sure that's what you meant

maxpatiiuk: yes, that's what I meant - thanks

this way, by default, consumers of this component still get a loader, but they don't get a bug if they choose to provide a custom `children` function

maxpatiiuk (approved):

emenslin (approved):
Looks good, no more skeleton loader. 

https://github.com/specify/specify7/assets/150940840/e4e2a4d7-e9f6-4ad7-a60d-9f52ef383278

bronwyncombs (approved):
Working as expected! 
This branch vs `edge` on the right: 

https://github.com/specify/specify7/assets/135047322/c8f2e12a-0004-40c9-b6a6-31898579c6fe

PR #4535, opened 2024-02-14
refactor(Atoms): simplify time units

maxpatiiuk:
Having SECOND be 1000 rather than 1 simplifies the code logic and actually catches a few mistakes in the code

i.e before `setTimeout(callback, HOUR)` would actually mean every 3.6s rather than every hour because you, counterintuitive had to do `setTimeout(callback, HOUR * MILLISECONDS)` - this PR fixes that

### Checklist

### Testing instructions

- [x] @realVinayak could you please review the code, specifically, the Stats and Attachments code?
- [ ] Make sure the Bulk Attachment Upload progress bar works fine (same as it did before this PR)
- [ ] Make sure relative dates in the workbench datasets dialog display correctly (same as it did before this PR)


specifyweb/frontend/js_src/lib/components/Molecules/DateElement.tsx
```diff
@@ -40,12 +40,10 @@ function DateElementSafe({
   React.useEffect(() => {
     let timeout: NodeJS.Timeout | undefined;
 
-    function updateRelativeDate() {
+    function updateRelativeDate(): void {
```
maxpatiiuk, line 43: this function had a bug before actually, since it used `HOUR` rather than `HOUR * MILLISECONDS` before
this refactor would make things less confusing

CarolineDenis (approved):

specifyweb/frontend/js_src/lib/components/AttachmentsBulkImport/PerformAttachmentTask.ts
```diff
@@ -119,7 +118,7 @@ export function PerformAttachmentTask({
                 workRef.current.retrySpec[currentUploadingIndex] ??= 0;
                 const nextTry =
                   workRef.current.retrySpec[currentUploadingIndex];
-                workRef.current.retrySpec[currentUploadingIndex] += 1;
+                workRef.current.retrySpec[currentUploadingIndex] += SECOND;
```
realVinayak, line 121: revert back to += 1

I used time units wherever needed, so it is safe to assume that no place (in any of batch upload) hard codes time!

bronwyncombs (approved):
- [X] Make sure the Bulk Attachment Upload progress bar works fine (same as it did before this PR)

https://github.com/specify/specify7/assets/135047322/22f843ac-4392-4222-a336-a88c263d09e3

- [X] Make sure relative dates in the workbench datasets dialog display correctly (same as it did before this PR)
Screenshot 2024-03-13 at 10 21 59 AM

Issue #2912, opened 2023-01-27
Enforce scope properly for attachment and other tables


grantfitzsimmons:
https://github.com/specify/specify6/blame/8c853f91b76ff7633c8318aff5008e1822ca10a0/src/edu/ku/brc/specify/datamodel/Attachment.java

This is a mess, `collectionmemberid` integer fields are not the way to go

maxxxxxdlp:
Can confirm it's a giant mess.
TLDR: leave it be the way it is, and when sp7 is no longer imprisoned by sp6, reengineer the scoping from scratch.

Findings:
Some tables have direct relationship to scoping tables (CollectionObject, Collection, Discipline, Division, Institution). Scoping for those is enforced in sp6. Scoping for them is enforced in sp7 only in query results and when going to that record directly by the URL - not enforced in other places.
There is an exception for query builder. The query builder scopes to Division only for some tables. See https://github.com/specify/specify7/commit/f79e70cf19c07354f7e1f8ee5167e82fcf9b27ae

All of the below scoping is enforced in sp6 only using some custom logic for every case. None of this is implemented in sp7 (unless mentioned otherwise):
1. Children tables inherit scope from parent (Accession Authorization is scoped to same level as Accession)
2. Attachments are scoped using `scopeType` ([represents id of the](https://github.com/specify/specify6/blame/8c853f91b76ff7633c8318aff5008e1822ca10a0/src/edu/ku/brc/specify/datamodel/Attachment.java#L74-L78)) and `scopeId` fields (represents the ID of the record in that scope). Note, the `scopeId` may refer to the record that no longer exists as delete blockers don't check for those (since `scopeId` is just a numeric field.
3. `SpPermission` table stores sp6 permissions. It has it's own weird solution for scoping.
4. There is an `SpPrincipal` table. `RecordSet`, `AppResource`, `SpPermission`, `Workbench` tables have relationship to it. The table appears to be used for permissions or scoping or both.
5. The following tables have a `collectionMemberId` **integer** field:
    ```json
    [
      "Borrow",
      "BorrowAgent",
      "BorrowMaterial",
      "BorrowReturnMaterial",
      "CollectingEventAttachment",
      "CollectingEventAttr",
      "CollectingTripAttachment",
      "CollectionObject",
      "CollectionObjectAttachment",
      "CollectionObjectAttr",
      "CollectionObjectAttribute",
      "CollectionObjectCitation",
      "CollectionObjectProperty",
      "Container",
[...]

PR #4542, opened 2024-02-14
Display only record sets for the collection

### LINKED ISSUE #4520: Record sets appear in the wrong collection
**Describe the bug**
Record sets that should not appear are showing in the wrong collection

It shows:
`You do not have access to any Collection containing this resource through the currently logged in account`

**To Reproduce**

1. Log in as `petrifiedroadkill` and select "NHMA Vertebrate"
1. Go to https://fitznhmaoct2023-edge.test.specifysystems.org/specify/record-set/344/
2. See that there are record sets visible
3. Go to https://fitznhmaoct2023-v793.test.specifysystems.org/specify/overlay/record-sets/
4. See that there are no record sets visible

Here's an example of one of the record sets visible in `edge` but not in `v7.9.3.1`:

|RecordSetID|TimestampCreated|TimestampModified|Version|CollectionMemberID|AllPermissionLevel|TableID|GroupPermissionLevel|Name|OwnerPermissionLevel|Remarks|Type|SpPrincipalID|CreatedByAgentID|ModifiedByAgentID|SpecifyUserID|InfoRequestID|
|-----------|----------------|-----------------|-------|------------------|------------------|-------|--------------------|----|--------------------|-------|----|-------------|----------------|-----------------|-------------|-------------|
|344|2020-01-16 14:40:08.000|2020-01-16 14:40:08.000|0|32769||1||Aarhus-Ento_1_2020-1-16_14:28|||0||1|1|1||

See that the `CollectionMemberID` references the "NHMA Entomology" collection, which this user does not have access to anymore:

|collectionId|CollectionName|
|------------|--------------|
|32769|NHMA Entomology|

This indicates record sets are not being filtered by collection properly.

**Screen Recording:**

https://github.com/specify/specify7/assets/37256050/1a5e3ccc-e8c6-478b-9ebd-a4f7fdf91d73

CarolineDenis:
Fixes #4520

### Checklist

### Testing instructions

go to record sets dialog overlay 
verify that all record sets proposed are part of the collection you are connected in


maxpatiiuk (changes requested):
the new code does not call `mapValue()` entirely - is that intentional?

some parts in it are really important (for example `tableName === 'Attachment' || typeof scopingField === 'object'` fixes a bug

can you please make the necessary changes needed to fix this bug directly in the mapValue function rather than bypassing it?

specifyweb/frontend/js_src/lib/components/DataModel/collection.ts
```diff
@@ -107,7 +107,7 @@ function mapValue(
     return value === true &&
       (tableName === 'Attachment' || typeof scopingField === 'object')
       ? 'true'
-      : undefined;
+      : (value as string).toString();
```
maxpatiiuk, line 110: hmm, this doesn't look right
the condition above `value === true && (tableName === 'Attachment' || typeof scopingField === 'object')` makes the return of this function true if `value` is true AND tableName or scopingField match some value

after your code change, if that condition doesn't match the function would still return 'true' even if the tableName or scopingField condition didn't match (because of `(value as string).toString()`), so you essentially made all of this redundant: `value === true && (tableName === 'Attachment' || typeof scopingField === 'object') ? 'true' :`
(which is incorrect, as that code as there for a reason)

in the other case, when `!value`, your code `(value as string).toString()` would return `'false'`, but that should be treated by the back-end same as committing domainFilter (the undefined value) - if back-end treats those a different than that might be a back-end bug

so that I can better understand the changes you are making, can you explain to me what is the bug that is causing https://github.com/specify/specify7/issues/4520 and how you are trying to solve it? or, if you didn't yet find the cause, what you tried/whether you need help

maxpatiiuk (changes requested):

CarolineDenis:
NOTES: 
tests have been modify for test/try out reasons need to be reverted if code changes

.vscode/settings.json
```diff
@@ -0,0 +1,23 @@
+[
```
maxpatiiuk, line 1: I might be unfamiliar with vscode, but this file doesn't look like
isn't vscode settings file supposed to be a object, rather than an array with objects?

and while you are at it, you can also merge in my local sp7's settings.json with this one:
```json
{
  "typescript.tsdk": "./specifyweb/frontend/js_src/node_modules/typescript/lib",
  "terminal.integrated.cwd": "specifyweb/frontend/js_src/lib/",
  "prettier.prettierPath": "./specifyweb/frontend/js_src/node_modules/prettier",
  "todo-tree.tree.scanMode": "workspace only",
  "eslint.experimental.useFlatConfig": true,
  "eslint.workingDirectories": ["./specifyweb/frontend/js_src"],
  "search-node-modules.path": "/specifyweb/frontend/js_src/node_modules",
  "cSpell.words": ["Syncer"]
}
```

specifyweb/frontend/js_src/lib/components/DataModel/__tests__/collection.test.ts
```diff
@@ -52,7 +52,7 @@ describe('fetchCollection', () => {
     resource_uri: getResourceApiUrl('Locality', 1),
   };
   overrideAjax(
-    '/api/specify/locality/?limit=1&localityname=Test&orderby=-latlongaccuracy&yesno1=True',
+    '/api/specify/locality/?limit=1&localityname=Test&orderby=-latlongaccuracy&yesno1=True&domainfilter=false',
```
maxpatiiuk, line 55: this change should be reverted as `domainfilter=false` should be equivalent to not having that value
if it is not, then it's a back-end bug, that should be fixed on the back-end, not front-end

if nothing else, think of this - this change makes every single resource request in specify a few bites larger. and if there is a bug there where `&domainfilter=false` is not the same as omitting it, it should be fixed at the back-end to benefit people who are using the API directly (and we do have such)

maxpatiiuk (changes requested):

.vscode/settings.json
melton-jason, line 1: Is committing a Workspace VSCode settings file to `production` something we wish to do? 
I agree there might be some value in doing so, but it implies that everyone should use the same extensions when developing Specify. The most value I can see for doing so is for extensions like `prettier` and `ESLint`, but it may cause headaches for other developers if they have additional extensions installed, differently formatted, different settings, etc. because it would likely raise merge conflicts everytime they switch or checkout another branch. 

My own Workspace settings looks like this: 
```json
{
  "[typescript]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode",
    "editor.formatOnSave": true
  },
  "editor.wordWrap": "on",
  "editor.wordWrapColumn": 80,
  "typescript.tsdk": "./specifyweb/frontend/js_src/node_modules/typescript/lib",
  "eslint.experimental.useFlatConfig": true,
  "eslint.workingDirectories": [{ "pattern": "./specifyweb/frontend/js_src/" }],

  "jest.enable": false,
  "jest.runMode": {
    "runAllTestsOnStartup": true,
    "type": "on-save",
    "testFileOnly": true
  },
  "jest.jestCommandLine": "npm run unitTests --",
  "jest.rootPath": "./specifyweb/frontend/js_src/",

  "better-comments.multilineComments": true,
  "better-comments.tags": [
    {
      "tag": "FIXME",
      "color": "#FF2D00",
      "strikethrough": false,
      "underline": false,
      "backgroundColor": "transparent",
      "bold": true,
      "italic": false
    },
    {
      "tag": "BUG",
      "color": "#15a32d",
      "strikethrough": false,
      "underline": false,
      "backgroundColor": "transparent",
      "bold": true,
      "italic": false
    },
    {
      "tag": "FEATURE",
      "color": "#d47d2c",
      "strikethrough": false,
      "underline": false,
      "backgroundColor": "transparent",
      "bold": true,
      "italic": false
    }
  ],

  "sqltools.connections": [
    {
      "mysqlOptions": {
        "authProtocol": "default"
      },
      "previewLimit": 50,
      "server": "localhost",
      "port": 3306,
      "askForPassword": true,
      "driver": "MariaDB",
      "name": "Specify Container",
      "database": "specify",
      "username": "root"
    }
  ],
  "sqltools.format": { "linesBetweenQueries": 2 },
  "[python]": {
    "editor.defaultFormatter": "ms-python.autopep8",
    "editor.formatOnSave": false
  }
}
```

Regardless, if we do go ahead with the changes we would likely need to write documentation which at the very least lists out the Extensions used in the settings file. 

Alternatively, we can include the `.vscode` directory and all files/directories therein within the `.gitignore` file, and allow people to manage their own settings. 
Though even in this case, we should recommend Extensions such as Prettier and ESLint. These extensions should theoretically no longer be "required" as they were before the "Lint Code with ESLint and Prettier" was developed.

melton-jason (changes requested):
I have verified that there is no backend bug with regards to `domainfilter`. 

Omitting `domainfilter` from the parameters of the collection query returns the same results as being explicitly setting `domainfilter=false`. 
i.e., the following api calls are identical: 
```
http://localhost/api/specify/collectionobject/?catalognumber__lte=000000050

http://localhost/api/specify/collectionobject/?catalognumber__lte=000000050&domainfilter=false
```

Because of the reasons @maxpatiiuk mentioned in https://github.com/specify/specify7/pull/4542#discussion_r1496808544, whenever the frontend passes a `domainFilter: false` in the options of a `fetchCollection`, the API call made to the backend should omit the `domainFilter`. 
Alternatively, we can revert the changes made in https://github.com/specify/specify7/pull/3304 and keep the mapValue as is (that is, always set domainfilter to whatever the developer passed to `fetchCollection`, while still handling the attachment case). The changes to mapValue were likely made in #3304 not only because specifying `domainfilter=false` is the same as omitting the parameter, but also to be smart about fetching collections (and avoid HierarchyExceptions thrown by the backend. See https://github.com/specify/specify7/issues/3564)

The issue is that `domainFilter` is being set to true when fetching the RecordSets, but is set to undefined in mapValue: 
https://github.com/specify/specify7/blob/b226bae7e62984cf2813da4eef5fda2c126e928f/specifyweb/frontend/js_src/lib/components/Toolbar/RecordSets.tsx#L83-L87


The problem with Record Sets specifically is that the `getScope` and `getScopingRelationship` methods for `tables.RecordSet` return `undefined`, so the domainfilter is being omitted automatically in mapValue:

https://github.com/specify/specify7/blob/b226bae7e62984cf2813da4eef5fda2c126e928f/specifyweb/frontend/js_src/lib/components/DataModel/collection.ts#L105-L110

The `RecordSet` table only has the `collectionmemberid` field to determine which collection the RecordSet belongs to. 
https://files.specifysoftware.org/schema/version/2.10/#recordset

Thus we need to modify the frontend `getScope` method for a Table to handle such a case, modify the `mapValue` function to handle this case, or add the collectionmemberid as `advancedFilters` when fetching the Record Sets: 
```ts
fetchCollection(
          'RecordSet',
          {
            specifyUser: userInformation.id,
            type: 0,
            limit,
            domainFilter: true,
            ...
          },
          {
            collectionmemberid: schema.domainLevelIds.collection,
          }
        )
```
(and revert other changes made in this PR regarding `mapValue`, unless we decide to restore the implementation as it was before #3304)

.vscode/settings.json
maxpatiiuk: > Is committing a Workspace VSCode settings file to production something we wish to do?

I have vscode in my personal global gitignore so it doesn't get committed accidentally - https://github.com/maxpatiiuk/dotfiles/blob/main/git/.gitignore_global#L9

Whether to check in settings.json or not is a decision each repository has to make
pros:
- the settings are consistent between devs (everyone has eslint, prettier, same config for TODO tags, etc)
- improvements to the settings by one dev benefit all
cons:
- we are not expecting everyone to use vscode, so these changes won't apply to everyone anyway
- it might be overly restrictive to force the settings on everyone, as generally we have been relatively relaxed with what editors and settings each dev uses

so there are pros and cons each way. I would be fine with not making this decision at the moment, but, @CarolineDenis feel free to do a small write up on how to setup python debugger - feel free to write that in https://github.com/specify/specify7/tree/production/specifyweb/frontend/js_src#js_src or in https://github.com/specifysystems/code-principles

maxpatiiuk:
@melton-jason thanks for the findings and the clarity on the situation - much appreciated
do you know the conditions under which back-end throws HierarchyException/complains about needless domainfilter?

in my view, there are two options:
- make front-end always pass domainfilter true, and make back-end ignore domainfilter when it's not applicable for a given table
- add an `isScoped` method to SpecifyModel that would check whether a [getScope returns not undefined](https://github.com/specify/specify7/blob/b226bae7e62984cf2813da4eef5fda2c126e928f/specifyweb/frontend/js_src/lib/components/DataModel/specifyTable.ts#L470) or [table has some scoping fields](https://github.com/specify/specify7/issues/2912#issuecomment-1406828873) that aren't already considered by getScope

melton-jason:
The backend raises a HierarchyException when `domainfilter` is set to true and the table does not have a field in `['collectionmemberid', 'collectionobject', 'collection', 'discipline', 'division', 'institution']`

https://github.com/specify/specify7/blob/b226bae7e62984cf2813da4eef5fda2c126e928f/specifyweb/specify/filter_by_col.py#L36-L48

> make front-end always pass domainfilter true, and make back-end ignore domainfilter when it's not applicable for a given table

Assuming there are no unexpected edge cases, this would be a two-line change. The function supports a `strict` parameter which can be set to False to return the queryset without any scoping applied if no direct scoping field exists on the table. 
https://github.com/specify/specify7/blob/b226bae7e62984cf2813da4eef5fda2c126e928f/specifyweb/specify/filter_by_col.py#L17

We would just need to pass in `strict=False` as an argument to the function call in the following places:
https://github.com/specify/specify7/blob/b226bae7e62984cf2813da4eef5fda2c126e928f/specifyweb/specify/api.py#L803-L804
https://github.com/specify/specify7/blob/b226bae7e62984cf2813da4eef5fda2c126e928f/specifyweb/specify/api.py#L877-L878

However, this also modifies the behavior to users who are utilizing the API in the case that collection queries which once errored would now return results, even if `domainFilter` was set to true. 
For instance, the backend currently has no way to scope a table like `Disposal`, so it may be possibly confusing fetching a collection of Disposals would now return every Disposal even if domainFilter was set to true. 
This could possibly be alleviated by logging a message explaining the situation. 

> add an isScoped method to SpecifyModel that would check whether a [getScope returns not undefined](https://github.com/specify/specify7/blob/b226bae7e62984cf2813da4eef5fda2c126e928f/specifyweb/frontend/js_src/lib/components/DataModel/specifyTable.ts#L470) or https://github.com/specify/specify7/issues/2912#issuecomment-1406828873 that aren't already considered by getScope. then test that this matches what back-end expects

This may work, although the issue here probably lies in making the solution not complicated. 
There are already two scope-related functions for tables, and adding a third will not make it easier for future developers (us included) to make proper use of them. 
Although this complication may be warranted until we move away from Specify6 and attempt to homogenize ccoping across Specify. Scoping and how it is implemented currently poses a complicated and messy problem, and it is likely wishful thinking to assume there'd be a simple and clean solution.

CarolineDenis:
Fixed in: https://github.com/specify/specify7/pull/4576

melton-jason:
There are 13 total tables which are affected by this bug (have a `collectionmemberid` field and table.getScope() returns undefined):
- Borrow
- BorrowAgent
- BorrowMaterial
- BorrowReturnMaterial
- CollectionObjectAttribute
- Container
- InfoRequest
- PreparationAttribute
- Project
- RecordSet
- SpExportSchemaMapping
- SpFieldValueDefault
- SpSymbiotaInstance

maxpatiiuk:
> However, this also modifies the behavior to users who are utilizing the API in the case that collection queries which once errored would now return results, even if domainFilter was set to true.

turning something that previously errored into a non-error is not necessarily a breaking change

> For instance, the backend currently has no way to scope a table like Disposal, so it may be possibly confusing fetching a collection of Disposals would now return every Disposal even if domainFilter was set to true.

okay, I agree with this

> There are already two scope-related functions for tables, and adding a third will not make it easier for future developers (us included) to make proper use of them.

when the choice is between bugs and some complication, probably complication is the lesser evil
though, among the affected tables, I guess only RecordSet, Container and Borrow could theoretically be fetched using fetchCollection - the others would be visible on the form instead
there are no used by sp7 at all: SpExportSchemaMapping, SpFieldValueDefault, SpSymbiotaInstance

so I am fine with adding isScoped() or not doing anything for now, but fixing up scoping (and making it more customizable? 🥺) is a must as soon as sp6 is deprecated.
cc @CarolineDenis

PR #4543, opened 2024-02-14
Remove the prompt field option from weblinks

### LINKED ISSUE #4525: Web link behavior has changed– prompt now reads field as text
**Describe the bug**
On `v7.9.3`, this web link immediately took the user to the URL part + field value.
Now on `edge` this web link prompts the user to enter a value for `altCatalogNumber`, which is not resolved to a field name nor taking the value of the field.

image

```xml
  <weblinkdef>
    <name><![CDATA[RBGELivingCat]]></name>
    <desc><![CDATA[]]></desc>
    <baseURLStr><![CDATA[https://data.rbge.org.uk/living/<altCatalogNumber>]]></baseURLStr>
    <args>
      <weblinkdefarg>
        <name><![CDATA[altCatalogNumber]]></name>
        <prompt><![CDATA[true]]></prompt>
        <isEditable><![CDATA[false]]></isEditable>
      </weblinkdefarg>
    </args>
    <usedByList/>
  </weblinkdef>
```

([WebLinks resource](https://herbrbge252024-edge.test.specifysystems.org/specify/resources/app-resource/25/web-link/))

Fundamentally, we need to decide what the behavior of "Prompt field" is. Is it 

**To Reproduce**

1. Go to https://herbrbge252024-edge.test.specifysystems.org/specify/view/collectionobject/35846/?recordsetid=650
2. Scroll down until you see the field "RBGE Living Collection Accession No."
3. Click on the web link button next to the field
4. See the described behavior

**Expected behavior**
It should link out to the site as it did before. 
If `<prompt><![CDATA[true]]></prompt>` is configured (as it is here), it should at least retain the value from the specified field _and_ show the appropriate caption for the field, not just the name of the field entered in the 'Prompt field' definition.

image

This does accurately recreate the behavior in Specify 6:

https://github.com/specify/specify7/assets/37256050/af9592bc-47eb-413d-8fa1-1c72dfd78df5

image

**Screenshots**
image

You are shown `altCatalogNumber` verbatim (as it is interpreted as a text string, and `xml-editor` accepts any string as it assumes this is a label) rather than it being interpreted as a field name.

https://github.com/specify/specify7/assets/37256050/009c523a-54c4-4a56-bbe6-76daed0574f3

CarolineDenis:
Fixes #4525

### Checklist

### Testing instructions
 go to weblink editor 
verify that the feature "Prompt Field" is not available anymore
verify that the other features are still working


CarolineDenis:
@maxpatiiuk, 
I started to comment out the code related to the prompt field feature (I will remove it of course) 
Can you check if it seems coherent already? Is there places I missed? 
I was thinking that we should change the code for `function Part` in WebLinks/Definition.tsx because of the lines: 
https://github.com/specify/specify7/blob/d2ca1fd1a62bd14851c569cd93d26e63cfd803a2/specifyweb/frontend/js_src/lib/components/WebLinks/Definition.tsx#L182C9-L187C11

What do you think? 
I will also add a comment in the Weblink component with a link to this issue/pr

maxpatiiuk (changes requested):
even if we decide to remove the prompt field from the UI, you shouldn't remove it from the XML spec so that Specify 7 can still read the existing sp6 XML correctly and not corrupt it on changes

specifyweb/frontend/js_src/lib/components/WebLinks/spec.ts
```diff
@@ -161,38 +159,39 @@ function parseField(item: RawWebLink, part: string): ParsedWebLink {
...
 ): Pick<RawWebLink, 'parameters' | 'url'> {
   const augmented = parts
     .filter((part) => part.type !== 'Field' || part.field.length > 0)
-    .map((parameter, index) =>
+    .map((parameter) =>
       parameter.type === 'Field'
         ? {
             name: parameter.field.map(({ name }) => name).join('.'),
             title: parameter.field.map(({ label }) => label).join(' > '),
             shouldPrompt: false,
```
maxpatiiuk, line 180: you should remove all references to `shouldPrompt` in this function - since sp7 won't be supporting prompt fields anymore, it should not be modifying the value of `shouldPrompt`, but just leave the way it is in the XML to not break sp6

maxpatiiuk (changes requested):

maxpatiiuk:
> even if we decide to remove the prompt field from the UI, you shouldn't remove it from the XML spec so that Specify 7 can still read the existing sp6 XML correctly and not corrupt it on changes

disregard this comment - I wrote it before I looked at the code to remember how things work.

we can totally remain compatible with sp6 without any breaking changes by just declaring shouldPrompt prop the way it is in the args spec, and then removing every other usage of shouldPrompt (every place where we were reading the value of shouldPrompt, or trying to overwrite the value of shouldPrompt)

specifyweb/frontend/js_src/lib/components/WebLinks/index.tsx
```diff
@@ -59,13 +52,9 @@ export function WebLinkField({
   const [builtUrl, setUrl] = React.useState<
     RA<string | { readonly prompt: string }> | undefined
```
maxpatiiuk, line 53:
```suggestion
  RA<string | undefined
```

can you do a global case-insenstive search on "prompt" just in case we missed any other places?

specifyweb/frontend/js_src/lib/components/WebLinks/spec.ts
```diff
@@ -91,7 +89,10 @@ const argumentSpec = f.store(() =>
       syncers.maybe(syncers.xmlContent),
       syncers.default(localized(''))
     ),
-    // Specify 7 only
+    /**
+     *  Specify 7 only, usage of this feature have been
+     * removed in https://github.com/specify/specify7/issues/4525
```
maxpatiiuk, line 94:
```suggestion
     * Specify 6 only. It was temporary implemented in Specify 7, but was removed
     * in https://github.com/specify/specify7/issues/4525
```

specifyweb/frontend/js_src/lib/components/WebLinks/spec.ts
```diff
@@ -161,48 +161,29 @@ function parseField(item: RawWebLink, part: string): ParsedWebLink {
       type: 'FormattedResource',
       formatter: field.title,
     };
-  return {
-    type: 'PromptField',
-    // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
-    label: localized(field?.title || field?.name || part),
-  };
+  return { type: 'UrlPart', value: localized('') };
```
maxpatiiuk, line 164: nice fallback 👍
this should behave as if user gave an empty response to a prompt field, which would exactly match the behavior before sp7 supported this feature

specifyweb/frontend/js_src/lib/components/WebLinks/spec.ts
```diff
@@ -91,7 +89,10 @@ const argumentSpec = f.store(() =>
       syncers.maybe(syncers.xmlContent),
       syncers.default(localized(''))
     ),
-    // Specify 7 only
+    /**
+     *  Specify 7 only, usage of this feature have been
+     * removed in https://github.com/specify/specify7/issues/4525
+     */
     shouldPrompt: pipe(
```
maxpatiiuk, line 96: rename this to `legacyShouldPrompt` - that's the naming I used for every syncer prop that is needed fo sp6 but not sps7

specifyweb/frontend/js_src/lib/components/WebLinks/spec.ts
```diff
@@ -220,6 +201,7 @@ function reconstructWeblink(
               name: localized(argument.name),
               title: localized(argument.title),
               legacyIsEditable: false,
+              shouldPrompt: undefined,
```
maxpatiiuk, line 204: can you please double check how would the XML for the "prompt" tag in the weblink look like after this change?
i.e did every weblink before this have a `<prompt>false</prompt>` or `<prompt>true</prompt>` tag?

if so, would a new web-link created by specify have `<prompt>false</prompt>` (correct) or not have `<prompt>` at all (potentially breaking sp6?)

maxpatiiuk (changes requested):

maxpatiiuk (approved):
looks good (on a condition that https://github.com/specify/specify7/pull/4543#discussion_r1496783588 is verified and is good)

emenslin (approved):
Looks good, 'prompt field' does not show up anymore. 

https://github.com/specify/specify7/assets/150940840/c4ec6ca9-cd61-4ffa-9df5-99b3bc8a50e8

pashiav (approved):
Looks good!
Screen Shot 2024-02-22 at 3 27 19 PM

PR #4546, opened 2024-02-15
feat(FormTable): reduce layout shifts

### LINKED ISSUE #4545: feat(FormTable): reduce layout shifts
In this video, see how Grant accidentally clicked on the "Form Meta" button at 0:06 - https://github.com/specify/specify7/issues/4511#issue-2127883881

That was caused by a layout shift from the header row disappearing.
In order to minimize layout shifts, I would propose that once FormTable's header row has been displayed, instead of disappearing once it's no longer needed, it should just become invisible

maxpatiiuk:
Fixes #4545

See how removing the needless layout shift reduces the possibility for accidental misclick as described in #4545:

https://github.com/specify/specify7/assets/40512816/090b7b1c-dbc1-4518-bea5-917817d2bcce

(especially bad in cases when accidental misclick causes the record to be deleted)

@grantfitzsimmons what do you think about this feature? ship it or not?
also, should we hide the form table header when all records are expanded? (right now it is hidden only when there are 0 resources or when there is 1 resource and that resource is expanded)

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
@@ -71,7 +71,7 @@ export function FormTable<SCHEMA extends AnySchema>({
   onClose: handleClose,
   sortField,
   onFetchMore: handleFetchMore,
-  isCollapsed,
+  isCollapsed = false,
   preHeaderButtons,
 }: {
   readonly relationship: Relationship;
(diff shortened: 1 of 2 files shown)


CarolineDenis (approved):

pashiav (approved):
Looks good!

https://github.com/specify/specify7/assets/118556884/602edb96-acf0-4a00-9a5f-fff306e6e7c6

emenslin (approved):
Looks good, does not shift as much. 

https://github.com/specify/specify7/assets/150940840/d734b89a-96b0-4626-9939-db025e42e9ae

PR #4551, opened 2024-02-16
Fix issue for doubled resolved quantity in loan return

### LINKED ISSUE #4505: Loan return breaks counts
**Describe the bug**
When using a record set to create a loan and then returning it it multiplies the count available and makes the count unavailable negative. 

**To Reproduce**

1. Create a record set with count preparations 
2. Create a loan using that record set 
3. Save loan 
4. Return loan 
5. Check record set count 
6. See error 

**Expected behavior**
There should not be negative counts.

**Screenshots**

https://github.com/specify/specify7/assets/150940840/768fede3-3de0-42c2-ab25-3238a2d56567

- OS: Windows 11
- Browser:  Chrome
- Specify 7 Version: edge

CarolineDenis:
Fixes #4505
Fixes #4551 too ?
### Checklist

### Testing instructions

Create a record set with count preparations
Create a loan using that record set
Save loan
Return loan
Check record set count

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
@@ -228,8 +228,8 @@ function PreparationReturn({
               const loanReturn = new tables.LoanReturnPreparation.Resource();
               loanReturn.set('loanPreparation', preparation.url());
               loanReturn.set('remarks', remarks);
-              loanReturn.set('quantityResolved', returns);
               loanReturn.set('quantityReturned', returns);
+              loanReturn.set('quantityResolved', returns);
               loanReturn.set(
                 'receivedBy',
                 loanReturnPreparation.current.get('receivedBy')
--- a/specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
@@ -229,7 +229,7 @@ function PreparationReturn({
               loanReturn.set('loanPreparation', preparation.url());
               loanReturn.set('remarks', remarks);
               loanReturn.set('quantityReturned', returns);
-              loanReturn.set('quantityResolved', returns);
+              loanReturn.set('quantityResolved', resolve);
               loanReturn.set(
                 'receivedBy',
                 loanReturnPreparation.current.get('receivedBy')


CarolineDenis:
NOTES: 
Need to investigate the business rules for LoanReturnPreparations (@melton-jason)

maxpatiiuk (approved):
👍 as long as Jason approves

@melton-jason should we trigger business rules on next-tick to run rules only after all changes are applied rather than for each change individually?
would that avoid bugs like this?

melton-jason:
That may work, although we would still need to be considerate of the order in which the quantityResolved and quantityReturned fields are applied. 

I take it you are suggesting a way to 'skip' the businessrules being run for these specific field sets? 

Besides your proposed solution, here are two more which can also be considered:
- Allow the `noBusinessRules` option to be passed to `resource.set` and `resource.bulkSet` calls (essentially a more specific implementation of the option which can be passed when a resource is initialized/created or when calling `resource.getRelated`)
  - ``` resource.bulkSet({"quantityResolved": 1, "quantityReturned": 1}, {"noBusinessRules": true}) ``` 
- Add a method to `ResourceBase` which accepts a function which would perform all modifications within the function as if business rules were enabled. An example usage might be:
```ts
loanReturn.withoutBusinessRules(() => {
   loanReturn.set("quantityResolved", 1);
   loanReturn.set("quantityReturned", 1);
   // loanReturn quantityResolved is still 1
)
// normally this would be
loanReturn.set("quantityResolved", 1);
loanReturn.set("quantityReturned", 1);
// loanReturn quantityResolved is 2
```

For both scenarios some modifications would need to be made to both businessRules.ts and ResourceBase. Currently, the only way to disable rules is on resource initialization. 
See
https://github.com/specify/specify7/blob/7041601872f434f61b524f392a9bcf8f2f86d5b8/specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.js#L118

realVinayak (approved):

melton-jason (approved):
This is fine for a temporary fix which respects business rules. 
However, to (better) avoid something similar happening in the future, we would likely need to implement something as suggested in https://github.com/specify/specify7/pull/4551#issuecomment-1952876220 or https://github.com/specify/specify7/pull/4551#pullrequestreview-1886373091. 

Though I would not consider the implementation to be a blocker for this PR specifically, so a separate Issue can be opened and this can be merged.

PR #4555, opened 2024-02-16
Display collection id and user id in denied error dialog

### LINKED ISSUE #4553: Permission denied dialog miss information
**Describe the bug**
In edge when you get the permission denied dialog it does not include all the information and just says "no permission" 

**To Reproduce**

1. Log in as a user with limited permissions 
2. Do something you don't have permission to do 
3. See error with dialog 

**Expected behavior**
The dialog should have the same information that it has in 7.9.3

**Screenshots**
Edge
Screenshot 2024-02-16 095703

7.9.3
Screenshot 2024-02-16 095712

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge

CarolineDenis:
Fixes #4553

### Checklist

### Testing instructions

Log in as a user with limited permissions
Do something you don't have permission to do
See that dialog contains user id and collection id


specifyweb/frontend/js_src/lib/components/Formatters/formatters.ts
```diff
@@ -158,11 +161,18 @@ async function formatField(
 
   let formatted: string | undefined = undefined;
   const hasPermission = hasPathPermission(fields ?? [], 'read');
-  if (hasPermission) {
-    const data = await fetchDistantRelated(parentResource, fields);
+  const data = await fetchDistantRelated(parentResource, fields);
```
maxpatiiuk, line 164: wouldn't this throw an exception if `!hasPermission`?

specifyweb/frontend/js_src/lib/components/Formatters/formatters.ts
```diff
@@ -158,11 +161,18 @@ async function formatField(
...
+
+  function handleData() {
     if (data === undefined) return { formatted: '' };
     const { resource, field } = data;
     if (field === undefined || resource === undefined) return { formatted: '' };
+    return { resource, field };
+  }
```
maxpatiiuk, line 171: if this code is called in both branches, without any arguments, then why does it even need to be a function - could just be inlined?

moreover, both branches include this:
```
    const { resource, field } = handleData() ?? {};
    if (!resource || !field) return { formatted: '' };
```

why is that code inside of branches, when it can be put before the branches (given that it's common to both branches)

maxpatiiuk (commented):
code should probably look like this instead:

```ts
if(hasPermission) {
  // code as it was before, without modifications
}
else
  formatted = tryBest ? naiveFormatter(resource.specifyTable.name, resource.id) : userText.noPermission();
```

pashiav (approved):
Displays IDs. Looks good!
issue-4553

bronwyncombs (approved):
Looks good!

Screenshot 2024-02-19 at 10 29 17 AM
Screenshot 2024-02-19 at 10 29 03 AM

maxpatiiuk (changes requested):

maxpatiiuk (changes requested):
did you mean to push some code?
I don't see any recent commits addressing the last comments I left, nor explanation of why they would/would not be addressed

specifyweb/frontend/js_src/lib/components/Formatters/formatters.ts
maxpatiiuk: responding to the messages you sent me in slack:
> Isn't it like it was in 9.3?
before, we had a permission check:
`const hasPermission = hasPathPermission(fields ?? [], 'read');`
and then only if `hasPermission`, was fetchDistantRelated called
where as with your modification, fetchDistantRelated is always called

maxpatiiuk: > It doesn't throw an error.

~~Could you please double check the network or the console tab?
We have the logic for when one error occurs, other related errors (like permission errors) are suppressed so as not to overwhelm the user needlessly.
Or, can you tell me how to replicate the permission error you are encountering?~~

I double-checked the code, and this code change does indeed introduce a bug.
In the case of this dialog, the current collection and current user resource are provided by the back-end even if user does not have read permission to specify user and collection table (because this bare minimum is necessary for things to work).
And in this dialog, the formatter is invoked to just get collectionName field from Collection and user name from SpecifyUser - since these resources are already fetched, and no requests to related tables are happening, no extra network requests are happening, and so no permission error is occurring.

But if you try to format a resource, where the formatter accesses resources the user has no permission to (i.e try to format Determinaion, when user has no read access to Taxon), you would see an error

maxpatiiuk (approved):
👍

Issue #3169, opened 2023-03-14
Add institutional preferences


maxxxxxdlp:
Similar to collection preferences added by stats page.

The tricky part to this is the migration:
- Need to create a Django migration that gets the original defaults from remote prefs and global prefs
- Also, as part of this, should migrate away from `moment` and `dayjs` libraries and move to a more modern library (`date-fns` was great option last time I checked, but more research is needed) - the current way of specifying a date format is different that what's expected by other libraries so migration would have to do string replacement, rather than simply using the same value
- Add a pref for ability to add institution prefs and enforce that in the app resource editor

realVinayak:
@maxxxxxdlp @grantfitzsimmons should changes done in specify6 to the files be seen in specify7 after this? I think they 7 should maintain it's own

grantfitzsimmons:
I'm okay with migrating existing Specify 6 preferences to 7 and breaking away. I don't see us supporting cross compatible preferences as 6 deletes any preferences it doesn't recognize

maxxxxxdlp:
100% agree with @grantfitzsimmons

Migrate all as best as we can one time on specify update, and then keep them as separate as possible

BUT, take this migration as a time to:
- rethink and standardize the naming (the naming used in remote/global prefs right now is totally inconsistent) - probably would make most sense to reuse the naming convention used in user preferences, and to follow as similar structure to that as possible
- improve existing prefs during the transition (improve names, make values clearer, etc - whatever you see can be improved, now is the best time)
- think through all the other prefs we could introduce so that we can introduce a lot at once (I did that with user preferences) - to discover the edge cases early on and to figure out which categories/subcategories you should have (it's easier to see what categories you should have when you have 30 prefs rather than 3) so that you don't have to add additional migrations in the future to change that
- DATES - The date format is specified in the prefs right now. This is the moment to migrate from moment.js to a newer date library. `luxon` and `date-fns` seem to be the two most popular at the moment - evaluate these to see what works best for us. The syntax for specifying the date format would change. Fortunately, there is only a limited number of date formats we currently support (probably the same as sp6) - https://github.com/specify/specify7/blob/6bc3785ddd2352d57b4ce4bd1937414cea4022a0/specifyweb/stored_queries/format.py#L273-L352 - that makes migration easier (just create a dictionary of mappings from old date to new equivalent date syntax in the new library)
[...]

PR #4556, opened 2024-02-19
Highlight preference text in app resource

### LINKED ISSUE #4496: Syntax highlighting for `preferences` app resources is missing
image

https://geology-edge.test.specifysystems.org/specify/resources/app-resource/4/

On the right is `v7.8.9`

CarolineDenis:
Fixes #4496

### Checklist

### Testing instructions

open app resource 
open global preferences 
verify that the text color is present


specifyweb/frontend/js_src/lib/components/AppResources/EditorComponents.tsx
```diff
@@ -241,7 +241,7 @@ export function useCodeMirrorExtensions(
     const language =
       mode === 'json'
         ? [json(), jsonLinter(handleLinted)]
-        : mode === 'properties'
+        : mode === 'properties' || mode === 'txt'
```
maxpatiiuk, line 244: while it might fix the issue, it's not a correct fix

if you look at this, `txt` is used when code is unable to determine that the resource is a `properties` file:

https://github.com/specify/specify7/blob/8ce714a19c2099c9c675194a6f6eb9aa70dc7213/specifyweb/frontend/js_src/lib/components/AppResources/hooks.ts#L210-L211

thus, we should fix the logic so that it properly detects this file as properties
and if we can't do that, then not fix this at all, as properties files won't be needed once https://github.com/specify/specify7/issues/3169 is implemented (yet another reason to do it)

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/AppResources/EditorComponents.tsx
CarolineDenis: see comment below

maxpatiiuk:
> Name is 'Global Preferences' and not preferences

that doesn't sound right
in this function the label is set to Global Preferences, but the name should not be changed -https://github.com/specify/specify7/blob/b226bae7e62984cf2813da4eef5fda2c126e928f/specifyweb/frontend/js_src/lib/components/AppResources/tree.ts#L101-L122

and according to this comment from Grant, the name is still `preferences`: https://github.com/specify/specify7/issues/4496#issuecomment-1935148705

can you double-check what's going on?

maxpatiiuk:
> the name currently is whatever the user has defined (which to me makes sense) and as discussed for the view def in

no.
properties file are only used by specify if their name matches "preferences" exactly - all other properties files are ignored by sp6 and sp7
so, not having highling for files that are ignored by specify sounds good to me (in fact, it's better that way, as it suggest to the user that they are doing something wrong)

and again, looking at https://github.com/specify/specify7/issues/4496#issuecomment-1935148705, the name is `preferences` in both instances, yet the syntax highlighting got broken, so it looks like a bug rather than a missing feature - can you double check that you can replicate the bug when app resource name is `preferences`?

and looking at https://github.com/specify/specify7/blob/d88e98c30df812c36aca6740235d733c735f6a39/specifyweb/frontend/js_src/lib/components/AppResources/hooks.ts#L210, it is looking for blank mime type and name `preferences`, which matches what https://github.com/specify/specify7/issues/4496#issuecomment-1935148705 has - can you check where the bug lies?

> NOTES: mimeType for global pref is: text/x-java-properties

maybe this is the issue? looking at the code, it seems that specify expects mime type to be empty
maybe the bug is than that we have `text/x-java-properties` rather than undefined in here?
https://github.com/specify/specify7/blob/d88e98c30df812c36aca6740235d733c735f6a39/specifyweb/frontend/js_src/lib/components/AppResources/types.tsx#L200

CarolineDenis:
@maxpatiiuk, 
Yes I confirm that the mime type is 'text/x-java-properties' even in Grant's example. (Inspect it on the test panel) Same for other dbs when looking at the prefs

maxpatiiuk:
hmm if I look at the API response from this URL:
http://localhost/api/specify/spappresource/?limit=0

I see that the `preferences` resource has no mime type:

Screenshot 2024-02-23 at 19 37 38

That, combined with the fact that the code is explicitly checking for mimeType being empty in several places leads me to belive that it should indeed be empty for the `preferences` resource.

Can you double check where did you find the non-empty mime-type? could you check in a different database? maybe it's set on the front-end mistakingly?

If the `preferences` resource has no mime type than we should probably also unset it in here: https://github.com/specify/specify7/blob/504a00037e7b5a66f717072993f43ce020d8fd45/specifyweb/frontend/js_src/lib/components/AppResources/types.tsx#L199

Or is the bug in that front-end is looking for mimeType equal to empty string but back-end returns null?

CarolineDenis:
@maxpatiiuk, 
I tested several dbs and the mime type is always 'text/x-java-properties'. 

In the lines: (EditorWrapper.tsx) 

 ```
 const resourcesTree = useResourcesTree(resources);
  const directory = useDirectory(
    rawDirectoryKey,
    resourcesTree,
    resource,
    resources
  );
```

The mime type for preferences in `resourcesTree` and in `resources` in an empty string but it is `java-text` in `resource`. 
The results you have from you api url correspond to the resources, so it make sense. 

But, it's the resource data that we are using to define the mime type in this case.

maxpatiiuk:
After looking at the code, the bug was introduced in this commit right here:

https://github.com/specify/specify7/commit/9663b72ca44409bcd60c8dc3355b2f1d10f6d154#diff-c950899116a6e1676f42d41649679136b0e6970a772fad3c600c9d57ebec648fR195-R196

The intention was to infer the mime type for resource if it didn't have one

The code change in the commit above itself is fine, but it triggers a bug due to this:

https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/components/AppResources/filtersHelpers.ts#L66-L67

Which basically means, if resource is called `preferences`, and has no mime type, assume it is `otherPropertiesResource`

but then `otherPropertiesResource` is defined as having a mime type of `text/x-java-properties`:

https://github.com/specify/specify7/blob/ea4477670c82b4d1edbe3e31c78c079e4f99abc4/specifyweb/frontend/js_src/lib/components/AppResources/types.tsx#L199-L200

which leads to an inconsistency.

---

I see two solutions:

1. The mimeType `text/x-java-properties`, while a correct mime type for this type of file, does not actually appear in the database, so it could be argued that using it on the front-end is misleading/false - thus, we could change this value to null instead: https://github.com/specify/specify7/blob/ea4477670c82b4d1edbe3e31c78c079e4f99abc4/specifyweb/frontend/js_src/lib/components/AppResources/types.tsx#L200-L200
2. Continue using the "slightly faked" mime type, but then ensure we are consistently looking for it where that's expected, which would involve changing the empty string here to `appResourceSubTypes.otherPropertiesResource.mimeType` (rather than `text/x-java-properties` to make it more explicit that the value is coming from appResourceSubTypes, rather than the database): https://github.com/specify/specify7/blob/ea4477670c82b4d1edbe3e31c78c079e4f99abc4/specifyweb/frontend/js_src/lib/components/AppResources/hooks.ts#L210 (which is similar to the solution you proposed)

Feel free to go with either of those options.

Sorry to drag this out a bit - when pushing back on this PR, I was mainly interested in why this stoped working as nothing at the database level changed - understanding why it broke is crucial for having confidence that the new "solution" didn't break something we didn't expect (especially in absence of solid test coverage)

emenslin (approved):
Looks good, the text is now the correct color.
Screenshot 2024-02-27 125125

maxpatiiuk (approved):

pashiav (approved):
Looks good! Text is colored.
issue-4496

PR #4557, opened 2024-02-19
refactor(utils): optimize removeKey

maxpatiiuk:
Make `removeKey` significantly faster by optimizing the common case by removing the key from the object using destructing, rather than through iterating over each item.

### Checklist

- [x] Add automated tests

### Testing instructions

This change is covered by the existing automated tests

### DIFF
--- a/specifyweb/frontend/js_src/lib/utils/uniquifyName.ts
+++ b/specifyweb/frontend/js_src/lib/utils/uniquifyName.ts
@@ -7,7 +7,6 @@ import { escapeRegExp } from './utils';
 
 const format = {
   title: {
-    // FEATURE: allow users to customize this?
     prefix: ' (',
     suffix: ')',
   },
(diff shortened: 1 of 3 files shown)


specifyweb/frontend/js_src/lib/utils/uniquifyName.ts
```diff
@@ -55,7 +54,13 @@ export function getUniqueName(
...
       ? strippedName
       : `${strippedName}${uniquePart}`;
+
   return newName.length > maxLength
-    ? getUniqueName(name.slice(0, -1 * uniquePart.length), usedNames, maxLength)
+    ? getUniqueName(
+        name.slice(0, -1 * uniquePart.length),
+        usedNames,
+        maxLength,
+        type
```
maxpatiiuk, line 63: A minor bug - I forgot to pass `type` in here before
Fixed, and added a test case

PR #4562, opened 2024-02-19
Recompute delete blockers on parent when changes made to children

### LINKED ISSUE #4536: Delete blocks are not being unset when resource is deleted
1. Add Preparation without specifying a prep type - this creates a validation error
2. Remove the preparation from the form - thus, the field with error no longer exists, and should not prevent save anymore
3. The save is still blocked - should not be

https://github.com/specify/specify7/assets/40512816/83e7c962-3923-4bb7-8ed8-be2965bfe0e7

CarolineDenis:
Fixes #4536

### Checklist

### Testing instructions

Add Preparation without specifying a prep type - this creates a validation error
Remove the preparation from the form - thus, the field with error no longer exists, and should not prevent save anymore
Verify that save button is enabled

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
@@ -12,7 +12,7 @@ import { businessRuleDefs } from './businessRuleDefs';
 import { backboneFieldSeparator, djangoLookupSeparator } from './helpers';
 import type { AnySchema, AnyTree, CommonFields } from './helperTypes';
 import type { SpecifyResource } from './legacyTypes';
-import { setSaveBlockers } from './saveBlockers';
+import { propagateBlockerEvents, setSaveBlockers } from './saveBlockers';
 import { specialFields } from './serializers';
 import type { LiteralField, Relationship } from './specifyField';
 import type { Collection, SpecifyTable } from './specifyTable';
@@ -51,6 +51,7 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
     this.resource.on('change', this.changed, this);
     this.resource.on('add', this.added, this);
     this.resource.on('remove', this.removed, this);
+    this.resource.on('destroy', () => propagateBlockerEvents(this.resource));
   }
 
   public async checkField(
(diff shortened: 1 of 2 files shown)


realVinayak:
Notes:

This is likely the simplest way (lmk if there is another way!), but not that efficient (every deletion of dependent resource triggers recomputation of all save blockers on all other dependent resources). See

https://github.com/specify/specify7/blob/58432982d2b651b16e9d7d3791f2682d3a6277e6/specifyweb/frontend/js_src/lib/components/DataModel/saveBlockers.tsx#L159-L163

A more efficient way would be to save the blocking resources (like the previous implementation), and remove the save blockers belonging to the deleted resource.

maxpatiiuk:
> Doesn't look like it is too tricky to do that. Will just need to add the propagateBlockerEvents function call below in resourceApi.js if I am not misunderstanding

This sounds like a better solution

---

BUT, I would suggest hold off on fixing this until https://github.com/specify/specify7/issues/4286 as:
- whatever solution we come up with, might not work/not be needed once we migrate
- this has been an issue for long enough, so can wait a bit more (and it has a workaround)

CarolineDenis:
> > Doesn't look like it is too tricky to do that. Will just need to add the propagateBlockerEvents function call below in resourceApi.js if I am not misunderstanding
> 
> This sounds like a better solution
> 
> BUT, I would suggest hold off on fixing this until #4286 as:
> 
> * whatever solution we come up with, might not work/not be needed once we migrate
> * this has been an issue for long enough, so can wait a bit more (and it has a workaround)

@maxpatiiuk, so can we merge this solution for 9.4 to have a fix and think of a better solution for later?

maxpatiiuk (approved):

maxpatiiuk:
I'm good with this solution, and I don't think it will be that inefficient
Also, if I understood correctly, based on my talk with @realVinayak yesterday, the other solution he mentioned won't work

emenslin (approved):
Looks good, can save it properly now. 

https://github.com/specify/specify7/assets/150940840/c85435ce-afeb-4c0e-b31d-1d3b2ddf65ae

pashiav (approved):
Allows save. Looks good!

PR #4577, opened 2024-02-22
Move `without preparations` and `add unassociated item` buttons to the left of Interactions dialog

### LINKED ISSUE #4394: Move `Without Preparations` button on Create Loan/Gift dialog to prevent accidental clicks

The "Without Preparations" button on the add cat numbers dialog for loans/gifts is where a "Continue" or "Next" button often is. If you select `By entering Cat #s`, the true "Next" button appears, but it's easy to select "Without Preparations" instead from a UX POV. 

content closed:
Screenshot 2024-01-12 at 11 38 16 AM

content expanded:
Screenshot 2024-01-12 at 11 38 29 AM

We could render the "Without Preparations" option in the same way as the `By choosing record set` and `By entering Cat #s` options so that this content is hidden unless the user chooses this method. If adding an extra click in order to create a new Loan/Gift without preps seems hasty, simply moving the button above the first two options (most likely resizing it) would also make sense to me. 

Could also just move the button to the left side so it's less enticing to click.

sharadsw:
Fixes #4394

### Checklist

### Testing instructions

- Go to Interactions
- Click on Loan or Gift
  - [ ] Verify `Without preparations` button is moved to the left of the dialog and is gray
  - Click on `By choosing a record set`
    - Click on any record set
    - If record set has no preparations, click on continue without preparations
    - Click on + icon next to Loan preparations
    - [ ] Verify `Add unassociated item` button is also moved to the top of the dialog and is gray

image
image


maxpatiiuk (changes requested):
while the code change looks good, I would like to reconsider the UI change.
we don't have a full-width button on top of any dialog anywhere - it just seems weird.

instead, what we do, is put the "alternative/orthogonal" button on the left edge of the dialog
See for example the "Import" button in the new query creation code (feel free to look at how that place is implemented in the code and implement something similar here)

Additionally, consider making the button Gray rather than Blue. Blue indeed makes it seem like "this is the one and only button you should click in this dialog", where as in reality, that dialog has at least 2 other buttons that are just as, if not more important

grantfitzsimmons:
>instead, what we do, is put the "alternative/orthogonal" button on the left edge of the dialog

:+1:

sharadsw:
Moved the buttons to the left and changed the color to gray.

image
image

maxpatiiuk (approved):
👍

CarolineDenis (approved):

emenslin (approved):
Looks good, dialog shows up properly for both loans and gifts! 
Screenshot 2024-02-29 115140

PR #4585, opened 2024-02-24
feat(FormParse): allow "false" as default value for Boolean

### LINKED ISSUE #4581: Allow setting a default value for boolean/checkbox fields
> Is it possible to set a default value for a yesNo tickbox? When viewing an unticked box on a form it is impossible to know if the database value is NULL or 0 - this can cause users to not find all the relevant data if they don't set up the query correctly. This could be fixed if we could set the default value to 0 for unticked boxes but we don't see a way to do that via the form xml? (Having three possible values for a binary result is a little counter-intuitive!)

**Reported by:** CSIRO in October 2022

>These nullable Boolean fields appear to currently operate differently in Specify 6 versus 7 when checkboxes are not explicitly checked on record creation. When new records are created in Specify 6, the checkboxes default to No whereas in Specify 7, they default to null/blank.
>
>From the user perspective, it would be better to have the two versions operate as similarly as possible in the default configuration. 

**Reported by:** Garth at University of Michigan in February 2024

maxpatiiuk:
Fixes #4581

### Checklist

- [x] Add automated tests

### Testing instructions

1. Add a checkbox to the form with `default="true"`.
2. Open the form
3. Without touching the checkbox, save the resource to the database
4. Run a Query Builder query on the resource and make sure that in the query results, the checkbox field has true value
5. Do the above steps with `default="false"` and make sure that query results has false value
6. Do the above steps with no `default` attribute and make sure that query results cell for this value is empty


melton-jason (approved):

emenslin (changes requested):
This does not seem to be working, when I follow the instructions for any of the default values it always shows up as empty in the query.

maxpatiiuk:
Thanks for catching the issue - looks like there is an additional place that needed to be modified.
Though, this PR now affects a highly important file useResourceValue, and thus I would suggest we don't merge this PR in 7.9.4, but merge it after this release to give more time for general testing

emenslin (approved):
Looks good, the different default values are now showing up as expected!

Issue #4499, opened 2024-02-09
Save button error 


emenslin:
**Describe the bug**
When changing the accession number to an incorrect format the save button flashes red but then changes back to orange. 

**To Reproduce**

1. Go to 'accession'
2. Click on 'accession number'
3. Change to an incorrect format

**Expected behavior**
The save button should change to red immediately and stay that way like it does in 7.9.3

**Screenshots**
Edge behavior 

https://github.com/specify/specify7/assets/150940840/164666a4-ffc4-4cfb-829e-12a058d7dc9d

7.9.3.1 behavior 

https://github.com/specify/specify7/assets/150940840/5acca6c7-d6c9-4e63-b7e0-e48d1fa8af52

Alternatively, please fill out the following information manually:

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge
- Database Name: lsumzmammals9823

melton-jason:
Might be related indirectly to #4501

CarolineDenis:
Duplicate of: https://github.com/specify/specify7/issues/4501

PR #4588, opened 2024-02-27
Handle incorrect order of setting save blockers

### LINKED ISSUE #4501: Can bypass uniqueness rules
**Describe the bug**
When setting a uniqueness rule if you try to test it by matching the fields the save button turns red but you can still click save and get an error. 

**To Reproduce**

1. Create a uniqueness rule 
2. Input a value into that field
3. Create a new record with the same value
4. See save button is red 
5. Click save anyway
6. See error

**Expected behavior**
You should not be able to click the save button.

**Screenshots**

https://github.com/specify/specify7/assets/150940840/e3073897-35fa-4394-be15-d4f877a9a10b

**Crash Report**

[Specify 7 Crash Report - 2024-02-09T18_06_46.421Z.txt](https://github.com/specify/specify7/files/14225069/Specify.7.Crash.Report.-.2024-02-09T18_06_46.421Z.txt)

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge

### LINKED ISSUE #4508: Max Uncertainty Est field won't validate on form
**Describe the bug**

After adding GEOLocate marker to application, the form is unable to be saved because of the Max Uncertainty Est field, even though it doesn't actually tell you that. On `7.9.3.1`, the Save button will turn red momentarily but it dismisses when a value is entered into this field. On `edge`, nothing you do makes the form happy. 

**To Reproduce**

1. Go to new Locality
2. Add name and geography
3. Open GEOLocate and place marker
4. Save to Application
5. See error on `edge` 

**Expected behavior**

Allow to save form

**Screenshots**

https://github.com/specify/specify7/assets/135047322/13e7aa42-4f00-42a5-8203-cc4a94cdb9b6

- OS: macOS
- Browser: Chrome
- Specify 7 Version: `edge`
- Database Name: `fwri_1_9_24`

melton-jason:
Fixes #4501

Might fix #4508

The problem here is that `setSaveBlockers` overrides previously set blockers. 
There are two saveBlockers which were related to the issues: uniqueness rules and field parsing. 
If the value in the field violated a uniqueness rule and the rules were checked first the save blockers would be set. 
Then when the parse results of the field were being checked and the value matched the correct format, then the business rule save blockers were removed. 

This is not a long term solution, as this bug can easily appear again if a `setSaveBlockers(resource, field, [])` is ran after all saveBlockers for the field are set. 
Furthermore, the current implementation of saveBlockers only allows one error to be shown to a user at a time, if the blockers are set independently. For instance, if the value violates both uniqueness rules and field parsing, only the blocker which is set last will be set and shown to the user. Ideally, all saveBlockers for a given field should be shown to the user at once.

A better solution may be to require passing a `fieldBlockerKey` to uniquely identify each source of a blocker for a field and only modify the blockers with the specific `fieldBlockerKey` with each `setSaveBlockers`. We may be able to implement this already by utilizing the save blocker message. 
Additionally, saveBlockers already support passing a function which has the fetched blockers as a parameter. However, a distinction needs to be made between blockers to correctly set them in the intended way. 

### Checklist

- [ ] Add automated tests

### Testing instructions
Test #4501 (#4499 is a duplicate of #4501) and #4508 

Additionally, ensure that saveBlockers are being properly set when editing AppResource(s) or Viewset(s). In other words, try and invalidate the file (through syntax error, invalid attribute on a node, etc.)


specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
```diff
@@ -132,9 +132,7 @@ export function useResourceValue<
       );
       if (field === undefined) return;
 
-      // This assumes that there are no field blockers set by anything else
-      if (parseResults.isValid) setBlockers([]);
-      else setBlockers([parseResults.reason]);
+      if (!parseResults.isValid) setBlockers([parseResults.reason]);
```
maxpatiiuk, line 135: would this change prevent blockes from being removed from a field once user resolved the error?

perhaps a better solution would to modify `setBlockers([]);` to instead be something like this?

```ts
setBlockers(blockers=>blockers.filter(blocker=>previousBlockersRef.current.includes(blocker)));
```

that would act just like fieldBlockerKey in many cases, without the associated complexities

maxpatiiuk (commented):
something like `fieldBlockerKey` is something we had, but rewriting business rule manager to typescript to simplify things. Now I regret the decision.
At the time I didn't think it likely that two business rule issues could be present at the same time, and even if there are, in the process of resolving the first issue, the business rule would re-run, thus re-exposing the 2nd issue that was hidden behind at first

specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
melton-jason: Using a `previousBlockersRef` directly as a React ref would be the optimal solution, but is not currently possible as the same ref would need to be used in a non-React component: https://github.com/specify/specify7/blob/29e6646431665b1fe432d36e4cec181bcea0ae08/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts#L165

In the meantime I have implemented a solution proposed by @realVinayak in https://github.com/specify/specify7/pull/4588/commits/076fc3cf2b043158ea682bbd0797f9f9210db0c4. 

We could still likely implement the 'ref' idea in theory, although it would need to handle the case of running in non-React functions.

specifyweb/frontend/js_src/lib/components/DataModel/saveBlockers.tsx
```diff
@@ -131,12 +154,34 @@ export function getFieldBlockers(
...
   field: LiteralField | Relationship
 ): RA<string> {
-  const blockers = saveBlockers.get(resource)?.blockers ?? [];
+  const blockers = getResourceBlockers(resource).blockers;
   return blockers
     .filter((blocker) => blocker.field === field)
     .map(({ message }) => message);
 }
 
+function usePreviouslySetBlocker(
```
maxpatiiuk, line 163: this is not a react hook, so please don't name it as `use*`
quite confusing)
(and ESLint might get confused too)

specifyweb/frontend/js_src/lib/components/DataModel/saveBlockers.tsx
```diff
@@ -131,12 +154,34 @@ export function getFieldBlockers(
...
+  resource: SpecifyResource<AnySchema>,
+  field: LiteralField | Relationship
+): GetSet<FieldValue> {
+  const fieldAndValue = {
+    [field.name]: resource.get(field.name),
+  };
+  if (!previouslySetBlockers.has(resource))
```
maxpatiiuk, line 170: how is `previouslySetBlockers` different from `getResourceBlockers`?
i.e current save blockers are saved somewhere already, correct? so can we get them from there or does that not work?

emenslin (changes requested):
Fixes #4501 but does not fix #4508 and it messes up app resources.

https://github.com/specify/specify7/assets/150940840/e5ca43df-5dd0-429d-88a1-9a913ddb8304

specifyweb/frontend/js_src/lib/components/DataModel/saveBlockers.tsx
melton-jason: The difference is that `previouslySetBlockers` stores the previous value of the field the last time the resource and field was called with `setSaveBlockers`. 
The resource blockers only store the message of the blocker which by itself is not enough information to determine whether it is safe to reset the blockers for the field if whichever consumer of the function is trying to clear the blockers. 

The rationale behind `previouslySetBlockers` is that we know we should NOT reset any existing save blockers if the value for the field has not changed since the last time  `setSaveBlockers` has been called for that field. 
(i.e., we should only allow the blockers to be reset if the field value has changed since the last call to `setSaveBlockers`)

emenslin (approved):
Looks good, app resources is fixed!

CarolineDenis (approved):

pashiav (approved):
Fixed https://github.com/specify/specify7/issues/4501!
Could not reproduce https://github.com/specify/specify7/issues/4508 on edge on `lsumz` and `fwri` dbs.

Issue #1464, opened 2022-05-09
Distinct query results should be navigable 


grantfitzsimmons:
**Describe the bug**
Since record IDs are unique, when you are looking for distinct results, record IDs are not returned, and thus you can't link out to them.

There should be some correction to this behavior in the future if possible.

maxxxxxdlp:
How should this feature even work?
Should it show the first occurrence of a duplicate record and throw out the rest?
Or aggregate multiple records into one row?
Or select a random record?

grantfitzsimmons:
> How should this feature even work? Should it show the first occurrence of a duplicate record and throw out the rest? Or aggregate multiple records into one row? Or select a random record?

Very good points. I was imagining the first occurrence of a duplicate record should be navigable and used for the record set. Perhaps whether it is random or not is unimportant since the use of this feature would be for summarizing information rather than a comprehensive export.

benanhalt:
IMO the back end should detect if there is a unique field in base table included in a distinct query. In that case it would include the ids to allow navigation since the unique field would cause each row to correspond to a single base table record. This would enable a work around by turning on distinct for Andy's pet peeve about the rows being duplicated due to undisplayed fields. Possibly we could make distinct be the default, because why would you ever want duplicates? Maybe there would be performance penalties. I'm not sure. From the _Third Manifesto_:

> By definition, relations do not contain duplicate tuples. Yet SQL does permit duplicate “tuples” in its tables—another of the many reasons why SQL tables are not true relations [72]. In the Manifesto, by contrast, duplicate tuples are absolutely, categorically, and unequivocally outlawed. What we tell you three times is true.

https://www.dcs.warwick.ac.uk/~hugh/TTM/DTATRM.pdf

grantfitzsimmons:
Assuming the same reason is involved, exporting to a CSV is not supported. This should be addressed if there is a solution to this problem. It would be great to see this in the future!

benanhalt:
I don't think #1956 is related to this.

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/queries-7-7-4-cannot-click-records-when-distinct/820/2

CarolineDenis:
Will be fixed in: https://github.com/specify/specify7/issues/4607

acwhite211:
This issue is now moved into #4607

PR #4589, opened 2024-02-27
Avoid query duplicates by grouping by only display fields

### LINKED ISSUE #1450: Queries not respecting unchecked show fields in results for duplicating records
**Describe the bug**
When running a query with the show button unchecked for certain fields, the query result still duplicates records based on that field

**To Reproduce**
1. Go to Default Collection Object query in freshfish database
2. Add record 1234 to query.  Note that citation field has show unchecked
3. Run query
4. See error - CO 1234 is duplicated based on the record having two citations.

**Expected behavior**
I would expect queries to respect unchecked status of fields for duplication as in Sp 6

**Screenshots**
Capture2
Capture3

 - OS: [e.g. Windows 10]

**Reported By**
Name of your institution

**Additional context**

acwhite211:
Fixes #1450

This PR is based on the [issue-1450-2](https://github.com/specify/specify7/tree/issue-1450-2) branch, but also proposes the [issue-1450-3](https://github.com/specify/specify7/tree/issue-1450-3) branch as another potential solution.

Here is an example of the query result from the query described in the #1450 issue
```
[61087,"101010104","Balistidae","Abalistes",null,"K & R 6","04/21/1964","<Aggregator not defined.>","North America, United States, Missouri, Platte","Missouri River, beneath Fairfax Bridge","39.1580429077","-94.6241989136","<Aggregator not defined.>",null,null,"Point","0E-10",2846]
[61087,"101010104","Balistidae","Abalistes",null,"K & R 6","04/21/1964","<Aggregator not defined.>","North America, United States, Missouri, Platte","Missouri River, beneath Fairfax Bridge","39.1580429077","-94.6241989136","<Aggregator not defined.>",null,null,"Point","0E-10",2846]
```
After the changes made in this branch, just one of the duplicate rows is returned
```
[61087,"101010104","Balistidae","Abalistes",null,"K & R 6","04/21/1964","<Aggregator not defined.>","North America, United States, Missouri, Platte","Missouri River, beneath Fairfax Bridge","39.1580429077","-94.6241989136","<Aggregator not defined.>",null,null,"Point","0E-10",2846]
```

The motive of this solution is to avoid the user seeing multiple identical rows next to each other in the query results.  In order to avoid duplicate records like in #1450, this solution takes all of the display fields checked by the user and adds them to a GROUP BY statement, thus avoiding duplicates.

This solution's idea came from https://github.com/specify/specify7/issues/1464#issuecomment-1286985113
> When running a distinct query, can we mimic a distinct query using "GROUP BY" instead of "SELECT DISTINCT" and include the id in the query as a `CONCAT(id, ", ")` (which would be parsed on the front-end into separate records)?

I think this solution is a good idea, and implementing a front-end solution to handle a `GROUP_CONCAT(id, ", ")` field in the record set could be nice.  See branch [issue-1450-3](https://github.com/specify/specify7/tree/issue-1450-3) and commit [group_concat record id field for distinct queries](https://github.com/specify/specify7/commit/92b1a129d513e9f9f92bfa7eaeffa798af303358) where I coded the back-end part for this solution, but still needs front-end code to handle multiple record IDs.

The difference from this branch, is that the issue-1450-3 branch moves the record ID from the group by statement to the GROUP_CONCAT statement.  When showing the results, each row will need to provide links to all the records. After talking with Andy, he thought that checking the distinct checkbox would work if it provided links to each record.  The current design could be modified, but we need to be mindful of any other side effects in changing the original design intentions https://github.com/specify/specify7/blob/67d73f4663cfef67356e95487da61cae1ae43db9/specifyweb/stored_queries/execution.py#L577

Please post your opinion on which solution we should go with for now, branch `issue-1450-2` or branch `issue-1450-3`?
I'm thinking we should further develop the `issue-1450-3` branch.

### Checklist

### Testing instructions

Use the query builder to make queries like those in #1450 and make sure that there are no duplicate records in the results.


acwhite211:
I'm scratching this PR in favor of the issue-1450-3 branch solution.

PR #4594, opened 2024-02-28
Fix failing notification tests

melton-jason:
Fixes the failing `notification` backend tests in production which were caused as a byproduct of  #4495 and hardcoding the expected message id in the `test_get_notification_with_param_since` test (shown below). 

https://github.com/specify/specify7/blob/ea626945cd30ae3cb331ec03d1083ac0ed44da46/specifyweb/notifications/tests.py#L12-L36

The previous code expects `test_get_notification_with_param_since` to be ran first (i.e., the created message to have an if of 1), but the tests can executed in any order.

### DIFF
--- a/specifyweb/notifications/tests.py
+++ b/specifyweb/notifications/tests.py
@@ -37,27 +37,22 @@ class NotificationsTests(ApiTests):
         
     def test_delete_all(self):
         # Create some test messages
-        testMessage1 = Message.objects.create(
+
+        messages = [Message.objects.create(
             user=self.specifyuser,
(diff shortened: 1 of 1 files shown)


CarolineDenis (approved):

pashiav (approved):

melton-jason:
This PR does not need any review from UX-Testing, it is strictly only modifying the backend notification tests. 
Becuase of this, it can likely be merged into production with two approving reviews, although I would still await a third approval before this is merging.

maxpatiiuk (approved):

PR #4595, opened 2024-02-28
Fix Query Builder Aggregator Display

### LINKED ISSUE #4510: Aggregation ordering inconsistent between form and queries
**Describe the bug**
Despite there being one aggregation configured, the order of items in the aggregation differs based on whether it is viewed in the query or on the form.

**To Reproduce**
1. Go to https://sp7demofish2924-edge.test.specifysystems.org/specify/query/87/
2. Run the query
3. See the aggregation for "Collectors" in the query results
4. Browse in forms
5. See the aggregation in the "Collectors" field on the CE form

**Expected behavior**
The behavior should be consistent.

Note that in the examples below, the aggregator orders by `isPrimary` and only one collector _is_ primary. That collector is `Abe, Francine`, and they should appear last in the order since the ordering goes from lowest number to highest number (though it would be nice to configure #4507)

**Expected order:** `Caire, M | Bagley, Justin | Abe, Francine`
_Note that this is the order we see in the query builder in both `edge` and `v7.9.3`_

image

https://sp7demofish2924-edge.test.specifysystems.org/specify/resources/app-resource/103/aggregator/Collector/10

### `edge` (xml-editor):
image

Collectors in query:
`Caire, M | Bagley, Justin | Abe, Francine`

Collectors on the form:
`Bagley, Justin | Caire, M | Abe, Francine`

### `v7.9.3`:
image

Collectors in query:
`Caire, M | Bagley, Justin | Abe, Francine`

Collectors on the form:
` |  | `

**Reported By**
[CSIRO on Asana](https://app.asana.com/0/1201972579963460/1206502287128493/f)

image

acwhite211:
Fixes #4510

On the production branch, the query `https://sp7demofish2924-edge.test.specifysystems.org/specify/query/87/`
returns
```json
{
  "results": [
      [10924,null,10924,"Francine","Abe",true],
      [10924,null,10924,"Justin","Bagley",false],
      [10924,null,10924,"M","Caire",false]]
}
```
when the collectors aggregator fields should not be null, and should be ordered by `isPrimary` (according to the Collector Aggregator), so in this example `Caire, M | Bagley, Justin | Abe, Francine`

### Checklist

- [ ] UI User Testing

### Testing instructions
Create queries with aggregators, and make sure there are not any blank aggregator fields, when they shouldn't be.  So, nothing like ` |  |  | `.  Look at the example in the issue.

Also, create queries with aggregators, then compare the aggregator field in the query results vs the record's form.  The ordering should be generally the same for most 'sort by' fields.  When sorting by a Boolean (True/False) fields, like 'IsPrimary', the exact orderings aren't expected to be exactly the same, because of the ambiguity of the ordering of the rows stored in the database.


acwhite211:
It seems that the sub-query generated by SQLAlchemy for the aggregator was using `LIMIT 0`, thus not returning anything.  This might have been the only issue in the code causing the problem, after doing some testing, it seems to be solved.  Still looking into the aggregation ordering issue.

acwhite211:
SQLAlchemy crashes with an error when the collector aggregator's sort fields is set to 'IsPrimary'', or any other Boolean field.

Here is the problem query being generated by SQLAlchemy
```sql
SELECT DISTINCT 
    (SELECT IFNULL(GROUP_CONCAT(IFNULL(CONCAT(IFNULL(CONCAT(' ', agent_1.LastName), ''), IFNULL(CONCAT(', ', agent_1.FirstName), '')), '') ORDER BY collector.IsPrimary != ' | ' SEPARATOR 0), '') AS blank_nulls_1
    FROM collector 
    LEFT OUTER JOIN agent AS agent_1 ON agent_1.AgentID = collector.AgentID
    WHERE collector.CollectingEventID = collectingevent.CollectingEventID) AS anon_1
FROM collectingevent
WHERE collectingevent.DisciplineID = 3
LIMIT 0, 40;
```

It should be 
`ORDER BY collector.IsPrimary !=  0 SEPARATOR ' | '`
instead of 
`ORDER BY collector.IsPrimary !=  ' | ' SEPARATOR 0`
also, it would be better to use 
`collector.IsPrimary != FALSE`
instead of 
`collector.IsPrimary != 0`

It seems like it is setup properly, here in the expression created during the query building process
```
GROUP_CONCAT(IFNULL(concat(IFNULL(concat(:concat_1, agent_1."LastName"), ''), IFNULL(concat(:concat_2, agent_1."FirstName"), '')), '') ORDER BY collector."IsPrimary" != :IsPrimary_1 SEPARATOR :group_concat_1)
```

This seems correct, so it is odd that the parameters substituted ends up being wrong.
```
[parameters: (' ', ', ', ' | ', 0, 3, 40, 40)]
```

acwhite211:
Ok, looking at v7.9.3 sql output:
```sql
SELECT collectingevent.`CollectingEventID` AS `collectingevent_CollectingEventID`, (SELECT IFNULL(GROUP_CONCAT(IFNULL(concat(IFNULL(concat(%s, agent_1.`LastName`), ''), IFNULL(concat(%s, agent_1.`FirstName`), '')), '') ORDER BY collector.`IsPrimary` SEPARATOR %s), '') AS blank_nulls_1
    FROM collector LEFT OUTER JOIN agent AS agent_1 ON agent_1.`AgentID` = collector.`AgentID`
    WHERE collector.`CollectingEventID` = collectingevent.`CollectingEventID`) AS anon_1
FROM collectingevent
WHERE collectingevent.`DisciplineID` = %s
LIMIT %s, %s
-- INFO:sqlalchemy.engine.base.Engine:(' ', ', ', ' | ', 3, 0, 40)
```

So, the difference in the ORDER BY statement.
In v7.9.3
`ORDER BY collector.IsPrimary SEPARATOR %s`
In edge
`ORDER BY collector.IsPrimary != %s SEPARATOR %s`

I think I'll remove that condition being created in the ORDER BY statement, as it seems unnecessary.

acwhite211:
The group_concat function seems to be returning the correct query expression `GROUP_CONCAT(IFNULL(concat(IFNULL(concat(:concat_1, agent_1."LastName"), \'\'), IFNULL(concat(:concat_2, agent_1."FirstName"), \'\')), \'\') ORDER BY collector."IsPrimary" != :IsPrimary_1 SEPARATOR :group_concat_1)`

str(compiler.params)
`"{'concat_1': ' ', 'concat_2': ', ', 'group_concat_1': ' | ', 'IsPrimary_1': 0}"`

Here is the debugging console output right before `list(query)` runs:
```
compiled = query.statement.compile(dialect=dialect, compile_kwargs={"literal_binds": True})
str(compiled)
    "SELECT collectingevent.`CollectingEventID`, (SELECT IFNULL(GROUP_CONCAT(IFNULL(concat(IFNULL(concat(%s, agent_1.`LastName`), ''), IFNULL(concat(%s, agent_1.`FirstName`), '')), '') ORDER BY collector.`IsPrimary` != %s SEPARATOR %s), '') AS blank_nulls_1 \nFROM collector LEFT OUTER JOIN agent AS agent_1 ON agent_1.`AgentID` = collector.`AgentID` \nWHERE collector.`CollectingEventID` = collectingevent.`CollectingEventID`) AS anon_1 \nFROM collectingevent \nWHERE collectingevent.`DisciplineID` = 3 \n LIMIT 0, 40"
str(compiled.params)
    "{'concat_1': ' ', 'concat_2': ', ', 'group_concat_1': ' | ', 'IsPrimary_1': 0}"
str(compiled.positiontup)
    "['concat_1', 'concat_2', 'group_concat_1', 'IsPrimary_1']"
```

So the problem seems to be that somehow the :IsPrimary_1 and :group_concat_1  are set correctly in the params dictionary, but the params ordering should be `['concat_1', 'concat_2', 'IsPrimary_1', 'group_concat_1']`

acwhite211:
This seems to be an issue with using SQLAlchemy.  I worked with Vinny and found that changing the inherited class of our group_concat function from `expression.FunctionElement` to `expression.ColumnElement` seemed to fix the parameter ordering issue that SQLAlchemy was creating.  Maybe upgrading the version of SQLAlchemy and using `func.group_concat` instead of out own custom group_concat function might be a better idea to try out.

acwhite211:
See https://github.com/specify/specify7/issues/4510#issuecomment-1981730778
> When sorting by a Boolean field, each aggregated value will be either 0 or 1, so that leaves a lot of room for ambiguity when ordering in the aggregator.
>
> In the first example given:
Bagley, Justin | Caire, M | Abe, Francine -> in the form editor
Caire, M | Bagley, Justin | Abe, Francine -> in the query builder results
>
>Both 'Caire, M' and 'Bagley, Justin' have an IsPrimary value of 0, so it is ambiguous as to how to order them.
This one 'Abe, Francine' will always be last because it has an IsPrimary value of 1.
>
>It would seem that in order to avoid this specific issue, we would need to have a secondary sort field added to the query. Thus a secondary sort field option in the Record Formatters Aggregator App Resource.
>
>If that sounds like a good idea, we can go ahead and create a new issue to allow Specify Aggregators to have secondary sort fields.

maxpatiiuk (approved):

CarolineDenis (approved):

emenslin (changes requested):
Using the demofish query the query results vs the forms are still different.

https://github.com/specify/specify7/assets/150940840/3a406da2-62a5-467f-91f9-5309481839b4

acwhite211:
Sorry, @emenslin, I updated the testing instructions.  This PR is mainly addressing the blank aggregators problem reported by CSIRO.

When sorting by a Boolean (True/False) fields, like 'IsPrimary', the exact orderings aren't expected to be exactly the same as Forms, because of the ambiguity of the ordering of the rows stored in the database.  We will need to add secondary 'sprt by' fields in the future to avoid ambiguity.  See this [comment](https://github.com/specify/specify7/issues/4510#issuecomment-1981730778)

emenslin (approved):
No blank aggregator fields and the query results match the browse in forms results.

realVinayak:
Why did you copy-paste the code from [this](https://github.com/specify/specify7/tree/issue-4510-gcpath) branch? Not a problem though, just delete that branch then.

PR #4596, opened 2024-02-29
Query Builder Distinct Option Returns Links to Record IDs

### LINKED ISSUE #1450: Queries not respecting unchecked show fields in results for duplicating records

### LINKED ISSUE #4607: Record Links and Selectability Added to Distinct Query Builder
Add the functionality in the query builder to return links to record IDs when running distinct queries.  When duplicate rows are grouped together as one, there will be a button that links to a dialog box, which will pop up with a dialog box listing links to all the record IDs associated with the duplicate row.  If a row just has one record ID, then it has a normal link to the record, just like in non-distinct queries.

Also, make the row selection functionality work with distinct queries.  The selection checkbox functionality should work just like in non-distinct query results.  The selected rows can then be exported to csv.

This issues solves both #1450 and #1464.

In order to create distinct queries, but aggregating all the record IDs associated with each row, we need to use `GROUP BY` instead of `DISTINCT` in out queries.  So, we add all the fields from the `SELECT` statement (other than the record ID), and add them to the `GROUP BY` statement.  We also add `GROUP_CONCAT(DISTINCT <record_ID>)` in the `SELECT` statement.  Thus, we have query that returns unique rows and all the record IDs associated with each row.

**Testing**
Try using the queries from #1450 and #1464 for testing, and make sure that there are no duplicate records in the results. Also try to make sure that the file created by export csv looks good.

Any query known to create duplicates would be a good test case for this PR. Try running as many queries as you can using the distinct mode.

acwhite211:
Fixes #4607 

After talking with Andy, he thought that checking the distinct checkbox would work to fix #1450, if it provided links to each record. 

The motive of this solution is to avoid the user seeing multiple identical rows next to each other in the query results.  In order to avoid duplicate records like in #1450, this solution takes all of the display fields checked by the user and adds them to a GROUP BY statement, thus avoiding duplicates.  This is done instead of using DISTINCT so that a list of the record IDs can be returned.

This solution's idea came from https://github.com/specify/specify7/issues/1464#issuecomment-1286985113
> When running a distinct query, can we mimic a distinct query using "GROUP BY" instead of "SELECT DISTINCT" and include the id in the query as a `CONCAT(id, ", ")` (which would be parsed on the front-end into separate records)?

I think this solution is a good idea, and implementing a front-end solution to handle a `GROUP_CONCAT(id, ", ")` field in the record set could be nice.  So the front-end code needs to be able to handle multiple record IDs.  When showing the results, each row will need to provide links to all the associated record IDs.

The difference from this branch, is that the issue-1450-2 branch moves the record ID from the GROUP_CONCAT statement to the the group by statement.  The current design could be modified, but we need to be mindful of any other side effects in changing the original design intentions https://github.com/specify/specify7/blob/67d73f4663cfef67356e95487da61cae1ae43db9/specifyweb/stored_queries/execution.py#L577

___
***Edit:
When displaying distinct query results, it will have a column to the record id, so it will look the same as non-distinct queries. The difference if that we're going to have a dialog box, when clicking on the record form link, that will allow you to pick one of the records associated with that query result row. @realVinayak had the idea to base it on the pop-up dialog from [agent view](https://sp7demofish.specifycloud.org/specify/view/agent/66/), which appears when clicking on the 'Delete' button.
___

### Checklist

### Testing instructions

Use the query builder to make queries like those in #1450 and make sure that there are no duplicate records in the results.  Also try to make sure that the file created by export csv looks good.

Any query known to create duplicates would be a good test case for this PR.  Try running as many queries as you can using the distinct mode.


acwhite211:
Update on the front-end design.  When displaying distinct query results, it will have a column to the record id, so it will look the same as non-distinct queries.  The difference if that we're going to have a dialog box, when clicking on the record form link, that will allow you to pick one of the records associated with that query result row. @realVinayak had the idea to base it on the pop-up dialog from [agent view](https://sp7demofish.specifycloud.org/specify/view/agent/66/), which appears when clicking on the 'Delete' button.

acwhite211:
TODO: handle the case where the record ID is put into the select statement of the query.  In that case, we cannot have `SELECT GROUP_CONCAT(id, ','), id`, we'll need to remove one of these.

maxpatiiuk:
> Update on the front-end design. When displaying distinct query results, it will have a column to the record id, so it will look the same as non-distinct queries. The difference if that we're going to have a dialog box, when clicking on the record form link, that will allow you to pick one of the records associated with that query result row. @realVinayak had the idea to base it on the pop-up dialog from [agent view](https://sp7demofish.specifycloud.org/specify/view/agent/66/), which appears when clicking on the 'Delete' button. That is, show the record set for resources (like how delete blockers does)

My initial idea was that it would display one row, but the cell in the first column (the eye icon) would be split (using colspan) to accommodate multiple records inside that row)

but your solution is both simpler and would work better in cases when distinct row is representing hundreds of database record

---

also, would the "Browse in Forms" button work on the front-end then? or is it not useful enough in distinct mode?

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
```diff
@@ -178,7 +177,7 @@ function Row({
         </div>
       )}
       {result
-        .filter((_, index) => !hasIdField || index !== queryIdField)
+        .filter((_, index) => index !== queryIdField)
```
maxpatiiuk, line 180: I think there are more places in the query builder that change their logic based on whether hasIdField or not
@CarolineDenis @acwhite211  would either of you be willing to check if there are more possible places so that we can also refactor them? in general, there shouldn't be any references to `hasIdField` left, right?

specifyweb/stored_queries/group_concat.py
```diff
@@ -25,3 +28,35 @@ def process_clause(idx):
...
+    fields_and_aliases = []
+    for match in matches:
+        field_expr, alias, _ = match
+        field_expr = field_expr.strip()
+        if alias:
+            fields_and_aliases.append(alias)
+        elif field_expr:
+            fields_and_aliases.append(field_expr)
+
+    return fields_and_aliases
```
realVinayak, line 62: Bad, get rid of this. Why not rely more on SQLAlchemy?

specifyweb/stored_queries/execution.py
```diff
@@ -586,7 +588,7 @@ def build_query(session, collection, user, tableid, field_specs,
     query = QueryConstruct(
         collection=collection,
         objectformatter=ObjectFormatter(collection, user, replace_nulls),
-        query=session.query().distinct() if distinct else session.query(id_field),
+        query=session.query(func.group_concat(id_field, separator=',')) if distinct else session.query(id_field), # Also try id_field.distinct()
```
realVinayak, line 591: ` # Also try id_field.distinct`

No, don't lol. you can't just take a distinct on specific fields. Move it to group-by if you want to do that

realVinayak (changes requested):

realVinayak:
@acwhite211 
> TODO: handle the case where the record ID is put into the select statement of the query. In that case, we cannot have SELECT GROUP_CONCAT(id, ','), id, we'll need to remove one of these.

Can you specify which cases you want to handle here? If the user puts record id as a visible field, the ids in the group_concat will always be the same id as the record (you won't get another record's id). However, you could get the SAME record id duplicated (but that won't be fixed by specifically handling record id anyways)

specifyweb/stored_queries/group_concat.py
acwhite211: Ya, I'm definitely getting rid of this.  I was looking in out SQLAlchemy code for where we set the alias name of a field, so I could just put the field's alias name in the group by statement instead of the full expression.

specifyweb/stored_queries/execution.py
acwhite211: Here is a case of row row returned from a distinct query
`[ "20080,20080", "13265", "Characidae", "Acanthocharax sp.", null, "WGS 67-12", "03/17/1967",, "South America, Ecuador, Napo", 'Rio Aguarico, 400 yards South of "Munozlandia" at Santa Cecilia']`

In this case, where the two record IDs are the same, we don't want the user to be shown a potentially long dialog box of the same record ID.

> query=session.query(func.group_concat(id_field.distinct(), separator=',')) if distinct else session.query(id_field),

In SQLALCHEMY, this just makes the `GROUP_CONCAT(id, ',')` distinct, not any other field.  I can confirm this is the case, because the query result count does not change.

The previous row is now:
`[ "20080", "13265", "Characidae", "Acanthocharax sp.", null, "WGS 67-12", "03/17/1967",, "South America, Ecuador, Napo", 'Rio Aguarico, 400 yards South of "Munozlandia" at Santa Cecilia']`

CarolineDenis:
TODO: 
Screenshot 2024-03-01 at 4 35 44 PM

acwhite211:
I'm running into a problem for cases where the query builder involves `blank_nulls`.  I've been trying lots of different ways to detect these fields before adding them to the 'group by' statement, but with no luck so far.  Might have to resort to using regex expressions to parse `query.statement`, or maybe upgrade the version of SQLAlchemy so that there might be functionality to deal with this situation.

The error that ends up occurring is `sqlalchemy.exc.OperationalError: (MySQLdb.OperationalError) (1054, "Unknown column 'blank_nulls_1' in 'group statement'")`.

Here is the sql statement that gets generated by SQLAlchemy
```sql
SELECT GROUP_CONCAT(DISTINCT collectionobject.`CollectionObjectID`) AS group_concat_1,
       CAST(collectionobject.`CatalogNumber` AS DECIMAL(65)) AS anon_1,
       CASE
           WHEN (taxon_1.`TaxonTreeDefItemID` = 19) THEN taxon_1.`Name`
           WHEN (taxon_2.`TaxonTreeDefItemID` = 19) THEN taxon_2.`Name`
           WHEN (taxon_3.`TaxonTreeDefItemID` = 19) THEN taxon_3.`Name`
           WHEN (taxon_4.`TaxonTreeDefItemID` = 19) THEN taxon_4.`Name`
           WHEN (taxon_5.`TaxonTreeDefItemID` = 19) THEN taxon_5.`Name`
           WHEN (taxon_6.`TaxonTreeDefItemID` = 19) THEN taxon_6.`Name`
           WHEN (taxon_7.`TaxonTreeDefItemID` = 19) THEN taxon_7.`Name`
           WHEN (taxon_8.`TaxonTreeDefItemID` = 19) THEN taxon_8.`Name`
           WHEN (taxon_9.`TaxonTreeDefItemID` = 19) THEN taxon_9.`Name`
           WHEN (taxon_10.`TaxonTreeDefItemID` = 19) THEN taxon_10.`Name`
           WHEN (taxon_11.`TaxonTreeDefItemID` = 19) THEN taxon_11.`Name`
       END AS anon_2,
       taxon_1.`FullName`,
       collectionobject.`FieldNumber`,
       collectingevent_1.`StationFieldNumber`,
       DATE_FORMAT(collectingevent_1.`StartDate`, 
           CASE collectingevent_1.`StartDatePrecision`
               WHEN 2 THEN '%m/%Y'
               WHEN 3 THEN '%Y'
               ELSE '%m/%d/%Y'
           END) AS date_format_1,
       (SELECT IFNULL(GROUP_CONCAT(IFNULL(CONCAT(IFNULL(collector.`Text1`, ''), IFNULL(CONCAT(',', agent_1.`LastName`), '')), '') ORDER BY collector.`OrderNumber` SEPARATOR ','), '') AS blank_nulls_1
        FROM collector
        LEFT OUTER JOIN agent AS agent_1 ON agent_1.`AgentID` = collector.`AgentID`
        WHERE collector.`CollectingEventID` = collectingevent_1.`CollectingEventID`
        LIMIT 0) AS anon_3,
       geography_1.`FullName`,
       locality_1.`LocalityName`,
       locality_1.`Latitude1`,
       locality_1.`Longitude1`,
       (SELECT IFNULL(GROUP_CONCAT(IFNULL(CONCAT(IFNULL(preptype_1.`Name`, ''), IFNULL(CONCAT(',', preparation.`CountAmt`), '')), '') SEPARATOR ','), '') AS blank_nulls_2
        FROM preparation
        LEFT OUTER JOIN preptype AS preptype_1 ON preptype_1.`PrepTypeID` = preparation.`PrepTypeID`
        WHERE preparation.`CollectionObjectID` = collectionobject.`CollectionObjectID`
        LIMIT 0) AS anon_4
FROM collectionobject
LEFT OUTER JOIN determination AS determination_1 ON collectionobject.`CollectionObjectID` = determination_1.`CollectionObjectID`
LEFT OUTER JOIN taxon AS taxon_1 ON taxon_1.`TaxonID` = determination_1.`TaxonID`
LEFT OUTER JOIN taxon AS taxon_2 ON taxon_1.`ParentID` = taxon_2.`TaxonID`
LEFT OUTER JOIN taxon AS taxon_3 ON taxon_2.`ParentID` = taxon_3.`TaxonID`
LEFT OUTER JOIN taxon AS taxon_4 ON taxon_3.`ParentID` = taxon_4.`TaxonID`
LEFT OUTER JOIN taxon AS taxon_5 ON taxon_4.`ParentID` = taxon_5.`TaxonID`
LEFT OUTER JOIN taxon AS taxon_6 ON taxon_5.`ParentID` = taxon_6.`TaxonID`
LEFT OUTER JOIN taxon AS taxon_7 ON taxon_6.`ParentID` = taxon_7.`TaxonID`
LEFT OUTER JOIN taxon AS taxon_8 ON taxon_7.`ParentID` = taxon_8.`TaxonID`
LEFT OUTER JOIN taxon AS taxon_9 ON taxon_8.`ParentID` = taxon_9.`TaxonID`
LEFT OUTER JOIN taxon AS taxon_10 ON taxon_9.`ParentID` = taxon_10.`TaxonID`
LEFT OUTER JOIN taxon AS taxon_11 ON taxon_10.`ParentID` = taxon_11.`TaxonID`
LEFT OUTER JOIN collectingevent AS collectingevent_1 ON collectingevent_1.`CollectingEventID` = collectionobject.`CollectingEventID`
LEFT OUTER JOIN locality AS locality_1 ON locality_1.`LocalityID` = collectingevent_1.`LocalityID`
LEFT OUTER JOIN geography AS geography_1 ON geography_1.`GeographyID` = locality_1.`GeographyID`
LEFT OUTER JOIN collector AS collector_1 ON collectingevent_1.`CollectingEventID` = collector_1.`CollectingEventID`
LEFT OUTER JOIN agent AS agent_2 ON agent_2.`AgentID` = collector_1.`AgentID`
LEFT OUTER JOIN collectionobjectcitation AS collectionobjectcitation_1 ON collectionobject.`CollectionObjectID` = collectionobjectcitation_1.`CollectionObjectID`
LEFT OUTER JOIN referencework AS referencework_1 ON referencework_1.`ReferenceWorkID` = collectionobjectcitation_1.`ReferenceWorkID`
WHERE collectionobject.`CollectionID` = 4 AND determination_1.`IsCurrent` = TRUE
GROUP BY CAST(collectionobject.`CatalogNumber` AS DECIMAL(65)),
         CASE
             WHEN (taxon_1.`TaxonTreeDefItemID` = 19) THEN taxon_1.`Name`
             WHEN (taxon_2.`TaxonTreeDefItemID` = 19) THEN taxon_2.`Name`
             WHEN (taxon_3.`TaxonTreeDefItemID` = 19) THEN taxon_3.`Name`
             WHEN (taxon_4.`TaxonTreeDefItemID` = 19) THEN taxon_4.`Name`
             WHEN (taxon_5.`TaxonTreeDefItemID` = 19) THEN taxon_5.`Name`
             WHEN (taxon_6.`TaxonTreeDefItemID` = 19) THEN taxon_6.`Name`
             WHEN (taxon_7.`TaxonTreeDefItemID` = 19) THEN taxon_7.`Name`
             WHEN (taxon_8.`TaxonTreeDefItemID` = 19) THEN taxon_8.`Name`
             WHEN (taxon_9.`TaxonTreeDefItemID` = 19) THEN taxon_9.`Name`
             WHEN (taxon_10.`TaxonTreeDefItemID` = 19) THEN taxon_10.`Name`
             WHEN (taxon_11.`TaxonTreeDefItemID` = 19) THEN taxon_11.`Name`
         END,
         taxon_1.`FullName`,
         collectionobject.`FieldNumber`,
         collectingevent_1.`StationFieldNumber`,
         DATE_FORMAT(collectingevent_1.`StartDate`,
         CASE collectingevent_1.`StartDatePrecision`
             WHEN 2 THEN '%m/%Y'
             WHEN 3 THEN '%Y'
             ELSE '%m/%d/%Y'
         END),
         blank_nulls_1,  -- needs to be anon_3
         geography_1.`FullName`,
         locality_1.`LocalityName`,
         locality_1.`Latitude1`,
         locality_1.`Longitude1`,
         blank_nulls_2  -- needs to be anon_4
ORDER BY taxon_1.`FullName` ASC,
         collectionobject.`FieldNumber` ASC,
         determination_1.`IsCurrent` ASC,
         collectingevent_1.`StationFieldNumber` ASC
LIMIT 40 OFFSET 0;
```

Here is the code segment where I am trying to detect `blank_nulls` or 'anon_3'
build_query.py ->
```python
for fs in field_specs:
    ...
    query, field, predicate = fs.add_to_query(query, formatauditobjs=formatauditobjs)
     if fs.display:
         formatted_field = query.objectformatter.fieldformat(fs, field)
         query = query.add_columns(formatted_field)
         selected_fields.append(formatted_field) # adds to list of fields that will be added to the group by statement
```
I haven't found any way to detect 'blank_nulls' or 'anon_3' through the `fs`, `query`, `field`, `predicate`, or `formatted_field` objects.  I also haven't found anyway to detect if any of these objects are associated with type `class blank_nulls(expression.FunctionElement)`

@melton-jason @maxpatiiuk @realVinayak have any idea how this could be accomplished before resorting to parsing `query.statement` ? I'll keep working at resolving this, but wanted to know if anyone had a solution in mind.  This edge case seems to be the last issue holding back this PR.

realVinayak:
See https://github.com/specify/specify7/pull/4600. Can you look over changes in that PR and merge after resolving comments and doing the TODOs?

specifyweb/stored_queries/format.py
```diff
@@ -235,7 +236,9 @@ def aggregate(self, query: QueryConstruct,
         aggregated = blank_nulls(group_concat(formatted, separator, *order_by_expr))
 
 
-        return subquery.query.add_column(aggregated).limit(limit).label(None)
+        aggregator_label = f"aggregator_{self.aggregator_count}"
+        self.aggregator_count = (self.aggregator_count + 1) % 1000
```
realVinayak, line 240: Hmm why modulo?

specifyweb/stored_queries/tests.py
```diff
@@ -476,6 +476,29 @@ def test_relationships_in_switch_fields(self):
             query = query.query.add_columns(expr)
             self.assertCountEqual(list(query), [('text 1 value for this accession',), (' text 2 value for this accession role2',)])
 
+    def test_group_by_with_aggregators(self):
+      with models.session_context() as session:
+        prep_col = func.max(getattr(models.Preparation, models.Preparation._id))
+
```
realVinayak, line 482: No. Django creates a `test_DATABASE` during testing. You need to connect to that. What you have is trying to connect to actual database. See the backend tests failing, and read up more on django's test databases.

The test_session_context() instead connects to the test database (created by Django). Use it instead. 

So, use 
```
with FormatterAggregatorTests.test_session_context() as session:
```
instead of 
```
with models.session_context() as session:
```

specifyweb/stored_queries/tests.py
```diff
@@ -476,6 +476,29 @@ def test_relationships_in_switch_fields(self):
...
+        new_query = (
+            session.query(models.CollectionObject._id)
+            .select_from(models.CollectionObject)
+            .filter(models.CollectionObject.CollectionID == 1)
+            .add_column(query)
+            .group_by(query)
+        )
+        
+        str_query = str(new_query.statement.compile(compile_kwargs={"literal_binds": True}))
+        rr = list(new_query)
```
realVinayak, line 500: Hmm I'd say add a unit test that tests if sql statement generated is correct or not. See some other unit tests above that generate sql for expressions and compare it.

realVinayak (changes requested):

specifyweb/stored_queries/tests.py
acwhite211: Ya, we could do that, I was just using the test to check that it the query generated is valid.  So as long as no exception is created, the tests passes.

specifyweb/stored_queries/tests.py
acwhite211: 👍 ya, thanks, nice catch, forgot to change that when I moved the function.

specifyweb/stored_queries/format.py
acwhite211: 😁 lol, just to prevent the number getting too ridiculously big.  But ya, since a new ObjectFormatter object is created for each query, we don't need this.

acwhite211:
I'm working on updating unit tests, but front-end user testing can start now.  The query from #1450 is working now without errors.  The csv exporting of distinct queries is working as well.

emenslin (approved):
Looks good, exporting as csv also works and looks correct. 

https://github.com/specify/specify7/assets/150940840/c26af449-dc6d-4c20-a9d2-5851f5169388

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -256,14 +256,16 @@ export function QueryResults(props: QueryResultsProps): JSX.Element {
...
             />
-            <QueryToForms
-              results={results}
-              selectedRows={selectedRows}
-              table={table}
-              totalCount={totalCount}
-              onDelete={handleDelete}
-              onFetchMore={isFetching ? undefined : handleFetchMore}
-            />
+            {isDistinct ? null : (
```
maxpatiiuk, line 259: (sad face)
okay.
if distict queries won't support QueryToForms, do we also need to disable other buttons for distinct queries (query to map, and etc...?)

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
```diff
@@ -120,6 +118,13 @@ function Row({
   );
   const viewUrl = typeof resource === 'object' ? resource.viewUrl() : undefined;
 
+  const splitIds: RA<number> | undefined =
```
maxpatiiuk, line 121: (optional) useMemo?

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
```diff
@@ -120,6 +118,13 @@ function Row({
...
+    typeof result[0] === 'string' && result[0].includes(',')
+      ? result[0].split(',').map(Number)
+      : undefined;
+
+  const [isIdListOpen, _, __, handleToggle] = useBooleanState(false);
```
maxpatiiuk, line 126: (optional) while I normally advise for using useBooleanState, given that we don't yet have virtual rendering, the `Row` component may be called many times, and so needs to be as efficient as possible.
given that, you might want to use a regular useState here as it would have slightly less overhead than useBooleanState (because useBooleanState calls several other hooks inside of it, including useCallback 3 times)

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
```diff
@@ -169,16 +174,44 @@ function Row({
...
-              rel="noreferrer"
-            />
+            {splitIds === undefined ? (
+              <Link.NewTab
+                className="print:hidden"
+                href={viewUrl}
+                rel="noreferrer"
+              />
+            ) : (
+              <Button.Icon
```
maxpatiiuk, line 184: add `className="print:hidden"` (to match the Link.NewTab above)

then can press Cmd+p in the browser to test how that look

or alternatively could emulate print mode using devtools - https://developer.chrome.com/docs/devtools/rendering/emulate-css#emulate_css_media_type_enable_print_preview

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
```diff
@@ -169,16 +174,44 @@ function Row({
...
+                icon="viewList"
+                title={queryText.viewListOfIds()}
+                onClick={handleToggle}
+              />
+            )}
+            {isIdListOpen && splitIds !== undefined ? (
+              <Dialog
```
maxpatiiuk, line 191: wouldn't this be better implemented by using RecordSelectorFromIds?
I think that might be more user friendly (will let them preview the records without opening a new tab, and go between records, and easily see total count of records, and easy jump to first/last record)

plus, if you have many records, this dialog would become unusable/not user friendly, but a RecordSelectorFromIds dialog would work fine

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -577,7 +577,8 @@ function Wrapped({
               <QueryResultsWrapper
                 createRecordSet={
                   !isReadOnly &&
-                  hasPermission('/querybuilder/query', 'create_recordset') ? (
+                  hasPermission('/querybuilder/query', 'create_recordset') &&
+                  !queryResource.get('selectDistinct') ? (
```
maxpatiiuk, line 581: can you use `query.selectDistinct` instead?

`queryResource.get('selectDistinct')` is not reactive (might not update the component on change)

specifyweb/frontend/js_src/lib/localization/query.ts
```diff
@@ -875,4 +875,10 @@ export const queryText = createDictionary({
     'fr-fr': 'Faux ou vide',
     'ru-ru': 'Ложь или пусто',
   },
+  viewListOfIds: {
+    'en-us': 'View list of ids.',
```
maxpatiiuk, line 879: remove trailing dot? while not a big issue, we normally don't have it in the text that is used in `titles` (just for consistency, and to allow reusing title strings outside of titles)

```suggestion
    'en-us': 'View list of ids',
```

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
maxpatiiuk: and then could you rename this feature from "list of ids" to something more user friendly like "list of records", or "record collection" or etc?

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
realVinayak: Nice idea! Do you think it would be better to have just one state at the top? Then, we can pass a callback (as a prop ig) to row which calls it to set the list of ids? Basically, shift the state responsible for displaying ids at the top and pass a callback to set the state to children (rows). I think that should be more efficient?

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
CarolineDenis: for now yes, same behavior as before

CarolineDenis: For now just this change is sufficient

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
CarolineDenis: I tried with query.selectDistinct but then the issue is that when creating a query, select distinct, run the query, the createRecordSet is not displayed; but, if check the distinct box, even without running the query again, the button would appear as the state is changing. Hence my use of queryResource.get

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
CarolineDenis: It would complicate things a lot, as we need to handle the fact that a user can delete or modify a resource in the preview

CarolineDenis: I pushed a code using `RecordSelectorFromIds`, but I actually prefer the previous code that used `Dialog`. It's simpler, easier to manage, and less prone to bugs. While the list can be long, the user can scroll through it. However, I do acknowledge that for the user, it might be easier to slide between records rather than opening them one at a time.

acwhite211: I tested the dialog box solution out with a row that had about 100 record IDs, and it was simple enough to scroll through the list.  I am kind of drawn to the simplicity of the dialog solution.  @grantfitzsimmons what do you think?

grantfitzsimmons: @maxpatiiuk's comments are exactly right. Database IDs should only be exposed to the user as an absolute last resort.

The records should use the table format to determine how they appear. Instead of `1`, it could use the table format for that record if available. If there is no formatter defined, IDs can be shown instead.

"List of ids" should change to "List of Records" :100:

It would be great to have some kind of "browse in forms" mechanism in that dialog so the user can preview them, but I think the dialog box solution is a good first step.

grantfitzsimmons (changes requested):
> @maxpatiiuk's comments are exactly right. Database IDs should only be exposed to the user as an absolute last resort.
> 
> The records should use the table format to determine how they appear. Instead of `1`, it could use the table format for that record if available. If there is no formatter defined, IDs can be shown instead.
> 
> "List of ids" should change to "List of Records" 💯
> 
> It would be great to have some kind of "browse in forms" mechanism in that dialog so the user can preview them, but I think the dialog box solution is a good first step.

https://github.com/specify/specify7/pull/4596#discussion_r1515119017

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
maxpatiiuk: > Nice idea! Do you think it would be better to have just one state at the top? Then, we can pass a callback (as a prop ig) to row which calls it to set the list of ids? Basically, shift the state responsible for displaying ids at the top and pass a callback to set the state to children (rows). We would need to move dialog ofc to the parent too. I think that should be more efficient?

yes, that would be even better! that's what we do in the attachments gallery if I remember right

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
maxpatiiuk: > It would be great to have some kind of "browse in forms" mechanism in that dialog so the user can preview them, but I think the dialog box solution is a good first step.

I should have been more explicit - when I said "RecordSelectorFromIds", I meant "browse in forms" as "RecordSelectorFromIds" is how that feature is called in the code.
Implementing "browse in forms" here would be quite simple as that's exactly the use case RecordSelectorFromIds was created for

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
```diff
@@ -169,16 +174,48 @@ function Row({
...
+                table={table}
+                title={commonText.colonLine({
+                  label: queryText.queryResults(),
+                  value: table.label,
+                })}
+                totalCount={splitRecords.length}
+                onAdd={undefined}
+                onClone={undefined}
+                onClose={() => toggleIsListOfRecordsOpen(false)}
+                onDelete={undefined}
```
maxpatiiuk, line 209: I am fine with delete not being supported, here, but just wanted to mention that it won't be very complicated:

you store `splitRecords` in a state
onDelete you remove one item from that state
you could also look at how "Browse in forms" does it right now

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
CarolineDenis: Yes, I had a delete option before but we decided to not support it here.

CarolineDenis (approved):

acwhite211:
I talked with Andy, he tested and signed off on this issue.  Going to go ahead and close this issue.  Might in the future create a new issue to add a form editor button to distinct queries, if we want to.

PR #4597, opened 2024-02-29
Expand collector subform when adding new

### LINKED ISSUE #4578: Collectors subform showing up as read only
**Describe the bug**
In the collection object form when opening the collectors subform it shows up as read only fields and you have to expand to be able to type anything. 

**To Reproduce**

1. Go to collection object form
2. Click on the collectors subform 
3. See error

**Expected behavior**
The form should open up as fields that can be typed in without having to expand the form 

**Screenshots**
Edge behavior vs 7.9.3

https://github.com/specify/specify7/assets/150940840/2ca21cbe-8965-4998-ae3c-4a68812b8912

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge

CarolineDenis:
Fixes #4578

### Checklist

### Testing instructions

- Go to collecting event form
- Click on the collectors subform
- See that it is expanded 

Also test: 
add new prep in gift or loan
add other types of records in subview

Notes: 
with the code changes during the refactoring of IntegratedRecordSelector.tsx between 9.3 and 9.4 we had collection.add() being called twice for resources in formTable.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
@@ -240,7 +240,8 @@ export function IntegratedRecordSelector({
                 preHeaderButtons={collapsibleButton}
                 sortField={sortField}
                 viewName={viewName}
-                onAdd={(): void => {
+                onAdd={(resources): void => {
+                  if (!isInteraction) collection.add(resources);
                   if (typeof handleAdd === 'function') handleAdd();
                 }}
                 onClose={handleClose}
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
@@ -115,6 +115,7 @@ export function IntegratedRecordSelector({
       <RecordSelectorFromCollection
         collection={collection}
         defaultIndex={isToOne ? 0 : index}
+        formType={formType}
         isCollapsed={isCollapsed}
         isInteraction={isInteraction}
         relationship={relationship}
@@ -130,7 +131,6 @@ export function IntegratedRecordSelector({
           handleExpand();
           if (typeof urlParameter === 'string') setIndex(index.toString());
         }}
-        formType={formType}
         {...rest}
       >
         {({
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
@@ -130,6 +130,7 @@ export function IntegratedRecordSelector({
           handleExpand();
           if (typeof urlParameter === 'string') setIndex(index.toString());
         }}
+        formType={formType}
         {...rest}
       >
         {({
--- a/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
@@ -115,12 +115,11 @@ export function IntegratedRecordSelector({
       <RecordSelectorFromCollection
         collection={collection}
         defaultIndex={isToOne ? 0 : index}
-        formType={formType}
-        isCollapsed={isCollapsed}
-        isInteraction={isInteraction}
         relationship={relationship}
         onAdd={(resource) => {
           if (isInteraction) handleOpenDialog();
+          if (!isInteraction && formType !== 'formTable')
+            collection.add(resource);
           handleAdding(resource);
         }}
         onDelete={(...args): void => {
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -115,6 +115,7 @@ export function IntegratedRecordSelector({
       <RecordSelectorFromCollection
         collection={collection}
         defaultIndex={isToOne ? 0 : index}
+        formType={formType}
         isCollapsed={isCollapsed}
```
maxpatiiuk, line 119: another small issue not caused by this PR:
you pass `isCollapsed` to RecordSelectorFromCollection, but that component does not need it (though it's in the TypeScript type)
can you please change the type to not need isCollapsed and see how that affects other places that are using the same type and might need isCollapsed

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromCollection.tsx
```diff
@@ -103,7 +105,7 @@ export function RecordSelectorFromCollection<SCHEMA extends AnySchema>({
       const resources = isToOne ? rawResources.slice(0, 1) : rawResources;
       if (isDependent && isToOne)
         collection.related?.placeInSameHierarchy(resources[0]);
-      if (!isInteraction) collection.add(resources);
+      if (!isInteraction && formType !== 'formTable') collection.add(resources);
       handleAdd?.(resources);
```
maxpatiiuk, line 109: could you please refactor the code such that this component does not receive the isInteraction and formType props at all? (generally, try to reduce how many props a component receives, as that both improves maintainability of the code and improves performance due to fewer re-renders)
it should be pretty simple - just move this line that you touched:

```ts
if (!isInteraction && formType !== 'formTable') collection.add(resources);
```

into the handleAdd

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -240,7 +241,8 @@ export function IntegratedRecordSelector({
                 preHeaderButtons={collapsibleButton}
                 sortField={sortField}
                 viewName={viewName}
-                onAdd={(): void => {
+                onAdd={(resources): void => {
+                  if (!isInteraction) collection.add(resources);
```
maxpatiiuk, line 245: can you please double-check if this line should be added here, or on line 122
(if you add it on line 122, this could would be inside the handleAdd() function, and thus also affect the  line 184)

maxpatiiuk (approved):
your changes look good (assuming they fix the bugs), but there are several small refactors you could make to these files while you are at it

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
CarolineDenis: Do you mean remove the type isCollapsed from type RecordSelectorProps?

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
CarolineDenis: Yes I do need it there.

maxpatiiuk: I understand that you need it here
I am saying that if you would move this code, than it would apply in both places where handleAdd is called
can you check if this code is needed only here or in both places where handleAdd is called?

specifyweb/frontend/js_src/lib/components/FormSliders/IntegratedRecordSelector.tsx
```diff
@@ -240,7 +240,8 @@ export function IntegratedRecordSelector({
...
                   if (typeof handleAdd === 'function') handleAdd();
                 }}
```
maxpatiiuk, line 246: you shouldn't be calling collecion.add here as it's now part of handleAdd

```suggestion
                onAdd={handleAdd}
```

CarolineDenis: the handleAdd() in the FormTableCollection props is not called from line 122 (the onAdd prop of RecordSelectorFromCollection) but from line 132 in FormTable. Hence the need to keep this here. 
https://github.com/specify/specify7/blob/9044ffc1e29eea13e1d18ccf20cad10ef5290978/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx#L132

CarolineDenis: I tested this solution and it works for all cases it seems. (interactions, subviews, formTable, etc)

pashiav (approved):
Looks good! No longer appearing as read-only.

https://github.com/specify/specify7/assets/118556884/ead619b1-3606-4a4a-9b77-d91f8918bca0

CarolineDenis:
@pashiav, 
Can you also please test the other places mentioned in the testing instructions and confirm it works there too?

pashiav (approved):
Confirmed it works in Gift/Loan subviews!

bronwyncombs (changes requested):
Collectors subview opens as expanded grid in edit mode, but first field is not focused. This is the case for Funding agent on Collecting Trip, Authors on Reference Work, Members on Agent, so I assume this is the case for all agent subforms.  

https://github.com/specify/specify7/assets/135047322/92c9903f-50c9-4409-a946-abf7fe983728

Not sure if this is something we want to fix or not, but I know there was a change to focus first field when records open so it might be good for consistency's sake.

See andy's comment here

maxpatiiuk (approved):

bronwyncombs (approved):
Given the info in @CarolineDenis response, this is working as expected. Approve 👍

PR #4600, opened 2024-03-03
Use label instead of as_scalar()

realVinayak:
Fixes https://github.com/specify/specify7/pull/4596#issuecomment-1974070716

For some reason, SQLAlchemy is treating Labal and Scalar Selects differently when adding as a field in group_by. Looks like it takes the first field in scalar select as the field in groupby. The current workaround is to create "anon" label of an aggregated subquery (which is the only place which can create such a subquery). See SQLAlchemy [docs](https://docs.sqlalchemy.org/en/13/core/selectable.html#sqlalchemy.sql.expression.Select.label), looks like label() creates a scalar subquery anyways. Calling label(None) makes an anon label (like "anon_1")

However, this creates a bug in automated testing. I don't know if .label(None) is thread-safe. The current django unit tests run on a single core. If running on different cores in future, .label(None) could create non-deterministic labels. So, if a unit test relies on absolute equality, it would fail. For example, two separate unit tests may call .label(None) and get "anon_1" and "anon_2" labels, but the unit test expects "anon_1" and "anon_1". There is a straightforward fix for it, by keeping track of number of times aggregate() has been called, and making a label like "aggregated_COUNT" and then passing that label instead of None.

### Checklist

- [ ] Add automated tests

### DIFF
--- a/specifyweb/stored_queries/execution.py
+++ b/specifyweb/stored_queries/execution.py
@@ -10,7 +10,7 @@ from functools import reduce
 
 from django.conf import settings
 from django.db import transaction
-from sqlalchemy import sql, orm, func
+from sqlalchemy import sql, orm, func, select
 from sqlalchemy.sql.expression import asc, desc, insert, literal
 
 from specifyweb.stored_queries.group_concat import group_by_displayed_fields
@@ -384,7 +384,17 @@ def createPlacemark(kmlDoc, row, coord_cols, table, captions, host):
 
     return placemarkElement
 
-
+def test_qb(request):
+    with models.session_context() as session:
+        prep_col = func.max(getattr(models.Preparation, models.Preparation._id))
+        query = select([prep_col]
+                       ).select_from(models.Preparation).where(models.Preparation.CollectionObjectID == models.CollectionObject.collectionObjectId).correlate(models.CollectionObject).as_scalar()
+        new_query = session.query(models.CollectionObject._id).select_from(models.CollectionObject).filter(models.CollectionObject.CollectionID == request.specify_collection.id)
+        new_query = new_query.add_column(query)
+        new_query = new_query.group_by(query)
+        str_query = str(new_query.statement.compile(compile_kwargs={"literal_binds": True}))
+        rr = list(new_query)
+    raise Exception('aha')
 
 def run_ephemeral_query(collection, user, spquery):
     """Execute a Specify query from deserialized json and return the results
@@ -649,5 +659,5 @@ def build_query(session, collection, user, tableid, field_specs,
     if distinct:
         query = group_by_displayed_fields(query, selected_fields)
 
-    logger.debug("query: %s", query.query)
+    logger.warning("query: %s", query.query)
     return query.query, order_by_exprs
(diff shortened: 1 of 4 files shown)


specifyweb/stored_queries/execution.py
```diff
@@ -384,7 +384,17 @@ def createPlacemark(kmlDoc, row, coord_cols, table, captions, host):
...
+    with models.session_context() as session:
+        prep_col = func.max(getattr(models.Preparation, models.Preparation._id))
+        query = select([prep_col]
+                       ).select_from(models.Preparation).where(models.Preparation.CollectionObjectID == models.CollectionObject.collectionObjectId).correlate(models.CollectionObject).as_scalar()
+        new_query = session.query(models.CollectionObject._id).select_from(models.CollectionObject).filter(models.CollectionObject.CollectionID == request.specify_collection.id)
+        new_query = new_query.add_column(query)
+        new_query = new_query.group_by(query)
+        str_query = str(new_query.statement.compile(compile_kwargs={"literal_binds": True}))
+        rr = list(new_query)
+    raise Exception('aha')
```
realVinayak, line 397: @acwhite211 @CarolineDenis I added this snippet of code for debugging. Replace `query` by `query.label(None)` in 
```
new_query = new_query.group_by(query)
```
and see that going to `localhost/api/test_error/` the column error is no longer thrown. 


TODO for you: Get rid of this function, and corresponding code in views.py

specifyweb/stored_queries/execution.py
```diff
@@ -649,5 +659,5 @@ def build_query(session, collection, user, tableid, field_specs,
     if distinct:
         query = group_by_displayed_fields(query, selected_fields)
 
-    logger.debug("query: %s", query.query)
+    logger.warning("query: %s", query.query)
```
realVinayak, line 662: Agh. Restore to logger.debug

specifyweb/stored_queries/format.py
```diff
@@ -236,7 +235,7 @@ def aggregate(self, query: QueryConstruct,
         aggregated = blank_nulls(group_concat(formatted, separator, *order_by_expr))
 
 
-        return subquery.query.add_column(aggregated).limit(limit).as_scalar()
+        return subquery.query.add_column(aggregated).limit(limit).label(None)
```
realVinayak, line 238: We could have applied this .label(None) [here](https://github.com/specify/specify7/blob/f4a107d3d136d69cf45f3309aec450f58266dd9b/specifyweb/stored_queries/execution.py#L625-L627) too, but I think it is worth handling it at the level of aggregate() if someplace might also use it in the future

CarolineDenis (approved):

acwhite211 (approved):

acwhite211:
I went ahead and merged it into the issue-1450-3-patch branch, working on the TODOs and seeing if changing to .label(None) is a good solution.

realVinayak:
Sounds good. The `test_qb` contains the smallest query to recreate the problem rather than the jungle of qb backend. So it highlights that it is either a bug in the library or in the way models are setup. This also means that you fix it in `test_qb`, you can copy the same fix to query builder

PR #4606, opened 2024-03-04
Expand merging to Localities, Paleo Ctxt, Collecting Event

### LINKED ISSUE #4046: Expand merging table support
- [ ] Locality
- [ ] Paleo Context
- [ ] Collecting Event

and more.

We should do this once `xml-editor` is merged.

CarolineDenis:
Fixes #4046

### Checklist

- [ ] Add automated tests

### Testing instructions

- perform same tests as agent merging but for new tables 
- [ ] Verify that we can’t merge PC and CE records when they are configured to be embedded
- [ ] Test for https://github.com/specify/specify7/issues/4969 (see instructions in that issue) -- has a unit test

Utils: 
https://discourse.specifysoftware.org/t/shared-vs-embedded-records-configuration/1712/1


specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -40,6 +41,13 @@ import { InvalidMergeRecordsDialog } from './InvalidMergeRecords';
 import { mergingQueryParameter } from './queryString';
 import { Status } from './Status';
 
+export const recordMergingTables = new Set<keyof Tables>([
```
maxpatiiuk, line 44: query builder should also use this list of tables (don't know what list it does right now)

realVinayak: It uses this https://github.com/specify/specify7/blob/9044ffc1e29eea13e1d18ccf20cad10ef5290978/specifyweb/frontend/js_src/lib/components/Merging/definitions.ts#L5-L21. 
@CarolineDenis can you make them the same? The reason was because to prevent group members merge (and it is generic enough to handle cases for other tables)

specifyweb/frontend/js_src/lib/localization/merging.ts
```diff
@@ -254,4 +254,13 @@ export const mergingText = createDictionary({
...
     'uk-ua': "Пов'язані записи",
   },
+  localityMerging: {
+    'en-us': 'Locality merging',
+  },
+  paleoContextMerging: {
+    'en-us': 'Paleo context merging',
+  },
+  collectingEventMerging: {
+    'en-us': 'Collecting event merging',
```
grantfitzsimmons, line 264: Shouldn't these names be localized based on the table name assigned by the user in the schema? If possible, I think we should make it generalized so that it will be "`{Table Caption}` Merging" rather than  creating specific localization strings for each one.

CarolineDenis: yes this was just for a test :) 
I'll keep this comment open to not gorget to change that

maxpatiiuk:
> In the code above, relationship is a 'zero-to-one'.
> relationshipIsToMany(relationship) returns true because

zero-to-one is a "hack" relationship type introduced at some point in sp6 to do a quick fix for some user request, and now we have to live with the consequences (that's why quick fixes aren't a good idea most of the time)
I must admit, while Ben explained it to me several times, I didn't fully understand what zero-to-one means or why it's necessary. If I remember correctly, zero-to-one means "one-to-one" from one direction, but "many-to-many" in the reverse duration, or something like that, which is why in `relationshipIsToMany` we say `true` for `zero-to-one` fields.

But, @realVinayak or @melton-jason might be able to provide a better description that me based on their knowledge of back-end, business rules, databases and sp6

If it becomes really necessary, and we can't infer what zero-to-one means from the back-end code, we could contact Ben to ask him to explain it one more time

---

edit:

my messages to Ben about this:

> Ok, I implemented zero-to-one. Tell me if I did it correctly:
> they are handled as -to-many's:
> image
> but if you addded a -to-many item, you can not add another one (there is no Add button)
> 
image

Ben:
> That'll work. thanks

---

See also helpful information in https://specifydev.slack.com/archives/CC6V12D3J/p1617992659051300

---

Max:
> I just found a relationship type I haven't seen before: zero-to-one
> How should I treat these? are those like one-to-one but one directional?

Ben:

> Yes. I don't know why it is called that. It is kind of stupid. It is just a one-to-one.

melton-jason:
> might be able to provide a better description that me based on their knowledge of back-end, business rules, databases and sp6

https://github.com/specify/specify7/pull/4606#issuecomment-1980013013

The assesment you provided is correct to my knowledge. 

There are two `zero-to-one` relationships in Specify and both on the Locality table: 
`localityDetails` and `geoCoordDetails`. 

I don't think the backend of Specify 7 has any dedicated logic to handling `zero-to-one` relationships, besides marking the `to-"many"` side of the relationship as dependent on the `to-one` side.  
https://github.com/specify/specify7/blob/906d7397b4da911e7c5fcb421f9f4a7592e15583/specifyweb/specify/load_datamodel.py#L333-L336

At a database level, it is indistinguishable from a `one-to-many` relationship. 
That is, the LocalityDetail and GeoCoordDetail  tables contain a ForeignKey to the Locality table. There are no other constraints in place, so it is perfectly valid from the database's  perspective for a Locality to have more than one LocalityDetails or GeocoordDetails. 

However, for whatever reason the restriction was needed in Specify 6 to enforce the relationship as a `one-to-one`. (i.e. a maximum of one LocalityDetails/GeocoordDetails per Locality). 
This restriction is enforced and managed entirely within the application (MySQL or Mariadb has no concept of the relationship type or the restriction that is desired). 

In the Specify 6 code, the relationship was  denoted `ZeroOrOne`, but used `zero-to-one` when writing the `specify_datamodel.xml`. 

https://github.com/specify/specify6/blob/945d28f00288d2d64979848fb2fb63fda34bdd1b/src/edu/ku/brc/af/core/db/DBRelationshipInfo.java#L36

Although, when working with the ORM Specify6 uses (Hibernate), the columns still needed to be mapped as `OneToMany` (on Locality) and `ManyToOne` (on LocalityDetails/GeocoordDetails)

`localityDetails` relationship on Locality:
https://github.com/specify/specify6/blob/945d28f00288d2d64979848fb2fb63fda34bdd1b/src/edu/ku/brc/specify/datamodel/Locality.java#L1083-L1091

`locality` relationship on LocalityDetail:
https://github.com/specify/specify6/blob/945d28f00288d2d64979848fb2fb63fda34bdd1b/src/edu/ku/brc/specify/datamodel/LocalityDetail.java#L935-L940

It seems they were first added in https://github.com/specify/specify6/commit/3126d021d0496e6a30c1db31a8604f53f4594daf, but I can not find any documenttation or reasoning as to why this restriction was needed in the first place. 

The relationship can be approached conceptually as "A Locality can have zero or one associated LocalityDetail/GeocoordDetail records".

maxpatiiuk:
@melton-jason would it make sense to use the schemaoverwrites files to make sp7 front-end treat this as a -to-one relationship?
or would this cause issues because back-end would still return it as if it's a nested resources (thus will be an array rather than a direct resource in the API response)?
or should front-end act as if it's a one-to-many, but then have a business rule that causes a crash when more than one is added (causes a crash, so that it becomes obvious where in the code we forgot to restrict adding more than 1 resource for a zero-to-one)

specifyweb/frontend/js_src/lib/components/Merging/definitions.ts
```diff
@@ -18,4 +18,13 @@ export const recordMergingTableSpec: Partial<{
...
     dialogSpecificText: mergingText.agentContainsGroupDescription(),
   },
+  Locality: {
+    dialogSpecificText: mergingText.localityMerging(),
+  },
+  PaleoContext: {
+    dialogSpecificText: mergingText.paleoContextMerging(),
+  },
+  CollectingEvent: {
+    dialogSpecificText: mergingText.collectingEventMerging(),
```
CarolineDenis, line 28: @maxpatiiuk , 

I wanted to generate the strings dynamically using the table config labels but the data model is not instantiated yet when this file in run. Any other idea? 
Tried: 
`dialogSpecificText: `${
      strictGetTable('Locality').label
    } ${mergingText.merging()}`,
  },`
  Issue: this file calls getTables before the data model is being fetch
  
  Also tried: 
```
  dialogSpecificText: `${
     Object.keys(tables).length === 0 ? 'someStaticMessage' : strictGetTable('Locality').label
    } ${mergingText.merging()}`
```

Another question is, this dialogSpecificText was created for when the user tries to merge agent group, at the moment the text should not be displayed for locality - paleo and collecting event correct? So is there really a purpose of having a string here? should we just put an empty string?

maxpatiiuk: regardless of what checks or conditions you do, the fact of the matter is that this object is being created in the global scope, so it would be created as soon as the file is loaded (which would be during page load), before datamodel
and object is created just once - values are set only once, not on every access - it's not a function, so `Object.keys(tables).length === 0` doesn't make sense as this expression would be evaluated only once - on page load

if you want the expression to only be evaluated when it's first accessed, then simply turn it into a function:

```ts
  dialogSpecificText: () => mergingText.merging(tables.Locality.label),
```

(and change TypeScript type and usage accordingly to let you do this)

this is exactly what we do in user preferences definitions if you noticed - we are using "Taxon" as a subcategory heading, but datamodel might not be loaded yet

maxpatiiuk: you are correct. perharps we should make this a bit more clear

how about we change this type:

```ts
    readonly filterIgnore?: (
      resource: SerializedResource<Tables[TABLE_NAME]>
    ) => SerializedResource<Tables[TABLE_NAME]> | undefined;
    readonly dialogSpecificText: string;
```

to this:

```ts
    readonly unmergable?: {
      matches: (
        resource: SerializedResource<Tables[TABLE_NAME]>
      ) => boolean;
      message: LocalizedString;
    };
```

improvements:
1. specifying the message is not needed unless you also want to filter some records
2. names are more clear (filterIgnore -> unmergable.matches, dialogSpecificText -> message)
3. using LocalizedString over string
4. the function now returns a boolean rather than the resource itself - returning resource is "too powerful" as it's not clear just from function signature whether it can/would modify the resource or not - when it returns a boolean only, it's much more clear that it's just a small, pure, filter functions
5. in the future, if more than one different criteria for "unmergability" would exist, the unmergable object could just be converted into an array of objects to support multiple distinct messages

realVinayak: That's exactly what I was going to write! You beat me to it

realVinayak:
> Currently, it would be easiest for the frontend to treat `zero-to-one` relationships as a dependent `one-to-many` relationships. This is how the backend currently serializes and returns records with the relationship type: as an array containing one or no records.
> 
> While it would require modifications to both the frontend and backend, It might be easier to treat the relationship as a dependent `many-to-one` relationship, like the `collectionObjectAttribute` field on CollectionObject. Instead of returning an array with zero or one element, the value of the field would either be the serialized resource or `null`.
> 
> However, there is a small concern regarding the _name_ of the relationships: `localityDetails` and `geocoordDetails`. The pluralized name implies there can be more than one and that the relationship is a `to-many`. So perhaps the _most_ ideal solution is to make the relationship a true `one-to-many` and remove the one record maximum restriction. Of course, this would require compatibility with Specify 6 to be broken and assumes it makes sense to allow more than one LocalityDetail and GeocoordDetail record per Locality.

the current pr patches that for this. see relevant changes

specifyweb/frontend/js_src/lib/components/Merging/definitions.ts
CarolineDenis: :+1: thank you

specifyweb/frontend/js_src/lib/components/Merging/index.tsx
```diff
@@ -141,8 +141,8 @@ function RestrictMerge({
       records === undefined
```
maxpatiiuk, line 141: filterArray is no longer needed as the function no longer returns undefined

```ts
      records?.filter((record) =>
        recordMergingTableSpec[table.name]?.unmergable?.matches(
          record as never
        )
      )
```

maxpatiiuk (approved):
small simplification, otherwise looking good

emenslin (changes requested):
### Testing instructions

- perform same tests as agent merging but for new tables 
- [ ] Verify that we can’t merge PC and CE records when they are configured to be embedded

Looks good for the most part however I was able to merge embedded CE records, and when I tried to merge embedded PC records this happened. Since you aren't supposed to be able to merge I think the button should be grayed out instead of throwing an error when you try to do it. 

https://github.com/specify/specify7/assets/150940840/4097f7d3-74b2-4dba-8061-993c5c44b0dd

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -99,6 +100,20 @@ export function QueryResults(props: QueryResultsProps): JSX.Element {
...
+    hasTablePermission(table.name, 'update');
+  const canMergePaleoContext =
+    table.name === 'PaleoContext' && !isEmbeddedPaleoContext && canMerge;
+  const canMergeCollectingEvent =
+    table.name === 'CollectingEvent' && !isEmbeddedCollectingEvent && canMerge;
+  const canMergeOtherTables =
+    table.name !== 'PaleoContext' &&
+    table.name !== 'CollectingEvent' &&
+    canMerge;
+
```
realVinayak, line 116: Good work, getting there. This can be generalized. More specifically, doesn't make sense to merge one-to-one records. via the schema config, this happens to be a one-to-one to collection object. Instead of these conditionals, add a check if the table being merged happens to be in a one-to-one relationship with any other table. If so, don't merge.

emenslin (approved):
### Testing instructions

- perform same tests as agent merging but for new tables 
- [x] Verify that we can’t merge PC and CE records when they are configured to be embedded

Looks good, merge button is now removed from embedded PC and CE.

realVinayak (changes requested):

grantfitzsimmons (changes requested):
When merging several collecting event duplicates, the collectors are merged together, though each collector should only appear once on a single CE. It should automatically remove duplicate collector records with the same agent for the final merged record.

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -214,15 +229,16 @@ export function QueryResults(props: QueryResultsProps): JSX.Element {
...
-              hasTablePermission(table.name, 'update') && (
-                <RecordMergingLink
-                  selectedRows={selectedRows}
-                  table={table}
-                  onDeleted={handleDelete}
-                  onMerged={handleReRun}
-                />
-              )}
+            {(table.name === 'PaleoContext' && canMergePaleoContext) ||
+            (table.name === 'CollectingEvent' && canMergeCollectingEvent) ||
```
maxpatiiuk, line 233: when you define canMergePaleoContext or canMergeCollectingEvent you already check for table.name so you don't need to do so here

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
```diff
@@ -99,6 +100,20 @@ export function QueryResults(props: QueryResultsProps): JSX.Element {
...
+    hasPermission('/record/merge', 'update') &&
+    hasTablePermission(table.name, 'update');
+  const canMergePaleoContext =
+    table.name === 'PaleoContext' && !isEmbeddedPaleoContext && canMerge;
+  const canMergeCollectingEvent =
+    table.name === 'CollectingEvent' && !isEmbeddedCollectingEvent && canMerge;
+  const canMergeOtherTables =
+    table.name !== 'PaleoContext' &&
+    table.name !== 'CollectingEvent' &&
+    canMerge;
```
maxpatiiuk, line 115: could you move this logic into a function? you will give that function the table and it will return a boolean - based on the value of that boolean you will render or not RecordMergingLink

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
CarolineDenis: right!

maxpatiiuk (approved):

acwhite211:
> When merging several collecting event duplicates, the collectors are merged together, though each collector should only appear once on a single CE. It should automatically remove duplicate collector records with the same agent for the final merged record.

Here is the current unit test to Collecting Event.
https://github.com/specify/specify7/blob/64003757c17d1b9a1be76c95f05c77d15dd7eeec/specifyweb/specify/api_tests.py#L1300-L1367

@grantfitzsimmons I'm working on getting this working in the record merging code.  I've added to the unit test and just want to make sure these are the right test assertions that will show that this problem is solved:

```py
# Setup test collectors, where each collector points to the same agent
collector_1 = models.Collector.objects.create(
    isprintable=True, isprimary=True, ordernumber=1, agent=self.agent, collectingevent=collecting_event_1)
collector_2 = models.Collector.objects.create(
    isprintable=True, isprimary=True, ordernumber=1, agent=self.agent, collectingevent=collecting_event_2)
collector_3 = models.Collector.objects.create(
    isprintable=True, isprimary=True, ordernumber=1, agent=self.agent, collectingevent=collecting_event_3)

# ... after the merge request where collecting_event_2 if the old record and collecting_event_1 and collecting_event_3 are the new records.

# Assert that the collectors were updated correctly to the new collecting event.
# Also, assert that remove duplicate collector records with the same agent were deleted.
self.assertEqual(models.Collector.objects.get(agent=self.agent).count(), 1) # only one collector is left
self.assertEqual(models.Collector.objects.filter(id=collector_1.id).count(), 1) # assert collector_1 still exists
self.assertEqual(models.Collector.objects.filter(id=collector_2.id).count(), 0) # assert collector_2 does not exist
self.assertEqual(models.Collector.objects.filter(id=collector_3.id).count(), 0) # assert collector_3 does not exist
```

specifyweb/specify/record_merging.py
```diff
@@ -241,11 +245,35 @@ def record_merge_fx(model_name: str, old_model_ids: List[int], new_model_id: int
...
+                            if not protected_dm_table or not protected_dm_field or protected_dm_field.required:
+                                continue
+                            protected_model.objects.filter(**{protected_field: obj}).update(**{protected_field: None})
+                            try:
+                                obj.delete()
+                            except ProtectedError as e:
+                                raise
+                        else:
+                            raise
+
```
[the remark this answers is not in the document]
realVinayak: > If we did, then we would have to pass collection and agent into the these function, which already has enough parameters.

that's the least of the problems in this case.

> unless you can think of a case where delete_resource works when the current code does not?

what if the dependent have dependents? then you'll have to recursively delete.
case: (Collectionobject -> CollectingEvent in embedded). so, when merging collection objects, your code will fail.

not saying your code doesn't handle it (partially, your code does handle one-hop away dependents -- nothing beyond that) but just that it is reinventing the other function. I think it is worth doing it now (well, if I was aware of this function initially, I'd have done it as part of https://github.com/specify/specify7/pull/3839 for the initial release). The minute we add more tables, we'll have to revisit this, again -- and we'll come-up with the exact rule -- delete dependents on deletion, and cascade it -- just what the function does -- already. 

The dependents belonging to other objects (the old ids) will get deleted (by that function) when the main object gets deleted. So why not delete it via that function? Otherwise, you've the awkward logic in the current code.

Did some refactoring (without the messy collection and agent argument pass you brought up). see https://github.com/specify/specify7/pull/4940/commits/a158853dca756924ebc40140d4531e3547df5b45. Moved tests to a separate directory + file (it was 1500 LOC in api_tests, no way). the backend tests pass without any change.

realVinayak:
Hmm, https://github.com/specify/specify7/pull/4606#issuecomment-2121151200, wasn't frontend supposed to warn if duplicate collectors were being inserted? should have done so, as part of save blockers (uniqueness rules)

specifyweb/specify/api_tests.py
```diff
@@ -1234,6 +1234,220 @@ def _get_record_data(pre_merge=False):
         merged_data = fix_record_data(_get_record_data(True), target_model, target_model.name.lower(), new_record_id, old_record_ids)
 
         self.assertDictEqual(merged_data, _get_record_data())
+
+    def test_replace_localities(self):
```
realVinayak, line 1238: add more complicated unit tests, where data comes from the frontend. 

Have at least one unit test that triggers this function to actually fix orderings
https://github.com/specify/specify7/blob/82a3542153ba43c1cdfd4fec4bb60cc7654ceb44/specifyweb/specify/record_merging.py#L93

see https://github.com/specify/specify7/blob/82a3542153ba43c1cdfd4fec4bb60cc7654ceb44/specifyweb/specify/api_tests.py#L1004 where frontend throws bad ordering, but backend is smart enough to fix that (bc of that function).  In the data that frontend sends in the above unit test, check if order numbers are unique. at least one of this kind is necessary.

specifyweb/specify/record_merging.py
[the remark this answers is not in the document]
realVinayak: See https://github.com/specify/specify7/pull/4940, and let me know if it can be merged to this PR.

realVinayak (changes requested):
unit tests + https://github.com/specify/specify7/pull/4940

specifyweb/specify/record_merging.py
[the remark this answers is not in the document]
acwhite211: Ok, I see what you are saying now.  Ya, that looks good, I like how that simplifies the record merging code.  Go ahead and merge that branch into this PR

CarolineDenis:
> When merging several collecting event duplicates, the collectors are merged together, though each collector should only appear once on a single CE. It should automatically remove duplicate collector records with the same agent for the final merged record.

@grantfitzsimmons, 
Can you rephrase this? Can you provide the steps / examples to reproduce?

specifyweb/specify/record_merging.py
```diff
@@ -229,23 +233,18 @@ def record_merge_fx(model_name: str, old_model_ids: List[int], new_model_id: int
...
-                # Set new value for the field
-                setattr(obj, field_name_id, new_model_id)
+                try:
+                    # Set new value for the field
+                    setattr(obj, field_name_id, new_model_id)
+                except TypeError as e:
+                    if "Direct assignment to the reverse side of a related set is prohibited" in str(e):
+                        continue
+                    else:
+                        raise
```
realVinayak, line 247: hmm is this really needed? shouldn't we be skipping these relationships anyways? that is, in the place where you are filtering for relationships to the merge model, exclude to-many (from foreign to merge)?

realVinayak: also, possible, cache that result (see functools) -- the relationships to merge, as they won't change at runtime. but not a blocker, the above point is.

realVinayak (changes requested):
some things that'll be helpful to be clarified.

specifyweb/specify/record_merging.py
```diff
@@ -220,7 +221,10 @@ def record_merge_fx(model_name: str, old_model_ids: List[int], new_model_id: int
             field_name = col.lower()
             field_name_id = f'{field_name}_id'
             if not hasattr(foreign_model, field_name_id):
-                continue
+                if field_name.endswith('s') and hasattr(foreign_model, field_name):
+                    field_name_id = field_name
+                else:
+                    continue
```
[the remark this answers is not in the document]
realVinayak: why did they fail? this seems hacky don't you think?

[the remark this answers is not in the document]
realVinayak: can you mention the relationships which raise this exception?

realVinayak:
@acwhite211 see the simplifications. hack wasn't needed after all

specifyweb/specify/record_merging.py
```diff
@@ -221,10 +204,20 @@ def record_merge_fx(model_name: str, old_model_ids: List[int], new_model_id: int
...
+            # Example: handle case of updating a large amount of agent ids in the audit logs.
+            # Fix by optimizing the query by consolidating it here
+            if model_name.lower() in MERGING_OPTIMIZATION_FIELDS and \
+                table_name.lower() in MERGING_OPTIMIZATION_FIELDS[model_name.lower()]:
+                if field_name_id in MERGING_OPTIMIZATION_FIELDS[model_name.lower()][table_name.lower()]:
+                    query = Q(**{field_name_id: old_model_ids[0]})
+                    for old_model_id in old_model_ids[1:]:
+                        query.add(Q(**{field_name_id: old_model_id}), Q.OR)
+                    foreign_model.objects.filter(query).update(**{field_name_id: new_model_id})
+                    progress(1, 0) if progress is not None else None
```
realVinayak, line 220: @acwhite211  see https://github.com/specify/specify7/pull/4699, on why this was moved.

specifyweb/specify/tests/test_record_merging.py
```diff
@@ -34,6 +34,8 @@ def test_replace_agents(self):
             collectingevent=collecting_event
         )
 
+        self.collectionobjects[0].cataloger = agent_1
+        self.collectionobjects[0].save()
```
realVinayak, line 38: this is a unit test for https://github.com/specify/specify7/pull/4699.

realVinayak:
> NOTES: At this moment, the collector bug is fixed only when the records owe try to merge have identical data, meaning, if the created by agent is different or if an agent as a remark in one record and not the other it will not work

that case seems fine. if they aren't exact duplicates, it won't combine them into one. it flags duplicate agents now, so users can pick one.

grantfitzsimmons (changes requested):
image

When merging localities with Locality Detail records, clicking to view the record results in the detail being removed from the preview merge record. Once you try to add one, it crashes.

https://github.com/specify/specify7/assets/37256050/2063c07f-200d-4b9b-8a63-f6d27e06b6b0

[Specify 7 Crash Report - 2024-05-31T14_30_08.396Z.txt](https://github.com/user-attachments/files/15515146/Specify.7.Crash.Report.-.2024-05-31T14_30_08.396Z.txt)

realVinayak:
^ fixed: https://github.com/specify/specify7/pull/4606#pullrequestreview-2091001372

lexiclevenger (changes requested):
### Testing instructions

- [ ]  perform same tests as agent merging but for new tables 
- [x] Verify that we can’t merge PC and CE records when they are configured to be embedded
- [x] Test for https://github.com/specify/specify7/issues/4969 (see instructions in that issue) -- has a unit test

---

Embedded PC and CE records cannot be merged, and #4969 and https://github.com/specify/specify7/pull/4606#pullrequestreview-2091001372 appear to be fixed. 

However, when trying to merge records that have attachments, the merging process fails. This is the error message: 
[Merging e0747809-9f06-4e9b-a0c2-4926e54a847f Crash Report - 2024-06-03T21_30_06.739Z.txt](https://github.com/user-attachments/files/15539912/Merging.e0747809-9f06-4e9b-a0c2-4926e54a847f.Crash.Report.-.2024-06-03T21_30_06.739Z.txt)

This does not happen when merging Agent records with attachments in 7.9.5. Here, I try to merge two Agents, but it also happens with the other kinds of records.


https://github.com/specify/specify7/assets/164079735/f5e88373-f799-4cb8-ad18-b165e042d9d1

I tested this on KU_Fish_Demo and fwri_1_9_24.

emenslin (changes requested):
When you try to merge collecting events with collector conflicts the close button in the dialog acts as a save button. Also the dialog itself is not very intuitive and when you delete the merge collectors there are multiple like blank sections with just the lines. 

https://github.com/specify/specify7/assets/150940840/45e76422-7ec1-4536-82f4-4fd871aa6606

I was also not able to catch it on video but at one point when I deleted some of the merge collectors it ended up deleting the collectors in an individual CE record.

emenslin:
I managed to recreate it  deleting the collectors so here's the video of that from my previous comment


https://github.com/specify/specify7/assets/150940840/edaacd74-574d-4475-b302-454d2b79860a

emenslin:
In paleocontext even if the forms are exact copies the merge dialog appears like there are differences between the two. Not sure if this happens in other tables but I only noticed it with the pc table.
Screenshot 2024-06-04 123421

I cloned a pc record, made no changes, tried to merge and this happened

emenslin (changes requested):
Close button behavior looks better however collectors are still deleted 

https://github.com/specify/specify7/assets/150940840/780a3e37-b5bf-49c5-a6f6-6643e099337c

I was also able to find the pc issue mentioned before in the agent dialog so it seems like it might be in every table
Screenshot 2024-06-04 152014

alesan99 (changes requested):
- [ ] perform same tests as agent merging but for new tables 
- [x] Verify that we can’t merge PC and CE records when they are configured to be embedded
- [x] Test for https://github.com/specify/specify7/issues/4969 (see instructions in that issue) -- has a unit test

I also encountered a crash when removing all Locality Attachments when merging Localitites

https://github.com/specify/specify7/assets/8108710/73e6a308-53ba-4e5d-b713-b560e9568bd7

[Specify 7 Crash Report - 2024-06-05T14_05_15.914Z.txt](https://github.com/user-attachments/files/15592118/Specify.7.Crash.Report.-.2024-06-05T14_05_15.914Z.txt)

alesan99 (approved):
> I also encountered a crash when removing all Locality Attachments when merging Localitites

As I was writing up my issue it got fixed.
After testing it again its working fine now 👍

emenslin (approved):
### Testing instructions

- perform same tests as agent merging but for new tables 
- [x] Verify that we can’t merge PC and CE records when they are configured to be embedded
- [x] Test for https://github.com/specify/specify7/issues/4969 (see instructions in that issue) -- has a unit test

Collectors issue seems to be fixed, if deleted in the merge dialog they are not deleted from the individual record 
Screenshot 2024-06-05 095913

The show conflicting fields issue is in edge so I will write up a separate issue for that. 

Everything looks good but I would like another testing review just to make sure we caught everything since there's a lot to cover

lexiclevenger (approved):
### Testing instructions

- [x] perform same tests as agent merging but for new tables 
- [x] Verify that we can’t merge PC and CE records when they are configured to be embedded
- [x] Test for https://github.com/specify/specify7/issues/4969 (see instructions in that issue) -- has a unit test
---
Looks good! The requested changes also seem to be fixed.

Areyes42 (approved):
### Testing instructions

-   perform same tests as agent merging but for new tables
-  [x]  Verify that we can’t merge PC and CE records when they are configured to be embedded
-  [x]  Test for [Merging agents unchecks isCurrent checkbox in associated determinations #4969](https://github.com/specify/specify7/issues/4969) (see instructions in that issue) -- has a unit test
***
I encountered similar issues as mentioned in the previous comments when I first tested this PR, but it appears everything has been resolved now. Great job! 👏

acwhite211 (approved):

pashiav (approved):
### Testing instructions

- perform same tests as agent merging but for new tables 
- [X] Verify that we can’t merge PC and CE records when they are configured to be embedded
- [X] Test for https://github.com/specify/specify7/issues/4969 (see instructions in that issue) -- has a unit test

---

Looks good! Working as expected.

PR #4609, opened 2024-03-06
Better handle `createdByAgent` field for uniqueness rules 

### LINKED ISSUE #4608: Can save uniqueness rules when you shouldn't be able to
**Describe the bug**
When a uniqueness rule is scoped to 'created by agent' the save button is still orange and you can click save but it throws an error afterward. Not sure if this applies to other any other scopes. 

**To Reproduce**

1. Create a uniqueness rule 
2. Set the scope to 'created by agent'
3. Test uniqueness rule 

**Expected behavior**
It shouldn't let you save if there is a problem with the uniqueness rules. 

**Screenshots**

https://github.com/specify/specify7/assets/150940840/7b28d054-70ad-4a2c-ac05-77043ec4d8c2

**Crash Report**

If the bug resulted in an error message, please click on "Download Error
Message" and attach it here
[Specify 7 Crash Report - 2024-03-05T21_55_10.231Z.txt](https://github.com/specify/specify7/files/14501936/Specify.7.Crash.Report.-.2024-03-05T21_55_10.231Z.txt)

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge but also present in 7.9.3

melton-jason:
Fixes #4608 

This resolves two issues regarding `createdByAgent`. See https://github.com/specify/specify7/issues/4608#issuecomment-1981073302

Regarding setting `createdByAgent` so the frontend ORM can handle 
Conceptually, there may be some problems with the frontend setting `createdByAgent` before the record itself is created. Primarily in that the record _is not_ created. However, I do not think there is a practical difference in behavior or impact with this change on a data integrity viewpoint. 
(We can not do the same for `modifiedByAgent` because that field is not set when the record is created)

Regarding specifically handling the clone/carryforward for rules scoped to `createdByAgent`:
This solution seems hacky, but having `createdByAgent` as the terminating scope may be a case we need to handle. 
With respect to uniqueness rules: currently, the only fields _not_ being included in a clone/carry-forward are those which are scoped globally, or whose scopes end in one of the institutional scoping hierarchies. All other fields for uniqueness rules can be carried over to the new record regardless of scope. 
(See https://github.com/specify/specify7/pull/3996#issue-1888914474)

If there are other `to-one` relationships which would additionally need to covered, we can extract the relationships into a variable rather than hardcoding the fields. (much like `uniqueFields`)
https://github.com/specify/specify7/blob/906d7397b4da911e7c5fcb421f9f4a7592e15583/specifyweb/frontend/js_src/lib/components/DataModel/resource.ts#L263-L270

### Checklist

### Testing instructions
- For any table, create a uniqueness rule which is scoped to `createdByAgent`
- On a DataEntry form, attempt to invalidate the form by breaking the newly created uniqueness rule
  - This should properly result in an error
- Ensure the unique fields  for the rule are disabled in the `Configure fields to carry forward` Form Meta Menu
- Ensure that values in the fields are not being carried over when the resource is cloned or carry-forwarded.


CarolineDenis (approved):

realVinayak:
Hmm I don't think it makes sense to make a uniqueness rule to createdby (or going through createdby) in the first place, sounds like a stupid thing to do. Because, this would mean they need to switch users to make a record. Are there valid use cases of that? @grantfitzsimmons

grantfitzsimmons:
>Hmm I don't think it makes sense to make a uniqueness rule to createdby (or going through createdby) in the first place, sounds like a stupid thing to do. Because, this would mean they need to switch users to make a record. Are there valid use cases of that? @grantfitzsimmons

I'm a bit confused too. You can't create a second record in that table as that user once this is configured, correct?

>For any table, create a uniqueness rule which is scoped to createdByAgent

maxpatiiuk (commented):
fyi: if I remember back-end code correctly, it discards whatever createdByAgent front-end sent, and assigns it's own when you try to create a resource
back-end also doesn't let front-end modify this field after the fact

emenslin (approved):
Looks good, cannot clone or carry forward a field with a uniqueness rule, and the scoping works. 

https://github.com/specify/specify7/assets/150940840/2ef60c39-f1b2-431c-b526-2b08d8b9b4d4

bronwyncombs (approved):
Carry Forward limits selection to respect uniqueness rule: 
Screenshot 2024-03-11 at 12 16 38 PM

- Clone also does not include data for field with uniqueness rule defined 

Trying to enter duplicate field triggers error like expected: 

Screenshot 2024-03-11 at 12 17 44 PM

PR #4612, opened 2024-03-06
Disable delete btn and add btn when read only in formatter comp

### LINKED ISSUE #4610: Read-only view does not disable deleting formatter fields
https://github.com/specify/specify7/assets/37256050/82b7dced-7747-402b-98ad-532c7b7ade54

**Steps to recreate:**
1. Log in as `grant` on the `umherb_1_10_24` test panel database
2. Go to https://umherb11024-edge.test.specifysystems.org/specify/overlay/resources/app-resource/92/formatter/Determination/22
3. Edit the Determination table format and remove a field
4. See that you shouldn't have been able to click the trash can in the first place when it is in read-only view

Note that the changes do not persist, however read-only should preclude the user from deleting a field. No delete buttons should be clickable when in read-only view.

**System Information:**
[Specify 7 System Information - 2024-03-06T19_17_40.911Z.txt](https://github.com/specify/specify7/files/14514487/Specify.7.System.Information.-.2024-03-06T19_17_40.911Z.txt)

CarolineDenis:
Fixes #4610

### Checklist

### Testing instructions

Log in as grant on the umherb_1_10_24 test panel database
Go to https://umherb11024-edge.test.specifysystems.org/specify/overlay/resources/app-resource/92/formatter/Determination/22
Click on the Edit pencils next the Determination table format 
==> Verify that the content of the dialog is read only and that you cannot delete anything, reorder or add new field

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
@@ -151,8 +151,8 @@ function ConditionalFormatter({
               />
             </div>
             <span className="-ml-2" />
-            {trimmedFieldsLength > 0 && isExpanded ? (
-              <Button.Danger disabled={isReadOnly} onClick={handleDelete}>
+            {trimmedFieldsLength > 0 && isExpanded && !isReadOnly ? (
+              <Button.Danger onClick={handleDelete}>
                 {resourcesText.deleteDefinition()}
               </Button.Danger>
             ) : null}
@@ -173,9 +173,8 @@ function ConditionalFormatter({
           ) : null}
         </Label.Block>
       )}
-      {expandedNoCondition ? null : fields.length === 0 ? (
+      {expandedNoCondition || isReadOnly ? null : fields.length === 0 ? (
         <Button.Small
-          disabled={isReadOnly}
           onClick={(): void => {
             handleToggle();
             handleChanged(
@@ -215,21 +214,19 @@ function ConditionalFormatter({
             fields,
             (fields): void => handleChanged({ value, fields }, index),
           ]}
-          isReadOnly={isReadOnly}
           table={table}
         />
       ) : null}
       <span className="-ml-2 flex-1" />
-      <div className="inline-flex">
-        {trimmedFieldsLength === 1 || isExpanded ? null : (
+      {trimmedFieldsLength === 1 || isExpanded || isReadOnly ? null : (
+        <div className="inline-flex">
           <Button.Icon
-            disabled={isReadOnly}
             icon="trash"
             title={resourcesText.deleteDefinition()}
             onClick={handleDelete}
           />
-        )}
-      </div>
+        </div>
+      )}
       <div className="flex">
         {hasCondition && !isExpanded ? (
           <Button.Icon
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
@@ -152,7 +152,7 @@ function ConditionalFormatter({
             </div>
             <span className="-ml-2" />
             {trimmedFieldsLength > 0 && isExpanded ? (
-              <Button.Danger onClick={handleDelete}>
+              <Button.Danger onClick={handleDelete} disabled={isReadOnly}>
                 {resourcesText.deleteDefinition()}
               </Button.Danger>
             ) : null}
@@ -193,6 +193,7 @@ function ConditionalFormatter({
               index
             );
           }}
+          disabled={isReadOnly}
         >
           {resourcesText.addField()}
         </Button.Small>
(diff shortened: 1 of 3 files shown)


specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -152,7 +152,7 @@ function ConditionalFormatter({
             </div>
             <span className="-ml-2" />
             {trimmedFieldsLength > 0 && isExpanded ? (
-              <Button.Danger onClick={handleDelete}>
+              <Button.Danger disabled={isReadOnly} onClick={handleDelete}>
```
maxpatiiuk, line 155: instead of disabling, can we hide it?
pros:
- hiding it makes UI simpler
- we usually do disabling for things that could get enabled (i.e disabled because it's currently loading or because some field is empty). in this case, it's about user permissions, so that is more persistent (at very least won't change during the user's session), so hiding the element is a better option

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -175,6 +175,7 @@ function ConditionalFormatter({
       )}
       {expandedNoCondition ? null : fields.length === 0 ? (
         <Button.Small
+          disabled={isReadOnly}
```
maxpatiiuk, line 178: similar, hide it instead?

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -214,13 +215,15 @@ function ConditionalFormatter({
...
           ]}
+          isReadOnly={isReadOnly}
           table={table}
         />
       ) : null}
       <span className="-ml-2 flex-1" />
       <div className="inline-flex">
         {trimmedFieldsLength === 1 || isExpanded ? null : (
           <Button.Icon
+            disabled={isReadOnly}
```
maxpatiiuk, line 226: hide it?

specifyweb/frontend/js_src/lib/components/Formatters/Fields.tsx
```diff
@@ -25,9 +25,11 @@ import type { Formatter } from './spec';
 export function Fields({
   table,
   fields: [fields, setFields],
+  isReadOnly,
```
maxpatiiuk, line 28: don't pass read-only as a prop. use this instead:

```ts
```

by using ReadOnlyContext everywhere, every component is able to handle being used in a read-only environment in a consistent way - better for code reuse too

specifyweb/frontend/js_src/lib/components/Formatters/Fields.tsx
```diff
@@ -99,6 +101,7 @@ export function Fields({
       )}
       <div className="flex gap-2">
         <Button.Secondary
+          disabled={isReadOnly}
```
maxpatiiuk, line 104: hide it

specifyweb/frontend/js_src/lib/components/Formatters/Fields.tsx
```diff
@@ -187,6 +191,7 @@ function Field({
       <td>
         <Button.Small
           aria-label={commonText.remove()}
+          disabled={isReadOnly}
```
maxpatiiuk, line 194: hide it and the following

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
CarolineDenis: yes, you're totally right

CarolineDenis:
Test 1
Distinct: 
- [ ]  In the query builder results, enabling the distinct option should group rows with identical data.
- [ ]  For distinct queries, a link icon will appear at the front of each row with a single ID. Clicking this icon will navigate the user to a new page containing the resource form.
- [ ]  For rows with multiple IDs, a list icon will be displayed at the front. Clicking this icon will open a dialog with a "Browse in form" option, allowing users to slide between all records concatenated in the row.
- [ ]  Users can select individual rows in distinct query results.
- [ ]  Users can select individual rows for export to CSV in distinct query results.
- [ ]  Users can export all rows to CSV in distinct query results.

Non distinct:
- [ ]  In the query builder results, disabling the distinct option should not group rows with identical data.
- [ ]  For non distinct queries, a link icon will appear at the front of each row with a single ID. Clicking this icon will navigate the user to a new page containing the resource form.
- [ ]  Users can select individual rows in non distinct query results.
- [ ]  Users can select individual rows for export to CSV in non distinct query results.
- [ ]  Users can export all rows to CSV in non distinct query results.

Notes:

bronwyncombs (approved):
read-only permissions:
Screenshot 2024-03-07 at 2 46 31 PM

can't delete formatters, as expected: 
Screenshot 2024-03-07 at 2 47 16 PM

maxpatiiuk:
> Looks good, table is read only and you cannot edit or delete anything. Screenshot 2024-03-07 121151

Hmm, why do these screenshots still include a delete button in read-only mode? Should it be removed? @CarolineDenis

maxpatiiuk (approved):
all good (besides https://github.com/specify/specify7/pull/4612#issuecomment-1984917741)

PR #4618, opened 2024-03-08
Create Custom Django Model Save Function to Allow Overwritting Timestamp Fields

### LINKED ISSUE #1806: WB Modified Date not saved after upload
image

---

**Expected Behavior:**
* When a column within a specified table is linked to either `timestampModified` or `timestampCreated`, it should not be automatically replaced upon uploading new data. 
* In cases where a column is not associated with `timestampModified` or `timestampCreated`, the system should autonomously allocate both fields with the current timestamp. 
* When new records are generated in other tables during the same upload process, if no column is directly linked to `timestampModified` or `timestampCreated`, the system should automatically assign both fields with the current timestamp.
* `timestampModified` or `timestampCreated` should be assigned the current timestamp in all cases during WorkBench uploads **except** when a column is mapped to that field where that row contains a value

acwhite211:
Fixes #1806

Remove the `auto_now_add` attribute for the `timestampCreated` field, and the `auto_now` attribute from the `timestampModified` field.  Replace the functionality of these attributes by overriding the Django model `save()` function with custom logic that allows for these fields to be set in the save function without being overwritten by the current time, but will otherwise use the current time if the field is not specified.  The custom timestamp save function will be used by any model using the timestamp fields, which are most tables in Specify.

The changes being made to the two timestamp fields are detected using the `django-model-utils` library.  This provides a succinct way to detect changes to a fields and provides much more useful Django functionality that could be useful in the future.

This PR implements the timestamp save function as both a separate function set in the `build_models.py` file, but also as an abstract Django model.  In the near future, when the Specify Django models are defined statically, instead of dynamically created from the Specify 6 xml file, this abstract model class will be a better way of overwrite many Django model save functions through inheritance.

### Checklist

- [x] Add automated tests
- [ ] UI Testing

### Testing instructions
Upload data in the workbench with `timestampCreated` and `timestampModified` fields.  Then check to make sure those values are indeed the same values that were uploaded, and were not overwritten by the current time.

Make sure to go into preferences -> WorkBench -> No restrictions mode, and set it to true.  That way you will be able to map columns in the workbench to the timestamp fields, otherwise they are hidden, even when checking "Reveal Hidden Form Fields"

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,3 +9,4 @@ PyJWT==2.3.0
 django-auth-ldap==1.2.15
 jsonschema==3.2.0
 typing-extensions==4.3.0
+django-model-utils==4.4.0
--- a/specifyweb/specify/api.py
+++ b/specifyweb/specify/api.py
@@ -377,8 +377,10 @@ def cleanData(model, data: Dict[str, Any], agent) -> Dict[str, Any]:
             pass
 
     # timestampcreated should never be updated.
+    # ... well it is now ¯\_(ツ)_/¯
     try:
-        del cleaned['timestampcreated']
+        # del cleaned['timestampcreated']
+        pass
     except KeyError:
         pass
 
--- a/specifyweb/specify/api.py
+++ b/specifyweb/specify/api.py
@@ -378,6 +378,7 @@ def cleanData(model, data: Dict[str, Any], agent) -> Dict[str, Any]:
 
     # timestampcreated should never be updated.
     # ... well it is now ¯\_(ツ)_/¯
+    # New requirments are for timestampcreated to be overridable.
     try:
         # del cleaned['timestampcreated']
         pass
--- a/specifyweb/specify/api_tests.py
+++ b/specifyweb/specify/api_tests.py
@@ -2,6 +2,7 @@
 Tests for api.py
 """
 
+from calendar import c
 from datetime import datetime
 import json
 from unittest import skip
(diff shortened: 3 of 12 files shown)


acwhite211:
Going to end my work on this for today, but the basic idea so far to is override the save function such that if a model has a timestamp field, then it will default implement the behavior of `auto_now`, even though the timestamped Django field doesn't have the attribute.  For times when you don't want this behavior, call the save function like this `record.save(timestamp_override=True)`.  I went ahead and changed those save function calls that are in the Django workbench app, but the next thing to do is edit this so that the modified save function call only occurs when one of the timestamp fields is being explicitly edited (or maybe we'll just do this for all workbench calls).

acwhite211:
I found the module `django-model-utils` which allows for tracking field changes before saving, thus creating an easy way to dynamically change saving behavior based on which fields have been changed.

acwhite211:
Got the functionality working with `timestampmodified`.  There is an issue with `timestampcreated` during the call `record.create(...)`.  It's not getting picked up by the Field Tracker in this case.  I'm experimenting with overriding the `models.Manager -> create()` functionality.

CarolineDenis (approved):

emenslin (approved):
### Testing instructions

- [x] Upload data in the workbench with `timestampCreated` and `timestampModified` fields.  Then check to make sure those values are indeed the same values that were uploaded, and were not overwritten by the current time.
- [x] Make sure to go into preferences -> WorkBench -> No restrictions mode, and set it to true.  That way you will be able to map columns in the workbench to the timestamp fields, otherwise they are hidden, even when checking "Reveal Hidden Form Fields"

Looks good, timestampCreated and timestampModified do not get overwritten by the current time 
Screenshot 2024-05-13 151750

alesan99 (approved):
- [ ] Upload data in the workbench with `timestampCreated` and `timestampModified` fields.  Then check to make sure those values are indeed the same values that were uploaded, and were not overwritten by the current time.

- [x] Make sure to go into preferences -> WorkBench -> No restrictions mode, and set it to true.  That way you will be able to map columns in the workbench to the timestamp fields, otherwise they are hidden, even when checking "Reveal Hidden Form Fields"

Looks good, successfully uploaded

EXCEL_qZoKJzRT0K

sharadsw (approved):

PR #4624, opened 2024-03-09
Fix page size selector taking up a disproportionate amount of space

### LINKED ISSUE #4623: Page size list takes up a disproportionate amount of space
**Describe the bug**
The page size pick list takes up a disproportionate amount of space, which makes it a bit of an eyesore 

https://github.com/specify/specify7/assets/37256050/ece66899-2a6b-4117-9367-3cf2c2940343

```tsx
   <Select
              className={pageSize === infinity ? 'w-auto' : 'w-16'}
              value={pageSize}
              onValueChange={(rawNewPageSize): void => {
                const newPageSize = Number.parseInt(rawNewPageSize);
                const currentIndex = currentPage * pageSize;
                setPageSize(newPageSize);
                setCurrentPage(Math.floor(currentIndex / newPageSize));
              }}
```

Despite the `w-auto` or `w-16` class being applied based on the `pageSize`, it is overwritten by the `w-full` class that is applied. This can be solved by either making `w-auto` and `w-16` important or by removing the `w-full` class that is applied to `select` elements here:

https://github.com/specify/specify7/blob/74fe3adfa28086dc5fc3cbfa146ec45f1c39b833/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx#L323-L325

This would require a larger examination on what will be affected. I will open a PR that uses a simple important flag to resolve this, but after doing some testing, removing the `w-full` from those lines in `Form.tsx` did not result in unexpected behavior on select elements on the forms.

image

grantfitzsimmons:
Since we should use these flags sparingly, I'd like to hear from a dev if removing `w-full` from the `select` style in `Form.tsx` would be a better long-term solution

Fixes #4623

Left is this branch, right is the current `edge` behavior:

https://github.com/specify/specify7/assets/37256050/00ecc7e3-1275-449d-b7c7-01c61191bbf1

### Checklist

### Testing instructions
1. Verify that the page size list takes up an appropriate amount of room in the
   - [ ] Record set dialog
   - [ ] Queries dialog 
2. Make sure the paginator appears correctly in
   - [ ] Browse in forms
   - [ ] Record sets

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Molecules/Paginator.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Molecules/Paginator.tsx
@@ -48,7 +48,7 @@ export function usePaginator(
           )}
           <div className="flex flex-1 justify-end">
             <Select
-              className={pageSize === infinity ? 'w-auto' : 'w-16'}
+              className={pageSize === infinity ? '!w-auto' : '!w-16'}
               value={pageSize}
               onValueChange={(rawNewPageSize): void => {
                 const newPageSize = Number.parseInt(rawNewPageSize);


CarolineDenis (approved):

emenslin (approved):
1. Verify that the page size list takes up an appropriate amount of room in the
- [x]  Record set dialog
- [x]  Queries dialog
2. Make sure the paginator appears correctly in
- [x] Browse in forms 
- [x] Record sets

Looks good!

bronwyncombs (approved):
Beautiful

https://github.com/specify/specify7/assets/135047322/cad61c02-80ac-4613-ba5d-2d438a99cc81

https://github.com/specify/specify7/assets/135047322/fadc8ce7-8cc6-400a-b9de-1f11c8179d12

--- 

Pagination in browse in forms and record sets works and displays as expected

maxpatiiuk:
Could someone please also test how this affects cases when the selector has long content, longer than the size available
I believe w-full was there to prevent long pick list values from breaking forms 
(Though, in ideal world, we would have screenshot tests that would show how this change would improve or worsen the UI in various places)

PR #4631, opened 2024-03-09
Add padding between the add field button in the table format dialog

### LINKED ISSUE #4619: Add Field button touching separator box
**Describe the bug**
When you open a table formatter there is no space between the last separator row and the `add field` button. 

**To Reproduce**

1. Go to app resources 
2. Click on `record formatters` 
3. Open the default accession formatter
4. Scroll down to the `add field` button 
5. See error

**Expected behavior**
There should be a gap between the last separator and the `add field` button. 

**Screenshots**
Screenshot 2024-03-08 145936

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge
- Database Name: KUFish51623

grantfitzsimmons:
Fixes #4619

Also changes miscellaneous paddings to make it visually consistent

image

image

### Checklist

### Testing instructions
- [ ] Test to make sure #4619 is fixed
- [ ] Make sure padding is symmetrical throughout the table format editor
- [ ] Test both conditional table formats and non-conditional table formats in the editor
- [ ] Make sure the user interface is intuitive


bronwyncombs (dismissed):
Add field button looks good now! The rest of the padding seems symmetrical and consistent with regular and conditional formats. 

One request ☝️ 🤓 , if we could move the customize field format checkbox back to the right that'd be awesome

Screenshot 2024-03-11 at 9 35 50 AM

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -124,9 +124,9 @@ function ConditionalFormatter({
 
   return (
     <div
-      className={`flex
-        ${isExpanded || !hasCondition ? 'flex-col' : 'items-center'}
-        ${isExpanded ? 'gap-2' : ''}
+      className={`flex gap-2
+        ${isExpanded || !hasCondition ? 'flex-col' : ' items-center pb-2 pt-2'}
+        ${isExpanded ? 'gap-2 pt-2' : ''}
```
CarolineDenis, line 129: gap-2 is not necessary in line 129, you are applying it twice. 
Could you also move the pt-2 to line 127 ?
If yes it would remove the need of line 129 and the need of pt-2 in line 128.

specifyweb/frontend/js_src/lib/components/Formatters/Fields.tsx
```diff
@@ -117,7 +117,6 @@ export function Fields({
           >
             {resourcesText.addField()}
           </Button.Secondary>
-          <span className="-ml-2 flex-1" />
```
CarolineDenis, line 120: this was needed to have the customize field formatters check box on the bottom right of component. Why moving it next to the add field button?

CarolineDenis: Screenshot 2024-03-11 at 7 42 26 AM
Screenshot 2024-03-11 at 7 43 06 AM

grantfitzsimmons: Bronwyn prefers it the way it was as well, so I will return it.

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -68,7 +68,7 @@ export function Definitions({
         />
       ))}
       {!isReadOnly && hasCondition ? (
-        <div>
+        <div className="pt-4">
```
maxpatiiuk, line 71: Instead of padding, could you add gap-4 tot he parent element?

we generally use gap for spacing as it’s less bug prone than adding (gap is symmetrical regardless of how many children are inserted, where as padding is trickier to get right as you have to account for first or last element not needing padding)
We generally use padding for things like making buttons larger or adding symmetrical white space from  all directions in a container

maxpatiiuk (approved):
If the same can be accomplished using gap rather than padding, that would be awesome - otherwise ok

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
grantfitzsimmons: Due to `divide-y`, `gap-4` makes things really... unaesthetic.

image
image

I wanted to do this initially but settled on the current solution. Thoughts?

grantfitzsimmons: Due to `divide-y`, `gap-4` makes things really... unaesthetic.

image
image

I wanted to do this initially but settled on the current solution. Thoughts?

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
```diff
@@ -124,9 +124,9 @@ function ConditionalFormatter({
 
   return (
     <div
-      className={`flex
-        ${isExpanded || !hasCondition ? 'flex-col' : 'items-center'}
-        ${isExpanded ? 'gap-2' : ''}
+      className={`flex gap-2 pt-2
+        ${isExpanded || !hasCondition ? 'flex-col' : ' items-center pb-2'}
+        ${isExpanded ? '' : ''}
```
CarolineDenis, line 129: this line should be removed now that you have empty string in each branch

grantfitzsimmons: Fixed in https://github.com/specify/specify7/pull/4631/commits/a6835234a9eb538b0b9804213e150d5f6d207d44

emenslin (approved):
### Testing instructions
- [x] Test to make sure #4619 is fixed
- [x] Make sure padding is symmetrical throughout the table format editor
- [x] Test both conditional table formats and non-conditional table formats in the editor
- [x] Make sure the user interface is intuitive 

Everything looks nice and symmetrical now both with and without conditional formatters! 
Screenshot 2024-03-12 091624

pashiav (approved):
- [X] Test to make sure #4619 is fixed
- [X] Make sure padding is symmetrical throughout the table format editor
- [X] Test both conditional table formats and non-conditional table formats in the editor
- [X] Make sure the user interface is intuitive  

Looks good! The padding reduces misclicks for the user.

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
grantfitzsimmons: I'll mark this as resolved for now, but if you have a solution in mind in the future we can open a new PR and resolve it. Thank you @maxpatiiuk !

CarolineDenis (approved):

emenslin (approved):
### Testing instructions
- [x] Test to make sure #4619 is fixed
- [x] Make sure padding is symmetrical throughout the table format editor
- [x] Test both conditional table formats and non-conditional table formats in the editor
- [x] Make sure the user interface is intuitive

specifyweb/frontend/js_src/lib/components/Formatters/Definitions.tsx
maxpatiiuk: ah good point, I forgot about the divide
this is the best solution then, and similar to what we do in the workbench mapper and query builder
the other option is to insert a div with height of 1 in between each item to simulate "border", but your solution is simpler

specifyweb/frontend/js_src/lib/components/Formatters/Fields.tsx
```diff
@@ -54,7 +54,7 @@ export function Fields({
            *   table layout with list layout
            */
           className={`
-            grid-table min-w-[35rem] gap-x-4 gap-y-4
+            grid-table min-w-[35rem] gap-x-2 gap-y-2
```
maxpatiiuk, line 57: minor thing, but just a fyi that this is equivalent:

```suggestion
            grid-table min-w-[35rem] gap-2
```

PR #4641, opened 2024-03-12
Uniquely identify saveBlockers with keys

### LINKED ISSUE #4638: Uniqueness Check Failure with Group Validation Rules
CSIRO had a chance to test the fix in https://github.com/specify/specify7/pull/4593 on the `edge` Docker tag. Yesterday, Dan discovered an issue:

> Hi Grant,
> I have just done some testing and the good news is that I can't upload duplicates via the WB! However I did spot an issue that needs to be looked at:
> 
> I created 3 uniqueness rules, two independent checks for text1 and text2, and a group check for text3/4/5: 
> 
> image
> 
> 
> The WB gave me the correct errors for the various combinations, but I was able to get the form to not pick up on a duplicate value. I didn't see the issue when I just add the rules for text1 and text2, so I'm wondering if the bug is when a group check is also involved?
> 
> Thankfully the system gives an error and doesn't let you save the data, but it would be frustrating for a user because they'd lose their input for that specimen and have to start over. I took a video showing what I did.
>
> **Video:**
> [Uniqueness UI bug.mp4](https://drive.google.com/open?id=1Luh0FWwH12Heu7N-xcYpEOwYWjmDZyrq&usp=drive_fs)

**Reported by:** [CSIRO on Asana](https://app.asana.com/0/1201972579963460/1206531432191019/f)

melton-jason:
Fixes #4638 

See https://github.com/specify/specify7/issues/4638#issuecomment-1991752196

### Checklist

- [ ] Add automated tests

### Testing instructions

For any table: 
- [ ] Create or find a uniqueness with 1 or more uniqueness fields
- [ ] For the same table, create a uniqueness rule with at least two fields (with at least one field differing from the previous rule) with the same scope as the previous rule
- [ ] On the DataEntry form for the table, invalidate both rules
- [ ] Change the field value for a rule so that one uniqueness rule becomes valid while the other remains invalid
- [ ] Ensure the saveBlocker for the invalid rule (and any other saveBlockers on the form) is still set

For an example, see https://github.com/specify/specify7/issues/4638#issuecomment-1991752196

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
@@ -1,11 +1,8 @@
-import type { LocalizedString } from 'typesafe-i18n';
-
 import type { ResolvablePromise } from '../../utils/promise';
 import { flippedPromise } from '../../utils/promise';
 import type { IR, RA } from '../../utils/types';
 import { filterArray, overwriteReadOnly } from '../../utils/types';
 import { removeKey } from '../../utils/utils';
-import { formatConjunction } from '../Atoms/Internationalization';
 import { softFail } from '../Errors/Crash';
 import { isTreeResource } from '../InitialContext/treeRanks';
 import type { BusinessRuleDefs } from './businessRuleDefs';
@@ -13,7 +10,11 @@ import { businessRuleDefs } from './businessRuleDefs';
 import { backboneFieldSeparator, djangoLookupSeparator } from './helpers';
 import type { AnySchema, AnyTree, CommonFields } from './helperTypes';
 import type { SpecifyResource } from './legacyTypes';
-import { propagateBlockerEvents, setSaveBlockers } from './saveBlockers';
+import {
+  getFieldBlockerKey,
+  propagateBlockerEvents,
+  setSaveBlockers,
+} from './saveBlockers';
 import { specialFields } from './serializers';
 import type { LiteralField, Relationship } from './specifyField';
 import type { Collection, SpecifyTable } from './specifyTable';
@@ -71,8 +72,8 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
     this.fieldChangePromises[processedFieldName] = thisCheck;
 
     const checks: RA<Promise<BusinessRuleResult<SCHEMA> | undefined>> = [
+      ...this.checkUnique(processedFieldName),
       this.invokeRule('fieldChecks', processedFieldName, [this.resource]),
-      this.checkUnique(processedFieldName),
       isTreeResource(this.resource as SpecifyResource<AnySchema>)
         ? treeBusinessRules(
             this.resource as SpecifyResource<AnyTree>,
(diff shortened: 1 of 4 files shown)


CarolineDenis (approved):

emenslin (approved):
### Testing instructions

For any table: 
- [x] Create or find a uniqueness with 1 or more uniqueness fields
- [x] For the same table, create a uniqueness rule with at least two fields (with at least one field differing from the previous rule) with the same scope as the previous rule
- [x] On the DataEntry form for the table, invalidate both rules
- [x] Change the field value for a rule so that one uniqueness rule becomes valid while the other remains invalid
- [x] Ensure the saveBlocker for the invalid rule (and any other saveBlockers on the form) is still set

bronwyncombs (changes requested):
For **Collection Object Attribute** table: 
- [X] Create or find a uniqueness with 1 or more uniqueness fields
- [X] For the same table, create a uniqueness rule with at least two fields (with at least one field differing from the previous rule) with the same scope as the previous rule
Screenshot 2024-03-13 at 9 45 24 AM

- [ ] On the DataEntry form for the table, invalidate both rules

it consistently did not trigger uniqueness rules and sent to 404 on save: 

https://github.com/specify/specify7/assets/135047322/2582a1a4-d892-41b0-bacc-34dd99788117


- [ ] Change the field value for a rule so that one uniqueness rule becomes valid while the other remains invalid
- [ ] Ensure the saveBlocker for the invalid rule (and any other saveBlockers on the form) is still set

bronwyncombs (approved):
👍 looks good!!

For any table -- Coll Obj Attribute: 
- [X] Create or find a uniqueness with 1 or more uniqueness fields
- [X] For the same table, create a uniqueness rule with at least two fields (with at least one field differing from the previous rule) with the same scope as the previous rule
- [X] On the DataEntry form for the table, invalidate both rules
Screenshot 2024-03-14 at 9 49 42 AM
Screenshot 2024-03-14 at 9 49 48 AM

- [X] Change the field value for a rule so that one uniqueness rule becomes valid while the other remains invalid
Screenshot 2024-03-14 at 9 51 17 AM

- [X] Ensure the saveBlocker for the invalid rule (and any other saveBlockers on the form) is still set

realVinayak:
Curious why the previous method didn't work. Can you explain that?

specifyweb/frontend/js_src/lib/components/DataModel/saveBlockers.tsx
```diff
@@ -69,7 +67,7 @@ export const propagateBlockerEvents = (
 export function useSaveBlockers(
```
maxpatiiuk, line 67: the conflict with save blockers that resulted in the need for blockerKey, was that a conflict between multiple places setting blockers, or the same place overwriting the blockers it set before?

if it was a conflict between multiple different places, then in the useSaveBlockers, you could have just used `const id = useId('blocker'); const blockerKey = id(field?.name ?? '')` as your blocker key

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -183,7 +193,9 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
       )
     );
 
-    const results = rulesToCheck.map(async ({ rule }) => this.uniqueIn(rule));
+    const results: RA<Promise<BusinessRuleResult<SCHEMA>>> = rulesToCheck.map(
+      async ({ rule }) => this.uniqueIn(rule)
+    );
     void Promise.all(results).then((results) =>
```
maxpatiiuk, line 199: huh, it's weird that a function returns an array of promises rather than a single promise from Promise.all
why not do this instead?

```ts
const results = await Promise.all(results);

results.flatMap ... do something with results

return results;
```

also, since your checkUnique doesn't await the promise before returning the result, it may be possible for it to be called again before the previous call completed, thus creating a race condition

specifyweb/frontend/js_src/lib/components/DataModel/saveBlockers.tsx
```diff
@@ -113,42 +111,26 @@ export function useSaveBlockers(
 export function setSaveBlockers(
```
maxpatiiuk, line 111: are blockers ever set for more than one field in a single function call?
if not, then why not just prepend the blocker key with the field name automatically inside of setSaveBlockers rather than forcing every place to call getFieldBlockerKey (or risk forgetting to call it and thus creating a bug)

maxpatiiuk (commented):
Thanks for these changes, and I do regret removing blocker keys a few years ago.
I also like that blocker keys actually reduced line count a bit.
But could you please address my comments in a small follow-up PR if possible? (sorry, I couldn't review it earlier)

PR #4652, opened 2024-03-14
Check all 6 collections when select set as admin for 6

### LINKED ISSUE #2695: Setting a user as admin does not set their 6 collections

CarolineDenis:
Fixes #2695

### Checklist

### Testing instructions

- go in security and accounts 
- click on make admin button under specify 6 permissions 
- open set collections under specify 6 permissions 
==> verify that all collections are checked
==> verify that you can unchecked some collections and save


CarolineDenis:
@acwhite211, 

There is an issue with the api, 
`/api/set_admin_status/${user.id}/` seems to not be correctly setting the Sp5 admin status. Can you check this? 
To reproduce: 
go in security and accounts
click on **make admin** button under **specify 6 permissions**
see that the button is now **remove admin** 
open **set collections** under **specify 6 permissions**
see that all collections are checked
refresh page 
see that the button is back to **make admin**

specifyweb/frontend/js_src/lib/components/Security/AdminStatusPlugin.tsx
```diff
@@ -28,7 +31,17 @@ export function AdminStatusPlugin({
   const loading = React.useContext(LoadingContext);
   const [user] = useResource(resource);
   const isCurrentUser = userInformation.id === user.id;
-
+  const [allCollections] = useAsyncState(
+    React.useCallback(
+      async () =>
+        fetchCollection('Collection', { limit: 0, domainFilter: false }).then(
```
maxpatiiuk, line 37: no need to fetch collections again as you can just pass it down as a prop from the parent component - https://github.com/specify/specify7/blob/aa434e5d24f9b88b3a2c2d892d8619f5d35e815e/specifyweb/frontend/js_src/lib/components/Security/User.tsx#L164

acwhite211:
It seems the `set_admin_status` api adds the SpPincipalID of an Administator (which is 1), with the SpecifyUserID, to specifyuser_spprincipal.  This api seems to be working properly, and can be confirmed by the `is_admin` function.
```sql
-- example for setting the new user, with ID 12, to admin

-- set_admin
SELECT spprincipalid, 12
FROM   spprincipal
WHERE  name = 'Administrator';

-- is_admin - return 1 if true, no results if false
SELECT 1
FROM   specifyuser_spprincipal,
       spprincipal
WHERE  12 = specifyuser_spprincipal.specifyuserid
       AND specifyuser_spprincipal.spprincipalid = spprincipal.spprincipalid
       AND spprincipal.name = 'Administrator'
LIMIT  1;

-- list_admins
SELECT specifyuserid, specifyuser.name
FROM   specifyuser_spprincipal
       JOIN spprincipal USING (spprincipalid)
       JOIN specifyuser USING (specifyuserid)
WHERE  spprincipal.name = 'Administrator';

-- get users_collections_for_sp6
SELECT DISTINCT c.usergroupscopeid, c.collectionname
FROM   collection c
       INNER JOIN spprincipal p
               ON p.usergroupscopeid = c.usergroupscopeid
       INNER JOIN specifyuser_spprincipal up
               ON up.spprincipalid = p.spprincipalid
       INNER JOIN specifyuser u
               ON u.specifyuserid = up.specifyuserid
                  AND p.grouptype IS NULL
WHERE  up.specifyuserid = 12;  
```
The other step after setting the user to Admin is to add the collections the user has access to via PUT api call to `/context/user_collection_access_for_sp6/{user_id}/`

It seems that the problem in that call the `user_collection_access_for_sp6` PUT api removes the user admin entry in the `specifyuser_spprincipal` table.  Here is the line of code where this is done:

 https://github.com/specify/specify7/blob/c4f634c88caa2497ad302d91e2f80bf295d2c773/specifyweb/context/views.py#L76-L81

Here are the sql statements that are made by the `set_users_collections_for_sp6` function:
```sql
-- First delete the mappings from the user to the principals.
DELETE specifyuser_spprincipal
FROM   specifyuser_spprincipal
       JOIN spprincipal using (spprincipalid)
WHERE  specifyuserid = 17
       AND usergroupscopeid NOT IN ( 65536, 32768, 4 );

-- Next delete the joins from the principals to any permissions.
DELETE FROM spprincipal_sppermission
WHERE  spprincipalid NOT IN (SELECT spprincipalid
                             FROM   specifyuser_spprincipal);

-- Finally delete all the principals that aren't connected to
-- any user. This should just be the ones where the mappings
-- were deleted above.
DELETE FROM spprincipal
WHERE  grouptype IS NULL
       AND spprincipalid NOT IN (SELECT spprincipalid
                                 FROM   specifyuser_spprincipal);

-- Now to add any new principals. Which ones alerady exist?
SELECT usergroupscopeid
FROM   spprincipal
       JOIN specifyuser_spprincipal using (spprincipalid)
WHERE  grouptype IS NULL
       AND specifyuserid = 17;

-- Add a principal for collection ids that don't already exist
UPDATE spprincipal set usergroupscopeid = %s WHERE spprincipalid = %s;
INSERT specifyuser_spprincipal(SpecifyUserID, SpPrincipalID) VALUES (17, %s);
```

I think this might of just been an oversight, so I'm adding an extra condition to that query statement to exclude Admin principals.  We might want to design a more comprehensive solution, but this at least avoids removing the admin status.
```sql
DELETE specifyuser_spprincipal
FROM   specifyuser_spprincipal
       JOIN spprincipal using (spprincipalid)
WHERE  specifyuserid = 17
       AND usergroupscopeid NOT IN ( 65536, 32768, 4 )
       AND spprincipal.Name != 'Administrator'; -- added this condition
```

There still seems to be an issue of the 'User Group' picklist being set to the old value, rather than 'Manager'.  Looking the the `SpecifyUser` table, the value is set to 'Manger', but it shows up as `LimitedAccess` on the front-end.  Not sure why this is, the `user.userType` comes from the resource api.  Checking the `/context/user.json/` file served to the front-end, the `usertype: "Manager"` variable is correct, so somehow it is getting messed up before being displayed in the picklist.

emenslin (approved):
### Testing instructions

- go in security and accounts 
- click on make admin button under specify 6 permissions 
- open set collections under specify 6 permissions 
==> verify that all collections are checked
==> verify that you can unchecked some collections and save

Looks good, all collections are checked. 

https://github.com/specify/specify7/assets/150940840/d63e7021-2d84-4ac6-a2bb-60a2bad547f9

pashiav (approved):
### Testing instructions

- go in security and accounts 
- click on make admin button under specify 6 permissions 
- open set collections under specify 6 permissions 
==> verify that all collections are checked
==> verify that you can unchecked some collections and save
---
All collections are checked when made into an admin and you can edit the selected collections and save. Looks good! 
Screen Shot 2024-05-13 at 4 05 28 PM

PR #4659, opened 2024-03-15
Improve setting of default Tree Ranks

### LINKED ISSUE #4598: Incorrect tree rank is being set during editing and creation of tree records
UPDATE from @grantfitzsimmons:

I was able to recreate it thanks to more information provided by CSIRO. 

**This is something that must be resolved before 7.9.4 is released.**

Presumably introduced in https://github.com/specify/specify7/pull/4462?

image

Left is [v7.9.2](https://kufish212-v792.test.specifysystems.org/specify/tree/taxon/?conformation=%7E%7E1%7E15797%7E17025%7E5%7E21%7E135%7E685%7E10135---------), right is [edge](https://kufish212-edge.test.specifysystems.org/specify/tree/taxon/?conformation=~~1~15797~17025~5~21~135~685~10135---------)

Notice that the taxonomic rank is shifted one rank over (so even though `brevirostrum` is a node at the 'Species' rank, it shows 'Subspecies'.

Despite this, a query will return the correct rank as expected.

image

When you select a node in the tree, you will see immediately that it is "unsaved", indicating that there have been changes made. Unfortunately, this change is that Specify tries to change the rank by default.

https://github.com/specify/specify7/assets/37256050/b963d350-216f-4852-b97b-8a4fd952766f

Once saved, it updates the rank and now it is incorrect.

**Reported by:** Corinna at CSIRO ([Asana](https://app.asana.com/0/1201972579963460/1206699861047218/f))

---

Original description by @bronwyncombs:

> This might already be on your radar as we observed this in our Sandbox environment which is on the edge version. We are not seeing this in the stable 7.9.3.1 release.
> 
> We have noticed that the taxon rank displayed in the tree doesn’t align with the taxonomic rank displayed when you open the taxon node in Edit mode. We have seen this in fish, insect and wildlife – all collections have different numbers of taxon ranks. In the insect collection we have seen this go down all the way to Species being displayed as ‘subspecies’ in the Edit screen.
> 
> ETA: I just ran a quick query and data is fine there, so it seems to be an issue with the rank being pulled into the taxon node screen.
> image
> image
> image
> image

---

I haven't been able to recreate, but given the nature of this but I figured it should be reported for dev assessment.

melton-jason:
Fixes #4598 

For example the trees (specifically treedefitems) in the Taxon tree of Ichthyology discipline of `KUFish` are defined as the following:

Screenshot 2024-03-14 at 10 33 38 PM

A node at the `Family` level can have children of ranks of `Subfamily` or `Genus`. 
However, since `Genus` is enforced, this means that there _must_ be a rank at the `Genus` level in order to have taxon records at ranks lower than `Genus`.

In the example provided in https://github.com/specify/specify7/issues/4598#issuecomment-1998694168 (i.e., in KUFish), there are three possible ranks which children of `Genus` nodes can take: `Subgenus`, `Species`, `Subspecies`.

 The logic currently in `production` is to set the tree rank to be the lowest possible rank for a given parent. 
(In `KUFish`, the lowest possible rank for children of Genus nodes is Subspecies). 

This logic was not accounting for cases in which the record already had a rank. 

Now, the rank of a tree node is only set to the lowest possible rank given 
1. The value of the `definitionItem` field is undefined
OR
2. The value of the rank does not match that of the possible ranks for the tree record 

@specify/ux-testing 
There is a consideration in which the setting of the rank for a tree node may currently be an inconvenience. 
Is it more logical to consider populating the rank of a tree node to be the _highest_ possible rank rather than the lowest? 
For example, in `KUFish` currently when the `Add Child` button is pressed on a Genus Taxon record, the rank populated into the tree rank picklist will be Subspecies. Would it instead make more sense for this to be Subgenus? 

### Checklist

### Testing instructions

For any Tree (Taxon, Geography, Storage, Chronostrat, Lithostrat) 
- [ ] In the TreeViewer, ensure the correct rank populates into the rank picklist when editing/viewing a tree record
- [ ] When using the `Add Child` functionality, ensure the rank picklist remains blank but has the correct options available for ranks

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
+++ b/specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
@@ -99,10 +99,16 @@ export function TreeLevelComboBox(props: DefaultComboBoxProps): JSX.Element {
           )
           .then((items) => {
             if (destructorCalled) return undefined;
-            resource.set(
-              'definitionItem',
-              props.defaultValue ?? (items?.slice(-1)[0]?.value || '')
-            );
+            const definitionItem = resource.get('definitionItem');
+            if (
+              definitionItem === undefined ||
+              (items !== undefined &&
+                !items.map(({ value }) => value).includes(definitionItem))
+            )
+              resource.set(
+                'definitionItem',
+                props.defaultValue ?? (items?.slice(-1)[0]?.value || '')
+              );
             return void setItems(items);
           }),
       true
--- a/specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
+++ b/specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
@@ -99,16 +99,6 @@ export function TreeLevelComboBox(props: DefaultComboBoxProps): JSX.Element {
           )
           .then((items) => {
             if (destructorCalled) return undefined;
-            const definitionItem = resource.get('definitionItem');
-            if (
-              definitionItem === undefined ||
-              (items !== undefined &&
-                !items.map(({ value }) => value).includes(definitionItem))
-            )
-              resource.set(
-                'definitionItem',
-                props.defaultValue ?? (items?.slice(-1)[0]?.value || '')
-              );
             return void setItems(items);
           }),
       true


CarolineDenis:
@melton-jason 

> in KUFish currently when the Add Child button is pressed on a Genus Taxon record, the rank populated into the tree rank picklist will be Subspecies. Would it instead make more sense for this to be Subgenus

For me it make more sense to have Subgenus but let's have @grantfitzsimmons 's opinion

grantfitzsimmons:
> @melton-jason
> 
> > in KUFish currently when the Add Child button is pressed on a Genus Taxon record, the rank populated into the tree rank picklist will be Subspecies. Would it instead make more sense for this to be Subgenus
> 
> For me it makes more sense to have Subgenus but let's have @grantfitzsimmons 's opinion

Users would likely prefer the rank with the highest number of nodes present to appear by default (naturally whatever requires the least interaction would be preferred). In this case, the user would _almost_ always be entering a "Species" when creating a child under "Genus", but without that being specified in the database, either solution is not perfect.

> Is it more logical to consider populating the rank of a tree node to be the highest possible rank rather than the lowest?

It's more logical to populate the rank with the highest level rather than the lowest. Users can better understand why that specific rank is being populated because it is next in the tree.

melton-jason:
> Users would likely prefer the rank with the highest number of nodes present to appear by default (naturally whatever requires the least interaction would be preferred). In this case, the user would almost always be entering a "Species" when creating a child under "Genus", but without that being specified in the database, either solution is not perfect.

For now, I have removed the feature which sets the rank into the picklist when creating tree nodes.
The feature in theory is well-intentioned and helpful, but looks like it can lead to even bigger headaches when the trees themselves are misconfigured. 
In this case of adding a taxon record under `Genus` in KUFish for instance, it would make sense for `Species` to be populated. This means that `Species` should be enforced, but it is not. 

This would be less of an issue if the user could configure Trees in Specify (and change which ranks are enforced), but since there is no way to accomplish this within the application, it may be wise to wait on the feature to autopopulate the rank  depending on rank enforcement until the ability to configure ranks is implemented.

grantfitzsimmons (approved):
### Testing instructions

For any Tree (Taxon, Geography, Storage, Chronostrat, Lithostrat) 
- [X] In the TreeViewer, ensure the correct rank populates into the rank picklist when editing/viewing a tree record
- [X] When using the `Add Child` functionality, ensure the rank picklist remains blank but has the correct options available for ranks

maxpatiiuk:
Should we have changed the logic to consider a rang enforceable if `it is enforced, or rank id/name is present in a hardcoded list`?

PR #4660, opened 2024-03-15
Fix camel case issue for field in export feed

### LINKED ISSUE #4650: `ExportFeed` visual editor creating invalid definitions
**Describe the bug**
Creating or modifying an ﻿`ExportFeed` app resource using the visual editor results in export failures due to a KeyError in the code. The issue stems from attribute naming inconsistencies (i.e. `﻿collectionid` instead of `﻿collectionId`). To ensure functionality, the visual editor must use consistent camelcase for attribute names in line with existing resources.

## New `ExportFeed` app resource created with the visual editor

**This is not working.**

```xml
<?xml version="1.0" encoding="UTF-8"?>
<channel>
	<title>Test</title>
	<description>Test</description>
	<item collectionid="4" userid="1" notifyuserid="1" definition="DwCA_voucher" metadata="DwCA_voucher_metadata" days="7" filename="kudwca.zip" publish="true">
		<title>KU Fish Voucher</title>
		<id>ba7b743e-3263-4016-aebf-995f0d6a6e19</id>
	</item>
</channel>
```

After creating or adding to an existing `ExportFeed` app resource using the visual editor in `edge`, all exports will fail. The errors look like this on the several databases I tested this on:

```
"traceback":"Traceback (most recent call last):\n  File \"/opt/specify7/specifyweb/export/views.py\", line 169, in try_update_feed\n    update_feed(force=True, notify_user=user)\n  File \"/opt/specify7/specifyweb/export/feed.py\", line 53, in update_feed\n    collection = Collection.objects.get(id=item_node.attrib['collectionId'])\nKeyError: 'collectionId'\n"}
```

This error indicates a "KeyError" occurred in the code, specifically when trying to access a key ('collectionId') in a dictionary where this key is not present. That led me to look and see if this key is being created by the visual editor, and lo and behold, it is not.

The visual editor is creating a `collectionid` attribute in `<item>` rather than `collectionId` (notice the case difference). 

## Existing `ExportFeed` app resource created manually

```xml
<channel>
  <title>KUBI ichthyology RSS Feed</title>
  <description>RSS feed for KUBI Ichthyology Voucher and Tissue collections</description>
  <item collectionId="4" userId="2" notifyUserId="2" definition="DwCA_voucher" metadata="DwCA_voucher_metadata" days="7" filename="kui-dwca.zip" publish="true">
    <title>KU Fish</title>
    <id>8f79c802-a58c-447f-99aa-1d6a0790825a</id>
  </item>
</channel>
```

Notice `collectionid` is `collectionId`, `userid` is `userId`, `notifyuserid` is `notifyUserId`, and all the other names are consistent.

The visual editor needs to use the correct keys, so it should use camelcase properly to make it consistent with existing `ExportFeed` resources. If we shipped everything now, users would be unable to create working export feeds and be unable to use the "Update RSS Feed" function after doing so.

CarolineDenis:
Fixes #4650

### Checklist

### Testing instructions

- delete any existing export feed 
- save 
- create a new export feed using the visual editor 
- switch to xml editor 

- [ ] verify the collectionId and userId are camel case 
- [ ] verify that you can do an export

@grantfitzsimmons edit, also test:

- [ ] Navigate to a database with an existing export feed (one created before this PR)
- [ ] Verify that the export works in `edge`
- [ ] Update the RSS feed from the User Tools menu
- [ ] Verify that the export is completed successfully in `issue-4650`


specifyweb/frontend/js_src/lib/components/Syncer/fromSimpleXmlNode.ts
```diff
@@ -5,6 +5,8 @@ import { formatXmlNode } from './formatXmlNode';
 import type { SimpleChildren, SimpleXmlNode, XmlNode } from './xmlToJson';
 import { getOriginalSyncerInput } from './xmlUtils';
 
+// Needed for RSS export
+export const fieldNeedsCamelCase = ['collectionId', 'userId', 'notifyUserId'];
```
maxpatiiuk, line 9: this is not good as it will affect all attributes called `collectionId` or `userId` in any syncer spec, not just the export feed spec - it would be very unexpected when some attributes across totally unrelated syncer files don't get converted to lowercase while others do
if you do want to solve this on the front-end, the solution has to be specific to the export feed spec
but I think this should be solved on the back-end

maxpatiiuk (changes requested):
See https://github.com/specify/specify7/issues/4650#issuecomment-2002042520
Instead of adding an exeption on the front-end, we should fix this on the back-end because this is the only place across all XML files where we have camelCase XML attributes rather than lowercase or kebab-case

even worse, some of the attributes in the same XML file are lowercase and some camelCase, which is very inconsistent:
https://github.com/specify/specify7/blob/f216daef02219f9ba1210d3aac27b56ac9c85a02/specifyweb/export/feed.py#L39

to fix this inconsistency, without breaking existing XML resources, you can simply change this line:

https://github.com/specify/specify7/blob/f216daef02219f9ba1210d3aac27b56ac9c85a02/specifyweb/export/feed.py#L45

to

```py
collection = Collection.objects.get(id=item_node.get('collectionid', item_node.get('collectionId')))
```

(the above will try to get `collectionid` attribute if it exists, and if not, fallback to 

and same for the other 2

maxpatiiuk (approved):

emenslin (changes requested):
### Testing instructions

- delete any existing export feed 
- save 
- create a new export feed using the visual editor 
- switch to xml editor 

- [ ] verify the collectionId and userId are camel case 
- [x] verify that you can do an export

collectionId and userId are not camel case however I was able to export anyway. Is this the new expected behavior? And if so then the first check for testing instructions should be removed.

grantfitzsimmons:
Testers– please make sure to test this on DBs that already have working export feeds configured:

- [ ] Navigate to a database with an existing export feed (one created before this PR)
- [ ] Verify that the export works in `edge`
- [ ] Update the RSS feed from the User Tools menu
- [ ] Verify that the export is completed successfully

acwhite211:
I tested with and without camelCase for the RSS Export Feed on the KUFish database, and both seem to work fine.  Let me know if anyone finds a case or specific database where it is not working.

emenslin (approved):
### Testing instructions

- delete any existing export feed 
- save 
- create a new export feed using the visual editor 
- switch to xml editor 

- [x] verify the collectionId and userId are camel case 
- [x] verify that you can do an export

also test:

- [x] Navigate to a database with an existing export feed (one created before this PR)
- [x] Verify that the export works in `edge`
- [x] Update the RSS feed from the User Tools menu
- [x] Verify that the export is completed successfully in `issue-4650`

works both with and without camel case

combs-a (approved):
### Testing instructions

- delete any existing export feed 
- save 
- create a new export feed using the visual editor 
- switch to xml editor 

- [X] verify the collectionId and userId are camel case 
- [X] verify that you can do an export

Also worked with or without camel case for me, was all lowercase by default.

- [X] Navigate to a database with an existing export feed (one created before this PR)
- [X] Verify that the export works in `edge`
- [X] Update the RSS feed from the User Tools menu
- [X] Verify that the export is completed successfully in `issue-4650`

Worked perfectly, used sross on herb_rbge_2_5_2024. I had some issues with it on edge, where it'd take awhile to load, but it seems to work better on this PR? And same as above--it went to lowercase, but it seemed to work just fine regardless of camel case or not.

Also stumbled into some unrelated issues related to RSS feeds in general but overall it seems like this is working! Good job!

PR #4668, opened 2024-03-18
Display table format as select pick list

### LINKED ISSUE #4614: Make the formatter picklist UI more apparent as a picklist in the aggregator formatter menu
**Describe the bug**
I created a new DataObjFormatter, set is as "The Default Table Formatter", and save it.  I then went to the query builder and re-ran the query, but the same aggregator was being used.  It didn't switch over to using the new default aggregator until I deleted the old one.

I tried doing this again, and when I created the new formatter and checked it as default at the sametime, the change was reflected when rerunning the query.  But, then when I went to switch the default formatter back to the original (and pressed save), the change was not reflected when rerunning the query.

The DataObjFormatter UI seems to reflect the default formatter has been changed, but I suppose the default value might not be updated in the database.

**To Reproduce**

0. Tested on edge with the sp7demofish_2_9_24 database
1. Go to App Resources -> Record Formatters -> Collector -> Add
2. Press the "Save" button.
3. Create a "Collecting Event" query, and select "Collectors -> aggregated"
4. Try switching the default table formatter between the two (make sure to press Save), and see that when a new query runs, the field formatting in the query results remains unchanged.

**Expected behavior**
The fields in the query builder should be using the updated default formatting.

**Screenshots**
image
image
image
image
image

- Specify 7 Version: edge (7.9.4 pre-release, after xml-editor merge)
- Database Name: sp7demofish_2_9_24
- Collection name:  KU Fish Voucher

CarolineDenis:
Fixes #4614

### Checklist

### Testing instructions

- Go to App Resources -> Record Formatters -> Collector -> Add

- [ ] Verify that below "Table Format" the box holds 'Choose Table Format' holder and has a down arrow on the right to indicate a drop down

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Components.tsx
@@ -11,7 +11,7 @@ import type { GetSet, IR, RA } from '../../utils/types';
 import { filterArray, localized } from '../../utils/types';
 import { multiSortFunction, sortFunction } from '../../utils/utils';
 import { Ul } from '../Atoms';
-import { Input } from '../Atoms/Form';
+import { Input, Select } from '../Atoms/Form';
 import { ReadOnlyContext } from '../Core/Contexts';
 import type { LiteralField, Relationship } from '../DataModel/specifyField';
 import type { SpecifyTable } from '../DataModel/specifyTable';
@@ -40,6 +40,7 @@ import { getMappingLineData } from '../WbPlanView/navigator';
 import { navigatorSpecs } from '../WbPlanView/navigatorSpecs';
 import type { Aggregator, Formatter } from './spec';
 import type { FormatterTypesOutlet } from './Types';
+import { headerText } from '../../localization/header';
 
 export function FormattersPickList({
   table,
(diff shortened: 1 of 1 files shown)


maxpatiiuk (changes requested):
the reason datalist was used here instead of select is that the `formatters` list possibly doesn't contain all the formatters that could be picked in this list - sorry, I should have left a comment about that in the code

one possible case where this would happen:
> when editing a picklist resource, the "Formatter" field will call the FormattersPickList, but then the FormattersPickList would only have the formatters defined in the app resource that is currently loaded in the browser (the closest app resource to current scope) - which would not work if the opened a pick list that is defined for usage in a different collection as such pick list would be mean to be used with the formatters defined in that collection

options:
1. use datalist, but try to add the ▼ to make it look like select
2. add an "Other" option to the Select, which would then render an input below it - also have "Other" option be selected if current value is not present in the list
3. (the simplest) if current value is not among available pick list values, then display an `<option value={value}>` for it (as otherwise, the select won't show the current value correctly)
4. see if you can use PickListComboBox component in here as it would handle the case of current value not being in the list of options automatically

CarolineDenis:
There is an arrow already next to the pick list, just need to hover. This seems to be fine to me for now. Closing the issue. 
Screenshot 2024-03-19 at 6 35 53 AM

PR #4682, opened 2024-03-25
Add a tool to bulk move prepartions from one node to another

### LINKED ISSUE #4042: Bulk move preparation locations in the storage tree
After our discussion, we would need to create a new function for the Storage tree that allows you to change the preparation's storage locations from one node to another.

image

**Functionality:**
1. Select a storage node that contains 1 or more preparations
2. Click "Move Items" (I would love the image `Truck` heroicon)
3. Select the new location to move the items to
4. Click "Move items from {Current Location} here" in the toolbar
    image
5. Items are moved!

Both storage locations should still exist at the end of this interaction, just the preparations that were previously pointing to the first location should now point to the new one.

---

Add a feature where you can move all items in one storage location to another.

**Use Case:**
- Box has 81 positions.
- A sample can be put in one position.
- It would be nice to move that preparation to another position in another box easily.

**Requested By:** KU Mammals

CarolineDenis:
Fixes #4042

### Checklist

- [ ] Add automated tests

### Testing instructions

- Select a storage node that contains 1 or more preparations
- Click "Move Items" (Truck icon)
- Select the new location to move the items to
- Click "Move items from {Current Location} here" in the toolbar

- [ ] Verify items are moved


specifyweb/specify/tree_views.py
```diff
@@ -255,6 +255,20 @@ def move(request, tree, id):
                                    'old_value': old_fullname,
                                    'new_value': node.fullname}])
 
+def bulkMove(request, tree, id):
```
maxpatiiuk, line 258: add open api schema

maxpatiiuk (commented):
front-end changes thus far look good

maxpatiiuk (approved):
looks good, though someone should review the back-end code

specifyweb/frontend/js_src/lib/components/Permissions/definitions.ts
melton-jason, line 1: Should the bulkMove permission exist on every tree or only Storage? 
Currently the feature is only implemented for Storage nodes. Is it likely this feature will be used in other trees?

specifyweb/specify/tree_extras.py
```diff
@@ -10,7 +10,7 @@
 
 from specifyweb.businessrules.exceptions import TreeBusinessRuleException
 
-from  .auditcodes import TREE_MERGE, TREE_SYNONYMIZE, TREE_DESYNONYMIZE
+from  .auditcodes import TREE_BULK_MOVE, TREE_MERGE, TREE_SYNONYMIZE, TREE_DESYNONYMIZE
```
melton-jason, line 13: Currently, we are not using `TREE_BULK_MOVE`. 

Be sure to use the `mutation_log` function once the Preparations are successfully moved from one Storage node to another to insert the Action into the Audit Log (there are examples of this function for other Tree Actions in this file).
https://github.com/specify/specify7/blob/b331ce9c552e90d86fc4811a7089fab6ac29c07e/specifyweb/specify/tree_extras.py#L235-L237

https://github.com/specify/specify7/blob/b331ce9c552e90d86fc4811a7089fab6ac29c07e/specifyweb/specify/tree_extras.py#L286-L287

specifyweb/specify/tree_extras.py
```diff
@@ -299,6 +299,21 @@ def merge(node, into, agent):
...
+    logger.info('Bulk move preparations children %s into %s', node, into)
+    model = type(node)
+    if not type(into) is model: raise AssertionError(
+        f"Unexpected type of node '{into.__class__.__name__}', during bulk move. Expected '{model.__class__.__name__}'",
+        {"node" : into.__class__.__name__,
+        "nodeModel" : model.__class__.__name__,
+        "operation" : "bulkMove",
+        "localizationKey" : "invalidNodeType"})
+    target = model.objects.select_for_update().get(id=into.id)
+    if not (node.definition_id == target.definition_id): raise AssertionError("bulk move across trees", {"localizationKey" : "bulkMoveAcrossTrees"})
```
melton-jason, line 313: The `bulkMoveAcrossTrees` localization key is not resolved or recognized on the frontend. 

You will likely need to add the frontend implementation. 
See how similar tree actions are being resolved: 
 
https://github.com/specify/specify7/blob/b331ce9c552e90d86fc4811a7089fab6ac29c07e/specifyweb/frontend/js_src/lib/components/Errors/JsonError.tsx#L180-L183

Since these three error message are very similar, consider making a more generic `{operation} across trees` and remove the more specific error messages in favor of the generic one. 

Alongside the `{operation} across trees` message, you will need to pass in the operation name in the payload being sent with the error. 

For a similar example, see https://github.com/specify/specify7/blob/b331ce9c552e90d86fc4811a7089fab6ac29c07e/specifyweb/specify/tree_extras.py#L245-L248

specifyweb/specify/tree_extras.py
```diff
@@ -299,6 +299,21 @@ def merge(node, into, agent):
 
     assert False, "failed to move all referrences to merged tree node"
 
+def bulk_move(node, into, agent):
+    from . import models
+    logger.info('Bulk move preparations children %s into %s', node, into)
```
melton-jason, line 304: If this logger.info is to remain, I would recommend: 

```suggestion
    logger.info('Bulk move preparations from %s to %s', node, into)
```

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/Permissions/definitions.ts
CarolineDenis: no you're right we only use it for storage but didn't find yet a way to prevent this error: 
`Argument of type 'string' is not assignable to parameter of type 'keyof (Readonly<Record<"%", Readonly<Record<"%", boolean>>>> & { readonly '/preferences/user': Readonly<Record<"edit_protected", boolean>>; readonly '/preferences/statistics': Readonly<...>; } & { ...; })[`/tree/edit/${Lowercase<SCHEMA["tableName"]>}`]'.
  Type '"bulkMove"' is not assignable to type '"merge" | "move" | "synonymize" | "desynonymize" | "repair"'.`

specifyweb/specify/urls.py
```diff
@@ -30,6 +30,7 @@
         url(r'^(?P<id>\d+)/path/$', tree_views.path),
         url(r'^(?P<id>\d+)/merge/$', tree_views.merge),
         url(r'^(?P<id>\d+)/move/$', tree_views.move),
+        url(r'^(?P<id>\d+)/bulkMove/$', tree_views.bulk_move),
```
CarolineDenis, line 33: @melton-jason, 
Would it make sense here to have `r'^(?P<id>\d+)/bulk_move/$'` ? But then we would need to change the frontend bulkMove too...

melton-jason: Yes, most other endpoints on the backend use underscores as separators for the names in the url path rather than camelCase. 
`r'^(?P<id>\d+)/bulk_move/$'` matches the consistency used for other endpoints. 

You would need to change the frontend as well: 

https://github.com/specify/specify7/blob/74b5c9a5d9759ca590f3869368a21128ba7d2f99/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L27-L34

https://github.com/specify/specify7/blob/74b5c9a5d9759ca590f3869368a21128ba7d2f99/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L306-L309

https://github.com/specify/specify7/blob/74b5c9a5d9759ca590f3869368a21128ba7d2f99/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L338-L339

https://github.com/specify/specify7/blob/74b5c9a5d9759ca590f3869368a21128ba7d2f99/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L381-L382

https://github.com/specify/specify7/blob/74b5c9a5d9759ca590f3869368a21128ba7d2f99/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L421-L422

https://github.com/specify/specify7/blob/74b5c9a5d9759ca590f3869368a21128ba7d2f99/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L434-L435

https://github.com/specify/specify7/blob/74b5c9a5d9759ca590f3869368a21128ba7d2f99/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L450-L455

Alternatively, since the frontend prefers camelCase, you can create an Object/function which would map the frontend action name to the backend name and use the mapping before calling the endpoint:

https://github.com/specify/specify7/blob/74b5c9a5d9759ca590f3869368a21128ba7d2f99/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L319-L330

CarolineDenis: great thank you will do that

specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
```diff
@@ -286,6 +286,14 @@ function EditRecordDialog<SCHEMA extends AnyTree>({
...
 }
 
+const frontendToBackendMappingActions = {
+  bulkMove: 'bulk_move',
+  desynonymize: 'desynonymize',
+  merge: 'merge',
+  move: 'move',
+  synonymize: 'synonymize',
+};
+
```
melton-jason, line 296: Instead of redefining every `Action` in this Object (and losing type safety), you can likely leverage the spread operator and the action type to make this more accessibile for future developers. 

Consider the following modifications which can be made to the top of the file (and modifying the `Action` type): 

```ts
const treeActions = [
  'add',
  'bulkMove',
  'desynonymize',
  'merge',
  'move',
  'synonymize',
] as const;

type Action = typeof treeActions[number];

const frontendToBackendMappingActions: Readonly<Record<Action, string>> = {
  ...Object.fromEntries(treeActions.map((action) => [action, action])),
  bulkMove: 'bulk_move',
};
```

Now with the `...Object.fromEntries(treeActions.map((action) => [action, action]))`, we only need to specify the actions which need to be explicitly overwritten. Even if other Actions are added, modified, or removed, the mapping object gets updated automatically.

Additionally, this approach allows the following to be simplified to:

https://github.com/specify/specify7/blob/58ab2b2ae7b09ba13433bc9b8bd9d8de8c3a2080/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx#L289-L290

```ts
  if (!treeActions.includes(type)) throw new Error('Invalid action type');
```

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
```diff
@@ -157,6 +167,16 @@ export function TreeViewActions<SCHEMA extends AnyTree>({
           />
         </li>
       )}
+      {hasPermission(resourceName, 'bulkMove') && tableName === 'Storage' ? (
```
melton-jason, line 170: (optional) 

Can you swap the boolean expressions here? 
It is far more likely that the `tableName === 'Storage'` check is faster than checking permissions, so we can take advantage of [short-circuit evaluation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND#short-circuit_evaluation). 

```suggestion
      {tableName === 'Storage' && hasPermission(resourceName, 'bulkMove') ? (
```

melton-jason (changes requested):

CarolineDenis:
@specify/dev-testing @specify/ux-testing
Would it be beneficial to include the ids of the preparations which were bulk moved from the source node to the target node in the audit log entry for the Bulk Move operation?

specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
```diff
@@ -157,6 +167,17 @@ export function TreeViewActions<SCHEMA extends AnyTree>({
           />
         </li>
       )}
+      {tableName === 'Storage' &&
+      hasPermission(resourceName as '/tree/edit/storage', 'bulk_move') ? (
+        <li className="contents">
+          <Button.Icon
+            disabled={disableButtons || isRoot}
```
melton-jason, line 174: What is the reason to disable the bulk move action on the root node? 
At least from a functional standpoint, there is nothing wrong with allowing bulk moving preparations from the root node to other nodes. 

```suggestion
            disabled={disableButtons}
```

melton-jason (changes requested):
Looking good! Alongside awaiting feedback for https://github.com/specify/specify7/pull/4682#issuecomment-2066485610, just some more auditing code to consider. 

The Auditlog on the frontend has special code to make the format more readable for users. 

Can you add the bulk move audit log action to this array? 

https://github.com/specify/specify7/blob/340bb24aba8fcc5cde1a84d10019efa570149124/specifyweb/frontend/js_src/lib/components/PickLists/definitions.ts#L39-L47

The index of each element in this array corresponds to the auditcode on the backend (in this file: https://github.com/specify/specify7/blob/production/specifyweb/specify/auditcodes.py)

Thus, be careful not to change the ordering of any elements and insert the bulkMove action at index 7 (after tree desynonymize).

This will make it so that the bulk move action can be selected in the `Action` picklist of an Spauditlog query, and that in the query results, the frontend maps the audit code to a much more readable string instead of `7`.

emenslin (approved):
### Testing instructions

- Select a storage node that contains 1 or more preparations
- Click "Move Items" (Truck icon)
- Select the new location to move the items to
- Click "Move items from {Current Location} here" in the toolbar

- [x] Verify items are moved

Looks good, items moved to the new location (I love the truck icon as well)! I would like Grant to review this one though to make sure this is the expected behavior

melton-jason (approved):
🥳

PR #4688, opened 2024-03-26
Preserves aliased table

### LINKED ISSUE #4687: Circular table formats not detected/handled
**Describe the bug**
from issue #3266:

>  I.e, if accession includes accession agent formatter, and accession agent formatter includes accession formatter.
> 
> The fix:
> If the same resource appears the 2nd time in the formatting, exclude relationship fields from the formatter (thus, Accession will be formatted, Accession -> Accession Agents will be formatted, Accession -> Accession Agents -> Accession -> Accession Number will be formatted, but Accession -> Accession Agents -> Accession -> Accession Agents won't be
> 
> With the formatter/aggregator editor being available and so easy to use, and showing a live preview of formatter on any change, this issue is far more likely to happen. And once it does, the front-end will be rapidly sending requests to the back-end.

**To Reproduce**

Create an Accession format like this:

Screenshot 2024-03-25 at 3 01 32 PM

And an Accession Agent format like this: 

Screenshot 2024-03-25 at 3 12 01 PM

Query with these fields (formatted) & see the problem in results: 

Screenshot 2024-03-25 at 3 01 22 PM

**Expected behavior**

Results should display in query builder like they do on the "Preview" section of the table format dialog 

Not totally sure if it should prevent creating a "circular" table format like this, or just handle it differently/better ... 

- OS: macOS
- Browser: Chrome
- Specify 7 Version: `edge`
- Database Name: `herb_rbge_2_5_2024`
- Collection name: Herbarium

realVinayak:
Fixes #4687 

Backend does have logic for detecting and handling cyclical formatter and aggregator formatters. The actual problem was unrelated to that cycle.

The problem is that if you make a query like CollectionObject -> Accession -> (Accession formatted), and if Accession's formatter is aggregation of accession agents, then a subquery needs to be made. This subquery incorrectly didn't consider the outer accession (made from the join of CollectionObject -> Accession). So, the _entire_ accession table join accession agent was being used. 

Testing instructions: 
https://github.com/specify/specify7/issues/4687#issue-2206620927

### DIFF
--- a/specifyweb/stored_queries/format.py
+++ b/specifyweb/stored_queries/format.py
@@ -119,6 +119,10 @@ class ObjectFormatter(object):
         formatter = fieldNodeAttrib.get('formatter', None)
         aggregator = fieldNodeAttrib.get('aggregator', None)
         next_table_name = formatter_field_spec.table.name
+
+        # preserve sql table alias
+        formatter_field_spec = formatter_field_spec._replace(root_sql_table=orm_table)
+
         if formatter_field_spec.is_relationship():
             if previous_tables is not None and next_table_name in [table_name for table_name, _ in previous_tables]:
                 return (query, literal(
--- a/specifyweb/stored_queries/format.py
+++ b/specifyweb/stored_queries/format.py
@@ -125,9 +125,8 @@ class ObjectFormatter(object):
 
         if formatter_field_spec.is_relationship():
             if previous_tables is not None and next_table_name in [table_name for table_name, _ in previous_tables]:
-                return (query, literal(
-                    _text(f"<Cycle Detected.>: {'->'.join([*[str(_) for _ in previous_tables], next_table_name])}")),
-                        formatter_field_spec)
+                return query, literal(_text('')), formatter_field_spec
+
             new_query, new_expr, _, __ = formatter_field_spec.add_spec_to_query(
                 query,
                 formatter,
--- a/specifyweb/stored_queries/queryfieldspec.py
+++ b/specifyweb/stored_queries/queryfieldspec.py
@@ -58,7 +58,7 @@ def make_stringid(fs, table_list):
     return table_list, fs.table.name.lower(), field_name
 
 
-class QueryFieldSpec(namedtuple("QueryFieldSpec", "root_table join_path table date_part tree_rank tree_field")):
+class QueryFieldSpec(namedtuple("QueryFieldSpec", "root_table root_sql_table join_path table date_part tree_rank tree_field")):
     @classmethod
     def from_path(cls, path_in, add_id=False):
         path = deque(path_in)
@@ -80,6 +80,7 @@ class QueryFieldSpec(namedtuple("QueryFieldSpec", "root_table join_path table da
             join_path.append(node.idField)
 
         return cls(root_table=root_table,
+                   root_sql_table=getattr(models, root_table.name),
                    join_path=tuple(join_path),
                    table=node,
                    date_part='Full Date' if (join_path and join_path[-1].is_temporal()) else None,
@@ -130,6 +131,7 @@ class QueryFieldSpec(namedtuple("QueryFieldSpec", "root_table join_path table da
                 date_part = "Full Date"
 
         result = cls(root_table=root_table,
+                     root_sql_table=getattr(models, root_table.name),
                      join_path=tuple(join_path),
                      table=node,
                      date_part=date_part,
@@ -178,8 +180,7 @@ class QueryFieldSpec(namedtuple("QueryFieldSpec", "root_table join_path table da
         return field is not None and field.is_temporal()
 
     def build_join(self, query, join_path):
-        model = getattr(models, self.root_table.name)
-        return query.build_join(self.root_table, model, join_path)
+        return query.build_join(self.root_table, self.root_sql_table, join_path)
 
     def is_auditlog_obj_format_field(self, formatauditobjs):
         if not formatauditobjs or self.get_field() is None:
@@ -225,7 +226,7 @@ class QueryFieldSpec(namedtuple("QueryFieldSpec", "root_table join_path table da
 
         if self.tree_rank is None and self.get_field() is None:
             return (*query.objectformatter.objformat(
-                query, getattr(models, self.root_table.name), formatter), None, self.root_table)
+                query, self.root_sql_table, formatter), None, self.root_table)
 
         if self.is_relationship():
             # will be formatting or aggregating related objects
--- a/specifyweb/stored_queries/tests.py
+++ b/specifyweb/stored_queries/tests.py
@@ -88,15 +88,6 @@ class SQLAlchemySetup(ApiTests):
 
 
 
-    def setUp(self):
-        print("""
-            #BUG: If a test which compares the final sql query is added, then it could randomly fail
-            #in multithreaded tests because of usage of .label(None) in aggregate() 
-            #function in stored_queries/format.py 
-              """)
-        super().setUp()
-
-
 class SQLAlchemySetupTest(SQLAlchemySetup):
 
     def test_collection_object_count(self):
--- a/specifyweb/stored_queries/tests.py
+++ b/specifyweb/stored_queries/tests.py
@@ -358,7 +358,9 @@ class FormatterAggregatorTests(SQLAlchemySetup):
           >
             <switch single="true">
               <fields>
+                <field>role</field>
                 <field>accession</field>
+                <field>accession.accessionnumber</field>
               </fields>
             </switch>
           </format>
@@ -379,11 +381,11 @@ class FormatterAggregatorTests(SQLAlchemySetup):
         """
         object_formatter = self.get_formatter(formatter_def)
         accession_1 = spmodels.Accession.objects.create(
-            accessionnumber='a',
+            accessionnumber='Some_number',
             division=self.division)
         accession_agent_1 = spmodels.Accessionagent.objects.create(
             agent=self.agent,
-            role='role',
+            role='roleA',
             accession=accession_1,
         )
         with FormatterAggregatorTests.test_session_context() as session:
@@ -396,7 +398,7 @@ class FormatterAggregatorTests(SQLAlchemySetup):
             query = query.query.add_column(expr)
             self.assertCountEqual(list(query),
                                   [(
-                                   "<Cycle Detected.>: ('Accession', 'formatting')->('AccessionAgent', 'aggregating')->('AccessionAgent', 'formatting')->Accession",)]
+                                   "roleASome_number",)]
                                   )
 
     def test_relationships_in_switch_fields(self):


bronwyncombs (approved):
Definitely improved! This list of results is now at least a scrollable, and better than `v7.9`'s behavior of a recursion error crash 

Screenshot 2024-03-26 at 11 42 10 AM

@specify/ux-testing, thoughts? Plan to request a catch for this in the UI during format config later on?

CarolineDenis (approved):

specifyweb/stored_queries/format.py
```diff
@@ -119,11 +119,21 @@ def make_expr(self,
...
+        # preserve sql table alias
+        formatter_field_spec = formatter_field_spec._replace(root_sql_table=orm_table)
+
         if formatter_field_spec.is_relationship():
             if previous_tables is not None and next_table_name in [table_name for table_name, _ in previous_tables]:
-                return (query, literal(
-                    _text(f"<Cycle Detected.>: {'->'.join([*[str(_) for _ in previous_tables], next_table_name])}")),
-                        formatter_field_spec)
+                if query.detect_cycles:
+                    formatted_literal = literal(_text(f"<Cycle Detected.>: {'->'.join([*[str(_) for _ in previous_tables], next_table_name])}"))
```
maxpatiiuk, line 129: (optional)

```suggestion
                    formatted_literal = literal(_text(f"<Cycle Detected>: {'->'.join([*[str(_) for _ in previous_tables], next_table_name])}"))
```

specifyweb/stored_queries/execution.py
```diff
@@ -587,6 +587,7 @@ def build_query(session, collection, user, tableid, field_specs,
         collection=collection,
         objectformatter=ObjectFormatter(collection, user, replace_nulls),
         query=session.query(func.group_concat(id_field.distinct(), separator=',')) if distinct else session.query(id_field),
+        detect_cycles=settings.DEBUG
```
maxpatiiuk, line 590: `detect_cycles` is a bad variable name as it's supposed to always detect cycles, but then to only print debug information if this setting is true.

so really, this setting should be more like `debug=`, but probably shouldn't even exist because you can:
1. just reference settings.DEBUG directly in place where you were going to read this variable
2. OR remove this debug functionality entirely if you think it's not too likely that someone would benefit from this again in the future (and really, we should be using a python debugger or an SQL alchemy/MySQL debugger if there is such a thing)

specifyweb/stored_queries/format.py
```diff
@@ -119,11 +119,21 @@ def make_expr(self,
...
+                else:
+                    formatted_literal = literal(
+                        _text("<Invalid record formatter detected>") if formatter_field_spec.get_field().type == 'many-to-one' else
+                        _text("<Invalid record aggregator detected>")
+                    )
```
maxpatiiuk, line 134: I don't think we should be outputting `Invalid record formatter detected`.

reasons:
1. it's easy to accidentally create a cycle in the UI, and the UI doesn't provide an easy way to detect that it has happened
2. this string is very long

instead, I would suggest you match the front-end behavior and simply short-circuit the output once a cycle is detected (don't output anything/output empty string)

that is going to be the desirable behavior in most cases.
i.e a non-technical user wouldn't understand what a cycle in this context or why it's occurring. and resolving the cycle would force them to create almost identical duplicated formatters (one with a cyclical relationship, and another without)
where as simply silently short-circuiting would make this "just work"

maxpatiiuk (changes requested):

specifyweb/stored_queries/execution.py
realVinayak: > just reference settings.DEBUG directly in place where you were going to read this variable

Hmm nice idea, the only problem is that unit tests do not necessarily run in debug mode. They, for example, don't run right now in debug mode. I have a unit test that detects cycles. I want to make sure it detects the correct cycle (the full path)

specifyweb/stored_queries/format.py
realVinayak: After internal discussion, it was decided to keep the text.

specifyweb/stored_queries/execution.py
realVinayak: Although, you are right about variable naming.

> OR remove this debug functionality entirely if you think it's not too likely that someone would benefit from this again in the future (and really, we should be using a python debugger or an SQL alchemy/MySQL debugger if there is such a thing)

Then that would mean only devs will be able to figure it out what's happening. Right now, on the test panel, it would show the cycle detected, so anyone on our team can figure out the faulty formatter definition.

grantfitzsimmons (changes requested):
After feedback from the testing and user-facing members of the team, it was decided that showing `Invalid record formatter detected` is preferable over showing nothing or a full breakdown of detected cycles when not in debug mode.

maxpatiiuk (changes requested):
> After feedback from the testing and user-facing members of the team, it was decided that showing Invalid record formatter detected is preferable over showing nothing or a full breakdown of detected cycles when not in debug mode

It's not showing nothing. It's just not showing the same information twice.

Imagine an Accession formatter like this:
```js
`${accession.accessionNumber} ${accession.accessionAgents.each(format, ',')}`
```

and an Accession Agent formatter like this:
```js
`${accessionAgent.lastName} ${accessionAgent.accession}`
```

If we simply silently short-circuit the above when we detect a cycle, it would look like this:
formatting Accession Agent: `LastName AccessionNumber `
formatting Accession: `AccessionNumber LastName1  LastName2  LastName3`

The above looks quite usable and the user doesn't even need to know that their is a cycle - "it just works"™

---

now let's see how the above would work with the scary `<Invalid record formatter detected>`

formatting Accession Agent: `LastName AccessionNumber <Invalid record formatter detected>`
formatting Accession: `AccessionNumber LastName1 <Invalid record formatter detected> LastName2 <Invalid record formatter detected> LastName3 <Invalid record formatter detected>`
how is that helpful to the user???
it doesn't tell why the issue occurred, or how to resolve it. Worse, we didn't have to create an issue at all as we can just resolve it silently

---

Could you please reconsider the `Invalid record formatter detected` approach?
What are the reasons that it was preferred?

bronwyncombs:
@maxpatiiuk 

On behalf of UX, we thought that a `< invalid format detected>` message would signal a similar thought as `<format not defined>` aka changes in this app resource are necessary to return a more clear result from query builder. 

I think this is better than both the crash report you'd get in production, and a blank result as you suggest, because if the user really wants the result for the field they have a little something to go off of that isn't visually jarring like the crash or huge result lines that we're mitigating now. Further, a resolved blank field in a screenshot sent to support is much harder to debug than a field with this detection string.

maxpatiiuk (approved):
> I think this is better than both the crash report you'd get in production, and a blank result as you suggest, because if the user really wants the result for the field they have a little something to go off

this issue would only occur on cycles. if you have a cycle, then all the information the user wants is already included in the formatter at least once - short-circuiting the cycle just prevents the information from being displayed twice (or infinitely many times) (I don't think that's what users would ever want)

---

I still think just not displaying the data when we detect that we already printed the data for this table is a simple approach that creatures desirable results in many cases (rather than necessitating a support email)

Still, seems like you are already decided, so I won't block this further.
Though, consider changing the front-end logic too to reflect this behavior so as to be consistent

acwhite211:
Ok, let's start UX testing on this issue.  The functionality for returning the error message is looking good. @specify/ux-testing let me know if you find any problems.

I thinking about adding code that will just stop the formatting recursion and output the formatting up to that point without an error (like Max suggested), but have that functionality activated with a flag that we can enable in the future if we feel to do so.  Let me know if anyone thinks that is worth doing now?

acwhite211 (approved):

realVinayak:
Adding more unit tests to this PR would be a nice exercise for a new dev, since it would involve django and sql at the same time. 

> I thinking about adding code that will just stop the formatting recursion and output the formatting up to that point without an error (like Max suggested),

Implementation wise, you'd be just returning an empty string (instead of cycle detected / invalid formatted string).

pashiav (dismissed):
- [X] The edge will likely give a giant result string (it will take the entire accession table). The branch will not. 
---
The giant string is gone - looking good!!
Screen Shot 2024-04-04 at 9 15 44 AM

---
From https://github.com/specify/specify7/pull/4688#issuecomment-2030105887:
> On behalf of UX, we thought that a `< invalid format detected>` message would signal a similar thought as `<format not defined>` aka changes in this app resource are necessary to return a more clear result from query builder.

Instead of a blank result displayed to the user, there should be a `<invalid format detected>` message.

realVinayak:
@pashiav @grantfitzsimmons so you want the invalid formatted text at the top level? There is a difference in the screenshot attached [here](https://github.com/specify/specify7/pull/4688#issuecomment-2037505661), and what occurs on this branch. Just talking about backend here. Checking if that was intentional

pashiav:
Yes, to make sure the user is aware that the formatter they created is invalid.

realVinayak:
the behaviour now

image

realVinayak:
Not the biggest fan, since it makes it even more harder to figure out the issue

bronwyncombs (dismissed):
- [X] `<Incorrect Formatter Defined>` displays consistently for rows in results 
- [X] CSVs from these queries are exported and formatted appropriately
Screenshot 2024-04-05 at 9 21 27 AM

---

Changes: 
- [ ] Notification of invalid formatter configuration upon creation in Table Format editor preview is not yet handled 
Screenshot 2024-04-05 at 9 31 30 AM

- [x] Revert verbiage in results to "Invalid" as opposed to "Incorrect" 

Both of these should help to address concerns about the issue being unclear to the user.

acwhite211:
In order to allow the Table Format Preview to show the "\<Invalid Formatter Defined\>" error, we need to change it's functionality on the front-end to call the query api `/store_query/ephemeral`, instead of the specify table api `/api/specify/<table>/<id>/`

After talking with Jason, we decided that instead of using `fetchCollection`, we should instead use `createQuery` and `runQuery` when displaying the Table Format Preview.

https://github.com/specify/specify7/blob/ec31a51d1e6b28913dfd8a4b0222ad99055dfa94/specifyweb/frontend/js_src/lib/components/Formatters/Preview.tsx#L35-L49

realVinayak:
Should have made it clear before, but the text now appears on the frontend too.

emenslin (changes requested):
- [X] `<Invalid Formatter Defined>` displays consistently for query results 
Screenshot 2024-04-16 121513


- [X] CSVs from these queries are exported and formatted appropriately
- [x] Preview shows `<Invalid Formatter Detected>` 
Screenshot 2024-04-16 121537
---
Querycombobox behavior inconsistent. If a collection object already has an accession it shows up with the accession number and accession agent but if you remove that and try typing again it shows up as `<Invalid Formatter Defined>` (see video below) 

https://github.com/specify/specify7/assets/150940840/e62b32ec-9216-41d7-9a21-231ed6e59b6d

Also, shouldn't the accession name say `<Invalid Formatter Detected>` or is it supposed to show the format it currently does?
Screenshot 2024-04-16 123604

emenslin (approved):
- [x] Preview shows recursive format properly 
Screenshot 2024-04-16 151224

- [x] Queries show the correct format 
Screenshot 2024-04-16 144918

- [x] Exporting to CSV looks as expected 
- [x] querycomboboxes show up with the correct table format 
Screenshot 2024-04-16 144950

Probably the best solution we've come up with so far, looks good as far as I can tell, Id like one more testing review though before merging just to be sure that everything gets covered.

combs-a (approved):
- [X] Preview shows recursive format properly 

image

image

(I used *** to denote the AccessionAgent -> Accession, and --- for Accession -> AccessionAgent.)

- [X] Queries show the correct format 

In preview for Accession formatter:
image

In queries:
image

Compared the queries to edge queries, looks good!

- [X] Exporting to CSV looks as expected 
- [X] querycomboboxes show up with the correct table format 

image

Looking great! Good work! 👍

grantfitzsimmons (approved):

PR #4665, opened 2024-03-18
Restore past behavior for rank tree picklists

melton-jason:
Fixes #4664 
Fixes #4666 
Fixes #4676

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [x] Add relevant issue to release milestone

### Testing instructions

For any Tree (Taxon, Geography, Storage, Chronostrat, Lithostrat) 
- [ ] In the TreeViewer, ensure the correct rank populates into the rank picklist when editing/viewing a tree record
- [ ] When using the `Add Child` functionality, ensure the rank picklist is not blank and has the correct options available for ranks
- [ ] When setting the parent of a new Tree Record, the rank picklist should automatically populate with an acceptable rank
- [ ] The rank picklist for all trees should now be disbaled (readonly) when the tree record does not have a parent
- [ ] When viewing the root node for the tree, ensure the rank pisklist contains the correct rank

PR #4690, opened 2024-03-27
Set tree rank for new tree records after very short delay

### LINKED ISSUE #4666: Taxonomic Rank field displays API URL
**Describe the bug**
When viewing a Taxon record with no parent nodes (highest rank), the Taxonomic Rank field displays an API URL instead of the field being blank.

**To Reproduce**

1. Go to Taxon Tree
2. Click on the highest ranking node/node with no parents (https://fwri1924-edge.test.specifysystems.org/specify/view/taxon/12689/)
3. See the **Taxonomic Rank** field contains an API URL 

**Expected behavior**
Taxonomic Rank field should be blank if there are no parent nodes.
Behavior from **`v7.9.3`**:
expected

**Evidence**
issue

https://github.com/specify/specify7/assets/118556884/08e1c353-78f4-43c0-8ae7-fde99972f905

OS: macOS Monterey (12.6.5)
Browser: Chrome
Specify 7 Version: `edge`
Database Name: `fwri_1_9_24`

melton-jason:
Fixes #4666

> With regards to the other issue mentioned, here is what I have found in the code.
> 
> The issue is not with the field value on the resource object. Which is being set from
> 
> https://github.com/specify/specify7/blob/967d0d28c342e8d9d365f878053f0b82c90ed7b2/specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx#L123-L125
> 
> The underlying issue is with regards to parseResults in the useResourceValue hook. In this case `parseResults` returns `{isValid: false, value: <treeDefItemUrl>, reason: ''}`
> 
> https://github.com/specify/specify7/blob/967d0d28c342e8d9d365f878053f0b82c90ed7b2/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx#L116-L121
> 
> The problem is that the value is never being set on the HTML input element itself. Because of this, the [valueMissing](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState/valueMissing) attribute on the input's validity element remains `true` and thus `hasNativeErrors` returns true.
> 
> https://github.com/specify/specify7/blob/967d0d28c342e8d9d365f878053f0b82c90ed7b2/specifyweb/frontend/js_src/lib/utils/parser/parse.ts#L41-L45

https://github.com/specify/specify7/issues/4666#issuecomment-2021058501

### Checklist

### Testing instructions

- [ ] On a new Tree Data Entry form, ensure the rank populates correctly into the Tree Rank picklist (both the correct rank and no saveBlockers are set) when a parent is added 
- [ ] When using the `Add Child` button in the Tree Viewer, ensure the rank populates correctly into the Tree Rank picklist (both the correct rank and no saveBlockers are set)

Additionally, make sure the functionality from #4665 is kept: 

For any Tree (Taxon, Geography, Storage, Chronostrat, Lithostrat) 
- [ ] In the TreeViewer, ensure the correct rank populates into the rank picklist when editing/viewing a tree record
- [ ] When using the `Add Child` functionality, ensure the rank picklist is not blank and has the correct options available for ranks
- [ ] The rank picklist for all trees should now be disbaled (readonly) when the tree record does not have a parent
- [ ] When viewing the root node for the tree, ensure the rank pisklist contains the correct rank

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
+++ b/specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
@@ -120,9 +120,13 @@ export function TreeLevelComboBox(props: DefaultComboBoxProps): JSX.Element {
           .then((items) => {
             if (destructorCalled) return undefined;
             if (typeof resource.get('definitionItem') !== 'string')
-              resource.set(
-                'definitionItem',
-                props.defaultValue ?? items?.slice(-1)[0]?.value ?? ''
+              setTimeout(
+                () =>
+                  resource.set(
+                    'definitionItem',
+                    props.defaultValue ?? items?.slice(-1)[0]?.value ?? ''
+                  ),
+                1
               );
             return void setItems(items);
           }),


CarolineDenis (approved):

pashiav (approved):
### Testing instructions

- [X] On a new Tree Data Entry form, ensure the rank populates correctly into the Tree Rank picklist (both the correct rank and no saveBlockers are set) when a parent is added
- [X] When using the `Add Child` button in the Tree Viewer, ensure the rank populates correctly into the Tree Rank picklist (both the correct rank and no saveBlockers are set)
For any Tree (Taxon, Geography, Storage, Chronostrat, Lithostrat) 
- [X] In the TreeViewer, ensure the correct rank populates into the rank picklist when editing/viewing a tree record
- [X] When using the `Add Child` functionality, ensure the rank picklist is not blank and has the correct options available for ranks
- [X] The rank picklist for all trees should now be disbaled (readonly) when the tree record does not have a parent
- [X] When viewing the root node for the tree, ensure the rank pisklist contains the correct rank 

---

LookS great!!!! Good job, Jason!

maxpatiiuk (approved):
sad that these hacks are needed
we might want to do something about useResourceValue at some point
at least write good test coverage for that hook, as it's probably the most important hook we have

emenslin (approved):
### Testing instructions

- [x] On a new Tree Data Entry form, ensure the rank populates correctly into the Tree Rank picklist (both the correct rank and no saveBlockers are set) when a parent is added 
- [x] When using the `Add Child` button in the Tree Viewer, ensure the rank populates correctly into the Tree Rank picklist (both the correct rank and no saveBlockers are set)

Additionally, make sure the functionality from #4665 is kept: 

For any Tree (Taxon, Geography, Storage, Chronostrat, Lithostrat) 
- [x] In the TreeViewer, ensure the correct rank populates into the rank picklist when editing/viewing a tree record
- [x] When using the `Add Child` functionality, ensure the rank picklist is not blank and has the correct options available for ranks
- [x] The rank picklist for all trees should now be disbaled (readonly) when the tree record does not have a parent
- [x] When viewing the root node for the tree, ensure the rank pisklist contains the correct rank

PR #4694, opened 2024-03-27
Display header in grid table for first collapsed row

### LINKED ISSUE #4692: Grid view not showing headings by default
**Describe the bug**
In `v7`, grid view shows a heading above a list of records. In `edge`, this heading is missing, but can return if you expand and collapse a grid subview.

**To Reproduce**

1. Go to https://fitznhmaoct2023-edge.test.specifysystems.org/specify/view/collectionobject/608836/
2. See there is no heading above the record in **Preparations**
3. Expand "Preparations"
4. Collapse "Preparations"
5. See that the heading now appears

**Expected behavior**
It should be visible at all times as it was in `v7`.

In `v7`:

https://github.com/specify/specify7/assets/37256050/23a57ab9-e00c-476f-8426-9160416cfede

https://fitznhmaoct2023-v7.test.specifysystems.org/specify/view/collectionobject/608836/

**Screenshots**

In `edge`:

https://github.com/specify/specify7/assets/37256050/0e30a636-efbc-4086-8d48-a9f62d12a193

https://fitznhmaoct2023-edge.test.specifysystems.org/specify/view/collectionobject/608836/

CarolineDenis:
Fixes #4692

### Checklist

### Testing instructions

Go to https://fitznhmaoct2023-edge.test.specifysystems.org/specify/view/collectionobject/608836/

- [ ] See there is heading above the record in Preparations

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
@@ -190,8 +190,8 @@ export function FormTable<SCHEMA extends AnySchema>({
     mode !== 'view' && typeof handleDelete === 'function';
   const displayViewButton = !isDependent;
   const headerIsVisible =
-    (resources.length >= 1 && isExpanded[resources[0].cid] === undefined) ||
-    (resources.length >= 1 && isExpanded[resources[0].cid] === false);
+    (resources.length > 0 && isExpanded[resources[0].cid] === undefined) ||
+    (resources.length > 0 && isExpanded[resources[0].cid] === false);
 
   const headerWasVisibleRef = React.useRef(headerIsVisible);
   headerWasVisibleRef.current =
--- a/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
@@ -190,10 +190,8 @@ export function FormTable<SCHEMA extends AnySchema>({
     mode !== 'view' && typeof handleDelete === 'function';
   const displayViewButton = !isDependent;
   const headerIsVisible =
-    resources.length > 1 ||
-    (resources.length === 1 &&
-      (isExpanded[resources[0].cid] === undefined ||
-        isExpanded[resources[0].cid] === false));
+    (resources.length >= 1 && isExpanded[resources[0].cid] === undefined) ||
+    (resources.length >= 1 && isExpanded[resources[0].cid] === false);
 
   const headerWasVisibleRef = React.useRef(headerIsVisible);
   headerWasVisibleRef.current =
--- a/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
@@ -189,16 +189,16 @@ export function FormTable<SCHEMA extends AnySchema>({
   const displayDeleteButton =
     mode !== 'view' && typeof handleDelete === 'function';
   const displayViewButton = !isDependent;
-  const headerIsVisible =
-    (resources.length > 0 && isExpanded[resources[0].cid] === undefined) ||
-    (resources.length > 0 && isExpanded[resources[0].cid] === false);
+  // const headerIsVisible =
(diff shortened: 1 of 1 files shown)


pashiav (changes requested):
### Testing instructions

Go to https://fitznhmaoct2023-edge.test.specifysystems.org/specify/view/collectionobject/608836/

- [X] See there is heading above the record in Preparations
---

Headings appear, looking good!!
Screen Shot 2024-03-28 at 8 54 46 AM

---
When a Preparation is expanded, there is a large gap between the heading and the prep. The expected behavior is on the right on branch `edge`, which does not have as wide of a gap.
Screen Shot 2024-03-28 at 9 01 51 AM

CarolineDenis:
@maxpatiiuk , do we really want to keep the code to avoid the shift? It creates a lot of complications in the UI now

emenslin (approved):
### Testing instructions

Go to https://fitznhmaoct2023-edge.test.specifysystems.org/specify/view/collectionobject/608836/

- [x] See there is heading above the record in Preparations

Looks good!
Screenshot 2024-03-28 111413

grantfitzsimmons (approved):
> Go to [fitznhmaoct2023-edge.test.specifysystems.org/specify/view/collectionobject/608836](https://fitznhmaoct2023-edge.test.specifysystems.org/specify/view/collectionobject/608836/)
> 
>  - [x] See there is heading above the record in Preparations

Funny that the link is to a deployment on `edge` :joy:
We might want to have more robust testing instructions in other cases, but it works as expected!

maxpatiiuk:
> The gap with the headers invisible is normal though, it prevents the row to shift up, which prevent the user clicking somewhere else when rapidly deleting a row.

The gap should only be there if there was more than 1 row, and user removed all but the last one.
The gap should definitely not be there when there is more than 1 row - that would be a bug

maxpatiiuk (changes requested):
sorry if I am being slow, but I don't see what is the issue in https://github.com/specify/specify7/issues/4692#issuecomment-2026640173

And this PR seems to be introducing a bug. See https://github.com/specify/specify7/pull/4694#issuecomment-2026641720

grantfitzsimmons (changes requested):
> The gap with the headers invisible is normal though, it prevents the row to shift up, which prevent the user clicking somewhere else when rapidly deleting a row. (Fixed in https://github.com/specify/specify7/pull/4546)

As @maxpatiiuk said:
>The gap should only be there if there was more than 1 row, and user removed all but the last one.
>If user didn't remove the rows, the gap shouldn't be there, so there may be a bug with this PR (unless I am misunderstanding)

I believe he's correct

CarolineDenis:
NOTES: 
Decision: 
- Any time there is more than 1 row collapsed or not display header, if only one row and it's collapsed headers are displayed, if only one row but expanded than no headers

grantfitzsimmons:
After discussing this in-depth with @CarolineDenis– I felt that ultimately keeping the header at all times is preferable to having a gap _if_ our primary focus is reducing layout shifts.

Essentially, if you are visualizing a subview as a "table"/"grid", the expectation really should be that it is a table. If you want to simply show expanded record primarily, the form should be configured to display as a "form". 

This really begs the related (but too big for this issue) question– are we visualizing a table effectively in Specify 7? Maybe we need to add separators between the rows and columns.

specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx
```diff
@@ -271,7 +269,7 @@ export function FormTable<SCHEMA extends AnySchema>({
                 >
                   {isSortable && typeof fieldName === 'string' ? (
                     <Button.LikeLink
-                      tabIndex={headerIsVisible ? undefined : -1}
+                      tabIndex={undefined}
```
maxpatiiuk, line 272: remove this attribute as undefined is the default

maxpatiiuk (approved):
I like this behavior and it simplifies things
I think this is the behavior we had in the past in v7.7 until we changed it to remove redundant heading when you have only 1 row that is expanded.

Feel free to close https://github.com/specify/specify7/pull/4701 as unmerged

CarolineDenis:
let me know @grantfitzsimmons , @maxpatiiuk if we're good with this version

grantfitzsimmons (approved):

PR #4696, opened 2024-03-28
Display date picker for any value of accessible pref

### LINKED ISSUE #4691: Date picker hiding date precision on narrow forms
**To Reproduce**

1. Go to https://fitznhmaoct2023-edge.test.specifysystems.org/specify/view/collectionobject/new/ as `petrifiedroadkill` in the NHMA Malacology collection
2. Zoom in or make the window narrower
3. See that `edge` does not make enough room for "Full Date" or "Month / Year"

**Expected behavior**
The date picker should have "Full Date", "Month / Year", and "Year" fully visible.

**Screenshots**

https://github.com/specify/specify7/assets/37256050/1ff19c8d-606c-4514-bbdb-bcdf7ed3a2cf

[Specify 7 System Information - 2024-03-27T18_27_51.565Z.txt](https://github.com/specify/specify7/files/14779664/Specify.7.System.Information.-.2024-03-27T18_27_51.565Z.txt)

CarolineDenis:
Fixes #4691

### Checklist

### Testing instructions

- Go to preferences
- Check use accessible full date picker and  use accessible month picker 
- save 
- go to data entry form 

- [ ] Verify that the 3 date option are visible 

- Go to preferences
- Uncheck use accessible full date picker and  use accessible month picker 
- save 
- go to data entry form 

- [ ] Verify that the 3 date option are visible

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/PartialDateUi.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/PartialDateUi.tsx
@@ -10,6 +10,7 @@ import { getDateParser } from './dateUtils';
 import type { PartialDatePrecision } from './useDatePrecision';
 import { useDatePrecision } from './useDatePrecision';
 import { useMoment } from './useMoment';
+import { userPreferences } from '../Preferences/userPreferences';
 
 /** A date picker with precision selection (full, month-year, year) */
 export function PartialDateUi({
@@ -47,9 +48,16 @@ export function PartialDateUi({
 
   const isReadOnly =
     React.useContext(ReadOnlyContext) || resource === undefined;
+  const [useAccessiblePicker] =
+    userPreferences.use('form', 'ui', 'useAccessibleFullDatePicker') ||
+    userPreferences.use('form', 'ui', 'useAccessibleMonthPicker');
 
   return (
-    <div className="grid w-full grid-cols-[fit-content(40%)_1fr] gap-1">
+    <div
+      className={`${
+        useAccessiblePicker ? 'grid-cols-2' : 'grid-cols-[fit-content(40%)_1fr]'
+      } grid w-full gap-1`}
+    >
       {!isReadOnly && canChangePrecision ? (
         <DatePrecisionPicker moment={moment} {...precisionProps} />
       ) : undefined}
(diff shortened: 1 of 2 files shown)


pashiav (approved):
### Testing instructions
Go to preferences
Check use accessible full date picker and  use accessible month picker 
save 
go to data entry form 
- [X] Verify that the 3 date option are visible 

Go to preferences
Uncheck use accessible full date picker and  use accessible month picker 
Save 
go to data entry form 
- [X] Verify that the 3 date option are visible 
---

Looks good, all date precision selections are visible!
Screen Shot 2024-03-28 at 10 16 22 AM

emenslin (approved):
### Testing instructions

- Go to preferences
- Check use accessible full date picker and  use accessible month picker 
- save 
- go to data entry form 

- [x] Verify that the 3 date option are visible 

- Go to preferences
- Uncheck use accessible full date picker and  use accessible month picker 
- save 
- go to data entry form 

- [x] Verify that the 3 date option are visible 

Looks good!

maxpatiiuk (changes requested):
This was the intended behavior, not a bug.
See https://github.com/specify/specify7/issues/4691#issuecomment-2027403082

grantfitzsimmons (changes requested):
In @pashiav's review, the behavior looks undesirable:

> Looks good, all date precision selections are visible!
> Screen Shot 2024-03-28 at 10 16 22 AM
> 

Here's a head-to-head comparison between `edge`, `v7`, and this PR (`issue-4691`):

image

Losing more of the date in favor of seeing the precision isn't worth the trade off after further evaluation.

After reviewing this on a couple of databases (and with narrow windows as I always do), I am beginning to agree with Max's fix as an improvement over the current behavior and as a better fix than this PR. See his comment here: https://github.com/specify/specify7/issues/4691#issuecomment-2027403082

In @maxpatiiuk 's [comment here](https://github.com/specify/specify7/issues/4691#issuecomment-2027870148):
> no they didn't have enough room - they just forcefully pushed other columns out of the way.
> Notice how on the left screenshot, the Cataloger query combo box has fewer letters visible making it less usable. The fields in the last columns are also a bit narrower. At the same time, the fields in the middle column were way too wide.
> 
> if there is enough room, then everything would be displayed in full, without being cut off.

I think he makes a good point. It's clear that it was previously forcing the forms to give it more room and we've fixed that in `edge`. 

I think we should close this PR in favor of the current "fix" and consider @maxpatiiuk's suggestion here:

> If we need a good compromise that is both space-efficient, and looks great, we could replace the Select in the date picker with just a button-icon, that when clicked would either toggle the date format or open a pick list to select a date format.

PR #4702, opened 2024-03-29
Disable check boxes when read only

### LINKED ISSUE #4700: Read-only User can interact with Table Formatter
**Describe the bug**
Logged in as a user with read-only permissions, the user can interact with the Table Formatter dialog in App Resources. Users with read-only permissions should not have this ability.

Note: found while testing https://github.com/specify/specify7/pull/4612

**To Reproduce**
Branch: `edge`
1. Log in as user: **`grant`** on `umherb_1_10_24`
2. Go to App Resources => Record Formatters => Determination (https://umherb11024-edge.test.specifysystems.org/specify/resources/app-resource/92/formatter/Determination/)
3. Click **Determination (Default)**
4. Uncheck **Default** => See a red bar with the text "Please select a default record formatter for this table" appears behind the dialog
5. Uncheck **Conditional Format** then check it again => See that the original format disappears

**Expected behavior**
The dialog should not be interactable to read-only users. All check boxes should not be clickable.

**Evidence**

https://github.com/specify/specify7/assets/118556884/0ee9e24b-2bed-48f7-85c9-20b2db79fa0e

OS: macOS Monterey (12.6.5)
Browser: Chrome
Specify 7 Version: `edge`
Database Name: `umherb_1_10_24`

CarolineDenis:
Fixes #4700

### Checklist

### Testing instructions
- Connect as a read only user
- Go to App Resources => Record Formatters => Choose Table that has conditional formatter
- Click on a formatter

- [ ] Verify the 2 check boxes (Default and Conditional Format) are disabled 

- proceed with the same steps with another table

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
@@ -155,7 +155,6 @@ export const Input = {
         props.onChange?.(event);
       },
       readOnly: isReadOnly,
-      disabled: isReadOnly,
     })
   ),
   Text: wrap<
--- a/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
@@ -155,6 +155,7 @@ export const Input = {
         props.onChange?.(event);
       },
       readOnly: isReadOnly,
+      disabled: isReadOnly,
     })
   ),
   Text: wrap<
--- a/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
@@ -136,6 +136,7 @@ export const Input = {
     'input',
     {
       readonly onValueChange?: (isChecked: boolean) => void;
+      readonly onClick?: 'Use onValueChange instead';
       readonly readOnly?: 'Use isReadOnly instead';
       readonly isReadOnly?: boolean;
       readonly type?: 'If you need to specify type, use Input.Generic';
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Element.tsx
@@ -142,7 +142,7 @@ export function FormatterWrapper(): JSX.Element {
             <Input.Checkbox
               checked={getSet[0].isDefault}
               isReadOnly={isReadOnly}
-              onClick={(): void =>
+              onChange={(): void =>
                 setItems(
                   // Ensure there is only one default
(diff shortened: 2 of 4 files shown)


emenslin (approved):
### Testing instructions
- Connect as a read only user
- Go to App Resources => Record Formatters => Choose Table that has conditional formatter
- Click on a formatter

- [x] Verify the 2 check boxes (Default and Conditional Format) are disabled 

- proceed with the same steps with another table 

Cannot check either of the check boxes and cannot edit any other fields.

bronwyncombs (approved):
- Connect as a read only user **(guest, on `kufishtissue`, tissue collection)**
- Go to App Resources => Record Formatters => Choose Table that has conditional formatter
- Click on a formatter

- [X] Verify the 2 check boxes (Default and Conditional Format) are disabled 

- [X] proceed with the same steps with another table 

Save and delete buttons gone, and pick list selections for fields can't be changed, but they can be clicked on and opened to view other options. 

Restrictions correctly apply for read-only in XML editor too.

specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
```diff
@@ -155,6 +155,7 @@ export const Input = {
         props.onChange?.(event);
       },
       readOnly: isReadOnly,
+      disabled: isReadOnly,
```
maxpatiiuk, line 158: note, readOnly and disabled are not equivalent
this change affects all checkboxes across specify, and that is likely not desirable in all cases
read more about the difference: https://ux.stackexchange.com/a/118041/131886
for example, readonly checkboxes are tab-focusable, but disabled are not.

Instead, notice that on line 153 in this file we are ignoring the change event for any checkbox that is disabled or read-only. Could you check why does that not prevent the change event?

<details><summary>Soluton</summary>
<p>

The checkbox in the but report has `onClick` event listener rather than `onChange` or `onValueChange`.
Could you modify the typings for `Input.Checkbox` to disallow using the onClick event, OR, add similar logic we have for onChange (disabled and readonly check) to the onClcik event

</p>
</details>

maxpatiiuk (changes requested):

maxpatiiuk (changes requested):
fixes the bug in this case, but doesn't prevent it from happening again in the future.

could you please address this part:

> Could you modify the typings for Input.Checkbox to disallow using the onClick event, OR, add similar logic we have for onChange (disabled and readonly check) to the onClcik event


<p>

Add this line:

> readonly onClick?: 'Use onValueChange instead';

in here:

https://github.com/specify/specify7/blob/issue-4611/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx#L138

</p>
</details>

maxpatiiuk (approved):
🌈

PR #4704, opened 2024-03-29
Display synonyms node when hovering

### LINKED ISSUE #407: Synonyms should be shown when hovering over a node in a tree
Mouse overs for synonyms are not working in both directions – mouse overs showing preferred names for a synonym are working but not mouse overs for the synonym names of preferred names.

CarolineDenis:
Fixes #407

### Checklist

- [ ] Add automated tests

### Testing instructions

- go to any tree 
- hover over a node that has synonyms
- [ ] verify that the synonyms are being displayed

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
@@ -284,9 +284,9 @@ export function TreeRow<SCHEMA extends AnyTree>({
                         })
                       : fetchedChildrenName === undefined
                       ? undefined
-                      : `${treeText.synonyms()} ${fetchedChildrenName.join(
-                          ','
-                        )}`
+                      : treeText.synonyms({
+                          names: fetchedChildrenName.join(', '),
+                        })
                   }
                 >
                   {doIncludeAuthorPref &&
--- a/specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
@@ -11,8 +11,17 @@ import { getPref } from '../InitialContext/remotePrefs';
 import { userPreferences } from '../Preferences/userPreferences';
 import type { Conformations, KeyAction, Row, Stats } from './helpers';
 import { formatTreeStats, mapKey, scrollIntoView } from './helpers';
+import { genericTables } from '../DataModel/tables';
+import { AnySchema, AnyTree } from '../DataModel/helperTypes';
+import { SpecifyTable } from '../DataModel/specifyTable';
+import { useAsyncState } from '../../hooks/useAsyncState';
+import { hijackBackboneAjax } from '../../utils/ajax/backboneAjax';
+import { Http } from '../../utils/ajax/definitions';
+import { unsafeTriggerNotFound } from '../Router/Router';
+import { useLiveState } from '../../hooks/useLiveState';
+import { SpecifyResource } from '../DataModel/legacyTypes';
 
-export function TreeRow({
+export function TreeRow<SCHEMA extends AnyTree>({
   row,
   getRows,
   getStats,
@@ -53,7 +62,7 @@ export function TreeRow({
   readonly onAction: (action: Exclude<KeyAction, 'child' | 'toggle'>) => void;
   readonly setFocusedRow?: (row: Row) => void;
   readonly synonymColor: string;
-  readonly treeName: string;
+  readonly treeName: SCHEMA['tableName'];
   readonly hideEmptyNodes: boolean;
 }): JSX.Element | null {
   const [rows, setRows] = React.useState<RA<Row> | undefined>(undefined);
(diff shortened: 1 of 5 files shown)


specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -154,6 +163,33 @@ export function TreeRow({
           const stats =
             typeof nodeStats === 'object' &&
             formatTreeStats(nodeStats, row.children === 0);
+          const table = genericTables[treeName] as SpecifyTable<AnyTree>;
+          const [resource] = useLiveState<
+            SpecifyResource<AnySchema> | undefined
+          >(
+            React.useCallback(() => {
```
maxpatiiuk, line 170: sorry, but you keep making the same mistake ---- you can't call react hooks anywhere but on the top level of a component
you can't call it inside of callbacks, inside of loops, inside of conditions!

please read this article carefully - https://legacy.reactjs.org/docs/hooks-rules.html#:~:text=Only%20Call%20Hooks%20at%20the%20Top%20Level
feel free to also research online how react hooks work under the hood so that you can understand why such restriction exists

maxpatiiuk: instead, if you need some hook result to be used inside your loop, then either move the hook out of the loop, OR, if your hook relies on some data specific to that loop iteration, then move the loop body into it's own component and then inside of that component you can call your hooks at the top level

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -154,6 +163,33 @@ export function TreeRow({
...
+              const table = genericTables[treeName] as SpecifyTable<AnyTree>;
+              const parentNode = new table.Resource({ id: row.nodeId });
+              let node = parentNode;
+              return node;
+            }, [row.nodeId, treeName])
+          );
+          const [loadedResource] = useAsyncState(
```
maxpatiiuk, line 177: performance of calling this for each row is going to be bad.
tree performance is bad as is.
instead, this information might need to be provided by the back-end, or on-demand, or enabled only for certain ranks (like we do with tree stats right now)
otherwise, this would make a lot of network requests...

cc @realVinayak

maxpatiiuk (changes requested):
🔻

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -147,6 +157,49 @@ export function TreeRow({
...
+
+  const [fetchedChildrenName, setFetchedChildrenName] =
+    React.useState<RA<string>>();
+
+  React.useEffect(() => {
+    fetchedChildren().then(({ data: objects }) => {
+      const synonymsNames = objects.objects.map((node: Row) => node.name);
+      setFetchedChildrenName(synonymsNames);
+    });
+  }, [acceptedChildren]);
```
maxpatiiuk, line 201: refactor to use useAsyncState?

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -147,6 +157,49 @@ export function TreeRow({
...
+        (status) =>
+          status === Http.NOT_FOUND ? unsafeTriggerNotFound() : undefined
+      );
+    }, [resource]),
+    false
+  );
+
+  const acceptedChildren = loadedResource?.get('acceptedChildren') || '';
+
+  const fetchedChildren = async (): Promise<any> => {
```
maxpatiiuk, line 182: @realVinayak could you please advise if it would be better to do this on the back-end (when the rows themself are fetched) or on the front-end here (which would create a lot of extra network requests)?

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
```diff
@@ -147,6 +157,49 @@ export function TreeRow({
...
+        [Http.NOT_FOUND],
+        async () => resource.fetch(),
+        (status) =>
+          status === Http.NOT_FOUND ? unsafeTriggerNotFound() : undefined
+      );
+    }, [resource]),
+    false
+  );
+
+  const acceptedChildren = loadedResource?.get('acceptedChildren') || '';
```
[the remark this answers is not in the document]
maxpatiiuk: even better, since you only need the name field, we have a `fetchRows` function that would do that even more efficiently as it would let you return just the single field you want rather than the entire resource

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/TreeView/Row.tsx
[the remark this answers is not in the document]
CarolineDenis: after some tryouts, went for fetchCollection

CarolineDenis:
NOTES: 
Will try to do something on the back too

maxpatiiuk (changes requested):

maxpatiiuk (approved):
Pushed code that resolves these issues
Also, refactored fetchRows definition a bit to reduce duplication
And, added tests while I was at it

emenslin (changes requested):
### Testing instructions

- go to any tree 
- hover over a node that has synonyms
- [x] verify that the synonyms are being displayed 

Looks good, when hovering it shows both synonyms and preferred taxon 
Screenshot 2024-05-08 090131
Screenshot 2024-05-08 090145

The only issue is that even when there are no synonyms for a taxon the tooltip still shows, I think if there is no synonym/preferred taxon then there shouldn't be a tooltip. 
Screenshot 2024-05-08 090427

emenslin (approved):
### Testing instructions

- go to any tree 
- hover over a node that has synonyms
- [x] verify that the synonyms are being displayed 

Looks good, synonym only displayed if there are actually synonyms. 
image

combs-a (approved):
### Testing instructions
- go to any tree 
- hover over a node that has synonyms
- [X] verify that the synonyms are being displayed 

image
image

Looking good, both hovers work.

PR #4724, opened 2024-04-02
Add group number support in Taxon queries

### LINKED ISSUE #4723: Add support for 'Group Number' in Taxon queries
First component of https://github.com/specify/specify7/issues/4697

The same issue reported by Taylor is causing an issue with this query defined in Specify 6:
```json
{
  "tableList": "1,9-determinations,4",
  "stringId": "1,9-determinations,4.taxon.Family groupNumber",
  "fieldName": "Family groupNumber",
  "isRelFld": false,
  "sortType": 0,
  "position": 2,
  "isDisplay": true,
  "operStart": 11,
  "startValue": "",
  "isNot": false
}
```

As described in https://github.com/specify/specify7/issues/4697#issuecomment-2025741314, you can only search on "Author" and "Full Name" at a given rank, therefore Specify 7 is unable to successfully execute the query. If we extended support to searching on `groupNumber`, that would solve this case, but we should do it more generalized if possible.

[Specify 7 Crash Report - 2024-03-27T14_37_45.508Z.txt](https://github.com/specify/specify7/files/14824446/Specify.7.Crash.Report.-.2024-03-27T14_37_45.508Z.txt)

grantfitzsimmons:
Fixes #4723

This has been tested between both Specify 6 and 7. Existing queries that previously were non-functional created in 6 work now as well!

image

The previous configuration expected `<RankName> GroupNumber`, but Specify 6 constructs the string ID for these query items as `<RankName> groupNumber`.

### Checklist

### Testing instructions
_Note: To properly test some of these, you will need to use Specify 6 locally to configure such a query or you can import this query that was created in Specify 6:_
[GroupNumber 6 Query.json](https://github.com/specify/specify7/files/14838445/GroupNumber.6.Query.json)

- [ ] Run a query that had the field "Group Number" added at the Family level in  Specify 6
    - [ ] Verify the output is consistent with Specify 6 and that all query functions work as expected
- [ ] Add "Group Number" to various ranks in the Taxon tree and ensure it can be searched on and results are displayed properly
- [ ] Verify that hiding and unhiding "Group Number" in the schema configuration make it appear as expected


specifyweb/stored_queries/queryfieldspec.py
```diff
@@ -22,7 +22,7 @@
 DATE_PART_RE = re.compile(r'(.*)((NumericDay)|(NumericMonth)|(NumericYear))$')
 
 # Pull out author or groupnumber field from taxon query fields.
-TAXON_FIELD_RE = re.compile(r'(.*) ((Author)|(GroupNumber))$')
+TAXON_FIELD_RE = re.compile(r'(.*) ((Author)|(groupNumber))$')
```
maxpatiiuk, line 25: FYI: from what Ben told me, the only reason we limit what fields are accessible in tree queries is because that's what sp6 did
that reason might be less good today compared to how it was in 2013.
@realVinayak interested in tackling this at some point?

maxpatiiuk (approved):
🎉

CarolineDenis (approved):

emenslin (approved):
### Testing instructions
_Note: To properly test some of these, you will need to use Specify 6 locally to configure such a query or you can import this query that was created in Specify 6:_
[GroupNumber 6 Query.json](https://github.com/specify/specify7/files/14838445/GroupNumber.6.Query.json)

- [x] Run a query that had the field "Group Number" added at the Family level in  Specify 6
    - [x] Verify the output is consistent with Specify 6 and that all query functions work as expected
- [x] Add "Group Number" to various ranks in the Taxon tree and ensure it can be searched on and results are displayed properly
- [x] Verify that hiding and unhiding "Group Number" in the schema configuration make it appear as expected

Looks good, group number can now be queried in 7 
Screenshot 2024-05-08 094447

combs-a (approved):
### Testing instructions

- [X] Run a query that had the field "Group Number" added at the Family level in Specify 6
    - [X] Verify the output is consistent with Specify 6 and that all query functions work as expected
- [X] Add "Group Number" to various ranks in the Taxon tree and ensure it can be searched on and results are displayed properly
- [X] Verify that hiding and unhiding "Group Number" in the schema configuration make it appear as expected

Everything looks to be working correctly, used it between a local specify 6 DB and used the query Grant posted for instructions.

Results in 6 and 7:
image

image

When hidden:
image

And the query I used:
[GroupNumberTest.json](https://github.com/specify/specify7/files/15310212/GroupNumberTest.json)

Good work!

alesan99 (changes requested):
- [ ] Run a query that had the field "Group Number" added at the Family level in  Specify 6
    - [ ] Verify the output is consistent with Specify 6 and that all query functions work as expected
- [ ] Add "Group Number" to various ranks in the Taxon tree and ensure it can be searched on and results are displayed properly
- [ ] Verify that hiding and unhiding "Group Number" in the schema configuration make it appear as expected

Query results are different between SP 6 and 7.
chrome_NEDW2TPmcw
With one of the group number lines removed SP 7 matches.
chrome_fC4fTIQBdF
And this is how it looks on `v7` (without the GroupNumber lines)
chrome_9KyKNglaPq

specifyweb/stored_queries/queryfieldspec.py
realVinayak: See https://github.com/specify/specify7/pull/4929. Begun some work there. now need to add tests to ensure compatibility with 6.

alesan99 (approved):
After further testing it looks like we were only able to reproduce the inconsistency between 6 and 7 if that specific query was made in 7 first, which wouldn't really affect users I don't think.
I don't have any notes on how I originally encountered the issue but it seems like that must've been how. So I think this is good.

PR #4750, opened 2024-04-05
Enable select all/deselect all when typing in bulk, handle NaN

### LINKED ISSUE #4745: Loan 'Bulk Return' disables select all and deselect all buttons
**Describe the bug**
When returning a loan typing in the 'bulk return' field both the select all and deselect all buttons are disable until you type in the 'bulk resolve field' 

**To Reproduce**

1. Go to a loan with preparations 
2. Click on return loan 
3. Type a number into the 'bulk return'
4. See select all and deselect all disabled 

**Expected behavior**
The buttons should never be disabled unless all preparations are selected or deselected 

**Screenshots**

https://github.com/specify/specify7/assets/150940840/995a4b4c-4d8e-485b-8396-73bf42af4d8b

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge
- Collection name: mammals
- User Name: SpAdmin
[Specify 7 System Information - 2024-04-05T15_51_52.663Z.txt](https://github.com/specify/specify7/files/14887292/Specify.7.System.Information.-.2024-04-05T15_51_52.663Z.txt)

CarolineDenis:
Fixes #4745

### Checklist

### Testing instructions

- Go to a loan with preparations
- Click on return loan
- Type a number into the 'bulk return'

- [ ]  Verify select all and deselect all are enabled

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
@@ -103,9 +103,10 @@ function PreparationReturn({
       remarks: '',
     }))
   );
+
   const canDeselect = state.some(({ resolve }) => resolve > 0);
   const canSelectAll = state.some(
-    ({ resolve, unresolved }) => resolve < unresolved || Number.isNaN(resolve)
+    ({ resolve, unresolved }) => resolve < unresolved
   );
 
   const [bulkReturn, setBulkReturn] = React.useState(0);
@@ -133,12 +134,12 @@ function PreparationReturn({
           type === 'resolve'
             ? updateResolvedChanged({
                 returns,
-                resolve: resolvedCount,
+                resolve: Number.isNaN(resolvedCount) ? 0 : resolvedCount,
                 unresolved,
                 remarks: returnItem.remarks,
               })
             : updateReturnChanged({
-                returns: resolvedCount,
+                returns: Number.isNaN(resolvedCount) ? 0 : resolvedCount,
                 resolve,
                 unresolved,
                 remarks: returnItem.remarks,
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
```diff
@@ -105,7 +105,7 @@ function PreparationReturn({
   );
   const canDeselect = state.some(({ resolve }) => resolve > 0);
   const canSelectAll = state.some(
-    ({ resolve, unresolved }) => resolve < unresolved
+    ({ resolve, unresolved }) => resolve < unresolved || Number.isNaN(resolve)
```
maxpatiiuk, line 108: where is the NaN coming from?
it shouldn't be here - if it is, there may be a bug somewhere
and plus, even if you handle nan in this case, it won't be handled in all cases (i.e line 106 and others)

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
CarolineDenis: when the user select the bulk value and delete it to enter manually a new number, for a fraction of time the value for resolve or return is NaN

pashiav (approved):
- [X]  Verify select all and deselect all are enabled
---
Both Select All and Deselect All are enabled. Looks good!!

https://github.com/specify/specify7/assets/118556884/34884ede-157e-498e-97e8-2c9ecdaf0b68

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/Interactions/LoanReturn.tsx
maxpatiiuk: ah, yes, we have this issue in some places
ideally, the onValueChanged in Input.Number should be `number | undefined` rather than `number` (unless the input also has `isRequired={true}`, in which case `onValueChanged` should not be triggered while the input is empty

PR #4758, opened 2024-04-08
Allow decimals in Input field

### LINKED ISSUE #4754: Decimal fields prevent saving and display a misleading error until the fields are focused on
**Describe the bug**
Save blockers are triggered because a decimal field is not a whole number until it is focused, then the record can be saved once focus is lost.

**To Reproduce**
1. Go to https://herb32124-v793.test.specifysystems.org/specify/view/collectionobject/1253799
2. See that you cannot save
3. Click "Save" and notice your focus is now on "Lat Seconds" warning that you need to enter a valid value
4. Click "Save" again and notice your focus is now on "Long Seconds" giving the same warning
5. Click "Save" again and see that you can save

**Expected behavior**

**Screenshots**

`edge` on left, `v7.9.3.1` on right

https://github.com/specify/specify7/assets/37256050/3cca6781-e566-40e6-aabb-886d2873082d

https://herb32124-v793.test.specifysystems.org/specify/view/collectionobject/1253799

**Crash Report**

[Specify 7 System Information - 2024-04-05T21_00_25.376Z.txt](https://github.com/specify/specify7/files/14890734/Specify.7.System.Information.-.2024-04-05T21_00_25.376Z.txt)

- OS: macOS
- Browser: Arc
- Specify 7 Version: `edge`
- Database Name: `herb_3_21_24`
- Collection name: Herbarium
- User Name: `spadmin`

CarolineDenis:
Fixes #4754

### Checklist

### Testing instructions

- use herb32124
- open the record set named small set
- [ ] verify that no errors are thrown for the lat second field 
- [ ] verify you can switch from one record to another

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
@@ -162,11 +162,11 @@ function Field({
          */
         updateValue(input.value, event.type === 'paste');
       }}
-      // step={
-      //   validationAttributes.step === undefined
-      //     ? 'any'
-      //     : validationAttributes.step
-      // }
+      step={
+        validationAttributes.step === undefined
+          ? 'any'
+          : validationAttributes.step
+      }
     />
   );
 }
--- a/specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
@@ -123,44 +123,44 @@ function Field({
     'ui',
     'rightAlignNumberFields'
   );
-  return (
-    <Input.Generic
-      forwardRef={validationRef}
-      name={name}
-      {...validationAttributes}
-      className={
-        /*
-         * Disable "text-align: right" in non webkit browsers
-         * as they don't support spinner's arrow customization
-         */
-        parser.type === 'number' &&
-        rightAlignNumberFields &&
-        globalThis.navigator.userAgent.toLowerCase().includes('webkit')
-          ? `text-right ${isReadOnly ? '' : 'pr-6'}`
-          : ''
-      }
-      id={id}
-      isReadOnly={isReadOnly}
-      required={'required' in validationAttributes && !isInSearchDialog}
-      tabIndex={isReadOnly ? -1 : undefined}
-      value={value?.toString() ?? ''}
-      onBlur={
-        isReadOnly ? undefined : ({ target }): void => updateValue(target.value)
-      }
-      onValueChange={(value): void => updateValue(value, false)}
+
+  const commonProps = {
+    forwardRef: validationRef,
+    name: name,
+    className:
       /*
-       * Update data model value before onBlur, as onBlur fires after onSubmit
-       * if form is submitted using the ENTER key
+       * Disable "text-align: right" in non webkit browsers
+       * as they don't support spinner's arrow customization
        */
-      onChange={(event): void => {
-        const input = event.target as HTMLInputElement;
-        /*
-         * Don't show validation errors on value change for input fields until
-         * field is blurred, unless user tried to paste a date (see definition
-         * of Input.Generic)
-         */
-        updateValue(input.value, event.type === 'paste');
-      }}
-    />
+      rightAlignNumberFields &&
+      globalThis.navigator.userAgent.toLowerCase().includes('webkit')
+        ? `text-right ${isReadOnly ? '' : 'pr-6'}`
+        : '',
+    id: id,
+    isReadOnly: isReadOnly,
+    required: 'required' in validationAttributes && !isInSearchDialog,
+    tabIndex: isReadOnly ? -1 : undefined,
+    value: value?.toString() ?? '',
+    onBlur: isReadOnly ? undefined : ({ target }) => updateValue(target.value),
+    onValueChange: (value) => updateValue(value, false),
+    /*
+     * Update data model value before onBlur, as onBlur fires after onSubmit
+     * if form is submitted using the ENTER key
+     */
+    onChange: (event) => {
+      const input = event.target as HTMLInputElement;
+      /*
+       * Don't show validation errors on value change for input fields until
+       * field is blurred, unless user tried to paste a date (see definition
+       * of Input.Generic)
+       */
+      updateValue(input.value, event.type === 'paste');
+    },
+  };
+
+  return validationAttributes.type === 'number' ? (
+    <Input.Float {...commonProps} {...validationAttributes} />
+  ) : (
+    <Input.Generic {...commonProps} {...validationAttributes} />
   );
 }
--- a/specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
@@ -124,43 +124,44 @@ function Field({
     'rightAlignNumberFields'
   );
 
-  const commonProps = {
-    forwardRef: validationRef,
-    name: name,
-    className:
+  return (
+    <Input.Generic
+      forwardRef={validationRef}
+      name={name}
+      {...validationAttributes}
+      className={
+        /*
+         * Disable "text-align: right" in non webkit browsers
+         * as they don't support spinner's arrow customization
+         */
+        parser.type === 'number' &&
+        rightAlignNumberFields &&
+        globalThis.navigator.userAgent.toLowerCase().includes('webkit')
+          ? `text-right ${isReadOnly ? '' : 'pr-6'}`
+          : ''
+      }
+      id={id}
+      isReadOnly={isReadOnly}
+      required={'required' in validationAttributes && !isInSearchDialog}
+      tabIndex={isReadOnly ? -1 : undefined}
+      value={value?.toString() ?? ''}
+      onBlur={
+        isReadOnly ? undefined : ({ target }): void => updateValue(target.value)
+      }
+      onValueChange={(value): void => updateValue(value, false)}
       /*
-       * Disable "text-align: right" in non webkit browsers
-       * as they don't support spinner's arrow customization
+       * Update data model value before onBlur, as onBlur fires after onSubmit
+       * if form is submitted using the ENTER key
        */
-      rightAlignNumberFields &&
-      globalThis.navigator.userAgent.toLowerCase().includes('webkit')
-        ? `text-right ${isReadOnly ? '' : 'pr-6'}`
-        : '',
-    id: id,
-    isReadOnly: isReadOnly,
-    required: 'required' in validationAttributes && !isInSearchDialog,
-    tabIndex: isReadOnly ? -1 : undefined,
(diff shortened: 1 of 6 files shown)


maxpatiiuk (approved):

CarolineDenis:
@maxpatiiuk: 
Notes: 
Does not seem to fix the bug in #4754

maxpatiiuk:
Based on the video recording, the issue is with the `step` attribute being set to `1` when it shouldn't be

acwhite211 (approved):

specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
```diff
@@ -162,6 +162,11 @@ function Field({
          */
         updateValue(input.value, event.type === 'paste');
       }}
+      step={
+        validationAttributes.step === undefined
+          ? 'any'
+          : validationAttributes.step
+      }
```
maxpatiiuk: 1. I'm not sure (or at least didn't know) that `any` is a valid value for the step attribute - I though if you want to allow any step, you just pass undefined to this attribute
2. Sorry to say, but this is not a fix, just a bad workaround. This change would ignore the `step` attribute completely, even if it was set by the user in the form definition. Instead, you have to investigate where the value for `validationAttributes.step` is coming from in the case when it's not desirable.

<p>

Search for references to `step` in this file - https://github.com/specify/specify7/blob/issue-4789/specifyweb/frontend/js_src/lib/utils/parser/definitions.ts

`step` should have a default value of `1` for integer fields. `step` must not have a default value for fields that accept floating point.

Then, update the tests to catch this error for the future

</p>
</details>

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
CarolineDenis: this fix allows the user to enter any number in the number input if he didn't set a step in the xml. In the case he did set a step in the xml it will be respected

CarolineDenis: I pushed a code with changes in definitions.ts too. The behavior stays the same

CarolineDenis: but what I don't like with this version is that if you have no step defined and you change your input to be by example 9.3 you will have an alert Please choose a valid number, The nearest valid number are 9 and 10. Error we do not have with my first push. I'd be in favor of keeping the first code

bronwyncombs (approved):
- [X] verify that no errors are thrown for the lat second field 
- [X] verify you can switch from one record to another 


https://github.com/specify/specify7/assets/135047322/f207e477-26bf-4290-8df2-620206353939

emenslin (approved):
### Testing instructions

- use herb32124
- open the record set named small set
- [x] verify that no errors are thrown for the lat second field 
- [x] verify you can switch from one record to another 

Looks good, there is no longer an error and can switch between reconds. Also works on other databases. 
Screenshot 2024-04-16 111217

melton-jason (dismissed):
While this does correctly resolve this issue for float-related fields, it does not address the underlying problem and may lead to the same bug popping up again in the future if the root cause is not addressed. 

(Although there likely _is_ an issue currently with fields which are strictly integers. If the step attribute is not defined, this change sets the step attribute to `any`, which allows decimals in the integer fields; in this case the step attribute should remain the default `1`)

The problem, as @maxpatiiuk  mentioned in https://github.com/specify/specify7/pull/4758#issuecomment-2043936332 is that the `step` attribute is being set to `1` when it shouldn't be. 

From the Mozilla docs: 

> If not explicitly included, step defaults to 1 for number and range [input types]

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step

For modularity, the HTML validation-related attributes should be handled by the `parser` related code. 
Consider the case where another `number` type input is to be rendered by the frontend. Currently, the changes made in this PR would need to be applied to ensure #4754 does not occur again and the `step` attribute is correctly set for Float/Decimal fields. 
Ideally, this other place would just need to use the `parser`-related code to handle all of the parsing/attributes and not have to worry about applying this fix. 

The preferred solution would then be to modify the parser/parsing code to explicitly specify the step as `any` (as far as the HTML is concerned) when the field can accept floats and doesn't already specify a step in the `initialize` attributes for the cell on the form definition.  
(Importantly as noted before, integer-type fields should still default to have a step of 1)

The `validationAttributes` is coming from a function which maps Specify Parser attributes to HTML validation attributes: 

https://github.com/specify/specify7/blob/e85e6f474253cac6ca04e45649a2f59313fe855d/specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx#L119

The `parser` itself comes from `useResourceValue`: 
https://github.com/specify/specify7/blob/e85e6f474253cac6ca04e45649a2f59313fe855d/specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx#L104-L108

And the `defaultParser` is a simple parser constructed solely from the cell on the form definition:

https://github.com/specify/specify7/blob/e85e6f474253cac6ca04e45649a2f59313fe855d/specifyweb/frontend/js_src/lib/components/FormFields/index.tsx#L182-L203

`useResourceValue` is using the following `useParser` hook to construct the 'final' parser for the field:

https://github.com/specify/specify7/blob/e85e6f474253cac6ca04e45649a2f59313fe855d/specifyweb/frontend/js_src/lib/hooks/resource.tsx#L122-L143

Importantly, take note of the `resolveParser` function:

https://github.com/specify/specify7/blob/e85e6f474253cac6ca04e45649a2f59313fe855d/specifyweb/frontend/js_src/lib/utils/parser/definitions.ts#L263-L290

`resolveParser` constructs a Parser based on the `field` attributes as defined in the DataModel, which is exactly the function we are looking for: we want to explicitly define an `any` step for fields which are BigDecimal/float type fields.
Thus, the Parser returned from this function in these cases should include the any step attribute. 
 
With the change, any place on the frontend which wants to construct a Parser based on a field will benefit from the fix for #4754
(`mergeParsers` will also likely need modification to handle the case where the field parser and defaultParser defined on the form conflict)

specifyweb/frontend/js_src/lib/components/FormFields/Field.tsx
maxpatiiuk: > This change would ignore the step attribute completely, even if it was set by the user in the form definition

Ah you are right - sorry, I was too quick to jump on this

> I'm not sure (or at least didn't know) that any is a valid value for the step attribute

you are correct - `any` is a valid option for this attribute. And `1` is the implicit default value. I didn't know that

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step#:~:text=If%20not%20explicitly%20included%2C%20step%20defaults%20to%201%20for%20number%20and%20range%2C%20and%201%20unit%20type%20(minute%2C%20week%2C%20month%2C%20day)%20for%20the%20date/time%20input%20types.%20The%20value%20must%20be%20a%20positive%20number%20%2D%20integer%20or%20float%20%E2%80%94%20or%20the%20special%20value%20any

maxpatiiuk:
@CarolineDenis sorry about my comments - I spoke to quick.

Still, the suggestions that @melton-jason are valid and should be addressed

In addition, for Input.Float, we need to add a `step={props.step ?? 'any'}` so that it doesn't force integers when step is not provided:

https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx#L262

Also, this line contains a bug:

https://github.com/specify/specify7/blob/e85e6f474253cac6ca04e45649a2f59313fe855d/specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx#L247C18-L247C21

could you please change it to `props.step !== Math.floor(props.step)` (disallow any non-inter step values in Input.Integer - if needed, Input.Float should be used instead)

specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
```diff
@@ -70,7 +70,13 @@ export type FieldTypes = {
...
       readonly max: number | string | undefined;
       // These are used by numeric field only
-      readonly step: number | undefined;
+
+      /*
+       * 'any' is not a valid step attribute on the form, and not validated
+       * by the Syncer, but is required for the default step for floating
+       * point fields
+       */
+      readonly step: number | 'any' | undefined;
```
melton-jason, line 79: The following line was angry about the adjustments to the `Parser` type: 
https://github.com/specify/specify7/blob/e0f1234513aac3e379083524f84da47e539d1f66/specifyweb/frontend/js_src/lib/components/Forms/generateFormDefinition.ts#L272

I have resolved the error by adding the additional `'any'` in the FieldTypes type (which is used for the FormFieldDefinition type). 
Although, I would assume the `FieldTypes` type is supposed to represent the valid values of the attribute for the Cell Definition and `any` is not supported on the form for  the call attribute`initialize="step=value"`

So, if needed, instead of adding the `'any'` step type here, we can alternatively only modify the type definition in the code to autogenerate the form to keep the type definition for the 'Form' elements free of the `Parser` required modifications.

melton-jason:
@specify/ux-testing 
Can this be tested again? 

If you want to test this outside of `herb32124` and/or the database/`Small set` record set is no longer available or modified, here are general testing instructions. 

- Identify one or more fields on a Table which is/are of type `BigDecimal` or `Double`
- Add or ensure that the field(s) are present on the table's View Definition
- Set the `step` attribute on (one of) the field(s) to be any decimal number
- [ ] On a New Data Entry form for the table, ensure that only values matching the step attribute are valid for the 
- Remove the `step` attribute on (another of) the field(s)
- [ ] On a New Data Entry form for the table, ensure that any numeric value (both including/excluding decimal) is allowed in the modified field
- Find (or create) an existing record which contains values in fields which were modified on the form definition 
- [ ] Ensure the value in the field can only match what is supported by its `step` attribute
  - If the current value in the field is not valid according to it's step, it changes to accommodate the nearest valid step value

---

For example:
On the following section of an `Accession` form definition, 
`number1` should only allow increasing/decreasing values by 0.5 and allow a max precision of 0.5 (valid: 0, 0.5, 1, 1.5, ...)
`number2` should allow any numeric value. 
```xml 
		<row>
			<cell type="label" labelfor="num1"/>
			<cell type="field" id="num1" name="number1" uitype="spinner" initialize="step=0.5"/>
			<cell type="label" labelfor="num2"/>
			<cell type="field" id="num2" name="number2" uitype="spinner"/>
		</row>
```

pashiav (approved):
### Testing instructions

- use herb32124
- open the record set named small set
- [X] verify that no errors are thrown for the lat second field 
- [X] verify you can switch from one record to another 

---
Looks good!

lexiclevenger (approved):
use herb32124
open the record set named small set
- [x] verify that no errors are thrown for the lat second field
- [x] verify you can switch from one record to another

Looks good!

maxpatiiuk (commented):
👍

PR #4763, opened 2024-04-09
refactor(FormEditor): use .getRelationship() in viewSpec

maxpatiiuk:
Refactor code in response to code changes in PR #4748 (it was merged before I had a chance to fix it)

### Checklist

### Testing instructions

- Test that the thing #4748 was fixing is still fixed
- Test that an error is shown when you enter a name of the field that does not exist in the related table (see list of fields per table: https://sp7demofish.specifycloud.org/specify/data-model/)
- open forms with data entry


specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
```diff
@@ -406,11 +406,10 @@ const subViewSpec = (
...
           (raw: string) => {
             const cellName = cell.rest.node.attributes.name;
-            const cellRelationship = table?.fields
-              .filter((field) => field.isRelationship)
-              .find((table) => table.name === cellName) as
-              | Relationship
-              | undefined;
+            const cellRelationship =
+              typeof cellName === 'string'
+                ? table?.getRelationship(cellName)
```
CarolineDenis, line 411: this results in the same error in Sp, table?.getRelationship is not a function

maxpatiiuk: Actually, this was caused by a bug in the definition.

syncers.switch is called by viewSpec:

https://github.com/specify/specify7/blob/3ffb5e7fa9e14b7421ff946bb9bac7cfd75ee274/specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts#L61-L89

switch in turn uses spread syntax to augment the passed in arguments:
https://github.com/specify/specify7/blob/3ffb5e7fa9e14b7421ff946bb9bac7cfd75ee274/specifyweb/frontend/js_src/lib/components/Syncer/syncers.ts#L661

in the process, that breaks the `table` object - it's properties are preserved in a new object, but all of the methods are lost.

specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
```diff
@@ -406,12 +406,14 @@ const subViewSpec = (
...
             const cellName = cell.rest.node.attributes.name;
-            const cellRelationship = table?.fields
-              .filter((field) => field.isRelationship)
-              .find((table) => table.name === cellName) as
-              | Relationship
-              | undefined;
-            const cellRelatedTableName = cellRelationship?.relatedTable.name;
+            const cellRelationship =
+              typeof cellName === 'string'
+                ? syncers.field(table?.name).serializer(cellName)?.at(-1)
```
maxpatiiuk, line 411: by using syncers.field, an error will be thrown on invalid field name

specifyweb/frontend/js_src/lib/components/Syncer/syncers.ts
```diff
@@ -658,7 +659,7 @@ export const syncers = {
             ]) as TYPE_MAPPER[keyof TYPE_MAPPER]) ?? ('Unknown' as const);
         const spec = mapper[type] ?? mapper.Unknown;
         const { serializer } = syncers.object(
-          spec(cell, { ...extraPayload, rawType })
+          spec(cell, extraPayload, rawType)
```
maxpatiiuk, line 662: this was causing `extraPayload` object to loose it's prototype chain, thus table was loosing the `getRelationship` and all other methods

CarolineDenis (approved):

lexiclevenger (approved):
### Testing instructions

- [x] Test that the thing #4748 was fixing is still fixed
- [x] Test that an error is shown when you enter a name of the field that does not exist in the related table (see list of fields per table: https://sp7demofish.specifycloud.org/specify/data-model/)
- [x] open forms with data entry 

---

Looks good! The fix from #4748 still works, and an error is shown if a field does not exist in the related table.

PR #4767, opened 2024-04-09
Only use query builder in the search formatter

### LINKED ISSUE #4756: Preview Search tool in Record Formatters does not return correct results
**Describe the bug**
Record Formatters' Preview Search tool does not always return the correct results.

**To Reproduce**

1. Go to AccessionAgent table formatter: https://umherb11024-edge.test.specifysystems.org/specify/resources/app-resource/92/formatter/AccessionAgent/1
2. Copy the name of an existing Agent in the Preview 
3. Click **Search** => paste the existing agent's name into the Agent field (e.g. Gerofue)
4. Click the green **Search** button 
5. See that it returns no results.

**Expected behavior**
The Search tool should always return the correct results for any input.

**Evidence**
1. Video shows that _Gerofue_ exists but does not appear in results => then selecting Collector returns all the collector records => shows that _Andrea Matthies_ exists => shows that no results are returned when only searching for _Andrea Matthies_

https://github.com/specify/specify7/assets/118556884/dd599d8f-73dd-49f4-93db-3449c8d99f83

2. Example of a record existing in Loan but returning no results: https://umherb11024-edge.test.specifysystems.org/specify/resources/app-resource/92/formatter/Loan/33

https://github.com/specify/specify7/assets/118556884/130221c3-6e65-4eaa-bf72-5e186f9f2b84

Note: I checked CO table format and it returned the correct results

[Specify 7 System Information - 2024-04-08T15_01_35.011Z.txt](https://github.com/specify/specify7/files/14907284/Specify.7.System.Information.-.2024-04-08T15_01_35.011Z.txt)
OS: macOS Monterey (12.6.5)
Browser: Chrome
Specify 7 Version: `edge`
Database Name: `umherb_1_10_24`

CarolineDenis:
Fixes #4756

### Checklist

### Testing instructions

- Go to a table formatter
- Click Search
- [ ] Verify that the search dialog opens only the query builder tool 
- Proceed with the selection 
- [ ] Verify that the search is still accessible from other places in sp 7. i.e a combobox in data entry. Test also changing the pref to always use query builder in the search dialog

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/SearchDialog/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/SearchDialog/index.tsx
@@ -60,6 +60,7 @@ export function SearchDialog<SCHEMA extends AnySchema>(props: {
   readonly onClose: () => void;
   readonly searchView?: string;
   readonly onSelected: (resources: RA<SpecifyResource<SCHEMA>>) => void;
+  readonly onlyUseQueryBuilder?: boolean;
 }): JSX.Element | null {
   const [alwaysUseQueryBuilder] = userPreferences.use(
     'form',
@@ -67,7 +68,7 @@ export function SearchDialog<SCHEMA extends AnySchema>(props: {
     'alwaysUseQueryBuilder'
   );
   const [useQueryBuilder, handleUseQueryBuilder] = useBooleanState(
-    alwaysUseQueryBuilder
+    props.onlyUseQueryBuilder ? true : alwaysUseQueryBuilder
   );
   return useQueryBuilder ? (
     <QueryBuilderSearch
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Preview.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Preview.tsx
@@ -106,6 +106,7 @@ export function useResourcePreview(table: SpecifyTable): {
             table={table}
             onClose={handleClose}
             onSelected={setResources}
+            onlyUseQueryBuilder={true}
           />
         )}
       </div>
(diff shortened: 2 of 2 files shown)


combs-a (approved):
### Testing instructions

- Go to a table formatter
- Click Search
- [X] Verify that the search dialog opens only the query builder tool 
- Proceed with the selection 
- [X] Verify that the search is still accessible from other places in sp 7. i.e a combobox in data entry.
- [X] Test also changing the pref to always use query builder in the search dialog 

Everything looks good and the 'always use query builder' preference works great. Good work!

melton-jason (approved):
This solution is fine if we are settling on specifically only using the QueryBuilder search interface for selecting records in which to preview a formatter/aggregator. 

Although #4756 can still happen if the user can access the Search Dialog in other places in Specify. 
Though I am unsure if this would be an easy task to accomplish or not. 

For the future if it is needed, perhaps a more general solution is to always use the QueryBuilder interface if there is no SearchView for a specific table: do not render the default view for the table.
Alternatively, the express/simple search endpoint and frontend capability can be expanded to accommodate filtering on relationships (and theoretically values of fields/to-one-relationships in dependent resources of the table; or restricting the behavior)

specifyweb/frontend/js_src/lib/components/SearchDialog/index.tsx
```diff
@@ -60,14 +60,15 @@ export function SearchDialog<SCHEMA extends AnySchema>(props: {
   readonly onClose: () => void;
   readonly searchView?: string;
```
maxpatiiuk, line 61: (optional) instead of adding new prop, we could allow passing `false` for `searchView` to explicitly opt out of search view being shown

maxpatiiuk (approved):

PR #4637, opened 2024-03-11
Migrate Workbench to React

sharadsw:
Fixes #4611

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [x] Add relevant issue to release milestone

### Testing instructions
Test all features described here
https://discourse.specifysoftware.org/t/the-specify-7-workbench/540

PR #4777, opened 2024-04-11
Fix disambiguate all incorrectly affectly records

realVinayak:
Fixes **# My issue when doing conversion work**

The problem in the previous code is you could have something like 

image

Here, test1 has duplicates. So, if you click on the test1 on the top row and click disambiguate and perform "Apply All", it incorrectly chose rows _just_ based on the key. Then, it applied the disambiguation to the same mapping path as the one currently active. So, it will apply a disambiguation on the second row with the mapping path of the first one, which is incorrect as test2 will be incorrectly disambiguated as test1.

This leads to backend always choosing test1 for test2. Seems like backend should have logic for not looking at disambiguation when a single match. I don't know use cases for multiple disambiguations for the same record in a single row -- I just tried making the code as semantically same as before. Could get rid of that to just do on all disambiguations in a single row.

I'll defer this PR to 7.9.5 when workbench is in react but opening it right now because this was annoying bug for conversion work. In my case, it led to uniqueness rule exception (a very confusing one) because everything seemed fine on the surface.

### Checklist

- [ ] Add automated tests


maxpatiiuk (approved):
This looks good, but pinging @sharadsw as he is doing a major workbench refactor this now in https://github.com/specify/specify7/pull/4637

any changes we make to the workbench source code in the meanwhile may cause merge conflicts, and would have to be incorporated into the new refactored code

grantfitzsimmons:
Seems like an incredibly urgent issue based on the description. @CarolineDenis What do you think?

sharadsw (approved):

realVinayak:
wouldn't be suprised if there is bad data out there bc of this. we caught it now just bc of business rules (other cases where no business rules will not have been caught)

grantfitzsimmons (approved):

grantfitzsimmons:
@CarolineDenis failing tests 😢

PR #4782, opened 2024-04-12
Filter out empty conditional format def

### LINKED ISSUE #4774: Can save invalid conditional format definitions in table formatter
**Describe the bug**
For conditionally formatted tables, you can add any amount of invalid definitions and save it, which causes infinite load and error messages upon querying/trying to search in the formatted table.

**To Reproduce**

`edge`:
1. Go to App Resources -> Record Formatters -> Any table
2. Check 'Conditional Format'
3. Add a blank field or invalid value (example of 'invalid value' would be putting something besides true/false in a true/false field)
4. Close and save
5. Try to query with `formatted` or use the querycombobox for the table you changed the formatting of
6. See error messages appear, infinite load

`v7.9.3`:
1. Go to App Resources -> DataObjFormatters
2. Add a `<fields/>` or `<fields value="(invalid value here)"/>` tag to one of the formats
3. Save
4. Go through steps 5 and 6 from above

**Expected Behavior**
Unsure, but there shouldn't be error. Issue is also present in v7.9.3.

**Screenshots**
Saving and querying/checking query combo box:

https://github.com/specify/specify7/assets/71564869/eecdea49-839d-48df-b973-400d58dbce5c

Second example of issue:

https://github.com/specify/specify7/assets/71564869/b789f79d-1c3a-4fac-9706-849ea23d2995

In XML formatter (same between `edge` and `v7.9.3`):
image

**Crash Report**
[Specify 7 Crash Report - 2024-04-10T20 41 51.328Z.txt](https://github.com/specify/specify7/files/14937599/Specify.7.Crash.Report.-.2024-04-10T20.41.51.328Z.txt)

- OS: Windows 10
- Browser: Firefox
- Specify 7 Version: `edge`, `v7.9.3`
- Collection name: `Herp-PC`
- User Name: `testuser`

**Additional Context**
Similar in behavior to issue #4747 on the surface, but crash report cites IndexError rather than AttributeError.

_Originally posted by @combs-a in https://github.com/specify/specify7/pull/4757#pullrequestreview-1991848646_

CarolineDenis:
Fixes #4774

### Checklist

### Testing instructions

- Go to App Resources -> Record Formatters -> Any table
- Check 'Conditional Format'
- Add a blank field or invalid value 
- Close and save
- Try to query with formatted 
- [ ]Verify you can query and when opening back the record formatter the empty fields are gone

### DIFF
--- a/specifyweb/frontend/js_src/lib/localization/resources.ts
+++ b/specifyweb/frontend/js_src/lib/localization/resources.ts
@@ -621,27 +621,7 @@ export const resourcesText = createDictionary({
   conditionDescription: {
     'en-us': `
       This format will be used only if the condition field value equals this
-      condition.
-    `,
-    'de-ch': `
-      Dieses Format wird nur verwendet, wenn der Wert des Bedingungsfelds dieser
-      Bedingung entspricht.
-    `,
-    'es-es': `
-      Este formato se utilizará sólo si el valor del campo de condición es igual
-      a esta condición.
-    `,
-    'fr-fr': `
-      Ce format ne sera utilisé que si la valeur du champ de condition est égale
-      à cette condition.
-    `,
-    'ru-ru': `
-      Этот формат будет использоваться только в том случае, если значение поля
-      условия соответствует этому условию.
-    `,
-    'uk-ua': `
-      Цей формат використовуватиметься, лише якщо значення поля умови дорівнює
-      цій умові.
+      condition and is not null.
     `,
   },
   elseConditionDescription: {
--- a/specifyweb/frontend/js_src/lib/localization/resources.ts
+++ b/specifyweb/frontend/js_src/lib/localization/resources.ts
@@ -624,33 +624,6 @@ export const resourcesText = createDictionary({
       condition and is not null.
     `,
   },
-  elseConditionDescription: {
-    'en-us': `
-      This format will be used only if the condition field value equals this
-      condition or if no other format matches it.
-    `,
-    'de-ch': `
-      Dieses Format wird nur verwendet, wenn der Wert des Bedingungsfelds dieser
-      Bedingung entspricht oder wenn kein anderes Format damit übereinstimmt.
-    `,
-    'es-es': `
-      Este formato se utilizará sólo si el valor del campo de condición es igual
-      a esta condición o si ningún otro formato coincide con ella.
-    `,
-    'fr-fr': `
-      Ce format ne sera utilisé que si la valeur du champ de condition est égale
-      à cette condition ou si aucun autre format ne lui correspond.
-    `,
-    'ru-ru': `
-      Этот формат будет использоваться только в том случае, если значение поля
-      условия соответствует этому условию или если ему не соответствует никакой
-      другой формат.
-    `,
-    'uk-ua': `
-      Цей формат використовуватиметься, лише якщо значення поля умови дорівнює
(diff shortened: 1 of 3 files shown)


combs-a (changes requested):
### Testing instructions

- Go to App Resources -> Record Formatters -> Any table
- Check 'Conditional Format'
- Add a blank field or invalid value 
- Close and save
- Try to query with formatted 
- [X] Verify you can query and when opening back the record formatter the empty fields are gone 

Worked well for me--completely blank definitions were removed as well as definitions that had no fields added. It did not remove blank or invalid definitions that had a field (see below for clarity) but these were still ignored in queries.

image

It'd be good to remove the blank definitions regardless of whether or not they have fields, just to make sure everything's cleaned up, but not completely necessary since it doesn't affect queries (from what I can tell).

On a separate note, having some kind of notification/warning to tell the user that there's an invalid conditional format definition so they can go in and fix/remove an invalid, non-blank field would be great, if possible!

Either way, good work! This works great as-is.

emenslin (approved):
### Testing instructions

- Go to App Resources -> Record Formatters -> Any table
- Check 'Conditional Format'
- Add a blank field or invalid value 
- Close and save
- Try to query with formatted 
- [x]Verify you can query and when opening back the record formatter the empty fields are gone 

Everything seems to be working, no longer an error when querrying.

grantfitzsimmons:
We need to change the text from:

> This format will be used only if the condition field value equals this condition or if no other format matches it.

to

> This format will be used only if the condition field value equals this condition and cannot be null.

grantfitzsimmons:
I think we _should_ preserve the ability to have a blank definition in cases where you want no value to be returned when the field meets a specific condition. Because the issue reported by @combs-a is causing errors, I am comfortable removing that capability for now. However, it seems like a feature that would be desirable to retain if it were to function correctly.

grantfitzsimmons:
> Worked well for me--completely blank definitions were removed as well as definitions that had no fields added. It did not remove blank or invalid definitions that had a field (see below for clarity) but these were still ignored in queries.
> 
> image
> 
> It'd be good to remove the blank definitions regardless of whether or not they have fields, just to make sure everything's cleaned up, but not completely necessary since it doesn't affect queries (from what I can tell).


@combs-a This is acceptable behavior for now as this should be supported in the future:
https://github.com/specify/specify7/issues/4787 

For now, the user will just need to set a field and not leave it `null`.

combs-a (approved):

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/localization/resources.ts
```diff
@@ -621,54 +621,7 @@ export const resourcesText = createDictionary({
...
-    `,
-    'ru-ru': `
-      Этот формат будет использоваться только в том случае, если значение поля
-      условия соответствует этому условию или если ему не соответствует никакой
-      другой формат.
-    `,
-    'uk-ua': `
-      Цей формат використовуватиметься, лише якщо значення поля умови дорівнює
-      цій умові або якщо жоден інший формат їй не відповідає.
+      condition and is not null.
```
melton-jason, line 624: @specify/ux-testing
Would `empty` be better here? 
How clear to the average Specify User is the concept of nullity in this context? 

```suggestion
      condition and is not empty.
```

Issue #3238, opened 2023-03-21
Bulk clone records


grantfitzsimmons:
> When entering collection objects from newly collected material, it would be convenient if the collection object could be cloned multiple times.
> For example, if I want to insert 32 identical objects, it would be nice to have a “clone it X times” option where I could set X to 32 to generate 32 records.
> It would probably be inappropriate to have this option for all collections, but is it possible to make it optional on the collection object form, or even put it in a different subform?

**Requested By:** [@cicadina on the Speciforum](https://discourse.specifysoftware.org/t/multiple-records-cloning/1110)

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/multiple-records-cloning/1110/2

PR #4804, opened 2024-04-17
Add series data entry feature

### LINKED ISSUE #545: Series data entry

CarolineDenis:
Fixes #545 

### Checklist

- [ ] Add automated tests

### Testing instructions

- Create a new CO 
- Save 
- Go in the form meta, check the box "Show Bulk Carry Forward count" 
- [ ] Verify the box "Show Bulk Carry Forward count" is displayed only when carry forward is checked 
- [ ] Verify that the Bulk amount input is not displayed on the form if carry forward or "Show Bulk Carry Forward count" is not checked in the form meta
- Go back to the CO form 
- Enter the amount you want to bulk carry 
- Click on carry forward 
- [ ] Verify that a new RS is created 
- [ ] Verify that the new RS has, as a title, the cat number range 
- [ ] Verify once the new RS is created, it opens it


CarolineDenis:
NOTES: 
Records are create, a number of records depending on the input value 
- need to find a way to open those record in the fictions record set

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -204,17 +213,69 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
+        loading(
+          handleClick()
+            .then((result) => {
+              const ids: WritableArray<number> = [];
+              if (originalResourceId !== undefined) {
+                ids.push(originalResourceId);
+              }
+              if (Array.isArray(result)) {
+                result.forEach((newResource) => {
+                  if (handleAdd) handleAdd(newResource);
```
sharadsw, line 225:
```suggestion
                  handleAdd?.(newResource);
```

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -204,17 +213,69 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
+                  ids.push(newResource.id);
+                });
+              } else {
+                if (handleAdd) handleAdd(result);
```
sharadsw, line 229:
```suggestion
                handleAdd?.(result);
```

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -195,7 +201,10 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
   const copyButton = (
     label: LocalizedString,
     description: LocalizedString,
-    handleClick: () => Promise<SpecifyResource<SCHEMA>>
+    handleClick: () =>
+      | Promise<readonly SpecifyResource<SCHEMA>[]>
+      | Promise<SpecifyResource<SCHEMA>>,
```
sharadsw, line 206: Does `Promise<SpecifyResource<SCHEMA>>` need to be readonly as well? Also, would wrapping `<readonly SpecifyResource<SCHEMA>[]` in `RA<>` be more consistent?

sharadsw (approved):
Looks good to me! Only making some optional suggestions

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -195,7 +201,10 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
   const copyButton = (
     label: LocalizedString,
     description: LocalizedString,
-    handleClick: () => Promise<SpecifyResource<SCHEMA>>
+    handleClick: () =>
+      | Promise<RA<SpecifyResource<SCHEMA>>>
+      | Promise<SpecifyResource<SCHEMA>>,
```
maxpatiiuk, line 206:
```suggestion
    handleClick: () => Promise<RA<SpecifyResource<SCHEMA>> | SpecifyResource<SCHEMA>>
```

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -204,17 +213,69 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
+              title={commonText.bulkCarry()}
+              onClick={() => setBulkCarryInput(true)}
+            >
+              {commonText.bulkCarry()}
+            </Button.Save>
+          ) : null}
+          {resource.specifyTable.name === 'CollectionObject' &&
+          !isInRecordSet &&
+          showBulkCarryInput ? (
+            <Label.Inline>
```
maxpatiiuk, line 266: should this be in a dialog?
otherwise, you would need to add aria-controls and forward focus from the button to input on button click

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -225,7 +286,29 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
+                carryForwardAmount > 2
+                ? async () => {
+                    const clones = Array.from(
+                      { length: carryForwardAmount },
+                      async () => {
+                        const clonedResource = await resource.clone(false);
+                        const formatter = clonedResource.specifyTable
+                          .strictGetLiteralField('catalogNumber')
+                          .getUiFormatter()!;
+                        const wildCard = formatter.valueOrWild();
```
maxpatiiuk, line 299: move this code out of the Array.from() since it will produce the same result for each iteration of the loop (because all resources are of the same table)

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
maxpatiiuk: (optional) but it would be even better if you change the type to just always be an array (like we did in case of handleAdd() in other places)
that would simplify the type-checking. the consumer code can then check if array length is 1 or not

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -204,17 +213,69 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
         smoothScroll(form, 0);
-        loading(handleClick().then(handleAdd));
+        loading(
+          handleClick()
+            .then((result) => {
+              const ids: WritableArray<number> = [];
+              if (originalResourceId !== undefined) {
+                ids.push(originalResourceId);
+              }
+              if (Array.isArray(result)) {
```
maxpatiiuk, line 223: change the signature of handleAdd() to make it be callable with an array of resources rather than one resource
and use that to simplify the code here
that would also shift the responsibility to the consumer of Save for deciding how it wants to handle or not to handle the case of more than one resource being present (i.e imagine that some code right now does a redirect on handleAdd() - by calling handleAdd multiples times here, that code is now potentially broken; where as by changing handleAdd signature to include array, that place would now be forced to consider how the array case needs to be handled)

maxpatiiuk: and once handleAdd() is being called with an array, there won't be need for handleCarryBulk() anymore as it can be combined

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -306,8 +307,23 @@ function RecordSet<SCHEMA extends AnySchema>({
...
   ): Promise<void> {
+    if (
+      fromBulkCarry &&
+      typeof ids[0] === 'number' &&
+      typeof ids.at(-1) === 'number'
+    ) {
+      const startingResource = fetchResource('CollectionObject', ids[0]);
+      const endingResource = fetchResource('CollectionObject', ids.at(-1)!);
+      const startingResourceCatNumber = (await startingResource).catalogNumber;
+      const endingResourceCatNumber = (await endingResource).catalogNumber;
```
maxpatiiuk, line 321: once you change the code to remove handleCarryBulk and make handleAdd receive an array of resources, you won't need to do re-fetching of resources manually here as handleAdd is already being called with fetched resource

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
sharadsw: I think this code is meant to avoid for loops when cloning a resource `carryForwardAmount` of times. What would be a better way to do it?

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
sharadsw: Tagging you for a quick review to see if I'm on the right track with this

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
maxpatiiuk: to clarify, I mean that these two variables don't have to be re-computed on each iteration of the loop since they will be the same.
instead, define them above the lop, and use them inside the loop

```ts
                        const formatter = tables.CollectionObject
                          .strictGetLiteralField('catalogNumber')
                          .getUiFormatter()!;
                        const wildCard = formatter.valueOrWild();
```

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSelectorFromIds.tsx
```diff
@@ -238,13 +240,17 @@ export function RecordSelectorFromIds<SCHEMA extends AnySchema>({
...
+        // but isNewRecordSet has been removed in https://github.com/specify/specify7/pull/4904
+        // Can possibly just pass on the isInRecordSet prop received by this component?
         isInRecordSet={!isNewRecordSet}
         isLoading={isLoading || isExternalLoading}
         isSubForm={false}
         resource={resource}
         title={title}
         viewName={viewName}
-        onAdd={handleClone}
+        onAdd={(resources) => resources.map(r => handleClone?.(r))}
```
maxpatiiuk, line 253: make handleClose accept an array here too

```suggestion
        onAdd={handleClone}
```

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
maxpatiiuk: looking better, but still work to do
could you tag me again once the TODOs you added are resolved?

sharadsw: I'm not sure how to get rid of handleCarryBulk. I think I would still need to pass some sort of callback down from `RecordSet.tsx` to `ResourceView.tsx` to create a record set from an array of resources so that the consumer of `Save.tsx` (ResourceView) can define the behavior for when `resources.length > 1`?

Or I can define the behavior for `resources.length > 1` higher up the parent chain (maybe in `RecordSelectorFromIds` or `RecordSet` which could eliminate the need for the callback. What would be better?

sharadsw: moved the bulk carry logic into onClone in RecordSet as that's the same callback being used for adding new resources

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
sharadsw: not sure of the UI requirements for this PR. what do you think, @grantfitzsimmons ?

CarolineDenis: @maxpatiiuk, is there usages of forward focus already in the code?

maxpatiiuk: A dialog would be the simplest if we can, as then all a11y works automatically.

Otherwise have to:
- Add `autofocus` attribute to the input
- Add this to the input: `forwardRef={(input)=>input?.focus()}`

you could look at these examples also, or search for more calls of `.focus()` in the codebase:
https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/components/FormCells/FormTable.tsx#L145
https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/components/Errors/Toasts.tsx#L89

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -412,7 +416,32 @@ function RecordSet<SCHEMA extends AnySchema>({
                 })
             : undefined
         }
-        onClone={(newResource): void => go(totalCount, 'new', newResource)}
+        onClone={(resources) => {
+          resources.map((newResource) => go(totalCount, 'new', newResource));
```
maxpatiiuk, line 420: doens't make sense to call go multiple times since it will just rapidly jump, changing the URL.
instead, go to the first item of the added resources?

```suggestion
          go(totalCount, 'new', resources[0]);
```

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
CarolineDenis: agree

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
CarolineDenis: thank you!

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -214,27 +219,81 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
+                ? async () => {
+                    const formatter =
+                      tables.CollectionObject.strictGetLiteralField(
+                        'catalogNumber'
+                      ).getUiFormatter()!;
+                    const wildCard = formatter.valueOrWild();
+                    const clones = Array.from(
+                      { length: carryForwardAmount },
+                      async () => {
+                        const clonedResource = await resource.clone(false);
```
realVinayak, line 279: Put this out of this function? resource doesn't change

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -214,27 +219,81 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
               formsText.carryForwardDescription(),
-              async () => resource.clone(false)
+              resource.specifyTable.name === 'CollectionObject' &&
+                carryForwardAmount > 2
+                ? async () => {
+                    const formatter =
+                      tables.CollectionObject.strictGetLiteralField(
+                        'catalogNumber'
+                      ).getUiFormatter()!;
+                    const wildCard = formatter.valueOrWild();
```
realVinayak, line 275: Hmm should this feature only be allowed for auto-incrementing catalog numbers?

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -412,7 +416,37 @@ function RecordSet<SCHEMA extends AnySchema>({
...
+              sortFunction((r) => r.id)
+            );
+            loading(
+              createNewRecordSet(
+                sortedResources.map((resource) => resource.id)
+              ).then(async () => {
+                const startingResourceCatNumber =
+                  serializeResource(
+                    sortedResources[0] as SpecifyResource<CollectionObject>
+                  ).catalogNumber ?? '0';
```
realVinayak, line 436: actually, you might be better off using `defined(startingResourceCatNumber)`. Same for endingResourceCatNumber

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
realVinayak: Better yet, it is the same resource in the other branch. Definitely factor it out of this function

realVinayak (changes requested):
Looking good, minor things

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
realVinayak: Oh also, why do we need to serializeResource? Why not `sortedResources[0].get('catalognumber')`

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
CarolineDenis: yes

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
CarolineDenis: same error with defined

grantfitzsimmons (changes requested):
1. I think the string "Show Bulk Carry Forward input amount" should be changed to "Show Bulk Carry Forward count", but open to other options or pushback on this

---

After discussing with Caroline, **these have both been spun off into separate issues.**

_3. It doesn't function inside of a record set as far as I can tell. The new bulk added records don't need to be added to the existing record set, but the feature should work when inside of a record set already. It can create a new record set after the process completes._

   https://github.com/specify/specify7/assets/37256050/d08d83b9-108d-4c64-b12a-5f9d3d83ae1e

_4. The resulting record set is not ordered as one would expect. The resulting set should be ordered by catalog number ascending._

   https://github.com/specify/specify7/assets/37256050/9c3f74a4-b066-4495-8e37-8b3a31dbea2c

CarolineDenis:
NOTES: 
- at the moment we are not ordering the resources in a record set, we would need to implement a new way to order the records based on their ids (@grantfitzsimmons )

grantfitzsimmons:
> NOTES:
> 
> * at the moment we are not ordering the resources in a record set, we would need to implement a new way to order the records based on their ids (@grantfitzsimmons )

Seems related to https://github.com/specify/specify7/issues/4967, maybe we could add support for `OrderNumber` and populate it with the appropriate values?

grantfitzsimmons (approved):
This part looks good– we will likely need to add https://github.com/specify/specify7/issues/5041 and https://github.com/specify/specify7/issues/4967 to "complete" this feature

emenslin (changes requested):
### Testing instructions

- Create a new CO 
- Save 
- Go in the form meta, check the box "Show Bulk Carry Forward count" 
- [x] Verify the box "Show Bulk Carry Forward count" is displayed only when carry forward is checked 
- [ ] Verify that the Bulk amount input is not displayed on the form if carry forward or "Show Bulk Carry Forward count" is not checked in the form meta
- Go back to the CO form 
- Enter the amount you want to bulk carry 
- Click on carry forward 
- [ ] Verify that a new RS is created 
- [ ] Verify that the new RS has, as a title, the cat number range 
- [ ] Verify once the new RS is created, it opens it 

Sometimes when trying to carry forward it throws an error. I think it has to do with the preparation table but it could be something else 


https://github.com/specify/specify7/assets/150940840/17a18b75-7474-4a8a-a0d0-cf54c580acc9





Screenshot 2024-07-02 084902


[Specify 7 Crash Report - 2024-07-02T13_49_06.908Z.txt](https://github.com/user-attachments/files/16069594/Specify.7.Crash.Report.-.2024-07-02T13_49_06.908Z.txt)

It still creates the records but it does not create the record set. 
Without any information in the preparations table it seems to overall be working and a rs is created but I'm not sure if other tables would also cause errors. 

--- 

Also the bulk carry forward count is always shown even if you uncheck the box, it only gets removed when the 'show carry forward button' box is unchecked 


https://github.com/specify/specify7/assets/150940840/91563d56-bb50-4e1c-83a9-2cab88f84f56

acwhite211:
I have a potential solution on this branch https://github.com/specify/specify7/tree/issue-545-collection-dispatch-bulk
It adds a collection_dispatch_bulk function, so carry forward can make just one API call, instead potentially a very large amount.

https://github.com/specify/specify7/blob/af1caec8cd9ba9f43dc9942f66ea257a494ae8d1/specifyweb/specify/urls.py#L20

https://github.com/specify/specify7/blob/af1caec8cd9ba9f43dc9942f66ea257a494ae8d1/specifyweb/specify/api.py#L280-L308

acwhite211:
Looks like the database deadlock is due to this lock during autonumbering
https://github.com/specify/specify7/blob/080b32ceaec14d747291bbb020d6329f48dcedc0/specifyweb/specify/autonumbering.py#L45-L49

specifyweb/specify/api.py
```diff
@@ -239,10 +241,33 @@ def collection_dispatch(request, model) -> HttpResponse:
...
-                            model, json.load(request),
-                            request.GET.get('recordsetid', None))
+        try:
+            obj = post_resource(request.specify_collection,
+                                request.specify_user_agent,
+                                model, json.load(request),
+                                request.GET.get('recordsetid', None))
+        # Handle deadlocks by retrying the operation, fixes deadlock issues to some extent.
+        # It's better to use collection_dispatch_bulk instead for bulk operations.
+        # Note: This is a temporary fix, remove when the front-end is updated to use bulk collection_dispatch_bulk.
```
realVinayak, line 251: Yep, nice idea, was going to commit about the same.

specifyweb/specify/api.py
```diff
@@ -252,6 +277,36 @@ def collection_dispatch(request, model) -> HttpResponse:
...
+                request.specify_collection,
+                request.specify_user_agent,
+                model,
+                obj_data,
+                request.GET.get("recordsetid", None),
+            )
+            resp_objs.append(_obj_to_data(obj, checker))
+        
+        resp = HttpResponseCreated(toJson(resp_objs),
+                                   content_type='application/json')
```
realVinayak, line 303: you'll have to do `safe=False` here

specifyweb/specify/api.py
```diff
@@ -252,6 +277,36 @@ def collection_dispatch(request, model) -> HttpResponse:
 
     return resp
 
+def collection_dispatch_bulk(request, model) -> HttpResponse:
```
realVinayak, line 280: Something to consider: lock_tables will commit the current transaction. we'll not have inconsistent states but it is not ideal to keep a connection and committing multiple times (from the user's perspective, the records should be atomic...). But, this was also a problem before so no code changes are necessary

realVinayak (changes requested):
Backend looks fine. The issue with deadlock was concurrent collection object save() calls. That is, doing save() [awaiting](https://github.com/specify/specify7/tree/issue-545-serial) the previous did fix the problem. But, a backend solution will be better for performance

specifyweb/specify/api.py
```diff
@@ -252,6 +277,36 @@ def collection_dispatch(request, model) -> HttpResponse:
...
+    """
+    Do the same as collection_dispatch, but for bulk POST operations.
+    Call this endpoint with a list of objects of the same type to create.
+    This reduces the amount of API calls needed to create multiple objects, like when creating multiple carry forwards.
```
melton-jason, line 284: While not needed in this PR, maybe it would be worth extending this bulk operation endpoint to also include batch editing and deletion

specifyweb/specify/api.py
```diff
@@ -252,6 +277,36 @@ def collection_dispatch(request, model) -> HttpResponse:
...
+    """
+    resp: HttpResponse
+    checker = table_permissions_checker(request.specify_collection, request.specify_user_agent, "read")
+
+    if request.method == 'POST':
+        data = json.load(request)
```
melton-jason, line 290: Not sure if the two are semantically equivalent (it looks to be the case), but I would rather use a `json.loads(request.body)` here instead of `json.load`. 

https://docs.djangoproject.com/en/5.0/ref/request-response/#django.http.HttpRequest.body. 

Or by using `json.load` do we also handle the case where the client does not use a traditional body and instead uses form data?

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -412,7 +416,33 @@ function RecordSet<SCHEMA extends AnySchema>({
                 })
             : undefined
         }
-        onClone={(newResource): void => go(totalCount, 'new', newResource)}
+        onClone={(resources: RA<SpecifyResource<SCHEMA>>): void => {
+          go(totalCount, 'new', resources[0]);
+          if (resources.length > 1) {
```
maxpatiiuk, line 421: if a new record set is created, shouldn't you `go()` to the newly created record set?
or does it already do that?

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -214,27 +221,82 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
-              formsText.clone(),
-              formsText.cloneDescription(),
-              async () => resource.clone(true)
-            )}
+          {resource.specifyTable.name === 'CollectionObject' &&
+          (isInRecordSet === false || isInRecordSet === undefined) &&
+          isSaveDisabled &&
+          showCarry &&
+          showBulkCarry ? (
+            <Label.Inline>
```
maxpatiiuk, line 235: this Label does not have any actual label text?
if you can't put label text for UI reasons, then remove Label and add placeholder and aria-label attributes

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -214,27 +221,82 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
+              <Input.Generic
+                type="number"
+                value={carryForwardAmount}
+                onValueChange={(value): void =>
+                  carryForwardAmount === undefined
```
maxpatiiuk, line 240: shouldn't this check for `value === undefined` instead?

maxpatiiuk: any why are you using Input.Generic here? if you use Input.Number, then the number will be parsd for you

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -214,27 +221,82 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
+                          'catalogNumber',
+                          wildCard as never
+                        );
+                        return clonedResource;
+                      }
+                    );
+
+                    const clones = await Promise.all(clonePromises);
+
+                    const backendClones = await ajax<
```
maxpatiiuk, line 274: the code is clonning a resource x times and then sending that as JSON to the back-end
do we have the capability to do cloning on the back-end? i.e can we just send a base record and a number to the back-end and let it do the rest? would be more bandwidth efficient 🙂

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
CarolineDenis: it does it yes

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
CarolineDenis: @acwhite211

acwhite211: Ya, I had a similar thought.  The request body size would get pretty massive when someone does a carry-forward with 100 plus copies.  I added an api `/api/specify/bulk/collectionobject/99/` that calls a collection_dispatch_bulk_copy function.  The api input is the same as the colleection_dispatch, you just add a number to the end of the api request for the amount of copies.  When could change it to be `/api/specify/collectionobject?copies=99`, that might be more clear, let me know what you think.

acwhite211: So, we decided that while the new API works in reducing the request body size, the front-end currently still needs the response body to return all the newly created records to create a record set.  This defeats the purpose of the new `collection_dispatch_bulk_copy` function.  So, we'll keep the API for now, but we'll need to wait to call it until the front-end can handle not receiving all the created records in the response.  We could implement it now only if we are for sure that we don't need the front-end to create a record set of the new data.

alesan99 (approved):
- Create a new CO 
- Save 
- Go in the form meta, check the box "Show Bulk Carry Forward count" 
- [x] Verify the box "Show Bulk Carry Forward count" is displayed only when carry forward is checked 
- [x] Verify that the Bulk amount input is not displayed on the form if carry forward or "Show Bulk Carry Forward count" is not checked in the form meta
- Go back to the CO form 
- Enter the amount you want to bulk carry 
- Click on carry forward 
- [x] Verify that a new RS is created 
- [x] Verify that the new RS has, as a title, the cat number range 
- [x] Verify once the new RS is created, it opens it 

Aside from the record set issue discussed above, the settings are now being respected reliably 👍👍👍

https://github.com/specify/specify7/assets/8108710/170c729f-cda1-475c-80dc-b100c6aa11c6

acwhite211 (approved):

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -232,17 +232,18 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
-                    ? setCarryForwardAmount(1)
-                    : setCarryForwardAmount(Number.parseInt(value))
-                }
-              />
-            </Label.Inline>
+            <Input.Integer
+              aria-label={formsText.bulkCarryForwardCount()}
+              className="!w-fit"
+              min={1}
+              placeholder="1"
```
maxpatiiuk, line 239: hmm, placeholder of 1 is still not clear what this field is about
could you either set placeholder to `formsText.bulkCarryForwardCount()`, or add a `title={formsText.bulkCarryForwardCount()}`
edit: I have just noticed that there is a button visible right after the field - that makes it more clear what the field is about so the way it is now is ok!

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -232,17 +232,18 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
+              value={carryForwardAmount}
+              onValueChange={(value): void =>
+                value === undefined
+                  ? setCarryForwardAmount(1)
+                  : setCarryForwardAmount(value)
+              }
```
maxpatiiuk, line 245: is value ever undefined? not according to it's typescript type it seems

or did you mean to handle the case when the user erased the value from the field? we don't have that case well in specify now but probably should - `value` would be NaN in such cases, not undefined

maxpatiiuk (approved):

Areyes42 (changes requested):
### Testing instructions

-   Create a new CO
-   Save
-   Go in the form meta, check the box "Show Bulk Carry Forward count"
-   [x] Verify the box "Show Bulk Carry Forward count" is displayed only when carry forward is checked
-   [x] Verify that the Bulk amount input is not displayed on the form if carry forward or "Show Bulk Carry Forward count" is not checked in the form meta
-   Go back to the CO form
-   Enter the amount you want to bulk carry
-   Click on carry forward
-   [ ] Verify that a new RS is created
-   [ ] Verify that the new RS has, as a title, the cat number range
-  [ ]  Verify once the new RS is created, it opens it
***
I tested this PR yesterday and it was working fine. It's now throwing a TypeError when trying to bulk carry. Tested on multiple COs and DBs.

[Specify 7 Crash Report - 2024-07-09T16_28_03.361Z.txt](https://github.com/user-attachments/files/16146710/Specify.7.Crash.Report.-.2024-07-09T16_28_03.361Z.txt)

https://github.com/specify/specify7/assets/130881183/90440fea-6963-4fbc-b60f-20013a815fea

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
CarolineDenis: yes I wanted to handle to case where we have the user erasing the field (like the issue we had in bulk resolve bulk loaned) will change to NaN

Areyes42 (approved):
### Testing instructions

-   Create a new CO
-   Save
-   Go in the form meta, check the box "Show Bulk Carry Forward count"
-  [x]  Verify the box "Show Bulk Carry Forward count" is displayed only when carry forward is checked
-   [x] Verify that the Bulk amount input is not displayed on the form if carry forward or "Show Bulk Carry Forward count" is not checked in the form meta
-   Go back to the CO form
-   Enter the amount you want to bulk carry
-   Click on carry forward
-   [x] Verify that a new RS is created
-   [x] Verify that the new RS has, as a title, the cat number range
-   [x] Verify once the new RS is created, it opens it
***
Retested, everything looks good!

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
```diff
@@ -214,27 +221,83 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
...
+                    const wildCard = formatter.valueOrWild();
+
+                    const clonePromises = Array.from(
+                      { length: carryForwardAmount },
+                      async () => {
+                        const clonedResource = await resource.clone(false);
+                        await clonedResource.set(
+                          'catalogNumber',
+                          wildCard as never
+                        );
```
melton-jason, line 268: This makes it so that trying to bulk carry forward catalog numbers with a non-incrementing format will result in an error.

Consider the two following formats for catalognumber:

```xml
  <format system="false" name="CatalogNumberTest" class="edu.ku.brc.specify.datamodel.CollectionObject" fieldname="catalogNumber">
    <field type="year" size="4" value="YEAR" byyear="true"/>
    <field type="separator" size="1" value="-"/>
    <field type="numeric" size="3" inc="false"/>
  </format>
```

In this case, the `wildCard` will be `2024-###`, which is an invalid catalognumber and the backend throws an exception stating this: 

`FormatMismatch at /api/specify/bulk/collectionobject/ value '2024-###' doesn't match formatter YEAR-### `

```xml
  <format system="false" name="CatalogNumberRegex" class="edu.ku.brc.specify.datamodel.CollectionObject" fieldname="catalogNumber" default="true">
    <field type="regex" size="6" minsize="4" value="[0-9]{4}[a-zA-Z]{0,2}" pattern="####[AA]"/>
  </format>
```
In this case, `wildCard` will be the `value` of the regex itself and the backend will fail with a similar exception: 

`FormatMismatch at /api/specify/bulk/collectionobject/ value '[0-9]{4}[a-zA-Z]{0,2}' doesn't match formatter [0-9]{4}[a-zA-Z]{0,2}`

melton-jason: I am not sure how we'd want to resolve this. 
(Besides disabling this feature for those cases, but that is far from the ideal). 

I would imagine an ideal workflow might be to navigate to a RecordSet-like interface with the cloned records (without saving them to the backend) when they wish to bulk carry forward. 
And having a button (which could be the normal Save Button), which only allows saving when all records have no save blockers. 

Do we know if Specify 6 handles these cases?

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
melton-jason: What is stopping us from returning the same response (an array of the created resources) from the `new collection_dispatch_bulk_copy` endpoint?

We are still calling `post_resource` for each record we create, which returns the created object. 

https://github.com/specify/specify7/blob/bc06abf437cae5fd0debb3838e3c0a64b9ed184e/specifyweb/specify/api.py#L289-L300

Was there somehow trouble regarding serializing the objects? It should be exactly the same as the current endpoint we are using: 

https://github.com/specify/specify7/blob/bc06abf437cae5fd0debb3838e3c0a64b9ed184e/specifyweb/specify/api.py#L268-L278

specifyweb/specify/api.py
```diff
@@ -252,6 +252,53 @@ def collection_dispatch(request, model) -> HttpResponse:
...
+
+def collection_dispatch_bulk_copy(request, model, copies) -> HttpResponse:
+    checker = table_permissions_checker(request.specify_collection, request.specify_user_agent, "read")
+
+    if request.method != 'POST':
+        return HttpResponseNotAllowed(['POST'])
+
+    data = json.loads(request.body)
+    data = dict(filter(lambda item: item[0] != 'id', data.items())) # Remove ID field before making copies
+    # resp_objs = [] # For now, we don't need to return the created objects to avoid sending large amounts of data
```
melton-jason, line 288: > For now, we don't need to return the created objects to avoid sending large amounts of data 

If we are concerned about response size, then the current endpoint would need to be modified and optimized for storage as well. 
It returns the exact same thing: an array of serialized resources. 

Either return an array of serialized resources here if we don't care about response size, or find a way to reduce the size on the current endpoint (possibly by just returning an array of ids?)

specifyweb/specify/schema.py
```diff
@@ -461,7 +461,102 @@ def table_to_endpoint(table: Table) -> List[Tuple[str, Dict]]:
...
+                            "schema": {
+                                "type": "number",
+                                "minimum": 1,
+                            },
+                            "description": "Number of copies to create",
+                        },
+                    ],
+                    "requestBody": {
+                        "required": True,
+                        "description": "An array containing a JSON representation of the objects to create",
```
melton-jason, line 531:
```suggestion
                        "description": "A serialized JSON representation of the object to clone",
```

specifyweb/specify/schema.py
```diff
@@ -461,7 +461,102 @@ def table_to_endpoint(table: Table) -> List[Tuple[str, Dict]]:
...
+                            "description": "An array containing the newly created objects",
+                            "content": {
+                                "application/json": {
+                                    "schema": {
+                                        "type": "array",
+                                        "items": {
+                                            "$ref": f"#/components/schemas/{table.django_name}",
+                                        }
+                                    }
+                                }
```
melton-jason, line 551: This is not a possible response for the endpoint currently

specifyweb/frontend/js_src/lib/components/FormSliders/RecordSet.tsx
```diff
@@ -412,7 +416,33 @@ function RecordSet<SCHEMA extends AnySchema>({
...
+                    sortedResources.at(-1) as SpecifyResource<CollectionObject>
+                  ).catalogNumber ?? '0';
+                recordSet.set(
+                  'name',
+                  `Batch #${startingResourceCatNumber} - #${endingResourceCatNumber}`
+                );
+                await recordSet.save();
+              })
+            );
+          }
```
melton-jason, line 444: We only want to execute this if the table is `CollectionObject`, no? 
This code could conceptually be from any table, which is why you received the errors mentioned in
https://github.com/specify/specify7/pull/4804/files#r1649414105. 

Make sure this code is _only_ run if recources.length > 1 and the table is `CollectionObject`.
Then you can proceed with the suggestions in https://github.com/specify/specify7/pull/4804/files#r1649414105 (no need to serialize the resources, just use `resource.get('catalogNumber')`). 

If typescript is not smart enough to infer that `sortedResources` is of type `RA<SpecifyResource<CollectionObject>>`, then we can do type casting because we _know_ they types will always be correct within the if statement.

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
realVinayak: > In this case, the wildCard will be 2024-###, which is an invalid catalognumber and the backend throws an exception stating this:

Isn't that a valid catalog number though? Pretty sure that's accepted....

I think it is rare enough for users to want to use this in the first place..... makes it even more unlikely for rege ones..

melton-jason: > Isn't that a valid catalog number though? Pretty sure that's accepted....

No, it's not accepted by the backend or frontend. 
In the format, that part of the formatter is explicitly defined as a non-incrementing numeric field. 
The pattern for the field remains `2024-###`, but is marked as invalid and only numeric digits are allowed. 

If it's supposed to be accepted, or was at some point, then should the backend just treat `###` as it would any auto-incrementing numeric formatter? 

--- 

> I think it is rare enough for users to want to use this in the first place..... makes it even more unlikely for rege ones..

Probably a fair point. I am unsure about how many people used this feature in Specify 6 or requested this (apart from the mentions in #545 and #3238), though I do know that Entomologists would most likely use and benefit from this feature. 

At the very least, instead of erroring we should disable this feature then. 

I'll take a look into how Specify 6 handles these cases.

realVinayak: Ah, I see that is not incremented now. I don't think it is worth considering. Worse edge cases have not been handled elsewhere

specifyweb/frontend/js_src/lib/components/FormMeta/CarryForward.tsx
```diff
@@ -142,6 +143,48 @@ export function CarryForwardConfig({
...
+export const tableValidForBulkClone = (table: SpecifyTable): boolean =>
+  table === tables.CollectionObject &&
+  !(
+    tables.CollectionObject.strictGetLiteralField('catalogNumber')
+      .getUiFormatter()
+      ?.fields.some(
+        (field) =>
+          field.type === 'regex' ||
+          (field.type === 'numeric' && !field.canAutonumber())
+      ) ?? false
```
realVinayak, line 185: This should be `true` instead of false? 

Also, why doesn't `getUiFormatter().canAutonumber()` not be enough?

realVinayak (dismissed):

specifyweb/frontend/js_src/lib/components/FormMeta/CarryForward.tsx
melton-jason: I used `false` instead of `true` to be more permissive rather than restrictive if `getUiFormatter` returns undefined. 

---

`getUiFormatter().canAutonumber()` would be enough to handle the second case (doing this would also handle the case for an AlphaNumeric field on a UIFormatter which contains an auto-incrementing section), but not if the Formatter contains a `regex` field, which is a case we we'd probably want to disable. 

i..e, they can have a regex type field in a formatter which would cause an error

```xml
 <field type="regex" size="6" minsize="4" value="[0-9]{4}[a-zA-Z]{0,2}" pattern="####[AA]"/>
```

Even if it's a case we shouldn't consider fixing (or just aren't right now for whatever reason), I think it's worth disabling in the UI if we _know_ it's going to always cause an error. 
Similar to https://github.com/specify/specify7/pull/5001#pullrequestreview-2116275954 and https://github.com/specify/specify7/pull/5001#issuecomment-2167039171

Errors can absolutely be beneficial to users and to us, but I think it's just a hindrance in this case. 
Let me know what you think. This was merged before I got back to this review, so hopefully not to late to make changes if needed.

grantfitzsimmons:
Need to now implement https://github.com/specify/specify7/pull/4952

specifysoftware:

PR #4726, opened 2024-04-02
Restore bad tree structure saveBlocker

melton-jason:
Fixes #4717

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [x] Add automated tests
- [x] Add relevant issue to release milestone

### Testing instructions
- [ ] On a new Taxon record, set the Parent to be a rank which is not the lowest in the tree
- [ ] Clear the Parent QueryComboBox (the tree rank should clear as well) 
- [ ] Set the Parent of the Taxon to be a node which is the lowest rank in the tree
- [ ] A saveBlocker should be set on the Parent field which reports `Bad Tree Structure` and the Tree Rank picklist should become disabled
- [ ] Change the parent to a Taxon in which the possible ranks do not include the current rank (definitionItem) of the new Taxon record. (See  https://github.com/specify/specify7/pull/4726#pullrequestreview-1974967848)
- [ ] The tree rank should be set accordingly to the lowest enforceable rank 

- [ ] On an existing tree record, ensure the form opens without unload protect being triggered (as in #3259)
- [ ] The tree rank should not be overwritten with a new value, and should use the existing rank of the record as shown in the Tree Viewer (especially when there are multiple possible ranks the node could take, see https://github.com/specify/specify7/pull/4659#issue-2187680064)

PR #4811, opened 2024-04-19
Don't unset rank for root level tree node

### LINKED ISSUE #4807: Viewing highest ranked node causes unload protect
**Describe the bug**
When clicking on the highest node in a tree to view the record if you try to close out of it you get the unload protect.

**To Reproduce**

1. Go to any tree (e.g. storage) 
2. Click on the highest ranking node-> edit
3. Try to close out of the dialog

**Expected behavior**
You should be able to close out of the dialog since no changes were made 

**Screenshots**

Edge

https://github.com/specify/specify7/assets/150940840/d7144941-b527-45d2-80a1-775f8b193beb

7.9.3

https://github.com/specify/specify7/assets/150940840/284f1555-ff45-493d-b2e7-0b2983f5786f

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge
- Collection name: mammals
- User Name: Spadmin
[Specify 7 System Information - 2024-04-18T16_27_44.330Z.txt](https://github.com/specify/specify7/files/15027699/Specify.7.System.Information.-.2024-04-18T16_27_44.330Z.txt)

melton-jason:
Fixes #4807

Related #4726 

### Checklist

### Testing instructions
In the TreeViewer for any tree table: 
- [ ] Open the form for the tree root level node and ensure the rank is correctly set and unload protect is not triggered
- [ ] Open the form for any existing tree node at the lowest level and ensure the rank is correctly set and unload protect is not triggered
- [ ] Open the form for any existing tree node at any intermediate rank (not highest or lowest) and ensure the rank is correctly set and unload protect is not triggered

---

- Inspect a tree node which is invalid (such as the Reptilia or Amphibia nodes from `sdnhm_herps_2_8_24`: https://sdnhmherps2824-issue-4807.test.specifysystems.org/specify/tree/taxon/?conformation=%7E%7E1--) 
- [ ] When viewing the record, the Tree Rank (definitionItem) should not change to match the lowest possible enforceable rank
- [ ] The api link of the previous definitionIten remains in the picklist, with the options for the 'correct' ranks remaining as items in the picklist
- [ ] A `bad-tree-structure` saveBlocker should be set on the Parent field indicating the error


emenslin (approved):
### Testing instructions
In the TreeViewer for any tree table: 
- [x] Open the form for the tree root level node and ensure the rank is correctly set and unload protect is not triggered
- [x] Open the form for any existing tree node at the lowest level and ensure the rank is correctly set and unload protect is not triggered
- [x] Open the form for any existing tree node at any intermediate rank (not highest or lowest) and ensure the rank is correctly set and unload protect is not triggered

Looks good, unload protect is not triggered anywhere on the trees!

https://github.com/specify/specify7/assets/150940840/f4b121a1-b50b-4882-8aaa-58582136777b

bronwyncombs (approved):
on `fwri` -- geography & taxon tested
In the TreeViewer for any tree table:
- [x] Open the form for the tree root level node and ensure the rank is correctly set and unload protect is not triggered
- [X] Open the form for any existing tree node at the lowest level and ensure the rank is correctly set and unload protect is not triggered
- [X] Open the form for any existing tree node at any intermediate rank (not highest or lowest) and ensure the rank is correctly set and unload protect is not triggered

combs-a:
I was going through level by level in sdnhm_herps_2_8_24, and found that I was still getting unload protections on Amphibia and Reptilia, as well as the subspecies that display by themselves on the tree. Checking against the Placeholder branch (which goes all the way down) shows that it's not a specific issue with class, since that one does not have an unload protection issue.

image

image

This issue is in `edge`, but not `v7.9.3`. Might also be a database specific thing?

Everything else was working correctly!

specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
```diff
@@ -142,11 +142,14 @@ export function TreeLevelComboBox(props: DefaultComboBoxProps): JSX.Element {
...
+        const invalidDefinitionItem =
+          typeof definitionItem !== 'string' ||
+          !(items?.map(({ value }) => value).includes(definitionItem) ?? true);
+
         if (
           (items !== undefined ||
             typeof resource?.get('parent') !== 'string') &&
-          (typeof definitionItem !== 'string' ||
-            !items?.map(({ value }) => value).includes(definitionItem))
+          invalidDefinitionItem
```
realVinayak, line 152: Maybe things can be simplified if we do not have a resource.onchange here. Or does that create more bugs? The items would be set by the previous effect so this second effect can just have items as part of the dependency.

grantfitzsimmons:
From @melton-jason:
> Regardless, I would imagine this specific case to be an edge/outlier: The application prevents the user from creating taxon records like Reptilia or Amphibia. These records were created unorthodoxly (e.g, directly via SQL) or the configuration of the trees was changed after the records were created.
> (This case should still be handled better, but the underlying issue is not caused by this PR and is hopefully not likely to be encountered in the wild.
> Can an Issue be opened to maintain a better record of the problem?)

Presumably these were created before the tree definition was modified (a case that will become increasingly common as we add it to Specify 7 directly in the future). Much more likely than using SQL, and not uncommon in Specify 6. I'm not a fan of changing the rank automatically, even if it is "helpful", as there is never a time where an existing Taxon record should change rank in the tree even if it does not conform to the configuration in the tree definition.

>Would you prefer the rank to remain the same in this case and set a `bad-tree-structure- save blocker?

I would prefer it to trigger a blocker warning the user that they need to modify the record themselves.

specifyweb/frontend/js_src/lib/components/PickLists/TreeLevelPickList.tsx
melton-jason: Fixed in https://github.com/specify/specify7/pull/4811/commits/e02e4eb4bbac53b161c4d58ecfcebc9ce7f23efd

melton-jason:
> I would prefer it to trigger a blocker warning the user that they need to modify the record themselves. You can determine if that should be a different issue or it is outside the scope of this PR, but we should not try to be "helpful" when opening an existing record unless absolutely necessary

I agree that modifying the record without any indication to the user should only be performed when **absolutely** expected. 
We don't want to to make accidental changes which were unintended. 

The behavior now ONLY for the case mentioned in https://github.com/specify/specify7/pull/4811#issuecomment-2066958808 is the following: 
- When viewing the record, the Tree Rank (definitionItem) should not change to match the lowest possible enforceable rank
- Instead, the api link of the previous definitionItem will remain in the picklist, with the options for the 'correct' ranks remaining as items in the picklist
- A `bad-tree-structure` saveBlocker should be set on the Parent field indicating the error

(testing instructions updated)

If it is desired, we can also always fetch the name and title for the definitionItem of the tree resource (regardless of whether it is valid), and include that in the picklist. The saveBlocker will still be set when the rank is set to a value which makes the tree structure invalid

combs-a (changes requested):
### Testing instructions
In the TreeViewer for any tree table: 
- [X] Open the form for the tree root level node and ensure the rank is correctly set and unload protect is not triggered
- [X] Open the form for any existing tree node at the lowest level and ensure the rank is correctly set and unload protect is not triggered
- [X] Open the form for any existing tree node at any intermediate rank (not highest or lowest) and ensure the rank is correctly set and unload protect is not triggered

Tested these on sdnhm_herps using the placeholder nodes, didn't run into any issues.

I don't see this as big enough to request changes over (since this is how it is in `v7.9.3`), but you *can* currently add a parent to the root node (both in sdnhm and in lsumzmammals). It gets blocked by a bad tree structure error, so it shouldn't cause crashes. Maybe setting the parent of root nodes to read only sometime in the future would be nice though.

---

- Inspect a tree node which is invalid (such as the Reptilia or Amphibia nodes from `sdnhm_herps_2_8_24`: https://sdnhmherps2824-issue-4807.test.specifysystems.org/specify/tree/taxon/?conformation=%7E%7E1--) 
- [X] When viewing the record, the Tree Rank (definitionItem) should not change to match the lowest possible enforceable rank
- [X] The api link of the previous definitionIten remains in the picklist, with the options for the 'correct' ranks remaining as items in the picklist
- [X] A `bad-tree-structure` saveBlocker should be set on the Parent field indicating the error

All tests passed, but the API link being in the tree is kind of strange, even though it does make it obvious that it should probably be changed. I'm not sure what'd be the best for users, but I think that having the rank displayed regardless of validity like Jason said may be a bit better

Just going to request changes on this for now for further discussion before merging. Good work! 👍

maxpatiiuk (approved):

melton-jason:
> All tests passed, but the API link being in the tree is kind of strange, even though it does make it obvious that it should probably be changed. I'm not sure what'd be the best for users, but I think that having the rank displayed regardless of validity like Jason said may be a bit better
> 
> Just going to request changes on this for now for further discussion before merging. Good work! 👍

https://github.com/specify/specify7/pull/4811#pullrequestreview-2018270322

For now, this case will remain as-is with the API link for the upcoming `v7.9.4` release, but can be addressed in the future.

pashiav (changes requested):
### Testing instructions
In the TreeViewer for any tree table: 
- [X] Open the form for the tree root level node and ensure the rank is correctly set and unload protect is not triggered
- [X] Open the form for any existing tree node at the lowest level and ensure the rank is correctly set and unload protect is not triggered
- [X] Open the form for any existing tree node at any intermediate rank (not highest or lowest) and ensure the rank is correctly set and unload protect is not triggered
---
- Inspect a tree node which is invalid (such as the Reptilia or Amphibia nodes from `sdnhm_herps_2_8_24`: https://sdnhmherps2824-issue-4807.test.specifysystems.org/specify/tree/taxon/?conformation=%7E%7E1--) 
- [ ] When viewing the record, the Tree Rank (definitionItem) should not change to match the lowest possible enforceable rank
- [ ] The api link of the previous definitionIten remains in the picklist, with the options for the 'correct' ranks remaining as items in the picklist
- [ ] A `bad-tree-structure` saveBlocker should be set on the Parent field indicating the error

---
Great work so far!!

_- Inspect a tree node which is invalid (such as the Reptilia or Amphibia nodes from `sdnhm_herps_2_8_24`: https://sdnhmherps2824-issue-4807.test.specifysystems.org/specify/tree/taxon/?conformation=%7E%7E1--)_ 

Though the buttons for **Clone**, **Add**, and **Save** seem to be blocked, they can still be clicked. When clicking **Clone** or **Add** then clicking **Close**, the unload protect is triggered.

 Also, the taxonomic rank is displayed as a API URL.

`issue-4807` | `edge`



https://github.com/specify/specify7/assets/118556884/247bcd21-a9cf-445a-8647-54ff75d2a4d8



Amphibia got moved to Kingdom from Class after Save on `edge`.
Screen Shot 2024-04-24 at 12 30 32 PM

combs-a (approved):

melton-jason:
> Though the buttons for Clone, Add, and Save seem to be blocked, they can still be clicked. When clicking Clone or Add then clicking Close, the unload protect is triggered.

https://github.com/specify/specify7/pull/4811#pullrequestreview-2020506125

I would not consider this bug to be a product of this PR. 
As of now,  the Copy, Carry Forward, and Clone buttons become red and and have the `not-allowed` cursor whenever the form contains a saveBlocker. 
They become disabled whenever the resource is new, has been changed, or is saving. 

At the moment, these conditions (showing red / being disabled) can be mutually exclusive when an existing record contains a saveBlocker and has not been changed. 

What would the preferred behavior here be? 
I would imagine the buttons should also become disabled when there is a saveBlocker on the form, or the not-allowed cursor can be relaxed in this case if we want to allow cloning, carry forward, or adding. 

Regardless, this can be addressed outside of this PR as it is (to an extent) not related or caused by the changes in this PR.

PR #4867, opened 2024-05-01
Revised terminology in dialogs

grantfitzsimmons:
This PR exists to make the application consistent with the terminology used in our documentation.


specifyweb/frontend/js_src/lib/localization/resources.ts
```diff
@@ -733,28 +733,13 @@ export const resourcesText = createDictionary({
     'uk-ua': 'Запустити в колекції',
   },
   createNewForm: {
-    'en-us': 'Create new form',
-    'de-ch': 'Definition',
-    'es-es': 'Crear nuevo formulario',
-    'fr-fr': 'Créer un nouveau formulaire',
-    'ru-ru': 'Создать новую форму',
-    'uk-ua': 'Створити нову форму',
+    'en-us': 'Create a new view definition',
```
maxpatiiuk, line 736: Hmm, I was actually hoping to retire `view definition` in favor of `form`

I.e `Form Editor` is clear to everyone
`View definition editor` is long, a pain to translate, and not clear
plus, it's ambiguous - `view the editor of definitions` or `the editor for view definitions`

view is a term that comes from sp6 and java. forms comes from html

CarolineDenis:
I like Max's suggestion to change the view def in forms

grantfitzsimmons:
> I like Max's suggestion to change the view def in forms

I've thought a lot about it, and I think it makes much more ambiguous and disjoint from what the user sees in the XML.
When I say "Forms", it immediately becomes unclear whether I'm referring to the app resource "Form definition" or an individual `<viewdef>` ("Form"). 

> I.e Form Editor is clear to everyone
> View definition editor is long, a pain to translate, and not clear
> plus, it's ambiguous - view the editor of definitions or the editor for view definitions

When you select a **Form Definition** app resource from the menu, you see a list of tables. You can click on the table you wish to edit to view its definitions. You are editing a specific view definition within the form definition for a particular table.

The XML doesn't reflect the proposed nomenclature and the many hours I have put into creating documentation and recording videos for this update have taught me that making this as clear as possible is very important (though difficult). I tried to explain it before these changes, but it was almost a necessity as "forms" didn't make it clear what I was referring to, and when editing XML it just didn't make sense.

If the user has to specifically reference a particular "form" in this case, do I say "Put the name of the form in the `viewname=""` attribute for the cell"?

```
<row>
   <cell type="subview" viewname="CollectingEventSub" id="21"  name="collectingEvent" colspan="14" rows="5"/>
</row>
```

To me, it's immediately ambiguous as to whether they need to put the name of the form definition (app resource) or a specific "form" under a specific tables "forms". Each "form" would also be captured within a `<viewdef>`, further abstracting the user from the actual logic that is confusing enough as it is in the first place.

What about `subview`? The `view` elements? These are things the user will see and interact with often, and naming it differently at this stage would likely be a mistake. As long as the user has to edit this manually, we need to make it clear.

When defining a `formtable` view definition (which _still_ must be done manually in `v7.9.4` to define a separate definition for the columns of a collapsed table), it also would upset the terminology used in the system itself.

Once editing XML is obsolete, we should switch. As you agree, "Form" is a much better name for the concept. Until a strictly visual editor is available, it's best to keep these concepts named separately to ensure clarity for users interacting with the system.

specifyweb/frontend/js_src/lib/localization/resources.ts
grantfitzsimmons: https://github.com/specify/specify7/pull/4867#issuecomment-2101313908

maxpatiiuk:
Agree
Good points - we are not ready yet to abstract away sp6 XML yet

CarolineDenis (approved):

PR #4873, opened 2024-05-02
Add min width to aggregator preview

### LINKED ISSUE #4858: Table Aggregation dialog resizing when it should not
**Describe the bug**
When you change Limit between 0-4 in Table Aggregation, the dialog changes size according to the value put in--0 starts as the biggest, 1 is the smallest, and then it grows in size between 2-4, where 4 is back to the size of 0.

Additionally, changing any other element causes rapid resizing where it shrinks and then immediately adjusts itself, causing a flashing effect.

**To Reproduce**

1. Go to Schema Config
2. Create a new Table Aggregator
3. Change its limit between 0 and 4
4. See it resize
5. Edit any of the other fields
6. See the rapid resizing

This also happens in App Resources -> Record Formatters -> Table Aggregation.

**Expected behavior**
Shouldn't resize at all.

**Screenshots**
Limit issue:

https://github.com/specify/specify7/assets/71564869/b0563c4c-15e4-43d1-a10b-635acc3e23dd

Rapid resizing:

https://github.com/specify/specify7/assets/71564869/8a5e1a97-d745-40b3-b31e-4d8f5b7b7c25

- OS: Windows 10
- Browser: Firefox
- Specify 7 Version: `edge`
- Collection name: `Herp-PC`
- User Name: `testuser`

**Additional context**
The second issue with the rapid resizing could be dangerous for anyone who's photosensitive, so this should be fixed ASAP. Though I went slowly in the example above, holding down a key/typing quickly makes it significantly worse.

[Specify 7 System Information - 2024-04-29T21 48 42.911Z.txt](https://github.com/specify/specify7/files/15155333/Specify.7.System.Information.-.2024-04-29T21.48.42.911Z.txt)

sharadsw:
Fixes #4858 

### Checklist

### Testing instructions

- Go to Schema Config / App Resources
- Create a new Table Aggregator
  - Change the limit between 0 and 4
    - [ ] Verify the dialog does not rapidly resize 
  - Add a sufficiently large suffix or edit any of the other fields
    - [ ] Verify the dialog does not rapidly resize 

Note: There still is a rapid re-render but it's inside previews rather the entire dialog. To solve that we possibly need to rewrite the underlying component which is also being used by Table Formatter and WebLinks.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Aggregator.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Aggregator.tsx
@@ -114,6 +114,10 @@ function AggregatorPreview({
   );
   return typeof aggregator.table === 'object' &&
     hasTablePermission(aggregator.table.name, 'read') ? (
-    <ResourcePreview doFormatting={doFormatting} table={aggregator.table} />
+    <ResourcePreview
+      doFormatting={doFormatting}
+      isAggregator
+      table={aggregator.table}
+    />
   ) : null;
 }
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Preview.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Preview.tsx
@@ -19,7 +19,10 @@ import { SearchDialog } from '../SearchDialog';
 
 const defaultPreviewSize = 4;
 
-export function useResourcePreview(table: SpecifyTable): {
+export function useResourcePreview(
+  table: SpecifyTable,
+  isAggregator: boolean = false
+): {
   readonly resources: GetOrSet<RA<SpecifyResource<AnySchema>> | undefined>;
   readonly children: (
     callback: (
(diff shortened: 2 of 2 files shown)


emenslin (approved):
### Testing instructions

- Go to Schema Config / App Resources
- Create a new Table Aggregator
  - Change the limit between 0 and 4
    - [x] Verify the dialog does not rapidly resize 
  - Add a sufficiently large suffix or edit any of the other fields
    - [x] Verify the dialog does not rapidly resize 

Looks good, dialog no longer resizes! 

https://github.com/specify/specify7/assets/150940840/cf843d9d-eedd-4847-b7be-9046ae4b806b

CarolineDenis (approved):

maxpatiiuk (approved):
in the video attached to the issue, it looked like the problem is that the preview is expanding the size of the dialog

the best solution I would suggest in such case is to make preview never increase the size of it's parent container, but wrap to multiple lines as needed.

this way, you don't have to hardcoded a cherry-picked width for the dialog size

i.e. see https://stackoverflow.com/questions/20993947/prevent-a-specific-child-div-from-expanding-the-parent-div

for an example implementation, see how the collection name button is implemented in the side menu:

Screenshot 2024-05-02 at 16 44 34

regardless of how long the collection name is, it will never grow the side menu width - but all other menu items (like Workbench and Interactions) do dictate the side menu width

---

I am ok with the solution you have right now, but the above mentioned solution might be better

PR #4776, opened 2024-04-10
Handle case when default formtable requires default view

melton-jason:
Fixes #4749 

> Thankfully there is some consistency amongst the table-only views. Notably, every `altview` in the `altviews` of the view points to a view definition which is of type `formtable`. When reading a form definition for a subview on the frontend, if every altview points to a `formtable` viewdef in a view, we can likely fallback to using the `table.view` view for rendering the expanded definition of the subview (the collapsed view can still remain the original view).

Alternative implementation of #4760 which follows the above proposed solution from https://github.com/specify/specify7/pull/4760#issuecomment-2046022176

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [ ] Add automated tests
- [x] Add relevant issue to release milestone

### Testing instructions
- In the instance, make sure there are no custom views defined for Collectors
- On the view definition for the CollectingEvent, make sure the `collectors` relationship is using the `Collectors` view (ignore this step if using the default): 
```xml
<cell type="subview" viewname="Collectors" id="collect" name="collectors" colspan="13" defaulttype="table" rows="3"/>
```
- Ensure the Collector subview correctly renders the default `Collector` view (pictured below) when the Subview is: 
  - [ ] In Subform Mode
  - [ ] In Grid Mode and the cell is expanded
  - [ ] Also ensure the correct collapsed view is shown when the Subview is in Grid Mode and the cell is collapsed
- On the CollectingEvent form definition, modify the viewname for the `collectors` relationship to be `Collector`
  - [ ] Ensure the collapsed view when the Subview is in Grid Mode is using the `Collector` view definition

- [ ] On the default Agent form, ensure the GroupPerson/Members subview appears as it did in `v7.9+`

---
**Default Collector View form:** 

Screenshot 2024-04-11 at 10 56 06 AM

**Default collapsed Collector View:** 

Screenshot 2024-04-11 at 10 58 28 AM

**Collapsed view definition for `Collector` View:**

Screenshot 2024-04-11 at 11 18 47 AM

PR #4740, opened 2024-04-05
Use correct views for every type of subview

melton-jason:
Fixes #4737 

See https://github.com/specify/specify7/issues/4737#issuecomment-2038639953

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [ ] Add automated tests
- [x] Add relevant issue to release milestone

### Testing instructions
For any relationship which is `to-many` (can be rendered as a SubView in both form and table modes. e.g, on CollectionObject: `preparations`, `determinations`, `dnaSequences`,  etc.)
- [ ] Create a new view definition for the Related Table of the relationship (Preparation, Determination, ...) 
- [ ] Modify the view definition to make it distinct from the default view of the table
- [ ] On the form for the `one` side of the `to-many` relationship, ensure the `viewname` attribute matches that of the newly created view/viewdef and the `defaulttype` attribute is set to `table`
- [ ] On the DataEntry form of the `one` side of the relationship, ensure the correct view definition is used for the subview in all of the following cases:
  - [ ]  The collapsed view of a row in `table` mode
  - [ ] The expanded view of a row in `table` mode
  - [ ] Subform/Form mode

(For a concrete example see https://github.com/specify/specify7/issues/4737#issue-2225602012)

PR #4888, opened 2024-05-05
Use table `defaulttype` for some subviews if not present on cell definition

### LINKED ISSUE #4878: `-to-many` subviews showing in a form by default instead of a grid
**Describe the bug**
For example, "Collection Object Properties" and "Collectors"  are now shown in form view rather than in a grid 

**To Reproduce**

1. Go to any database using the default forms
2. Go to Collecting Event
3. See that the Collectors subview is appearing as a form rather than a grid (compare `v7.9.3.1` to `edge`)

**Expected behavior**
Collectors (and other -to-many rels) should be shown in grid view automatically rather than in form view just as Preparations is.

**Screenshots**
Same form definition:

image

---

image

```xml
		<row>
			<cell type="subview" viewname="Collectors" id="12" name="collectors" colspan="12" rows="3"/>
		</row>
```

See that the `defaulttype=""` attribute is not present in the default forms in Specify 6 despite appearing like a table in the app.

[Paleo Views.xml.zip](https://github.com/specify/specify7/files/15193134/Paleo.Views.xml.zip)

**Crash Report**

[Specify 7 System Information - 2024-05-02T19_37_05.676Z.txt](https://github.com/specify/specify7/files/15193150/Specify.7.System.Information.-.2024-05-02T19_37_05.676Z.txt)

`bauerjen` in the "University of Michigan Invertebrate Paleontology" collection

melton-jason:
Fixes #4878 
Related to the changes introduced in #4740 and #4776

See https://github.com/specify/specify7/issues/4878#issuecomment-2093605099

This solution is the 'messiest' I have considered for this case, but covers the most cases and theoretically a user will never encounter #4878. 

A cleaner, but more prone to users encountering the bug, solution involves implementing a hard-coded list of views which should be rendered as a table (see https://github.com/specify/specify7/commit/5c91a89e633130899af46b3446aaa282ef41920d). 

This solution runs the risk that these default views can be copied (or a similar view/viewdef can be manually created/copied) and the viewname can be changed. Thus, if the cell which renders the view does not specify a `defaulttype`, Specify will always assume that the defaulttype will be `form`. 

@specify/ux-testing, if this is not likely, then I'd vastly prefer the simpler solution until we can define default forms for Specify 7. 

(Also note currently that even if the affected view/viewdefs are copied, they can not be edited via the Visual Editor, although they _can_ be added to a ViewSet when added + saved. 
This behavior can be reproduced by attempting to copy the Collector view titled `Collectors`)

### Checklist

- [x] Add automated tests

### Testing instructions
- On the form for CollectingEvent, ensure there is a cell to render the `collectors` relationship as a SubView
- Use the for the `viewname` attribute, use `Collectors` and make sure the is not a `defaulttype` defined on the cell
- [ ] Open the CollectingEvent form and make sure that Collectors opens as a table
- On the CollectingEvent form, modify the `collectors` cell and add `defaulttype="form"`
- [ ] Open the CollectingEvent form and make sure that Collectors opens as a form

Example `collectors` SubView on a CollectingEvent form with no `defaulttype`:

```xml
<cell type="subview" viewname="Collectors" id="5" name="collectors" colspan="13" rows="3"/>
```

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormCells/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/index.tsx
@@ -154,7 +154,7 @@ const cellRenderers: {
                   typeof viewDefinition === 'object'
                     ? resolveViewDefinition(
                         viewDefinition,
-                        viewDefinition.defaultSubviewFormType ?? formType,
+                        formType,
                         propsToFormMode(isReadOnly, isInSearchDialog)
                       )
                     : undefined
--- a/specifyweb/frontend/js_src/lib/components/FormCells/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormCells/index.tsx
@@ -154,7 +154,7 @@ const cellRenderers: {
                   typeof viewDefinition === 'object'
                     ? resolveViewDefinition(
                         viewDefinition,
-                        formType,
+                        viewDefinition.defaultSubviewFormType ?? formType,
                         propsToFormMode(isReadOnly, isInSearchDialog)
                       )
                     : undefined
--- a/specifyweb/frontend/js_src/lib/components/FormEditor/Create.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormEditor/Create.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { useOutletContext } from 'react-router';
 import { useNavigate } from 'react-router-dom';
 import type { LocalizedString } from 'typesafe-i18n';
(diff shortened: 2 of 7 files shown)


CarolineDenis (approved):

CarolineDenis:
This seems fine for now. I Know we couldn't find anything 'better'. Thank you for this solution. 
I still think that as soon as we have the default file in 7 we should revert those changes and add the defaulttype=formtable to the relevant subviews.

sharadsw (approved):
Good job on the PR!

PR #4890, opened 2024-05-06
Change unavailable prep string and filter results

### LINKED ISSUE #4483: Create Loan displays "No preparations are available for the following records:" when preps are available
**Describe the bug**
When creating a loan by entering a Catalog Number with preparations available, "_No preparations are available for the following records: [catalog number]_" is displayed. The user can still click **Continue** and see that there are preparations available to interact with.

**To Reproduce**
1. Find a collection object with preparations available (https://fwri6123-latest.test.specifysystems.org/specify/view/collectionobject/354192/)
2. Go to Interactions → Loan (https://fwri6123-latest.test.specifysystems.org/specify/overlay/interactions/create/Loan/)
3. Enter in the CO's catalog number
4. See line displayed underneath **Update** button

**Expected behavior**
Should not display line when there are preparations available.

**Evidence**

https://github.com/specify/specify7/assets/118556884/db897467-374e-4d72-83d1-938b8e1cc0c1

OS: macOS Monterey (12.6.5)
Browser: Chrome
Specify 7 Version: `edge`
Database Name: `fwri_6_1_23`

CarolineDenis:
Fixes #4483 

### Checklist

### Testing instructions

- go in interactions
- click on loan 
- choose by cat # 
- enter a cat # that has some prep type unavailable 
- verify that the warning string is now "No preparations are available for at least one type of preparation in the following records:"

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
@@ -293,9 +293,8 @@ export function InteractionDialog({
                     {interactionsText.addUnassociated()}
                   </Button.Info>
                 ) : actionTable.name === 'Loan' &&
-                  !(
-                    state.type === 'MissingState' && prepsData?.length === 0
-                  ) ? (
+                  state.type === 'MissingState' &&
+                  prepsData?.length === 0 ? (
                   <Link.Info href={getResourceViewUrl('Loan')}>
                     {interactionsText.withoutPreparations()}
                   </Link.Info>
--- a/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/InteractionDialog.tsx
@@ -153,6 +153,9 @@ export function InteractionDialog({
     const unavailablePrep = prepsData.filter(
       (prepData) => Number.parseInt(prepData[10]) === 0
     );
+    const availablePrep = prepsData.filter(
+      (prepData) => Number.parseInt(prepData[10]) > 0
+    );
     const unavailable =
       typeof entries === 'object'
         ? entries.filter((entry) =>
@@ -162,8 +165,8 @@ export function InteractionDialog({
 
     if (missing.length > 0 || unavailable.length > 0) {
       setState({ type: 'MissingState', missing, unavailableBis: unavailable });
-      setPrepsData(prepsData);
-    } else showPrepSelectDlg(prepsData);
+      setPrepsData(availablePrep);
+    } else showPrepSelectDlg(availablePrep);
   }
 
   const showPrepSelectDlg = (prepsData: RA<PreparationRow>): void =>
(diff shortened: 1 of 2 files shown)


melton-jason (commented):
This is better! 
But I am still not sure how useful the message is in the first place. 
I would imagine the knowledge that only _some_ preparations were available for a CollectionObject is not very helpful, especially if this is already visible in the next dialog where they can view the preparations. To further clarify, I think if a message like this is shown to the user, it should provide information which should make them consider altering their current workflow. 
The only case where I think this is applicable is when a CollectionObject has no available preparations across all types of preparations.  

@specify/ux-testing @specify/dev-testing, the current implementation _is_ ideal however, if users care more about the _type_ of preparation for a CollectionObject they are using in an Interaction. Is this the case? Or do users care more particularly about the CollectionObject as a whole rather than the types of Preparations?  
If the former is the case, then ignore this suggestion.

Consider the following workflow: 

Screenshot 2024-05-06 at 10 35 48 AM

Screenshot 2024-05-06 at 10 36 00 AM

I would propose the following changes, which I hope would be more intuitive: 

Revert the message to `No preparations are available for the following records:` and only display CollectionObjects which have no available preparations in the message.
Then in the next screen where Preparations are being chosen, omit the Preparations for CollectionObjects which were shown in the previous dialog to have no available Preparations. 
(In the workflow example, the first dialog would state that there are no available Preparations for the CollectionObject with catalognumber `000000002`, and in the second dialog there would be no Preparations for that CollectionObject visible). 

If desired, this could be taken a step further and any Preparation which is entirely unavailable could be omitted in the second dialog.

CarolineDenis:
@melton-jason, 
What I did: 
- if the user enters a cat # that has no prep, a warning will be displayed, the user will be able to click on continue without preparation 
- if the user enters a cat # that has some type prep with no preps available, a warning will be displayed, the user will be able to click on continue, the interaction dialog with the rows will open without the prep type unavailable 
- if the user enters two cat #, one with some available prep types, one with nothing available, a warning will be displayed, the user will be able to click on continue, the interaction dialog with the rows will open without the prep type unavailable and without the second cat # 
- I think keeping this current new string makes more sense

emenslin (changes requested):
### Testing instructions

- go in interactions
- click on loan 
- choose by cat # 
- enter a cat # that has some prep type unavailable 

- [x] verify that the warning string is now "No preparations are available for at least one type of preparation in the following records:"

Issue is that the without preparations button is missing from the dialog 
Screenshot 2024-05-13 105619

---
Although the functionality of this PR works I feel like the initial dialog could be unnecessary because the user can see all the information in the next dialog it could be confusing to why it's there. What would be the effect of removing it?

Screenshot 2024-05-13 112127
Screenshot 2024-05-13 112146

CarolineDenis:
@emenslin , 

Because that dialog let you know that there is nothing available for #42562 and that at least one prep type for #2 as nothing available, those 2 will not be represented in the second dialog as we now filter the unavailable ones

pashiav (approved):
> * _"No preparations are available for at least one type of preparation in the following records:"_ does not appear for a record with some preps unavailable.

There was an issue with the database. Fixed by Caroline! Looks good.

Areyes42 (approved):
### Testing instructions

-   go in interactions
    
-   click on loan
    
-   choose by cat #
    
-   enter a cat # that has some prep type unavailable
    
-  [x]  verify that the warning string is now "No preparations are available for at least one type of preparation in the following records:"
***
Looks good, warning is properly displayed!
Screenshot 2024-05-14 at 1 58 19 PM

emenslin (approved):

PR #4892, opened 2024-05-07
Add missing business rules from Specify 6

### LINKED ISSUE #4866: Make Agent Type field default to Person
In an Agent record, the Agent Type field defaults to Person in Specify 6, but not in Specify 7.

The Agent Type field should default to Person to match Specify 6 behavior.

Current Specify 7 behavior:
Screenshot 2024-05-01 at 2 27 31 PM

Specify 6:
Screenshot 2024-05-01 at 2 26 21 PM

melton-jason:
Fixes #4866

### Checklist

- [ ] Add automated tests

### DIFF
--- a/specifyweb/businessrules/orm_signal_handler.py
+++ b/specifyweb/businessrules/orm_signal_handler.py
@@ -1,9 +1,13 @@
+import logging
+
 from typing import Callable, Literal, Optional, Hashable
 
 from django.db.models import signals
 from django.dispatch import receiver
 
 from specifyweb.specify import models
+
+logger = logging.getLogger(__name__)
 
 # See https://docs.djangoproject.com/en/3.2/ref/signals/#module-django.db.models.signals
 MODEL_SIGNAL = Literal["pre_init", "post_init", "pre_save",
@@ -14,6 +18,10 @@ def orm_signal_handler(signal: MODEL_SIGNAL, model: Optional[str] = None, **kwar
     def _dec(rule):
         receiver_kwargs = kwargs
         if model is not None:
+            if not hasattr(models, model): 
+                logger.warn(f"{model} not a valid model. Skipping rule")
+                return
+
             receiver_kwargs['sender'] = getattr(models, model)
 
             def handler(sender, **kwargs):
--- a/specifyweb/businessrules/rules/collector_rules.py
+++ b/specifyweb/businessrules/rules/collector_rules.py
@@ -1,6 +1,7 @@
 from django.db.models import Max
 from specifyweb.businessrules.orm_signal_handler import orm_signal_handler
 from specifyweb.specify.models import Collector
+from specifyweb.businessrules.exceptions import BusinessRuleException
 
 
 @orm_signal_handler('pre_save', 'Collector')
@@ -12,3 +13,11 @@ def collector_pre_save(collector):
                 collectingevent=collector.collectingevent)
             top = others.aggregate(Max('ordernumber'))['ordernumber__max']
             collector.ordernumber = 0 if top is None else top + 1
+
+    if collector.agent is not None:
+        collector.division = collector.agent.division
+    else:
+        raise BusinessRuleException("collector must be associated with an agent",
+                                    {"table": "Collector",
+                                     "fieldName": "agent",
+                                     "collectorid": collector.id})
--- a/specifyweb/businessrules/rules/determiner_rules.py
+++ b/specifyweb/businessrules/rules/determiner_rules.py
@@ -2,17 +2,20 @@ from django.db.models import Max
 from specifyweb.businessrules.orm_signal_handler import orm_signal_handler
 from specifyweb.specify import models
 
-# This check is provided to support the Specify 6.8.01
-# datamodel (schema version 2.9). When support for that
-# version is dropped it can be removed.
-if hasattr(models, 'Determiner'):
+# Determiners were added in the Specify 6.8.01
+# datamodel (schema version 2.9) update. 
+# orm_signal_handler skips rules if the model does not exist 
+# on the primary datamodel object
 
-    @orm_signal_handler('pre_save', 'Determiner')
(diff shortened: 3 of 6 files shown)


grantfitzsimmons:
@CarolineDenis Should this be added to the timeline?

CarolineDenis:
@melton-jason 

> Should we make a migration to include these rules for instances which already have uniqueness rules created?

==> yes

combs-a:
I feel like on the uniqueness rule migration part, I don't like the idea that it might mess with pre-existing workflows. If it's missed in the patch notes/not relevant to users for awhile, people might be become confused when they finally run into the changed behavior.

I think that skipping creating the rules that would already be invalid/be duplicates will cause less issues. I think the migration is necessary though, so hopefully the ones that don't already have business rules now having them won't break anything 🙏

melton-jason:
Hey everyone! A migration has been added to this PR to create the uniqueness rules! A rule is created _only if_ there are not pre-existing duplicates for the rule. 

The PR description and testing instructions have been updated! 

Please note the first testing instruction (also provided below), which has a step that should be taken _before_ migrations have been run on a database!

> - 💡 Before creating a copy of the database for migrations, ensure that at least one of the uniqueness rules in the above "Uniqueness Rule" section will be invalid
>   - For an example, create two Disposals with the same disposalNumber, or add two AccessionAuthorizations with the same permit to an Accession, etc.

emenslin (changes requested):
- [x] Ensure that the added uniqueness rules (detailed in the above "Uniqueness Rule" section) are created when the rule is enforcable (no duplicates before running the migration)
- [x] Ensure that uniqueness rules which are not enforceable (already have duplicates) are skipped and not created

- Open a new Agent form
- [ ] Ensure agentType correctly defaults to "Person" 
- [ ] Save the new Agent, reload the page, and ensure that the Agent still has a "Person" agentType
- [x] Ensure that the previous agentType is respected


-  On the Determination form (which can be rendered as a Subview on CollectionObject), add a new Determiner
- [x] Ensure that the `Determiner -> isPrimary` is checked
- [x] Ensure that the new Determiner now has `isPrimary` checked and the old Determiner no longer has `isPrimary` checked
- On the CollectingTrip form (which can be rendered as a Subview on the CollectingEvent form), add a new Funding Agent
- [x] Ensure that the `FundingAgent -> isPrimary` is checked
- [x] Ensure that the new FundingAgent now has `isPrimary` checked and the old FundingAgent no longer has `isPrimary` checked

--- 
Everything looks good except agents do not default to person

https://github.com/user-attachments/assets/ee9cd11b-f9d0-471f-aa20-fb623407cf50

emenslin (dismissed):
- [x] Ensure that the added uniqueness rules (detailed in the above "Uniqueness Rule" section) are created when the rule is enforcable (no duplicates before running the migration)
- [x] Ensure that uniqueness rules which are not enforceable (already have duplicates) are skipped and not created

- [x] Ensure agentType correctly defaults to "Person" 
- [x] Save the new Agent, reload the page, and ensure that the Agent still has a "Person" agentType
- [x] Ensure that the previous agentType is respected
- Add a default other than `Person` to the field on the form definition
- [ ] Open a new Agent form and ensure that the field correctly defaults to the value specified on the form definition
- [x] Please also test other PickLists (agentTitle, prepType, the default "Department" picklist, etc.) and ensure that no new bugs have been introduced 🙏 



- [x] Ensure that the `Determiner -> isPrimary` is checked
- [x] Ensure that the new Determiner now has `isPrimary` checked and the old Determiner no longer has `isPrimary` checked
- [x] Ensure that the `FundingAgent -> isPrimary` is checked
- [x] Ensure that the new FundingAgent now has `isPrimary` checked and the old FundingAgent no longer has `isPrimary` checked

--- 

Cannot set a default for the agent type picklist but you can for other picklists.

Issue Branch: 

https://github.com/user-attachments/assets/59cdd451-902e-4a3e-86f7-764c21f95050


Main: 

https://github.com/user-attachments/assets/edb5d517-f3f5-4c8d-bd66-68152bb73518

melton-jason:
@emenslin 
Yes you're right! It looks like the business rule to set the default type is always being executed before the logic which sets the default based on the form definition. This is largely a systematic Issue with Specify currently: there currently is no concept of "precedence"--- whichever gets set first will be the value for the field. When Specify then checks to see if it should set the default value for the other value (on the form definition in this case), it sees the field already has a value and thus doesn't get set. 

In cases like these, I would always assume the default value specified on the form definition should overwrite any values set on resource creation/initialization of business rules. Would you agree? 
I'll get to work on a fix which applies the above assumption, but let me know your thoughts on it!

emenslin:
> In cases like these, I would always assume the default value specified on the form definition should overwrite any values set on resource creation/initialization of business rules. Would you agree?

Yes, I think that the value the user specifies in the form definition should override any of the predefined defaults. I'm thinking about how it works in other areas and in most cases when the user sets something themselves it will always reference that and not the defaults (i.e. if a user makes their own CO form definition then the form will look like that and not the default CO form).

melton-jason:
Hey everyone! 
I think we are going to push the setting of Person as default agentType to the next release. It may take some additional time to properly resolve the Issue, and it may be beneficial to get the other changes this branch provides into `main` and the soon-to-be `v7.11` release. 

There will a PR open to address the setting of conflicting defaults (as described in https://github.com/specify/specify7/pull/4892#issuecomment-2904421685) soon! 

For now, can this be tested again without that functionality? I have removed it from this branch and updated the testing instrunctions.

grantfitzsimmons:
@melton-jason not so sure [`213edbf` (#4892)](https://github.com/specify/specify7/pull/4892/commits/213edbf72be9ed6f7e847c91cf8eb78b32952307) is working anymore, create a new agent on sp7demofish and see: https://sp7demofish.specifycloud.org/specify/view/agent/new/

PR #4901, opened 2024-05-10
Prevent user from saving CO without current determination

### LINKED ISSUE #4819: Collection Objects without any current determination can be saved
**Describe the bug**
When you add and delete a determination, no determination will be current anymore. Present in v7.9.3 as well.

**To Reproduce**

1. Go to Data Entry->Collection Object
2. Add multiple determinations
3. Delete the determination with isCurrent checked
4. Check the other determinations to see that there is no longer any marked as current
5. See that you can save

**Expected behavior**
No behavior was implemented for this previously. Should maybe have a save blocker or something to prevent from saving without a current determination marked? Unsure of the implications of this, but knowing the up-to-date determination seems important.

**Screenshots**
image

- OS: Windows 10
- Browser: Firefox
- Specify 7 Version: `edge`
- Collection name: `Herp-PC`
- User Name: `testuser`

[Specify 7 System Information - 2024-04-22T21 53 37.884Z.txt](https://github.com/specify/specify7/files/15068910/Specify.7.System.Information.-.2024-04-22T21.53.37.884Z.txt)

CarolineDenis:
Fixes #4819

### Checklist

### Testing instructions

- Go to Data Entry->Collection Object
- Add multiple determinations
- Play around with deleting, unchecking and adding a current determination
- [ ] Verify save is blocked when no current determinations exist
- [ ] Verify save is enabled when a current determination is added


specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -207,17 +210,37 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
...
           determination.collection !== undefined &&
           !determination.collection.models.some(
             (c: SpecifyResource<Determination>) => c.get('isCurrent')
           )
         ) {
-          determination.set('isCurrent', true);
+          return {
+            isValid: false,
+            reason: resourcesText.currentDeterminationRequired(),
+            saveBlockerKey: 'determination-isCurrent',
```
maxpatiiuk, line 223: could you create a variable for this key and reuse it?
(to guard against accidental typos)

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
sharadsw: fixed

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -207,17 +212,47 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
...
+          saveBlockerKey: CURRENT_DETERMINATION_KEY,
+          resource: determination.collection?.related,
+        };
       },
     },
+    onRemoved: (determination: SpecifyResource<Determination>): void => {
+      // Block save when a current determination is deleted
+      if (determination.get('isCurrent'))
+        setSaveBlockers(
+          determination.collection?.related ?? determination,
```
melton-jason, line 240: (optional)
In this case, we know that `determination.collection` will always be defined. 
We can directly access the Collection which the resource was added to by using the second parameter of the `onRemoved` function: 
https://github.com/specify/specify7/blob/4e5722aabb4ae3e8afd2ee487f4bea68eaff83d1/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts#L31-L34

We trigger this `onRemoved` function whenever the `removed` Backbone event is triggered: 
https://github.com/specify/specify7/blob/4e5722aabb4ae3e8afd2ee487f4bea68eaff83d1/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts#L59-L60

Backbone handles passing the resource instance and the collection as arguments with the event. 
See the Backbone docs: https://backbonejs.org/#Events-catalog

```suggestion
    onRemoved: (determination, collection): void => {
      // Block save when a current determination is deleted
      if (determination.get('isCurrent'))
        setSaveBlockers(
          collection.related ?? determination,
```

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -207,17 +212,47 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
...
+          determination.specifyTable.field.isCurrent,
+          [resourcesText.currentDeterminationRequired()],
+          CURRENT_DETERMINATION_KEY
+        );
+    },
+    onAdded: (determination: SpecifyResource<Determination>): void => {
+      // Clear any existing save blocker on adding a new current determination
+      if (determination.get('isCurrent'))
+        setSaveBlockers(
+          determination.collection?.related ?? determination,
```
melton-jason, line 250: (optional) 
see the above comment about directly using the collection instead of `determination.collection`

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -207,17 +212,47 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
...
+          determination.specifyTable.field.isCurrent,
+          [],
+          CURRENT_DETERMINATION_KEY
+        );
+    },
```
melton-jason, line 255: (optional) 
If a little refactoring can be done, the code from Determination `customInit` should really be simplified and included here, in the `onAdded` callback. 

https://github.com/specify/specify7/blob/4e5722aabb4ae3e8afd2ee487f4bea68eaff83d1/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts#L149-L165

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -207,17 +212,47 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
...
+            saveBlockerKey: CURRENT_DETERMINATION_KEY,
+            resource: determination.collection.related,
+          };
         }
-        return { isValid: true };
+        return {
+          isValid: true,
+          saveBlockerKey: CURRENT_DETERMINATION_KEY,
+          resource: determination.collection?.related,
+        };
```
melton-jason, line 233: (just a note) 
While this is functional for now, some refactoring in the future may be required. 
Currently, there is not a way to handle if there are multiple blockers which need to be updated with a single field update.
We _could_ utilize `setSaveBlockers` directly, but there may be certain logic in `businessRules` which would be important to execute. And preferably, the `setSaveBlocker` functionality and thinking should be as separate as possible from BusinessRules (i.e., only worry about the BusinessRuleResult and let the internal logic handle setting the saveBlockers). 

Eventually, business rules should be refactored to be able to handle accommodating the case.

melton-jason: An additional refactoring note: the other callbacks for business rules (`customInit`, `onRemoved`, and `onAdded`): 

https://github.com/specify/specify7/blob/4e5722aabb4ae3e8afd2ee487f4bea68eaff83d1/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts#L27-L35

should be refactored to have the same return type as the `fieldChecks` callback for the same reasons as stated in the discussion comment above

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -469,6 +469,7 @@ export function attachBusinessRules(
 export type BusinessRuleResult<SCHEMA extends AnySchema = AnySchema> = {
   readonly localDuplicates?: RA<SpecifyResource<SCHEMA>>;
   readonly saveBlockerKey?: string;
+  readonly resource?: SpecifyResource<SCHEMA>;
```
melton-jason, line 472: (optional)
I understand that this is needed for the case that a change occurs and a save blocker needs to be set on another resource other than the one attached to the businessrule manager running the rule. 

Can you add a comment here explaining that, or rename the variable to make its purpose a little clearer? 
Something like: 
```suggestion
  readonly resourceToSaveBlock?: SpecifyResource<SCHEMA>;
```

melton-jason (approved):
This approach is fine for now given how businessRules and saveBlockers are currently implemented 😅 

Still is a mess with how verbose and complicated it is to add 'simple' behavior such as this.
Thus most of these comments are just related to refactoring and nothing functional.

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
sharadsw: This is very verbose indeed. Also this part is duplicated here and in isCurrent fieldCheck and could be written into a function
```
if (determinaton.collection !== undefined) { 
         determinaton.collection.models.forEach( 
           (other: SpecifyResource<Determination>) => { 
             if (other.cid !== determinaton.cid) 
               other.set('isCurrent', false); 
           } 
         ); 
       } 
```

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
sharadsw: added a comment

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
sharadsw: Looks like we don't need to uncheck other determinations in `customInit` at all. If we set `isCurrent` to true when creating a determination, it triggers the `isCurrent` fieldCheck which then unchecks isCurrent for other determinations in the collection

melton-jason (approved):
Much simpler, thanks for refactoring this!

Areyes42 (approved):
### Testing instructions

-   Go to Data Entry->Collection Object
-   Add multiple determinations
-   Play around with deleting, unchecking and adding a current determination
-   [x] Verify save is blocked when no current determinations exist
-   [x] Verify save is enabled when a current determination is added
***
Looks good!
Screenshot 2024-06-03 at 11 29 03 AM

combs-a (approved):
- Go to Data Entry->Collection Object
- Add multiple determinations
- Play around with deleting, unchecking and adding a current determination
- [X] Verify save is blocked when no current determinations exist
- [X] Verify save is enabled when a current determination is added

Looks good! Worked with unchecking and deleting, as well as newly added determinations. No unexpected behavior during my checks.

PR #4805, opened 2024-04-18
Open interaction dialog when adding prep in ExchangeOut

CarolineDenis:
Fixes #922

### Checklist

- [ ] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [x] Add relevant issue to release milestone

### Testing instructions

- Go to interactions 
- Open Exchange Out 
- Add a prep
- [ ] verify the interaction dialog opens
- [ ] verify that you can add a prep by selecting "By choosing a record set" 
- [ ] verify that you can add a prep by selecting "By entering catalog number" 
- [ ] verify that you can add an unassociated item 
- [ ] test adding a preparation for ExchangeIn

PR #4905, opened 2024-05-10
Prevent from carrying preps in interaction during cloning

### LINKED ISSUE #4012: Cloning a Loan carries the Loan Preparations along
image

**Steps to Recreate:**
1. Create a loan with the maximum number of preps an object has
2. Clone the loan
3. Save

```
BusinessRuleException at /api/specify/loan/ ('loan preparation quantity exceeds availability (None: 2 0)', 
```

Should cloning a loan record carry forward the loan preparations? I don't think so.

[Specify 7 Crash Report - 2023-09-14T20 39 29.482Z.txt](https://github.com/specify/specify7/files/12613199/Specify.7.Crash.Report.-.2023-09-14T20.39.29.482Z.txt)

CarolineDenis:
Fixes #4012

### Checklist

### Testing instructions

- Create a loan with the maximum number of preps an object has
- Clone the loan
- [ ] verify that the preparations were not carried over to the new 'cloned' loan 
- [ ] verify the same for other interactions


specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
```diff
@@ -225,12 +225,37 @@ export function resourceOn(
...
+] as const;
+type InteractionTable = typeof interactionTablesWithPreps[number];
+const interactionTablesPrepsFieldName: RR<InteractionTable, string> = {
+  Disposal: 'disposalPreparations',
+  Loan: 'loanPreparations',
+  Gift: 'giftPreparations',
+  ExchangeOut: 'exchangeOutPreps',
+  ExchangeIn: 'exchangeInPreps',
+  Borrow: 'borrowPreparations',
+};
```
maxpatiiuk, line 244: do you need to hardcode this list?
imagine someone adding a new table - how would they know that this place in the code might need to be modified before the table is added or else a bug is introduced???
that is why you should avoid hardcoding things and creating edge cases where possible - instead compute things on the fly

could you please dynamically construct it on the fly (i.e if field name is a relationship to the Preparations table, then don't clone it)

---

the only exception to computing things on the fly is when doing so is computationally expensive - but in those cases, you should add a test case that will do the computation to make sure the hardcoded list is not out of date with the dynamically computed list
(i.e front-end stores a hardcoed list of all the permission policies that sp7 back-end defines so as not to have to fetch those from the back-end on each app load - yet, when in development, it will make sure that the hardcoded list stored on the front-end matches exactly what back-end provides)

specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
```diff
@@ -225,12 +225,37 @@ export function resourceOn(
...
 export function getFieldsToNotClone(
   table: SpecifyTable,
   cloneAll: boolean
 ): RA<string> {
-  const fieldsToClone = getCarryOverPreference(table, cloneAll);
+  let fieldsToClone = getCarryOverPreference(table, cloneAll);
   const uniqueFields = getUniqueFields(table);
+  if (
+    interactionTablesWithPreps.includes(table.name as InteractionTable) &&
+    cloneAll
```
maxpatiiuk, line 253: are you sure this logic should apply only in the `cloneAll` cases?
it seems like it should apply in `!cloneAll` case too...

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
melton-jason: The type `InteractionTable` already exists on the frontend (called `AnyInteractionPreparation`), and is constructed dynamically based on the `Tables` type. 

https://github.com/specify/specify7/blob/3fb636eaaec678d7ef4d2e175ac62f105b226cdb/specifyweb/frontend/js_src/lib/components/DataModel/helperTypes.ts#L58-L69

---

For defining a constant with this type, take a look at my changes in #4805. 
Specifically, `interactionPrepTables`: 

https://github.com/specify/specify7/blob/5638811529378ccbf1544e91d9b78c0ead81614c/specifyweb/frontend/js_src/lib/components/Interactions/helpers.ts#L7-L15

If needed/desired, there is also code to get the main Interaction tables which have preparations: 

https://github.com/specify/specify7/blob/5638811529378ccbf1544e91d9b78c0ead81614c/specifyweb/frontend/js_src/lib/components/Interactions/helpers.ts#L20-L26

--- 

#4805 also has examples of using/getting the relationship from Interaction -> InteractionPreparation: 
https://github.com/specify/specify7/blob/5638811529378ccbf1544e91d9b78c0ead81614c/specifyweb/frontend/js_src/lib/components/Interactions/PrepDialog.tsx#L234-L241

CarolineDenis:
NOTES: 
Continue work here once #4805 is merged

specifyweb/frontend/js_src/lib/components/DataModel/resource.ts
```diff
@@ -225,12 +226,25 @@ export function resourceOn(
...
   className.split('.').at(-1) ?? '';
 
+type InteractionTable = typeof interactionsWithPrepTables[number];
+const interactionTablesPrepsFieldName: RR<InteractionTable, string> = {
+  Disposal: 'disposalPreparations',
+  Loan: 'loanPreparations',
+  Gift: 'giftPreparations',
+  ExchangeOut: 'exchangeOutPreps',
+  ExchangeIn: 'exchangeInPreps',
+};
```
maxpatiiuk, line 236: could you please create this list dynamically?
like so:
```suggestion
const interactionTablesPrepsFieldName = f.store(()=>
  filterArray(
    Object.values(genericTables)
    .map((table=>{
      const relationshipToPreparation = table.relationships.find(({relatedTable})=>relatedTable === table.Preparation);
      return relationshipToPreparation ? [table.name, relationshipToPreparation.name] : undefined;
    })
  )
);
```

melton-jason (commented):
@maxpatiiuk @CarolineDenis review my latest commit (https://github.com/specify/specify7/commit/30d20735fd42a7dea2cde3746a83a8e8dc0a8e8b) which should helpfully simplify things while remaining (mostly) dynamic. 

(The types for the `interactionPrepTables` and `interactionsWithPrepTables` constants are constructed dynamically and will cause Type Check to fail if there are any missing constants. 
See https://github.com/specify/specify7/pull/4805/files#diff-b4381f08a5a89522f0cf59b58b611e39536e912fc3b3618176b168417db4fd1c)

My only concern with this approach is that this may lead to unintended behavior with `autoMerge` when the tables are able to be merged. 

https://github.com/specify/specify7/blob/c62b400cec7732b856e25451df042bb8898a0d51/specifyweb/frontend/js_src/lib/components/Merging/autoMerge.ts#L161-L187

We can easily reuse the logic in `getFieldsNotToClone` and instead do some slight adjustments to which fields are disabled for carryForward if this will be a problem. 

Currently the disabled fields in the FormMeta CarryForward Config are those that are unique: 

https://github.com/specify/specify7/blob/c62b400cec7732b856e25451df042bb8898a0d51/specifyweb/frontend/js_src/lib/components/FormMeta/CarryForward.tsx#L166

maxpatiiuk (approved):

emenslin (approved):
### Testing instructions

- Create a loan with the maximum number of preps an object has
- Clone the loan
- [x] verify that the preparations were not carried over to the new 'cloned' loan 
- [x] verify the same for other interactions 

Tried all interactions with preparations, looks good! 


https://github.com/specify/specify7/assets/150940840/9127f4c2-a1a4-458e-a9fe-387295c9c885

alesan99 (approved):
- Create a loan with the maximum number of preps an object has
- Clone the loan
- [x] verify that the preparations were not carried over to the new 'cloned' loan 
- [x] verify the same for other interactions 

Preps aren't being carried over 👍👍 Also applies to various quantities
Tested with Loans, Disposals, Gifts, Exchange In/Out

https://github.com/specify/specify7/assets/8108710/2b7f585c-d0c1-4341-b9c2-287f50b9ae2a

combs-a (approved):
### Testing instructions

- Create a loan with the maximum number of preps an object has
- Clone the loan
- [X] verify that the preparations were not carried over to the new 'cloned' loan 
- [X] verify the same for other interactions 

https://github.com/specify/specify7/assets/71564869/78e0c7b4-38f6-4ae9-af07-5ea9fa564e5c

Looks good, worked for all relevant interactions

Areyes42 (approved):
### Testing instructions

-   Create a loan with the maximum number of preps an object has
-   Clone the loan
-   [x] verify that the preparations were not carried over to the new 'cloned' loan
-  [x]  verify the same for other interactions
***
Looks good!

https://github.com/specify/specify7/assets/130881183/c7791b26-dad6-4cff-8154-76883fae04a5

specifysoftware:

Issue #3750, opened 2023-07-08
Enable uploading Collection Relationships in the WorkBench when Collection Object is the base table


grantfitzsimmons:
> The update (`v7.8.12`) introduced the ability to upload to the Collection Relationship table directly as the base table rather than from the Collection Object level. 

Willem at SAIAB is requesting the ability to upload and create collection relationships when the Collection Object table is the base table by simply uploading through the right side rels matching to collection objects on the other side of the relationship by `catalogNumber`.

See the mapping that he had created before realizing this is not currently possible:
![image](https://github.com/specify/specify7/assets/37256050/712a1570-fa1c-4127-a043-1bd04e50799e)

The mapping specifies the relationship name and the catalog number, which theoretically should be all that you need.

I spoke to @melton-jason about this and he believes there _may_ be a way to do this currently, but most likely with a less intuitive workflow. The mapping he created should work out of the box.

**Requested By:** Willem at SAIAB

> It will be wonderful if you can do that because the workflow will obviously be much less complicated.

zoewarner1:
Hi @grantfitzsimmons  - just wondering if Jason was able to provide a way to do this.
I'm re-jigging some WB templates for our Fish collection. Specimens and Tissues are in separate collections. Most tissues are related to a Specimen, which is known at time of uploading the Tissue. 
Currently load the Tissue csv twice, one to create the specimens, and then again to create the relationship to Specimen. I tried the above shown in screenshot, to reduce it to one upload. It ended up creating two Collection Objects in Tissue collection, one with the Specimen catalogue number, the other with the Tissue cat. no. and linked them together.

realVinayak:
Actually, I tested it out, and if my understanding of collection relationships is correct, then this bug is resolved in https://github.com/specify/specify7/pull/4929 (I heavily refactored and reimplemented collection relationships in workbench, and this was a side-effect). That is, the mapping worked out of the box for me.

Here is the mapping I used, and was able to successfully upload into different collection.

![image](https://github.com/user-attachments/assets/0009a21b-04d5-485e-ba68-ad1a232e8815)

![image](https://github.com/user-attachments/assets/d67c30c1-2736-4a6c-90d4-d0ebe3c2bb82)

![image](https://github.com/user-attachments/assets/8121e4cc-2ba7-4d4e-973d-91424af3e853)


</br>
</br>
@zoewarner1 your best bet would be to wait that release out. I'll need to add unit tests for this.

zoewarner1:
[...]

Issue #2565, opened 2022-12-08
Loan preparations cannot be associated with existing loans


grantfitzsimmons:
> I created an empty loan and then tried to add the preparations via the loanpreparation table in workbench

![image](https://user-images.githubusercontent.com/37256050/206487498-fb947e2b-cb0e-439c-abe3-df4749771a40.png)

---

A more in-depth look at another instance of this issue:

> I've added the spreadsheet I used for the workbench upload with a screenshot of the mapping. The collection objects all have one preparation but the "Dan-test124" loan already exists. We'll have hundreds of preps to add to a loan at a time (even though this example just has the one row) - can you spot any differences in the setup from your testing?

![image](https://user-images.githubusercontent.com/37256050/206487561-4b6012c1-d62f-47a3-a9b3-b86490ac2d4c.png)

![image](https://user-images.githubusercontent.com/37256050/206487587-8259236f-0d35-48ee-9887-f003b7d48194.png)

![image](https://user-images.githubusercontent.com/37256050/206487697-0c8995d2-ee4d-4d54-9628-022de0a159df.png)

[testing loanprep workbench - Copy.xlsx](https://github.com/specify/specify7/files/10186869/testing.loanprep.workbench.-.Copy.xlsx)

grantfitzsimmons:
It matches an existing loan exactly– no trailing whitespace

grantfitzsimmons:
https://sp7demofish.specifycloud.org/specify/workbench/3/

Recreated it just now- see for yourself!

Existing loan: https://sp7demofish.specifycloud.org/specify/view/loan/251/

maxxxxxdlp:
Looks like it wants to create new records:

![Screenshot 2022-12-08 at 12 49 37](https://user-images.githubusercontent.com/40512816/206541597-a715af60-1784-4253-a346-cff58adfb074.png)

Must match:
![Screenshot 2022-12-08 at 12 58 24](https://user-images.githubusercontent.com/40512816/206543491-efc8393f-0d20-47b8-860f-6c717533e94b.png)

maxxxxxdlp:
Need to debug why back-end wants to create new loans rather than match existing

grantfitzsimmons:
From @chanulee1:

> Can add loan preparations to an existing loan but going to preparations show that they are not loaned. Happens when adding with both catalog numbers and records sets. Also, adding preparations does not trigger the save button. (Edit: Appears to be duplicate of #2565)
> 
[...]

PR #4929, opened 2024-05-18
Batch Edit, Workbench, Query Builder backend improvements

realVinayak:
Some backend improvements

Workbench
- Refactored collection relationships. code was tricky before.
- Arbitrary nested to-many support
- Makes multiple matches for to-many more smarter
- Batch Edit
- Apparently also https://github.com/specify/specify7/issues/3750

Query Builder
- Generic relationship and fields in trees. At least the backend support for now.

### Checklist

- [ ] Add automated tests

### Testing instructions

# Batch-editing

## Implementation and design

1. The current implementation uses workbench, and query builder. 
2. The workbench and batch-edit dataset are differntiated at the user level using a new "isupdate" field in spdataset table. DEV note: There is no difference at the code level -- everything is as general as possible. In fact, the isupdate field is only used at the code level to follow a special rollback procedure
3. The batch-edit datasets can seen using the batch-edit overlay accessible via the batch-edit menu item. Currently, the only way to create a new dataset is via the query builder interface.
4. To make a batch-edit dataset, go to the query builder, and add the relevant fields to the query. Some fields and relationships are not supported. Nested-to-many, for instance, are supported in workbench, but aren't in the batch-edit. Special fields like nodenumber and highestchildnodenumber, fullname field in tree are also not supported.
5. Other than relationships mentioned above, every field is supported. If an unsupported field like nodenumber is added, it is rendered as readonly. However, you cannot make nested-to-manys visible (this is different than being able to map it) --- you can map nested-to-manys, and even filter on them. As long as they are hidden, it'll not block. You can also arbitrarily add formatted and aggregated fields (they are unmapped, and are ignored)

## Testing Instructions

- Make a query with columns in the base table, and select relationships to edit. In the current, I am using CO with the base table. There are 4 different types of relationships, in general. 
    - To-one dependent (for ex. collectionobjectattribute),
    - To-one independent (for ex. Cataloger, CollectingEvent [when not embedded])
    - To-many dependent (for ex. determinations, preparations),
    - To-many independent (for ex. None for CO as the base table)

- ### Fields
    Modifying any field is possible, other than nodenumber, fullname, and highestchildnodenumber. If some fields get updated, only those fields are highlighted
- ### To-one dependent (for ex. collectionobjectattribute)
    These relationships get directly updated, and are not matched. If the to-one is not in the db, it'll create one. 
    This also includes collectingevent when embedded. <br/>
    <u>Test cases to consider</u>:
    - When mapped, the record is directly updated.
    - When mapped, if the record is not present, it'll be created, if not null values are present.
    - If the record previously had values, and the values are removed (making the cells completely empty), the to-one dependent record will be deleted. Since it is possible that there may be other fields in the database (but not in the query), we may accidentally delete the record. Eample: user selected collectionobject -> collectionObjectAttribute -> remarks. And say they set remarks to empty in the spreadsheet, it is is possible that integer1 field in collectionObjectAttribute may have some value. To prevent accidental deletion, by default, we look at all the fields in the database for that record (other than system fields), to determine whether we can delete the record or not. This behaviour is controlled by a remote preference. (described in a later section):

- ### To-many dependent (for ex. determinations)

    Same as to-one dependent. These relationships get directly updated. If the corresponding record is not present, a new one gets created. 

    <u>Test cases to consider</u>:
    - When mapped, the record is directly updated.
    - When mapped, if the record is not present, it'll be created, if not null values are present.
    - If the cell data is removed, and if every other field is empty in the database (can be disabled via a preference), the record will be deleted.

- ### To-one independent (for ex. cataloger)

    These relationships get matched, and uploaded (if match is not found). During upload, it performs a clone of the record (cloning all the non-unique fields, and dependents). The clone takes into account relationships also mapped. That is, if agent needs to be cloned, and you have mapped agentspecialty, it'll take the agentspecialty mapped (rather than cloning previous's agentspecialty).

    <u>Test cases to consider:</u>

    - Start from a collectionobject with a cataloger, and map some fields. Change some of the values (like, say, lastname and firstname) to of agents that are present in the db. verify that agent gets matched. Note that the match can be performed with just the visible fields, or can also include fields in the database, not included in the query. This is controlled via a preference. By default, to be cautious in matching, it uses just the fields visible in the query.
    - If it is unable to match, it'll clone the existing agent, with data from the sheet. Make an agent with addresses/specialties/variants. Make sure the workbench is able to clone the agent correctly, and if you've provided some dependents in the mapping, it takes it.
    - Similar to workbench, you could customize the match behaviour by changing the matching options (like "never ignore", "ignore when blank", and "always ignore")

- ### To-many independent

    Same as to-many dependent. The only difference is that we always perform an update (we don't delete these). If a mapped record is not present, it'll create one, without any matching. 

    <u>Test cases to consider: </u>
    - Make collection objects, and assign them collectingevent. Do a query using collectingevent as the base table, and add fields of the CO table. Verify that resetting all fields does not delete the collection object (you'll also need to disable the preference that says to look at all fields for null checks)
    - If a record is not present, it'll create one, if there is a non-empty field.    

- ### Trees

    - Workbench method:<br/>

    - Update method: <br/>

    In both of the above methods, fullname, nodenumber, highestchildnodenumber is completely readonly. 

## Results

There are 4 new different type of results; 
- ### NoChange
Reported when the record was meant to updated, but no change occurred. That is, all the values from the db were the same. This is not visible to the user.

- ### Updated
Reported when the record's fields were changed. This does not consider relationships (they are reported with different result)

- ### Deleted
Reported when a record is deleted. Happens when a dependent's cells are all empty.

- ### MatchedAndChanged
Reported when a to-one independent was matched to another record, different than the current one.

- The side panel also shows the results per table, for different categories.

## Preferences

There are three different preference options.

- ### Remote Preferenences (2)

- <b> Defer For Match </b>
Set by `sp7.batchEdit.deferForMatch`.This preference controls whether database fields are included for matching or not. Defaults to false.

- <b> Defer For Null </b>
Set by `sp7.batchEdit.deferForNull`.This preference controls whether database fields are included for determining if the record is null or not. For dependents, null records are deleted, so this preference is used to control the caution batch-edit follows

- ### User Preferenences (1)

- <b> Number of query rows </b>

Determines how many number of query results are used for batch-edit. Defaults to 5000.

## Rollbacks

Rollbacks are complicated to perform. In the current design, whenever user creates a batch-edit dataset, via the query builder, it makes two datasets. User can only see one of them. The second is a "backer" of the first, and contains a FK to the first (so we can find backer of a dataset later). When rollback is requested, for every row in the main one, we find the original row in the backer, and perform the regular batch-edit update on it. Essentially, it applies original snapshot.

This is highly experimental, so it is recommended to always take a backup of the db, but this should work in a good amount of cases.

## Misc

- Queries from record set are supported.


grantfitzsimmons:
@CarolineDenis	Is this something that we can include in 7.9.7?

grantfitzsimmons:
@realVinayak Please mark this as ready for testing once you're ready and send it to the #sp7-engineering channel! Thank yoU!

specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
```diff
@@ -1936,6 +1937,40 @@ export const userPreferenceDefinitions = {
       },
     },
   },
+  batchEdit: {
+    title: batchEditText.batchEdit(),
+    subCategories: {
+      query: {
```
maxpatiiuk, line 1943: (optional) if this limit is so easily changeable by the user, instead of being a limit, should we just display a warning when someone tries to batch edit more records than this?

(optional) and then we might not even need a pref for this as at that point we could just have a `const batchEditWarningSize = 5_000;`.

specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
```diff
@@ -1936,6 +1937,40 @@ export const userPreferenceDefinitions = {
...
+            visible: true,
+            defaultValue: 5000,
+            type: "java.lang.Double",
+            parser: {
+              min: 0
+            }
+          })
+        }
+      },
+      editor: {
```
maxpatiiuk, line 1958: could you provide context on why there is a pref to disable this?

specifyweb/frontend/js_src/lib/components/Toolbar/WbsDialog.tsx
```diff
@@ -129,159 +133,246 @@ function TableHeader({
   );
 }
 
-/** Render a dialog for choosing a data set */
-export function DataSetsDialog({
+
+type WB_VARIANT = keyof Omit<typeof datasetVariants, "bulkAttachment">;
```
maxpatiiuk, line 137:
```suggestion
type WbVariant = keyof Omit<typeof datasetVariants, "bulkAttachment">;
```

to be consistent with our naming

specifyweb/frontend/js_src/lib/components/Toolbar/WbsDialog.tsx
```diff
@@ -129,159 +133,246 @@ function TableHeader({
...
+  const {fetchUrl, sortConfig: sortConfigSpec, canEdit, uiSpec, route, metaRoute, canImport} = datasetVariants[wbVariant];
   const [unsortedDatasets] = useAsyncState(
     React.useCallback(
-      async () =>
-        ajax<RA<DatasetBriefPlan>>(
-          `/api/workbench/dataset/${showTemplates ? '?with_plan' : ''}`,
-          { headers: { Accept: 'application/json' } }
-        ).then(({ data }) => data),
-      [showTemplates]
+      async () => ajax<RA<DatasetBriefPlan>>(formatUrl(fetchUrl, {}), { headers: { Accept: 'application/json' } }).then(({data})=>data),
```
maxpatiiuk, line 153: looks like formatUrl is not needed here since you don't pass any options?

specifyweb/frontend/js_src/lib/components/Toolbar/WbsDialog.tsx
```diff
@@ -129,159 +133,246 @@ function TableHeader({
...
+                      ? (event): void => {
+                          event.preventDefault();
+                          handleDataSetSelect(dataset.id);
+                        }
+                      : undefined
+                  }
+                >
+                  <TableIcon
+                    label
+                    name={dataset.uploadplan?.baseTableName ?? 'Workbench'}
```
maxpatiiuk, line 229: instead of hardcoding `Workbench` here, could you use something like `tables.Workbench.label` instead?

specifyweb/frontend/js_src/lib/components/Toolbar/WbsDialog.tsx
```diff
@@ -129,159 +133,246 @@ function TableHeader({
...
+      undoStartDescription:  wbText.beginRollbackDescription(),
+      doSuccessfulDescription: wbText.uploadSuccessfulDescription(),
+      undoFinishedDescription: wbText.dataSetRollbackDescription(),
+      doing: wbText.uploading()
+    }
+  }
+} as const;
+
+// Defines a shared interface to access dataset variants
+export const datasetVariants = {
```
maxpatiiuk, line 303: does this belong in a separate file of its own? seems important enough

specifyweb/frontend/js_src/lib/components/Toolbar/WbsDialog.tsx
```diff
@@ -129,159 +133,246 @@ function TableHeader({
...
+  canEdit: ()=>hasPermission('/workbench/dataset', 'update'),
+  route: (id: number)=>`/specify/workbench/${id}`,
+  metaRoute: (id: number)=>`/specify/overlay/workbench/${id}/meta/`,
+  canCreate: ()=>hasPermission('/workbench/dataset', 'create'),
+  canTransfer: ()=>hasPermission('/workbench/dataset', 'transfer'),
+  canUpdate: ()=>hasPermission('/workbench/dataset', 'update'),
+  canDo: ()=>hasPermission('/workbench/dataset', 'upload'),
+  canUndo: ()=>hasPermission('/workbench/dataset', 'unupload'),
+  canValidate: ()=>hasPermission('/workbench/dataset', 'validate'),
+  uiSpec: {
```
maxpatiiuk, line 284: rename to `localization` or `text`?

specifyweb/frontend/js_src/lib/components/Toolbar/WbsDialog.tsx
```diff
@@ -129,159 +133,246 @@ function TableHeader({
...
+const baseWbVariant = {
+  fetchUrl: '/api/workbench/dataset/',
+  sortConfig: {
+    key: 'listOfDataSets',
+    field: 'name'
+  },
+  canImport: ()=>hasPermission('/workbench/dataset', 'create'),
+  canEdit: ()=>hasPermission('/workbench/dataset', 'update'),
+  route: (id: number)=>`/specify/workbench/${id}`,
+  metaRoute: (id: number)=>`/specify/overlay/workbench/${id}/meta/`,
```
maxpatiiuk, line 277: rename to getRoute/getMetaRoute?

function names should usually begin with verbs, not nouns

specifyweb/frontend/js_src/lib/components/WbImport/helpers.ts
```diff
@@ -160,9 +161,9 @@ const MAX_NAME_LENGTH = 64;
 export async function uniquifyDataSetName(
   name: string,
   currentDataSetId?: number,
-  datasetsUrl = '/api/workbench/dataset/'
+  datasetsUrl: keyof typeof datasetVariants = 'workbench'
```
maxpatiiuk, line 164: you have WB_VARIANT (WbVariant) type in a different file. should you export it and use it here?

specifyweb/frontend/js_src/lib/components/WbPlanView/__tests__/navigator.test.ts
```diff
@@ -20,396 +20,405 @@ theories(getMappingLineData, [
...
-          voucherRelationships: {
-            optionLabel: 'Voucher Relationships',
-            isEnabled: true,
-            isRequired: false,
-            isHidden: false,
-            isDefault: false,
-            isRelationship: true,
-            tableName: 'VoucherRelationship',
-          },
+        "customSelectSubtype": "simple",
```
maxpatiiuk, line 23: hard to see the diff here, especially since a bunch of quotes were added (that's why we have Prettier)
could you please summarize what changed here?

specifyweb/frontend/js_src/lib/components/WbUtils/index.tsx
```diff
@@ -127,6 +129,28 @@ export function WbUtilsComponent({
...
+        utils={utils}
+      />
+      <Navigation
+        label={wbText.deletedCells()}
+        name="deletedCells"
+        totalCount={cellCounts.deletedCells}
+        utils={utils}
+      />
+      <Navigation
+        label={wbText.matchAndChanged()}
```
maxpatiiuk, line 147: it's not nice that `And` is capitalized in this button label

you can add `normal-case` class name to these buttons to prevent it from auto-capitalizing every word, thus giving you control over casing

specifyweb/frontend/js_src/lib/components/WbUtils/index.tsx
```diff
@@ -127,6 +129,28 @@ export function WbUtilsComponent({
         totalCount={cellCounts.newCells}
         utils={utils}
       />
+      {
```
maxpatiiuk, line 132: what is the difference between "Matched and Changed" vs "Updated"?

specifyweb/frontend/js_src/lib/components/WbUtils/index.tsx
```diff
@@ -14,13 +14,15 @@ import type { WbUtils } from './Utils';
 
 export function WbUtilsComponent({
```
maxpatiiuk, line 15: some raw ideas for how to reduce the visual size of this component:
- have a small button icon to collapse the "search", "replace", "search settings" and "search results" elements
- replace the navigation button labels with icons, with title being shown in a tooltip on hover

specifyweb/frontend/js_src/lib/components/WorkBench/CellMeta.ts
```diff
@@ -347,6 +368,15 @@ export class WbCellMeta {
...
       ),
+      updatedCells: cellMeta.reduce(
+        (count, info)=>count + (this.getCellMetaFromArray(info, 'isUpdated') ? 1 : 0), 0
+      ),
+      deletedCells: cellMeta.reduce(
+        (count, info)=>count + (this.getCellMetaFromArray(info, 'isDeleted') ? 1 : 0), 0
+      ),
+      matchedAndChangedCells: cellMeta.reduce(
+        (count, info)=>count + (this.getCellMetaFromArray(info, 'isMatchedAndChanged') ? 1 : 0), 0
+      )
```
maxpatiiuk, line 379: (optional) a small performance optimization - could you short circuit the computation of all of these if you are not in an batch editor?

also, we could make this a bit less "functional" but changing code like this:
```ts
let modifiedCells = 0;
let ... = 0;
...

for(const meta of cellMeta) {
  if(this.getCellMetaFromArray(info, 'isDeleted')) modifiedCells += 1;
  /// ...
  if(isInBatchEditor) {
    if(...) updatedCells += 1;
  }
}
``

File: specifyweb/frontend/js_src/lib/localization/batchEdit.ts [hunk 15]
```diff
@@ -0,0 +1,59 @@
...
+  },
+  startRevertDescription: {
+    'en-us':
+      'Rolling back the dataset will re-update the values, delete created records, and create new records',
+  },
+  commitSuccessfulDescription: {
+    'en-us': `Click on the "Results" button to see the number of records affected in each database table`,
+  },
+  dateSetRevertDescription: {
+    'en-us': `This Rolledback Data Set is saved, however, it cannot be edit. Please re-run the query`,
```
maxpatiiuk, line 54: I think it's `rolled back`
https://ell.stackexchange.com/questions/106534/rollbacked-or-rolled-back-the-edit-and-what-about-double-click
also, probably doesn't need to be capitalized

specifyweb/frontend/js_src/lib/localization/workbench.ts
```diff
@@ -383,12 +383,7 @@ export const wbText = createDictionary({
     `,
   },
   startUpload: {
-    'en-us': 'Begin Data Set Upload?',
-    'ru-ru': 'Начать загрузку набора данных?',
-    'es-es': '¿Comenzar carga de conjunto de datos?',
-    'fr-fr': "Commencer le téléchargement de l'ensemble de données ?",
-    'uk-ua': 'Почати завантаження набору даних?',
-    'de-ch': 'Hochladen des Datensatzes beginnen?',
+    'en-us': 'Begin Data Set {type:string}?',
```
maxpatiiuk, line 386: while I like the dry-ness of this solution, I am beginning to realize that being too smart with reusing localization strings is not great:
- it makes autogenerated Google Translate translations worse
- it doesn't give translators context about what the "type" should be, thus leaving them unable to translate this correctly
- some languages might not be flexible enough to accommodate this pattern

is "type" here just one of 2 possible values? if so I would suggest we have separate strings for these

maxpatiiuk (commented):
I really like that you added test cases for code you touched, even if you didn't write it.
I really like that you wrote code in a very readable way, out of small utilities, and refactored existing code to split it into smaller pieces - made the code much easier to read.

Though, you need to run Prettier!

lexiclevenger (changes requested):
>  If an unsupported field like nodenumber is added, it is rendered as readonly.

Adding an unsupported field and clicking Batch Edit throws an error. I tried to add fullname to this query:
Screenshot 2024-08-23 at 2 59 09 PM
[Specify 7 Crash Report - 2024-08-23T19_59_00.369Z.txt](https://github.com/user-attachments/files/16733150/Specify.7.Crash.Report.-.2024-08-23T19_59_00.369Z.txt)

lexiclevenger:
> > If an unsupported field like nodenumber is added, it is rendered as readonly.
> 
> Adding an unsupported field and clicking Batch Edit throws an error. I tried to add fullname to this query: Screenshot 2024-08-23 at 2 59 09 PM [Specify 7 Crash Report - 2024-08-23T19_59_00.369Z.txt](https://github.com/user-attachments/files/16733150/Specify.7.Crash.Report.-.2024-08-23T19_59_00.369Z.txt)

I realized this occurs when trying to add any Taxon field without adding "name", not just the unsupported fields.

lexiclevenger (changes requested):
The following validation error occurs when I try to validate changes in this data set despite no changes actually being made to the tree structure:
`Error occurred during Validation
{
  "uploaderstatus": {
    "operation": "validating",
    "taskid": "a1f94f22-356e-4827-ac2f-1e6b6814f4b2"
  },
  "taskstatus": "FAILURE",
  "taskinfo": "BusinessRuleException('Tree structure changed, please re-run the query', {}, ['Taxon', ['Subspecies'], None])"
}`

Screenshot 2024-08-23 at 3 46 27 PM

Screenshot 2024-08-23 at 3 34 04 PM

https://fwri1924-wbimprovements.test.specifysystems.org/specify/workbench/449

specifyweb/frontend/js_src/lib/localization/workbench.ts
realVinayak: Agreed, it was just too convenient to reuse them.

specifyweb/frontend/js_src/lib/components/WbUtils/index.tsx
realVinayak: Updated means one of the mapped columns changed. So, collectionobject's remarks field changed -- that'll be updated.

Matched and Changed means if you mapped cataloger, and it matched it to an agent different than the current one. In this case, the collectionobject will have NoChange (completely of no color), but the agent will be of different color (because matched and changed). This was done because otherwise you'd be highlighting the entire collectionobject, which introduces ambiguity: did you actually change all the fields, or just the relationship?

maxpatiiuk: ah, makes sense.
vanilla workbench has this ambiguity when you add determinations to existing collection object

specifyweb/frontend/js_src/lib/localization/batchEdit.ts
lexiclevenger: Also change "edit" to "edited"

emenslin (changes requested):
I am not even entirely sure what is happening but this data set is basically impossible to work with because of how broken it is. 
Here's a screen recording of most of the incorrect behavior I experienced, there's definitely more that's messed up but this is what I got on video

https://github.com/user-attachments/assets/b52229c4-40a4-4c2f-93b4-4cb37e413dba

Here's the query I used to make this data set (in lsumzmammals9823): 
[Batch edit test.json](https://github.com/user-attachments/files/16764048/Batch.edit.test.json)

and then this is the specific data set: https://lsumzmammals9823-wbimprovements.test.specifysystems.org/specify/workbench/80

All of these issues I can consistently recreate with this query and data set

emenslin:
Simple search is broken 

https://github.com/user-attachments/assets/c06c8c55-232d-427c-a182-539c58a42495



[Specify 7 Crash Report - 2024-08-27T15_27_11.527Z.txt](https://github.com/user-attachments/files/16764524/Specify.7.Crash.Report.-.2024-08-27T15_27_11.527Z.txt)

emenslin:
Visual issues: 
-	‘deleted cells’ navigator color matches an accent color in preferences but all the other navigator colors are standard, was this intentional? Either way even the default is a little hard to read in dark mode 
Custom color preferences: 
image


Default preferences: 
Screenshot 2024-08-27 115252


-	‘new cells’ and ‘updated cells’ are the same color in dark mode but very different colors in light mode 
Dark mode:

Screenshot 2024-08-27 103302

Light mode:
image


-	‘matched and changed cells’ navigator is too light and hard to read in dark mode 
Screenshot 2024-08-27 103838

Bugs: 
-	When clicking ‘create record set’ it's named ‘WB upload’ 

https://github.com/user-attachments/assets/2df0a076-43bd-41eb-8cba-7d011d0d3c3c


-	In batch edit you are allowed to edit the timestamp created but it does not actually change it if you upload the data set. This should be a read only field 

https://github.com/user-attachments/assets/2d537f7d-75b5-4828-8abe-304359ff30c8


-	If you try to batch edit a locality data set with latitude and longitude values it adds zeros to the end which makes the cells show up as ‘updated cells’ after validation even though the cells were not edited 

https://github.com/user-attachments/assets/b1b84308-e230-483a-8ca4-55fbea7ac1ef

Last point, I feel like using the word 'commit' might not be super clear to the users. I am open to hearing other opinions but I think switching it to something like 'upload' would make a more sense. @grantfitzsimmons what are your thoughts on this?

grantfitzsimmons:
@emenslin:
>Last point, I feel like using the word 'commit' might not be super clear to the users. I am open to hearing other opinions but I think switching it to something like 'upload' would make a more sense. @grantfitzsimmons what are your thoughts on this?

Ha! Ironically I suggested 'Commit' as it was the clearest term for what we were doing to me. 

I think we could use 'Update' instead if you would prefer, but I think 'Upload' is misleading.

emenslin:
> Ha! Ironically I suggested 'Commit' as it was the clearest term for what we were doing to me.
> 
> I think we could use 'Update' instead if you would prefer, but I think 'Upload' is misleading.

Of these 3 my personal favorite would be update but if you think commit is the clearest term for the user then that works too

realVinayak:
Nice catches, especially the float and timestampcreated. Yeah, I might just make it readonly, that along with timestampmodified. Let me know if there are other fields that should be readonly.

Oops, I didn't consider dark mode.

emenslin:
First point: several tables are missing from the query builder (e.g sp audit log, collection, discipline, etc.) that are present in edge. You can still manually type them in the link to get there but they should be on the list 

Fields I think should be read-only:
- ID fields should be read only or you should just not be able to select them for batch edit as they currently throw an error when they are selected. 

This is with collection object ID but i also noticed a similar error with accession ID so I imagine it's with every ID field 

https://github.com/user-attachments/assets/f75a711a-edde-420f-a65b-7576b33e7e06

- Preferred taxon through determinations. You can't edit that through determinations in the forms but you can do it through taxon so I don't know what we want to do with this one.
- I would also consider modified by agents and created by agents 
- Tree ranks as some tree ranks can be changed if there are ranks that can be skipped but not all can so it's something to think about 
- A lot of the loan preparation fields (e.g is resolved, quantity, quantity returned, quantity resolved,etc.) changing these results in an error when trying to validate as well 
- Version should also be read only
- spaudit log, it is very messed up if you try to edit it but you shouldn't be able to anyway 
- All/most system tables 

There are probably more but these are the ones I caught for the time being. Some of these I think can be discussed but I think it's better to make them read only first and open them later if people want it.

lexiclevenger (changes requested):
Theresa brought to my attention that the Results view seems to be showing some of the wrong fields. Here, only a new Agent record was created, but it shows the column for the formatted Locality field and does not show the column for lastName:



https://github.com/user-attachments/assets/647d65a4-d236-4da9-bdf9-4bca0c843b3e



https://spnhcdatabase-wbimprovements.test.specifysystems.org/specify/workbench/67

emenslin (changes requested):
Okay a few more points: 
-	The create a record set behavior is weird. From what I have tested it seems like if you don't change any of the fields for whatever the base table for the data set is but change other fields and then try to create a record set it just doesn't work. I think either it should not create a record set or it should just create a record set with all the records in the base table. Just something that is not just creating a blank record set of the base table would be good in my opinion 

https://github.com/user-attachments/assets/904cd4f2-eb31-45bd-affc-cb155484ff0d


-	Can still edit modifiedbyagent and createdbyagent and it doesn’t actually change anything when you upload. Did you decide not to remove this field or was it an accident? 
-	Can hide all fields in an audit log query and then still press batch edit but it throws an error. I don’t think anyone would do this so it isn’t a big deal but is there anyway to just disable the button for the audit log? 

https://github.com/user-attachments/assets/ba9e76bb-e62b-42f8-aa81-69407149fc3c


- If you try to batch edit a taxon query and go back to the data mapper all the fields get unmapped. Not sure if this happens in any other fields but I only noticed it with the taxon tree. I couldn't recreate it in edge but weirdly enough I just ran into basically the same thing in https://github.com/specify/specify7/pull/5091#pullrequestreview-2269942980 

https://github.com/user-attachments/assets/08078875-73cf-45ab-92c0-438598477542



None of these are major just a few little things I noticed. It looks like everything else I brought up in my first few comments was fixed though so good job

realVinayak:
@lexiclevenger that actually seems like a bug in production too. See: https://sp7demofish.specifycloud.org/specify/workbench/2/. 

Looks like a bug in their library. Here: https://github.com/specify/specify7/blob/f58f0d43c4441ee1718ab92823f9a1da3d98bed2/specifyweb/frontend/js_src/lib/components/WbActions/useResults.ts#L71

It, internally, gets the physical column, but that one is quite off (during debugging, it just gave back the same physical column as the visual column). Obviously, this also means that removing below resolves the issue.
https://github.com/specify/specify7/blob/f58f0d43c4441ee1718ab92823f9a1da3d98bed2/specifyweb/frontend/js_src/lib/components/WbActions/useResults.ts#L65

Will need to debug this more, but this might be one of the few "won't fix", since it is an bug in workbench too.

realVinayak:
@emenslin 
> If you try to batch edit a taxon query and go back to the data mapper all the fields get unmapped

Refreshing the page should resolve that. here, and in that other pr.

emenslin:
> 4. Refreshing the page should resolve that. here, and in that other pr.

 Refreshing does not fix it here, I'm not sure about the other pr though

https://github.com/user-attachments/assets/6727cf49-ef80-4c65-9387-3178f59e6080


There is also now a migrations issue in edge where if the db has been used for batch edit you cannot use workbench. Not sure what exactly we can do about it before this pr gets merged but I wanted to make note of it here. 
Edge: 

https://github.com/user-attachments/assets/92baec91-13fa-4303-a02e-03f72af5e44c


[Specify 7 Crash Report - 2024-08-30T14_42_31.496Z.txt](https://github.com/user-attachments/files/16818607/Specify.7.Crash.Report.-.2024-08-30T14_42_31.496Z.txt)

realVinayak:
@emenslin The migration issue should be resolved now. We should, in future, be using [default](https://docs.djangoproject.com/en/5.1/ref/models/fields/#db-default) in django 5.0. Once it builds, just open the database under this branch, and that should be enough.

emenslin (changes requested):
- Can edit collection information but it throws a validation error so collection (along with discipline, division, and institution) should be read only

https://github.com/user-attachments/assets/408c6095-5a55-409a-b52a-891145839cc7


- Also sorry I forgot to address this in my previous comment but the main issue I have with leaving the various preparation fields as editable is that you cannot validate, even if you don't edit anything it throws an error 

https://github.com/user-attachments/assets/e55f0b39-0880-44c6-9ac4-639860c3cfd3



- You can also commit a data set that has determination is current set to false but cannot create a collection object without a current determination 
Batch edit: 

https://github.com/user-attachments/assets/03d6ca62-8d1f-4666-a8ae-a296d7b00431



Data entry: 

https://github.com/user-attachments/assets/58167bb8-978e-41d2-813c-06537ce931c2


- There's some weird shifting when scrolling some data sets

https://github.com/user-attachments/assets/97a6e00d-39f7-4674-aa74-7bd0f57240fd


- Adding a nonexistent taxon causes a validation error. If it should throw an error it should show in the cell after validation or at the very least the error should be more clear

https://github.com/user-attachments/assets/e8b3c39b-945e-4511-974e-151796e7ea93


- Lastly, can you give some more information on the remote preferences? I am not seeing any difference when having one or the other or neither but I'm not sure if I just don't understand or if they aren't working

realVinayak:
> collection (along with discipline, division, and institution) should be read only

smart. But I'm guessing only as a relationship? Meh, I might just make it for all (even as base table). there might be use cases for batch-editing these, but we don't encounter more than 20 anyways

emenslin:
> smart. But I'm guessing only as a relationship? Meh, I might just make it for all (even as base table). there might be use cases for batch-editing these, but we don't encounter more than 20 anyways

Yes I think even as a base table these should all be read only

pashiav (commented):
`umherb_1_10_24` db has migration issues with this PR. 

Talked to @realVinayak: Migrations should have been done automatically, but it seems to have issues related to attachmentdataset. 

`wb-improvements`

https://github.com/user-attachments/assets/5e19f0a4-eda0-4248-8093-a221b7fd2b4a


`edge` expected behavior 
Screen Shot 2024-09-03 at 9 36 55 AM

[Specify 7 Crash Report - 2024-08-28T14_40_09.214Z.txt](https://github.com/user-attachments/files/16850451/Specify.7.Crash.Report.-.2024-08-28T14_40_09.214Z.txt)

alesan99 (changes requested):
I haven't thoroughly tested everything yet but I noticed a couple of things so far:

1. 
The Batch Edit button in the Query view is based on the currently selected fields, rather than the columns that appear in the last preformed query. This means you can accidentally attempt to Batch Edit unsupported tables if you don't re-run the query.
**Not pictured:** You can even proceed to Batch Edit without an error as long as your new query has the same number of columns as the last results.

https://github.com/user-attachments/assets/877a2874-f388-43bd-a8c2-d8a9cac826d6

2. 
The gears on the Data Mapper are still functional in Batch Edit. This lets you change the Upload plan after its already done.

https://github.com/user-attachments/assets/c2a30c34-b206-4e64-8001-7602193972d2

3.
This ones minor, but (empty string) appears and is treated like a change even though it seems to stay as a true empty string afterwards. Not exactly sure what causes this one.
chrome_d8BFCvKFcN

realVinayak:
@alesan99 nice catches. Can you send a link to the dataset for the third one?

I'm not sure for the first one. what do you think? @emenslin @specify/ux-testing should we force users to save the query in general (that'll actually catch this case + make sure they don't get the save blocker + very simple to implement) before they can batch-edit. Otherwise, I can make it take the last saved query's fields.

alesan99:
> @alesan99 nice catches. Can you send a link to the dataset for the third one?

https://kubirds82523-wbimprovements.test.specifysystems.org/specify/workbench/92
I'm having trouble recreating it but this is the original query:
https://kubirds82523-wbimprovements.test.specifysystems.org/specify/query/59/

emenslin:
> should we force users to save the query in general (that'll actually catch this case + make sure they don't get the save blocker + very simple to implement) before they can batch-edit. Otherwise, I can make it take the last saved query's fields.

I think we should force the user to run the query before being able to batch edit. I think because there are a lot of fields that are read only I can only imagine the potential problems that could occur with this issue. I think by disabling the batch edit button until the query is rerun it will catch all the potential issues and make things less confusing for the user.

emenslin:
Still existing bugs/improvements: 

- [ ] Collection, discipline, division, and institution should be read only in all cases (base table and relationship) 
- [ ] Trying to validate a data set that includes loan preparation fields throws an error (related to #2565) 
- [ ] You can commit a data set that has determination 'is current' set to false but cannot create a collection object without a current determination 
- [ ] Weird shifting when scrolling some data sets
- [ ] Trying to create a taxon without all necessary fields causes an unlear validation error, the error should explain what fields are missing
- [ ] `umherb_1_10_24` db has migration issues with this PR
- [ ] The Batch Edit button in the Query view is based on the currently selected fields, rather than the columns that appear in the last preformed query. Batch edit should be disabled until query is rerun if changes have been made 
- [ ] The gears on the Data Mapper are still functional in Batch Edit. This lets you change the Upload plan after its already done
- [ ] (empty string) appears and is treated like a change even though it seems to stay as a true empty string afterwards
- [ ] Link to batch edit documentation in the batch edit dialog if no data sets exist (similar to batch attachment upload) 
- [ ] UI for `deferForMatch=true` and `deferForNullCheck=true` instead of being in remote preferences

realVinayak:
@emenslin @specify/ux-testing 

Implemented the UI for batch-edit prefs, stored as part of the upload plan. Adds a default one. The previous upload plans are still valid (backwards compatible). Thoughts?
image

emenslin:
> @emenslin @specify/ux-testing
> 
> Implemented the UI for batch-edit prefs, stored as part of the upload plan. Adds a default one. The previous upload plans are still valid (backwards compatible). Thoughts? image

I think that the wording can probably be improved but I like this implementation a lot more than having to edit remote prefs. I think the first check is fairly clear but the second one is confusing and I don't know if I would've understood what it meant if you hadn't explained it to me before.

grantfitzsimmons (commented):
From @acbentley:

> Have been playing around some with batch editing and it looks and works great.  However one thing that bothers me is that after you run a query and click batch edit, it redirects you to another page and the original query disappears.  So, if you have an issue or want to roll back your changes, the only way to view the results of the query is to reconstruct the query again which seems like a time waste.  Is there any way that the original query tab could remain open so that you could go back to it if needed?  Maybe there is some reason why this is not possible or inadvisable that I am not thinking of?

combs-a (commented):
Still going through this but wanted to leave a comment; if there's a true/false field like Current in Determinations set, it's a bit required to have the "ignore invisible fields for determining whether a record is empty" on to delete the record. If you add in Current to try and delete it by emptying the field it's flagged as an error since the Schema Config requires it.  Not a big thing, just wondering if that's ideal to need that specific preference enabled for that function?

I'm testing in `sdnhm_herps` so if this isn't a thing for other DBs let me know.

Also for the wording, I'd probably do something like this on the second one:

- [ ] Use only visible fields for determining if a record is 'empty'

specifyweb/stored_queries/queryfieldspec.py
```diff
@@ -177,33 +181,33 @@ def from_stringid(cls, stringid, is_relation):
         extracted_fieldname, date_part = extract_date_part(field_name)
         field = node.get_field(extracted_fieldname, strict=False)
```
realVinayak, line 183: Yikes, you're quite brave for merging production in this branch. Most of the code looks not too bad. Except this one, where I thought some insider info will be necessary.

I know getting rid of TreeRankQuery seems like an easy way, but trust me, TreeRankQuery makes batch-edit very simple. Don't take my word for it. Go to the sibling file batch_edit.py, and you'll see just how minimal changes were necessary to support dataset construction when trees are selected. (IIRC, there are like just 3 places, quite isolated from an abstraction perspective).

Here's the idea behind TreeRankQuery. Basically, each rank is considered as a relationship from tree to itself. So, Kingdom is a to-one relationship from Taxon to Taxon.  So, when user selects Collectionobject -> Determination -> Taxon (species, fullname), join path becomes

```
determination, taxon, species, fullname
```

Pros
1. No need for tree_rank and tree_field fields.
2. It's quite easy to test if join path ends with relationship (just check the last field is relationship or not). You'd previously also have to think about tree_rank and tree_field.
3. The code, before this merge, **already** constructs correct queries for something like `determination, taxon, species, createdby, firstname`, effectively allowing relationships from tree ranks. Try doing that with tree_rank and tree_field!
4. Most importantly (/s), it makes batch-edit quite simple. For every row, batch-edit dataset construction needs to look at what the IDs are. When you have TreeRankQuery, tree ranks are effectively just like any other relationship. 

Cons:
1. Hard to merge from production, which is a valid reason

sharadsw: Thanks! I was reluctant to merge prod here initially but the alternative seemed to be a bigger headache. I'll look into incorporating TreeRankQuery in the PR that follows #5417

PR #4939, opened 2024-05-20
Prevent from saving invalid latitude and longitude

### LINKED ISSUE #4822: Locality form allows user to input invalid latitude/longitude values
**Describe the bug**
When creating a locality, invalid values can be entered into the coordinate fields.

Latitude should be between -90 and 90
Longitude should be between -180 and 180

**To Reproduce**

1. Go to Data Entry
2. Create new Locality
3. Enter a value outside -90 and 90 in the Latitude field
4. Enter a value outside -180 and 180 in the Longitude field
5. See that there is no error

**Expected behavior**
User should be blocked from saving the form after inputting invalid values as in v7.

**Screenshots**

https://github.com/specify/specify7/assets/164079735/169b994b-114e-4528-8fed-0976a2c1672f

[Specify 7 System Information - 2024-04-23T22_05_39.686Z.txt](https://github.com/specify/specify7/files/15083522/Specify.7.System.Information.-.2024-04-23T22_05_39.686Z.txt)

- OS: Mac
- Browser: Chrome
- Specify 7 Version: edge
- Database Name: KUPaleoBot_8_31_2023
- Collection name: Paleobotany
- User Name: sppaleobotadmin

CarolineDenis:
Fixes #4822

### Checklist

### Testing instructions

- Go to Data Entry
- Create new Locality
- Enter a value outside -90 and 90 in the Latitude field
- Enter a value outside -180 and 180 in the Longitude field
- [ ] Verify that Save is blocked and a tooltip displays the reason on the field


specifyweb/frontend/js_src/lib/localization/locality.ts
```diff
@@ -292,4 +292,10 @@ export const localityText = createDictionary({
     'uk-ua': 'DD MM SS.SS N/S/E/W (32 45 42.84 N)',
     'de-ch': 'DD MM SS.SS N/S/O/W (32 45 42.84 N)',
   },
+  validLatitude: {
```
maxpatiiuk, line 295: (optional) would it make sense to reuse one of these strings?
https://github.com/specify/specify7/blob/4e5722aabb4ae3e8afd2ee487f4bea68eaff83d1/specifyweb/frontend/js_src/lib/localization/backEnd.ts#L178-L211

specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
```diff
@@ -74,6 +75,11 @@ function Coordinate({
     [resource, coordinateField, updateValue, step, fieldType]
   );
 
+  const [_blockers, setBlockers] = useSaveBlockers(
+    resource,
+    resource.specifyTable.fields.find((field) => field.name === coordinateField)
```
maxpatiiuk, line 80: don't do .fields.find() - use the specifyTable.strictGetField() util instead

and, on line 38 the strictGetField is already called - extract that usage into a variable and reuse it here

specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
```diff
@@ -97,6 +103,15 @@ function Coordinate({
...
     isChanging.current = true;
 
+    const latLongBlockers = isValid
+      ? []
+      : [
+          fieldType === 'Lat'
+            ? localityText.validLatitude()
+            : localityText.validLongitude(),
+        ];
+    setBlockers(latLongBlockers, fieldType);
```
maxpatiiuk, line 113: on line 34 you have a `setValidation` callback, that should also set validation blockers - can you check if you can reuse that?
and I think the call to setBlockers/setValidation should be in useEffect so that it's not made on each render

maxpatiiuk: though, `setValidation` is already called here so validation should be working:
https://github.com/specify/specify7/blob/4e5722aabb4ae3e8afd2ee487f4bea68eaff83d1/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx#L88-L89

instead of adding extra validation code, could you please check why the existing validation code is not working?

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
sharadsw: Looks like `setValidation` gets called twice somehow and so validation is not working as expected. 
This line should be blocking save correctly:
https://github.com/specify/specify7/blob/4e5722aabb4ae3e8afd2ee487f4bea68eaff83d1/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx#L89
But `setValidation` gets called once again due to this useEffect with an empty message which unsets the save blocker. I'm guessing the useEffect gets triggered due to some re-render in `useResourceValue` but still need to figure out why.
https://github.com/specify/specify7/blob/f4fc143eb5b31e1865d31e26533aa5297b3930bc/specifyweb/frontend/js_src/lib/hooks/useValidation.tsx#L116

sharadsw: I think a straight-forward but maybe not ideal solution would be to make `useResourceValue` return `setBlockers` and use that in `LatLongUi` instead of `setValidation`. What do you think?

maxpatiiuk: yes, it seems like the setBlockers() call in useResourceValue changes the blockers array, which are passed to useValidation and in turn causing it to re-set (in this case empty) validation messages.
A fix would be to integrate the useValidation and useSaveBlockers closer together to prevent them from stepping on each other
Since setBlockers will indirectly call setValidation, making useResourceValue return setBlockers seems like a good idea - give it a try.

Edge cases to consider:
- there may be timing differences between when error is shown on the screen by setValidation vs setBlocker, and how the error is shown to the user (it is shown right away, or only only if field looses/gains focus?)
- setValidation accepts a key parameter (added by Jason in https://github.com/specify/specify7/commit/4879cd59fb28dbc63d7864bafc1ea3649c740b21). I think that instead of returning setBlockers directly, it might be better for useResourceValue to return something like `React.useCallback((blockers:provideTypingHere)=>setBlockers(blockers, 'customValidation'),[setBlockers]);` (could replace `customValidation` with any other string or variable) - point being that in places where useResourceValue is used, you won't be inconvenienced by having to come up with a key.

---

additional context:

The purpose of the useValidation hook (used by useResourceValue behind the scenes to get setValidation) is to set validation messages on some html input.
If there is validation error in an HTML input, the submit button will get disabled (by browser), and so saving would be prevented until errors are fixed (with great a11y benefits and screen reader support in the process)

specifyweb/frontend/js_src/lib/localization/locality.ts
sharadsw: I don't mind the new string personally. I think it's more concise as `Actual: (value)` is not really needed on the frontend

specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
sharadsw: I'm choosing to return `setBlockers` directly as `LatLongUi.tsx` would need to specify a different key for longitude and latitude. I could add an optional key param to your callback but I'm not sure we're gaining anything from that

specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
```diff
@@ -240,5 +241,6 @@ export function useResourceValue<
     validationRef,
     setValidation,
```
maxpatiiuk, line 242: hmm, what I had in mind would be more like this:

```suggestion
    setValidation: React.useCallback((messages)=>setBlockers(messages, 'validation'), [setBlockers]),
```

this avoids the need to add new props (and the confusion of `do I pick setBlockers or setValidation?) and brings this bug fix to all existing code too

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx
sharadsw: Using 'validation' as the key every time doesn't work well as I need to set blockers on longitude as well as latitude. If I use the same key, it will remove any existing blocker from the other field. Also, `setValidation`'s signature comes from the return type of `useValidation`:
https://github.com/specify/specify7/blob/43c4b7484498e32521e123591794ea06b4614dc1/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx#L56-L64

and so the `message` param could either be an array or a string. Removing `ReturnType<typeof useValidation>` and changing the signature for `useValidation` breaks typecheck in a lot of other places. The `field` prop for `useResourceValue` accepts undefined and so I've added a check for that in `setValidation` and so this is what I came up with in the end:

https://github.com/specify/specify7/blob/de082ea89cceb1a82a041e691df651fbffbe0c76/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx#L241-L247

maxpatiiuk (approved):
looks good.
could you please construct a list of places where the setValidation returned by useResourceValue was consumed and post that list in the testing instructions
(i.e it is used in the field validation in the forms - doesn't mean every field on every form needs to be tested, but test one or two)
when testing, make sure the following behavior on this branch matches that in production:
- what happens if you enter invalid value (without moving focus out of the field)
- what happens if you enter invalid value and right away move focus out of the field
- what happens if you enter an invalid field, move focus out of the field into a different one, and then back into the invalid field
- what happens if you try to save/submit the form with an invalid field being present

sharadsw:
Fortunately, none of the other usages of `useResourceValue` use `setValidation` so we're good!

combs-a (approved):
### Testing instructions

- Go to Data Entry
- Create new Locality
- Enter a value outside -90 and 90 in the Latitude field
- Enter a value outside -180 and 180 in the Longitude field
- [X] Verify that Save is blocked and a tooltip displays the reason on the field

Looks good! Tested 90, -90, 180 and -180, then other things in and out of range including decimals. Didn't run into any issues.

image

Areyes42 (approved):
### Testing instructions

-   Go to Data Entry
-   Create new Locality
-   Enter a value outside -90 and 90 in the Latitude field
-   Enter a value outside -180 and 180 in the Longitude field
-   [x] Verify that Save is blocked and a tooltip displays the reason on the field
***
Looks good! I tested it with various formats and numbers, and the save flagger works as expected. One minor and highly unlikely scenario is if a value is placed after the [13th decimal point](https://www.reddit.com/r/theydidthemath/comments/w2e1vl/request_how_precise_are_latitudelongitude/). There is no save blocker for this, but I believe the parser just rounds off such minute values. Great job!
Screenshot 2024-06-06 at 11 14 33 AM

melton-jason (approved):
This solution works for now, but it is perhaps not the most ideal: users can still save a Locality with invalid ranges for coordinates when they are not using the LatLongUI plugin. 

That is, when the latitude1 and longitude1 fields are used directly, this error would not be caught.

Screenshot 2024-06-07 at 8 30 42 AM

I would imagine something like a business rule on those specific fields for Locality would ensure #4822 is applied systemically? 

However these changes to `useResourceValue` might still be beneficial in the longterm: so that the issue we discovered here (see https://github.com/specify/specify7/pull/4939#discussion_r1625033788) does not happen again. 

I think the best solution then would be to keep the changes to `useResoruceValue`, but move setting the blockers on the Locality to a business rule instead. 
This change is fine however if it is very likely that people will use the LatLongUI plugin rather than the `latitude1` and `longitude1 fields directly. 

(Unless this behavior is only desired when using the LatLongUI plugin, but I would probably advocate for consistency across coordinate values: especially if other tools like GeoMap, spatial queries, etc. rely on accurate values)

PR #4977, opened 2024-06-03
Fix Tree Rank Edit Verification Steps

### LINKED ISSUE #4964: Error when saving changes in tree definition forms
**Describe the bug**
After making and saving changes in definition forms, an error message is shown briefly and the page reloads without saving the changes.

**To Reproduce**

1. Go to Trees
2. Click on the pencil button next to a tree
3. Make changes to the name or remarks field and save

**Expected behavior**
No error should occur when saving changes, and the page should not reload unless changes are saved.

**Screenshots**

https://github.com/specify/specify7/assets/164079735/79c4a463-7584-4fe7-aa63-f8f5fbdc4210

[Specify 7 System Information - 2024-05-29T20_10_41.191Z.txt](https://github.com/specify/specify7/files/15490327/Specify.7.System.Information.-.2024-05-29T20_10_41.191Z.txt)

A crash report cannot be downloaded due to the page reload.

- OS: Mac
- Browser: Chrome
- Specify 7 Version: edge
- Database Name: Fitz_NHMD_Oct_2023
- Collection name: Vertebrate Paleontology
- User Name: petrifiedroadkill

acwhite211:
Fixes #4964 

Fix the tree rank edit functionality and verification steps for the UI workflow of Treeview->Treedef->Treedefitems->Add.  This is the alternate workflow that editing directly on the rank column headers in the tree view.

### Checklist

### Testing instructions

https://github.com/specify/specify7/issues/4964#issue-2324178402

### DIFF
--- a/specifyweb/businessrules/rules/tree_rules.py
+++ b/specifyweb/businessrules/rules/tree_rules.py
@@ -7,7 +7,7 @@ from specifyweb.specify.tree_ranks import *
 
 logger = logging.getLogger(__name__)
 
-@orm_signal_handler('pre_save')
+# @orm_signal_handler('pre_save')
 def pre_tree_rank_initiation_handler(sender, obj):
     if is_instance_of_tree_def_item(obj) and obj.pk is None: # is it a treedefitem? 
         if obj.pk is None: # is it a new object?
@@ -16,7 +16,7 @@ def pre_tree_rank_initiation_handler(sender, obj):
         else:
             verify_rank_parent_chain_integrity(obj, RankOperation.UPDATED)
 
-@orm_signal_handler('post_save')
+# @orm_signal_handler('post_save')
 def post_tree_rank_initiation_handler(sender, obj, created):
     if is_instance_of_tree_def_item(obj) and created: # is it a treedefitem?
         post_tree_rank_save(sender, obj)
@@ -32,15 +32,15 @@ def cannot_delete_root_treedefitem(sender, obj):
                  "node": {
                      "id": obj.id
                  }})
-        pre_tree_rank_deletion(sender, obj)
-        verify_rank_parent_chain_integrity(obj, RankOperation.DELETED)
+        # pre_tree_rank_deletion(sender, obj)
+        # verify_rank_parent_chain_integrity(obj, RankOperation.DELETED)
 
-@orm_signal_handler('post_delete')
+# @orm_signal_handler('post_delete')
 def post_tree_rank_deletion_handler(sender, obj):
     if is_instance_of_tree_def_item(obj): # is it a treedefitem?
         post_tree_rank_deletion(obj)
 
-@orm_signal_handler('pre_save')
+# @orm_signal_handler('pre_save')
 def set_is_accepted_if_prefereed(sender, obj):
     if hasattr(obj, 'isaccepted'):
         obj.isaccepted = obj.accepted_id == None


alesan99 (approved):
- [x] Go to Trees
- [x] Click on the pencil button next to a tree
- [x] Make changes to the name or remarks field and save

Reproduced the original issue on production and now it works correctly with this PR 👍
Takes a while to save but no error appears before refreshing, and changes to remarks remain.

combs-a:
- Go to Trees
- Click on the pencil button next to a tree
- [X] Make changes to the name or remarks field and save successfully
- [X] Check that the changes are still there when you reopen edit menu 

Recreated issue on ciscollections database in Minerals collection in production and it was fixed. o7

combs-a (approved):

Areyes42 (approved):
-  [x]  Go to Trees
-   [x] Click on the pencil button next to a tree
-   [x] Make changes to the name or remarks field and save
***
I tested multiple databases and couldn't replicate the original issue in production, but the issue is still resolved in this PR. 😎👍

melton-jason (changes requested):
Is there ever a case when creating or updating a TreeDefItem which also needs to ensure that every parent does not have more than one child rank? 

Screenshot 2024-06-03 at 2 30 28 PM

Take a look at the datamodel defined for [TaxonTreeDefItem](https://files.specifysoftware.org/schema/version/2.10/#taxontreedefitem) for example. 

Couldn't someone do something like: 
```py
example_item = Taxontreedefitem.objects.get(id=some_id)
example_item.parent = some_other_taxontreedefitem
# let's say some_other_taxontreedefitem already has a child taxontreedefitem
example_item.save()
# now some_other_taxontreedefitem has more than one child
```
The above code would not throw the exception, but the tree structure would be invalid. Is there some other place which would handle the invalid tree state? 
(I used update above, but same can technically be reproduced for creating).

--- 

Also, there is a frontend issue in https://github.com/specify/specify7/issues/4964#issue-2324178402. 
The frontend is _ALWAYS_ reloading the page when the user attempts to save/delete a treedefitem. 
This is very bad when an exception occurs: all of the error context is immediately lost. 

Even if this PR resolves the Issue with the backend, if any other exception happens with creating/updating/deleting tree ranks, users (and us!!!) would have a very hard time getting any crash report or error context.

alesan99 (changes requested):
Upon further testing, the error still occurs on Fitz_NHMD_Oct_2023.
1. Add remarks to the Taxon Tree
2. After loading for a while, an error will quickly appear before the page refreshes

While I was able to recreate the issue on KUBirds_8_25_23 on `edge` before, I have not been able to recreate it again on there. It looked like it got fixed when I tried it with this PR but I may have just been unable to recreate the issue again. (Or maybe it did get fixed but it still happens on certain databases?)

specifyweb/specify/tree_ranks.py
```diff
@@ -305,6 +305,7 @@ def verify_rank_parent_chain_integrity(rank, rank_operation: RankOperation):
             if rank_id is not None:
                 parent_to_children_dict.setdefault(parent_id, []).append(rank_id)
 
-    for parent_id, children in parent_to_children_dict.items():
-        if len(children) > 1 and parent_id is not None:
-            raise TreeBusinessRuleException(f"Parent rank {parent_id} has more than one child rank")
+    if rank_operation == RankOperation.DELETED:
+        for parent_id, children in parent_to_children_dict.items():
+            if len(children) > 1 and parent_id is not None:
+                raise TreeBusinessRuleException(f"Parent rank {parent_id} has more than one child rank")
```
realVinayak, line 311: hmm i'm kinda thinking the rest of the checks can be enforced by the db itself, no? the ones at lines 300-307. (so you can just get rid of them)

I know you want to "check count of children". but, this can be done more simply. HINT: get a list of parents. what needs to happen for the a node to have multiple children?

also, the child count can be simplified. don't want to

acwhite211:
Some of the verification steps and functions somehow got messed up.  I'm going to fix them now.

Also, @melton-jason are you referring to needing a business rule like we have here, or something else?

https://github.com/specify/specify7/blob/f4fc143eb5b31e1865d31e26533aa5297b3930bc/specifyweb/businessrules/rules/tree_rules.py#L10-L46

melton-jason:
> Some of the verification steps and functions somehow got messed up. I'm going to fix them now.
> 
> Also, @melton-jason are you referring to needing a business rule like we have here, or something else?
> 
> https://github.com/specify/specify7/blob/f4fc143eb5b31e1865d31e26533aa5297b3930bc/specifyweb/businessrules/rules/tree_rules.py#L10-L46

To clarify, I meant that with this change we only check the treeDefitem's children count when calling `verify_rank_parent_chain_integrity` with the Delete operation: previously it was being checked on every operation.
There might be some cases for which we should enforce the rule on Update and Create operations (like that provided in https://github.com/specify/specify7/pull/4977#pullrequestreview-2094713208). 
Unless this is intentional and we do want to check this aspect of validity only on Delete operations.

---

As far as businessrules are concerned, I was specifically referring to the [TreeRank class](https://github.com/specify/specify7/blob/0ce5e5c58964b3a262b5d8710636cc13ceed34b1/specifyweb/specify/tree_extras.py#L698-L735) defined in tree_extras.py, although the same would apply to any signals which use the `verify_rank_parent_chain_integrity` function. 

Actually, I would think a lot of the code here would be happening twice currently when a treeDefItem is saved/deleted. 
Save for the below code which raises a TreeBusinessRuleException when deleting the root treeDefItem, all of the code in business rules is duplicated with the TreeRank class. 

https://github.com/specify/specify7/blob/0ce5e5c58964b3a262b5d8710636cc13ceed34b1/specifyweb/businessrules/rules/tree_rules.py#L27-L34

Would this code not be executed twice, once for the TreeRank `save` and `delete` methods, and once again in the signals for businessrules? 😅

If the reasoning for including these is that the `delete` method override does not get executed with bulk delete operations (from this discussion: https://github.com/specify/specify7/pull/4257#discussion_r1599244273), then at the very least can we remove as many signals as possible to keep overhead/performance of operations at a minimum? As well as include a comment for the delete signals stating why they are needed when we have the `delete` method on TreeRank. 
(Although with the current implementation, the 'optimal' solution performance-wise then would probably be to just handle the deletion of treeDefItems via signals and remove the method override on the TreeRank class).

acwhite211:
So, the reason for some of the issues on the Fitz_NHMD_Oct_2023 database is because there are a lot of duplicate ranks with different treedef values.  This was an issue with the deletion process, because when added it was changing parents of multiple ranks.  The latest commit fixes this.  Still working on more changes.
```
|GeographyTreeDefItemID|GeographyTreeDefID|ParentItemID|Name     |RankID|
|----------------------|------------------|------------|---------|------|
|1                     |1                 |            |Earth    |0     |
|2                     |1                 |87          |Continent|100   |
|3                     |1                 |2           |Country  |200   |
|4                     |1                 |3           |State    |300   |
|5                     |1                 |4           |County   |400   |
|6                     |1                 |5           |District |500   |
|23                    |5                 |87          |Earth    |0     |
|24                    |5                 |23          |Continent|100   |
|25                    |5                 |24          |Country  |200   |
|26                    |5                 |25          |State    |300   |
|27                    |5                 |26          |County   |400   |
|36                    |7                 |87          |Earth    |0     |
|37                    |7                 |36          |Continent|100   |
|38                    |7                 |37          |Country  |200   |
|39                    |7                 |38          |State    |300   |
|40                    |7                 |39          |County   |400   |
|46                    |9                 |87          |Earth    |0     |
|47                    |9                 |46          |Continent|100   |
|48                    |9                 |47          |Country  |200   |
|49                    |9                 |48          |State    |300   |
|50                    |9                 |49          |County   |400   |
|51                    |10                |87          |Earth    |0     |
|52                    |10                |51          |Continent|100   |
|53                    |10                |52          |Country  |200   |
|54                    |10                |53          |State    |300   |
|55                    |10                |54          |County   |400   |
|56                    |11                |87          |Earth    |0     |
|57                    |11                |56          |Continent|100   |
|58                    |11                |57          |Country  |200   |
|59                    |11                |58          |State    |300   |
|60                    |11                |59          |County   |400   |
|61                    |12                |87          |Earth    |0     |
|62                    |12                |61          |Continent|100   |
|63                    |12                |62          |Country  |200   |
|64                    |12                |63          |State    |300   |
|65                    |12                |64          |County   |400   |
|66                    |13                |87          |Earth    |0     |
|67                    |13                |66          |Continent|100   |
|68                    |13                |67          |Country  |200   |
|69                    |13                |68          |State    |300   |
|70                    |13                |69          |County   |400   |
|76                    |15                |87          |Earth    |0     |
|77                    |15                |76          |Continent|100   |
|78                    |15                |77          |Country  |200   |
|79                    |15                |78          |State    |300   |
|80                    |15                |79          |County   |400   |
|81                    |16                |87          |Earth    |0     |
|82                    |16                |81          |Continent|100   |
|83                    |16                |82          |Country  |200   |
|84                    |16                |83          |State    |300   |
|85                    |16                |84          |County   |400   |
|87                    |1                 |1           |test     |50    |
```

realVinayak:
> This was an issue with the deletion process, because when added it was changing parents of multiple ranks. The latest commit fixes this. Still working on more changes.

My bad! I didn't expect people to have that so asked you to remove it

acwhite211:
Notes:
- The front-end refreshing issue Jason pointed out is a good point.  I believe Caroline looked into that and found a problem with implementing that.  Maybe we can re-look into that and create a new issue for the next release.
- I forgot to mention that the signal handling code is not called anymore, I removed their connecting initialization code in buid_models.py, instead it was changed to the abstract method override.  So the verification function is not called twice.  I liked the signals implementation too, but Vinny wanted to use the abstract classes.  Sorry, I just wanted to keep that signals code there for future reference, and maybe change it in the future.  
- The UI workflow path of Treeview->Treedef->Treedefitems->Add was getting errors in the back-end because the request didn't have a parent rank given, which usually signifies that it should be added as the root.

lexiclevenger (approved):
**Testing instructions**
https://github.com/specify/specify7/issues/4964#issue-2324178402

---
Everything looks good now, great job!

emenslin (changes requested):
Still does not work on Fitz_NHMD_Oct_2023 

https://github.com/specify/specify7/assets/150940840/18380870-2f3a-4dcd-9f58-11e9eac3df06

acwhite211:
> Still does not work on Fitz_NHMD_Oct_2023

Hmm, I ran it locally on that database and it seemed to work.  I tried it on the test panel and just got a timeout error, didn't find any errors in the back-end log.  I'll try it again in a sec.

realVinayak:
why not add a unit test for it?

sharadsw (approved):

PR #4998, opened 2024-06-12
Fix unload protect being triggered in locality record sets

### LINKED ISSUE #4995: Unload protect triggered when switching between records of a record set
**Describe the bug**
When switching between records in a record set the unload protect gets triggered. This does not happen in 7.9.5 and I believe was working earlier in edge so potentially broken by pr #4977. Not sure if this happens with all record sets but at least seems to happen with most locality record sets. 

**To Reproduce**

1. Go to a locality record set (e.g. https://kufish51623-edge.test.specifysystems.org/specify/view/locality/6858/?recordsetid=10) 
2. Use the arrows to navigate through a few records 
3. See unload protect 

**Expected behavior**
Unload protect should not be triggered 

**Screenshots**

https://github.com/specify/specify7/assets/150940840/a98f70bc-4c0e-41ca-ad5c-29b46f1d4ee6

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge
- Collection name: mammals
- User Name: spadmin

sharadsw:
Fixes #4995 

Caused by #4939 
The changes in that PR sets the validation blocker more times than it needs to which somehow causes an unload protect to trigger. My guess is it's a timing error.

### Checklist

- [ ] Add automated tests

### Testing instructions

- Go to a locality record set
- Navigate between records
- [ ] Verify unload protect does not trigger

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
@@ -141,7 +141,6 @@ function Coordinate({
     fieldType,
     step,
     handleFormatted,
-    setValidation,
     parser,
   ]);
 


melton-jason (commented):
As noted in https://github.com/specify/specify7/issues/4995#issue-2344362545, the issue only occurred when switching between records (most likely in a record set). 
Unload protect would not be triggered on a record if the page was refreshed. 

The unload protect is being triggered by the following line in the LatLongUI plugin: 
https://github.com/specify/specify7/blob/bb328a30a92eae17a13854356c481a5163e14759/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx#L113. 

For example, navigate to https://lsumzmammals9823-edge.test.specifysystems.org/specify/view/locality/1/?recordsetid=478 which has the following values for LatLongUI Plugin related fields: 
```json
{
  "lat1text": "30",
  "latitude1": 30,
  "long1text": "-93",
  "longitude1": -93,
  "originallatlongunit": 0,
  "srclatlongunit": 0,
}
```

Navigate to the next record https://lsumzmammals9823-edge.test.specifysystems.org/specify/view/locality/2/?recordsetid=478
which has the following field values: 
```json
{
  "lat1text": "30.837",
  "latitude1": 30.837,
  "long1text": "-91.218",
  "longitude1": -91.218,
  "originallatlongunit": 0,
  "srclatlongunit": 0,
}
```

The unload protect on the second record is being triggered by in two instances: 
- lat1text is being set to `30`
- long1text is being set to `-93`

These values are coming from the previous resource's values for those fields. 
(the field values are eventually 'reset' to their expected values). 

This is happening because `setValidation` callback in the useEffect dependency array (linked below) is being updated far more frequently since it was changed in https://github.com/specify/specify7/pull/4939. 
https://github.com/specify/specify7/blob/bb328a30a92eae17a13854356c481a5163e14759/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx#L144

Before #4939, the `setValidation` from `useResourceValue` was only being updated whenever the `InFormEditorContext` was updated. 
https://github.com/specify/specify7/blob/bb328a30a92eae17a13854356c481a5163e14759/specifyweb/frontend/js_src/lib/hooks/useValidation.tsx#L112-L114

Now, `setValidation` from `useResourceValue` is being updated whenever the resource or field changes. 
https://github.com/specify/specify7/blob/bb328a30a92eae17a13854356c481a5163e14759/specifyweb/frontend/js_src/lib/components/DataModel/saveBlockers.tsx#L105-L107

This overlaps with: 
https://github.com/specify/specify7/blob/bb328a30a92eae17a13854356c481a5163e14759/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx#L135-L138

Areyes42 (approved):
### Testing instructions

-   Go to a locality record set
-   Navigate between records
-   [x] Verify unload protect does not trigger
***
Looks good, unload protect doesn't get triggered anymore! 👍

https://github.com/specify/specify7/assets/130881183/5c37e6f2-972e-4d4c-bf93-d99943adbe43

combs-a (commented):
Unsure what's going on with this case since it's also present in v7.9.5, but unload protect is being triggered on a few record sets I just made through queries:

https://github.com/specify/specify7/assets/71564869/748219d8-9acf-4a40-8e85-8f6520b761fb

https://github.com/specify/specify7/assets/71564869/6746c9cf-945d-4cb4-803f-df4426d7a4dc

Just commenting for now w/o requesting changes since it might be a separate issue from this since it follows Elizabeth's second case more. I can test more thoroughly when I get in tomorrow.

maxpatiiuk:
(optional) we could make setValidation here consumer a fieldRef rather than field state, thus not triggering a needless re-render
https://github.com/specify/specify7/blob/bb328a30a92eae17a13854356c481a5163e14759/specifyweb/frontend/js_src/lib/hooks/useResourceValue.tsx#L241

combs-a (approved):
Looking more into this, it seems like the issue I was having before was a separate one ( PR #3959 ) and was able to make a new record set that has the right bug (no unload protect in v7.9.5, but does have it in edge).

https://github.com/specify/specify7/assets/71564869/b62ca390-5824-459c-b184-e5748aabe79d

### Testing instructions

- Go to a locality record set
- Navigate between records
- [X] Verify unload protect does not trigger

Looks good! 🎉

sharadsw:
Created some record sets with WB and queries but I'm not able to reproduce the bug. I think the underlying cause for those is definitely different than the one for this issue. I think you're correct about it being related to #3259

PR #5001, opened 2024-06-12
Fix readonly mode triggering permission denied error in FormMeta

### LINKED ISSUE #4997: Permission denied error when attempting to enable/disable read-only mode
**Describe the bug**
When enabling read-only mode on any form, a 'Permission Denied' error briefly appears, even if the user has the necessary role or permission (E.g. Institution Admin). Furthermore, clicking the form meta button to disable read-only mode triggers the same error, preventing access to the form meta menu. In this case, both the 'Dismiss' and 'Go To Home Page' buttons redirect to the Specify homepage, making them redundant for this scenario.

**To Reproduce**

1. Go to any pre-existing form (e.g. Collection Object)
2. Click on the form meta button in the top right
3. Click on 'Enable Read-Only Mode'
4. See error briefly while loading
5. Click on the form meta button in the top right again
6. See permission error again

**Expected behavior**
This issue does not occur in v7.9.5, and switching read-only mode functions without errors. Regarding the 'Permission Denied' error, the dismiss button should redirect users back to the original form they opened, rather than the homepage.

**Screenshots**
`Edge`

https://github.com/specify/specify7/assets/130881183/7e1a560c-a3eb-45f8-985c-2e065fbcaf8e 

`v7.9.5`

https://github.com/specify/specify7/assets/130881183/d42a2e0e-143a-4374-bacb-e3b479eb2658

- OS: macOS 14.5
- Browser: Chrome
- Specify 7 Version: edge
- Database Name: KUFish_5_16_23
- User Name: spfishadmin

sharadsw:
Fixes #4997

### Checklist

- [ ] Add automated tests

### Testing instructions

- Go to any pre-existing form (e.g. Collection Object)
- Click on the form meta button in the top right
- Click on 'Enable Read-Only Mode'
- Click on form meta again
- [ ] Verify there is no permission denied error and read only mode can be disabled again
- [ ] Re-run the same steps for a form with merge records enabled (Agent, Locality, PaleoContext, CollectingEvent)

Additional test:
- In Readonly mode, make a query on LoanPreparation
- [ ] Verify the page loads correctly and there is no permission denied error dialog

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormMeta/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormMeta/index.tsx
@@ -24,9 +24,9 @@ import { isTreeResource } from '../InitialContext/treeRanks';
 import { interactionTables } from '../Interactions/config';
 import { recordMergingTableSpec } from '../Merging/definitions';
 import { Dialog } from '../Molecules/Dialog';
+import { hasTablePermission } from '../Permissions/helpers';
 import {
   ProtectedAction,
-  ProtectedTable,
   ProtectedTool,
 } from '../Permissions/PermissionDenied';
 import { UnloadProtectsContext } from '../Router/UnloadProtect';
@@ -218,14 +218,10 @@ function MetaDialog({
             </ProtectedTool>
             <ProtectedAction action="update" resource="/record/merge">
               <ProtectedAction action="delete" resource="/record/merge">
-                <ProtectedTable
-                  action="update"
-                  tableName={resource.specifyTable.name}
-                >
-                  {resource.specifyTable.name in recordMergingTableSpec && (
-                    <MergeRecord resource={resource} />
-                  )}
-                </ProtectedTable>
+                {resource.specifyTable.name in recordMergingTableSpec &&
+                hasTablePermission(resource.specifyTable.name, 'update') ? (
+                  <MergeRecord resource={resource} />
+                ) : undefined}
               </ProtectedAction>
             </ProtectedAction>
           </>
--- a/specifyweb/frontend/js_src/lib/components/FormMeta/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormMeta/index.tsx
@@ -24,7 +24,7 @@ import { isTreeResource } from '../InitialContext/treeRanks';
 import { interactionTables } from '../Interactions/config';
 import { recordMergingTableSpec } from '../Merging/definitions';
 import { Dialog } from '../Molecules/Dialog';
-import { hasTablePermission } from '../Permissions/helpers';
+import { hasPermission, hasTablePermission } from '../Permissions/helpers';
 import {
   ProtectedAction,
   ProtectedTool,
@@ -216,14 +216,12 @@ function MetaDialog({
                 ))}
               </ProtectedAction>
             </ProtectedTool>
-            <ProtectedAction action="update" resource="/record/merge">
-              <ProtectedAction action="delete" resource="/record/merge">
-                {resource.specifyTable.name in recordMergingTableSpec &&
-                hasTablePermission(resource.specifyTable.name, 'update') ? (
-                  <MergeRecord resource={resource} />
-                ) : undefined}
-              </ProtectedAction>
-            </ProtectedAction>
+            {resource.specifyTable.name in recordMergingTableSpec &&
+            hasPermission('/record/merge', 'update') &&
+            hasPermission('/record/merge', 'delete') &&
+            hasTablePermission(resource.specifyTable.name, 'update') ? (
+              <MergeRecord resource={resource} />
+            ) : undefined}
           </>
         }
         header={formsText.recordInformation()}
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx
@@ -16,11 +16,11 @@ import { tables } from '../DataModel/tables';
 import type { RecordSet, SpQuery, SpQueryField } from '../DataModel/types';
 import { ErrorBoundary } from '../Errors/ErrorBoundary';
 import { TableIcon } from '../Molecules/TableIcon';
-import { hasToolPermission } from '../Permissions/helpers';
 import {
-  ProtectedAction,
-  ProtectedTable,
-} from '../Permissions/PermissionDenied';
+  hasPermission,
+  hasTablePermission,
+  hasToolPermission,
+} from '../Permissions/helpers';
 import { SaveQueryButtons, ToggleMappingViewButton } from './Components';
 import { useQueryViewPref } from './Context';
 import { QueryEditButton } from './Edit';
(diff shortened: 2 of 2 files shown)


maxpatiiuk (commented):
hmm, could you please explain what was causing the bug and why this fixes it?
it's hard without any context to know if this is a good fix or not.
i.e is there an issue with this particular check in this one place, or is it a more systematic failure that could affect all such checks anywhere in the codebase?
and what change lead to this error to start appearing where as it did not before?

melton-jason (commented):
While this does fix the issue in this case, the ProtectedTable would still cause the error dialog in another case: when viewing any LoanPreparation Query. 

Is this intended? I would instead think disabling or hiding the button in this case would be more user-friendly (rather than make the query completely non-functional in ReadOnly Mode) 

https://github.com/specify/specify7/assets/64045831/7cfc3e6b-286a-4b2f-acee-19cbe2afde85

https://github.com/specify/specify7/blob/bb328a30a92eae17a13854356c481a5163e14759/specifyweb/frontend/js_src/lib/components/QueryBuilder/Header.tsx#L112-L128

This raises questions about the general behavior of permissions in Specify, especially relating to tools such as these. I don't think that raising an error in these cases is ever helpful (at least in the current state. It would be better if the Error Dialog was dismissible). 

In these cases, I would imagine the intended behavior is to either disable the tool/button, or not render it in the first place. 

In my opinion, raising the error dialog should only occur when the user _explicitly_ tries to do something against their permissions, such as navigating to a form they don't have read access to. Small tools/buttons such as the `Enable Read Only Mode` or `Return Loan Records` buttons are already in the Interface while the user is trying to accomplish another task entirely. 

In the case of `Return Loan Records` in the QueryBuilder, the user is likely there to Query on records rather than Return the Loans Preparations, so violating the table create permission should be approached with less severity/notice to the user. 
We already hide the button without raising an error in the InteractionsDialog. 

https://github.com/specify/specify7/blob/bb328a30a92eae17a13854356c481a5163e14759/specifyweb/frontend/js_src/lib/components/Interactions/InteractionsDialog.tsx#L66-L88
--- 
Also, I would think the `Dismiss` button on the Permission Denied Dialog redirecting the user back to the home page is unintentional.

maxpatiiuk:
Thank you for the details

> In these cases, I would imagine the intended behavior is to either disable the tool/button, or not render it in the first place.

Yeah, ideally the permission denied dialog would only be shown in two cases:
- The user directly navigated to a URL of the action they don't have a permission for (in which case you need to explicitly tell them that, rather than display an empty page)
- Knowing that the user does not have a permission is for some reason important (rather than simply hiding the feature). One such use case I remember is when we added a new feature (might have been agent merging), but the permission for which wasn't included by default (so no one had is except for the admins), thus people might not even know that this feature was added because they won't see it in the UI. Where as seeing a disabled "Merge records" button in the UI will at least be a cause for some investigation (or a support email)

Given the above, I think we should try to follow this rule:
- When displaying a button/link to an action the user has no permission for, use one of the has permission functions (i.e hasTablePermission) to conditionally render the UI only if permission is given, and else don't display anything
- On the other hand, when the user is at the URL for the page they have no permission to, the content of the page needs to be wrapped in one of the component checks (i.e ProtectedTable), which will render the UI, or an error dialog which permission is missing. Think of it as we are wrapping the area in one of the ProtectedAction/ProtectedTable/... component to designate it as a "protected area".

If we are careful about where we use function check and where we use the component check, most issues with readonly should be resolved. If that is not enough however, we could add a check inside the component check to not display an error dialog if user is in readonly mode

Given the above, it becomes more clear that the use of Protected component checks for a button is incorrect - permission function check should be used instead.

---

As far as the other places were permission checks are used, it would be great if someone could check them (maybe one of the student programmers), though that would have to be compared to the other priorities

specifyweb/frontend/js_src/lib/components/FormMeta/index.tsx
```diff
@@ -218,14 +218,10 @@ function MetaDialog({
...
               <ProtectedAction action="delete" resource="/record/merge">
-                <ProtectedTable
-                  action="update"
-                  tableName={resource.specifyTable.name}
-                >
-                  {resource.specifyTable.name in recordMergingTableSpec && (
-                    <MergeRecord resource={resource} />
-                  )}
-                </ProtectedTable>
+                {resource.specifyTable.name in recordMergingTableSpec &&
```
maxpatiiuk, line 221: could you also convert the ProtectedAction calls above into function checks?

melton-jason (approved):
I agree with https://github.com/specify/specify7/pull/5001#issuecomment-2167039171 in that we should also go through the other uses of the `Protected` components (ProtectedAction, ProtectedTool,  ProtectedTable) and the `hasPermission` functions to ensure no other error like this is present. 

It _would_ be a great exercise for another student developer! 

---

Also from https://github.com/specify/specify7/pull/5001#pullrequestreview-2116275954

> Also, I would think the `Dismiss` button on the Permission Denied Dialog redirecting the user back to the home page is unintentional. 

I would imagine this is a bug. 

--- 

Although neither of these Issues likely _have_ to be resolved in this PR: the PR might become too big, and testing instructions would be a nightmare! (As well as keep potential merge conflicts to a minimum). 

This smaller scope works for now. Nice changes!

emenslin (approved):
### Testing instructions

- Go to any pre-existing form (e.g. Collection Object)
- Click on the form meta button in the top right
- Click on 'Enable Read-Only Mode'
- Click on form meta again
- [x] Verify there is no permission denied error and read only mode can be disabled again
- [x] Re-run the same steps for a form with merge records enabled (Agent, Locality, PaleoContext, CollectingEvent)

Additional test:
- In Readonly mode, make a query on LoanPreparation
- [x] Verify the page loads correctly and there is no permission denied error dialog

Looks good, I did not run into any other errors with read only

Areyes42 (approved):
### Testing instructions

-   Go to any pre-existing form (e.g. Collection Object)
-   Click on the form meta button in the top right
-   Click on 'Enable Read-Only Mode'
-   Click on form meta again
-  [x]  Verify there is no permission denied error and read only mode can be disabled again
-   [x] Re-run the same steps for a form with merge records enabled (Agent, Locality, PaleoContext, CollectingEvent)

Additional test:

-   In Readonly mode, make a query on LoanPreparation
-   [x] Verify the page loads correctly and there is no permission denied error dialog
***
Looks good now, no longer throwing error dialogs

lexiclevenger (approved):
### Testing instructions

- Go to any pre-existing form (e.g. Collection Object)
- Click on the form meta button in the top right
- Click on 'Enable Read-Only Mode'
- Click on form meta again
- [x] Verify there is no permission denied error and read only mode can be disabled again
- [x] Re-run the same steps for a form with merge records enabled (Agent, Locality, PaleoContext, CollectingEvent)

Additional test:
- In Readonly mode, make a query on LoanPreparation
- [x] Verify the page loads correctly and there is no permission denied error dialog

---
Looks good!

https://github.com/specify/specify7/assets/164079735/271c6239-2e44-4261-bbe8-cce99ef0f6bd

I tested querying for Loan Preparations and some other tables with no issue.

PR #5008, opened 2024-06-14
Add new tables and fields for MOTs

### LINKED ISSUE #4640: Add support for 'Multiple Object Types'
It is not currently possible to manage several different types of collection objects in a single collection that use different names and attribute fields. Managing a single collection with multiple materials using this definition has never before been possible in Specify. Each discipline can currently only have one taxon tree and no types are given to an object.

This prevents users from querying specimen records managed by the same people in the same space in addition to preventing them from using the same incrementing/standardized catalog numbering scheme. The workflow for cataloging specimens that are similar/housed together but named differently becomes difficult when cataloging across multiple collections. 

## Components
- [x] https://github.com/specify/specify7/issues/4982
- [x] https://github.com/specify/specify7/issues/4981
- [x] https://github.com/specify/specify7/issues/4833
- [x] https://github.com/specify/specify7/issues/4829

## Desired Solution

This would involve creating a new table and relationship field on Collection Object.
- A new field would be added to Collection Object in the data model named `objectType`
- A new table would be added named `collectionobjecttype`
- This table would contain a number of fields used for application business logic
   - `CollectionObjectTypeID` (int) - Database ID
   - `Name` (text) - Name of the type (i.e. Fossil, Rock, Meteorite, etc.)
   - _`IsLoanable` (boolean) - Indicates whether this type of material can have its preparations be transacted in interactions (to be reconsidered)_
   - `Version` (int) - Version to identify which revision it is
   - `TimestampCreated` (datetime)
   - `TimestampModified` (datetime)
   - `CreatedByAgentID` (FK to Agent) - Created by agent
   - `ModifiedByAgentID` (FK to Agent) - Modified by agent
   - `CollectionID` (FK to Collection) - Used to scope the table to the collection it was created in
   - `TaxonTreeDefID` (FK to TaxonTreeDef) - Indicates which taxon tree definition contains the names associated with this object

### Business Logic & Requirements

- Each `collectionobjecttype` would be scoped to a particular collection

- An `objectType` would be chosen from a pick list on the Collection Object form 

- When creating a new Determination record, the system should check the `objectType.taxonTreeDef` field to find the tree to search for the user's entered name when searching in the `taxon` or `preferredTaxon` query combo boxes

- `objectType` should not be a required field but can be made to be using the form editor.

- Conditional forms must be able to use the value from the `objectType` field to determine the layout in an intuitive way for users
   - If the user can define a form definition as follows:
   ```xml
     <cell type="panel" coldef="100px,2px,76px" colspan="6">
       <rows>
         <!-- This view definition will be displayed in Specify 6, or if all other conditions don't match -->
         <row>
           <cell type="label" label="Without any object type, this will display." />
         </row>
       </rows>
       <rows condition="objectType=Rock">
         <row>
           <cell type="label" label="This is a rock!" />
         </row>
       </rows>
       <rows condition="objectType=Mineral">
         <row>
           <cell type="label" label="This is a mineral!" />
         </row>
       </rows>
     </cell>
   ```
   That would be satisfactory. This would enable users to create forms or display subviews based on the value, which would also enable users to create specific determination forms or collection object attribute forms based on the value of `objectType`.

_To be reconsidered:_
- If the `IsLoanable` field is `TRUE`, preparations associated with the object can be used in interactions
- If the `IsLoanable` field is `FALSE`, preparations associated with the object cannot be used in interactions.

### Interface

 - The user should be able to create a new `collectionobjecttype` record easily in the interface.
   - The user should be able easily name the object type
   - On this form, the user should be able to search for a taxon tree from any discipline and choose it from the list

- The query builder should display all tree ranks across all trees that have objects in the current collection at once

- The user should be able to switch between all taxon trees used in a single discipline from the tree viewer

- If a query is built in a collection with Collection Objects with various different `objectType` values and different taxon trees associated with the `collectionobjecttype` records linked by `objectType`, the query builder must be able to search across all trees at once.

- If the object type is changed, the determination should be cleared and the user must select a new name

- When the user creates a new determination, the taxon tree searched when entering a `taxon` or `preferredTaxon` value must be based on the object type's tree (if `objectType` is present), otherwise it should use the discipline default

- When creating a new Collection Object record, if there is no `objectType` value, it will use the default taxon tree for the current discipline.
   - If there is an `objectType` value, the taxon tree linked to that object type will be used when creating the determination.

### Data Unique to a Type (Attribute)

After discussing internally approaches to handling attribute data, it seems the frictionless and most effective solution (at least until the discontinuation of Specify 6) is to continue managing attribute data in a single table between all object types.

To be able to successfully display certain fields when a specific object type is selected, we need to leverage the conditional form rendering system that will be added in the 7.9.4 update.

image

This means that all attribute fields will be "unhidden" in the schema and visible all at once when building queries. It would be up to the user to define conditional forms that display only the fields they desire to show for that particular object type.

This allows users to store "Weight" or any attribute in a single field across different object types, simplifying searches across various `objectType`s.

In summary, no changes would be necessary in this area and it is up to the user to define conditional forms based on type to hide or show certain attribute data based on the object type. The object type only influences what names are returned when creating a determination.

### Trees
- [x] https://github.com/specify/specify7/issues/4829

Because the `TaxonTreeDefID` (FK to TaxonTreeDef) indicates which taxon tree definition contains the names associated with an object, adding support for MOTs requires a fundamental change in the scoping hierarchy in Specify.

We would need to add the ability to view and switch between several taxon trees in a single discipline.

Because the trees would need to share a common schema configuration, the model I propose is to have it so that every tree is **specific to that discipline**, so even if there are 3, 4, or more trees in a discipline, they cannot be accessed outside that discipline.

https://github.com/specify/specify7/assets/37256050/0198fe8a-6269-4e2f-806c-286f8b8bf7d8

It could be as simple as a picklist that switches between the tree definitions so that the user can edit and view all of the trees linked to the various objects in that collection.

Specify could return all object types in the current collection and show only the trees that are connected to types in that list.

#### Schema and Scoping

At present, the interface schema is scoped to the discipline. This means that all object types must share the same field captions and descriptions for all data model fields in the application. That means `text1` cannot be "Size" for rocks and "Gleam" for minerals, it must be the same between both object types.

Ultimately this means that collection objects within the same collection (despite having multiple object types) must conform to the same format for all resources scoped to collection, discipline, division, and institution. 

**The following will be shared between all collection objects in a single collection regardless of object type:**

---

#### Institution, 
Accession*, Journal, Permit, Reference Work, Storage

#### Division
Accession*, Agent, Collector, Exchange In, Exchange Out, Funding Agent, Group Person, Repository Agreement

#### Discipline
Attribute Def, Collecting Event, Collecting Event Attribute, Collecting Trip, Collecting Trip Attribute, Exchange In Prep, Exchange Out Prep, Field Notebook Page, Field Notebook Page Set, Gift, Gift Agent, Gift Preparation, Loan, Loan Agent, Loan Preparation, Loan Return Preparation, Locality, LocalityCitation, LocalityNameAlias, PaleoContext, Shipment, Sp Export Schema, Sp Locale Container, Geography, Geologic Time Period, Lithostratigraphy

#### Collection
Field Notebook, Pick List, Prep Type, Sp App Resource Dir

---

The exception to this now is that multiple object types requires that Specify is capable of modifying, displaying, and searching across several trees from within a single discipline. This changes the scoping hierarchy to some extent as it is now "flexible" in a way it was not previously.

#### Queries

Queries in Specify currently resolve the tree rank based on a constructed string ID. This string ID unfortunately does not resolve the rank by a unique identifier but rather by the user inputted `name` field in the `taxontreedefitem` record.

For example, here is a query in Specify 7 that shows "Phylum", "Class", "Family" and "(any rank)".

image

When you look at the string IDs associated with this query, you see the following:

```xml
<?xml version="1.0" ?>
<query name="Tree Query" contextTableId="1">
	<field stringId="1,93.collectionobjectattribute.text11" oper="1" value="10m" isNot="false" isRelFld="false"/>
	<field stringId="1,9-determinations,4-preferredTaxon.taxon.Phylum" oper="8" value="" isNot="false" isRelFld="false"/>
	<field stringId="1,9-determinations,4-preferredTaxon.taxon.Class" oper="8" value="" isNot="false" isRelFld="false"/>
	<field stringId="1,9-determinations,4-preferredTaxon.taxon.Family" oper="8" value="" isNot="false" isRelFld="false"/>
</query>
```

That means if there are two trees with a rank named "Family", the query would only know to search that rank by the relationship and rank name (`1,9-determinations,4-preferredTaxon.taxon.Family`). This decision has manifested in other issues in the past (see https://github.com/specify/specify7/issues/3351, https://github.com/specify/specify7/issues/3424, and  https://github.com/specify/specify7/issues/3340)

The other consideration with trees is that you may be unable to build a single export for all materials since each column can only be mapped to a DWC term once. If you divide the tree ranks into separate trees and select a rank, then choose a tree and create multiple string IDs and query items, you can only map one full name column to the `scientificName` field.

If we change the construction of the string IDs, we would likely break queries in Specify 6 and require a migration to handle this issue.

For these reasons, I imagine it would be easier and preferred to show all ranks for all trees associated with objects in that discipline in a single list rather than segmenting them. 

My main concern is the user experience and interface when seeing all tree ranks at once in a single list. If possible, it would be better to address these concerns without compromising compatibility with Specify 6, ensuring accurate exports, and enabling simultaneous cross-collection queries. At this time I do not feel it is essential to segment the trees into several lists but my opinion/group opinion may evolve as we develop.

## Strawman Diagram

[Diagram](https://drive.google.com/file/d/16CIfabeLqRLCSy_0uqpJK-1tGIX4FLJq/view?usp=sharing)

MOC Diagram

### LINKED ISSUE #4981: Complete schema changes to add new tables and fields for MOTs
Component of https://github.com/specify/specify7/issues/4640

> This would involve creating a new table and relationship field on Collection Object.
> - A new field would be added to Collection Object in the data model named `objectType`
> - A new table would be added named `collectionobjecttype`
> - This table would contain a number of fields used for application business logic
>    - `CollectionObjectTypeID` (int) - Database ID
>    - `Name` (text) - Name of the type (i.e. Fossil, Rock, Meteorite, etc.)
>    - _`IsLoanable` (boolean) - Indicates whether this type of material can have its preparations be transacted in interactions (to be reconsidered)_
>    - `Version` (int) - Version to identify which revision it is
>    - `TimestampCreated` (datetime)
>    - `TimestampModified` (datetime)
>    - `CreatedByAgentID` (FK to Agent) - Created by agent
>    - `ModifiedByAgentID` (FK to Agent) - Modified by agent
>    - `CollectionID` (FK to Collection) - Used to scope the table to the collection it was created in
>    - `TaxonTreeDefID` (FK to TaxonTreeDef) - Indicates which taxon tree definition contains the names associated with this object

We should also include a number of empty text fields just in case.

acwhite211:
Fixes #4981 

Add the new CollectionObjectType table to the specify models, database, and datamodel.

### Checklist

### Testing instructions

- [x] Test CollectionObjectType in the QueryBuilder (functionality for this table will come in a later PR).
- [x] Test that the table appears in the data-model view http://localhost/specify/data-model/#collectionobjecttype


grantfitzsimmons:
This should be fixing https://github.com/specify/specify7/issues/4981

realVinayak (approved):
Changes look good. Minor things that'll be nice to resolve (in comments)

Do you expect multiple schema changes to be made on this concept? Something to consider is that each subsequent change will require a migration. I think it will better that the changes get applied in production inside one migration. If we merge this in production as it is then,
1. Each schema change will be a migration, increasing number of migrations, more number of things that can go wrong when MOT gets merged to production
2. If there is a hot bug fix needed, it will be a mess to do it properly if partial MOT is inside production at any point.

I'll recommend keeping all the changes in this branch. Yes, subsequent changes will still require a migration. But, before release, we can make merge those migrations into one migration. So, in realty, only one migration will run for production users after release.

TLDR: Consider not merging into production (even after 7.9.6 gets released), and make this branch the base for all MOT development so migrations can be [squashed](https://docs.djangoproject.com/en/dev/topics/migrations/#migration-squashing) into one.

specifyweb/specify/datamodel.py
```diff
@@ -7843,7 +7843,7 @@
 
         ],
         relationships=[
-            Relationship(name='specifyuser', type='many-to-one', required=True, relatedModelName='SpecifyUser', column='specifyuser_id')
+            Relationship(name='specifyuser', type='many-to-one', required=True, relatedModelName='SpecifyUser', column='specifyuserid')
```
realVinayak, line 7846: Be careful, this will require a migration

realVinayak: And the rest of ones before this PR

acwhite211:
Ya, I too feel a little uneasy about the number of Django migrations needed for 7.9.7. I was thinking about solving multiple issues in one branch, or at least just merge issue into a 7-9-7-temp branch.  Also, we don't want to be merging this code into production before 7.9.6 is released.  The problem would then be that we would have to manually make database changes on the test-panel so that testers could try out the branches.

specifyweb/specify/datamodel.py
```diff
@@ -8161,6 +8161,38 @@
...
+            Field(name='text1', column='Text1', indexed=False, unique=False, required=False, type='java.lang.String', length=255),
+            Field(name='text2', column='Text2', indexed=False, unique=False, required=False, type='java.lang.String', length=255),
+            Field(name='text3', column='Text3', indexed=False, unique=False, required=False, type='java.lang.String', length=255)
+        ],
+        indexes=[
+
+        ],
+        relationships=[
+            Relationship(name='collection', type='many-to-one', required=False, relatedModelName='Collection', column='Collection_ID'),
+            Relationship(name='taxontreedef', type='many-to-one', required=False, relatedModelName='TaxonTreeDef', column='TaxonTreeDef_ID'),
```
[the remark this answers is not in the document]
acwhite211: Good question, I talked to Grant, we are just doing taxon trees.  In the future we might add Geography trees, but it will likely be above discipline in scope.  So, I'm going to leave it as is for now.

CarolineDenis (approved):

specifyweb/specify/datamodel.py
```diff
@@ -8154,10 +8154,42 @@
 
         ],
         relationships=[
-            Relationship(name='collection', type='many-to-one', required=False, relatedModelName='Collection', column='Collection_ID'),
-            Relationship(name='specifyuser', type='many-to-one', required=True, relatedModelName='SpecifyUser', column='SpecifyUser_ID'),
-            Relationship(name='createdbyagent', type='many-to-one', required=False, relatedModelName='Agent', column='CreatedByAgent_ID'),
-            Relationship(name='modifiedbyagent', type='many-to-one', required=False, relatedModelName='Agent', column='ModifiedByAgent_ID')
+            Relationship(name='collection', type='many-to-one', required=False, relatedModelName='Collection', column='CollectionID'),
+            Relationship(name='specifyUser', type='many-to-one', required=True, relatedModelName='SpecifyUser', column='SpecifyUse_ID'),
```
sharadsw, line 8158: Typo here? `SpecifyUse_ID`. Also does it need an underscore or does it have to be removed like the other ID columns?

acwhite211:
Closing this PR, being replaced with https://github.com/specify/specify7/pull/5032

CarolineDenis:
@acwhite211, canyon close the associated branch too? 
Screenshot 2024-07-02 at 8 38 17 AM

PR #5033, opened 2024-06-23
chore: Update to Prettier 3

maxpatiiuk:
Update Prettier to version 3
After this PR, I will gradually update other dev dependencies too (ESLint, with better configuration, TypeScript v5, Tailwind with new features, and etc) - will make sure to split these into smaller thematic updates to make reviewing easier, and to make reverting changes if necessary easier too

Command used to run Prettier on every repository file:

```sh
# running from js_src folder
npx prettier --config prettier.config.js --ignore-path ../../../.gitignore --ignore-path ./.prettierignore --write ../../../
```

### Checklist

- [ ] Add automated tests

### Testing instructions

This PR consists of 4 commits to make reviewing easier:
- First commit updates our Prettier configuration to the [latest](https://www.npmjs.com/package/@maxpatiiuk/prettier-config)
  - no settings changes, other than refactoring to use ESM over CommonJS, and removing deprecated options
- 2nd commit fixes a typo in `CITATION.cff` where `keywords` key was duplicated twice
- 3rd commit: in one localization util we called prettier API - that API changed a bit in v3 so that it now returns a promise
- 4th commit is done completely by Prettier - 0 manual changes, and thus hopefully 0 things broken. Not expecting a thorough code review for this one as it's all automated changes. Main sources of changes:
  - Prettier 3 started indenting nested ternary expressions for better readability
  - Prettier 3 started turning `typeof a[b]` into `(typeof a)[b]` to make operator precedence order clearer
  - Prettier 3 started adding trailing commas in types and interfaces - trailing commas make refactoring code either since you can move lines around without causing syntax errors. In the future PR we can enable trailing commas in functions and other places too (not doing it in this PR as that caused a +5k line diff)

I tried my best to minimize the merge conflicts this PR causes. If you still get merge conflicts in the future as result of this PR, to resolve the conflict:
- accept the change from your branch rather than the change made by this PR
- re-run Prettier on the affected files (or let CI do that automatically)


specifyweb/frontend/js_src/.prettierignore
maxpatiiuk, line 1: There are some files in the repository root that Prettier knows how to reformat - updating this prettierignore. Now we can run it on the entire repository, not just the front-end part (it doesn't touch the Python files though)

realVinayak: ^actually, I'll see if we can add a python linter too (the backend is currently inconsistently linted).

CarolineDenis (approved):

specifyweb/frontend/js_src/.prettierignore
maxpatiiuk: That would be great!
But probably a separate PR
We used to use black, flake8, darglint, pydocstyle and pyupgrade on the back-end
[see](https://github.com/specify/specify7/blob/prettier-3/.pre-commit-config.yaml#L78)
but that faided into disrepair as I was focused on the front-end and there was no one to maintain it on the back-end
some of the python tools I named can probably be replaced by prettier - don't know which one is better so will let you decide. some of these tools are supplemental to a formatter though, so would be nice to embrace them.

sharadsw (approved):

grantfitzsimmons (commented):
Worked for me! A little late :joy:

PR #5032, opened 2024-06-21
initial schema creation for geospecify

acwhite211:
Fixes #4981
Fixes #4827
Fixes #4976

This is the initial schema create for geo-specify, all in one Django migration.  Right now the new models are being added to the `sp7_models` Django app, but we might be able to change that soon and use the `specify` app instead.  This PR adds the `CollectionObjectType`, `CollectionObjectGroup`, and `CollectionObjectGroupJoin` models to the Specify 7 schema.  It also modifies the `CollectionObject` model by adding two new fields.

This PR also magically creates an initial Django migration for all the inherited schema from Specify 6.  This is done  by editing the django_migrations table with an SQL statement that is triggered in the docker entry point file.

image

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [x] Add relevant issue to release milestone

### Testing instructions

When switch to the branch, run `docker compose down` and `docker compose up --build`

For testing on the test panel, if there are no errors when starting up the instance then the migration ran without any problems.  Check to make sure that the follow tables exist in the QB:

- [x] CollectionObjectType - https://hujinnhcgeo-geodude.test.specifysystems.org/specify/query/new/collectionobjecttype/
- [x] CollectionObjectGroupType - https://hujinnhcgeo-geodude.test.specifysystems.org/specify/query/new/collectionobjectgrouptype/
- [x] CollectionObjectGroup - https://hujinnhcgeo-geodude.test.specifysystems.org/specify/query/new/collectionobjectgroup/
- [x] CollectionObjectGroupJoin - https://hujinnhcgeo-geodude.test.specifysystems.org/specify/query/new/collectionobjectgroupjoin/

Verify that each Collection has a default CollectionObjectType
- [x] Verify each Collection has its CollectionObjectType pointing to something, so they are not null.  Do this by doing a query on all collections and verify that each has something in its CollectionObjectType field.  Each CollectionObjectType name should match the name of the discipline associated with the collection.  You'll have to go through each discipline to verify this for all the collections unfortunately.

Before testing the new Trees, generate them by calling the `/api/specify_trees/add_geo_default_trees/` API call.  There is no front-end way of calling this API at the moment, so it will need to be done with a curl command.  Message me when testing and I can run the command for that test panel instance.  (Maybe we can wait to test this in a later PR where the front-end is implemented)
[...]

PR #5016, opened 2024-06-18
Multi-Discipline Tree Schema and API

acwhite211:
Fixes #4975 

Add schema to create a Many-to-One relationship between TaxonTreeDef and Discipline, so that Disciplines can have multiple Taxon Trees.  The api will also be changed to accommodate the new schema.

### Checklist

- [ ] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [x] Add relevant issue to release milestone

### Testing instructions

TBD

PR #5036, opened 2024-06-24
Add new feature to change taxon tree type in tree viewer

### LINKED ISSUE #4829: Add support for multiple trees of the same type within a single discipline
This is one component of https://github.com/specify/specify7/issues/4640

Please see the issue for a more detailed write-up with additional considerations.

> ### Trees 
> 
> Because the `TaxonTreeDefID` (FK to TaxonTreeDef) indicates which taxon tree definition contains the names associated with an object, adding support for MOTs requires a fundamental change in the scoping hierarchy in Specify.
> 
> We would need to add the ability to view and switch between several taxon trees in a single discipline.
> 
> Because the trees would need to share a common schema configuration, the model I propose is to have it so that every tree is **specific to that discipline**, so even if there are 3, 4, or more trees in a discipline, they cannot be accessed outside that discipline.
> 
> https://github.com/specify/specify7/assets/37256050/0198fe8a-6269-4e2f-806c-286f8b8bf7d8
> 
> It could be as simple as a picklist that switches between the tree definitions so that the user can edit and view all of the trees linked to the various objects in that collection.
> 
> Specify could return all object types in the current collection and show only the trees that are connected to types in that list.

### LINKED ISSUE #4978: Add ability to switch between Taxon trees in the tree viewer

This would involve a pick list in the tree viewer interface that allows a user to switch between which tree they would like to see within the discipline.

### LINKED ISSUE #4979: Add support for multiple trees in the query builder

From @realVinayak:
> regarding query builder backend, this can be very simply (and naively) accomplished by passing the object type and getting the appropriate treedef here: https://github.com/specify/specify7/blob/4e5722aabb4ae3e8afd2ee487f4bea68eaff83d1/specifyweb/stored_queries/query_construct.py#L42
> Figure out why it will "just work". the above assumes that defitems cannot be shared (like same rank ITSELF wouldn't appear in two trees). Like if Kingdom is treedefitemid=1 in one treedef, it is also not treedefitemid=1 in another treedef. basically, how the trees in specify works currently (so no worries I believe)
> 
> BUT, I think that'll be inefficient. You can make it considerably more efficient (think why the above isn't) very simply, by including the objecttype equality to the join criterion (well, technically just the first node, but you get the point)
> (https://github.com/specify/specify7/blob/4e5722aabb4ae3e8afd2ee487f4bea68eaff83d1/specifyweb/stored_queries/query_construct.py#L48)
> 
> ^ofc that's for querying a specific tree. querying all trees at once is simpler (but I guess you can't say (any tree) -> species), like you could have (Fossil) -> species)
> 
> theoretically, doing something like (any tree) -> species is also straightforward (will think about perf impact). you could have an  OR condition here (look up all treedefitemids which have species, put them in the OR below
> https://github.com/specify/specify7/blob/4e5722aabb4ae3e8afd2ee487f4bea68eaff83d1/specifyweb/stored_queries/query_construct.py#L56

CarolineDenis:
Fixes #4829
Fixes #4978 
Fixes #4979 

### Checklist

### Testing instructions
==> you need to add a new taxon tree in your db to test this PR 

Tree Viewer
- Open the taxon tree viewer 
- [ ] If the db as only one taxon tree, verify that the header is a H2 
- [ ] If the db as several taxon trees, verify that the header is a drop down
- [ ] Verify that the header drop down contain all the trees of the db 
- Change the tree type
- [ ] Verify the tree has changed 
- click on a node, add a child 
- [ ] verify that the ranks in the drop down are from the correct tree 

QB
- Open the a new Taxon query 
- [ ] Verify it is the same case when doing: New CO query > Determination > Taxon 
- [ ] Query on any rank from any tree
- [ ] Query on a rank present in 2 or more trees and verify all nodes are doing return


specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts
```diff
@@ -97,7 +97,18 @@ export const treeRanksPromise = Promise.all([
           )
       : []
   )
+  /*
```
CarolineDenis, line 100: @acwhite211

acwhite211: I changed discipline.taxontreedef from a one-to-one to many-to-one relationship, no django migration required for that, in this commit https://github.com/specify/specify7/pull/5032/commits/b7f9afb4c9ddcf63bb4471f3ba9e629db643c1b1

melton-jason:
@acwhite211 @CarolineDenis 

Here is a summary of the major changes I did

Commits:
https://github.com/specify/specify7/pull/5036/files/fc36e4c1b779e172f35010c63c5a9a1c5be11f76..dcc229a65f5af2f1797937136d8d52710c26ec70

### Backend
I have gone ahead and implemented the endpoint I suggested in https://github.com/specify/specify7/commit/a520069a159688db1210fab96c28c6390a3fba8a#commitcomment-143518914. 

The endpoint returns all of the tree-structure related information for every tree the user has read access to. 
That is, for each type of tree (Taxon, Storage, Geography, Lithostrat, Geologictimeperiod), an array of all definitions (TreeDefs) as well as all ranks (TreeDefItems) for each definition sorted by rankid. 

The frontend previously was doing a serious of requests to create an object similar to what this endpoint now returns. 

https://github.com/specify/specify7/blob/b893ef792d0d4195156569e8c29690e798ddcdc1/specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts#L58-L104

Feel free to share what you think: 
https://github.com/specify/specify7/blob/c64b8b63cf9928488f27a2644e18e7d372304fa4/specifyweb/specify/tree_views.py#L420-L444

Also did some other small refactoring where I saw fit: primarily improving typing. 

### Frontend

The frontend work consisted of using the newly defined endpoint, getting the cache for the definition names up and running, and changing the pre-existing code to be compatible with all of the changes we have done to `treeRanksPromise` (and to lay the foundations for other changes in this PR). 

Instead of having a single object in the cache which keeps track of each Tree's definition the user was last using, now we are creating a separate cache entry for each tree. This should make it simpler and easier to both access and set. 

https://github.com/specify/specify7/pull/5036/files/fc36e4c1b779e172f35010c63c5a9a1c5be11f76..dcc229a65f5af2f1797937136d8d52710c26ec70#diff-af2e7f4cf2d56aca1781676234d2c93967a908b230cc5a04eb37a1309815e2b2R83-R84

💡  Although I would like to use the definition ids rather than the name internally to keep track of the 'current' tree. If the names of the trees change, this would automatically handle the case. 

The code is completely functional right now: if the user has more than one Tree of a specific type created, they can switch which tree they are using. in the TreeView.

Other places which used the `treeDefinitions` promise and object should now also be compatible with the changes. We will just need to touch these places up when we work out resolving the 'default' Tree (currently it uses the first tree in the array of all trees).

CarolineDenis:
@melton-jason, 
Thank you for these changes, the new API makes the frontend code so much simpler

>  Although I would like to use the definition ids rather than the name internally to keep track of the 'current' tree. If the names of the trees change, this would automatically handle the case.

Yes, I think that would be good 

> resolving the 'default' Tree (currently it uses the first tree in the array of all trees).
We'll have to define where we set this default

CarolineDenis:
Need to work on the taxonomic rank pick list when adding new taxon: 
Currently the 'default' taxon tree def is used 
Screenshot 2024-06-27 at 10 13 06 AM

specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts
```diff
@@ -3,32 +3,36 @@
...
+  }>;
+};
+
+export let treeDefinitions: {
+  readonly [TREE_NAME in AnyTree['tableName']]: RA<{
+    readonly definition: SerializedResource<Tables[`${TREE_NAME}TreeDef`]>;
     readonly ranks: RA<SerializedResource<Tables[`${TREE_NAME}TreeDefItem`]>>;
-  };
+  }>;
 } = undefined!;
```
sharadsw, line 36: The type looks duplicated? How about
```
let treeDefinition: TreeInformationBackend = undefined;
```
Also, I think the type name `TreeInformationBackend` could be clearer. How about just `TreeInformation` or `TreeDefinitions`

specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts
```diff
@@ -118,20 +104,50 @@ function getTreeScope(
...
 export function getTreeDefinitionItems<TREE_NAME extends AnyTree['tableName']>(
   tableName: TREE_NAME,
-  includeRoot: boolean
-): typeof treeDefinitions[TREE_NAME]['ranks'] | undefined {
-  const definition = caseInsensitiveHash(treeDefinitions, tableName);
-  return definition?.ranks.slice(includeRoot ? 0 : 1);
+  includeRoot: boolean,
+  rawDefinitionName?: string,
+  getAllRanks?: boolean
+): typeof treeDefinitions[TREE_NAME][number]['ranks'] | undefined {
```
sharadsw, line 110: Type can be changed to `TreeInformationBackend[TREE_NAME][number]['ranks'] | undefined` ?

specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts
```diff
@@ -118,20 +104,50 @@ function getTreeScope(
...
   TREE_NAME extends AnyTree['tableName']
 >(
   tableName: TREE_NAME,
-  includeRoot: boolean
-): typeof treeDefinitions[TREE_NAME]['ranks'] =>
+  includeRoot: boolean,
+  // FIXME: replace this with the correct default tree
+  rawDefinitionName?: string,
+  getAllRanks?: boolean
+): typeof treeDefinitions[TREE_NAME][number]['ranks'] =>
```
sharadsw, line 143: Same as above. Type can be changed to avoid typeof

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -47,34 +50,47 @@ import { Tree } from './Tree';
...
+          readonly definition: SerializedResource<GeographyTreeDef>;
+          readonly ranks: RA<SerializedResource<GeographyTreeDefItem>>;
+        }>)
       : undefined;
 
-  if (treeName === undefined || !isTreeTable(treeName)) return <NotFoundView />;
-  return (
+  const [
+    currentDefinitionName = treeDefinitionArray === undefined
+      ? undefined
```
maxpatiiuk, line 70: hmm, I'm not sure this is legal js syntax - assignment to undefined is not allowed

alesan99 (changes requested):
Attempting to edit the current tree definition from the edit button in the tree viewer results in the prompt closing. Looks like typing any key in any field causes it to close.
 
https://github.com/specify/specify7/assets/8108710/8091700a-742b-4802-9406-1074b860a75b

alesan99:
Also switching the Taxon tree then switching to a different type of tree causes a crash.

https://github.com/specify/specify7/assets/8108710/79be06ed-1cd6-449b-9979-c1141c48634c

[Specify 7 Crash Report - 2024-07-08T15_40_16.041Z.txt](https://github.com/user-attachments/files/16131426/Specify.7.Crash.Report.-.2024-07-08T15_40_16.041Z.txt)

emenslin (changes requested):
### Testing instructions
==> you need to add a new taxon tree in your db to test this PR 

Tree Viewer
- Open the taxon tree viewer 
- [x] If the db as only one taxon tree, verify that the header is a H2 
- [x] If the db as several taxon trees, verify that the header is a drop down
- [x] Verify that the header drop down contain all the trees of the db 
- Change the tree type
- [x] Verify the tree has changed 
- click on a node, add a child 
- [x] verify that the ranks in the drop down are from the correct tree 


I was however able to recreate Alejandro's issues with editing the title and the error with switching trees 


QB
- Open the a new Taxon query 
- [ ] Verify that all ranks from all tree are present 
- [x] Verify the ranks are separated with a sub header for each tree 
- [x] Verify it is the same case when doing: New CO query > Determination > Taxon 

All trees were not present. In the video as you can see there are 3 trees but in the QB there are only 2 sub headers. 

https://github.com/specify/specify7/assets/150940840/5915d059-2ec8-4824-98c7-0e78d9b340db

---

I know that the QB scoping is not done yet but these were my initial thoughts so I wanted to include it in my review even if it might become obsolete with the future changes 

Along with that I also think the subheader layout in the queries is not the best choice. I feel like they blend in and when there are several ranks for the trees (like the invertebrate taxonomy tree in the video) it is a lot of scrolling to fond the other tree, I can only imagine how much scrolling it would be for db with more trees that are even larger. 

I suggest some sort of drop down menu (like a picklist) to be able to select the specific tree so you don't have to scroll as much.

specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts
```diff
@@ -116,22 +104,65 @@ function getTreeScope(
...
 export function getTreeDefinitionItems<TREE_NAME extends AnyTree['tableName']>(
   tableName: TREE_NAME,
-  includeRoot: boolean
-): typeof treeDefinitions[TREE_NAME]['ranks'] | undefined {
-  const definition = caseInsensitiveHash(treeDefinitions, tableName);
-  return definition?.ranks.slice(includeRoot ? 0 : 1);
+  includeRoot: boolean,
+  definitionFilter?: Partial<
+    SerializedResource<FilterTablesByEndsWith<'TreeDef'>>
+  >
```
maxpatiiuk, line 132: idea: instead of this filter could we simply have `treeDefinitionId: number | undefined`?

pros:
- it's requires so this would avoid bugs when the caller of this function forgets to provide it
- simplifies the code - we can add back complexity when if it would be needed

same goes for getTreeDefinitions

specifyweb/frontend/js_src/lib/components/Toolbar/TreeRepair.tsx
```diff
@@ -92,16 +93,33 @@ export function TreeSelectDialog({
...
+                const treeDefinition = deserializeResource(
+                  getTreeDefinitions(
+                    treeName,
+                    cachedDefinitionId === undefined
+                      ? undefined
+                      : {
+                          id: cachedDefinitionId,
+                        }
+                  )[0].definition
+                );
```
maxpatiiuk, line 110: doing deserializeResource on each render may be a bit wasteful
could you put lines 86-110 into a useMemo?

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
```diff
@@ -54,7 +57,7 @@ export function Tree<SCHEMA extends AnyTree>({
   readonly treeDefinitionItems: RA<
     SerializedResource<FilterTablesByEndsWith<'TreeDefItem'>>
   >;
-  readonly tableName: SCHEMA['tableName'];
+  readonly tableName: TREE_NAME;
```
maxpatiiuk, line 60: hmm, could you clarify why this change was made?

specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts
maxpatiiuk: even better, you can have "getCache('tree', `definition${tableName}`)," be the default value for treeDefinitionId, thus making the default "just work"

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -225,11 +255,27 @@ function TreeView<SCHEMA extends AnyTree>({
     <Container.Full>
       <header className="flex items-center gap-2 overflow-x-auto sm:flex-wrap sm:overflow-x-visible">
         <TableIcon label name={table.name} />
-        <H2 title={treeDefinition.get('remarks') ?? undefined}>
-          {treeDefinition.get('name')}
-        </H2>
+        {definitions.length > 1 ? (
+          <Select
```
maxpatiiuk, line 259: need aria-label and title attributes

specifyweb/frontend/js_src/lib/components/TreeView/index.tsx
```diff
@@ -225,11 +255,27 @@ function TreeView<SCHEMA extends AnyTree>({
...
+            ))}
+          </Select>
+        ) : (
+          <H2 title={(treeDefinition.remarks as string) ?? undefined}>
+            {treeDefinition.name as string}
+          </H2>
+        )}
         <ResourceEdit
-          resource={treeDefinition}
+          resource={deserializeResource(treeDefinition)}
```
maxpatiiuk, line 278: this will create a new object on each render, which is possible going to break ResourceEdit, or cause performance issues - put this into a useMemo?

specifyweb/frontend/js_src/lib/components/WbPlanView/CustomSelectElement.tsx
```diff
@@ -682,9 +682,7 @@ export function CustomSelectElement({
             {...selectGroupData}
             hasArrow={has('arrow')}
             hasIcon={has('icon')}
-            selectGroupLabel={
-              customSelectSubtype === 'simple' ? selectGroupLabel : undefined
-            }
+            selectGroupLabel={selectGroupLabel}
```
maxpatiiuk, line 685: looks good, but could you double check that the UI impact of this on the workbench mapper and query builder is ok?

specifyweb/frontend/js_src/lib/components/WbPlanView/LineComponents.tsx
```diff
@@ -37,6 +37,7 @@ export type HtmlGeneratorFieldData = {
   readonly isDefault?: boolean;
   readonly isRelationship?: boolean;
   readonly tableName?: keyof Tables;
+  readonly tableTreeDefName?: string;
 };
```
maxpatiiuk, line 41: (off topic) haha, the name of this type `HtmlGeneratorFieldData`, is from the dark days before React when this was fed into a function that spit out HTML strings 🥲
we should probably rename it

specifyweb/frontend/js_src/lib/tests/ajax/static/api/specify_trees.json
maxpatiiuk, line 1: would it make sense to include in the test data at least one case where tree has multiple definitions so that such case can be covered by tests?

specifyweb/frontend/js_src/lib/utils/cache/definitions.ts
```diff
@@ -80,6 +80,8 @@ export type CacheDefinitions = {
   } & {
     readonly /** Open nodes in a given tree */
     [key in `conformations${AnyTree['tableName']}`]: Conformations;
+  } & {
+    readonly [key in `definition${AnyTree['tableName']}`]: number;
```
maxpatiiuk, line 84: sorry if I missed it, but I don't see where this cache entry is set - I only see it being read
is the code for setting it not developed yet?
and how is the case when this isn't set yet handled (first start of specify in incognito tab)

specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts
```diff
@@ -61,47 +61,35 @@ export const treeRanksPromise = Promise.all([
...
+      readonly ranks: RA<SerializedRecord<Tables[`${TREE_NAME}TreeDefItem`]>>;
+    }>;
+  }>('/api/specify_trees/', {
+    headers: { Accept: 'application/json' },
+  }).then(({ data }) => {
+    disciplineTrees = allTrees.filter((treeName) =>
+      Object.keys(data)
+        .map((key) => key.toLowerCase())
+        .includes(treeName.toLowerCase())
+    );
```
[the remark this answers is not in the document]
CarolineDenis: first one is safer

specifyweb/frontend/js_src/lib/components/WbPlanView/CustomSelectElement.tsx
CarolineDenis: yep verified

specifyweb/frontend/js_src/lib/components/WbPlanView/LineComponents.tsx
CarolineDenis: will create a new issue for it

specifyweb/frontend/js_src/lib/tests/ajax/static/api/specify_trees.json
CarolineDenis: yes it would

specifyweb/frontend/js_src/lib/utils/cache/definitions.ts
CarolineDenis: it's set when the user changes the taxon tree in the tree viewer

CarolineDenis: it will be the first available for now and we will add some new logic and a new way to define the default tree that is being used

specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts
melton-jason: > idea: instead of this filter could we simply have treeDefinitionId: number | undefined?

yes, I agree in nearly cases just `treeDefinitionId: number | undefined` would be better. And in all uses, we do just use the id. 

I had assumed that some point in the future we may need to return a list of certain TreeDefItems or TreeDefs fitting some criteria, but that is probably not that likely (and the caller can do so anyways if needed)

melton-jason: > even better, you can have "getCache('tree', definition${tableName})," be the default value for treeDefinitionId, thus making the default "just work"

Did you have something like this in mind? 
https://github.com/specify/specify7/pull/5036/commits/82fb29d8477e64e2b9c659a86c3258ff22549ceb

maxpatiiuk: looks great! that reduced a bit of complexity

specifyweb/frontend/js_src/lib/components/Leaflet/localityRecordDataExtractor.ts
```diff
@@ -136,7 +136,8 @@ async function recursiveResourceResolve(
       return [];
     const tableRanks = strictGetTreeDefinitionItems(
       treeTableName as 'Geography',
-      false
+      false,
+      undefined
```
maxpatiiuk, line 140: passing undefined here seems needless as that would be the default behavior when attribute is not provided

is your intention to not-apply the default value? default value still applies if the passed attribute is undefined

specifyweb/frontend/js_src/lib/components/Toolbar/TreeRepair.tsx
```diff
@@ -67,6 +66,67 @@ export function TreeSelectDialog({
   const [treeRanks] = usePromise(treeRanksPromise, true);
   const [isFinished, setIsFinished] = useBooleanState();
 
+  const treeList = React.useMemo(
```
maxpatiiuk, line 69: this useMemo is useless right now because a new `getLint` function is passed on each render
could you please refactor to only have this part in the useMemo as this part is less likely to change often:

```ts
      getDisciplineTrees()
        .filter((treeName) =>
          permissionName === 'repair'
            ? hasPermission(`/tree/edit/${toLowerCase(treeName)}`, 'repair')
            : hasTreeAccess(treeName, 'read')
        )
        .map((treeName) => {
          const treeDefinition = deserializeResource(
            getTreeDefinitions(treeName)[0].definition
          );
          return [treeName, treeDefinition] as const;
      }
```

and outside of the useMemo you would map over the results to render the actual elements

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/Leaflet/localityRecordDataExtractor.ts
maxpatiiuk: yes, this might be one of the only few good reasons to use `null`
i.e `undefined` could be though of as `I provide no value - feel free to use the default`
where as `null` is `I explicitly tell you that there is no value!`
but your idea of using `'all'` is much better as it's more readable than seeing `null`

specifyweb/frontend/js_src/lib/components/InitialContext/treeRanks.ts
```diff
@@ -116,22 +106,61 @@ function getTreeScope(
   );
 }
 
+export function getTreeDefinitions<TREE_NAME extends AnyTree['tableName']>(
+  tableName: TREE_NAME,
+  treeDefinitionId: number | undefined | 'all' = getCache(
+    'tree',
+    `definition${tableName}`
+  )
+): typeof treeDefinitions[TREE_NAME] {
```
sharadsw, line 115: (optional) Would `TreeInformation[TREE_NAME]` work here instead? If so, there's a couple more places I noticed with `typeof treeDefinitions` that can be replaced.

sharadsw (approved):

melton-jason (approved):
This specific feature is functional and the code looks okay, so I will approve this. 

HOWEVER, we are far from done with tree changes, although most of the changes will be hard to undertake before #5016 is implemented. 

I know there is a separate PR already open to handle the WorkBench side of this equation: #5091. 

I am not a big fan of how the tree ranks are currently displayed in the WorkBench/QueryBuilder mapper. This can probably be improved (later if needed). 

Screenshot 2024-07-16 at 10 58 38 AM

(Although currently you can not even Query on ranks from other Trees without an error). 

Here are some major places (on the backend) which need modifications once #5016 (or its equivalent) is available: 

QueryBuilder: 
https://github.com/specify/specify7/blob/df4eccdebf7a6b2380bb4345178d83ff3f342e76/specifyweb/stored_queries/query_construct.py#L13-L16

WorkBench: 
https://github.com/specify/specify7/blob/df4eccdebf7a6b2380bb4345178d83ff3f342e76/specifyweb/workbench/upload/scoping.py#L186-L205

Scoping: 
https://github.com/specify/specify7/blob/df4eccdebf7a6b2380bb4345178d83ff3f342e76/specifyweb/specify/filter_by_col.py#L27-L34

https://github.com/specify/specify7/blob/df4eccdebf7a6b2380bb4345178d83ff3f342e76/specifyweb/stored_queries/execution.py#L53-L91

Taxon Tiles: 
https://github.com/specify/specify7/blob/df4eccdebf7a6b2380bb4345178d83ff3f342e76/specifyweb/barvis/views.py#L15-L49

Some of the frontend work is done in this PR, we just need to decide how we want to resolve which tree to interact with within the UI. 
(i.e., when using QueryComboBox, Workbench/QueryBuilder Mapper, Data Entry form, etc.)

Nice work so far!

acwhite211:
Using QB with Taxon seems to be working well now.
Need to double check the logic for getting tree definitions in this file:
https://github.com/specify/specify7/blob/issue-4829/specifyweb/specify/tree_utils.py

specifyweb/specify/tree_utils.py
melton-jason, line 1: I don't think there should be as much reliance on CollectionObjectType for scoping tree records. 

I think Taxon records should first still be scoped to discipline. 
That is I think `/api/specify/taxon/?domainfilter=true` should return all Taxon records across all TreeDefs defined in the Discipline, rather than just returning only Taxon records associated with the `Collection.CollectionObjectType.taxonTreeDef`. 
The same applies for QueryBuilder searches.

The consumer of the API should then further filter results as needed.

melton-jason (changes requested):

specifyweb/stored_queries/query_construct.py
```diff
@@ -4,17 +4,12 @@
...
+from specifyweb.stored_queries import models
 
 logger = logging.getLogger(__name__)
 
-
-def get_treedef(collection, tree_name):
-    return (collection.discipline.division.institution.storagetreedef
-            if tree_name == 'Storage' else
-            getattr(collection.discipline, tree_name.lower() + "treedef"))
-
```
realVinayak, line 17: I don't think what you have will work. IIRC, we were going to support queries to ALL trees, correct? So if you get "species", it'll return species in Mineral and Fossil tree. This doesn't do that...

Previously, I mentioned adding that COT in tree stringid. We don't technically need that anymore (if we're searching for all trees), but what you have wouldn't work. For instance, try searching for a rank = 'blahh'. you won't get the results for _all_ the trees. 

I'll recommend looking at how it constructs tree joins to figure out why. You even have a debugger!

realVinayak: You _really_ need to lookup all the treedefids, and then get the most depth here (https://github.com/specify/specify7/blob/dff3564be2f8e4b5a5a60ba761f2cbac995801e8/specifyweb/stored_queries/query_construct.py#L43), and then look at ALL the treedefitemids (https://github.com/specify/specify7/blob/dff3564be2f8e4b5a5a60ba761f2cbac995801e8/specifyweb/stored_queries/query_construct.py#L56) that match a particular rank

Basically, that should be an OR (or an IN, doesn't matter. So, use OR -- more efficient)

CarolineDenis: > we were going to support queries to ALL trees, correct? So if you get "species", it'll return species in Mineral and Fossil tree. This doesn't do that...

correct

realVinayak: 👍, the issue I mentioned should be verified and resolved then. I can't set it up on testpanel due to this [commit](https://github.com/specify/specify7/actions/runs/10115757864/job/27977226991) presumably

specifyweb/barvis/views.py
```diff
@@ -1,29 +1,49 @@
...
+                Determination.objects.filter(taxon_id=OuterRef("id"), iscurrent=True)
+                .values("taxon_id")
+                .annotate(count=Count("*"))
+                .values("count"),
+                output_field=IntegerField(),
+            )
+        )
+        .values("id", "rankid", "parent_id", "name", "current_determination_count")
+    )
+    result = toJson(list(taxons))
```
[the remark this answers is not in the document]
realVinayak: before that, test it on a big tree to see if the json response's string causes memory issues, not optional. we already had oom issues due to oversights in tree sizes

specifyweb/barvis/views.py
```diff
@@ -1,29 +1,49 @@
...
+    taxons = (
+        Taxon.objects.filter(definition__in=taxon_tree_defs)
+        .annotate(
+            current_determination_count=Subquery(
+                Determination.objects.filter(taxon_id=OuterRef("id"), iscurrent=True)
+                .values("taxon_id")
+                .annotate(count=Count("*"))
+                .values("count"),
+                output_field=IntegerField(),
+            )
```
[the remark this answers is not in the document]
realVinayak: > This Subquery returns None

Are you sure? It should all be just one query, and you have an aggregate there. Why would it return none?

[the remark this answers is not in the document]
melton-jason: > can you send responses with None?
Sorry, I didn't mean the actual response being returned was None if that is what you are referring to. I meant that the `current_determination_count` was NULL occasionally for the rows in the response. 

I am not sure why the Subquery that Alec had here returned None in some cases, but it does.  

```py
current_determination_count=Subquery(
                Determination.objects.filter(taxon_id=OuterRef("id"), iscurrent=True)
                .values("taxon_id")
                .annotate(count=Count("*"))
                .values("count"),
                output_field=IntegerField(),
            )
```

Here is the generated SQL of the Query. 

```sql
SELECT `taxon`.`taxonid`, `taxon`.`RankID`, `taxon`.`ParentID`, `taxon`.`Name`, (SELECT COUNT(*) AS `count` FROM `determination` U0 WHERE (U0.`IsCurrent` AND U0.`TaxonID` = `taxon`.`taxonid`) GROUP BY U0.`TaxonID` ORDER BY NULL) AS `current_determination_count` FROM `taxon` WHERE `taxon`.`TaxonTreeDefID` IN (SELECT V0.`taxontreedefid` FROM `taxontreedef` V0 WHERE V0.`taxontreedefid` IN (SELECT DISTINCT U0.`TaxonTreeDefID` FROM `collectionobjecttype` U0 WHERE U0.`CollectionID` = 4))
```

Looks like the GROUP BY in the Subquery might be causing these NULL results for `current_determination_count`?

realVinayak:
See https://github.com/specify/specify7/pull/5151/

specifyweb/specify/tree_utils.py
```diff
@@ -27,6 +27,14 @@ def get_treedef(collection: spmodels.Collection, tree_name: str) ->  List[Tuple[
     )
 
     result = list(search_query)
+
+    # Handle null taxontreedef.discipline
+    if len(result) == 0 and tree_name.lower() == 'taxon':
+        result = list(spmodels.CollectionObjectType.objects\
+            .filter(collection=collection)\
+            .annotate(item_counts=Count('taxontreedef__treedefitems'))\
+            .values_list('taxontreedef_id', 'item_counts'))
```
realVinayak, line 36: Hmm should trees corresponding to COT always be searched? If yes, then this should be refactored to always include them (not just when result is empty). You can do quite a different ways of doing this. I think the "correct" way would be to modify `get_search_filters` to always return a Q() object. For taon trees, do `Q(discipline=collection.discipline) | Q(cotypes__collection=collection)`. See model definition: https://github.com/specify/specify7/blob/0557807fea85d44019919a4d9368a98065e88bbd/specifyweb/specify/models.py#L7509

realVinayak (dismissed):

specifyweb/specify/tree_utils.py
realVinayak: Plus, if you do this here, it'll automatically resolve a possible bug when scoping..

realVinayak: Resolved in https://github.com/specify/specify7/pull/5036/commits/b62b7353e0e0ede3ee993edc46473eb1dde15cbe

specifyweb/specify/tree_utils.py
realVinayak: Currently, we rely on both. See latest changes to those. There were some reported issues where treedefs didn't have discipline set (but did have cot correctly set, and thus needed to be included). In practise (assuming the migrations set-up the def id correctly https://github.com/specify/specify7/pull/5036/commits/b4cca060dc93a6f126e4e85327b5dae0a52934a9), I don't suspect it to be an issue

specifyweb/specify/migrations/0002_geo.py
```diff
@@ -67,7 +68,10 @@ def create_default_collection_types():
         # Update CollectionObjects' collectionobjecttype for the discipline
         Collectionobject.objects.filter(collection=collection).update(collectionobjecttype=cot)
         collection.collectionobjecttype = cot
-        collection.save()
+        try:
+            collection.save()
+        except BusinessRuleException as e:
+            continue
```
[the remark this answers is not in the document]
acwhite211: Ya, I ran into this error for one of the databases.  I'll try to recreate it again, I believe it was in the hujinnhc test panel database migration.  A complete solution that should be done is to handle and resolve the business rule exception so that the collection field is not null.

realVinayak:
@CarolineDenis, @grantfitzsimmons, how should the stats page be changed to account for multiple trees? It should currently display all the tree ranks in a single category. Nothing for old stats should break

acwhite211 (approved):

acwhite211:
@specify/ux-testing this PR is ready for testing.

emenslin (changes requested):
### Testing instructions
==> you need to add a new taxon tree in your db to test this PR 

Tree Viewer
- Open the taxon tree viewer 
- [ ] If the db as only one taxon tree, verify that the header is a H2 
- [ ] If the db as several taxon trees, verify that the header is a drop down
- [ ] Verify that the header drop down contain all the trees of the db 
- Change the tree type
- [ ] Verify the tree has changed 
- click on a node, add a child 
- [ ] verify that the ranks in the drop down are from the correct tree 

QB
- Open the a new Taxon query 
- [ ] Verify it is the same case when doing: New CO query > Determination > Taxon 
- [ ] Query on any rank from any tree
- [ ] Query on a rank present in 2 or more trees and verify all nodes are doing return 

Cannot open trees and queries with taxon don't work as expected, tried on kuentomology, SAIAB, and hujinnhc

https://github.com/user-attachments/assets/c1f5b8eb-40b4-44fa-959f-19fb53c74bcd


[Specify 7 Crash Report - 2024-07-31T13_48_08.648Z.txt](https://github.com/user-attachments/files/16442668/Specify.7.Crash.Report.-.2024-07-31T13_48_08.648Z.txt)



https://github.com/user-attachments/assets/2b4f79d4-50a0-41a5-b53a-dbb9e8bf3e3f



https://github.com/user-attachments/assets/b388d819-6911-48a3-a1fe-105b7bf20cf4


[Specify 7 Crash Report - 2024-07-31T13_49_41.264Z.txt](https://github.com/user-attachments/files/16442691/Specify.7.Crash.Report.-.2024-07-31T13_49_41.264Z.txt)

realVinayak:
@emenslin , or I guess @specify/ux-testing, clear your cache out before testing

realVinayak:
actually, this'll be in production too.....

emenslin:
> @emenslin , or I guess @specify/ux-testing, clear your cache out before testing

I did clear my cache and it's still happening

emenslin:
Still not working, I get this error once logged into the db 
Screenshot 2024-07-31 101224


[Specify 7 Crash Report - 2024-07-31T15_12_25.055Z.txt](https://github.com/user-attachments/files/16443894/Specify.7.Crash.Report.-.2024-07-31T15_12_25.055Z.txt)

realVinayak:
Sorry, I didn't get to test it locally before pushing. Resolved now.

emenslin (approved):
### Testing instructions
==> you need to add a new taxon tree in your db to test this PR 

Tree Viewer
- Open the taxon tree viewer 
- [x] If the db as only one taxon tree, verify that the header is a H2 
- [x] If the db as several taxon trees, verify that the header is a drop down
- [x] Verify that the header drop down contain all the trees of the db 
- Change the tree type
- [x] Verify the tree has changed 
- click on a node, add a child 
- [x] verify that the ranks in the drop down are from the correct tree 

QB
- Open the a new Taxon query 
- [x] Verify it is the same case when doing: New CO query > Determination > Taxon 
- [x] Query on any rank from any tree
- [x] Query on a rank present in 2 or more trees and verify all nodes are doing return 

Everything looks good from what I can tell, did not run into any of the previous errors this time

Areyes42 (approved):
### Testing instructions

\==> you need to add a new taxon tree in your db to test this PR

Tree Viewer

-   Open the taxon tree viewer
-   [x] If the db as only one taxon tree, verify that the header is a H2
-   [x] If the db as several taxon trees, verify that the header is a drop down
-   [x] Verify that the header drop down contain all the trees of the db
-   Change the tree type
-   [x] Verify the tree has changed
-   click on a node, add a child
-  [x]  verify that the ranks in the drop down are from the correct tree

QB

-   Open the a new Taxon query
-   [x] Verify it is the same case when doing: New CO query > Determination > Taxon
-   [x] Query on any rank from any tree
-   [x] Query on a rank present in 2 or more trees and verify all nodes are doing return
***
Tree viewer and QB looks good across multiple different DBs that I tested with. Didn't come across any major issues 👍

lexiclevenger (approved):
### Testing instructions
==> you need to add a new taxon tree in your db to test this PR 

Tree Viewer
- Open the taxon tree viewer 
- [x] If the db as only one taxon tree, verify that the header is a H2 
- [x] If the db as several taxon trees, verify that the header is a drop down
- [x] Verify that the header drop down contain all the trees of the db 
- Change the tree type
- [x] Verify the tree has changed 
- click on a node, add a child 
- [x] verify that the ranks in the drop down are from the correct tree 

QB
- Open the a new Taxon query 
- [x] Verify it is the same case when doing: New CO query > Determination > Taxon 
- [x] Query on any rank from any tree
- [x] Query on a rank present in 2 or more trees and verify all nodes are doing return 

---

Looks good, everything seems to be working as expected. 👍

realVinayak:
@acwhite211 could have at least tried to add unit tests 🤦

specifysoftware:

PR #5042, opened 2024-06-25
Block save when ExportFeed has invalid user id

### LINKED ISSUE #4955: RSS Export Feed directs to an Error 404 page if `userId` or `notifyUser` values are invalid
**Describe the bug**
When you put in an invalid ID into the `userId` or `notifyUser` fields for an RSS Export Feed using the XML editor, opening it again afterwards directs to an error 404 page.

**To Reproduce**

1. Go to App Resources -> Create a new/open an RSS Export Feed -> Switch to XML Editor.
2. Change the `userId` or `notifyUser` values to an invalid one (one that does not match any Specify User) and save.
3. Navigate away from the RSS Feed and then reopen it.
4. See 404 error.

**Expected behavior**
There should not be an error 404 page, since this prevents even being able to fix the RSS feed.

**Screenshots**

https://github.com/specify/specify7/assets/71564869/c3d4b821-d021-45c9-af52-a88d0fb34ed6

- OS: Windows 10
- Browser: Firefox
- Specify 7 Version: `edge`
- Collection name: `Herp-PC`
- User Name: `testuser`

**Additional context**
This is a new issue due to the visual RSS feed; it's likely trying to find a user that doesn't exist for the "Run as user" and "Send completion notification to user" fields, and since it cannot, it goes to an error 404.

If you need to delete the files for testing purposes, you can switch the database to `v7.9.4` or do it from the back end.

sharadsw:
Fixes #4955 

NOTES:
The 404 page occurs due to the way our ajax fetch is written. The QCBX in `UserPicker` fetches the resource at the specified resource url. When the url is invalid, ajax receives a 404 error which is handled by rendering a 404 page in production
https://github.com/specify/specify7/blob/013f81295c9e83cd6bd86a565d9b21950b5a76da/specifyweb/frontend/js_src/lib/components/Errors/FormatError.tsx#L152-L162

My workaround for this is to fetch the resource in `UserPicker` using `fetchResource` with `strict = false` (to avoid a 404 page) and if the url is invalid, we clear the QCBX and block save.

### Checklist

- [ ] Add automated tests

### Testing instructions

- Go to App Resources -> Create a new/open an RSS Export Feed -> Switch to XML Editor.
- Change the userId or notifyUser values to an invalid one (one that does not match any Specify User) and save.
- [ ] Verify RSS Export Feed can be opened and there is no 404 page
- [ ] Verify userId/notifyUserId QCBX is cleared and save is blocked in the visual editor
- [ ] Verify the user can be changed and the Export Feed config can be saved

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/ExportFeed/Editor.tsx
+++ b/specifyweb/frontend/js_src/lib/components/ExportFeed/Editor.tsx
@@ -20,10 +20,12 @@ import { icons } from '../Atoms/Icons';
 import { ReadOnlyContext } from '../Core/Contexts';
 import { getField } from '../DataModel/helpers';
 import {
+  fetchResource,
   getResourceApiUrl,
   idFromUrl,
   resourceOn,
 } from '../DataModel/resource';
+import { getFieldBlockerKey, useSaveBlockers } from '../DataModel/saveBlockers';
 import { tables } from '../DataModel/tables';
 import { CollectionPicker } from '../Header/ChooseCollection';
 import { AutoGrowTextArea } from '../Molecules/AutoGrowTextArea';
@@ -360,14 +362,31 @@ function UserPicker({
   readonly isRequired: boolean;
 }): JSX.Element {
   const resource = React.useMemo(() => new tables.Agent.Resource(), []);
-  React.useEffect(
-    () =>
-      void resource.set(
-        'specifyUser',
-        id === undefined ? null : getResourceApiUrl('SpecifyUser', id)
-      ),
-    [resource, id]
-  );
+  const field = getField(tables.Agent, 'specifyUser');
+
+  const [_, setBlockers] = useSaveBlockers(resource, field);
+
+  React.useEffect(() => {
+    void resource.set(
+      'specifyUser',
+      id === undefined ? null : getResourceApiUrl('SpecifyUser', id)
+    );
+    if (id === undefined) return;
+    // Fetch resource to check for invalid SpecifyUser id entered through XML editor
+    fetchResource('SpecifyUser', id, false).then((res) => {
+      if (res?.id !== id)
+        void resource.set(
+          'specifyUser',
+          res === undefined ? null : getResourceApiUrl('SpecifyUser', id)
+        );
+      if (res === undefined)
+        setBlockers(
+          [resourcesText.invalidSpecifyUser()],
+          getFieldBlockerKey(field, 'invalidSpecifyUser')
+        );
+      else setBlockers([], getFieldBlockerKey(field, 'invalidSpecifyUser'));
+    });
+  }, [id]);
 
   const setIdRef = React.useRef(setId);
   setIdRef.current = setId;
@@ -384,7 +403,7 @@ function UserPicker({
 
   return (
     <QueryComboBox
-      field={getField(tables.Agent, 'specifyUser')}
+      field={field}
       forceCollection={undefined}
       formType="form"
       id={undefined}
--- a/specifyweb/frontend/js_src/lib/localization/resources.ts
+++ b/specifyweb/frontend/js_src/lib/localization/resources.ts
@@ -840,6 +840,6 @@ export const resourcesText = createDictionary({
     'uk-ua': 'Потрібне поточне визначення.',
   },
   invalidSpecifyUser: {
-    'en-us': 'Invalid specify user'
-  }
+    'en-us': 'Invalid specify user',
+  },
 } as const);
--- a/specifyweb/frontend/js_src/lib/localization/resources.ts
+++ b/specifyweb/frontend/js_src/lib/localization/resources.ts
@@ -839,4 +839,7 @@ export const resourcesText = createDictionary({
     'ru-ru': 'Требуется текущее определение.',
     'uk-ua': 'Потрібне поточне визначення.',
   },
+  invalidSpecifyUser: {
+    'en-us': 'Invalid specify user'
+  }
 } as const);


specifyweb/frontend/js_src/lib/components/ExportFeed/Editor.tsx
```diff
@@ -360,14 +362,31 @@ function UserPicker({
...
+          res === undefined ? null : getResourceApiUrl('SpecifyUser', id)
+        );
+      if (res === undefined)
+        setBlockers(
+          [resourcesText.invalidSpecifyUser()],
+          getFieldBlockerKey(field, 'invalidSpecifyUser')
+        );
+      else setBlockers([], getFieldBlockerKey(field, 'invalidSpecifyUser'));
+    });
+  }, [id]);
```
maxpatiiuk, line 389: to be safe, if user id changed, you should disregard the previous request, rather than set a blocker based on a stale value

```suggestion
    fetchResource('SpecifyUser', id, false).then((res) => {
      if(destructorCalled) return;
      if (res?.id !== id)
        void resource.set(
          'specifyUser',
          res === undefined ? null : getResourceApiUrl('SpecifyUser', id)
        );
      if (res === undefined)
        setBlockers(
          [resourcesText.invalidSpecifyUser()],
          getFieldBlockerKey(field, 'invalidSpecifyUser')
        );
      else setBlockers([], getFieldBlockerKey(field, 'invalidSpecifyUser'));
    });
    let destructorCalled = false;
    return ():void => {
      destructorCalled = true
    };
  }, [id]);
```

specifyweb/frontend/js_src/lib/localization/resources.ts
```diff
@@ -839,4 +839,7 @@ export const resourcesText = createDictionary({
     'ru-ru': 'Требуется текущее определение.',
     'uk-ua': 'Потрібне поточне визначення.',
   },
+  invalidSpecifyUser: {
+    'en-us': 'Invalid specify user',
```
maxpatiiuk, line 843: capitalize Specify
even better - don't hardcode table names in the UI localization - table names should be coming from schema localization

see how we do that in other places:
https://github.com/specify/specify7/blob/00f63833a3ec941907f5e2ac66fc8c3407976c45/specifyweb/frontend/js_src/lib/localization/forms.ts#L131-L140
and then search for usages of paleoRequiresGeography in the code to see how it get's called - do something similar here too if possible

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/ExportFeed/Editor.tsx
maxpatiiuk: Now that I think of it, instead of writing this code here to handle this one case, I think we should make QueryComboBox itself add the save blocker if resource is not found - that way, this issue is fixed not just for this one place, but for every place where QueryComboBox is used

specifyweb/frontend/js_src/lib/localization/resources.ts
sharadsw: I think we can just reuse this string instead of using table names. What do you think?
https://github.com/specify/specify7/blob/b893ef792d0d4195156569e8c29690e798ddcdc1/specifyweb/frontend/js_src/lib/localization/forms.ts#L604-L611

specifyweb/frontend/js_src/lib/components/ExportFeed/Editor.tsx
sharadsw: That would be ideal. It was my initial thought as well but it's hard to avoid the 404 page due to the way our ajax fetch is written. Using `fetchResource` with `strict = false` lets us get away with the 404 but `QueryComboBox` does not use it. I think it uses `rgetPromise` instead? I briefly looked into refactoring it to use a similar `strict` param but couldn't figure it out.

I think the current approach is fine as I don't think there are any other places where QCBX can have an invalid resource. If we run into more cases like these we can look into modifying QCBX to handle it.

maxpatiiuk: look for usages of hijackBackboneAjax in the code to see how you can make this work with rgetPromise

maxpatiiuk: > any other places where QCBX can have an invalid resource

that could be any place if the database was edited by hand, or migration didn't complete correctly

sharadsw: I tried it with `hijackBackbone`. I'm able to avoid the 404 and clear the QCBX but it does not trigger an unload protect and the save blocker doesn't show up until you click on the field. I'm gonna open a different PR for that approach so that we at least have this working solution as a hotfix if needed

maxpatiiuk (approved):
accepting this as a hotfix, but please follow up with a solution where this logic is moved into query combo box
I can help you with the save blocker issue you are seeing if needed

combs-a (changes requested):
- Go to App Resources -> Create a new/open an RSS Export Feed -> Switch to XML Editor.
- Change the userId or notifyUser values to an invalid one (one that does not match any Specify User) and save.
- [X] Verify RSS Export Feed can be opened and there is no 404 page
- [X] Verify userId/notifyUserId QCBX is cleared and save is blocked in the visual editor
- [X] Verify the user can be changed and the Export Feed config can be saved

Looks good! Save button isn't red when save is invalid, but it does still block. It should be changed so that it matches other save buttons

image

lexiclevenger (approved):
### Testing instructions



- Go to App Resources -> Create a new/open an RSS Export Feed -> Switch to XML Editor.
- Change the userId or notifyUser values to an invalid one (one that does not match any Specify User) and save.
- [x] Verify RSS Export Feed can be opened and there is no 404 page
- [x] Verify userId/notifyUserId QCBX is cleared and save is blocked in the visual editor
- [x] Verify the user can be changed and the Export Feed config can be saved
---
Looks good! I don't see any issues other than the save button mentioned above.
Screenshot 2024-07-25 155523

combs-a (approved):
Saveblocker is appearing now, looks good!

image

CarolineDenis (approved):

specifysoftware:

Issue #23, opened 2012-05-02
Add support for customizing field formats


benanhalt:
This is supported by the thick client and isn't needed for the foreseeable future.

grantfitzsimmons:
Part of #2932, not something most 7 users can do easily anymore since the 'thick client' is being used less and less

grantfitzsimmons:
We need to add a visual editor for customizing this app resource:
https://github.com/specify/specify6/blob/master/config/backstop/uiformatters.xml

PR #5044, opened 2024-06-25
Handle scoping for Picklists based on Field in Table

melton-jason:
Fixes #4989

The Issue here was that for Picklists based on field values for a table, the frontend was always trying to enforce scoping for for records of the table the "fictitious" Picklist Items were based off of. 

See https://github.com/specify/specify7/issues/4989#issuecomment-2151059896
Related:  #3564

### Checklist

- [ ] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [ ] Add automated tests
- [ ] Add relevant issue to release milestone

### Testing instructions
- Create a PickList with type "Field From Table" 
- Choose a Table which can not be scoped
  - Some examples include: 
    - AccessionAuthorization
    - CollectionRelationship
    - CollectionRelType
    - Disposal
    - SpQuery
    - TaxonTreeDefItem  
- Choose any field from the Table
- In any field on any table in the Schema Configuration, assign the PickList to the field
  - [Related Speciforum Docs](https://discourse.specifysoftware.org/t/editing-forms-in-specify-7/1557#combobox-32)  
- [ ] Navigate to any form of the table which contains the field assigned to the PickList and ensure no error is thrown

Issue #4989, opened 2024-06-05
`HierarchyException` when loading Collectionreltype


grantfitzsimmons:
**Describe the bug**
When displaying the `collectionRelType` field as a query combo box, the user is presented with the following error:

```
HierarchyException at /api/specify_rows/collectionreltype/ queryset model 
Collectionreltype has no hierarchy field
```

This leads to the form being unable to load and the user no longer being able to use the application.

<img width="618" alt="image" src="https://github.com/specify/specify7/assets/37256050/9fc38af3-8997-46d8-871f-2299de4f601d">

**Form XML**
```
<cell type="field" id="collectionRelType" name="collectionRelType" uitype="querycbx" initialize="name=CollectionRelType;isrequired=true;newbtn=false;editbtn=false" isrequired="true"/>
```

**To Reproduce**

1. Go to https://mpabi20240604-edge.test.specifysystems.org/specify/view/collectionobject/2135/?recordsetid=41
2. Wait for the record to load
3. See crash

**Screenshots**

Left is `[v7.9.3.1](https://mpabi20240604-v793.test.specifysystems.org/specify/view/collectionobject/2135/?recordsetid=41)` and right is `[edge](https://mpabi20240604-edge.test.specifysystems.org/specify/view/collectionobject/2135/?recordsetid=41)` (during `v7.9.6` testing)

<img width="1510" alt="image" src="https://github.com/specify/specify7/assets/37256050/159c2a55-121e-4016-8634-2063523fcd78">

**Crash Report**
[Specify 7 Crash Report - 2024-06-05T17_42_31.246Z.txt](https://github.com/user-attachments/files/15594456/Specify.7.Crash.Report.-.2024-06-05T17_42_31.246Z.txt)

**Reported By**
University of Michigan

> IMPACT: This prevents me from keeping our newest project, the MPABI [michigan pathogen] database from upgrading to the newest versions of Specify [beyond 7.9.1.1].

User reports the following:
Problematic in `v7.9.4`
Originally working in `v7.9.1.1`
Tested to seemingly be working ok in `v7.9.3.1` but the test was not exhaustive.

melton-jason:
This issue was caused by #3304. 
Specifically, this happens because as of https://github.com/specify/specify7/pull/3240, the frontend overrides the `CollectionRelType -> name` field to be a picklist with items matching the `name` of `CollectionRelTypes`. 

[...]

PR #4952, opened 2024-05-24
Add series tool in query builder

CarolineDenis:
Fixes #2000 
Fixes #6166 

### Checklist

- [ ] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [ ] Add relevant issue to release milestone

### Testing instructions

- Open the QB with CO table 
- Add catalog num field and some other field 
- [ ] verify series checkbox is present only when Cat num is one of the field 
- Check Series 
- [ ] verify Distinct is read only 
- Uncheck Series 
- [ ] Verify you can check Distinct and that Series become readonly 
- Uncheck Distinct and check Series 
- Run the query 
- [ ] Verify that all the records that have the same data for each field in your query and that have consecutive cat num are group together 
- [ ] Verify that the consecutive cat numbers are displayed as a range 

---
Tips: 
- Create a new CO with a determination and a preparation in a db with auto-numbering for cat num 
- Save 
- Use bulk Carry and create 5 identical copies 
- Create another new CO with all the exact same info as the previous one but skip one cat num 
- This will give you some record to test with series 

---
With Series 
Screenshot 2025-02-03 at 8 50 57 AM

Without Series
Screenshot 2025-02-03 at 8 50 49 AM

NOTES: 
Sorting the column cat num is not available for series

PR #5075, opened 2024-07-07
feat(FieldFormatters): add visual editor

maxpatiiuk:
Fixing an issue from 2012 😊

Fixes https://github.com/specify/specify7/issues/23

Screenshot 2024-07-06 at 17 18 46

### Checklist

- [x] Add automated tests

### Testing instructions

Generally visual editor for Field Formatters should act and look very similar to the one for Record formatters since it shares a lot of code with it.

To test:
- Verify that each field formatter type (constant, any character, ...) can be set in the UI, without user-experience issues, and after saving, can be read back correctly
- Verify that a formatter created/edited using a visual editor works in Specify 7 and Specify 6 (in forms and query results)
  - For reference, here is [how to configure a field formatter in Specify 6](https://discourse.specifysoftware.org/t/add-regular-expression-format-in-specify-6/560)
- P.S: don't try to enter an emoji character like on the screenshot above - Specify 7 back-end does not support emojis in app resources 😢

- Test prs in this list: https://github.com/specify/specify7/pull/5075#issuecomment-3347937852 (go over the main feature not in details) 

[Documentation on field formatters](https://discourse.specifysoftware.org/t/field-formatting-and-auto-numbering/663)


specifyweb/frontend/js_src/lib/components/FieldFormatters/FieldFormatter.tsx
```diff
@@ -0,0 +1,183 @@
...
+        mapping={[
+          mapping,
+          (mapping): void => {
+            if (mapping !== undefined && mapping?.length > 1)
+              softError('Expected mapping length to be no more than 1');
+            const field = mapping?.[0];
+            if (field?.isRelationship === true) {
+              softError(
+                'Did not expect relationship field in field formatter mapping'
+              );
```
maxpatiiuk, line 81: these cases should not happen - the UI doesn't allow them. but handling them here just in case (and to satisfy TypeScript)

specifyweb/frontend/js_src/lib/components/SchemaConfig/Format.tsx
```diff
@@ -314,3 +322,48 @@ function WebLinkEditing({
...
+  value,
+  schemaData,
+}: {
+  readonly table: SpecifyTable;
+  readonly value: string | null;
+  readonly schemaData: SchemaData;
+}): JSX.Element | null {
+  const index = schemaData.uiFormatters
+    .filter((formatter) => formatter.tableName === table.name)
+    .findIndex(({ name }) => name === value);
```
realVinayak, line 337: (not part of any review - was just curious about it)

This is equivalent to 
```
  const index = schemaData.uiFormatters
    .findIndex(({tableName, name}) => tableName === table.name && name == value)
```

Is there a benefit of using the approach in the PR? Will the second part be faster? - maybe JS engine is smart enough to combine the predicates - SQL is (too much SQL in the past days!)

maxpatiiuk: these are not equivalent
if we do only .findIndex(), we get the index in the original array
by filtering first, we get index in the filtered down array (i.e index of a formatter for that particular table, rather than among all tables)

realVinayak: Ah I see, I missed that array will be different

grantfitzsimmons:
❤️

grantfitzsimmons (changes requested):
1. When creating a field format like `GiftNumber` using the regular expression `/^(KUI|KUBI|NHM)$/`, I am unable to set it up correctly.

See that I am supposed to put the 'hint' on the left and 'pattern' on the right:
image

image

This works as expected when using a regex validator, but when setting it as the pattern it is not recognized properly. 

```xml
	<format system="true" name="GiftNumber" class="edu.ku.brc.specify.datamodel.Gift" fieldname="giftNumber" title="" default="true">
		<autonumber>edu.ku.brc.af.core.db.AutoNumberGeneric</autonumber>
		<field type="regex" size="4" value="/^(KUI|KUBI|NHM)$/" byyear="true" pattern="KUBI"/>
	</format>
```

---

2. After modifying other sections of the format, those changes are not made immediately visible for the 'Example Field'. Instead, it uses the previous field format before the edits have been made.

See that it shows "Required Format: KUBI" in the tooltip:

image

(This tooltip is most often obscured by the browser's "match the requested formatter" error)
https://github.com/specify/specify7/assets/37256050/d34bcb16-85d5-43fc-8c68-56bef2e36cc6

After making it simply numeric:

image

---

3. What does the 'Auto Numbering' do that the 'Auto-number' checkbox does not?

image

maxpatiiuk:
> When creating a field format like GiftNumber using the regular expression /^(KUI|KUBI|NHM)$/, I am unable to set it up correctly.

Looks like the expected regex between Sp6 and Sp7 differed.
In Specify 6, the regex is run on each field part separately, so `/^` and `$/` are necessary
In Specify 7, all regexes are joined together into on regex, thus, having `/^` or `$/` will break things if there is more than one part in your formatter.

Fixed:
- When reading from .xml, both Specify 7 front-end and back-end will trim any `/^` or `$/` from the regex pattern to not break the pattern when multiple parts are joined together
- When front-end saves the updated .xml, the syncers will make sure to add `/^` and `$/`, even if not originally specified by the user

In practice, that means:
- When using the visual field format editor, it doesn't matter anymore if you put or not put `/^` and `$/` in a regex field as the value will be normalized automatically
- Backwards compatibility with Specify 6

maxpatiiuk:
> After modifying other sections of the format, those changes are not made immediately visible for the 'Example Field'. Instead, it uses the previous field format before the edits have been made.

Fixed

maxpatiiuk:
> What does the 'Auto Numbering' do that the 'Auto-number' checkbox does not?

Ups, the "Auto Numbering" checkbox at the top is not supposed to be exposed in the UI - removed
(it corresponds to whether .xml had a tag like `<autonumber>edu.ku.brc.specify.dbsupport.CollectionAutoNumber</autonumber>`, which is an sp6-only construct - sp7 cares about it only in so far as creating sp6-compatiable field formatters)

grantfitzsimmons:
@maxpatiiuk Can you resolve the merge conflicts?

lexiclevenger (changes requested):
### Testing instructions

Generally visual editor for Field Formatters should act and look very similar to the one for Record formatters since it shares a lot of code with it.

To test:
- Verify that each field formatter type (constant, any character, ...) can be set in the UI, without user-experience issues, and after saving, can be read back correctly
- Verify that a formatter created/edited using a visual editor works in Specify 7 and Specify 6 (in forms and query results)
  - For reference, here is [how to configure a field formatter in Specify 6](https://discourse.specifysoftware.org/t/add-regular-expression-format-in-specify-6/560)
- P.S: don't try to enter an emoji character like on the screenshot above - Specify 7 back-end does not support emojis in app resources 😢
---
Super exciting! I have a few recommendations:

1. If no default formatter is selected, the message says "record formatter. I would change this to "field formatter"
Screenshot 2024-07-19 at 1 39 31 PM

1. The default for the "size" field for most field types is 0, and this can be saved as long as there is a value in the "hint" field. The value will be shown in the formatter, but it will make the format invalid if that value is entered in the field. In the Specify 6 field formatter, the "size" value cannot go below 1; I think adding this in 7 would be a good solution.
Screenshot 2024-07-19 at 1 42 45 PM



3. The user can input and save any value at any length in the "hint" field (see screenshot above) for the numeric field type, but this is misleading because "#" is the only character that actually works, and it will only appear as many times as the value in the "size" field. Maybe this field could be read-only since it's not really functional.

I'd like to hear what others have to say about 2 and 3; they don't conflict with the xml editor, and the "Example field" displays their consequences correctly, but I think these solutions would make the visual editor more user-friendly.

maxpatiiuk:
> If no default formatter is selected, the message says "record formatter. I would change this to "field formatter"

Good point. To have the message be applicable both for field formatters and record formatters, I edited it to say:

"Please designate one of the formatters as default"

> The default for the "size" field for most field types is 0, and this can be saved as long as there is a value in the "hint" field. The value will be shown in the formatter, but it will make the format invalid if that value is entered in the field. In the Specify 6 field formatter, the "size" value cannot go below 1; I think adding this in 7 would be a good solution.

Made "size" value not go below 1

> The user can input and save any value at any length in the "hint" field (see screenshot above) for the numeric field type, but this is misleading because "#" is the only character that actually works, and it will only appear as many times as the value in the "size" field. Maybe this field could be read-only since it's not really functional.

Fixed. "hint" is now readonly for number fieldsq

lexiclevenger (changes requested):
Looks good! Another thing I noticed is that clicking on the "edit" button next to a format in schema config often takes you to the wrong one.

https://github.com/user-attachments/assets/1f8d27fc-53dd-407b-9bef-90657f9370fd

https://fwri1924-field-editor.test.specifysystems.org/specify/schema-config/en/Gift/

specifyweb/frontend/js_src/lib/components/SchemaConfig/Format.tsx
maxpatiiuk: haha, looks like I shouldn't have done the `.filter()` here after all, as that caused https://github.com/specify/specify7/pull/5075#pullrequestreview-2194509397

the issue is that when computing indexes for record formatters or web links, I don't filter by table - but I forgot about that. And since UI Formatters reuse a lot of code, they should not filter by table either when computing the index

maxpatiiuk:
Fixed!
For testing, to reproduce the bug look at formatters for tables that are not first in the alphabetical order (so all except for Accession 😊)

lexiclevenger (changes requested):
The edit button takes you to the correct resource now! One more issue I didn't notice before (sorry!) is that the save button is enabled as soon as the "Field Formatters" dialog is opened even if no changes have been made.


https://github.com/user-attachments/assets/3eab9e34-0f41-4c65-a4f7-ce814c656fe8

maxpatiiuk:
Fixed merge conflicts.
Rebased with latest changes.
Fixed the bug with "Save" button being active as soon as the tool is opened.

@lexiclevenger Ready for testing again

grantfitzsimmons:
@CarolineDenis Can we fold this into a milestone?

emenslin (changes requested):
I didn't get the chance to do a full test but I did notice a few small things that I think should be fixed or at least further discussed. 

First just a question but should the formatter respect what the size is or what is typed in the hint field? Also I think this was mentioned in another review but could we maybe rename hint to something else, unless this is a 6 thing I feel like it's not super clear what it means. 

- [ ] When mapping to a field it doesn't automatically close when you click it, not a big deal but it's different than other places so it's somehting to consider.

Field formatters:

https://github.com/user-attachments/assets/f43a88e6-a0c5-4181-9497-38caba8448a3


Record formatter:

https://github.com/user-attachments/assets/3ca54c2b-f2f8-427f-9612-f79f23ec39ff


- [ ] There is no size limit for the 'numeric' field and if you make the number high enough it eventually crashes the browser

https://github.com/user-attachments/assets/775b7016-9230-4177-bf71-318cf5f63563


- [ ] You can remove the number from the size field and it won't throw an error even for fields where you shouldn't be able to edit it

https://github.com/user-attachments/assets/e9afd7fb-261b-4baf-af50-608c34afd4c2


- [ ] Similar to the last point, anything filled out will carry over if you change the type 

https://github.com/user-attachments/assets/601f0bb3-33d9-4137-80db-136a8933eccf


- [ ] You can type in the example field 

https://github.com/user-attachments/assets/a67cf3e0-6319-4962-9d7e-950abf1b04a8


- [ ] You can edit the records in the preview

https://github.com/user-attachments/assets/f471e0a0-33bb-4e36-9a1d-f75cee525a25


- [ ] When you select year nothing shows in the example field, is this intentional? 

https://github.com/user-attachments/assets/411dc51f-849a-4759-9034-52d21e045d40


Overall though I am excited for this, it looks good!

emenslin (changes requested):
Thank you for the explanation of the types it was very helpful! I think 'value' instead of 'hint' would make more sense as it is likely what the users are more familiar with but I am open to other opinions @grantfitzsimmons 

> Set max size to 9999.

This works technically but it messes up the dialog, is there a reason the size would need to be this big in the first place? Also you can keep typing in the size field above 9999 and eventually crash the browser still, it just turns red if you click out of it. 

https://github.com/user-attachments/assets/b47acc8c-75ae-4c98-867c-5c694a51ad34



https://github.com/user-attachments/assets/ba65619a-fe85-41fa-8dc8-9d47ef4b7a40



> The record being editable is useful in the Record formatter and Web Link formatter as there the value could be coming from a different field (or multiple fields) and so it is useful to be able to edit different record fields and see how they affect the formatted result/weblink url

I thought this issue was fixed but I was thinking of #4891 which stopped you from editing/creating new records in form definitions, however, #4795 is the open issue for fixing this behavior in record formatters and has not been fixed yet. I get what you're saying though and I think a discussion should be had on if we want users to be able to edit forms this way or not. 



Information is still carried over when you switch the type, it does change when you select year or numeric but the other ones still carry over. This mostly just seems like a problem because form some of the types you can't change the size but when the info is carried over the size will just be whatever it was for the previous type.


https://github.com/user-attachments/assets/06419d76-ab0c-4708-b754-ed04e19c87f3



--- 

New issues: 

If you have the size field selected and scroll up and down the number in the field changes. This is possibly #4245 resurfacing. 

https://github.com/user-attachments/assets/3728fdcc-f776-4e3d-b680-e004abe0634e



Not sure if this is supposed to happen but when I set a numeric type to auto-incrementing and put it on the fax field and tried to actually use and save on the form I got an error. Assuming this is intentional, we should probably try to find another solution other than the user getting an error in the forms (maybe some sort of save blocker in the field formatter dialog?) 

https://github.com/user-attachments/assets/dff6a7e6-916a-4e25-a3f1-8bf086cba3b8


[Specify 7 Crash Report - 2024-12-16T15_51_26.733Z.txt](https://github.com/user-attachments/files/18153124/Specify.7.Crash.Report.-.2024-12-16T15_51_26.733Z.txt)


When typing in the hint field for some of the types it changes the size depending on how many characters you type but for others it doesn't, is this the expected behavior? 

https://github.com/user-attachments/assets/67bbab4e-ae85-44da-be7d-a304959f5a8b


This behavior is fine but is there a reason why clicking the add button in schema config just takes you to app resources instead of pulling up a dialog in schema config? 

https://github.com/user-attachments/assets/52e01b1d-f0df-4153-8c9d-2408205c75a5


The existing system formatters can have multiple formatters be defaults for one table but for new ones only one can be default. This is not necessarily a problem as no errors are thrown or anything but I don't know if it could cause any problems. 

https://github.com/user-attachments/assets/34c29472-434d-451c-85d2-395295043d39


The description for field formatters in the resource says: 

Screenshot 2024-12-16 103622

And I think this is pretty good but still a little confusing, could it be a little similar to how the description is for table formats/aggregations to also keep things consistent? 

Screenshot 2024-12-16 103737


I think we should consider removing some tables from the field formatters because right now you can add one to any table. I can see this causing problems but I might be missing some potential use cases. 

Same thing with some of the fields. You can add a formatter to the GUID field which should be autogenerated and it just causes a lot of problems. 


https://github.com/user-attachments/assets/ef52d5e3-4b57-470f-a3dd-c2c0cbdfcffb

maxpatiiuk:
> Thank you for the explanation of the types it was very helpful! I think 'value' instead of 'hint' would make more sense as it is likely what the users are more familiar with but I am open to other opinions

Sounds good. Renamed to "Value"

> This works technically but it messes up the dialog, is there a reason the size would need to be this big in the first place? Also you can keep typing in the size field above 9999 and eventually crash the browser still, it just turns red if you click out of it.

Good point. Will reduce to 99. We can increase it back if we receive feedback (or people can edit it directly in XML)

> Information is still carried over when you switch the type, it does change when you select year or numeric but the other ones still carry over. This mostly just seems like a problem because form some of the types you can't change the size but when the info is carried over the size will just be whatever it was for the previous type.

Doesn't seem to be a problem - the size is reduced as needed when switching for example from numeric to year:

https://github.com/user-attachments/assets/34aec3f7-0ef8-461b-a4ed-313f2492174c

> If you have the size field selected and scroll up and down the number in the field changes. This is possibly https://github.com/specify/specify7/issues/4245 resurfacing.

Looks like that bug was fixed for floating point fields, but not for integer fields - fixed. Thanks for catching!

> Not sure if this is supposed to happen but when I set a numeric type to auto-incrementing and put it on the fax field and tried to actually use and save on the form I got an error. Assuming this is intentional, we should probably try to find another solution other than the user getting an error in the forms (maybe some sort of save blocker in the field formatter dialog?

Can reproduce.
@melton-jason The error says "<modelName> has no hierarchy field"
Is auto numering only supported for tables with hierarchy fields? Otherwise, can auto-numbering be global? or can it piggy-back on the "parent" table hierarchy (in this case Address is a dependent record of Agent, which is attached to a division)

> When typing in the hint field for some of the types it changes the size depending on how many characters you type but for others it doesn't, is this the expected behavior?

Expected. For constant parts, the "Value" is literally the exact and only value that can be put into that part - so the size must equal the "Value".
For Alpha-numeric fields, "Value" is just a placeholder/hint, and doesn't have a relation to "size". For more details, see https://github.com/specify/specify7/pull/5075#issuecomment-2543232621

> This behavior is fine but is there a reason why clicking the add button in schema config just takes you to app resources instead of pulling up a dialog in schema config?

An inconsistency! Fixed

> The existing system formatters can have multiple formatters be defaults for one table but for new ones only one can be default. This is not necessarily a problem as no errors are thrown or anything but I don't know if it could cause any problems.

It may cause inpredictability/inconsistency as to which will actually be the default (as only one will be used by specify by default). The same issue is present in Record formatters.

The way I decided to handle them there is:
- if there are multiple defaults already in .xml definition, then keep it so to not break any behavior the user may be relying on
- if user tries to set a new default, unset all other defaults

If you wish I can make it more proactive, and display a warning message if multiple defaults are present. For now, we warn in the opposite case:

Screenshot 2024-12-25 at 14 49 31

> And I think this is pretty good but still a little confusing, could it be a little similar to how the description is for table formats/aggregations to also keep things consistent?

Made it more consistent. Do say if you have more feedback

> Same thing with some of the fields. You can add a formatter to the GUID field which should be autogenerated and it just causes a lot of problems.

Good point. I excluded all read only and virtual fields from the list (for example CatalogNumber.age)
Unfortunately, guid is not virtual, and not read only in the schema (not even hidden), so there is no easy way to exclude it. Still, a good chunk of fields is excluded.
I belive GUID is not read only because it can be manually entered for some tables, like Taxon

melton-jason:
> > Not sure if this is supposed to happen but when I set a numeric type to auto-incrementing and put it on the fax field and tried to actually use and save on the form I got an error. Assuming this is intentional, we should probably try to find another solution other than the user getting an error in the forms (maybe some sort of save blocker in the field formatter dialog?) 
> > 
> > https://github.com/user-attachments/assets/dff6a7e6-916a-4e25-a3f1-8bf086cba3b8
> 
> Can reproduce.
> @melton-jason The error says " has no hierarchy field"
> Is auto numering only supported for tables with hierarchy fields? Otherwise, can auto-numbering be global? or can it piggy-back on the "parent" table hierarchy (in this case Address is a dependent record of Agent, which is attached to a division)

From https://github.com/specify/specify7/pull/5075#pullrequestreview-2506791959 and https://github.com/specify/specify7/pull/5075#issuecomment-2562018456

Yes, looks like auto-numbering right now is only supported for tables with hierarchy fields. 

Specifically for auto-numbering, the backend uses the `filter_by_colleciton` function (so this error is the same one as https://github.com/specify/specify7/issues/4989 and https://github.com/specify/specify7/issues/3564)

https://github.com/specify/specify7/blob/1d6b1ee8bc51e506e4f4d71a2e8a6bd28992991d/specifyweb/specify/uiformatters.py#L32-L33

The function has support for a `strict` argument (which defaults to true if unset), so the easiest solution in this case would just be less strict about scoping (i.e., make the scoping global). 

The potential consequence of doing so is that if there are two or more records which are supposed to have differing scopes, they will share a global autonumbering scheme. 
I think this solution would work in most cases, but there are times when scoping becomes tricky (artifacts of Specify 6 aside). 

Anything besides making the scoping less strict will require some implementation of https://github.com/specify/specify7/pull/5044#pullrequestreview-2425174036

--- 

> or can it piggy-back on the "parent" table hierarchy (in this case Address is a dependent record of Agent, which is attached to a division)

I think Address is a really great example to consider scoping for auto-numbering (and other applications)!
(For general number of cases, maybe not actual feasibility of being useful) 

Building on this, Address _also_ has independent relationships in the form of:
- Division -> address
- Institution -> address
- InstitutionNetwork -> address

Because of the dependent relationship of `Agent -> addresses`,  we know that - if desired - auto-numbering of fields within address can even be conceptually restricted to a per-agent basis if desired, though I'd assume we'd try and use the "parent" record's scope. 
Should the dependent addresses be included when determining the next number for addresses in independent relationships? 

In the case of the independent relationships, should the Division addresses be considered separately from the Institution/InstitutionNetowork addresses? That is, if we add a new Division, do we only consider the addresses of other Divisions or just all addresses in general? (which might be the same in this case, but might not always be the case across all relationships).

A global approach is by far the easiest to implement and think about, at least from a developmental and design standpoint. 
**_As long as it would be expected by most users_**, that's the route I would take.

maxpatiiuk:
> Yes, looks like auto-numbering right now is only supported for tables with hierarchy fields.

I see. That might match sp6 behavior.
@grantfitzsimmons have you heard of users requesting we expand auto numbering support?
Otherwise, I will keep the scope of this PR in check and restrict auto numbering in UI to what us allowed by the back-end - we can reconsider once we get a chance to re-engineer the scoping system

emenslin (changes requested):
Sorry I did not have time for a full review but here's a couple things I noticed: 
If you shrink the dialog the preview background shrinks but if the value is long enough it extends past the background. The value should always have that background behind it. 

https://github.com/user-attachments/assets/c6690f5c-1aca-4050-b39b-4718f60d4ac0

Going back to this point: 
> Information is still carried over when you switch the type, it does change when you select year or numeric but the other ones still carry over. This mostly just seems like a problem because form some of the types you can't change the size but when the info is carried over the size will just be whatever it was for the previous type.

It is not a huge deal especially if it is hard to fix but if possible I would like the carrying over to be fixed. I can imagine it being annoying if you click on the wrong thing and then having to take the extra step to delete it all. It doesn't matter too much since it is properly handled, just something I could potentially see users complain about 

https://github.com/user-attachments/assets/1da0cc92-0bd4-455d-829c-80c5e96d5457

maxpatiiuk:
> If you shrink the dialog the preview background shrinks but if the value is long enough it extends past the background

Fixed by force-wrapping long values.

> It is not a huge deal especially if it is hard to fix but if possible I would like the carrying over to be fixed

Could you clarify what you mean by fixed?
Do you want carrying over to be completely disabled? It is there as feature - if you say it is an anti feature, we can remove it
If by fixed you mean (looking at your example video) the size should revert back to 8 after being truncated to 4 be, the that is possible but a bit tricky.

lexiclevenger (changes requested):
The fix for https://github.com/specify/specify7/pull/5075#pullrequestreview-2565821373 looks good!

https://github.com/user-attachments/assets/4ddcb891-2c5b-4a1e-b0e0-830a1f35183b

However, there is still an issue with entering large numbers in the size field, though. Here, I enter "99999," and it throws an error. I can then save the format, breaking it in the visual editor completely. 

https://github.com/user-attachments/assets/873044d6-c371-4b88-ae67-1c5ff48c952f


I think the carrying over is fine--it is also a feature for certain field types in the Specify 6 visual editor. I think the Year and Numeric field types are the only bothersome ones. Would it be possible to disable it for just those field types? It might be even better if the last value for 'size' and 'value' entered by the user could also be returned if the field type is changed again. I'm not sure if this would be 100% intuitive either, though. What do you think, @emenslin?

emenslin:
The carrying over is not really a feature anywhere in 7 as far as I know which is why I think it should be removed. From my perspective I feel like it could be confusing to the user to have values carried over, but if people thing otherwise then I don't mind it staying in. From @maxpatiiuk comment 

>If by fixed you mean (looking at your example video) the size should revert back to 8 after being truncated to 4 be, the that is possible but a bit tricky.

I do not think this should be the case though. By fixed I did mean to remove it entirely but if other people do think it should be a feature then that's okay. The main reason I wanted it to be removed is because I did find it annoying when testing but if we don't think the users will care about it then I don't mind keeping it in. @grantfitzsimmons do you have any thoughts on this?

emenslin (changes requested):
I didn't get a chance for a full review but I wanted to comment what I've found so far. 


Previous issue that came back 

- [ ] Error message says 'record formatter' instead of either formatter or field formatter (first mentioned [here](https://github.com/specify/specify7/pull/5075#pullrequestreview-2189080837))

image

---

Catalog number formats issues 

- [ ] This might not be a big deal; however, the format marked as the default does not match the COT default format. 
- [ ] `Visual editor is not available for this resource` for a default catalog number format 
Screenshot 2025-09-29 115758

- [ ] `Preview not available` also for a default catalog number format 

Screenshot 2025-09-29 115957

---

Autonumber 

- [ ] Auto-number is disabled on a numeric field for certain tables, but auto-number by year seems to work fine. On other tables such as CO, accession, gift, loan, etc. auto-number is enabled for both numeric and by year. Could have to do with this comment https://github.com/specify/specify7/pull/5075#issuecomment-2563163573 but I am not sure so I wanted to mention it 

https://github.com/user-attachments/assets/158eded3-7b9c-4103-8d73-81f9f3436f9b


Link to the DB I was using: https://ojsmnh20250910-field-editor.test.specifysystems.org/specify/

emenslin:
I started on the list of PRs but I did not finish so this only includes the PRs I had a chance to get through. 

- [x] https://github.com/specify/specify7/pull/3501
- [x] https://github.com/specify/specify7/pull/4804
- [x] https://github.com/specify/specify7/pull/4952
- [x] https://github.com/specify/specify7/pull/5240
- [ ] https://github.com/specify/specify7/pull/5242 

Bulk carry forward is enabled for alphaNumeric catalog numbers in this PR but not in main 
Main: 
image
Issue branch: 
image

- [x] https://github.com/specify/specify7/pull/5253
- [x] https://github.com/specify/specify7/pull/5355
- [x] https://github.com/specify/specify7/pull/5400
- [x] https://github.com/specify/specify7/pull/5431
- [x] https://github.com/specify/specify7/pull/5476
- [x] https://github.com/specify/specify7/pull/5485
- [x] https://github.com/specify/specify7/pull/5489
- [x] https://github.com/specify/specify7/pull/6115
- [x] https://github.com/specify/specify7/pull/6206
- [x] https://github.com/specify/specify7/pull/6245
- [x] https://github.com/specify/specify7/pull/6246
- [x] https://github.com/specify/specify7/pull/6252
- [x] https://github.com/specify/specify7/pull/6255

emenslin (approved):
All the above issues I mentioned are fixed! I tested the rest of the PRs I didn't get a chance to before and also removed some either I decided weren't relevent or there are already open known issues with them, looks good. 
- [x] https://github.com/specify/specify7/pull/6285
- [x] https://github.com/specify/specify7/pull/6300
- [x] https://github.com/specify/specify7/pull/6306
- [x] https://github.com/specify/specify7/pull/6324
- [x] https://github.com/specify/specify7/pull/6373
- [x] https://github.com/specify/specify7/pull/6419
- [x] https://github.com/specify/specify7/pull/6469
- [x] https://github.com/specify/specify7/pull/6483
- [x] https://github.com/specify/specify7/pull/6710
- [x] https://github.com/specify/specify7/pull/6721
- [x] https://github.com/specify/specify7/pull/6994
- [x] https://github.com/specify/specify7/pull/7080
- [x] https://github.com/specify/specify7/pull/7143
- [x] https://github.com/specify/specify7/pull/7179
- [x] https://github.com/specify/specify7/pull/7341
- [x] https://github.com/specify/specify7/pull/7452

PR #5090, opened 2024-07-10
Adjust notification timing for development mode

combs-a:
This PR adjusts the interval for when notifications are fetched in development mode, to make it easier to filter through requests in the debug console while testing/debugging/etc.

I set the development time interval to 1 minute, but that can be adjusted easily if needed.

### Checklist

- [ ] Add automated tests

### Testing instructions

- [ ] Check that production mode still has requests about every 5 seconds.
- [ ]  Check that development mode has requests about every minute.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
@@ -8,7 +8,7 @@ import { sortFunction } from '../../utils/utils';
 import { formatUrl } from '../Router/queryString';
 import type { GenericNotification } from './NotificationRenderers';
 
-let INITIAL_INTERVAL = (process.env.NODE_ENV === 'development') ? 60000 : 5000;
+const INITIAL_INTERVAL = (process.env.NODE_ENV === 'development') ? 60000 : 5000;
 const INTERVAL_MULTIPLIER = 1.1;
 
 export function useNotificationsFetch({
--- a/specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
@@ -8,7 +8,7 @@ import { sortFunction } from '../../utils/utils';
 import { formatUrl } from '../Router/queryString';
 import type { GenericNotification } from './NotificationRenderers';
 
-const INITIAL_INTERVAL = 5000;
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
```diff
@@ -5,10 +5,12 @@ import { ajax } from '../../utils/ajax';
 import { formatDateForBackEnd } from '../../utils/parser/dateFormat';
 import type { IR, RA } from '../../utils/types';
 import { sortFunction } from '../../utils/utils';
+import { MINUTE, SECOND } from '../Atoms/timeUnits';
 import { formatUrl } from '../Router/queryString';
 import type { GenericNotification } from './NotificationRenderers';
 
-const INITIAL_INTERVAL = 5000;
+const INITIAL_INTERVAL =
+  process.env.NODE_ENV === 'development' ? Number(MINUTE) : 5 * SECOND;
```
maxpatiiuk, line 13: MINUTE is already a number, so no need to convert it?

```suggestion
  process.env.NODE_ENV === 'development' ? MINUTE : 5 * SECOND;
```

maxpatiiuk (approved):
This looks good, but if you are interested in an even better solution, there are some more sophisticated web APIs we can use to make notifications fetching more efficient and to reduce the noise in the network panel. See:
https://github.com/specify/specify7/issues/2606
(that issue is marked as fixed because a basic improvement was done, but more improvements are still possible)

specifyweb/frontend/js_src/lib/components/Notifications/hooks.tsx
combs-a: Sorry about late reply--ESLint/Prettier wanted to make that change, I had it w/o Number before 😅 we're all good to change it back

alesan99 (approved):
### Testing instructions

- [x] Check that production mode still has requests about every 5 seconds.
- [x]  Check that development mode has requests about every minute.

Looks good! 👍👍 The network logs definitely slow down in dev mode.

grantfitzsimmons (approved):

Areyes42 (approved):
### Testing instructions

-  [x]  Check that production mode still has requests about every 5 seconds.
-  [x]  Check that development mode has requests about every minute.
***
Everything looks good, awesome job Anya! 👏

grantfitzsimmons:
🎉 😄

PR #5091, opened 2024-07-10
Add a feature to choose type of taxon tree in WB

### LINKED ISSUE #4980: Add support in the WorkBench for specifying which tree you want to upload to

To identify which tree will be used to create new records, we will need to have an object type or specified tree included in the  data set.

Also need to handle: 
**Describe the bug**
After the latest changes for multiple trees and MOTs are present on my database, my Taxon record set no longer functions. I also can't create a new record set based on this old import. I cannot recreate before the migrations were run.

I've uploaded this database to the test panel, with the user in this case being `spadmin`. 
Database is `geo_swiss_2024_08_05`.

**To Reproduce**

1. Go to an existing Taxon record set
2. See the `No Taxon matches the given query.` error message
3. Go to the WorkBench data set where you uploaded the taxon records
4. Try to create a 

**Expected behavior**
Existing taxon record sets should still be usable and I should still be able to create a record set from taxonomy uploaded in an earlier version (before MOTs).

**Screenshots**

https://github.com/user-attachments/assets/98999946-93c2-4728-ad9e-e3983eb13920

**Crash Report**
Arc, macOS

Database is `geo_swiss_2024_08_05`. User is `spadmin`

`edge`
Git SHA: https://github.com/specify/specify7/commit/9f37028b2e20f10eb9972d50f4058790b3dcca9f

`No Taxon matches the given query.`: [Specify 7 Crash Report - 2024-08-05T16_25_20.425Z.txt](https://github.com/user-attachments/files/16500287/Specify.7.Crash.Report.-.2024-08-05T16_25_20.425Z.txt)

CarolineDenis:
Fixes #4980 

### Checklist

### Testing instructions
I created a database named 'sp7demofish_mots.sql' on the test panel, this has some sample Taxon trees created to use in this PR's test.  Use the screenshots for extra testing details.

Here is a test panel instance to use: https://sp7demofishmots-issue-4980.test.specifysystems.org

- [x] Setup the WB test by adding a new taxon tree and ranks.  Also, create a new CollectionObjectType.
- [x] Add new CollectionObjects with a Determination and CollectionObjectType. Do a workbench upload with CollectionObject as the base table.  Create workbench columns like "Det->Det 1->TestTree->TestKingdom->Name" (Edit: ignore this step, COType is now hidden in v7.9.7)
- [x] Add new CollectionObjects with Determination, this time without specifying a CollectionObjectType.
- [x] Add new Taxons to Ranks of different Taxon trees. Do a workbench upload with Taxon as the base table.  Create workbench columns like "Taxon->TestTree->TestKingdom->Name".
- [x] Add new CollectionObjects from Determinations.  Do a workbench upload with Determination as the base table.  Create workbench columns like "Determination->Taxon->TestTree->TestKingdom->Name".
- [x] Check that the Taxon Trees have been updated in the Tree Viewer. 

image
image
image
image
image
image
image
image


CarolineDenis:
TODO: 
- fix upload plan in WB 

(@acwhite211 )

specifyweb/frontend/js_src/lib/components/WbPlanView/navigator.ts
```diff
@@ -254,7 +256,19 @@ export function getMappingLineData({
...
+          .filter((field) => {
+            if (
+              internalState.mappingLineData.length === 0 &&
+              taxonType !== undefined
+            ) {
+              return field[1].tableTreeDefName === taxonType;
+            }
+            return true;
+          })
+          .map(([rawKey, fieldData]) => [rawKey, fieldData])
```
maxpatiiuk, line 270: this .map() seems to have no effect?

maxpatiiuk (approved):

realVinayak:
IIRC, we were going to make it so that users select treedef when they open CO form. It'd absolutely simplify things here too...

specifyweb/frontend/js_src/lib/components/WbPlanView/navigator.ts
CarolineDenis: yes might be the same as in issue 4829.

CarolineDenis:
NOTES: 
- need backend fixes 

@acwhite211

CarolineDenis:
> IIRC, we were going to make it so that users select treedef when they open CO form. It'd absolutely simplify things here too...
> 
> Also, this is simpler and more straightforward on the backend than the query builder...provided upload plan identifies the tree for us

This was a discussion we had during a meeting but we decided not to go with this solution

CarolineDenis:
NOTES:

Need to remove: when selecting Taxon as base table in the WB, the user is now presented a list of available tree to upload to

Implement a first column in the query mapper with the trees and only when selected the ranks, same in the query lines

acwhite211:
Here is an example of the upload plan I designed the back-end to handle:
```json
{
  "baseTableName": "taxon",
  "taxonTreeId": "1",
  "uploadable": {
    "treeRecord": {
      "ranks": {
        "Kingdom": {
          "treeNodeCols": {
            "name": "Kingdom",
            "author": "Kingdom - Author"
          },
          "treedefid": "1"
        },
        "Phylum": {
          "treeNodeCols": {
            "name": "Phylum",
            "author": "Phylum - Author"
          },
          "treedefid": "1"
        },
        "Kingdom Test": {
          "treeNodeCols": {
            "name": "Kingdom Test",
            "author": "Kingdom Test - Author"
          },
          "treedefid": "2"
        }
      }
    }
  }
}
```

The taxonTreeId it the id of the base table taxon id, and within the ranks, the treedefid  is the treedefid of the rank.  If the treedefid is not given for the rank, the taxontreedefid is used, or we just search for the rank by name and throw an error if there are duplicates.  On the backend, I'm allowing for taxonTreeId and treedefid to be optional, so the backend will try and infer what they should be, and then will throw an error if it cannot determine the treeid.  If inferences are made, the upload plan is edited on the backend to add those values.

I experimented with adding definition in the treeNodeCols, but the front-end threw errors about putting a relationship in the treeNodeCols.  Let me know if we should try and get that to work, rather than using treedefid outside of treeNodeCols.

Let's discuss further in dev meeting.

realVinayak:
> The taxonTreeId it the id of the base table taxon id, and within the ranks

Oh man, i'll need to adjust batch-edit for that! But, not too much.

acwhite211:
This PR can be tested by the UX Testers once the front-end request to the WB back-end is added.

acwhite211 (approved):

specifyweb/workbench/upload/treerecord.py
```diff
@@ -19,33 +19,138 @@
 
 logger = logging.getLogger(__name__)
 
+class TreeRank:
```
realVinayak, line 22: Can you combine this or get rid of this, in any way, possibly using `scopingAttrs`? Try to keep props usage to minimal, and only if none of others workout.

acwhite211:
@specify/ux-testing  @specify/testing This PR is ready for testing.  Use the sp7demofish_cots database on the test panel, here is a currently running instance https://sp7demofishmots-issue-4980.test.specifysystems.org/. Let me know if you have any questions about the testing steps, I added screenshots in the testing instruction that will hopefully help.

specifyweb/frontend/js_src/lib/tests/fixtures/uploadplan.1.json
```diff
@@ -283,34 +283,40 @@
...
-                        }
+                        },
+                        "treeId": 1
                       },
                       "Family": {
                         "treeNodeCols": {
                           "name": "Family"
-                        }
+                        },
+                        "treeId": 1
```
realVinayak, line 293: Why do I care about this? You cannot, ever, select ranks from different trees, correct? Like, say, Phylum from tree1 and Species from tree 4? You are overcomplicating design, making things easier to break, for something that just isn't needed. So, get rid of treeId from here, and just make it at the top for treerecord.

realVinayak: @acwhite211 do the adjustments on the backend, and see if it can be simplified.

specifyweb/workbench/upload/treerecord.py
realVinayak: At least make it namedtuple. It's not about "obsession with readonly". If I want to refactor this, I'd need to assume that you are sneakily mutating something. make stuff easier for others, make it named tuple.

specifyweb/workbench/views.py
```diff
@@ -626,22 +627,35 @@ def upload(request, ds, no_commit: bool, allow_partial: bool) -> http.HttpRespon
...
-        ds.save(update_fields=['uploaderstatus'])
-
-    return http.JsonResponse(async_result.id, safe=False)
-
+        data = json.loads(request.body) if request.body else {}
+        background = True
+        if 'background' in data:
+            background = bool(data['background']) # {"background": false}
+
+        if background:
```
realVinayak, line 635: good job! i change views for it, this will be helpful

specifyweb/workbench/upload/treerecord.py
```diff
@@ -60,15 +165,15 @@ def get_treedefs(self) -> Set:
     def bind(self, collection, row: Row, uploadingAgentId: Optional[int], auditor: Auditor, cache: Optional[Dict]=None, row_index: Optional[int] = None) -> Union["BoundTreeRecord", ParseFailures]:
         parsedFields: Dict[str, List[ParseResult]] = {}
         parseFails: List[WorkBenchParseFailure] = []
-        for rank, cols in self.ranks.items():
+        for tree_rank_record, cols in self.ranks.items():
             nameColumn = cols['name']
-            presults, pfails = parse_many(collection, self.name, cols, row)
-            parsedFields[rank] = presults
+            presults, pfails = parse_many(collection, self.name, cols, row) # TODO: fix
```
realVinayak, line 170: TODO: fix

"fix"? Please mention what needs to be.

specifyweb/workbench/upload/treerecord.py
```diff
@@ -19,33 +19,138 @@
...
     def unparse(self) -> Dict:
         return { 'baseTableName': self.name, 'uploadble': self.to_json() }
 
 class ScopedTreeRecord(NamedTuple):
     name: str
-    ranks: Dict[str, Dict[str, ExtendedColumnOptions]]
+    ranks: Dict[TreeRankRecord, Dict[str, ExtendedColumnOptions]]
     treedef: Any
     treedefitems: List
     root: Optional[Any]
```
realVinayak, line 156: simplify all this. what's the point of "TreeRankRecord"? Why is getting treedef and treedefitems based on basetreedef not enough? If we ever decide to support ranks across trees, you can message me "I told you so"

realVinayak (changes requested):

specifyweb/workbench/upload/treerecord.py
acwhite211: > The front-end's design allowed for selecting ranks from multiple trees, so the back-end got adjusted to handle it.  It would have been better probably to just use the unique db id for Treedefid and Treedefitemid to identify ranks throughout Specify, rather than by name.

specifyweb/frontend/js_src/lib/tests/fixtures/uploadplan.1.json
acwhite211: The front-end's design allowed for selecting ranks from multiple trees, so the back-end got adjusted to handle it.  It would have been better probably to just use the unique db id for Treedefid and Treedefitemid to identify ranks throughout Specify, rather than by name.

specifyweb/workbench/upload/upload_plan_schema.py
```diff
@@ -217,14 +232,17 @@
 
 def parse_plan_with_basetable(collection, to_parse: Dict) -> Tuple[Table, Uploadable]:
     base_table = datamodel.get_table_strict(to_parse['baseTableName'])
-    return base_table, parse_uploadable(collection, base_table, to_parse['uploadable'])
+    extra = {}
+    if 'treeId' in to_parse.keys():
+        extra['treedefid'] = to_parse['treeId']
```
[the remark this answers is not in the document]
acwhite211: > Here is an example of the upload plan I designed the back-end to handle:
> 
> ```json
> {
>   "baseTableName": "taxon",
>   "treeId": 1, # base tree table, ranks default to this tree when no treeId is given with the rank
>   "uploadable": {
>     "treeRecord": {
>       "ranks": {
>         "Kingdom": {
>           "treeNodeCols": {
>             "name": "Kingdom",
>             "author": "Kingdom - Author"
>           },
>           "treeId": 1 # the treedefid of the rank
>         },
>         "Phylum": {
>           "treeNodeCols": {
>             "name": "Phylum",
>             "author": "Phylum - Author"
>           },
>           "treeId": 1 # the treedefid of the rank
>         },
>         "Kingdom Test": {
>           "treeNodeCols": {
>             "name": "Kingdom Test",
>             "author": "Kingdom Test - Author"
>           },
>           "treeId": 2 # the treedefid of the rank
>         }
>       }
>     }
>   }
> }
> ```
> 
> The treeId it the id of the base table taxon id, and within the ranks, the treedefid is the treedefid of the rank. If the treedefid is not given for the rank, the taxontreedefid is used, or we just search for the rank by name and throw an error if there are duplicates. On the backend, I'm allowing for taxonTreeId and treedefid to be optional, so the backend will try and infer what they should be, and then will throw an error if it cannot determine the treeid. If inferences are made, the upload plan is edited on the backend to add those values.
> 
> I experimented with adding definition in the treeNodeCols, but the front-end threw errors about putting a relationship in the treeNodeCols. Let me know if we should try and get that to work, rather than using treedefid outside of treeNodeCols.
> 
> Let's discuss further in dev meeting.

specifyweb/workbench/upload/treerecord.py
acwhite211: The rank is identified by TreeRankRecord (rank_name, treedef_id) now instead of just rank_name.

realVinayak:
> The front-end's design allowed for selecting ranks from multiple trees, so the back-end got adjusted to handle it.

And? not for the SAME tree though, so all that treerank isn't needed. you can go ahead with this, I'll simplify this when I merge batch-edit

emenslin (changes requested):
If you try to map a taxon through either the determination base table or the collection object base table it will always remove the rank from the mapping after you save and go back to the data mapper. 

https://github.com/user-attachments/assets/ef8ee235-7f0e-4876-9192-a01b2ea59666


Also collection object type is not showing up in the workbench so you cannot map it

acwhite211:
Thanks @emenslin for finding that issue. Looks like we need to add a front-end fix in the workbench so that when you reopen the data mapper, the display will be able to show the correct rank based on the treeId of the rank in the upload plan.  

@sharadsw It should be similar to the code for creating the upload request, we just need to handling loading the info back to the data mapper viewer.

This screenshot is before saving the upload plan:
image

This screenshot is from reopening the upload plan in the data mapper:
image

emenslin (changes requested):
If you try to do a workbench upload with multiple different taxon trees it will only take whatever the first tree is and update that one and not the others, this is with either CO or taxon as the base table 

https://github.com/user-attachments/assets/3ae7438a-23b3-4786-a523-be622ff56502

sharadsw:
Pushed a fix for the mapper not saving the tree rank.

sharadsw:
@emenslin The mapper should be fixed now and no longer lose the rank from the mapping.

acwhite211:
I've been working on fixing the case where new taxon are being uploaded with multiple ranks from multiple trees in the upload plan. The upload plan is being set correctly, the bug right now is that it just uses the treedef of the first tree rank in the upload plan.  I've been experimenting with some different ideas, but the problem is that a solution will either take a big overhaul to the WB code, or be incredibility hacky.  We should talk over what we want to happen here.  Both solutions aren't great, and I don't want to make the merge conflicts even worse for batch-edit with a big code change.  We might want to consider this an edge case for now and fix it at a later time.

specifyweb/workbench/upload/treerecord.py
```diff
@@ -19,33 +20,134 @@
...
+    def check_rank(self, tree: str) -> bool:
+        return TreeRank.create(self.rank_name, tree, self.treedef_id).check_rank()
+    
+    def validate_rank(self, tree) -> None:
+        TreeRank.create(self.rank_name, tree, self.treedef_id).validate_rank()
 
 class TreeRecord(NamedTuple):
     name: str
-    ranks: Dict[str, Dict[str, ColumnOptions]]
+    ranks: Dict[Union[str, TreeRankRecord], Dict[str, ColumnOptions]]
```
realVinayak, line 118: I'll highly recommend getting rid of the union here. Make it consistent (doesn't matter which you choose - TreeRankRecord or str).

acwhite211: Good suggestion, I was going for backwards compatibility here.  So you can still do it the old way if the rank name is unique in all trees, if not, it will just throw an error.

specifyweb/workbench/upload/treerecord.py
```diff
@@ -158,10 +158,54 @@ def disambiguate(self, disambiguation: DA) -> "ScopedTreeRecord":
...
+            col
+            for col, value in row.items()
+            if value is not None and value != "" and col in rank_names
+        }
+
+        # Handle cases with multiple or no ranks in the row
+        if len(ranks_in_row_not_null) > 1:
+            return self, WorkBenchParseFailure('multipleRanksInRow', {}, list(ranks_in_row_not_null)[0])
+        if not ranks_in_row_not_null:
+            return self, WorkBenchParseFailure('noRanksInRow', {}, '')
```
realVinayak, line 181: This is not correct. It will be considered a NullRecord. There should be some unit tests failing complaining about that. In any case, just pick a random tree. In your current implementation, you are always forcing to have a record (when, in reality, relationship doesn't need to be present)

specifyweb/workbench/upload/treerecord.py
```diff
@@ -158,10 +158,54 @@ def disambiguate(self, disambiguation: DA) -> "ScopedTreeRecord":
...
+
+        # Create a mapping of rank names to their treedef IDs
+        ranks = {rank.rank_name: rank.treedef_id for rank in self.ranks.keys()}
+        rank_names = set(ranks.keys())
+
+        # Find non-null rank columns in the row
+        ranks_in_row_not_null = {
+            col
+            for col, value in row.items()
+            if value is not None and value != "" and col in rank_names
```
realVinayak, line 174: Don't check for `value != ''`. it's already handled before you get here. It makes me think "do other places need this?"

specifyweb/workbench/upload/treerecord.py
```diff
@@ -158,10 +158,54 @@ def disambiguate(self, disambiguation: DA) -> "ScopedTreeRecord":
...
+        target_rank_treedef_id = ranks[target_rank_name]
+
+        # Query for the target rank treedef item
+        treedefitem_query = tree_rank_model.objects.filter(name=target_rank_name, treedef_id=target_rank_treedef_id)
+
+        if not treedefitem_query.exists() or treedefitem_query.count() != 1:
+            return self, WorkBenchParseFailure('invalidRank', {'rank': target_rank_name}, target_rank_name)
+
+        # Fetch the required data
+        treedefitems = list(tree_rank_model.objects.filter(treedef_id=target_rank_treedef_id))
```
realVinayak, line 194: sort by rankid

realVinayak (changes requested):

specifyweb/workbench/upload/treerecord.py
```diff
@@ -490,6 +490,8 @@ def _upload(self, to_upload: List[TreeDefItemWithParseResults], matched: Union[M
 
     def _do_insert(self, model, **kwargs):
         obj = model(**kwargs)
+        if hasattr(obj, 'fullname') and obj.fullname is None:
+            obj.fullname = obj.name
```
realVinayak, line 494: Why was this change needed? can you mention the unit test this aimed to solve?

realVinayak: by default, fullnames get reset at the very end anyways. If they don't, then you probably have a bug in changed_trees in upload.py

realVinayak (changes requested):

acwhite211:
@CarolineDenis @sharadsw 
Found a new front-end bug that happens when adding two rank columns with the same name but from different trees.
Added the following columns with CO base table:
1. cat number
2. determinations -> 1 -> taxon -> Fish taxonomy tree -> kingdom -> name
3. determinations -> 1 -> taxon -> Test tree -> kingdom -> name

When saving the upload plan, the front end request body to workbench/dataset is:
```
{"visualorder":[0,-1,1]}
```
It shouldn't have a negative number in the request body.  Seems to happen when ranks from different trees have the same name.

Also, the upload plan is missing a column.  This is probably caused by the rank names being the same, but in different trees.  Fixing this request might fix the visualorder problem.
```
{
  "uploadplan": {
    "baseTableName": "collectionobject",
    "uploadable": {
      "uploadTable": {
        "wbcols": {
          "catalognumber": "Cat #"
        },
        "static": {
          
        },
        "toOne": {
          
        },
        "toMany": {
          "determinations": [
            {
              "wbcols": {
                
              },
              "static": {
                
              },
              "toOne": {
                "taxon": {
                  "treeRecord": {
                    "ranks": {
                      "Kingdom": { # There needs to be two Kingdom columns
                        "treeNodeCols": {
                          "name": "Kingdom (2)"
                        },
                        "treeId": 5
                      }
                    }
                  }
                }
              }
            }
          ]
        }
      }
    }
  }
}
```

The treeRecord should be something like:
```
"treeRecord": {
  "ranks": {
    "Kingdom": {
      "treeNodeCols": {
        "name": "Kingdom"
      },
      "treeId": 1
    },
    "Kingdom (2)": {
      "treeNodeCols": {
        "name": "Kingdom"
      },
      "treeId": 5
    }
  }
```

Here are some screenshots:
image

image

specifyweb/workbench/upload/treerecord.py
acwhite211: Ya, I think there has to be a better place for this, but it solved an issue of the co->determination->taxon showing up correctly in the record results.  For some reason when the taxon.fullname was left null, then the issue occurred.  I'll look for a better solution.

specifyweb/workbench/upload/treerecord.py
acwhite211: From debugging, `value != ""` was found to be needed.  An example  row value I found was:
```
{'FossilKingdom': '', 'Kingdom': 'test5', 'RockKingdom': '', 'Cat #': '#########', 'MineralKingdom': ''}
```

acwhite211:
I've been messing around with the API and database more. In order to avoid overwriting keys in the upload plan request, we need to format the rank keys to include the treedefid in the cases where two ranks from different trees have the same name.  I was thinking a request format like this would work:
```json
{
  "uploadplan": {
    "baseTableName": "collectionobject",
    "uploadable": {
      "uploadTable": {
        "wbcols": {
          "catalognumber": "Cat #"
        },
        "static": {
          
        },
        "toOne": {
          
        },
        "toMany": {
          "determinations": [
            {
              "wbcols": {
                
              },
              "static": {
                
              },
              "toOne": {
                "taxon": {
                  "treeRecord": {
                    "ranks": {
                      "Kingdom~>1": {
                        "treeNodeCols": {
                          "name": "Kingdom - Fish taxonomy tree"
                        },
                        "treeId": 1
                      },
                      "Kingdom~>5": {
                        "treeNodeCols": {
                          "name": "Kingdom - Test Tree"
                        },
                        "treeId": 5
                      }
                    }
                  }
                }
              }
            }
          ]
        }
      }
    }
  }
}
```

The front-end needs to be changed to transform conflicting rank keys, so the two `Kingdom` ranks become `Kingdom~>1` and `Kingdom~>5`.  It's also problem a good idea to add tree context to the column name in the upload table view, so the two `Kingdom` column names become `Kingdom - Fish taxonomy tree` and `Kingdom - Test Tree`.

So the front-end needs to be able to adjust the ranks key names in the upload schema request, as well as when reloading the upload plan view in the UI.

Here are screenshots of an example:
image

image

emenslin (approved):
### Testing instructions
I created a database named 'sp7demofish_mots.sql' on the test panel, this has some sample Taxon trees created to use in this PR's test.  Use the screenshots for extra testing details.

Here is a test panel instance to use: https://sp7demofishmots-issue-4980.test.specifysystems.org

- [x] Setup the WB test by adding a new taxon tree and ranks.  Also, create a new CollectionObjectType.
- [x] Add new CollectionObjects with a Determination and CollectionObjectType. Do a workbench upload with CollectionObject as the base table.  Create workbench columns like "Det->Det 1->TestTree->TestKingdom->Name" (Edit: ignore this step, COType is now hidden in v7.9.7)
- [x] Add new CollectionObjects with Determination, this time without specifying a CollectionObjectType.
- [x] Add new Taxons to Ranks of different Taxon trees. Do a workbench upload with Taxon as the base table.  Create workbench columns like "Taxon->TestTree->TestKingdom->Name".
- [x] Add new CollectionObjects from Determinations.  Do a workbench upload with Determination as the base table.  Create workbench columns like "Determination->Taxon->TestTree->TestKingdom->Name".
- [x] Check that the Taxon Trees have been updated in the Tree Viewer. 

Looks good, everything now seems to be validated and can be uploaded properly. The behavior is a little weird without being able to select a COT but I know that will be fixed in the next release so I will approve this. I would like to make sure there's another testing review though just to make sure I didn't miss anything

lexiclevenger (approved):
### Testing instructions
I created a database named 'sp7demofish_mots.sql' on the test panel, this has some sample Taxon trees created to use in this PR's test.  Use the screenshots for extra testing details.

Here is a test panel instance to use: https://sp7demofishmots-issue-4980.test.specifysystems.org

- [x] Setup the WB test by adding a new taxon tree and ranks.  Also, create a new CollectionObjectType.
- [x] Add new CollectionObjects with a Determination and CollectionObjectType. Do a workbench upload with CollectionObject as the base table.  Create workbench columns like "Det->Det 1->TestTree->TestKingdom->Name" (Edit: ignore this step, COType is now hidden in v7.9.7)
- [x] Add new CollectionObjects with Determination, this time without specifying a CollectionObjectType.
- [x] Add new Taxons to Ranks of different Taxon trees. Do a workbench upload with Taxon as the base table.  Create workbench columns like "Taxon->TestTree->TestKingdom->Name".
- [x] Add new CollectionObjects from Determinations.  Do a workbench upload with Determination as the base table.  Create workbench columns like "Determination->Taxon->TestTree->TestKingdom->Name".
- [x] Check that the Taxon Trees have been updated in the Tree Viewer. 

---

Looks good, I didn't run into any issues!

Areyes42 (approved):
### Testing instructions

I created a database named 'sp7demofish\_mots.sql' on the test panel, this has some sample Taxon trees created to use in this PR's test. Use the screenshots for extra testing details.

Here is a test panel instance to use: [https://sp7demofishmots-issue-4980.test.specifysystems.org](https://sp7demofishmots-issue-4980.test.specifysystems.org/)

-  [x]  Setup the WB test by adding a new taxon tree and ranks. Also, create a new CollectionObjectType.
-   [x] Add new CollectionObjects with a Determination and CollectionObjectType. Do a workbench upload with CollectionObject as the base table. Create workbench columns like "Det->Det 1->TestTree->TestKingdom->Name" (Edit: ignore this step, COType is now hidden in v7.9.7)
-   [x] Add new CollectionObjects with Determination, this time without specifying a CollectionObjectType.
-   [x] Add new Taxons to Ranks of different Taxon trees. Do a workbench upload with Taxon as the base table. Create workbench columns like "Taxon->TestTree->TestKingdom->Name".
-   [x] Add new CollectionObjects from Determinations. Do a workbench upload with Determination as the base table. Create workbench columns like "Determination->Taxon->TestTree->TestKingdom->Name".
-   [x] Check that the Taxon Trees have been updated in the Tree Viewer.
***
Looks good on my end 👍

combs-a (dismissed):
### Testing instructions

- [x] Setup the WB test by adding a new taxon tree and ranks.  Also, create a new CollectionObjectType.
- [X] Add new CollectionObjects with Determination, this time without specifying a CollectionObjectType.
- [ ] Add new Taxons to Ranks of different Taxon trees. Do a workbench upload with Taxon as the base table.  Create workbench columns like "Taxon->TestTree->TestKingdom->Name".
- [X] Add new CollectionObjects from Determinations.  Do a workbench upload with Determination as the base table.  Create workbench columns like "Determination->Taxon->TestTree->TestKingdom->Name".
- [X] Check that the Taxon Trees have been updated in the Tree Viewer. 

Tested locally.

Ran into the issue where it uploads to the wrong tree if you have multiple trees involved in one row.

https://github.com/user-attachments/assets/e4498477-57ca-430e-b1ac-aa01e09a603f

Additionally, I separated out the above ranks into an expected format (and this may just be a database thing since many of my trees do not have root nodes) but it skipped over every column except for the end one. It did, however, upload it into the right tree.

https://github.com/user-attachments/assets/285397a7-255a-4d6c-89bc-4e4ee37d2723

It didn't happen when the ranks were different; it just happens when the ranks have the same name.

Another note: not sure how often it'll come up when it comes to user experience so it may not be necessary, but leaving the name of the tree in the column title might be better so that it's always clear which tree you're uploading into, even if you don't have duplicate ranks. Below is an example of why a lack of tree name could get confusing:

image

Everything is working well besides that--good work so far!

specifyweb/workbench/upload/treerecord.py
```diff
@@ -243,6 +243,9 @@ def rescope_tree_from_row(self, row: Row) -> Tuple["ScopedTreeRecord", Optional[
         # Based on the target treedef, get the treedefitems and root for the tree    
         treedefitems = list(tree_rank_model.objects.filter(treedef_id=target_rank_treedef_id).order_by("rankid"))
         root = tree_node_model.objects.filter(definition_id=target_rank_treedef_id, parent=None).first()
+        if root is None:
+            logger.warning(f"No root found for treedef {target_rank_treedef_id}")
+            return self, WorkBenchParseFailure('noRoot', {}, None)
```
realVinayak, line 248: I'll give you the benefit of doubt _again_, but workbench is supposed to handle no root cases (and it has done that since years ago you touched it). Now, after knowing this, don't you feel this is code smell? 

Don't you think that there could be a much bigger bug somewhere that we are not thinking about? Don't just patch things. I don't particularly care that we are behind timeline for this (what if the real bug causes a much bigger issue downstream, taking more testing resources + dev time?)

Just do things properly please, no hacks.

realVinayak (changes requested):

CarolineDenis:
NOTES: 
- We have been testing on several dbs adding new det to taxon table with a root node and with no root node, it creates as expected the 'Uploaded' root node for the ones that did not have a root. 
i.e:
 tree 1 Kingdom Name (root node present in the tree) 
 tree 1 Genus Name (root node present in the tree) 
 tree 2 Kingdom Name (root node NOT present in the tree) 
 tree 2 Phylum Name (root node NOT present in the tree) 
 
- the code for the 'no root' warning has been removed and the behavior will be the same as before (no root node, a new 'Uploaded' node is created) 

- there is some kind of issue with the herps trees in the db @combs-a

sharadsw (approved):

specifyweb/workbench/upload/scoping.py
```diff
@@ -206,19 +208,30 @@ def apply_scoping_to_treerecord(tr: TreeRecord, collection) -> ScopedTreeRecord:
...
 
     root = list(getattr(models, table.name.capitalize()).objects.filter(definitionitem=treedefitems[0])[:1]) # assume there is only one
 
+    scoped_ranks: Dict[TreeRankRecord, Dict[str, ExtendedColumnOptions]] = {}
+    for r, cols in tr.ranks.items():
+        if isinstance(r, str):
+            r = TreeRank.create(r, table.name, treedef.id if treedef else None, tr.base_treedef_id).tree_rank_record()
+        scoped_ranks[r] = {}
+        for f, colopts in cols.items():
+            scoped_ranks[r][f] = extend_columnoptions(colopts, collection, table.name, f)
```
realVinayak, line 228:
```suggestion
        scoped_ranks[r] = {f: extend_columnoptions(colopts, collection, table.name, f) for (f, colopts) in cols.items()}
```

specifyweb/workbench/upload/treerecord.py
```diff
@@ -55,20 +173,105 @@ def disambiguate(self, disambiguation: DA) -> "ScopedTreeRecord":
...
+
+        if len(set(tr.treedef_id for tr in self.ranks.keys())) == 1:
+            return self, None
+
+        # Get rank columns that are not null in the row
+        ranks_columns_in_row_not_null: List[TreeRankCell] = self._get_not_null_ranks_columns_in_row(row)
+
+        # Determine the target treedef based on the columns that are not null
+        targeted_treedefids = set([rank_column.treedef_id for rank_column in ranks_columns_in_row_not_null])
+        if targeted_treedefids is None or len(targeted_treedefids) == 0:
```
realVinayak, line 230: Why will `targeted_treedefids` be None? Remove of that check, or are you defending against python too 😅?

specifyweb/workbench/upload/treerecord.py
```diff
@@ -157,10 +361,25 @@ def _handle_row(self, must_match: bool) -> UploadResult:
...
+        def has_non_null_values(parse_results):
+            return any(
+                value is not None
+                for result in parse_results
+                for value in result.filter_on.values()
+            )
+
+        def get_parse_results(tdi):
+            tree_rank_record = TreeRankRecord(tdi.name, tdi.treedef_id)
+            return self.parsedFields.get(tree_rank_record, [])
```
realVinayak, line 373: Ugh, no. Please, just do a self.parsedFields[tree_rank_record] instead of get. You really gotta stop with all this defensive programming man. Yes, I can see that it'll always be there, but wouldn't you sleep well when you know that the code doesn't have undefined behavior? Isn't the absensce of tree_rank_record in self.parsedFields indicator of a bug somewhere else? Heck this should be an assert -- which you can enforce by using self.parsedFields[tree_rank_record]. 


You think you are creating a "robust" system but you are instead making a blunder where code works when it isn't supposed to (unless, right now, you can detail what happens in all the cases where tree_rank_record isn't present) -- especially to other parts.

specifyweb/workbench/upload/treerecord.py
```diff
@@ -157,10 +361,25 @@ def _handle_row(self, must_match: bool) -> UploadResult:
             return UploadResult(match_result, {}, {})
 
     def _to_match(self) -> List[TreeDefItemWithParseResults]:
+        def has_non_null_values(parse_results):
```
realVinayak, line 364: nice job breaking this up

specifyweb/workbench/upload/treerecord.py
```diff
@@ -55,20 +173,105 @@ def disambiguate(self, disambiguation: DA) -> "ScopedTreeRecord":
...
+    
+    def _get_not_null_ranks_columns_in_row(self, row: Row) -> List[TreeRankCell]:
+        # Get rank columns that are not null in the row
+        ranks_columns_in_row_not_null = []
+        for row_key, row_value in row.items():
+            if not row_value:
+                continue
+            for tree_rank_record, cols in self.ranks.items():
+                for col_name, col_opts in cols.items():
+                    formatted_column = f'{col_opts.column} - {col_name}' if col_name != 'name' else col_opts.column
```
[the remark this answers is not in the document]
realVinayak: Can you double check why col_ops.column is not enough (especially now when treename is included in columns)?

realVinayak (changes requested):
Thanks for testing out the root issue!!

specifyweb/workbench/upload/treerecord.py
```diff
@@ -223,7 +221,6 @@ def validate_rank(self, tree) -> None:
 class TreeRecord(NamedTuple):
     name: str
     ranks: Dict[Union[str, TreeRankRecord], Dict[str, ColumnOptions]]
-    base_treedef_id: Optional[int] = None
```
realVinayak: Awesome job removing this!

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingHelpers.ts
```diff
@@ -75,14 +94,44 @@ export const formattedEntry = `${schema.fieldPartSeparator}formatted`;
...
  *
  */
 export const formatTreeRank = (rankName: string): string =>
-  `${schema.treeSymbol}${rankName}`;
+  `${schema.treeRankSymbol}${rankName}`;
+
+// Delimiter used for rank name keys i.e: <rankname>~><treeId>
+const RANK_KEY_DELIMITER = '~>';
+
+/**
```
maxpatiiuk, line 117: if we have this, do we even need the treeDefinitionSymbol?

can we simplify things by adding a tree def id to every usage of a tree rank?

maxpatiiuk: once we add these symbols, it would be very hard to remove them as it would break previously creating upload plans.

if i am not wrong, it appears that this PR introduces two new mapping path symbols, when one would suffice.

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingHelpers.ts
```diff
@@ -75,14 +94,44 @@ export const formattedEntry = `${schema.fieldPartSeparator}formatted`;
...
+ * Returns a formatted tree rank name along with its tree id: (e.x $Kingdom => $Kingdom~>1)
+ * Used for generating unique key names when there are multiple trees with the same rank name.
+ * Opposite of getRankNameWithoutTreeId
+ * See: https://github.com/specify/specify7/pull/5091#issuecomment-2328037741
+ */
+export const formatTreeRankWithTreeId = (
+  rankName: string,
+  treeId: number
+): string => `${rankName}${RANK_KEY_DELIMITER}${treeId}`;
```
maxpatiiuk, line 126: it might be more intuitive to have tree id before the rank name
after all, rank name is inside a tree, not the other way around

is the a reason to do it otherwise?

maxpatiiuk (dismissed):
nice feature!

most of the code looks good.
have questions about the newly added mapping path symbols

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingHelpers.ts
sharadsw: We already do pass a treedefId to every usage of a tree rank. There was an issue where if the mapping has different trees with the same rank name, it would generate an incorrect upload plan because the key name would be the same for all mappings with that rank name and so the key would just get overwritten.

More info on that in these comments: 
https://github.com/specify/specify7/pull/5091#issuecomment-2327526079
https://github.com/specify/specify7/pull/5091#issuecomment-2328037741

`RANK_KEY_DELIMITER` is not really a part of the mapping but rather a part of the building/parsing the upload plan. Maybe placing it in `mappingHelpers.ts` was a bad idea and should get moved to `uploadPlanBuilder.ts`?

Other than that, I'm not sure we could use the same symbols for mapping and for generating unique key names as those are different usages in my opinion

melton-jason: For the usage of this, the test file for the uploadplan provides an example. 

https://github.com/specify/specify7/blob/10324c97095b34c296b37d3a18a354dba7a95e6c/specifyweb/frontend/js_src/lib/tests/fixtures/uploadplan.1.json#L276-L322

Here are my thoughts on this: 

This approach _does_ make it pretty simple to parse the tree definition id from the ranks, but I think it could be made a little more intuitive.

https://github.com/specify/specify7/blob/10324c97095b34c296b37d3a18a354dba7a95e6c/specifyweb/workbench/upload/treerecord.py#L137-L146

I think if we are modifying the upload plan this much already, it may be worthwhile to format the upload plan in a way which more closely resembles the added functionality. 

Particularly, I think an "inheritance" approach to mapping would make the most sense. 
Something like the following would be the most ideal I think: 

```json
{
   "ranks":{
      "Taxonomy":{
         "Class":{
            ...
         },
         "treeId":1 # You can keep treeId here if you want, but we should be able to identify the treeDef by name as well
      },
      "2" :{ # Or, maybe a little less intuitive: just use the treeDefId as a key 🤷 
         "Class":{
            ...
         }, 
         "treeId": 2 # and remove this
      }
   }
}
```

Of course, the implementation of this would be far more difficult to pull off so it doesn't have to follow the above exactly (`ranks` could be an array for example), but "it would be very hard to remove them as it would break previously creating upload plans" is a real concern (https://github.com/specify/specify7/pull/5091#discussion_r1751222126).

We are kind of repeating ourselves here as well, which can make things even more confusing: the treeDefId is in two places in the upload plan and _only_ needed in the rank name to resolve duplicate rank names across trees: which seems more like a design problem at odds with the newer approach than anything else.

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingHelpers.ts
melton-jason: Would something like the following be a little more intuitive and require the same amount of work as the current implementation to parse correctly? 

```json
"Taxonomy~>Subspecies": { 
       "treeNodeCols": { 
         "name": "Subspecies", 
         "author": "Subspecies Author" 
       }, 
       "treeId": 1 
     }, 
"1~>Subspecies": { # or alternatively
       "treeNodeCols": { 
         "name": "Subspecies", 
         "author": "Subspecies Author" 
       }, 
     }, 
```

sharadsw: I like the look of `<treename>~><rankname>`. Should be a simple enough change on the frontend
```JSON
"Taxonomy~>Subspecies": { 
       "treeNodeCols": { 
         "name": "Subspecies", 
         "author": "Subspecies Author" 
       }, 
       "treeId": 1 
     }, 
```
Would it break already existing upload plans on the backend? I think maybe a reason we went with rankname~>treeId was to be as backwards compatible as possible?

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingHelpers.ts
realVinayak: Actually, it's much more simple to make it an array of trees.

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingHelpers.ts
realVinayak: >  I think maybe a reason we went with rankname~>treeId was to be as backwards compatible as possible?

Agree with Jason, we shouldn't take a crappy design just because it "looks" similar to previous one

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingHelpers.ts
sharadsw: Shouldn't be too difficult to make it an array on the frontend

specifysoftware:

PR #5094, opened 2024-07-11
Add support for Geography Code at any level in queries

### LINKED ISSUE #5093: Add support for 'Geography Code' in Geography queries
Second component of https://github.com/specify/specify7/issues/4697
Expansion of https://github.com/specify/specify7/issues/4723

The [DWC term `countryCode`](https://dwc.tdwg.org/terms/#dwc:countryCode) recommends the following:

> Recommended best practice is to use an ISO 3166-1-alpha-2 country code. Recommended best practice is to leave this field blank if the dcterms:Location spans multiple entities at this administrative level or if the dcterms:Location might be in one or another of multiple possible entities at this level. Multiplicity and uncertainty of the geographic entity can be captured either in the term dwc:higherGeography or in the term dwc:locality, or both.

No matter which child geography you are exporting, if you want to map to `countryCode` it is necessary to show the geography code of the node at the country level.

**Requested By:** Many users, most recently Jordi at MCNB

grantfitzsimmons:
Fixes #5093

### Testing instructions

1. Build a query that retrieves the `geographyCode` field at any level in the Geography hierarchy.
2. Ensure the results are returned appropriately at any level.

See how it should appear here:

image

image


lexiclevenger (approved):
### Testing instructions

1. Build a query that retrieves the `geographyCode` field at any level in the Geography hierarchy.
2. Ensure the results are returned appropriately at any level.
---

Looks good!

Screenshot 2024-07-11 at 12 33 19 PM

emenslin (approved):
### Testing instructions

1. Build a query that retrieves the `geographyCode` field at any level in the Geography hierarchy.

- [x] Ensure the results are returned appropriately at any level.

Looks good, results return as expected 
Screenshot 2024-07-11 140021

alesan99 (approved):
### Testing instructions

- [x]  Build a query that retrieves the `geographyCode` field at any level in the Geography 
hierarchy.
- [x]  Ensure the results are returned appropriately at any level.

Working nicely👍👍👍

chrome_90Rk5gu81q

CarolineDenis (approved):

specifyweb/stored_queries/queryfieldspec.py
```diff
@@ -118,8 +121,7 @@ def from_stringid(cls, stringid, is_relation):
                 tree_rank = tree_id_match.group(1)
                 tree_field = 'ID'
             else:
-                tree_field_match = TAXON_FIELD_RE.match(extracted_fieldname) \
-                                   if node is datamodel.get_table('Taxon') else None
+                tree_field_match = TAXON_FIELD_RE.match(extracted_fieldname) if node is datamodel.get_table('Taxon') else GEOGRAPHY_FIELD_RE.match(extracted_fieldname) if node is datamodel.get_table('Geography') else None
```
melton-jason, line 124: (optional) 
I can envision this line getting pretty long (at maximum a clause for every tree). 

If you want to keep readability, can you extract this into a separate function or object (dictionary) which can resolve a Table (and/or field name depending on how you want to approach this) to the correct regular expression matches? 

The end result would be something like: 
```py
tree_field_match = something
```
instead of the long conditional value we have currently.

melton-jason (approved):
Nice work! 

Would it be worth hiding these queries by default in Specify 6? 
We have a function on the frontend, `isModern`, which sets the Query's `isFavorite` attribute to false (thus hiding it by default in Specify 6) if the Query contains any Specify 7 exclusive features. 

https://github.com/specify/specify7/blob/df4eccdebf7a6b2380bb4345178d83ff3f342e76/specifyweb/frontend/js_src/lib/components/QueryBuilder/helpers.ts#L379-L396

specifyweb/stored_queries/queryfieldspec.py
realVinayak: I think it'll be better to remove the restriction in the first place?

grantfitzsimmons: That's ultimately the next step. I was drafting a PR locally that does exactly that, but I haven't been satisfied with the implementation yet. Maybe I'll revisit this in the coming weeks. If another dev wants to do this, feel free! Just LMK.

grantfitzsimmons:
>We have a function on the frontend, isModern, which sets the Query's isFavorite attribute to false (thus hiding it by default in Specify 6) if the Query contains any Specify 7 exclusive features.

I didn't really see a need for it, see this:

image

specifyweb/stored_queries/queryfieldspec.py
realVinayak: @grantfitzsimmons excited to see your implementation! it might be easier for you to base it off https://github.com/specify/specify7/tree/wb_improvements, since it parses out tree ranks more nicely (instead of having the concept of tree_rank and tree_field, which makes things trivial things harder -- like if queryfield ends with a relationship or not), it makes a phantom query-builder relationship, making things much more simpler. 

The implementation I've in there also supports relatonships from rank (so you can do determination -> Species -> created by) which is not possible in the production implementation (not your fault, legacy implementation). The biggest blocker, as with most things, is a frontend implementation. I didn't touch it yet since geo guys will change relevant portions....

specifysoftware:

PR #5097, opened 2024-07-14
feat(KeyboardShortcuts): add customizable shortcuts for common actions

### LINKED ISSUE #1746: Add keyboard shortcuts for common actions

maxpatiiuk:
Begins to fix: #1746

### Checklist

- [x] Add automated tests

### Testing instructions

- Added `Keyboard Shortcuts` item to User tools - verify it looks good
- For each shortcut on the `Keyboard Shortcuts` page that has a default value assigned, make sure that the default keyboard shortcuts we have set do not already have an action assigned to them (we don't want to overwrite default browser or OS actions, unless we really have to, or unless user would expect us to do so)
  - i.e open any non-specify webpage (google.com) and see if this key does anything (this needs to be verified in non-specify pages because if specify collides with a system shortcut, specify will take priority, and so you may not even know that it collides)
  - Verify this on Windows in Chrome and Firefox
  - Verify this on macOS in Chrome and Safari
  - (if possible) Verify this on Linux in Firefox
- Check if any other common action is missing a keyboard shortcut that you believe we should add
- Verify the ability to set a keyboard shortcut to any page (see https://github.com/specify/specify7/pull/5097#issuecomment-2562140320)


grantfitzsimmons:
CSIRO mentioned users are having repetitive strain injuries due to this not being implemented

maxpatiiuk:
Bugs:
- Can add same shortcut twice for the same action
- When adding the keyboard shortcut, can't reach the "Remove" button with keyboard alone

maxpatiiuk:
Added ability to set shortcut to any URL:

https://github.com/user-attachments/assets/48a445d9-af3a-4332-ad55-53ccb1191221

Use cases:
- can set a keyboard shortcut to any menu item/user tool
- can set a keyboard shortcut to a record set/query that you use often
- can set a keyboard shortcut to an external website (morphosource/gbif/...)
  - this was trivial to implement, and makes the feature much more powerful - will see how it will end up being used
- brings us closer to https://github.com/specify/specify7/discussions/5686

By default, I only included keyboard shortcut to open the keyboard shortcut viewer, and will add shortcut to go to the home page. For the rest, we can let people set those to what makes sense to them (I avoided setting too many be default as things can conflict with os/browser/browser extensions, unless we go for 3-4 key shortcuts)

TODO:
- [ ] come up with a default keyboard shortcut to go to the home page

sharadsw (approved):

emenslin (changes requested):
Can we make the save button orange or something in the shortcuts to url dialog? I think if you're editing a url it can be easy to not click the save button assuming that when you close out of the dialog the bigger save button will be orange. Or we could also consider having some sort of blocker or dialog if that would make more sense. I feel like when first testing the dialog I didn't even notice the save button appeared next to it but that could just be me. 

https://github.com/user-attachments/assets/d47d1ecb-6829-447d-8cd0-d14e9f7dc968


On top of that if you create a new shortcut and don't press the gray save button, close the dialog, the save button on the naviagation menu turns orange but if you click that and go back to the shortcut it says 'No key binding assigned'

https://github.com/user-attachments/assets/51598069-891c-490a-9261-e186e4907622


If you add a shortcut and then try to add another without saving the dialog freaks out 

https://github.com/user-attachments/assets/8dbbdb6c-f7cd-411f-9040-8a1d68101294



Also I feel like this layout can be a little confusing with the buttons and the spacing. Is the intention of the add button after a shortcut has been made for people to be able to add multiple shortcuts to one URL? And if that is the case could they all be in one line instead of having the add button below the shortcut and add button? 
image


Also is there a plan to add more to this in the future? I feel like it looking just like preferences but only having the keyboard shortcuts looks a little odd especially considering the keyboard shortcuts opens a separate dialog so it just feels a little bare and confusing. 
image

CarolineDenis:
@maxpatiiuk, is this PR waiting on final testing or there are still some code changes needed?

maxpatiiuk:
> Can we make the save button orange or something in the shortcuts to url dialog?

Great point. Fixed

> On top of that if you create a new shortcut and don't press the gray save button, close the dialog, the save button on the naviagation menu turns orange

Fixed. The button no longer turns orange unless you hit "Save" inside the dialog

> If you add a shortcut and then try to add another without saving the dialog freaks out

Ah, geez. Fixed!

Now, as soon as you start editing any keyboard shortcut, all others become disabled - leaving you with only 2 options - complete entering the shortcut and hit "Save" or hit "Remove"

image

> Also is there a plan to add more to this in the future? I feel like it looking just like preferences but only having the keyboard shortcuts looks a little odd especially considering the keyboard shortcuts opens a separate dialog so it just feels a little bare and confusing.

Hmm, not sure now you got the page to look this way.

The keyboard shortcuts page should look more complete:

Screenshot 2025-07-26 at 08 51 40

It is possible that the page failed to render after this bug happened:
> If you add a shortcut and then try to add another without saving the dialog freaks out


---

Merge conflicts resolved. Ready for testing.

grantfitzsimmons:
Closing this in favor of https://github.com/specify/specify7/pull/8521

PR #5123, opened 2024-07-22
Reimplement timestamp logic

### LINKED ISSUE #5108: `TimestampModified` not being updated when modifying records
**Describe the bug**
When modifying an existing record, the `timestampModified` is not being updated.
This issue started in `v7.9.6` and was likely introduced due to the changes in https://github.com/specify/specify7/pull/4618

**To Reproduce**

1. Create a new record in any table that has a `timestampModified` field and save the record **or** find an existing record
3. Make modifications to the record
4. Save
5. See that the `timestampModified` value has not been changed

https://github.com/user-attachments/assets/ffe9acfa-3cf7-4149-b9f1-d73b2905acdb

**Expected behavior**
The `timestampModified` should be set to the current timestamp when changes are made.

**Reported By**
University of Kansas (Andy and Caleb)

Andy:
>Can indeed confirm that the record he mentions does not have an updated modified on date.  have also confirmed that this is happening in my database too.  I just edited record 1234 in my voucher collection and saved it and it still has a modified date of 10/17/2019.  Something is amiss!!

realVinayak:
Fixes #5108 

### Checklist

- [x] Add automated tests

### DIFF
--- a/specifyweb/specify/model_timestamp.py
+++ b/specifyweb/specify/model_timestamp.py
@@ -1,18 +1,35 @@
-from django.db import models
 from django.utils import timezone
-from django.conf import settings
+from django.db.models import Model
 
+timestamp_fields = [('timestampmodified', True), ('timestampcreated', False)]
 
+fields_to_skip = [field[0] for field in timestamp_fields if not field[1]]
 
-timestamp_fields = {'timestampmodified', 'timestampcreated'}
-
-def pre_save_auto_timestamp_field_with_override(obj):
+def save_auto_timestamp_field_with_override(save_func, args, kwargs, obj):
     # If object already is present, reset timestamps to null.
+    model: Model = obj.__class__
+    is_forced_insert = kwargs.get('force_insert', False)
+    fields_to_update = kwargs.get('update_fields', None)
+    if fields_to_update is None:
+        fields_to_update = [
+            field.name for field in model._meta.get_fields(include_hidden=True) if field.concrete
+            and not field.primary_key
+            ]
 
-    if obj.id is None:
-        return
-    
-    for field in timestamp_fields:
+    if obj.id is not None:
+        fields_to_update = [
+            field for field in fields_to_update
+            if field not in fields_to_skip
+            ]
+        
+    current = timezone.now()
+    _set_if_empty(obj, timestamp_fields, current, obj.pk is not None)
+    new_kwargs = {**kwargs, 'update_fields': fields_to_update} if obj.pk is not None and not is_forced_insert else kwargs
+    return save_func(*args, **new_kwargs)
+
+def _set_if_empty(obj, fields, default_value, override=False):
+    for field, can_override in fields:
         if not hasattr(obj, field):
             continue
-        setattr(obj, field, None)
+        if (override and can_override) or getattr(obj, field) is None:
+            setattr(obj, field, default_value)
--- a/specifyweb/specify/model_timestamp.py
+++ b/specifyweb/specify/model_timestamp.py
@@ -2,50 +2,17 @@ from django.db import models
 from django.utils import timezone
 from django.conf import settings
 
-from model_utils import FieldTracker
 
-def pre_save_auto_timestamp_field_with_override(obj, timestamp_override=None):
-    # Normal behavior is to update the timestamps automatically when saving.
-    # If timestampcreated or timestampmodified have been edited, don't update them to the current time.
-    cur_time = timezone.now()
-    timestamp_override = (
-        timestamp_override
-        if timestamp_override is not None
-        else getattr(settings, "TIMESTAMP_SAVE_OVERRIDE", False)
-    )
-    timestamp_fields = ['timestampcreated', 'timestampmodified']
+
+timestamp_fields = {'timestampmodified', 'timestampcreated'}
+
+def pre_save_auto_timestamp_field_with_override(obj):
+    # If object already is present, reset timestamps to null.
+
+    if obj.id is None:
+        return
+    
     for field in timestamp_fields:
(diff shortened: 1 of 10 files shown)


grantfitzsimmons:
This must be tested:

> image
> 
> ---
> 
> **Expected Behavior:**
> * When a column within a specified table is linked to either `timestampModified` or `timestampCreated`, it should not be automatically replaced upon uploading new data. 
> * In cases where a column is not associated with `timestampModified` or `timestampCreated`, the system should autonomously allocate both fields with the current timestamp. 
> * When new records are generated in other tables during the same upload process, if no column is directly linked to `timestampModified` or `timestampCreated`, the system should automatically assign both fields with the current timestamp.
> * `timestampModified` or `timestampCreated` should be assigned the current timestamp in all cases during WorkBench uploads **except** when a column is mapped to that field where that row contains a value

realVinayak:
we also don't need https://github.com/specify/specify7/blob/c3d0ab14521f03191f6a0f46f111c6ce213ac2fb/requirements.txt#L12 anymore, @acwhite211

emenslin (approved):
**Expected Behavior:**

- [x] When a column within a specified table is linked to either `timestampModified` or `timestampCreated`, it should not be automatically replaced upon uploading new data. 
- [x] In cases where a column is not associated with `timestampModified` or `timestampCreated`, the system should autonomously allocate both fields with the current timestamp. 
- [x] When new records are generated in other tables during the same upload process, if no column is directly linked to `timestampModified` or `timestampCreated`, the system should automatically assign both fields with the current timestamp.
- [x] `timestampModified` or `timestampCreated` should be assigned the current timestamp in all cases during WorkBench uploads **except** when a column is mapped to that field where that row contains a valu

I worked mostly with the CO table and the Agents table and everything looks good. All cases mentioned in the expected behavior work and the timestamp modified field is now updated when a record is edited.

acwhite211:
> we also don't need
> https://github.com/specify/specify7/blob/c3d0ab14521f03191f6a0f46f111c6ce213ac2fb/requirements.txt#L12
> anymore

Ya, since we're not using the FieldTracker anymore for tracking changes in the timestamp fields, we can remove this requirement.

alesan99 (approved):
- [x] When a column within a specified table is linked to either `timestampModified` or `timestampCreated`, it should not be automatically replaced upon uploading new data. 
- [x] In cases where a column is not associated with `timestampModified` or `timestampCreated`, the system should autonomously allocate both fields with the current timestamp. 
- [x] When new records are generated in other tables during the same upload process, if no column is directly linked to `timestampModified` or `timestampCreated`, the system should automatically assign both fields with the current timestamp.
- [x] `timestampModified` or `timestampCreated` should be assigned the current timestamp in all cases during WorkBench uploads **except** when a column is mapped to that field where that row contains a valu

I can confirm timestamps are being updated again for collection objects 👍 I uploaded the same data I did when comparing to edge.
The timestamps are also filled in accordingly when there is no corresponding columns.

CarolineDenis (approved):

Areyes42 (approved):
-   [x] When a column within a specified table is linked to either `timestampModified` or `timestampCreated`, it should not be automatically replaced upon uploading new data.
-   [x] In cases where a column is not associated with `timestampModified` or `timestampCreated`, the system should autonomously allocate both fields with the current timestamp.
-   [x] When new records are generated in other tables during the same upload process, if no column is directly linked to `timestampModified` or `timestampCreated`, the system should automatically assign both fields with the current timestamp.
-   [x] `timestampModified` or `timestampCreated` should be assigned the current timestamp in all cases during WorkBench uploads **except** when a column is mapped to that field where that row contains a value
***
I tested on COs, agent, and locality tables, and it looks like all the timestamps are now properly updated compared to the behavior in edge 👌

lexiclevenger (approved):
**Expected Behavior:**

- [x] When a column within a specified table is linked to either `timestampModified` or `timestampCreated`, it should not be automatically replaced upon uploading new data. 
- [x] In cases where a column is not associated with `timestampModified` or `timestampCreated`, the system should autonomously allocate both fields with the current timestamp. 
- [x] When new records are generated in other tables during the same upload process, if no column is directly linked to `timestampModified` or `timestampCreated`, the system should automatically assign both fields with the current timestamp.
- [x] `timestampModified` or `timestampCreated` should be assigned the current timestamp in all cases during WorkBench uploads **except** when a column is mapped to that field where that row contains a value
---
Looks good! Timestampmodified is also updated again when a record is modified through the data entry form.

specifyweb/specify/model_timestamp.py
```diff
@@ -1,51 +1,35 @@
...
+from django.db.models import Model
+
+timestamp_fields = [('timestampmodified', True), ('timestampcreated', False)]
+
+fields_to_skip = [field[0] for field in timestamp_fields if not field[1]]
+
+def save_auto_timestamp_field_with_override(save_func, args, kwargs, obj):
+    # If object already is present, reset timestamps to null.
+    model: Model = obj.__class__
+    is_forced_insert = kwargs.get('force_insert', False)
```
melton-jason, line 11: (just a note here) 

note that in the future `force_insert` will not always necessarily be a boolean. 

Looks like starting in `5.0`, it can be a tuple of Model-derived classes: https://docs.djangoproject.com/en/5.0/ref/models/instances/#ref-models-force-insert

Although even then I don't think this changes the functionality here unless an empty tuple is passed as a value to the `force_insert` kwarg (causing `not is_forced_insert` to return True). 

Something to keep in mind for when we update Django!

melton-jason (approved):
Looks solid!

specifyweb/specify/model_timestamp.py
realVinayak: Yep, it was fun trying to figure that out through the legacy docs

Issue #5098, opened 2024-07-15
Fix Schema Config and View Definition for New SP7 Tables


acwhite211:
For the new Specify 7 only tables added to v7.9.7, we need to add the new fields and tables to the Schema Configurations.

Fix error 'Unable to find SpLocaleConatiner for Role' error when trying to access any of the sp7_only tables in Schema Config.

We also need to add other SP7 only tables that are missing from the Schema config, like spUserRole and spAttachmentDataSet.

Also, we might need to create viewset definitions for the new tables as well.  This will avoid the 404 error for http://localhost/context/view.json?name=CollectionObjectGroupJoin.

**To Reproduce**

1. Go to Schema Config
2. Open up one of the new tables like CollectionObjectType, CollectionObjectGroupType, CollectionObjectGroup, and CollectionObjectGroupJoin.
3. See the error.
4. This also affects the new field collectionobjecttype in the CollectionObject table, and collectionobjecttype in the Collection table.

grantfitzsimmons:
<img width="728" alt="image" src="https://github.com/user-attachments/assets/5fbeac2c-c0c8-41a8-87cb-7d4afd2eb32c">

Note that these relationships should most likely not be visible by default for all collections.

acwhite211:
I wrote a function that creates default values for the schema config, generated from the datamodel.  There might be some fields in the schema config that got generated that we'll want to adjust.

CarolineDenis:
@sharadsw, 
Can you verify that this is fixed with your PR? If yes, can you close it?

CarolineDenis:
Closed in favor of https://github.com/specify/specify7/issues/5258

PR #5127, opened 2024-07-22
Add business rule to validate Determination taxon with COType

### LINKED ISSUE #5111: MOTs business logic: Determination subview must use same TaxonTree as COType to be valid
Related to #5107, part of #4982

> - When creating a new Determination record, the system should check the objectType.taxonTreeDef field to find the tree to search for the user's entered name when searching in the taxon or preferredTaxon query combo boxes

Alternatively, use business rules to validate whether chosen Taxon belongs to the correct COType tree definition

sharadsw:
Fixes #5111

Using business rules to check whether treeDef of a Determination Taxon is same as treeDef of the selected COType in a CO form. Save is blocked if they are not the same.

### Checklist

- [ ] Add automated tests

#### Set up the COType field in the CO form
Ideally, COType would be set as a picklist. However, until #5098 is fixed, it is not possible to configure one for COType. Instead for this PR, COType can be set up as a QCBX. 
- Go to App Resources > Type Searches (found under Global Resources)
- Add the following line in the XML editor:
```xml
<typesearch tableid="1015"  name="CollectionObjectType" searchfield="name" displaycols="name" format="%s" uifieldformatter="" dataobjformatter="" system="true"/>
```
- Click on the Form Definition tab
- Select the correct `.views` file for the collection you are in
- Go the the Collection Object form definition
- Add the following lines inside an existing `<row>`:
```xml
<cell type="label" labelfor="111"/>
<cell type="field" id="111" name="collectionObjectType" uitype="querycbx" />
```
Your CO form should look something like this
image

#### Testing the PR
- Make sure to use a db with multiple trees in the same discipline and multiple COTypes in the same collection
- Go to the CO form
- Select a COType 
- Add a determination
- Choose a Taxon which does not belong to the COType's treeDefiniton
- [ ] Verify save is blocked
- Choose a Taxon which belongs to the COType's treeDefiniton
- [ ] Verify save is enabled


specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -165,6 +169,37 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
...
+              .rgetPromise('collectionObjectType', true)
+              .then((coType: SpecifyResource<CollectionObjectType>) => {
+                /*
+                 * Have to set save blockers directly here to get this working.
+                 * Since following code has to wait for above rgetPromise to resolve, returning a Promise<BusinessRuleResult> for validation here is too slow and
+                 * does not get captured by business rules.
+                 */
+                if (
+                  idFromUrl(coType.get('taxonTreeDef')) ===
+                  idFromUrl(taxon?.get('definition') ?? '')
```
melton-jason, line 185: (optional) 
`idFromUrl` is not needed here. 

The following should be equivalent, 

```suggestion
                  coType.get('taxonTreeDef') ===
                  (taxon?.get('definition') ?? '')
```

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
melton-jason, line 1: If you have the time, can you write some unit tests for this new rule?

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -165,6 +169,37 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
                 .then(async (accepted) =>
                   accepted === null ? taxon : getLastAccepted(accepted)
                 );
+
+            const collectionObject = determination.collection
+              ?.related as SpecifyResource<CollectionObject>;
+            collectionObject
+              .rgetPromise('collectionObjectType', true)
+              .then((coType: SpecifyResource<CollectionObjectType>) => {
```
melton-jason, line 177: The `coType` can technically be `null` here, such as when the `collectionObjectType` is not yet set on the CollectionObject. 
(Not sure how we want to handle that case, but worthy a mention). 

Also, it is still possible to have an invalid `Determination -> taxon` here.
This Business Rule is only ran when the Determination changes. The user can change insert the Determination first, and _then_ change the `CollectionObejct -> coType` which uses a different TaxonTreeDef compared to the `Determination -> Taxon`.

melton-jason (changes requested):
While there is functionally nothing wrong with this Business Rule, I don't think it addresses what the requirement initially was for #5111:

> When creating a new Determination record, the system should check the objectType.taxonTreeDef field to find the tree to search for the user's entered name when searching in the taxon or preferredTaxon query combo boxes

I think the implementation of this would be to scope the results of the Taxon QueryComboBox to the correct tree depending on the `objectType`. 
Is this functionality going to be included in another PR? 

Right now from the changes in `stored_queries/execution.py`, I think the returned results from the QueryComboBox are from all Taxon Trees. 

Is this the new requirement? To return all Taxon records when searching and instead have this businessrule to determine whether the determination is correct or not? 

This is not the ideal from the user perspective, but it _does_ eliminate a lot of complexity.
Still, if this the approach we are going with then I think we can further improve quality-of-life (such as by ordering the results of the QueryComboBox by Tree). 

(Conceptually, wouldn't the ideal implementation default the filtering of results in QueryComboBoxes for Taxon records to the TaxonTreeDef on `Collection -> TaxonTreeDef`? On the Determination Form/SubView, we can then further filter the results based on the `CollectionObject -> coType -> taxonTreeDef`. 
Definitely not easy, but if we go with a simpler, less complex approach I think we should still try and make it as intuitive as possible).

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
sharadsw: `collectionObjectType` is set as a required field in the schema due to which it becomes a required field in the form as well, even if you set `isrequired=false` in the form definition. So ideally there would always be a `collectionObjectType` - especially if the plan is to set a default based on discipline?

#5100 should handle the case when `collectionObjectType` is changed after selecting a `Determination -> taxon`.

melton-jason: > So ideally there would always be a collectionObjectType

I agree that for existing CollectionObjects, collectionObjectType will always have a value. 
I specifically meant for new CollectionObjects which have not yet selected a `collectionObjectType` yet, the value will always be null/undefined. 

In this case, they can set a `Determination -> taxon` on the CollectionObject before they select a CollectionObjectType. 
We can let #5100 handle that case if we want. 

I think the intention was to have the CollectionObjectType on the CollectionObject default to whatever the Collection -> CollectionObjectType is. 

In that case, yes it would be another business rule (😵) to set that default, if we want to autofill the type on the CollectionObject.
(Either let, or always have the user explicitly select which CollectionObjectType to use. Whichever approach fits more with our requirements).

sharadsw: Setting default cotype with business rule should be super straightforward. When creating a new CO:
- Default COType will be set already (maybe a different PR for that in case we go with scoping QCBX instead of this PR?)
- Determination -> taxon will be validated to make sure taxon treeDef is same as coType treeDef
- Changing COType after that will clear all fields (won't clear coType) due to #5100 

Sounds good to me.

emenslin (changes requested):
See https://github.com/specify/specify7/pull/5036#pullrequestreview-2210237277

CarolineDenis (approved):

emenslin (approved):
#### Testing the PR
- Make sure to use a db with multiple trees in the same discipline and multiple COTypes in the same collection
- Go to the CO form
- Select a COType 
- Add a determination
- Choose a Taxon which does not belong to the COType's treeDefiniton
- [x] Verify save is blocked
- Choose a Taxon which belongs to the COType's treeDefiniton
- [x] Verify save is enabled

Looks good, works as expected

lexiclevenger (approved):
#### Testing the PR
- Make sure to use a db with multiple trees in the same discipline and multiple COTypes in the same collection
- Go to the CO form
- Select a COType 
- Add a determination
- Choose a Taxon which does not belong to the COType's treeDefiniton
- [x] Verify save is blocked
- Choose a Taxon which belongs to the COType's treeDefiniton
- [x] Verify save is enabled
---
Looks good!
Screenshot 2024-08-01 at 3 45 45 PM

melton-jason (approved):

specifysoftware:

PR #4764, opened 2024-04-09
Add Datamodel and Django Model definitions

acwhite211:
Fixes #4358 and #2813

This PR adds datamodel and Django model code definitions, instead of relying on reading the 'specify_datamodel.xml' file from Specify 6.  So, the schema is now statically defined in our python Django code, instead of dynamically be created at runtime.  There is also code that can generated the static python code from the 'specify_datamodel.xml', so we can regenerate the code again in the future easily if we need to.  Unit tests have been created to ensure the equivalence  between the old and new datamodels.  There was another branch `django-schema` that had a lot of code from the development process, but I cleaned it up and simplified it into this branch.  There were some problems with adding the sql alchemy schema model classes, this isn't a pressing matter so this is something we can do in the future.

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [x] Add automated tests
- [x] Add relevant issue to release milestone

### Testing instructions

Unfortunately there isn't anything specific to test.  I would say the best way to go about testing this from the UX side is to play around with forms and queries, and make sure no database table or field related errors occur.  Also, for the sp7_only tables, test them out in the query builder.

sp7_only tables:
- Spuserexternalid
- Spattachmentdataset
- UniquenessRule
- UniquenessRuleField
- Message
- Spmerging
- UserPolicy
- Role
- LibraryRole
- UserRole
- RolePolicy
- LibraryRolePolicy
- Spdataset

PR Testing Checklist:
- [x] QB works on tables like Agent, Collecting Object, and Locality
- [x] QB works on the new sp7_only tables
- [x] The Database Schema page works without errors (UserTools->DatabaseSchema) localhost/specify/data-model
- [x] Test out simple agent merging to make sure it still works properly

PR #5133, opened 2024-07-23
Display system tables by default in query list menu option

### LINKED ISSUE #5128: `Sp Audit Log` no longer shows up in query builder
**Describe the bug**
The sp audit log table does not show up in the list of tables to query on and the only way to query on it is by editing the url or using the embedded query builder to do it by record 

**To Reproduce**

1. Go to queries 
2. Click on `new`
3. Select the edit pencil next to `new query`
4. Scroll down to the bottom of each list 
5. See `Sp Audit Log` (and other tables) do not show up on either list 

**Expected behavior**
All tables should show up in either the selected or possible behavior 

7.9.3.1 (last release it was present) 
Screenshot 2024-07-22 154654

**Screenshots**
edge
Screenshot 2024-07-22 154710

Screenshot 2024-07-22 154807

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge (goes back to 7.9.4) 
- Collection name: mammals
- User Name: spadmin

CarolineDenis:
Fixes #5128

### Checklist

### Testing instructions

- Go to queries
- Click on new
- Select the edit pencil next to new query
- [ ] Verify Sp Audit Log (and other tables) show

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
+++ b/specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
@@ -82,11 +82,15 @@ export function tablesFilter(
   showAdvancedTables: boolean,
   { name, overrides }: SpecifyTable,
   // Don't exclude a table if user already has it selected
-  selectedTables: RA<keyof Tables> | undefined = undefined
+  selectedTables: RA<keyof Tables> | undefined = undefined,
+  parent?: 'DataEntryList' | 'QueryList'
 ): boolean {
   if (selectedTables?.includes(name) === true) return true;
 
-  const isRestricted = overrides.isHidden || overrides.isSystem;
+  const isRestricted =
+    parent === 'QueryList'
+      ? overrides.isHidden
+      : overrides.isHidden || overrides.isSystem;
   if (!showHiddenTables && isRestricted) return false;
   const hasAccess = hasTablePermission(name, 'read');
   if (!showNoAccessTables && !hasAccess) return false;
--- a/specifyweb/frontend/js_src/lib/components/DataEntryTables/Edit.tsx
+++ b/specifyweb/frontend/js_src/lib/components/DataEntryTables/Edit.tsx
@@ -130,6 +130,7 @@ function CustomEditTables({
           : formsText.configureInteractionTables()
       }
       isNoRestrictionMode={false}
+      parent="DataEntryList"
       tables={tables}
       onChange={handleChange}
       onClose={handleClose}
--- a/specifyweb/frontend/js_src/lib/components/Toolbar/QueryTablesEdit.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Toolbar/QueryTablesEdit.tsx
@@ -31,6 +31,7 @@ export function QueryTablesEdit({
       defaultTables={defaultQueryTablesConfig}
       header={queryText.configureQueryTables()}
       isNoRestrictionMode={isNoRestrictionMode}
+      parent="QueryList"
       tables={tables}
       onChange={setTables}
       onClose={handleClose}
@@ -51,6 +52,7 @@ export function TablesListEdit({
   tables: selectedTables,
   onChange: handleChange,
   onClose: handleClose,
+  parent,
 }: {
   readonly isNoRestrictionMode: boolean;
   readonly defaultTables: RA<keyof Tables>;
@@ -58,11 +60,19 @@ export function TablesListEdit({
   readonly tables: RA<SpecifyTable>;
   readonly onChange: (table: RA<SpecifyTable>) => void;
   readonly onClose: () => void;
+  readonly parent: 'DataEntryList' | 'QueryList';
 }): JSX.Element {
   const selectedValues = selectedTables.map(({ name }) => name);
   const allTables = Object.values(genericTables)
     .filter((table) =>
-      tablesFilter(isNoRestrictionMode, false, true, table, selectedValues)
+      tablesFilter(
+        isNoRestrictionMode,
+        false,
+        true,
+        table,
+        selectedValues,
+        parent
+      )
     )
     // TODO: temp fix, remove this, use to hide geo tables for COG until 9.8 release
     .filter((table) => !HIDDEN_GEO_TABLES.has(table.name))


maxpatiiuk (commented):
instead of this, should we simply make audit log tables not be system tables?
there are many system tables that rightfully should remain hidden (at very least to reduce noise and confusion). there are also many sp6-only tables that we don't want to see in Sp7

CarolineDenis:
NOTES: 

Should be in the query list: 

- Collection
- Discipline
- Division
- SpAuditLog

emenslin (approved):
### Testing instructions

- Go to queries
- Click on new
- Select the edit pencil next to new query
- [x] Verify Sp Audit Log - Collection -  Discipline - Division show

Looks good, the 4 tables above now show in the query builder
Screenshot 2024-07-24 105220

combs-a (approved):
- Go to queries
- Click on new
- Select the edit pencil next to new query
- [X] Verify Sp Audit Log - Collection -  Discipline - Division show

Looks good, everything is showing and you can query just fine

image

image

Areyes42 (approved):
### Testing instructions

-   Go to queries
-   Click on new
-   Select the edit pencil next to new query
-   [x] Verify Sp Audit Log - Collection - Discipline - Division show
***
Looks good, and they still query correctly!

sharadsw (approved):

grantfitzsimmons (approved):

specifyweb/specify/datamodel.py
```diff
@@ -6192,7 +6192,7 @@
         classname='edu.ku.brc.specify.datamodel.SpAuditLog',
         table='spauditlog',
         tableId=530,
-        system=True,
+        system=False,
```
maxpatiiuk, line 6195: do we also need to remove it from here?
https://github.com/specify/specify7/blob/69391723254d4ea71df4701a869c225dc00b80dd/specifyweb/specify/load_datamodel.py#L484C6-L484C12

if so, why are there two places designating what tables are system? that is a recipe for bugs

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
```diff
@@ -85,6 +90,7 @@ export function tablesFilter(
   selectedTables: RA<keyof Tables> | undefined = undefined
 ): boolean {
   if (selectedTables?.includes(name) === true) return true;
+  if (tablesToIncludeInQueryList.has(name)) return true;
```
maxpatiiuk, line 93: "tablesToIncludeInQueryList" but this function is used in more than query list...
it's used everywhere were tables are shown in specify

instead, this function has a `selectedTables` list that you can use to allow list certain tables, regardless of all other filters

maxpatiiuk: then, whether to force-include hierarchy tables like Collection/Discipline/Divison depends on the use case - is the use case for user to be able to edit them? if so, you should probably only include them if the user has edit permission to them.

or you could simply say that if user is a super admin then show all hidden tables even if not explicitly said so in the user preferences

maxpatiiuk: or do you want these tables to be included in all lists across specify (in workbench mapper too and everywhere else?). if so, we need to revisit the definition of system tables. a system table seemed to be before a table that user ordinary has no need to touch - if that is not the case for these tables, then should we mark them as not system?

or is there a reason to keep them as system on the back-end? in that case you can mark them as non-system on the front-end only in this file:

https://github.com/specify/specify7/blob/issue-5138/specifyweb/frontend/js_src/lib/components/DataModel/schemaOverrides.ts

maxpatiiuk (changes requested):
even if this fixes the immediate bug, let's not quickly jump to creating a new list of tables - every such list we create is one more place where things can go out of date or bugs can slip in. it would be better if we can adjust the existing system to fit the new requirement, rather than add a new system as an extra layer on top...

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
CarolineDenis: yes those 3 can be included in all places. Changing the isSystem in the schemaOverrides creates a lot of issues in other parts of sp and it doesn't worth the trouble for just 3 tables to add in a list, hence our choice to go with this short list variable.

CarolineDenis: the other thing is that we do not want all the tables but only those 4 which would not be the case if we use the isAdmin check right?

CarolineDenis: maybe adding the tables for admin managers? (pushed a test below)

specifyweb/specify/datamodel.py
melton-jason: Yes, it can (and should) be removed from there.
Although thankfully, it wouldn't have caused any bugs. 

Prior to #4764, that was used to mark which tables are systems when the datamodel was autogenerated from the `specify_datamodel.xml` file from Specify 6. 
Now, it is only used when the datamodel code is regenerated (which should now never be done without some refactoring and a note of which changes we have since made to the `datamodel.py` file). 

I had hoped that the implementation of #4764 would allow the Models and/or Datamodel to be autogenerated instead of manually edited (much like how the frontend Datamodel is generated), but that is not the case currently.

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
```diff
@@ -85,8 +86,12 @@ export function tablesFilter(
   selectedTables: RA<keyof Tables> | undefined = undefined
 ): boolean {
   if (selectedTables?.includes(name) === true) return true;
+  const userIsAdmin =
+    userInformation.isadmin && userInformation.usertype === 'Manager';
```
maxpatiiuk, line 90: usertype is a sp6 concept - sp7 only users may not have it set
could yo ucheck only for `userInformation.isadmin`?

maxpatiiuk (approved):

specifyweb/frontend/js_src/lib/components/SchemaConfig/Tables.tsx
CarolineDenis: ok :+1:

specifysoftware:

PR #5137, opened 2024-07-24
Add search on synonyms in the QB

### LINKED ISSUE #752: Add ability for query to find synonymized taxa
Currently, 7 only searches on the value that is entered into the search fields.  The 6 query has a checkbox that allows users to also bring back any taxa that have been synonymized with the taxon search term.
So, if the taxa entered in the search is invalid, it will also bring back the valid term.
If the taxa entered in the search is valid, it will also bring back the invalid terms synonymized with it.

CarolineDenis:
Fixes #752

### Checklist

### Testing instructions

- open query builder and select a tree table 
- make a search on any rank using the name as a field (chose a node that has a synonym in your tree) 
- check the box search synonyms 
- run the query 
- [ ] verify that the node is being return as well as its synonym


specifyweb/stored_queries/query_construct.py
```diff
@@ -17,7 +17,12 @@ def _safe_filter(query):
         return query.first()
     raise Exception(f"Got more than one matching: {list(query)}")
 
-class QueryConstruct(namedtuple('QueryConstruct', 'collection objectformatter query join_cache tree_rank_count internal_filters')):
+def get_treedef(collection, tree_name):
+    return (collection.discipline.division.institution.storagetreedef
+            if tree_name == 'Storage' else
+            getattr(collection.discipline, tree_name.lower() + "treedef"))
```
grantfitzsimmons, line 23: Just curious– what is the purpose of this check? 

`storagetreedef` is the same even if you construct it with 
```py
getattr(collection.discipline, tree_name.lower() + "treedef"))
```

acwhite211 (approved):

Iwantexpresso (approved):
- [x] verify that the node is being return as well as its synonym 
image

this looks great so far!

grantfitzsimmons (changes requested):
The fundamental issue here is that 'Search Synonyms' is used most often for searching _Collection Objects_ rather than tree names via a tree base table. The inadequate request this PR is based on unfortunately did not clarify this well enough.

See https://github.com/specify/specify7/issues/752#issuecomment-2754634819:

> Hey guys, this one is quite important. If I create a collection object query and search for a preferred taxon name, it does not return collection objects with the original taxon name that are different (i.e. junior synonyms) to the preferred taxon name. The functionality is as expected in Specify 6.
>
> Also related to this is that when querying collection objects from a taxon name in the taxon tree, the behaviour differs from Specify 6. In 6 you get records where either the taxon name or the preferred taxon name match the searched name, but in 7 it only searches on taxon name, so searches for a senior synonym don't return collection objects identified as the junior synonyms.

See that if I search for a synonym in a database (e.g. `Opsanus tau`) with Collection Object as the base table, I can still check "Search Synonyms":
Image

In the results:
Image

This same behavior is expected in 7.

---

Using a copy of `herb_rbge`, I see the following:

**Left:** `v7.11.2.1`, **Right:** `issue-752`
Image

Same query, exported and imported in both: [Taxon Synonyms.json](https://github.com/user-attachments/files/24412060/Taxon.Synonyms.json)

On the `v7.11.2.1`, only results where synonyms do have an ID that isn't empty are returned (`67532`). On `issue-752`, all taxon records are returned (`370611`). 

Image

In neither scenario is "Search Synonyms" checked.

---

Searching directly for a `Full Name` for a synonymized node (e.g. `Ericaceae`) does yield correct results when querying `Taxon` directly:

Image
Image

bhumikaguptaa (approved):
- [X] verify that the node is being returned as well as its synonym

I am able to see the node and the synonym. Looks good to me!

Image

PR #5139, opened 2024-07-24
Handle 404 errors in QueryComboBox

### LINKED ISSUE #5138: Handle 404 error in QueryComboBox
QueryComboBox fetches the resource for the field used for displaying the QCBX value. It does not account for 404 errors when fetching which leads to an uncaught exception, redirecting to a 404 page in a production environment (see #4955).

To avoid this, QCBX needs to handle the 404 error and clear its value and set a save blocker indicating the value is invalid.

sharadsw:
Fixes #5138 

Adding a `strict` param to `rgetPromise` like `fetchResource` does to pass 404 as an expected error and avoiding an uncaught exception
https://github.com/specify/specify7/blob/6a364e00f96d5c81f0853c65b3d12ed0918fc202/specifyweb/frontend/js_src/lib/components/DataModel/resource.ts#L42-L63

- QCBX will now always fetch its resource with `strict=false`. We will no longer see a 404 crash on production when it fetches an invalid resource. Instead, the QCBX value will be cleared and there should be save blocker indicating there is an invalid value
- This should not break existing usages of `rgetPromise` as the strict param is only optional and defaults to true. Older usages will still have the old behavior

Before this PR - 404 crash on invalid resource:

https://github.com/user-attachments/assets/98520da5-5a55-4322-802f-fe6bf7ce6f75

In this PR:

https://github.com/user-attachments/assets/a1be8084-b48a-4e37-be79-fb2a6f3df4ee

### Checklist

- [ ] Add automated tests

### Testing instructions

- Initialize a QCBX to an invalid resource
- For example:
  - Go to a CO form definition
  - Initialize cataloger to an invalid resource. You can do this with by adding a `initialize` property to the cataloger cell
 ```
initialize="default=/api/specify/agent/2222212211/"
```
  - Your cell could look something like:
 ```
<cell type="field" id="111" name="cataloger" uitype="querycbx" initialize="default=/api/specify/agent/2222212211/"/>
```
- Go to Data Entry > Collection Object
- [ ] Verify app does not crash
- [ ] Verify the QCBX is empty
- [ ] Verify there is a save blocker saying `Invalid value`


specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
```diff
@@ -171,30 +172,42 @@ export function QueryComboBox({
...
+                  };
+                }
+              })
+              .catch((_) => {
+                setValidation([formsText.invalidValue()]);
+                return {
+                  label: localized(''),
+                  resource: undefined,
+                };
+              })
```
sharadsw, line 210: As of now, the code never reaches this catch block as the 404 response is not considered to be an error. There is no 404 dialog in RSS Export Feed anymore (context: #4955) and the QCBX instead displays the formatted invalid resource and pretends it as valid. 

We can hit the catch block if we also throw an error when `status === Http.NOT_FOUND` in line 66 over here:
https://github.com/specify/specify7/blob/6a364e00f96d5c81f0853c65b3d12ed0918fc202/specifyweb/frontend/js_src/lib/utils/ajax/backboneAjax.ts#L64-L69
In which case, the error gets caught in QCBX, value gets cleared and a save blocker is set but somehow it does not trigger an unload protect and the blocker is not visible until you click on the field. My guess is there needs to be a re-render in the parent component for the unload protect to trigger.

specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.ts
```diff
@@ -747,13 +748,15 @@ export const ResourceBase = Backbone.Model.extend({
...
-        }));
+      return (this._fetch = hijackBackboneAjax(
+        options === undefined || options.strict ? undefined : [Http.NOT_FOUND],
+        () =>
+          Backbone.Model.prototype.fetch.call(this, options).then(() => {
+            this._fetch = null;
+            // BUG: consider doing this.needsSaved=false here
+            return this;
+          })
+      ));
```
sharadsw, line 759: I'm not sure what the side-effects are if we make the 404 change in `backboneAjax.ts` I mentioned in the other comment. There doesn't seem to be any way to avoid the 404 redirect without using `hijackBackboneAjax()`. Without `hijackBackboneAjax()`, the error does not get passed up the chain high enough to reach QCBX. 

The error is initially caught here: https://github.com/specify/specify7/blob/6a364e00f96d5c81f0853c65b3d12ed0918fc202/specifyweb/frontend/js_src/lib/utils/ajax/response.ts#L76-L78

and in a prod env, it redirects to a 404 page and the error seems to remain uncaught:
https://github.com/specify/specify7/blob/013f81295c9e83cd6bd86a565d9b21950b5a76da/specifyweb/frontend/js_src/lib/components/Errors/FormatError.tsx#L152-L162

sharadsw (commented):
Here's what I have so far @maxpatiiuk. Would appreciate some pointers

specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.ts
maxpatiiuk: > I'm not sure what the side-effects are if we make the 404 change in backboneAjax.ts I mentioned in the other comment. There doesn't seem to be any way to avoid the 404 redirect without using hijackBackboneAjax(). Without hijackBackboneAjax(), the error does not get passed up the chain high enough to reach QCBX.

I think you should be calling handleAjaxError() with expected errors [] not in this file but in the query combo box file. here: https://github.com/specify/specify7/blob/69391723254d4ea71df4701a869c225dc00b80dd/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx#L174-L175

that way, this change will not unexpectedly affect any other place, and you will have an error thrown like expected

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
maxpatiiuk: after adding hijackBackboneAjax like described in https://github.com/specify/specify7/pull/5139#discussion_r1690614195, it looks like the save blocker is set correctly and the save buttons turn red:

Screenshot 2024-07-24 at 17 17 51

could you verify if you also get this behavior after making that change?

maxpatiiuk:
For future reference, here is how you can temporary disable MySQL integrity protection:

https://stackoverflow.com/a/15501754/8584605

useful for testing 404 errors on related records

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
sharadsw: Hmm, save does get blocked correctly in the CO form. However, in a prod environment the page redirects to a 404 page, which doesn't solve the issue in #4955.

Save blocker still doesn't work correctly in RSS export feed although the values did get cleared
image
and the blocker is only visible after clicking on the field --- but it doesn't matter as in prod the page will still redirect to 404
image

Making the change here: https://github.com/specify/specify7/pull/5139#discussion_r1690241121 along with wrapping `hijackBackboneAjax` over the resource fetch directly in `resourceApi.ts` helps avoid the 404

sharadsw: I believe if we use `hijackBackboneAjax` higher up the call stack as opposed to wrapping it over `resource.fetch()`, it gets set as undefined before all chained promises are resolved
https://github.com/specify/specify7/blob/8d0052c7a8f37a1f2eb14a1e730279bd117a7608/specifyweb/frontend/js_src/lib/utils/ajax/backboneAjax.ts#L29

and then the `expectedErrors` over here is an empty list instead of `[404]` and so it gets treated as an uncaught error
https://github.com/specify/specify7/blob/8d0052c7a8f37a1f2eb14a1e730279bd117a7608/specifyweb/frontend/js_src/lib/utils/ajax/response.ts#L11-L26

maxpatiiuk: > Hmm, save does get blocked correctly in the CO form. However, in a prod environment the page redirects to a 404 page, which doesn't solve the issue in https://github.com/specify/specify7/issues/4955.

ah, sorry! I forgot that the behavior is different in production

---

You are correct
The cause of the issue is that rget is actual awaiting two promises:
1. Loading the resource itself (even if it's already loaded, this still creates a promise):
2. Loading the related resource:
https://github.com/specify/specify7/blob/9faedc70ac8cb78dbd676a840770e8ebf1476500/specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.ts#L521-L522

the assumption that `hijackBackboneAjax()` makes is that the variables it sets (`expectedErrors`, among others) are going to be retrieved in the same cycle - but that is not the case as the rget has two promises, and the one we are interested in comes 2nd.

The issue is that `hijackBackboneAjax()` is called concurrently by multiple query combo boxes (i.e accession), and so if accession resource is fetched earlier, the `hijackBackboneAjax()` does cleanup afterward, unsettling `expectedErrors` and other variables - before the collector request was sent out with correct `expectedErrors`

In all other places where hijackBackboneAjax is used, it's used directly on `.fetch()`, so this issue is not present.

two options:
- 1st (not great): make rget only call fetch() if resource is not yet fetched. not great as then this bug would still be present when working with resource that is not yet fetched
- 2nd (close to what you have right now): hijackBackboneAjax itself is a hack. relying on global variables is not great, and even worse when working with async actions. a better solution is to pass fetch options along with requests that do the fetching.

we don't want to do too many changes to resourceApi.ts as we are hoping to do major refactor to it soon, but in an ideal world the change you made with introducing the "options" parameter would be made for all methods in resourceApi that make network requests.

for now, could you please make the following changes:
- only call hijackBackboneAjax() in .fetch() if some options were actually provided - a performance optimization since .fetch() is very much on a hot path
- post a comment in https://github.com/specify/specify7/issues/4286 summarising the issue ("we need more control over how resourceApi makes network requests ...") and say that the new resourceApi API we will design should have fetching options on all async methods. since the new API will be using `ajax()` directly, it will be able to easily pass on the options, without the need for hacky `hijackBackboneAjax()`

specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.ts
```diff
@@ -746,14 +747,21 @@ export const ResourceBase = Backbone.Model.extend({
...
+    else {
+      const fetchCallback = () =>
+        Backbone.Model.prototype.fetch.call(this, options).then(() => {
           this._fetch = null;
           // BUG: consider doing this.needsSaved=false here
           return this;
-        }));
+        });
+      if (options === undefined || options.strict)
+        return (this._fetch = fetchCallback);
```
maxpatiiuk, line 758: shouldn't this be a function call instead?

```suggestion
        return (this._fetch = fetchCallback());
```

we don't have any type-checking in this file since there would be a few thousand errors)
till https://github.com/specify/specify7/issues/4286

maxpatiiuk (approved):
the code looks good,
and so this might now be ready for testing

though, could you please leave a comment on the related issue like mentioned in https://github.com/specify/specify7/pull/5139/files#r1692360885 ?

CarolineDenis (approved):

melton-jason (changes requested):
Nice work! 🥳 
Almost there; I really only have one concern with a `fetch` call in `getRelated`

melton-jason (approved):

emenslin (approved):
- [x] Verify app does not crash
- [x] Verify the QCBX is empty
- [x] Verify there is a save blocker saying `Invalid value`

Looks good, no more 404 error.
Screenshot 2025-05-15 113903

lexiclevenger (approved):
- [x] Verify app does not crash
- [x] Verify the QCBX is empty
- [x] Verify there is a save blocker saying `Invalid value`
---
Looks good!

Screenshot 2025-05-15 at 1 43 58 PM

PR #5140, opened 2024-07-24
Update WB line headers when changing mapping field

### LINKED ISSUE #1858: WorkBench mapper not saving assigned determination number
chrome_88Us8P2sTu

In my data set, I mapped the field text8 to the #3 Determination and it automatically reset to the #2 Determination after saving.

version: 7.7.0
browser: Chrome

There were 3 mapped determinations that included other fields.

Both the mapping and data set are here:
[WVA 0-10000 (7-18 Export).zip](https://github.com/specify/specify7/files/9144601/WVA.0-10000.7-18.Export.zip)

CarolineDenis:
Fixes #1858

### Checklist

### Testing instructions
TODO

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/WbPlanView/mappingReducer.ts
+++ b/specifyweb/frontend/js_src/lib/components/WbPlanView/mappingReducer.ts
@@ -326,14 +326,23 @@ export const reducer = generateReducer<MappingState, MappingActions>({
         mappingView: newMappingPath,
       };
 
+    const isMappingComplete = mappingPathIsComplete(newMappingPath);
+
+    const newHeader = isMappingComplete
+      ? generateMappingPathPreview(action.parentTableName!, newMappingPath)
+      : state.lines[line].headerName;
+
+    const updatedLines = deduplicateMappings(
+      modifyLine(state, line, {
+        mappingPath: newMappingPath,
+        headerName: newHeader,
+      }),
+      state.openSelectElement?.line ?? false
+    );
+
     return {
       ...state,
-      lines: deduplicateMappings(
-        modifyLine(state, line, {
-          mappingPath: newMappingPath,
-        }),
-        state.openSelectElement?.line ?? false
-      ),
+      lines: updatedLines,
       autoMapperSuggestions: undefined,
       changesMade: true,
       mappingsAreValidated: false,
--- a/specifyweb/frontend/js_src/lib/components/WbPlanView/mappingReducer.ts
+++ b/specifyweb/frontend/js_src/lib/components/WbPlanView/mappingReducer.ts
@@ -27,6 +27,7 @@ import type {
   SelectElementPosition,
 } from './Mapper';
 import { emptyMapping } from './mappingHelpers';
+import { generateMappingPathPreview } from './mappingPreview';
 import type { MatchBehaviors } from './uploadPlanParser';
 
 const modifyLine = (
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/WbPlanView/mappingReducer.ts
```diff
@@ -251,12 +255,21 @@ export const reducer = generateReducer<MappingState, MappingActions>({
     )
       return state;
 
+    const newLine = replaceItem(state.lines, focusedLine, {
+      ...state.lines[focusedLine],
+      mappingPath: mappingViewMappingPath,
+    });
+
+    const newHeader = generateMappingPathPreview(
+      baseTableName as keyof Tables,
```
maxpatiiuk, line 264: you should be able to get the base table name from `state`

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingReducer.ts
```diff
@@ -251,12 +255,21 @@ export const reducer = generateReducer<MappingState, MappingActions>({
...
+      newLine[focusedLine].mappingPath
+    );
```
maxpatiiuk, line 266: hmm, doesn't look right
the `headerName` is supposed to be coming from the data set
the only time you would be setting header name to result of `generateMappingPathPreview()` is when you are trying to add a new column to the data set - which doesn't seem to be the case here?

maxpatiiuk (changes requested):

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingReducer.ts
CarolineDenis: we need to reset the header when the user changes the fields in an existing line

CarolineDenis:
before: 

https://github.com/user-attachments/assets/9c56ff54-bc76-4412-85e2-7e42dc0d8d79

now: 

https://github.com/user-attachments/assets/16d8f75c-b578-4b21-8bf5-797af704fb78

specifyweb/frontend/js_src/lib/components/WbPlanView/mappingReducer.ts
CarolineDenis: not sure, how would you? 
export type MappingState = State<
  'MappingState',
  {
    readonly showMappingView: boolean;
    readonly showHiddenFields: boolean;
    readonly mappingView: MappingPath;
    readonly mappingsAreValidated: boolean;
    readonly lines: RA<MappingLine>;
    readonly focusedLine: number;
    readonly changesMade: boolean;
    readonly mustMatchPreferences: IR<boolean>;
    readonly autoMapperSuggestions?: RA<AutoMapperSuggestion>;
    readonly openSelectElement?: SelectElementPosition;
    readonly validationResults: RA<MappingPath>;
  }
>;

maxpatiiuk: > not sure, how would you?

you are right!
sorry, I forgot that it is coming as a prop, and not part of the state

maxpatiiuk:
This is wrong behavior
Headers are defined by the user in the CSV - we do not touch them.
See:

https://github.com/user-attachments/assets/42d3b950-8507-4e0e-aea7-4c7abb010041


The ONLY headers that we touch are the headers for newly added columns.
BUT, we are already renaming those, so there is no need to do that again. See:
https://github.com/specify/specify7/blob/6e3bbb30dc8a3a43a327a7adfa04ecc1d02708b4/specifyweb/frontend/js_src/lib/components/WbPlanView/helpers.ts#L51-L55

PR #5144, opened 2024-07-25
Django command fix_missed_timestamp_modified

acwhite211:
Retroactively fix the missed timestamp-modified fields that occurred between the 7.9.6 and 7.9.6.2 releases.  Do this by querying the Spauditlog table, filtering out workbench logs and logs outside the date range, and then update the instance's timestampmodified if the instance's timestampmodified is None or earlier than the log's timestampmodified.

Will need to run `docker exec -it specify7-specify7-1 ve/bin/python manage.py fix_missed_timestamp_modified`


acwhite211:
There is probably going to be some special cases we need to account for, but the basic idea is easy to implement with the Django ORM.

realVinayak:
No body should look at this code

PR #4958, opened 2024-05-28
Fix permission error in development

melton-jason:
Fixes this console error when webpack is in 'development' mode

Screenshot 2024-05-28 at 10 07 57 AM

From 
https://github.com/specify/specify7/blob/7cf9feaf3e2b3a3ba040e69fc24eadce65d49547/specifyweb/frontend/js_src/lib/components/Permissions/index.ts#L88-L90

The frontend expects the ordering of the permissions for each action to be the same on the backend and the frontend. 
The `bulk_move` permission for '/tree/edit/storage' has a different order: 

Frontend: 
https://github.com/specify/specify7/blob/7cf9feaf3e2b3a3ba040e69fc24eadce65d49547/specifyweb/frontend/js_src/lib/components/Permissions/definitions.ts#L34-L41

Backend: 
https://github.com/specify/specify7/blob/7cf9feaf3e2b3a3ba040e69fc24eadce65d49547/specifyweb/specify/tree_views.py#L380-L387 

Alternatively, we an refactor the function such that the order of the individual actions does not matter. 

### Checklist

- [ ] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [ ] Add automated tests
- [ ] Add relevant issue to release milestone

### Testing instructions
(no UI/UX testing required for this PR)

- [ ] With webpack in development mode, ensure the `'Front-end has outdated list of operation policies'` error is not thrown

PR #5145, opened 2024-07-26
chore: reduce console noise

maxpatiiuk:
All of the following log messages are now removed:

Screenshot 2024-07-25 at 18 32 03

Screenshot 2024-07-25 at 18 35 31

Besides fixing/silencing those errors/warnings, I removed the logging of how long each entry point took to fetch.
I initially thought that it would be helpful for keeping an eye on performance and caching, but my eyes have just "learned to ignore" these messages, defeating the point.
If you think they are helpful, I can leave them be - otherwise, less noise!

### Checklist

### Testing instructions

Verify there is less noise in the console


melton-jason:
Great Idea! There was a PR already to address the permission error,  #4958, but I think that should be closed in favor of this (I'm a fan of including how to resolve the error in the error message in this case. Saves future time of trying to find out how to do so)

CarolineDenis (approved):

sharadsw (approved):

PR #5151, opened 2024-07-27
Implement searching across all trees of a discipline

realVinayak:
Fixes #(something??)

Implements the backend part behind https://github.com/specify/specify7/pull/5036. 

In the current implementation, it queries for a rank (say, Species) across all treedefs in a discipline. So, if you do `Species`, and `Fineral` and `Mossil` both contain `Species`, they both trees will be searched, along with any field to them. If say `Mossil` did not contain the rank Species, then it would not be included in the query result (this is done to avoid the case where ALL nodes from `Mossil` would have been otherwise included). However, it is smart enough to include `Mossil` if any tree rank from `Mossil` is added. @grantfitzsimmons, @CarolineDenis is that what the requirement is? 

EDIT: This does handle geography and other trees just like before

### Checklist

- [ ] Add automated tests

### DIFF
--- a/specifyweb/barvis/views.py
+++ b/specifyweb/barvis/views.py
@@ -31,17 +31,11 @@ def taxon_bar(request):
     # Implementing the previous SQL query in Django ORM:
     taxon_tree_defs = get_taxon_treedefs(request.specify_collection)
     taxons = (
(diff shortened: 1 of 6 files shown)


realVinayak:
Closing this, since changes have been manually added to main pr

PR #5176, opened 2024-08-05
Use new table icons in Workbench column headers

### LINKED ISSUE #2864: Old table icons are used in the WorkBench
image

sharadsw:
Fixes #2864 

### Checklist

- [ ] Add automated tests

### Testing instructions

- Create a workbench dataset
- [ ] Verify new icons are being used in the column headers
- [ ] Verify icons work fine for new geo tables (eg: COG, COGType, age tables etc.)
- [ ] Verify table icons look the same as before everywhere else (Data Entry, Schema Config etc.)
- [ ] Verify icons look fine after changing font size or other text options


specifyweb/frontend/js_src/lib/components/WorkBench/hotProps.tsx
```diff
@@ -62,23 +61,20 @@ export function useHotProps({
...
       const tableName =
         (typeof mappingCol === 'number'
           ? mappings?.tableNames[mappingCol]
-          : undefined) ?? tableIcon?.split('/').slice(-1)?.[0]?.split('.')?.[0];
-      const tableLabel = isMapped
-        ? f.maybe(tableName, getTable)?.label ?? tableName ?? ''
-        : '';
-      // REFACTOR: use new table icons
+          : undefined) ??
+        tableIconUrl?.split('/').slice(-1)?.[0]?.split('.')?.[0];
```
maxpatiiuk, line 71: .slice() never returns undefned - no need for `?.`

```suggestion
        tableIconUrl?.split('/').slice(-1)[0]?.split('.')[0];
```

maxpatiiuk (approved):
looks good!

somewhat related:
since you refactored the workbench, could you double-check if some of these legacy icons can now be removed?
https://github.com/specify/specify7/blob/32d330173e2863db7bafb711ab39dad2660b95c9/specifyweb/frontend/js_src/lib/components/Atoms/Icons.tsx#L28-L31

sharadsw:
Removed the ban legacy icon

specifyweb/frontend/js_src/lib/components/Molecules/SvgIcon.tsx
```diff
@@ -61,6 +61,8 @@ export function SvgIcon({
       <g>
         <text
           dominantBaseline="central"
+          alignmentBaseline="middle"
+          baselineShift="-10%"
```
sharadsw, line 64: The svg text does not vertically align without this for column headers. The `baselineShift` is a magic number but it perfectly aligns the text.  Without it, the vertically alignment is *juuust* so slightly to the top.

Without `alignmentBaseline` and `baselineShift`:
Screenshot 2024-08-06 at 10 33 55 AM

With both:
Screenshot 2024-08-06 at 10 48 25 AM

Without `baselineShift`:
image

maxpatiiuk: I am ok with this fix, but I would like you to double check one thing:
The svg icons work fine everywhere else, but break in this one place.
I suspect thus the issue is not with the SVG icons, but with the interference with the handsontable styling.
could you verify if some style from handsontable is interfering with our icons? (use the styles panel in the dev tools for that)

then, if you find the cause, you can add the fixup CSS into this file:
https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/css/workbench.css

I leave it up to you to pick which solution you prefer more

maxpatiiuk:
The current solution is good, but we need to test it:
- does the text looks ok in other places (i.e workbench mapper and the data entry table list?)
- if you change the font size and other UI settings, do icons still display fine, or not?

CarolineDenis (approved):

pashiav (approved):
### Testing instructions
- Create a workbench dataset
- [X] Verify new icons are being used in the column headers
- [X] Verify icons work fine for new geo tables (eg: COG, COGType, age tables etc.)
- [X] Verify table icons look the same as before everywhere else (Data Entry, Schema Config etc.)
- [X] Verify icons look fine after changing font size or other text options
---
This looks so better now! Everything is working as expected.

Screenshot 2024-11-21 at 9 52 45 AM

alesan99 (approved):
### Testing instructions

- Create a workbench dataset
- [x] Verify new icons are being used in the column headers
- [x] Verify icons work fine for new geo tables (eg: COG, COGType, age tables etc.)
- [x] Verify table icons look the same as before everywhere else (Data Entry, Schema Config etc.)
- [x] Verify icons look fine after changing font size or other text options

chrome_sCQTYlN8ZL
The new icons are being used and look good 👍👍
Everywhere else where icons are used seemed to be functioning as usual

PR #5196, opened 2024-08-06
Unblock save when no current determinations exist

### LINKED ISSUE #5189: Cannot create a new CO without a determination after a COT is set
**Steps to recreate:**
1. Select a Collection Object Type
2. Try to save the record
3. See error: `Determination - Current: A current determination is required.`

image

**System Information:**
macOS, Arc
Git SHA: https://github.com/specify/specify7/commit/9f37028b2e20f10eb9972d50f4058790b3dcca9f

[Specify 7 System Information - 2024-08-05T17_13_03.215Z.txt](https://github.com/user-attachments/files/16500712/Specify.7.System.Information.-.2024-08-05T17_13_03.215Z.txt)

sharadsw:
Fixes #5189 

### Checklist

- [x] Add automated tests

### Testing instructions

- Create a CollectionObject / Open an existing one
- Add determinations
- Remove all determinations manually or by changing CollectionObjectType
- [ ] Verify save does not get blocked due to having no current determination

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
@@ -90,6 +90,7 @@ describe('Collection Object business rules', () => {
         {
           taxon: getResourceApiUrl('Taxon', otherTaxonId),
           preferredTaxon: getResourceApiUrl('Taxon', otherTaxonId),
+          isCurrent: true,
         } as SerializedResource<Determination>,
       ],
       resource_uri: collectionObjectUrl,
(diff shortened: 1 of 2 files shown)


CarolineDenis (approved):

Areyes42 (approved):
### Testing instructions

-   Create a CollectionObject / Open an existing one
-   Add determinations
-   Remove all determinations manually or by changing CollectionObjectType
-  [x]  Verify save does not get blocked due to having no current determination
***
Save blocker no longer appears for COs without current determinations or with removed determinations 🤙

https://github.com/user-attachments/assets/a76ce5e5-0a2c-4a0c-ad43-b461e458f39c

specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
```diff
@@ -168,6 +169,64 @@ describe('Collection Object business rules', () => {
...
+    const collectionObject = getBaseCollectionObject();
+    const determinations =
+      collectionObject.getDependentResource('determinations');
+
+    determinations?.add(new tables.Determination.Resource());
+
+    // Old determination gets unchecked as current
+    expect(determinations?.models[0].get('isCurrent')).toBe(false);
+    // New determination becomes current by default
+    expect(determinations?.models[1].get('isCurrent')).toBe(true);
```
melton-jason, line 183: (just a note) 

How reliable is it that the first Determination in the collection `at models[0]` is the one defined on the `baseCollectionObject`? 

I know a comparator for a Collection can be explicitly defined, and BackBone implicitly keeps the Collection sorted when objects are added (see https://github.com/jashkenas/backbone/blob/4e642088c4810938a0510fa89456ef2619f446da/backbone.js#L1032-L1051), but I am not sure of the behavior when a comparator is not defined for a Collection (I assume it sorts by their internal `cid`, or when the resources/models are added to the Collection?) 

If there is a consistent behavior, we can keep this as is! 
Otherwise, we can probably include some identifier on each of the Determinations and filter them from the Collection when needed (this would also handle the case when the BaseCollectionObject's determinations are modified) .

melton-jason (approved):
Love that tests are being included here! 🥳

specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
sharadsw: Looks like it gets sorted by its `cid`

PR #5200, opened 2024-08-07
Add COGType picklist to COG and add COGTypeType picklist

### LINKED ISSUE #5118: Add support for searching COG Type on COG form
**Describe the bug**
image

No results are returned, despite there being a record:

image

Currently, COG types cannot be searched using a QCBX. This needs to be remedied before it is released widely.

## How it should work

image

To make this happen in the current release, I did the following:

1. Added to the `TypeSearches` app resource:
```xml
    <typesearch tableid="1018"   name="CollectionObjectGroupType"              searchfield="name" displaycols="name"    format="%s"     dataobjformatter=""/>
```

2. Added the following table format:
```xml
	<format name="CollectionRelType" title="CollectionRelType" class="edu.ku.brc.specify.datamodel.CollectionRelType" default="true">
		<switch single="true">
			<fields>
				<field>name</field>
			</fields>
		</switch>
	</format>
```

CarolineDenis:
Fixes #5118 

### Checklist

### Testing instructions

TODO

### DIFF
--- a/specifyweb/specify/migrations/0004_cogtype_picklist.py
+++ b/specifyweb/specify/migrations/0004_cogtype_picklist.py
@@ -70,14 +70,14 @@ class Migration(migrations.Migration):
     ]
 
     def apply_migration(apps, schema_editor):
-        create_cotype_picklist()
-        create_cotype_splocalecontaineritem()
-        create_cotype_splocaleitemstr()
+        create_cogtype_picklist()
+        create_cogtype_splocalecontaineritem()
+        create_cogtype_splocaleitemstr()
 
     def revert_migration(apps, schema_editor):
-        revert_cotype_picklist()
-        revert_cotype_splocaleitemstr()
-        revert_cotype_splocalecontaineritem()
+        revert_cogtype_picklist()
+        revert_cogtype_splocaleitemstr()
+        revert_cogtype_splocalecontaineritem()
 
     operations = [
         migrations.RunPython(apply_migration, revert_migration, atomic=True)
--- a/specifyweb/specify/migrations/0004_cogtype_picklist.py
+++ b/specifyweb/specify/migrations/0004_cogtype_picklist.py
@@ -1,0 +1,84 @@
+from django.db import migrations
+from specifyweb.specify.models import (
+    Collection,
+    Splocalecontainer,
+    Splocalecontaineritem,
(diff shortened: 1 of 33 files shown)


CarolineDenis:
Reminder: 
- The COG has a COGType pick list (defined by user) 
- The COGType has a Type pick list (read-only, one of those three values: Discrete - Consolidated - Drill Core)

specifyweb/specify/migrations/0004_cogtype_picklist.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Create a Splocalecontaineritem record for each CollectionObjectGroup Splocalecontainer
+    # NOTE: Each discipline has its own CollectionObjectGroup Splocalecontainer
+    for container in Splocalecontainer.objects.filter(name='collectionobjectgroup', schematype=0):
+        Splocalecontaineritem.objects.get_or_create(
+            name=FIELD_NAME,
+            picklistname=PICKLIST_NAME,
+            type='ManyToOne',
+            container=container,
+            isRequired=False
```
CarolineDenis, line 42: do we want this true or false?

sharadsw (changes requested):
A note on picklist formatter:
Like in the COType PR, we will need a default formatter for the picklist. Looks like #5118 requires the formatter to be `cogtype -> type` but I'm not sure that would enough since each COGtype has an associated type (Drill Core, Discrete, Consolidated) and so the picklist would just have multiple entries for the same types. Maybe a better formatter would be `name + type` but either way we would need to create a default formatter (which involves changing a xml file in Sp6)

sharadsw (approved):

sharadsw:
NOTE:
Pushed commits to fix the default picklist created in the geo migration and also the fix the update queries for cog -> cogtype and cogtype -> type. 
- Cog -> cogtype will be a type 1 picklist containing names of all cogtypes in the db
- Cogtype -> type will be a type 0 picklist containing values Drill Core, Discrete and Consolidated

Any default formatter we define for the Cog -> cogtype picklist in future will have to be named `CollectionObjectGroupType` for it to be used automatically for this picklist.

CarolineDenis:
Added to https://github.com/specify/specify7/pull/5257/files

PR #5214, opened 2024-08-12
Add support for sorting columns in prep dialog

### LINKED ISSUE #5142: Allow sorting items in the create loan dialog
image

It would be useful to sort by Prep Type so the correct items can be chosen

**Requested by:** RBGE

grantfitzsimmons:
Fixes #5142 

This makes it so that the columns in the preparation select dialog are now sortable. This includes only **Catalog Number**, **Taxon**, **Preparation**, **Available**, and **Unavailable**. 'Selected' is not sortable at this time.

**AI Disclaimer:** This PR was drafted with help from GPT-4o mini. This is largely an experiment to see how capable it is.

I'm going to work on this in my free time. If we see this being an urgent priority, please have a staff developer begin work on this!

### Checklist

- [ ] Add automated tests

### Testing instructions

Please test this side-by-side with the latest `v7` tagged release. If an issue is encountered, please verify that it also does not happen in `edge` before requesting changes.

- [ ] Create a series of new interaction records by clicking the **Interactions** menu item
   - [ ] Create a new **Loan**, **Gift**, **Exchange In**, **Exchange Out**, and **Disposal** record based on a record set and based on a series of catalog numbers. 
   - [ ] **Repeat the process for each of these interactions:**
	   - [ ] In the new interaction dialog, select several rows at random.
	   - [ ] Enter a specific quantity in the 'selected' column and write it down alongside the catalog number, taxon name, preparation type, and quantities.
	   - [ ] Sort the rows by the headers (Catalog Number, Taxon, Preparation, Available, and Unavailable)
	   - [ ] Select more items and do the same step, capturing the specific preparation information
	   - [ ] Upon sorting, make sure the appropriate 'selected' values remain with the appropriate preparation
	   - [ ] Create the new interaction record
	   - [ ] Verify that under the '{Interaction} Preparations' subview that the correct 'selected' quantities appear with the appropriate loan preparation.
	   - [ ] Navigate to the CO records directly and verify that the preparations are linked to the correct interaction.
	   - [ ] **For Loans:**  Test that the 'Return Loan' function works as expected. _Note that the columns are not sortable in this dialog._
   - [ ] Make sure that 'Bulk Select' works as expected in all contexts

Please test this as extensively as possible. Let me know if you have any questions.


specifyweb/frontend/js_src/lib/components/Interactions/PrepDialog.tsx
```diff
@@ -63,14 +64,20 @@ export function PrepDialog({
...
-    React.useCallback(
-      () => Array.from({ length: preparations.length }).fill(0),
-      [preparations.length]
-    )
+  // Change to use an object for selected state allowing null values
+  const [selected, setSelected] = useLiveState<{
+    [key: string]: number | null;
+  }>(
+    React.useCallback(() => {
+      return preparations.reduce((acc, preparation) => {
```
maxpatiiuk, line 72: a bit more type safe:

```suggestion
      return preparations.reduce<{ [key: string]: number | null }>((acc, preparation) => {
        acc[preparation.preparationId] = null;
        return acc;
      }, {});
```

and we have a utility type for this as it's used very often:

```suggestion
      return preparations.reduce<R<number | null>>((acc, preparation) => {
        acc[preparation.preparationId] = null;
        return acc;
      }, {});
```

specifyweb/frontend/js_src/lib/components/Interactions/PrepDialog.tsx
```diff
@@ -63,14 +64,20 @@ export function PrepDialog({
...
   }, [rawPreparations, itemCollection]);
 
-  const [selected, setSelected] = useLiveState<RA<number>>(
-    React.useCallback(
-      () => Array.from({ length: preparations.length }).fill(0),
-      [preparations.length]
-    )
+  // Change to use an object for selected state allowing null values
+  const [selected, setSelected] = useLiveState<{
+    [key: string]: number | null;
```
maxpatiiuk, line 69: (optional) instead of null, could you refactor the code you added in this file to use `undefined`? that would be more consistent with what we use in the rest of the codebase

specifyweb/frontend/js_src/lib/components/Interactions/PrepDialog.tsx
```diff
@@ -133,12 +180,28 @@ export function PrepDialog({
...
+              setBulkValue(0);
+              setSelected((prev) => {
+                const newSelected = { ...prev };
+                sortedPreparations.forEach((prep) => {
+                  newSelected[prep.preparationId] = null;
+                });
+                return newSelected;
+              });
+            } else {
+              const count = parseInt(newCount, 10);
```
maxpatiiuk, line 195: hmm, the `Input.Integer` is supposed to call `onValueChange` with `number` not `string` as far as I remember
this this parseInt here and the `if (newCount === '') {` check above is not right 🤔

maxpatiiuk (changes requested):
👍

PR #5224, opened 2024-08-14
Set default cotype on the frontend

### LINKED ISSUE #5223: Set a default CollectionObjectType on the frontend
The default CollectionObjectType for a CollectionObject will be the one set in the Collection record. The backend already sets this as default but it would be useful to set it on the frontend as well so the user can see a default being set.

sharadsw:
Fixes #5223

### Checklist

- [ ] Add automated tests

### Testing instructions

- Go to Data Entry -> Collection Object
- [ ] Verify the default CoType on the form is same as the Collection's CoType


CarolineDenis (approved):

sharadsw:
Hmm, this breaks a couple of resource api tests which doesn't sound good
https://github.com/specify/specify7/blob/1419cd784acdb1bc2f5189ddb3ea26e77d7d619e/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/resourceApi.test.ts#L201-L219

specifyweb/frontend/js_src/lib/components/DataModel/__tests__/resourceApi.test.ts
```diff
@@ -199,20 +199,30 @@ describe('rgetCollection', () => {
...
 describe('needsSaved', () => {
-  test('changing field makes needsSaved true', () => {
+  test('changing field makes needsSaved true', async () => {
     const resource = new tables.CollectionObject.Resource({
       id: collectionObjectId,
     });
+
+    // Need to wait until the business rule fetches Collection -> CollectionObjectType (https://github.com/specify/specify7/pull/5224)
+    // See: https://stackoverflow.com/a/51045733
+    await new Promise(process.nextTick);
```
sharadsw, line 209: The test was failing as `resource.needsSaved` doesn't update until the `rgetPromise` in this business rule resolves.

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -156,6 +157,18 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
...
+      // Set the default CoType
+      if (collectionObject.get('collectionObjectType') === undefined)
+        collectionObject
+          .rgetPromise('collection')
+          .then((collection: SpecifyResource<CollectionTable> | null) => {
+            if (collection !== null)
+              collectionObject.set(
+                'collectionObjectType',
+                collection.get('collectionObjectType')
+              );
```
sharadsw, line 170: `CollectionObject -> Collection` shouldn't be null according to DataModel/types.ts but there are lots of tests where CO gets initialized as an empty resource and so it doesn't have a linked Collection/Cotype and so I changed the type here to allow null

sharadsw (commented):
Had to do some workarounds here to get the tests working. Lots of tests involving CO don't have a linked CollectionObjectType or a linked `Collection -> CollectionObjectType`. Some tests used data from static test files, which I have updated to include Collection/Cotype.

Let me know if there's a better way to do any of these.

emenslin (commented):
### Testing instructions



- Go to Data Entry -> Collection Object
- [x] Verify the default CoType on the form is same as the Collection's CoType

This does work but I am a little confused about the query behavior. When querying with cot as the base table it works as I expect showing the 2 cot I made in this collection, but when querying with collection as the base table cot is formatted and only shows the default. Since we added the ability to create multiple cot shouldn't it be aggregated when collection is the base table? 

Querying on collection: 
image

Querying on cot: 
image

pashiav (approved):
### Testing instructions



- Go to Data Entry -> Collection Object
- [X] Verify the default CoType on the form is same as the Collection's CoType

I have the same concerns as https://github.com/specify/specify7/pull/5224#pullrequestreview-2292904057, but it works!

sharadsw:
@emenslin That does look strange. I think it's happening because each collection record in the db gets linked to only one collectionObjectType (the default one) and so using collection as the base table only shows the defaults. Not sure if that can be changed unless we change the schema

grantfitzsimmons:
> @emenslin That does look strange. I think it's happening because each collection record in the db gets linked to only one collectionObjectType (the default one) and so using collection as the base table only shows the defaults. Not sure if that can be changed unless we change the schema

The issue is that the schema caption for the default COT relationship field should clarify this. Currently, it appears ambiguous, suggesting there is only one COT in the collection.

melton-jason:
> > @emenslin That does look strange. I think it's happening because each collection record in the db gets linked to only one collectionObjectType (the default one) and so using collection as the base table only shows the defaults. Not sure if that can be changed unless we change the schema
> 
> The issue is that the schema caption for the default COT relationship field should clarify this. Currently, it appears ambiguous, suggesting there is only one COT in the collection.

Yep, if I remember correctly the `Collection -> CollectionObjectType` relationship is only to denote the default collection object type for the Collection. 
Multiple CollectionObjectTypes can exist in a single Collection, but there can only be one default.

sharadsw:
I agree that the schema caption could be clearer. We can make that change as a part of #5290

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
melton-jason: If this is something that is going to be fetched a lot, maybe it would be worth including the Collection -> CollectionObjectType in some context call? 

Specifically, maybe include the CollectionObjectType in the `/context/domain.json` endpoint? 

https://github.com/specify/specify7/blob/5c639bf75575e46df2b5a574269abe2e4a24fe81/specifyweb/context/views.py#L331-L345

https://github.com/specify/specify7/blob/5c639bf75575e46df2b5a574269abe2e4a24fe81/specifyweb/frontend/js_src/lib/components/DataModel/schema.ts#L83-L94

That way, if we can reasonably assume the schema endpoint has been fetched by the time the businessrule is being run, we can simply do something like: 

```ts
collectionObject.set(
  'collectionObjectType',
   schema.defaultCollectionObjectType
);

// maybe fallback to the promise if schema is undefined? 
```

melton-jason (approved):
This implementation works, but maybe it would be worth changing businessrules a little (in this case specifcially for `customInit` rules) so we can `await businessRuleManager.pendingPromise` instead of `await new Promise(process.nextTick)` in tests. 

Also, maybe it'd be worth just including the Collection -> CollectionObjectType in a domain context call so we can synchronously access the collection's default CollectionObjectType wherever the schema has already been fetched. 
This would save a _lot_ of calls to the backend, and be significantly faster from the user's perspective.

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -156,6 +157,16 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
           new tables.CollectingEvent.Resource()
         );
       }
+
+      // Set the default CoType
+      if (typeof schema.defaultCollectionObjectType === 'number')
```
melton-jason, line 162: You still need the `collectionObject.get('collectionObjectType') === undefined` check here (and you can include `null` as well if that is part of the requirements): otherwise we overwrite the collectionObjectType on the collectionObject if the type is different from the default collectionObjectType.

melton-jason (changes requested):

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
sharadsw: Is it not safe to assume a CO will always be initialized with an undefined COType?

melton-jason: Yes, for new COs it's safe to assume that the COType will be undefined if it is not set. 

`customInit` as a rule is called regardless of whether the SpecifyResource is new or not: specifically, it's called whenever businessrules are attached to a resource, which happens when BackBone initializes a resource. 

https://github.com/specify/specify7/blob/92c94292c6d70fb867994d88ca503ea56572ea8f/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts#L54-L63

https://github.com/specify/specify7/blob/92c94292c6d70fb867994d88ca503ea56572ea8f/specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts#L462-L468

https://github.com/specify/specify7/blob/92c94292c6d70fb867994d88ca503ea56572ea8f/specifyweb/frontend/js_src/lib/components/DataModel/resourceApi.ts#L128

So `customInit` is also called on CollectionObjects which are not new, and they can have a non-undefined collectionObjectType

melton-jason (commented):
Looks good! Take a look at my recent changes. 
Instead of just returning the ID of the CollectionObjectType in the `/context/domain.json` endpoint, I thought it might be easier if we just directly format it as the the api uri. 

I updated the static `/context/domain.json` endpoint for the frontend tests to include the `defaultCollectionObjectType` property. 
I also realized we weren't using a LOT of the static files in the `*/tests/fixtures/` directory, so I removed all the files which are unused.

PR #5248, opened 2024-08-23
Filter Taxon QCBX by CollectionObjectType tree

### LINKED ISSUE #4982: Add business logic for MOTs
Component of https://github.com/specify/specify7/issues/4640

Review the full business logic described in the main topic and let me know if there are any questions.

> ### Business Logic & Requirements
> 
> - Each `collectionobjecttype` would be scoped to a particular collection
> 
> - An `objectType` would be chosen from a pick list on the Collection Object form 
> 
> - When creating a new Determination record, the system should check the `objectType.taxonTreeDef` field to find the tree to search for the user's entered name when searching in the `taxon` or `preferredTaxon` query combo boxes
> 
> - `objectType` should not be a required field but can be made to be using the form editor.
> 
> 
> - Conditional forms must be able to use the value from the `objectType` field to determine the layout in an intuitive way for users
>    - If the user can define a form definition as follows:
>    ```xml
>      <cell type="panel" coldef="100px,2px,76px" colspan="6">
>        <rows>
>          
>          <row>
>            <cell type="label" label="Without any object type, this will display." />
>          </row>
>        </rows>
>        <rows condition="objectType=Rock">
>          <row>
>            <cell type="label" label="This is a rock!" />
>          </row>
>        </rows>
>        <rows condition="objectType=Mineral">
>          <row>
>            <cell type="label" label="This is a mineral!" />
>          </row>
>        </rows>
>      </cell>
>    ```
>    That would be satisfactory. This would enable users to create forms or display subviews based on the value, which would also enable users to create specific determination forms or collection object attribute forms based on the value of `objectType`.

### LINKED ISSUE #5213: Determination `taxon` query combo box should be filtering by COT tree
Originally a component of https://github.com/specify/specify7/issues/4640

>- When the user creates a new determination, the taxon tree searched when entering a `taxon` or `preferredTaxon` value must be based on the object type's tree (if `objectType` is present), otherwise it should use the discipline default

I've discussed this with Caroline and it will be done sometime before the MVP if possible. I have found this to be a really important user interface issue.

**Issue:** The Taxon QCBX on the Determination form does not filter based on the selected Collection Object Type (COT)'s associated taxon tree.

**Solution:** Filter the QCBX based on the associated Collection Object’s (CO) COT’s tree.

**Rationale:** When searching for a taxon name, the results currently show names from all taxon trees. This is cumbersome as it

1. Slows the application for the user as it loads dozens of names that are not relevant or desired based on the type
2. Enables users to select a name that prevents the record from being saved.
3. The Collection Object Type can be an obstacle when its sole purpose is to validate the tree node

This should not be defined in any special way via the XML– it should only use the database relationships. If no type is selected, it should follow the default type for the collection.

sharadsw:
Fixes #5213
Fixes #4982

The cotype -> treedef value can be retrieved from the Determination resource used by the Taxon QCBX, which will be used to filter Taxon values

### Checklist

- [ ] Add automated tests

### Testing instructions

- Use a db with multiple trees in the same collection
- Go to Data Entry > Collection Object
- Enter a CollectionObjectType
- Add a determination
- [ ] Verify Taxon field of the determination gets filtered by the selected COType's tree definition

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
@@ -24,7 +24,10 @@ import { serializeResource } from '../DataModel/serializers';
 import type { Relationship } from '../DataModel/specifyField';
 import type { SpecifyTable } from '../DataModel/specifyTable';
 import { tables } from '../DataModel/tables';
-import { CollectionObject, CollectionObjectType } from '../DataModel/types';
+import type {
+  CollectionObject,
+  CollectionObjectType,
+} from '../DataModel/types';
 import { format, naiveFormatter } from '../Formatters/formatters';
 import type { FormType } from '../FormParse';
 import { ResourceView, RESTRICT_ADDING } from '../Forms/ResourceView';
--- a/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
@@ -23,6 +23,8 @@ import {
 import { serializeResource } from '../DataModel/serializers';
 import type { Relationship } from '../DataModel/specifyField';
 import type { SpecifyTable } from '../DataModel/specifyTable';
+import { tables } from '../DataModel/tables';
+import { CollectionObject, CollectionObjectType } from '../DataModel/types';
 import { format, naiveFormatter } from '../Formatters/formatters';
 import type { FormType } from '../FormParse';
 import { ResourceView, RESTRICT_ADDING } from '../Forms/ResourceView';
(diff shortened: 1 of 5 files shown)


specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
```diff
@@ -247,6 +252,26 @@ export function QueryComboBox({
...
+              )
+          : undefined,
+      [resource, resource?.collection?.related?.get('collectionObjectType')]
+    ),
+    false
+  );
+
   // FEATURE: use main table field if type search is not defined
   const fetchSource = React.useCallback(
     async (value: string): Promise<RA<AutoCompleteItem<string>>> =>
```
maxpatiiuk, line 277: do you need to add a check in this function for `treeDefinition` being loaded?
i.e don't start fetching resource until treeDefinition is resolved

that also means you need to turn treeDefinition into a tri-state value (some value to indicate that you are still fetching, some value to indicate there is nothing to fetch, and the actual result object)

maxpatiiuk (commented):
👍

emenslin (approved):
### Testing instructions



- Use a db with multiple trees in the same collection
- Go to Data Entry > Collection Object
- Enter a CollectionObjectType
- Add a determination
- [x] Verify Taxon field of the determination gets filtered by the selected COType's tree definition

sharadsw:
TODO: Remove determination save blocker that was added as a workaround previously

pashiav (approved):
### Testing instructions



- Use a db with multiple trees in the same collection
- Go to Data Entry > Collection Object
- Enter a CollectionObjectType
- Add a determination
- [X] Verify Taxon field of the determination gets filtered by the selected COType's tree definition

CarolineDenis (approved):

PR #5251, opened 2024-08-27
QB Api change for handling MOTs

### LINKED ISSUE #4979: Add support for multiple trees in the query builder

### LINKED ISSUE #5392: Change API to support for multiple trees in the query builder
This issue is an extension of #4979.  This PR defines the new protocol in the QB API for specifying the treedefid of a rank. The new QB request schema for MOTs is <table_id>.<table_name>.<treedef_id>,<rank>,<field>. The code is backwards compatible, so it still allows that old schema to be used. The front-end needs to be changed to send this new QB request schema when dealing with MOTs.

acwhite211:
Fixes #4979
Fixes #5392

This should have been included in #5036 PR.  This PR defines the new schema in the QB for specifying the treedefid of a rank. The new QB request schema for MOTs is `<table_id>.<table_name>.<treedef_id>,<rank>,<field>`.  The code is backwards compatible, so it still allows that old schema to be used.  The front-end needs to be changed to send this new QB request schema when dealing with MOTs.

### Checklist

### Testing instructions

TBD

### DIFF
--- a/specifyweb/specify/tree_utils.py
+++ b/specifyweb/specify/tree_utils.py
@@ -20,13 +20,13 @@ def get_search_filters(collection: spmodels.Collection, tree: str):
         discipline_query |= Q(id=tree_at_discipline.id)
     return discipline_query
 
-def get_treedefs(collection: spmodels.Collection, tree_name: str) ->  List[Tuple[int, int]]:
+def get_treedefs(collection: spmodels.Collection, tree_name: str, treedef_id=None) ->  List[Tuple[int, int]]:
     # Get the appropriate TreeDef based on the Collection and tree_name
 
     # Mimic the old behavior of limiting the query to the first item for trees other than taxon.
     # Even though the queryconstruct can handle trees with multiple types.
     _limit = lambda query: (query if tree_name.lower() == 'taxon' else query[:1])
-    search_filters = get_search_filters(collection, tree_name)
+    search_filters = get_search_filters(collection, tree_name) if treedef_id is None else Q(id=treedef_id)
 
     lookup_tree = lookup(tree_name)
     tree_table = datamodel.get_table_strict(lookup_tree)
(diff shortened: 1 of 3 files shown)


acwhite211:
Still need to add front-end code to use the new API request schema when using taxon ranks.

grantfitzsimmons:
Include in 7.9.7 if front-end can be finished this week, push to 7.9.8 if not

CarolineDenis:
@acwhite211 
Why was this not being tested and merged? 
Could you create an issue in 9.9 for it please?

acwhite211:
This solution does seem to be needed anymore for QB requests when dealing with MOTs.  If we ever want to change to this solution, we'll also need implement front-end changes to deal with the new request schema.

PR #5266, opened 2024-09-10
Add Specify 6.8.03 Static Files to Specify 7

### LINKED ISSUE #4828: Add Specify 7 static files
Specify 7 falls back on Specify 6's static files quite often for establishing defaults throughout the application. This creates a dependence on Specify 6 updates to change basic defaults for discipline-specific form definitions, the list of interaction tables the user can choose from (https://github.com/specify/specify6/issues/1240), and many other important places.

Simple issues like https://github.com/specify/specify6/issues/1230 cannot be resolved without a Specify 6 update which requires major work for both our development team and our users.

When we add new disciplines or want to set new defaults, we need to have the ability to serve static files via Specify 7 rather than relying entirely on Specify 6. If we want to update the default list of interactions or default forms, we should be able to do so easily by having Specify look at that definition first.

### LINKED ISSUE #5258: Add new geo view def + missing default
A) The TreeDefItem form definitions are missing uiType by default.

     1. Go to App Resources -> New or existing Form Definition
     2. Click show all tables and select any TreeDefItem
     3. Click Create and select [treename]TreeDefItem (i.e. TaxonTreeDefItem) from under Copy default view definition
     4. See "Required attribute uiType is missing" error for multiple fields
 
B) Create custom form definition for CoJo table to fix child relationship fields

The CoJo form view is not displaying the childcog and childco relationship fields because they are not required fields. The      situation is that one of these child fields must to filled out, but not both. Given this situation, we cannot have both of the fields be required. We need to create a custom default form definition for the CoJo table so that these two fields appear.

C) Create viewset definitions for the new geo tables

D) The name field in a CollectionObjectType form defaults to the picklist created in https://github.com/specify/specify7/pull/5170 instead of a text field

acwhite211:
Fixes #4828 
Fixes #5258 

Copy the config directory from Specify 6 into Specify 7 so that the Specify 6 docker container doesn't need to be spun up with the Specify 7 container.  Also, edit the `THICK_CLIENT_LOCATION` in the specify settings so that the `SPECIFY_CONFIG_DIR` gets set to the new path.  The nginx configuration file also needs to be edited to the new config directory path.

This is simply moving the default static files from Specify 6 to Specify 7. This will not override custom client specific configs that have been added to the database, since Specify 7 first looks in the database and then in the filesystem:
https://github.com/specify/specify7/blob/f58f0d43c4441ee1718ab92823f9a1da3d98bed2/specifyweb/context/app_resource.py#L53-L71

Note for this Draft PR, not all the files in the Specify 6 config directory are needed.  So identifying the files that are not actually ever used by Specify 7 and then pruning those from the Specify 7 config directory still needs to be done as of the creation of this Draft PR.

If we only had to worry about hosted instances, it would be nice just to put these files in S3.  Another option to explore, if anyone wants to, would be creating a Django migration to add these default configs as blobs into the database and adding a default field.

### Checklist

- [ ] Add automated tests

### Testing instructions

TBD


maxpatiiuk:
Once you identified and removed the needless static files, make sure to merge this PR with squash so that unused sp6 static files they don't clutter up the specify 7 repository history

Screenshot 2024-09-10 at 16 51 13

acwhite211:
We'll need to edit the Specify 7 nginx web-server to not depend on the Specify 6 container as well.

maxpatiiuk (changes requested):
there still appears to be a lot of files in this PR that are not needed by specify 7.

could you search the back-end and front-end for usages, and delete the needless files?
or would you prefer Jason or I look over these?

acwhite211:
> there still appears to be a lot of files in this PR that are not needed by specify 7.
> 
> could you search the back-end and front-end for usages, and delete the needless files? or would you prefer Jason or I look over these?

Here are the notes I made from going through the code-base, searching file and directory names.

Keep:
```
accessions
backstop
bird
botany
common
defaultperms
fish
herpetology
invertebrate
invertpaleo
mammal
paleobotany
vertpaleo
```

Maybe Keep:
```
insect
reptile
vascplant
fungi
disciplines.xml
icons.xml
icons_datamodel.xml
icons_disciplines.xml
icons_imgproc.xml
icons_plugins.xml
querybuilder.xml
schema_localization.xml
schema_version.xml
```

Probably Remove:
```
ireportconfig.xml
lichens_label.jrxml
specify_workbench_upload_def.xml
```

Remove:
```
SymbiotaDwcMeta.xml
VertNetCacheMapping.xml
agent_update_wb_datamodel.xml
bubble_info.xml
build_ipad_xreftables.sql
collectingevent_update_wb_datamodel.xml
collectionobject_import_wb_datamodel.xml
collectionobject_update_wb_datamodel.xml
command_registry.xml
darwin_core_map.xml
datamodel_automappings.xml
dwcdefaultmap.xml
es_skipfields.xml
floats_to_decimals.sql
geonames.sql.zip
install_info.xml
mime_icons.xml
new_attch_tables.sql
phonet.en
plugin_registry.xml
prefs_init.xml
locality_update_wb_datamodel.xml
preparation_update_wb_datamodel.xml
rstools_registry.xml
show_fields.xml
specify_tableid_listing.xml
specify_workbench_datamodel.xml
specify_workbench_remappings.xml
synonym_cleanup_en.html
tdwg_dwcterms.xsd
text_factory_mapping.xml
thumbnail_generators.xml
wb_registry.xml
wbschema_localization.xml
ipad_exporter
nameresources
dbdrivers.xml
hiddentables.xml
usertypes.xml
```

maxpatiiuk:
Ah, what I mean is that inside each of these folders there are several files that are probably not needed anymore:

```yaml
accessions
backstop
bird
botany
common
defaultperms
fish
herpetology
invertebrate
invertpaleo
mammal
paleobotany
vertpaleo
```

dataentry_task.xml, search_config.xml, startup_panel.xml and etc

there also appear to be many reports/labels files - you will need to check with Grant on these but I was under impression that all reports are hand-crafted or done in jasper studio/irepo, so I am not sure what these files are about

acwhite211:
Note, it looks like the icon xml files are used:

image

maxpatiiuk:
@acwhite211 yes, icons are still used, but only for compatibility with sp6 forms.
This PR might be an excellent time to remove that feature.

Sp6 static files include a fixed set of low resolution icons that people can include in their forms.
The icon*.xml files files are mapping the icon names to icon file names for different icon sizes.

Instead of using these legacy icon names, Specify 7 already supports specifying icon image URLs. Those URLs could point to your local asset server, or any external file server.

Thus the workflow could be:
- less than a dozen of those sp6 icons are included in the default forms - specify 7 could have hardcoded logic to resolve those 10 icons to some file on files.specifysoftware.org
- going forward, we advice people to put icon URLs instead, and instruct them how to get a URL from an asset server image
- when we have the visual form editor, there would be a UI option for uploading a new asset or picking an existing asset to get it's URL

Note, this change means that icons that icons display in sp6 won't be displayed in sp7 without some form changes. If that is too much of a breaking change at this point, then we can delay the deprecation till later.
That hover means this PR would need to also include Specify 6 static icons - without those, we are not separated from sp6 static files.

config/accessions/startup_panel.xml
maxpatiiuk, line 1: not used

config/accessions/statistics_panel.xml
maxpatiiuk, line 1: not used

config/accessions/stats_summary_panel.xml
maxpatiiuk, line 1: not used

config/accessions/search_config.xml
maxpatiiuk, line 1: not used

grantfitzsimmons (changes requested):
Without any modification, this branch is not working for me. 

Initially, when running `docker compose up --build`, it fails because the `specify6` service was defined yet invalid as the `specify6` volume was not being created.

```yml
  specify6:
    image: specifyconsortium/specify6-service:6.8.03
    volumes:
      - "specify6:/volumes/Specify"
```

After commenting the above bit out, I try to run it again. Immediately, I see this error:
```
specify7-1         | FileNotFoundError: [Errno 2] No such file or directory: '/opt/specify7/specify.jar'
```

I check to see that this is actually missing:

```
docker-compose run --rm specify7 ls -l /opt/specify7/config
```

and see
```bash
~/GitHub/specify7 on issue-4828 ⇣1 *1 +1 !2 ❯ docker-compose run --rm specify7 ls -l /opt/specify7/config                                                                                            at 01:37:27 PM
WARN[0000] /Users/g584f396/GitHub/specify7/docker-compose.yml: `version` is obsolete
[+] Creating 1/0
 ✔ Container specify7-webpack-1  Running                                                                                                                                                                       0.0s
total 11900
lrwxrwxrwx  1 root root       49 Aug 17 02:57 DataExporter_64.desktop -> .install4j/install4j_15dyqoe-DataExporter.desktop
lrwxrwxrwx  1 root root       55 Aug 17 02:57 ImportFileSplitter_64.desktop -> .install4j/install4j_15dyqoe-ImportFileSplitter.desktop
lrwxrwxrwx  1 root root       52 Aug 17 02:57 SpBackupRestore_64.desktop -> .install4j/install4j_15dyqoe-SpBackupRestore.desktop
lrwxrwxrwx  1 root root       45 Aug 17 02:57 SpWizard_64.desktop -> .install4j/install4j_15dyqoe-SpWizard.desktop
lrwxrwxrwx  1 root root       47 Aug 17 02:57 Specify4GB_64.desktop -> .install4j/install4j_15dyqoe-Specify4GB.desktop
lrwxrwxrwx  1 root root       60 Aug 17 02:57 SpecifyDBSecurityWizard_64.desktop -> .install4j/install4j_15dyqoe-SpecifyDBSecurityWizard.desktop
lrwxrwxrwx  1 root root       44 Aug 17 02:57 Specify_64.desktop -> .install4j/install4j_15dyqoe-Specify.desktop
lrwxrwxrwx  1 root root       46 Aug 17 02:57 SpiReport_64.desktop -> .install4j/install4j_15dyqoe-SpiReport.desktop
drwxr-xr-x  2 root root     4096 Aug 17 02:57 bin
drwxr-xr-x 21 root root     4096 Aug 17 02:57 config
drwxr-xr-x  2 root root     4096 Aug 17 02:57 demo_files
drwxr-xr-x  7 root root     4096 Aug 17 02:57 help
drwxr-xr-x  2 root root     4096 Aug 17 02:57 hibernate_libs
drwxr-xr-x  2 root root     4096 Aug 17 02:57 iReportLibs
drwxr-xr-x  2 root root     4096 Aug 17 02:57 libs
-rw-r--r--  1 root root    15590 Mar  1  2023 license_agreement.html
-rw-r--r--  1 root root     4996 Mar  1  2023 non-mac-application-adapter.jar
drwxr-xr-x  2 root root     4096 Mar  1  2023 plugins
-rw-r--r--  1 root root      446 Mar  1  2023 readme.html
-rw-r--r--  1 root root 12097156 Mar  1  2023 specify.jar
-rwx------  1 root root    14045 Mar  1  2023 uninstall
drwxr-xr-x  2 root root     4096 Aug 17 02:57 wwlibs
```

So we can verify that `specify.jar` is accessible within the container at `/opt/specify7/config/specify.jar`.

Now it seems the `SPECIFY_THICK_CLIENT` location may be causing the issue. I can fix the issue I see in the `specify7-1` container upon startup by changing it in the `specify_settings.py` file:

```py
# Specify 7 requires the files from a Specify 6 install.
# This setting should point to a directory containing an installation
# of Specify 6 of the same version as the Specify database.
THICK_CLIENT_LOCATION = '/opt/specify7'
```

Adding `/config` looks like this:

```py
THICK_CLIENT_LOCATION = '/opt/specify7/config'
```

Now no startup errors! However, this of course doesn't solve the problem. Now the other files aren't accessible properly, so naturally an error appears:

image

@acwhite211 How have you been testing this? Am I missing something? When you can, please provide testing instructions.

grantfitzsimmons:
@maxpatiiuk 
> @acwhite211 yes, icons are still used, but only for compatibility with sp6 forms.

Icons are also used when thumbnails aren't generated for attachments and for table headings in the WorkBench (until https://github.com/specify/specify7/pull/5176)

>This PR might be an excellent time to remove that feature.

I agree that this should be done, but I believe this is not the right time due to other pressing priorities. We must focus our efforts on delivering GeoSpecify on schedule. This should be spun off into a different issue.

grantfitzsimmons:
>there also appear to be many reports/labels files - you will need to check with Grant on these but I was under impression that all reports are hand-crafted or done in jasper studio/irepo, so I am not sure what these files are about

@acwhite211 There is no need to keep any of the `*.jrxml` files for Specify 7. @maxpatiiuk is right, these were either hand-crafted or used as defaults in the past. Even in the latest version of Specify 6, you can no longer use these as templates.

acwhite211:
So, for now I added the `specify.jar` file to the config directory, but I'm wondering if i would be better to just remove the functions that use it, so that we don't need to have the jar file at all.

Since we would be freezing the Specify version to 6.8.03, we wouldn't need to have the whole jar file just to verify the version.

Here are the places that use it, so if we remove these functions, then we wouldn't need the jar file.

https://github.com/specify/specify7/blob/764f22993460845cfb8ad2bfbd930aa5cd45d568/specifyweb/specify/views.py#L130-L141

https://github.com/specify/specify7/blob/764f22993460845cfb8ad2bfbd930aa5cd45d568/specifyweb/specify/views.py#L144-L150

https://github.com/specify/specify7/blob/764f22993460845cfb8ad2bfbd930aa5cd45d568/specifyweb/specify/check_versions.py#L9-L29

https://github.com/specify/specify7/blob/764f22993460845cfb8ad2bfbd930aa5cd45d568/specifyweb/context/views.py#L625-L649

acwhite211:
It's working locally for me. Trying to figure out why the files aren't there in the GitHub test.

grantfitzsimmons (changes requested):
image

acwhite211:
The current commit works locally, just not on GitHub.

So just having `specify.jar` in the config directory worked for me locally, but I couldn't get it to work on GitHub without failing.  When the back-end test ran on GitHub, the error `FileNotFoundError: [Errno 2] No such file or directory: '/opt/specify7/config/specify.jar'` would always occur during the migrations.  So, I unzipped the jar file and put all of its files into a directory `/config/specify_jar_dir`, but it seems it also fails getting the files fails on GitHub.  Don't like this solution, hope to fix in the future.  I'm not exactly sure what is going on, I reverted the code back to the using the jar file, and then added code to have GitHub search the whole file system for the specify.jar file to see if that allows GitHub to find the file, but now there is an issue with the DockerHub token.

acwhite211:
Not sure why the dockerhub token is not working??

grantfitzsimmons (changes requested):
This is likely the case with other tables we've added recently...

image

On `issue-4828`

[Specify 7 System Information - 2024-09-28T02_00_19.978Z.txt](https://github.com/user-attachments/files/17171857/Specify.7.System.Information.-.2024-09-28T02_00_19.978Z.txt)

image


Not even camel case...

grantfitzsimmons (commented):
While I do think we need a `geology.views.xml` file, I think we need to define the COG, COGType, COT, and COJO tables globally (in `global.views.xml`, preferably), so we should transplant those portions there.

As for the CO, Taxon, COA, Prep, etc. forms, those should go in `geology.views.xml` if they are specific to that discipline in particular.

grantfitzsimmons (changes requested):
While I do think we need a `geology.views.xml` file, I think we need to define the COG, COGType, COT, and COJO tables globally (in `global.views.xml`, preferably), so we should transplant those portions there.

As for the CO, Taxon, COA, Prep, etc. forms, those should go in `geology.views.xml` if they are specific to that discipline in particular.

config/botany/default_sgr_matchers.xml
melton-jason, line 1: This file is not used by Specify 7. 

It looks like this was used in Specify 6 for the SGR (Specify and Scatter, Gather, Reconcile) plugin, which was a tool which utilized Lifemapper to compare records in Specify with GBIF. 

https://github.com/specify/specify6/blob/master/src/edu/ku/brc/specify/tasks/SGRTask.java

config/defaultperms/prefsperms.xml
melton-jason, line 1: This is not used in Specify 7. 

Looks like Specify 6 uses the `prefsperms.xml` file to create the default UserType permissions when building the database: 

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/config/init/DataBuilder.java#L2538-L2555

Specify 6 also merges permissions with these defaults whenever the app is restarted: 

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/config/init/DataBuilder.java#L2673-L2688

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/Specify.java#L2622

config/defaultperms/tasks.xml
melton-jason, line 1: Not used in Specify 7. 

This is actually very similar to `prefsperms.xml`  in that Specify 6 uses these to create the initial permissions, and merges the permissions with these defaults whenever the app is restarted

Creating Permissions: 

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/config/init/DataBuilder.java#L2538-L2555

Merging Permissions: 

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/config/init/DataBuilder.java#L2673-L2688

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/Specify.java#L2622

config/defaultperms/dataobjs.xml
melton-jason, line 1: Not used in Specify 7. 

Same story as with the above `prefsperms.xml` and `tasks.xml`! 

Specify 6 uses these to create the initial permissions, and merges the permissions with these defaults whenever the app is restarted

Creating Permissions:

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/config/init/DataBuilder.java#L2538-L2555

Merging Permissions:

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/config/init/DataBuilder.java#L2673-L2688

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/Specify.java#L2622

config/disciplines.xml
melton-jason, line 1: I don't think Specify 7 actually uses this `disciplines.xml`. 

Specify 7 has a hardcoded list of disciplines and their types already: 

https://github.com/specify/specify7/blob/2a0b516e3e249816a90fb0c196875670ca41cbf0/specifyweb/context/app_resource.py#L24-L41

config/icon_factory_mapping.xml
melton-jason, line 1: Specify 7 doesn't use this file. 

Specify 6 uses this file to map Java classes (like `edu.ku.brc.specify.datamodel.CollectionObject`) to icon names, as well as resolve "submappers" to Java classes and get the collection of datamodel classes they map. 

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/ui/RepresentativeIconFactory.java#L144-L196

config/ireportconfig.xml
melton-jason, line 1: Specify 7 doesn't use this file. 

Specify 6 doesn't use it either. The only references to it are commented out: 

https://github.com/specify/specify6/blob/3edca4fdb26a630e1be19724cf9d78b5d95ef371/src/edu/ku/brc/specify/tools/ireportspecify/MainFrameSpecify.java#L382-L394

config/specify_workbench_upload_def.xml
melton-jason, line 1: Specify 7 doesn't use this file. 

Looks liek Specify 6 used this in conjunction with `specify_workbench_datamodel.xml` to show which relationships could be mapped in the Specify 6 WorkBench.

maxpatiiuk:
@melton-jason Amazing job with the analysis!

config/backstop/collstats.xml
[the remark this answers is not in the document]
grantfitzsimmons: Just in case they need a hint on how to run some oddly defined SQL queries :joy:

specifyweb/settings/specify_settings.py
```diff
@@ -2,7 +2,7 @@
 # Specify 7 requires the files from a Specify 6 install.
 # This setting should point to a directory containing an installation
 # of Specify 6 of the same version as the Specify database.
-THICK_CLIENT_LOCATION = '/opt/Specify'
+THICK_CLIENT_LOCATION = '/opt/specify7'
```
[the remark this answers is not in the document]
melton-jason: I think the main problem here is how to resolve changes in our defaults with the corresponding files in the `THICK_CLIENT_LOCATION`. 

Assuming they make no changes to the `THICK_CLIENT_LOCATION` or we keep the `THICK_CLIENT_LOCATION` as `/opt/Specify`, users' will still be using their old defaults and will not have access to any additions or changes we make to the defaults. 

For new files (such as new appresources or views in a new discipline type), we can feasibly develop another fallback in the code when searching files to look in the `specify/config` files if an appresource or viewset is not found in the database or the user's file system. 
Alternatively, we can add a migration to add the files directly to the database, but I haven't yet gone through the possible complications of doing so. 

It's much trickier to handle updates to existing files. 
Consider the changes made to `common.views.xml` in https://github.com/specify/specify7/pull/5266/commits/d64f31bfc1c2fcb710a7f6cc5e55dc4a20c845ea. 
Even with the above solution for new files, there would be no easy way to modify the local `common.views.xml` to include the added `CollectionObjectType`, `CollectionObjectGroup`,  `CollectionObjectGroupJoin`, and `CollectionObjectGroupType` views + viewdefs (the views + viewdefs would conceptually need to be merged into a views file which may or may not have had changes from the original view) 

My first impression of a solution is to create a new viewset file, include the views we wish to add, and add the file to the `viewset_registry`. 
This still only addresses adding views which do not currently exist. 
There is still a problem of modifying existing views and appresources.

config/backstop/app_resources.xml
maxpatiiuk, line 1: while this file is used at the moment, could we refactor back-end to not use app_resources.xml files?

e.g aren't all UIFormatters XML files of the same name, thus removing the need for this file?

it seems that this file is used at present to tell specify what files are defined in a given folder - wouldn't it be better to simply check if the file is present on the file system?

config/backstop/global.views.xml
```diff
@@ -0,0 +1,3340 @@
...
+            class="edu.ku.brc.specify.datamodel.ReferenceWork"
+            resourcelabels="false">
+            <desc><![CDATA[The ReferenceWork SubForm.]]></desc>
+            <altviews>
+                <altview name="ReferenceWorkSub View" viewdef="ReferenceWorkSub" mode="view"  default="true"/>                
+                <altview name="ReferenceWorkSub Edit" viewdef="ReferenceWorkSub" mode="edit"/>                
+            </altviews>
+        </view>
+        
+        <view name="WorkbenchTemplate"
```
maxpatiiuk, line 224: should we remove views for sp6-only tables?

config/backstop/global.views.xml
maxpatiiuk, line 1: this entire file could potentially be removed - need to double check

config/backstop/system.views.xml
```diff
@@ -0,0 +1,685 @@
...
+        <view name="ExcelExportInfo"
+            class="java.util.Hashtable"
+            resourcelabels="false">
+            <desc><![CDATA[]]></desc>
+            <altviews>
+                <altview name="ExcelExportInfo Edit" viewdef="ExcelExportInfo" mode="edit" validated="true"/>
+            </altviews>
+        </view>
+        
+        <view name="ChangePassword"
```
maxpatiiuk, line 131: remove.
search for all views that don't have a `class="edu.ku.brc.specify.datamodel.*` attribute and consider removing them

config/bird/es_config.xml
maxpatiiuk, line 1: what is this used for?

config/bird/dataentry_task.xml
maxpatiiuk, line 1: I don't think this file is used anymore by newer specify 7 versions - need to check

config/schema_localization.xml
maxpatiiuk, line 1: is it safe to remove schema localization for sp6-only tables or is it too early for that?

grantfitzsimmons:
@maxpatiiuk We are finalizing this PR and will address your comments in a future PR. Thank you for your feedback! 😄

config/bird/dataentry_task.xml
melton-jason: This is still used at the moment by Specify 7: when configuring which tables appear in the DataEntry, the file is used for the "Copy Specify 6 Settings" option. 

Screenshot 2024-10-08 at 10 37 14 AM

I left a comment in my earlier review for this file, which has a little more detail: https://github.com/specify/specify7/pull/5266#discussion_r1785556984

config/backstop/global.views.xml
melton-jason: There are a lot of views for tables here that are only defined in this file: 

- Address
- AgentSpecialty
- Agent
- AgentVariant
- GroupPerson
- Author
- Journal
- ReferenceWork
- TaxonCitation
- Taxon
- Geography
- Storage
- GeologicTimePeriod
- LithoStrat
- ObjectAttachmenmt
- CollectionObjectAttachment

The other views defined here can likely be removed

maxpatiiuk:
that means all these useless files will end up in the repository's git history forever 😮‍💨

edit: I guess that's ok. Jason already filtered out most of the needless files

maxpatiiuk (approved):

CarolineDenis:
TODO: 
All of our Specify Cloud instances, self-hosted users, and test panel configurations assume that the Specify 6 Docker image will be the source of the static files that Specify 7 depends on. 
- [ ] We need to add support for Specify 7-sourced static files in our [docker-compositions repository](https://github.com/specify/docker-compositions/issues/31)
- [ ] We need to extend the Specify 7 Test Panel to allow staff to test both [post-static-files and post-static-files PR versions of Specify 7](https://github.com/specify/specify7-test-panel/issues/121).

acwhite211:
PR for editing Specify Cloud deployment https://github.com/specify/docker-compositions/pull/32
PR for editing Test Panel deployment https://github.com/specify/specify7-test-panel/pull/122

For the test-panel, the main change is to edit nginx rewrite for config to static-files volume path. The current solution will require all new branches to have config in the Specify 7 container, instead of in a separate Specify 6 container. Further work will be needed in order to support new and old branches where the config location is different for the nginx rewrite path to point to. Need a good way of specifying if a branch uses the new or old config setup.

CarolineDenis (approved):

sharadsw (approved):

PR #5300, opened 2024-10-02
Add kml selection exports

### LINKED ISSUE #3589: Query Builder export to KML ignores selected rows

alesan99:
Fixes #3589

Adds frontend implementation of KML exports using jsonToXml

WIP

### Checklist

- [ ] Add automated tests

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
@@ -22,6 +22,8 @@ import { QueryButton } from './Components';
 import type { QueryField } from './helpers';
 import { hasLocalityColumns } from './helpers';
 import type { QueryResultRow } from './Results';
+import { format } from '../Formatters/formatters';
+import { jsonToXml } from '../Syncer/xmlToJson';
 
 export function QueryExportButtons({
   baseTableName,
@@ -89,16 +91,23 @@ export function QueryExportButtons({
     'exportCsvUtf8Bom'
   );
 
+  function formatExportFileName(
+    file_extension: string
+  ): string {
+    return `${
+      queryResource.isNew()
+        ? `${queryText.newQueryName()} ${genericTables[baseTableName].label}`
+        : queryResource.get('name')
+    } - ${new Date().toDateString()}.${file_extension}`;
+  }
+
   /*
    *Will be only called if query is not distinct,
    *selection not enabled when distinct selected
    */
-  async function exportSelected(): Promise<void> {
-    const name = `${
-      queryResource.isNew()
-        ? `${queryText.newQueryName()} ${genericTables[baseTableName].label}`
-        : queryResource.get('name')
-    } - ${new Date().toDateString()}.csv`;
+  
+  async function exportCsvSelected(): Promise<void> {
+    const name = formatExportFileName('csv');
 
     const selectedResults = results?.current?.map((row) =>
       row !== undefined && f.has(selectedRows, row[0])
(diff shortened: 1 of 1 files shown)


realVinayak:
While I'm sure that this is done to mimic CSV on frontend, I think a better implementation would be just to make an API call, but this time adding  an extra filter to just select the ids that the user chose. Benefits:
1. Much simpler to implement
2. If in future we find a bug in KML on backend, we don't need to waste time fixing on frontend.
3. No need to reimplement the wheel
4. Most importantly (/s), I'll need to be doing something similar for batch-edit BUT I plan on making it a backnd way (snding filtrs). So, if you implement it right, I can reuse it! If you want, you can use `wb_improvements` to implement it straight-away (I'm not sure where it lies in priority, so ask @CarolineDenis too).

@maxpatiiuk thoughts? I think, even for CSV, it would have been better to directly use backend.

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -125,6 +135,148 @@ export function QueryExportButtons({
...
+    const filteredResults = filterArray(selectedResults);
+
+    const columnsName = fields
+      .filter((field) => field.isDisplay)
+      .map((field) =>
+        generateMappingPathPreview(baseTableName, field.mappingPath)
+      );
+
+
+    let placemarkTarget: any = [];
```
maxpatiiuk, line 158: reduce usages of `any`.
in this case the type `XmlNode` might be better

specifyweb/frontend/js_src/lib/components/QueryBuilder/Export.tsx
```diff
@@ -125,6 +135,148 @@ export function QueryExportButtons({
...
+
+    filteredResults?.forEach((result: any) => {
+      let dataTarget: any = [];
+
+      // <ExtendedData>
+      let extendedDataTarget: any = [];
+      
+      fields.forEach((field, index) => {
```
maxpatiiuk, line 166: this could be implemented using `.map()` too

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

that would match the code style in the rest of the codebase and reduce the need for explicitly specifying typescript types

specifyweb/frontend/js_src/lib/components/Syncer/xmlToJson.ts
```diff
@@ -50,7 +50,7 @@ export const xmlToJson = (element: Element): XmlNode => ({
 /**
  * Reverse conversion to JSON
  */
-export function jsonToXml(node: XmlNode): Element {
+export function jsonToXml(node: any): Element {
```
maxpatiiuk, line 53: please refrain from using `any` in our codebase.
`any` disabled type-checking, which negates the benefits of TypeScript

you can read a quick reference guide on TypeScript if you wish to be more familiar with TypeScript concepts: https://www.typescriptlang.org/docs/handbook/typescript-in-5-minutes.html
you can also use chatgpt to explain TypeScript errors

if you are using VS Code, you can also install this extension to make the error messages more readable: https://marketplace.visualstudio.com/items?itemName=yoavbls.pretty-ts-errors

maxpatiiuk (changes requested):
Good use of the utility functions that we have available!

Nevertheless, I agree with Vinny that if it would reduce duplication, we should do this work on the back-end.

---

Aside from that, better communication about who is working on what is always nice to avoid redundant effort

specifyweb/frontend/js_src/lib/components/Syncer/xmlToJson.ts
alesan99: Aha, I didn't think my code would be reviewed yet 😅 I promise this was a temporary hack as I was still figuring out the xml types.

alesan99:
> While I'm sure that this is done to mimic CSV on frontend, I think a better implementation would be just to make an API call, but this time adding an extra filter to just select the ids that the user chose. Benefits:

I agree. I already came across issues myself having to implement the same feature twice on my last CSV export PR.

I think the only downside is we'd lose the instant downloads from the frontend exports.
But I'd like to take a look into it and see if I'm able to do that.

specifyweb/frontend/js_src/lib/components/Syncer/xmlToJson.ts
realVinayak: Ah, that's partially my mistake, I didn't mean to trigger a code review. Nevertheless, these resources by Max are good starting points!

realVinayak:
I don't think it necessarily matters, especially for KML exports. For CSV, it is quite handy to do it within a few clicks, so don't change that (or users wouldn't like extra clicks needed now). For KML, definitely go the backend route. You'd learn about the query builder API quite a bit in the process!

maxpatiiuk (approved):
👍

emenslin (changes requested):
### Testing instructions

- In query builder, create a locality query with coordinate data (like Latitude1 and Longitude1).
- Run the query and "Create KML"
- [ ] Make sure the KML was exported correctly. The result should be the exact same as one created on production.
- Select any rows of the query results.
- [x] Make sure the KML includes ONLY the rows that were selected. (Also keep in mind that any rows without valid coordinate data will not be exported.)
- [x] Make sure CSV exports still work.


If you try to create KML without selecting any rows nothing happens. Creating a KML and selecting rows does work along with creating a CSV with or without selecting rows.

pashiav (approved):
- In query builder, create a locality query with coordinate data (like Latitude1 and Longitude1).
- Run the query and "Create KML"
- [X] Make sure the KML was exported correctly. The result should be the exact same as one created on production.
- Select any rows of the query results.
- [X] Make sure the KML includes ONLY the rows that were selected. (Also keep in mind that any rows without valid coordinate data will not be exported.)
- [X] Make sure CSV exports still work.
---

Looks great! Everything works as expected.

CarolineDenis (approved):

emenslin (approved):
### Testing instructions

- In query builder, create a locality query with coordinate data (like Latitude1 and Longitude1).
- Run the query and "Create KML"
- [x] Make sure the KML was exported correctly. The result should be the exact same as one created on production.
- Select any rows of the query results.
- [x] Make sure the KML includes ONLY the rows that were selected. (Also keep in mind that any rows without valid coordinate data will not be exported.)
- [x] Make sure CSV exports still work.

Looks good!!

PR #5354, opened 2024-10-31
Add get_spmodel_class to replace getattr for Specify models, temp solution

### LINKED ISSUE #5349: All ages tables throw error when opening uniqueness rules in Schema Config
Open schema config 
Choose one of the age table relativeAge - absoluteAge - absoluteAgeCitation, ... 
Open uniqueness rule tool 
Click add a rule 
See crash

acwhite211:
Fixes #5349

Avoid problems with getattr when getting model class names from Specify models due to capitalization differences. In order to maintain python class name readability, we create a utility function that can handling getting the class by name, even if the capitalization in the names don't match.

The get_spmodel_class functions is added in each file it is needed for now in order to avoid circular imports, I will make better in future.

### Checklist

- [ ] Add automated tests

### Testing instructions
Open schema config
Choose one of the age table relativeAge - absoluteAge - absoluteAgeCitation, ...
Open uniqueness rule tool
Click add a rule
See crash

### DIFF
--- a/specifyweb/specify/tree_ranks.py
+++ b/specifyweb/specify/tree_ranks.py
@@ -113,11 +113,23 @@ TREE_RANK_TO_ITEM_MAP = {
     'Tectonicunittreedefitem': 'Tectonicunit'
 }
 
+def get_spmodel_class(model_name: str):
+    try:
+        return getattr(spmodels, model_name.capitalize())
+    except AttributeError:
+        pass
+    # Iterate over all attributes in the models module
+    for attr_name in dir(spmodels):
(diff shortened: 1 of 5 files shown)


specifyweb/specify/tree_ranks.py
```diff
@@ -113,11 +113,23 @@
     'Tectonicunittreedefitem': 'Tectonicunit'
 }
 
+def get_spmodel_class(model_name: str):
```
CarolineDenis, line 116: the issue is not resolved

PR #5366, opened 2024-11-01
Fix tree structure issues due to an old WorkBench bug

### LINKED ISSUE #5131: Create a migration to fix tree structure issues due to an old WorkBench bug
Since the release of the WorkBench (v7.6.1) until sometime around the release of Specify v7.9.0, taxon records uploaded via the WorkBench did not have `isAccepted` as true by default.

This results in a "Bad tree structure" error, preventing children from being added to any node (or any descendent of any node) that has `IsAccepted` as false yet still is not a synonym.

image

The underlying issue is not apparent to the user nor to some troubleshooting this.

We should automatically fix all such cases.

You can return erroneous not accepted Taxon records here:

```sql
SELECT t1.TaxonID 
FROM taxon t1 
LEFT JOIN taxon t2 ON t1.AcceptedID = t2.TaxonID 
WHERE t1.IsAccepted = 0 AND t2.TaxonID IS NULL
```
This SQL query performs the following operations:

1. `SELECT t1.TaxonID`: This part of the query selects the `TaxonID` column from the `taxon` table, which is aliased as `t1`.

2. `FROM taxon t1`: This part of the query specifies the `taxon` table as the primary table, and assigns the alias `t1` to it.

3. `LEFT JOIN taxon t2 ON t1.AcceptedID = t2.TaxonID`: This part of the query performs a left join between the `taxon` table (aliased as `t1`) and the `taxon` table (aliased as `t2`). The join condition is that the `AcceptedID` column in `t1` must match the `TaxonID` column in `t2`.

4. `WHERE t1.IsAccepted = 0 AND t2.TaxonID IS NULL`: This part of the query applies a filter to the result set. It selects only the rows where:
   - `t1.IsAccepted` is 0 (i.e., the taxon is not accepted)
   - `t2.TaxonID` is `NULL` (i.e., there is no matching accepted taxon in the `t2` table)

This essentially retrieves the `TaxonID` values for all taxa that are not accepted, and for which there is no corresponding accepted taxon in the database. This should only include problem taxa.

To fix cases like this, we can simply run:
```sql
UPDATE taxon 
SET IsAccepted = TRUE 
WHERE TaxonID IN (
    SELECT t1.TaxonID 
    FROM taxon t1 
    LEFT JOIN taxon t2 ON t1.AcceptedID = t2.TaxonID 
    WHERE t1.IsAccepted = 0 AND t2.TaxonID IS NULL
);
```

but it seems like a good candidate for a migration so that this is handled on all databases once since the issue is most likely widespread and present on self-hosted institution databases as well.

grantfitzsimmons:
Fixes #5131

> [!WARNING]  

This came up again from a user that was self-hosted so I thought this could be a more meaningful fix.

This fixes an issue due to a bug in the WorkBench prior to 7.9.0 that did not set accepted nodes to IsAccepted = 1 when they were not synonyms.

This was done on all hosted databases, but self-hosted users are still potentially facing bad tree structure errors due to this.

The migration is also fully reversible.

```
specify@c82512644edc:/opt/specify7$ ve/bin/python manage.py migrate specify 0010_updateDelete_parentcojo
Operations to perform:
  Target specific migration: 0010_updateDelete_parentcojo, from specify
Running migrations:
  Unapplying specify.0011_fix_accepted_taxon... OK
```

### Testing instructions

_I made this hard since I fixed all cases on hosted databases and on the test panel... but the scope of this is pretty small. These testing instructions are overkill._

Essentially, this is running the following SQL statement:
`UPDATE taxon t1 SET IsAccepted = TRUE WHERE t1.IsAccepted = 0 AND t1.AcceptedId IS NULL`

**Before running the migration:**
1. On `production`, manually modify `taxon` records (via SQL) to set `IsAccepted` to `0` (false) in cases where that record is not a synonym
2. Try to create a child under that node in Specify (should get a "bad tree structure" error)

**After running the migration:**
1. After starting the container on `issue-5131`, try to create a child under that node in Specify (should no longer see a "bad tree structure" error
2. Check that the modified `taxon` record now is shown as accepted (`IsAccepted` = `1` (true)) in all cases where it is not a synonym.
3. Check that all `taxon` records that are synonyms are not now accepted.


specifyweb/specify/migrations/0011_fix_accepted_taxon.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 3.2.15 on 2024-11-01 19:21
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('specify', '0010_updateDelete_parentcojo'),
```
grantfitzsimmons, line 9: TODO: Update this and the migration number to the appropriate iteration based on the current `production` branch at the time of merge

CarolineDenis (approved):

maxpatiiuk (approved):

grantfitzsimmons:
@CarolineDenis This has been updated to be completed after the last current geo migration (after `0019_remove_parentCog`)

Areyes42 (changes requested):
**Before running the migration:**

- [x]  Try to create a child under that node in Specify (should get a "bad tree structure" error)

**After running the migration:**

- [x]  After starting the container on `issue-5131`, try to create a child under that node in Specify (should no longer see a "bad tree structure" error)
- [x]  Check that the modified `taxon` record now is shown as accepted (`IsAccepted` = `1` (true)) in all cases where it is not a synonym.
- [x]  Check that all `taxon` records that are synonyms are not now accepted.
***
Accepted states look good after migrations; however, it looks like this created a new issue with child node creation. When creating a new child, Specify disregards the taxon rank chosen in the creation dialog (See recordings). This behavior is not present in production.

Screenshot 2025-01-15 at 12 08 12 PM

issue-5131:

https://github.com/user-attachments/assets/5a0d782f-804a-4d59-a9ff-548f16479de0

Production:

https://github.com/user-attachments/assets/8b4976f8-f9ef-4a00-8d77-07bd7c2e575a

grantfitzsimmons:
@Areyes42 Can you test this again now that the latest changes from `production` are included in this branch?

Areyes42 (approved):
Looks good now!

specifyweb/specify/migrations/0011_fix_accepted_taxon.py
grantfitzsimmons: Added this to a new app instead, `patches`

combs-a (approved):
Post-running the migration, everything looks good and works perfectly. Set all of the tree nodes' isAccepted to 0, switched to the branch and rebuilt the containers, and they were all set to 1 except for synonymized nodes. Was able to add children to the tree again.

PR #5385, opened 2024-11-07
Rescope upload tree when there is only 1 tree in the upload plan

### LINKED ISSUE #5338: Cannot validate taxon in workbench if it is not the original tree
**Describe the bug**
On a db with more than one taxon tree if you try to validate/upload taxon through the workbench for the tree that is not the original tree it does not work. You can map the fields fine but when you try to validate it no cells are updated even though they should be and if you try to upload anyway nothing happens. 

**To Reproduce**

1. Go to any db with more than one taxon tree (e.g. kufish2_12) 
2. Go to workbench and select taxon as base table
3. Add fields from a tree that is not the original (e.g. in kufish2_12 the 'test' tree) 
4. Go to grid and fill out some cells 
5. Validate 
6. See cells aren't updated 

Alternatively you can also import a taxon data set and get the same results 

**Expected behavior**
The taxon should validate and upload no matter what tree they're in 

**Screenshots**

New tree:

https://github.com/user-attachments/assets/c57c7e00-5737-40c5-923d-fa44078e63bb

Validating when it's the original tree: 

https://github.com/user-attachments/assets/760d71a7-eb16-4fc9-977e-b94807bdff12

- OS: Windows 11
- Browser: Chrome
- Database Name: kufish2_12
- Collection name: voucher
- User Name: spfishadmin

sharadsw:
Fixes #5338

The upload tree wasn't being rescoped in cases where upload plan had only 1 tree, due to which upload would be scoped to the default tree of the collection which meant upload didn't work if the upload tree of the plan was not the same as the default tree

### Checklist

- [ ] Add automated tests

### Testing instructions

- Go to the workbench
- Create an upload plan with only 1 tree which isn't the original tree
- [ ] Verify upload works


CarolineDenis (approved):

sharadsw:
What should be the existing behavior in both cases? It should use the existing root?

pashiav (approved):
### Testing instructions



- Go to the workbench
- Create an upload plan with only 1 tree which isn't the original tree
- [X] Verify upload works
---
Looks good - upload works! 

Screenshot 2024-11-18 at 1 52 16 PM

Screenshot 2024-11-18 at 1 51 51 PM

alesan99 (approved):
### Testing instructions


- Go to the workbench
- Create an upload plan with only 1 tree which isn't the original tree
- [x] Verify upload works

chrome_HKQ0HgaNKk


Tried this on multiple trees that were not working with uploads previously, they now work 👍👍👍

Note: I got an error with "taskstatus: successful" once that quickly disappeared but I couldn't recreate it. I've seen it before so I think its just a quirk with the workbench

PR #5389, opened 2024-11-07
Use platform over appversion on global navigator object

melton-jason:
> This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/appVersion#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

From Mozilla on appVersion: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/appVersion 

Recently, localization tests began failing because of an invalid setting of `global.navigator`: 

https://github.com/specify/specify7/actions/runs/11707985658/job/32609369992

https://github.com/specify/specify7/blob/7fe1b8006a86ebdfd2271b6f80de7146c2103867/specifyweb/frontend/js_src/lib/tests/jsdom.js#L10-L14

This failing test was remedied in https://github.com/specify/specify7/commit/76871c704f7667f5ea3254693f00fc12358d1d5d, which functionally completely removed the reassignment of `global.navigator`. 

However, it appears that support for `navigator.appVersion` has been dropped in some capacity, as `navigator.appVersion`  was returning undefined:

https://github.com/specify/specify7/actions/runs/11726874323/job/32666586609

https://github.com/specify/specify7/blob/7fe1b8006a86ebdfd2271b6f80de7146c2103867/specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx#L67-L69

This PR cleans up the code in the JSDOM configuration file, and uses `navigator.platform` instead of `appVersion`. 

⚠️ Note that `navigator.platform` is itself deprecated: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform

However, this use case is explicitly mentioned in the Mozilla docs, and may be a better alternative until a better solution is found... 

> But there is one case where, among the options you could use, navigator.platform may be the least-bad option: When you need to show users advice about whether the modifier key for keyboard shortcuts is the ⌘ command key (found on Apple systems) rather than the ⌃ control key (on non-Apple systems)

### Checklist


CarolineDenis (approved):

specifyweb/frontend/js_src/lib/tests/jsdom.js
```diff
@@ -1,18 +1,17 @@
 import { JSDOM } from 'jsdom';
 
 // Create a new JSDOM instance
+/**
+ * FEATURE: Allow customizing the JSDOM userAgent, platform, and other global
+ * properties
+ * See https://github.com/jsdom/jsdom#advanced-configuration
+ * See https://github.com/specify/specify7/pull/5389
```
maxpatiiuk, line 8: that should probably be done using mocking:

in beforeAll in some test file set userAgent/platform/...
in afterAll undo the change

specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
```diff
@@ -64,7 +64,12 @@ const isDarkMode = ({
   isRedirecting,
 }: PreferencesVisibilityContext): boolean => isDarkMode || isRedirecting;
 
-const altKeyName = globalThis.navigator?.appVersion.includes('Mac')
+/**
+ * REFACTOR: navigator.platform is deprecated. Implement better feature
+ * detection for browsers using Mac.
+ * See https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform
+ */
+const altKeyName = globalThis.navigator?.platform.startsWith('Mac')
```
maxpatiiuk, line 72: this is exactly the use case that was quoted by you from the MDN article, so probably using navigator.platform in the right choice in this case?

maxpatiiuk (approved):

PR #5394, opened 2024-11-12
Add a tool to create root node in trees

### LINKED ISSUE #5391: Create a node root creation tool
When using the new tree creation tool in tree viewer new ranks get added but no root node which prevents the user to add any node. 

Add a way to create a root node in the tree viewer for new tree

CarolineDenis:
Fixes #5391

### Checklist

- [ ] Add automated tests

### Testing instructions

- open the taxon tree viewer 
- create a new tree
- [ ] verify there is now a + button where the root node is usually 
- click the + 
- [ ] verify a root node is created (This might take a few seconds, and then it will reload.) 
- add a child to the new root


alesan99 (approved):
### Testing instructions

- open the taxon tree viewer 
- create a new tree
- [x] verify there is now a + button where the root node is usually 
- click the + 
- [x] verify a root node is created (This might take a few seconds, and then it will reload.) 
- add a child to the new root 

Working!👍👍👍

Side note: The root names aren't being displayed correctly outside the tree at the moment, presumably because they do not have a fullName, but that's a separate issue

image

combs-a (changes requested):
### Testing instructions

- open the taxon tree viewer 
- create a new tree
- [X] verify there is now a + button where the root node is usually 
- click the + 
- [X] verify a root node is created (This might take a few seconds, and then it will reload.) 
- add a child to the new root

Everything's looking good, but when you add a child, the parent name is not root for some reason? Happened on multiple trees. The record formatter is just the default (Full Name for taxon) so it shouldn't be this name

image

Alejandro also ran into this so if it does get put into a separate issue that's fine, but if it can be addressed here I think that'd be better

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
```diff
@@ -106,6 +108,28 @@ export function Tree<
     [baseUrl, statsThreshold]
   );
 
+  const treeDefinition = treeDefinitionItems[0].treeDef;
+  const treeDefId = idFromUrl(treeDefinition);
+  const createRootNode = async (): Promise<void> => {
+    if (treeDefId === undefined) {
+      console.error('treeDefId is undefined');
```
maxpatiiuk, line 115: (optional) consider using softError instead of console.error
it is equivalent to console.error in production, but will produce a full-screen exception in development.
that is useful as it makes it less likely that bugs get unnoticed in development, while preventing non-critical errors from disrupting user's work

specifyweb/frontend/js_src/lib/components/TreeView/Tree.tsx
```diff
@@ -106,6 +108,28 @@ export function Tree<
...
+        {
+          method: 'POST',
+          headers: { Accept: 'application/json' },
+        }
+      )
+        .then(() => {
+          globalThis.location.reload();
+        })
+        .catch((error) => {
+          console.error('Error creating root node:', error);
```
maxpatiiuk, line 128: same

maxpatiiuk (approved):

specifyweb/specify/tree_extras.py
```diff
@@ -633,96 +639,73 @@ def print_paths(table, depth):
...
-        path=path_expr(table, depth),
-        parent_joins=parent_joins(table, depth),
-    ))
+    # nodenumbers according to the lexical ordering of the paths.
+    nodes = tree_model.objects.all().order_by('path')
+    with transaction.atomic():
+        for i, node in enumerate(nodes, start=1):
+            node.nodenumber = i
+            node.highestchildnodenumber = i
+            node.save()
```
realVinayak, line 693: Sorry for the stray comment, but I'd highly recommending not this approach (and, instead, using raw sql). Also, the previous approach renumbers the entire tree -- so is this (restricting to a specific tree, based on commit message) done for performance reasons? If so, then that's technically not perfect. Actually, this (restricting to a specific tree) will only be correct when you are doing moves (which keeps total count same)

Anything else, merge, add, delete changes the node numbers of that tree (either increments or decrements). In doing so, the rest of the trees need to be updated. Well, technically, you can just update the trees with root node numbers after the current tree in case of inserts, and the ones before for deletes, for merges, it's complicated. I don't think complexity is worth here, given that you guys would want to remove the node numbers entirely, so just keep it be the previous code.

PR #5399, opened 2024-11-14
Update README.md

### LINKED ISSUE #5398: Improve README
The current README could use a makeover to better direct new users on how to contact us, a better explanation of what we do, updated installation instructions, links to relevant documentation, and a clearer idea of the "big picture" of what we do.

grantfitzsimmons:
Fixes #5398

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,534 +1,72 @@
-
 # [Specify 7](https://www.specifysoftware.org/products/specify-7/)
 
-The [Specify Collections Consortium](https://www.specifysoftware.org) is pleased
-to offer Specify 7, the web implementation of our biological collections data
-management platform.
+## **Specify Collections Management Platform**
 
-We encourage members to use
-our [Dockerized compositions](https://github.com/specify/docker-compositions) of
-Specify 7. You can choose a version, make the necessary adjustments and then run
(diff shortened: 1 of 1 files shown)


maxpatiiuk (commented):
I know that the non-docter installation instructions were likely out of date, but it was the only way for non-members to setup specify (since docker compositions are reserved for members). I understand that we can't prioritize non-members, but perhaps we can move the non-docker instructions into a separate .md file and add a disclaimer?

PhatWheZ:
If you guys need assistance in refining and contributing to the docker-compositions keep me posted

grantfitzsimmons:
>I know that the non-docter installation instructions were likely out of date, but it was the only way for non-members to setup specify (since docker compositions are reserved for members). I understand that we can't prioritize non-members, but perhaps we can move the non-docker instructions into a separate .md file and add a disclaimer?

That is the current plan B, if plan A (give everyone `docker-compositions` access) does not succeed.

grantfitzsimmons:
@PhatWheZ You are always welcome to suggest changes to that repository, especially if you have ideas for improvement! I am happy to review them and provide feedback.

PR #4548, opened 2024-02-15
Locality and Geocoorddetail batch import tool

melton-jason:
Fixes #4319

The Locality and GeoCoordDetail batch import tool allows users to match and update the Latitude1, Longitidude1, and Datum fields for Locality records and associate new GeoCoordDetails for the matched Locality records. 

The primary motivation for the tool is such that the georefrencing for existing records in the database can take place outside of Specify (such as with [CoGe](https://coge.geo-locate.org/About.htm)), and there is no easy way to currently repatriate the data back into Specify and update those records. 

This process is facilitated by requiring the guids of the Locality records in the CSV. Specify will use these guids to match/find the corresponding Locality record(s). 
With the Locality record, the tool does the following: 
- If any `latitude1`, `longitude1`, and/or `datum` fields have values for the Locality record in the CSV, Specify will overwrite the current values in the database with those specified in the CSV
- If any other column in the CSV matches a non-relationship field name in the GeoCoordDetail table, Specify will delete the previous GeoCoordDetail associated with the Locality record and create a new one populated with field values from the CSV
- If there are any other columns in the CSV, Specify will display a warning to the user indicating there are unknown columns, but allow the user to proceed; the unknown columns will be ignored if the user proceeds. 

---
### Demo
Currently, access to the tool can be found in the User Tools menu: 

locality_import_button

The primary interface is very similar to that of importing a file to be used as a DataSet in the WorkBench. Specify supports different encodings and delimiters, and displays the first 100 rows of the file:

coge_preview

The `Import File` button can be pressed to initiate the parsing/upload. 

If there are any unknown columns in the dataset, a warning will first be displayed to the user so they can decide how to proceed: 

coge_column_warning

If there are any bad values discovered when parsing the results, Specify will display a downloadable message stating which line the error was found on along with a message with more information about how to resolve the parsing error:

parse_error

If the parsing and upload is successful, Specify will display a brief results page stating how many records of each table were affected, along with the option to create a RecordSet of modified Locality Records:

upload_results

--- 

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [ ] Add automated tests
- [x] Add relevant issue to release milestone

### Testing instructions

For testing purposes, here is an example set of data for use on the `KUFish` database: 
[kufish_coge_sample.csv](https://github.com/specify/specify7/files/15257843/kufish_coge_sample.csv)

[...]

PR #5417, opened 2024-11-26
Batch Edit: Support for editing basic fields

### LINKED ISSUE #5413: Batch Edit: Add support for editing simple fields
Component of #4929 

## Batch Edit Workflow
- Create a Query for the data you want to edit
- Click on Batch Edit in the query results
- Update any values 
- Validate
- Valid edited cells highlight as `Updated Cell` if its a simple field
- Click on commit to apply the changes

### Features
- Users must be able to batch edit data from a query
- Add Batch Edit as an option in the sidebar
- Allow batch editing simple fields in a table
- Enable rollback for edited fields
- Add permissions for Batch Edit
  - Users without the Batch Edit permission cannot access Batch Edit

Additional context from #4929:
> Modifying any field is possible, other than nodenumber, fullname, and highestchildnodenumber. If some fields get updated, only those fields are highlighted

### LINKED ISSUE #6209: Batch Edit: Lat/long values are updated even without making changes
Reported in #5417 

Lat/long values are counted as updated cells despite making no changes. However, after upload the values do not actually change. 

https://github.com/user-attachments/assets/4315a665-fe76-4edc-bea2-01735f74d747

The lat/long error is related to floating point precision and using wrapper classes to primitives. When latitude and longitude are compared to their previous values the comparison that happens is: Decimal('33.1493800000') == 33.14938 which cannot be true and is detected as a change. Relevant code for the comparison: 
https://github.com/specify/specify7/blob/1b2ab816c6b8229e66ebb8469f3d9bfea3003e7f/specifyweb/workbench/upload/upload_table.py#L1112-L1118

sharadsw:
Fixes #5413
Fixes #6209
Uses code from #4929

This PR pretty much has the same code as #4929 but updated to production. To minimize the scope for testing, batch edit for relationships is disabled in this PR and will be enabled in the follow up PR - batch edit for relationships. There are still some merge conflicts resulting from some newer code in prod which will be resolved in the next PR.

### Checklist

- [x] Add automated tests

### Testing instructions

- Create a query
- Click on Batch Edit. This will create a new Batch Edit dataset with columns involving relationships disabled (readonly)
- Change values you wish to edit
- Click Validate
- [ ] Verify changed values are highlighted as updated cells
- Click Commit. This will apply your changes
- [ ] Re-run the original query and verify the values changed as expected
- Click rollback
- [ ] Verify values were rolled back to original

#### Regression tests
- [ ] Retest #5489


realVinayak:
Nice observation. My fault for keeping an hidden feature secret: Another, use for data mapper is to control “ignore when blank”-type behavior per field basis. That’s actually supported in that PR ( even for batch edit data mapper, you can modify this per-field setting. Ask grant for more context )

the only things not read only are
1. Batch edit prefs 
2. Field behavior (never ignore, …)

search for usages of “readonlyspec”.

For purposes of this PR, you can theoretically not add data mapper (you also don’t need batch edit prefs, they don’t do anything when there are no relationships)

lexiclevenger (commented):
### Testing instructions

- [x] Verify changed values are highlighted as updated cells
- [x] Re-run the original query and verify the values changed as expected
- [x] Verify values were rolled back to original

Looks good! I tested the CO, CE, Agent, and Paleo Context tables.

---
I did find a couple weird things, but I'm not sure whether they are in the scope of this PR.

1. CO data sets with the Age field cause a Validation error:
`{
  "uploaderstatus": {
    "operation": "validating",
    "taskid": "de1c70b1-4e40-46ec-bd48-85d1f49b8d4f"
  },
  "taskstatus": "FAILURE",
  "taskinfo": "AttributeError(\"'Collectionobject' object has no attribute 'age'\")"
}`
https://geo20241205copymigration5413-issue-5413.test.specifysystems.org/specify/workbench/50

2. Trying to batch edit certain ranks in the Taxon table throws an error: 
[Specify 7 Crash Report - 2024-12-24T21_48_02.580Z.txt](https://github.com/user-attachments/files/18241999/Specify.7.Crash.Report.-.2024-12-24T21_48_02.580Z.txt)


https://github.com/user-attachments/assets/273e1a49-6634-4552-a2f0-f6ea0bd591a6

realVinayak (commented):
Nice, I'd recommending looking at the tree queries and the unit tests. For trees, I'd recommend getting TreeRankQuery to work (otherwise, just copy the previous code, and disable/refactor unit tests in stored_queries directory that run tree rank queries for tests.

specifyweb/stored_queries/queryfieldspec.py
```diff
@@ -281,24 +350,46 @@ def add_to_query(self, query, value=None, op_num=None, negate=False, formatter=N
...
+                    self.get_field(),
+                    orm_model,
+                    aggregator or formatter,
+                    cycle_detector,
+                )
         else:
             query, orm_model, table, field = self.build_join(query, self.join_path)
-            if self.tree_rank is not None:
-                query, orm_field = query.handle_tree_field(orm_model, table, self.tree_rank, self.tree_field)
+            if isinstance(field, TreeRankQuery):
```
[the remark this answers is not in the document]
sharadsw: yeah, tree queries broke. I'm gonna add TreeRankQuery back

specifyweb/stored_queries/tests/stored_queries_tests.py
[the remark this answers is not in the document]
sharadsw: thanks!

specifyweb/stored_queries/batch_edit.py
```diff
@@ -1012,6 +1013,7 @@ class BatchEditProps(TypedDict):
     recordsetid: Optional[int]
     session_maker: Any
     fields: List[QueryField]
+    omit_relationships: Optional[bool] = False
```
realVinayak, line 1016: because defaults don't work in typed dict, so this is actually illegal

realVinayak: Also, can you try running unit tests locally? it'll be faster, and you actually run specific tests (like just the ones which are erroring / failing). Refer to docs. 
https://docs.djangoproject.com/en/5.1/topics/testing/overview/
It'll be more faster to develop, rather than waiting for 10+ minutes

specifyweb/stored_queries/query_construct.py
```diff
@@ -71,25 +71,8 @@ def handle_tree_field(self, node, table, tree_rank, next_join_path, current_fiel
...
-        # for ancestor in ancestors:
-        #     field_spec = make_tree_field_spec(ancestor)
-        #     query, orm_field, field, table = field_spec.add_spec_to_query(query)
-        #     #field and table won't matter. rank acts as fork, and these two will be same across siblings
-        #     cases.append((getattr(ancestor, treedefitem_column) == treedefitem_param, orm_field))
-
-        # column = sql.case(cases)
-
-        # return query, column, field, table
+        column_name = next_join_path[0].name
```
realVinayak, line 74: Hmm could you leave a comment to revert back to the previous implementation in 4929? You'd need that to support multiple trees in batch-edit. I understand if you don't want to add that, but at least keep the commented out the code for future reference.

specifyweb/stored_queries/batch_edit.py
sharadsw: Yeah I've been running them locally but I've been getting some weird errors locally which don't actually happen on the build checks. Didn't know I could test them individually, thanks!

sharadsw:
@lexiclevenger Thanks! Age should be readonly now and tree queries hopefully shouldn't have any crashes

emenslin (changes requested):
- [x] Verify changed values are highlighted as updated cells
- [x] Re-run the original query and verify the values changed as expected
- [ ] Verify values were rolled back to original

I feel like this shouldn't have 'workbench' in it and should just say 'Batch Edit Data Sets' 

Screenshot 2025-01-10 134406


Search and replace can edit readonly fields. If you upload the changes don't carry over, however, the values do appear to be changed in the data set. 

https://github.com/user-attachments/assets/5ecc9002-c92d-4efb-a455-cb37043bbdc5


Results aren't showing up correctly 
image


I didn't make any changes so I'm not sure why this roll back failed but when a rollback fails it gives an error but then shows a sucess dialog 


https://github.com/user-attachments/assets/3002ab7f-e0c5-429b-8129-25cc73eaa5fe


Also, the rollback success dialog should say "it cannot be edited." instead of "it cannot be edit." 
image

Can't batch edit for taxon as it requires you to add rank -> name but you can't do that (see #5935) and on top of that it requires you to add every single rank so if you have multiple taxon trees it wants every rank from every tree. Based on this pr and what it is supposed to cover it might be worth it to just disable the batch edit button when a tree is the base table. 
image

I'm not sure if this happens elsewhere but in this example on the accession table I had a valid pick list value`In Process` but when I clicked batch edit, in the grid it became `inprocess` an invalid value

https://github.com/user-attachments/assets/8bd31b6b-2e8b-48d0-ac47-c2ec7d64dbed

combs-a (changes requested):
### Testing instructions

- Create a query
- Click on Batch Edit. This will create a new Batch Edit dataset with columns involving relationships disabled (readonly)
- Change values you wish to edit
- Click Validate
- [X] Verify changed values are highlighted as updated cells
- Click Commit. This will apply your changes
- [X] Re-run the original query and verify the values changed as expected
- Click rollback
- [X] Verify values were rolled back to original

Agree with @emenslin on the text changes, disabling the batch edit button for taxon for now.

About the rollback not being allowed when you didn't make changes, it might've been something along the lines of a required field being edited from previously having nothing. Not sure why it was missing a required field but 🤷 

I was able to recreate the issue with the picklists. Looks like it's pulling from the value field and not the title field for Picklist items.
image

Also able to recreate the search and replace.

realVinayak:
1. Eh, don't bother fixing search and replace, it should be clear because of highlighting that it does not actually get changed in the database. Maybe there is a setting in HOT to ignore readonly cells?

2. Ugh, results not showing up correctly might actually be a bug in HOT, I'm not sure. I think there was a related bug, I debugged and wrote it up in 4929 somewhere IIRC.

3. Hmm, if rollback fails, the "best" way to debug that would be to understand how rollback for batch edit works. When, initially, you run the query and make a dataset, the backend secretly, and shamefully, just makes two copies of the dataset. The former is the one that user modifies, and the latter is used for batch edit. When you commit your changes, and want to roll it back, the batch edit on the later (used internally) is performed. I wanted to test that more but whatever.

This is where it happens:
https://github.com/specify/specify7/blob/4d82298ac772122992afd7f767c3dc25e2e52faa/specifyweb/workbench/upload/upload.py#L513

I think what you could do, just to debug this, would be to set a breakpoint here
https://github.com/specify/specify7/blob/4d82298ac772122992afd7f767c3dc25e2e52faa/specifyweb/workbench/upload/upload.py#L588

Then, just find what row has the failure. Me and Grant talked about making the better. Technically, we know exactly what went wrong. Heck, for debugging, can you just add json of failed rows (or first failed row) to the exception it raises? 

The next lines are needed. Basically, when you have relationships, you could have uploaded a record that you don't need anymore, so you can pretty cutely just run original workbench rollback after this special batch edit rollback.

Another thing I wanted to do (as indicated in "cannot be edited") was to make it readonly after one rollback. Do it simply by removing the backer, and if the batch edit (isupdate = true), just make it readonly (add a flag from the backend?). Theoretically, this is not needed, as the process is idempotent, fancy,

4. I'm not guaranteering any of my code works for multiple trees, I didn't get to finish that up 😉. But, originally, no it doesn't ask for every rank. If you selected species (and if species is the last rank), it will not ask for more ranks. If you selected genus, and there are subspecies and species rank below it, then it would force you to select species and subspecies. Basically, all the ranks below the highest rank need to be selected. It gets complicated in cases of relationships (determinations -> taxon) otherwise.

Also, don't straight up disable in case of taxon. Only do that if there is a rank also selected. It differentiates between the two cases. I'd recommend going back to original PR and maybe ask Grant, if he has time, on how it differentiates them. There's a somewhat generic logic for it, shouldn't be hard to find, in the BatchEdit/index.tsx file.

5. I swear I tested picklists, and I thought it worked! In any case, a somewhat decent way (which could actually fix another issue floating around which asks for seeing title of picklist) would be to add a "case when" here https://github.com/specify/specify7/blob/4d82298ac772122992afd7f767c3dc25e2e52faa/specifyweb/stored_queries/format.py#L329-L336

I'm sorry for putting you through the last one, completely my fault. I really thought that picklist worked.

sharadsw:
- [x] Localization fixes
- [ ] ~Readonly cells can be search & replaced~ (Created #6100 since it's a problem with WB as well)
- [x] Results aren't visible
- [ ] ~Rollback errors~ (deferring to #6128 so we have one issue for all rollback errors)
- [ ] ~Tree rank errors~ (created #6127, Batch Edit will be disabled in tree queries for this PR)
- [ ] Picklist shows value instead of title

realVinayak:
Definitely recommend checking why rollback failed for [this ](https://github.com/specify/specify7/pull/5417#pullrequestreview-2543815320) case though, if the DB is around (it's simple enough to see what error is actually causing the issue at the code level, just that it's hard for showing it to the user) - I'm don't disagreeing with that picklists might be causing rollback failures, but for elizabeth's case, no change was done, so rollbacks should work (rollbacks also don't care about versioning).

I'm also curious what caused failure in elizabeth's case, since no change was made. If update (commit) was allowed, there's no reason why rollbacks should fail (because it is the same data).

sharadsw:
This can be tested again. Fixed localization stuff, results, picklist error and batch edit should now be disabled on tree tables. I have created separate issues for other errors.

I'm not able to reproduce the rollback error locally on that db, will try to investigate it on the test panel when its back up.

specifyweb/workbench/upload/tests/testscoping.py
```diff
@@ -130,7 +130,7 @@ def test_collection_rel_uploaded_in_correct_collection(self):
         result = do_upload(self.collection, rows, scoped_plan, self.agent.id)
 
         left_side_cat_nums = [n.zfill(9) for n in '32 23'.split()]
-        right_side_cat_nums = '999 888'.split()
+        right_side_cat_nums = [n.zfill(9) for n in '999 888'.split()]
```
realVinayak, line 133: This is wrong, the original values were correct. For context, this is related to collection relationships. On the right hand side, in this test, the values should not have been formatted (and extra 0s -- zfill -- should not have been needed).

Looks like collection is correct, but formatter is wrong. I haven't spent time debugging this, but can you try running this locally? This is an important bug because otherwise you'd be creating collectionobjects with wrong catalog number. CSIRO won't be happy about that. Here is how I would test it - with unit tests - you could test via UI too, but I think unit tests are easier + more surgical

1. Here put a print (yes, icky) here, `print(collection, tablename, fieldname, value, formatter)`
 https://github.com/specify/specify7/blob/ab12c9bdc544a1f94ed5d76f33a8fbf72f6d7bce/specifyweb/workbench/upload/parsing.py#L121

2. Run _this_ unit test. Assuming you are in docker container, it'd be something like

```
python3 manage.py test specifyweb.workbench.upload.tests.testscoping.ScopingTests.test_collection_rel_uploaded_in_correct_collection` --keepdb -v3
```

v3 is important for max verbosity and seeing print statement outputs.

I suspect that all you'd see is when tablename and fieldname are CO and catalognumber respectively, then `formatter` will be defined in one case (for first collection - left side) and then undefined for the other (right hand side). Further, you'd only be seeing only one collection, the left-side one (so it won't vary across the logs). Since formatter is undefined for right hand side collection catalognumber, this then causes the code here:  
https://github.com/specify/specify7/blob/ab12c9bdc544a1f94ed5d76f33a8fbf72f6d7bce/specifyweb/specify/parse.py#L50
to fetch the formatter for the wrong collection (that is, when formatting right-hand, it fetches formatter for left). This is because left side (base) is passed in bind(). That was also the rationale behind why I got rid of passing collection to bind() -- because it creates mistakes like this.

Can you, initially, confirm that this is the issue? Can you also send all the print outputs when you run that unit test?

Thanks for enabling unit tests!

specifyweb/stored_queries/format.py
```diff
@@ -335,6 +336,21 @@ def _fieldformat(self, specify_field: Field,
             _case = case(cases)
             return blank_nulls(_case) if self.replace_nulls else _case
         
+        if self.format_picklist:
+            schema_items = Splocalecontaineritem.objects.filter(
+                container__discipline=self.collection.discipline,
+                container__schematype=0,
+                container__name=table.table.lower(),
+                name=specify_field.name.lower(),
+            )
```
realVinayak, line 350: I think this is fine, but keep in mind that workbench has slightly different logic (it also checks collection picklists). Can you make them behavior same, ideally by wrapping the workbench logic in a single function and then reuse it here to fetch the picklist? Better yet, I'd recommend a function in specifyweb/specify so people later on can reuse this logic.

emenslin (changes requested):
- [x] Verify changed values are highlighted as updated cells
- [x] Re-run the original query and verify the values changed as expected
- [x] Verify values were rolled back to original

---

For tables where batch edit isn't supported yet (or never will be) can we keep it consistent? For the tables that were initially caught (e.g. collection, discipline, etc.) the batch edit button is there but disabled, for the trees there isn't a button at all. Can we chose one option for all of them? Personally I would prefer we keep the button but have it disabled but fully removing the button would also work depending on other @specify/ux-testing opinions. 

Trees (no button): 
Screenshot 2025-01-24 120631

Other tables (button disabled):
Screenshot 2025-01-24 120553

If you have a CO query that includes taxon in it you cannot batch edit it. We should consider a solution that would allow the user to create the data set and maybe make it read only for now like the other relationship fields. As I said in my previous review, it is not possible to map rank -> name in the query builder but users are going to want to see the taxon to utilize batch edit even if they cannot edit it. 
Screenshot 2025-01-24 122310

Also in that same screenshot I'm not sure why 'genus name' and 'species name' are listed twice, might be something else to look into. 

I had a validation fail and then there were a lot of other dialogs after I closed out of that one. This could be related to existing WB bugs but I wanted to bring it up in case it wasn't. 

https://github.com/user-attachments/assets/6394090d-873d-48e8-9d43-3023b8cb170c

The error is 
```
{
  "uploaderstatus": {
    "operation": "validating",
    "taskid": "a91fe85a-f80a-4632-9200-e5b229d77f65"
 },
  "taskstatus": "FAILURE",
  "taskinfo": "Exception(\"Attempting to change the scope of the record: Loan object (17) at BoundUpdateTable(name='Loan', static={}, parsedFields=[ParseResult(filter_on={'loannumber': '6'}, upload={'loannumber': '6'}, add_to_picklist=None, column='Loan Number', missing_required=None), ParseResult(filter_on={'loandate': datetime.date(2024, 12, 31)}, upload={'loandate': datetime.date(2024, 12, 31)}, add_to_picklist=None, column='Loan Date', missing_required=None), ParseResult(filter_on={'contents': 'a'}, upload={'contents': 'a'}, add_to_picklist=None, column='Loan - Contents', missing_required=None), ParseResult(filter_on={'remarks': '1'}, upload={'remarks': '1'}, add_to_picklist=None, column='Loan - Remarks', missing_required=None)], toOne={}, toMany={}, scopingAttrs={'discipline_id': 5, 'division_id': 2}, disambiguation=None, uploadingAgentId=948, auditor=Auditor(collection=<Collection: Collection object (131072)>, props=AuditorProps(allow_delete_dependents=True, batch_edit_prefs={'deferForMatch': True, 'deferForNullCheck': False}), audit_log=None, skip_create_permission_check=True, agent=<Agent: Agent object (948)>), cache={}, to_one_fields={'collectionobject': ['collectionobjectattribute'], 'collectingevent': ['collectingeventattribute'], 'attachment': ['attachmentimageattribute'], 'collectingtrip': ['collectingtripattribute'], 'preparation': ['preparationattribute']}, match_payload={'id': 17, 'ordernumber': None, 'version': 1}, strong_ignore=[]). \\n\\n Diff: {'contents': {'field_name': 'contents', 'old_value': None, 'new_value': 'a'}, 'remarks': {'field_name': 'remarks', 'old_value': None, 'new_value': '1'}, 'division_id': {'field_name': 'division_id', 'old_value': None, 'new_value': 2}}\")"
}
```
Not sure why I get the error in the first place, I edited the contents and remarks field and then pressed validate and it threw that error. 
Link to the data set: https://ojsmnh110242024migrations5417-issue-5413.test.specifysystems.org/specify/workbench/119 

In workbench after you upload if you right click on an uploaded cell it gives you a link to that record, in batch edit it says 'No upload results are available for this cell'. This isn't a big deal but we might want to either get the link working (like in workbench) or remove this feature from batch edit for now as this could be confusing. 

Batch edit: 
image

Workbench: 
image


Also I had a rollback fail again even though no changes were made. Here's a link to the data set: https://ojsmnh110242024migrations5417-issue-5413.test.specifysystems.org/specify/workbench/125 

image

specifyweb/stored_queries/format.py
sharadsw: Thanks! That was my initial thought as well but somehow didn't follow through with it. A general function in specifyweb/specify sounds like a great idea

specifyweb/workbench/upload/tests/testscoping.py
realVinayak: Sorry just saw this, didn't get a notification for some reason. Thanks for the output! At least we know what's wrong now.

> I remember I had to re-add collection to the function due to merge conflicts with a different PR

Yeah, looks like from https://github.com/specify/specify7/pull/5417/commits/be512b87e474317bf75cc7111de0087489e21bff, ha.

Passing collection is very shaky because it can be arbitrary based on collection relationships.  even more so after 4929 since it is no longer limited to base table (see https://github.com/specify/specify7/discussions/5842 which is fixed by 4929). I think a better way to handle this would be to just determine formatters from the row directly (in apply_scoping). We, after 4929, have access to row in that function, so in theory we can determine the formatter directly, rather than deferring till later (which causes the problem we see in the logs). Any chance I can just submit a patch by Monday? 

Definitely fine otherwise. I was thinking of reimplementing the changes from that PR but this time using row directly, rather than deferring and then using the row anyways. This will allow formatter to be completely determined at the time of scoping (so we don't have to pass collection in bind). The "determined" part comes from changes similar to what was done in https://github.com/specify/specify7/pull/4929/commits/2f32a9b5bc065952fbdb287d1e15b11c805aed97 (specifically note how formatter gets converted into a function in the scoping.py file in that commit).

sharadsw: > Any chance I can just submit a patch by Monday?

Go for it!

realVinayak: See https://github.com/specify/specify7/pull/6159. I ran unit tests locally and they pass (and also removed the zfill).

Literal patches, in case PR acts up.
[0001-workbench-Fix-formatters-in-CRs.patch](https://github.com/user-attachments/files/18550138/0001-workbench-Fix-formatters-in-CRs.patch)
[0002-workbench-add-unit-tests-for-COT-cnos.patch](https://github.com/user-attachments/files/18550139/0002-workbench-add-unit-tests-for-COT-cnos.patch)

CarolineDenis (approved):

specifyweb/workbench/upload/tests/testscoping.py
sharadsw: Thanks!

sharadsw:
@specify/ux-testing Thanks for testing! Summary of the changes:
- Batch Edit is now disabled instead of hidden for Trees. For CO queries involving Taxon, batch edit should be now enabled except when the query has a specific rank selected (it will still work when using `any rank`).
  - The missing rank error has duplicates because missing ranks are being calculated by checking ranks of all trees. Looks like this db has another tree with rank `Genus` with `Species` below it and so they get added to missing ranks. Will be fixed in #6127
- Fixed scoping error with Loan
- Created a separate issue for uploaded cell links: https://github.com/specify/specify7/issues/6164
- Cannot reproduce the rollback error locally. Will continue looking into it while the PR is tested

I updated the testing instructions to retest #5489 since this PR refactors code written in that PR

realVinayak:
Ah, for the rollback, I thought I wrote up a comment about it, looks like I forgot to publish it.

For elizabeth's recent rollback case, it failed because the row was a new row (that is also why it is colored blue). While the upload logic is generic, I forgot to account for this case (rollbacks in a batch-edit dataset for a new row). We currently shouldn't allow adding a new row in a batch-edit dataset. So, it is actually two bugs:
1. Frontend incorrectly allowing adding new rows (which are treated as workbench)
2. Backend crashing when a new row is found in the batch-edit during rollbacks.

So, it needs to be fixed on both sides for it:
1. Frontend
https://github.com/specify/specify7/blob/f5a319d4b8cadefd007e036d876c3718d65ed42f/specifyweb/frontend/js_src/lib/components/WorkBench/WbSpreadsheet.tsx#L131-L136 
Disable both of them when dataset.isupdate is true (so batch-edit cases). Elizabeth, did you use a different way to add new rows? This specifically disables "insert row above/below" when you right click.

2. Backend
https://github.com/specify/specify7/blob/f5a319d4b8cadefd007e036d876c3718d65ed42f/specifyweb/workbench/upload/upload.py#L561
Here, I'm pretty sure this place is returning None.
It should be something like:
```
pack = look_up_in_backer(row)
if pack is None: continue # Skip new rows.
r, be = pack
```

**Side note**

I also noticed another bug (related to UX). On commit success, it should be readonly, but it is not. Like once you upload in workbench, the entire sheet becomes readonly. Same thing should happen here after commit. Currently it is not, because HOT is a bit pesky. If you add readonly property at cell level, it always takes that (and overrides "global" properties from the top).

Replace this:
https://github.com/specify/specify7/blob/f5a319d4b8cadefd007e036d876c3718d65ed42f/specifyweb/frontend/js_src/lib/components/WorkBench/handsontable.ts#L46

By
```
const nullRecordsResults = dataset.uploadresult?.success === true ? {} : (identifyNullRecords?.(physicalRow, physicalColumn, property) ?? {});
```

specifyweb/workbench/upload/upload_table.py
```diff
@@ -970,10 +970,13 @@ def _do_upload(
...
             reference_record, attrs
         )
+
+        # Edge case: Scope change is allowed for Loan -> division.
+        # See: https://github.com/specify/specify7/pull/5417#issuecomment-2613245552
         if any(
             scoping_attr in concrete_field_changes
             for scoping_attr in self.scopingAttrs.keys()
-        ):
+        ) and not (self.name == "Loan" and "division_id" in concrete_field_changes):
```
realVinayak, line 979: nitpick:

Can you wrap this in a separate function that takes tablename and concrete_field_changes? Otherwise, it's quite distracting to see this (most of the logic in this file is generic). In the future, it would be easier to modify that function + we know what hacks we did

specifyweb/frontend/js_src/lib/components/WorkBench/WbSpreadsheet.tsx
```diff
@@ -129,14 +133,14 @@ function WbSpreadsheetComponent({
...
+                  disabled: () => isReadOnly || dataset.isupdate,
                 },
                 row_below: {
-                  disabled: () => isReadOnly,
+                  disabled: () => isReadOnly || dataset.isupdate,
                 },
                 remove_row: {
                   disabled: () => {
-                    if (isReadOnly) return true;
+                    if (isReadOnly || dataset.isupdate) return true;
```
realVinayak, line 143: Changes look good!

For this one, are you sure removing also needs to be disabled? It would be beneficial to allow removing rows, because it is possible that those rows might have some errors (validation ones) but you still want the other rows to be updated. In the current case, if we do not allow removal, then the user would have to make the batch-edit query again and reapply changes. 
+ we check for versions and do not allow updates if the version differs. So, it is definitely possible that the user simply wants to remove the row with the changed version.

specifyweb/stored_queries/batch_edit.py
```diff
@@ -0,0 +1,1184 @@
...
+    #     to_one = {
+    #         key: value.skim_plan(is_current_naive)
+    #         for (key, value) in self.to_one.items()
+    #     }
+    #     to_many = {
+    #         key: value.skim_plan(is_current_naive)
+    #         for (key, value) in self.to_many.items()
+    #     }
+    #     return self._replace(to_one=to_one, to_many=to_many, is_naive=is_current_naive)
+
```
realVinayak, line 473: Remove this commented out code? Sorry, I meant to that it earlier.

specifyweb/stored_queries/batch_edit.py
```diff
@@ -0,0 +1,1184 @@
...
+    indexed, query_fields = row_plan.index_plan()
+    # we don't really care about these fields, since we'have already done the numbering (and it won't break with
+    # more fields). We also don't caree about their sort, since their sort is guaranteed to be after ours
+    query_with_hidden = [
+        *query_fields,
+        *[field for field in fields if not field.display],
+    ]
+
+    with props["session_maker"]() as session:
+        rows = execute(
```
realVinayak, line 1057: I recently learned that this is not the most optimum way of fetching rows. We actually use a different way (even for exports) - which is to use yield_per
https://github.com/specify/specify7/blob/bad8e594f3395e4b656db00ccce03102c69a741c/specifyweb/stored_queries/execution.py#L221
This is because otherwise it fetches all at once, and dedicates memory for the entire set. 
See: https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.yield_per
Maybe refactor this usage to use yield_per?

specifyweb/workbench/upload/upload_table.py
```diff
@@ -473,118 +655,464 @@ def _do_upload(self, model, toOneResults: Dict[str, UploadResult], info: ReportI
...
+            for scoping_attr in self.scopingAttrs.keys()
+        )
+    
+    # Edge case: Scope change is allowed for Loan -> division.
+    # See: https://github.com/specify/specify7/pull/5417#issuecomment-2613245552
+    def _is_scope_change_allowed(self, concrete_field_changes):
+        if self.name == "Loan" and "division_id" in concrete_field_changes:
+            return True
+        
+        return False
```
realVinayak, line 1059: I really like these function names, this is perfect!

realVinayak (commented):
Nice so far!

Some more user issues that I remember:
1. Currently, there is no way to select specific rows and batch edit just them. This might be a requirement, maybe not. There is a work around to do so. The current implementation supports batch edit out of a record set. So, the user can:
 - Make a query.
 - Select the records, make record set out of it.
 - Query out of the record set, and batch edit on those results.

Quite a lot of steps. Check if that's acceptable for now, I've no clue if there have been any new expectations regarding BE. I think a decent implementation for this will be to follow the implementation in https://github.com/specify/specify7/pull/5300 (I actually left a similar comment on there).

2. Data sorting. Currently, if you have one table (so no relationships), sorting should **always** (of multiple columns) be respected (UX guys: test this). If you have relationships, and you also sort on some of the relationship's columns, it's essentially arbitrary (no guarantees regarding what user sees). From the code perspective, it is not arbitrary (see if you can figure those places out). I know that we don't support relationships, but this point actually applies regardless, since the same procedure is applied, just that we ignore relationships in upload plan. This might UX issue, a way to work around will be to sort in the workbench instead (in the spreadsheet).

3. Potential performance issues, as mentioned in a below comment in stored_queries/views.py. IIRC, I once tried batch editing andy's entire voucher collection (45k CO records with determination and cataloger) and it took 1.2 minutes to construct the dataset (I think that's too slow - I deferred profiling it for later, maybe need to that).

specifyweb/stored_queries/format.py
```diff
@@ -333,9 +333,23 @@ def _dateformat(self, specify_field, field):
...
+            return blank_nulls(_case) if self.replace_nulls else _case
+        
+        if self.format_picklist:
+            picklists, _ = get_picklists(self.collection, table.table, specify_field.name)
+            if picklists:
+                cases = [(field == item.value, item.title) for item in picklists[0].picklistitems.all()]
+                _case = case(cases)
+            
+                return blank_nulls(_case) if self.replace_nulls else _case
+
```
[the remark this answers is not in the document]
sharadsw: Thanks, made a note in #5713

sharadsw:
@realVinayak Thank you! I have made notes of your points in different issues

emenslin (changes requested):
- [x] Verify changed values are highlighted as updated cells
- [x] Re-run the original query and verify the values changed as expected
- [x] Verify values were rolled back to original
- [x] Retest #5489 

I did not have time to do a whole lot of testing but I did notice that on a locality data set the lat and long values show as updated even when you haven't edited them yet. I ran into this in the first batch edit pr but I don't know which comment I mentioned it in. This also happened with agent as the base table with some of the blank agent title records but I imagine it is the same issue

https://github.com/user-attachments/assets/4315a665-fe76-4edc-bea2-01735f74d747


Cannot open the interactions page

https://github.com/user-attachments/assets/bd95ff84-ee86-414b-b752-b3911a1b793a


Also the deafult form definitions are not showing up when you try to add a new one


https://github.com/user-attachments/assets/d13eacf6-6b62-47c2-88f6-7b994e09ab21


Production: 


https://github.com/user-attachments/assets/16ba8516-1cea-438d-b364-9fc4b17b906d

sharadsw:
NOTES:
- The lat/long error is related to floating point precision and using wrapper classes to primitives. When latitude and longitude are compared to their previous values the comparison that happens is: `Decimal('33.1493800000') == 33.14938` which cannot be true and is detected as a change. Will create a different issue for this since no data is lost or actually changed after upload. After upload `33.14938` gets set as `33.1493800000` anyway.
- The agent title error happens when it has data that doesn't belong to its picklist (the error must exist for other picklists as well). After [5b1474b](https://github.com/specify/specify7/pull/5417/commits/5b1474bd603aa1e7c3a882b406194badfb313a1f) if a cell has data that doesn't belong to its picklist, the cell is empty. This causes a change to be detected instead of catching a picklist validation error

sharadsw:
@emenslin Cannot reproduce the interactions dialog and app resource error locally on the same db and also other dbs. Can you try making a new copy of the original db for testing?

combs-a (commented):
- [X] Verify changed values are highlighted as updated cells
- [X] Verify values were rolled back to original

Didn't get to the regression test since I was testing other fields. 

I wasn't able to recreate this issue a second time for some reason, but I had a case in which batch edit edited a preparation count above the amount it had available.

i.e.: Used a prep record which had a count of 3 with only 2 available preps. A gift preparation record was created, and the quantity of this gift preparation was 2. Opening this in batch edit, I was able to set the quantity on the gift preparation to 5 and save with no warning.

Checking the previous records, the quantity on the gift prep was 5, and then the original preparation record now had a count of 5. Rolling back the batch edit *did not* fix the preparation record.

Here's the audit log for that, though the old and new values are not displayed I know for sure that it was 3 before that batch edit, and there shouldn't have been two edits; just one from adding gift preps.

image

emenslin:
Relating to the latlong issue, I know you said you already wrote up an issue for it but is there anyway it could be fixed in this pr or at least in the same release as batch edit? I know no values were actually changed but I feel like this could be very confusing and concerning to users. Although I used a simple example in the video this still happens if you were to make one change to the data set, the users would have no way of really knowing if they made a change or if it just looked like they did.

sharadsw:
@emenslin Yeah we can do it in the same release/PR. My guess is that the problem exists in other places as well (even for non-decimal values) and we need a more nuanced way to compare old and new values. Will add it to this PR if it's a small fix

realVinayak:
> we need a more nuanced way to compare old and new values.

Correct, but we already do (for API)! See 

https://github.com/specify/specify7/blob/1b2ab816c6b8229e66ebb8469f3d9bfea3003e7f/specifyweb/specify/api.py#L536-L544

Should just be a mechanical change to use it here. Also, https://github.com/specify/specify7/blob/1b2ab816c6b8229e66ebb8469f3d9bfea3003e7f/specifyweb/specify/api.py#L961 branch is irrelevant to us (trace out why)

EDIT: I just saw you pushed https://github.com/specify/specify7/pull/5417/commits/4814f1161de9989e71ab2025866de71a100dc487. Would still recommend change from api.py (to make things consistent)
/////////////////

There is actually a separate bug here in general. Previously, Elizabeth mentioned this error in 4929 review: https://github.com/specify/specify7/pull/4929#pullrequestreview-2361248477. That is, some localities were flagged as new. This happens because, when parsing longitude and latitude fields, we also artificially set long1text and fields. That is, you'd observe that if we do a workbench upload but just set latitude1 and longitude1 fields, long1text and latitude1text will also be set.

Relevant code is here:

https://github.com/specify/specify7/blob/1b2ab816c6b8229e66ebb8469f3d9bfea3003e7f/specifyweb/specify/parse.py#L201-L216

Generically, parsing 1 field can result in multiple new fields. This is why `ParseResult` (which contains result for just one field) has the  `upload` attribute as a dict. This is also why we have a nested loop when getting attributes that belong to a uploadTable 

Relevant code:
https://github.com/specify/specify7/blob/1b2ab816c6b8229e66ebb8469f3d9bfea3003e7f/specifyweb/workbench/upload/upload_table.py#L559-L563

Using the above, you might think "Oh, if we add extra lat1text and long1text, then doesn't this mean we would fail to match existing localities which have those fields set as null". That is, say `{"latitude1": 34.07}` is parsed as  `{"latitude1": 34.07, "lat1text": "34.07"}`, then during match, we would fail to match this to a locality with   `{"latitude1": 34.07, "lat1text": None}`. And, you'd be correct -- that's a bug. However, things weren't always this way. The above faulty behavior was introduced in https://github.com/specify/specify7/pull/4548. Before that, the extra lat1text (and long1text) were _just_ added to the filter part. Relevant ancient code:

https://github.com/specify/specify7/blob/4dbb95a6d0c204106b51e44f436cd10f539e50d7/specifyweb/workbench/upload/parsing.py#L284-L304

I'd encourage you, trace that code out from the v7.9.0 (and see how it is just added to the filter part) --- this is also the reason why we have separate `filer` and `upload` attribute on `ParseResult`.

All the above is precursor to the bug (in batch-edit):

When you have latitude1 and/or longitude1, the relevant text will also be edited -- but, the user would not know that it did (since we report that). I'm not sure about the best way to fix this. Can this be discussed internally and determined if this needs to be fixed at all? I'm neutral on it:
1. It changes data without user's permission (bad).
2. It fixes possible incorrect data (good). 

////////////////////////

Now, for the bug in https://github.com/specify/specify7/pull/4929#pullrequestreview-2361248477.

You can probably see what the source of Elizabeth's bug was. IIRC, it was definitely because lat1text and long1text was empty (but we were matching with artificially added values -- which were not null). Can this be tested again when batch-edit supports relationships? 

So, in conclusion, three bugs are related to locality.
1. Decimal and value equality check. Can be fixed with that code from api.py. 
2. Extra fields are edited, without the user knowing.
3. Extra fields are added when matching (partially caused from https://github.com/specify/specify7/pull/4548), which causes some localities to not be matched.

sharadsw:
@combs-a I'm not able to recreate it. I run into a validation error if I try to set the preparation quantity greater than the count or the availability:
Screenshot 2025-02-13 at 2 43 19 PM

What I did was:
- Create a CO with preparation count 5
- Create 2 Gift records
  - One with preparation qty 3
  - One with preparation qty 2
- Tried to update preparation qty > 5 in Batch edit

sharadsw:
@realVinayak Doesn't look like `fld_change_info` can be used here without refactoring it to do something similar as `is_equal` from https://github.com/specify/specify7/commit/4814f1161de9989e71ab2025866de71a100dc487. I'm happy to keep `is_equal` for now and note this in #6170

https://github.com/specify/specify7/blob/1b2ab816c6b8229e66ebb8469f3d9bfea3003e7f/specifyweb/specify/api.py#L536-L544
- The string comparison is unreliable when comparing objects with a primitive. Not sure how we can handle it while keeping it generic
  - We get a scope change error when comparing discipline/scope values since the comparison becomes `Discipline object (3) == 3 => false`
- For `Decimal` values, the old value remains a `Decimal` and the new value is a float and so it still gets detected as a change for lat/long
  - i.e `str(Decimal(33.1493800000)) == str(33.14938) => 33.1493800000 == 33.14938 => false`

Made a note of the lat/long text issues in #6126, thanks!

sharadsw:
@specify/testing The lat/long decimal issue should now be fixed. The lat/long text update + matching issues is added to #6126 and will be tested when we enable relationships.

realVinayak:
Ah yes, you're right -- it's more involved than I thought, sorry for wasting your time. I assumed that the previous code is bug-free, but it is not (I tested this on sp7demofish, and latitude are always considered to be changed -- that's another, unrelated to batch edit, bug). I agree, just keep what you have in that case.

Probably also need to document the lat/long text update (plus other fields that get updated, refer to the code for it). Maybe the discourse documentation is enough, or if there should be a warning when you use locality as the base table.

Just for reference (no changes needed):
> RE: We get a scope change error when comparing discipline/scope values since the comparison become

Can replace `field.name` by `field.attname` and that'll fix this. Attname will convert discipline to discpline_id (which will be 3),

combs-a (changes requested):
@sharadsw I've tried to recreate my issue again a few times, and wasn't able to as well--it should be kept in mind, but it might've just been something very specific.

- [X] Verify changed values are highlighted as updated cells
- [X] Re-run the original query and verify the values changed as expected
- [X] Verify values were rolled back to original

#### Regression tests
- [X] Retest #5489 

LatLon has been kind of fixed, but there are also potentially integers; ints are being converted to floats and that marks them as updated cells when they shouldn't be.

image

image

Also, I tried out COs that use COTs with the AlphaNumByYear type, and batch edit flagged the correct format as invalid, and incorrect formats as valid. Is fixing this within the scope of the PR? Wanted to check since the same thing works fine in the Workbench.

The incorrect format is AAAAAA, which was the default catno format before I changed it for FunnyTurtles.

image

combs-a (approved):
- [X] Verify changed values are highlighted as updated cells
- [X] Re-run the original query and verify the values changed as expected
- [X] Verify values were rolled back to original

#### Regression tests
- [X] Retest #5489 

<hr>

All tests passed, good work!

emenslin (approved):
- [x] Verify changed values are highlighted as updated cells
- [x] Re-run the original query and verify the values changed as expected
- [x] Verify values were rolled back to original
- [x] Retest #5489 

I didn't run into any new issues but since this is such a big PR I would like if we could get another testing review just to make sure nothing was missed.

lexiclevenger (approved):
- [X] Verify changed values are highlighted as updated cells
- [X] Re-run the original query and verify the values changed as expected
- [X] Verify values were rolled back to original

#### Regression tests
- [X] Retest #5489 

---
Looks good, nice job!

specifysoftware:

PR #5433, opened 2024-12-03
Display parentCOG on CO and COG form

### LINKED ISSUE #5428: Add COGs to the Collection Object form
Users should be able to easily see which Collection Object Group (COG) a Collection Object (CO) is a child of. 

From the CO form, users must be able to:
- Add a CO to a COG
- See which COG a CO is a child of
- Edit the COG a CO is a child of
- Remove a CO from a COG
- Carry forward the COG a CO is a member of when entering data in a series from the CO form

Each CO can only be a child of one COG, therefore it would be intuitive for the user to see a query combo box that allows them to change which COG the CO is a child of from the CO form (e.g. like adding or removing a CO from an Accession).

Seeing and modifying the COJO metadata (`isPrimary`, `isSubstrate`, `Text1`, `Text2`, etc.) should be possible, if even only from the COG form.

### LINKED ISSUE #5437: CO/COG children do not display the parent in the `parentCog` or `coParentCog` QCBX if not set using it
Currently, CO/COG children do not display the parent in the `parentCog` or `coParentCog` QCBX if not set using it. Since this is a direct relationship field, if not set using that QCBX, it will not make the link if it was added using the 'Children' subview automatically.

### LINKED ISSUE #5450: Adding COG to itself should not throw an error
**Describe the bug**
If you add a COG to itself using the children subview it throws an error and you lose any other changes that were made. 

**To Reproduce**

1. Go to collection object group 
2. Fill out name and type
3. Save
4. Use search/query to add COG to itself 
5. Try to save 
6. See error 

**Expected behavior**
There should be a save blocker or anything other than a full error that forces the user to leave the form. It is very easy to accidentally click the wrong COG and if a user was selecting multiple at once and accidentally selected the COG their on they would lose everything. 

**Screenshots**

https://github.com/user-attachments/assets/62af25a7-ff46-49c5-ae9b-b0f80138dc32

**Crash Report**

[Specify 7 Crash Report - 2024-12-05T21_52_41.376Z.txt](https://github.com/user-attachments/files/18030116/Specify.7.Crash.Report.-.2024-12-05T21_52_41.376Z.txt)

- OS: Windows 11
- Browser: Chrome
- Database Name: calvertmarinemuseum20241114
- Collection name: Invertebrates
- User Name: spadmin
- URL: https://calvertmarinemuseum20241114-production.test.specifysystems.org/specify/

CarolineDenis:
Fixes #5428
Fixes #5437
Fixes #5450 

This pr allow to visualize the parentCOG on CO and COG form + add save blocker when trying to add a cog to itself 

### Checklist

- [ ] Add automated tests

### Testing instructions

!To test this a few modifications in the xml are needed. See bellow. (the parentCog will be displayed as a COJO subview) 

!!! after testing, when copy pasting the testing instructions, please only keep the - [ ] verify... remove all the rest 

Create in CO 
- open CO form 
- save 
- create a new COG parent in the COJO subview
- save 
- [ ] verify you can see the new parentCOG 
- [ ] verify you can open the new COG with edit tool 
- [ ] verify you can see the CO in the COG when you open the record

Use existant in CO
- open CO form 
- save 
- search for a COG parent in the COJO subview
- save 
- [ ] verify you can see the parent COG 
- [ ] verify you can open the parent COG with edit tool 
- [ ] verify you can see the CO in the COG when you open the record

Create in COG 
- open COG form 
- save 
- create a new COG parent in the COJO subview
- save 
- [ ] verify you can see the new parentCOG 
- [ ] verify you can open the new COG with edit tool 
- [ ] verify you can see the CO in the COG when you open the record

Use existant in COG
- open COG form 
- save 
- search for a COG parent in the COJO subview
- save 
- [ ] verify you can see the parent COG 
- [ ] verify you can open the parent COG with edit tool 
- [ ] verify you can see the CO in the COG when you open the record

Delete parentCOG from the parent side for CO
- open CO form 
- save 
- search for a COG parent in the COJO subview or create new one
- save 
- open COG parent using the edit tool 
- delete child CO line 
- save parent cog form dialog 
- close 
- reload the page 
- [ ] verify the parent cog has been removed from the child form 
PS: it is not possible to delete a parent on the child form

Delete parentCOG from the parent side for COG
- open COG form 
- save 
- search for a COG parent in the COJO subview or create new one
- save 
- open COG parent using the edit tool 
- delete child CO line 
- save parent cog form dialog 
- close 
- reload the page 
- [ ] verify the parent cog has been removed from the child form 
PS: it is not possible to delete a parent on the child form

Save blocker for COG to itself using children
- open COG form 
- save 
- add a child cog using QB or search 
- search for the cog itself 
- add it as a child 
- [ ] verify you have a tool tip blocker saying that you cannot add a cog to itself 

Save blocker for COG to itself using parent
- open COG form 
- save 
- add a parent cog using QB or search in the cojo parent subview
- search for the cog itself 
- add it as a parent 
- [ ] verify you have a tool tip blocker saying that you cannot add a cog to itself 

Check for parent cog when adding child on parent 
- open COG form 
- save 
- add a child cog using QB or search or create
- add the child
- save the form 
- open the child 
- [ ] verify you can see the name of the parent cog 

To add in your CO and COG form xml: 

```
<row>
<cell type="subview" viewname="CollectionObjectGroupJoin_Sub" id="43" name="cojo" colspan="11" />
</row>
```

Create a new collectionobjectgroupjoin_sub view def 
```
<viewdef name="CollectionObjectGroupJoin_sub” class="edu.ku.brc.specify.datamodel.CollectionObjectGroupJoin" type="form" gettable="edu.ku.brc.af.ui.forms.DataGetterForObj" settable="edu.ku.brc.af.ui.forms.DataSetterForObj">
	<desc>The Collection Object Group Join Form</desc>
	<enableRules/>
	<columnDef>50px,2px,30px,5px,50px,2px,30px,5px,50px,2px,100px,0px,50px,2px,100px,p:g</columnDef>
	<rowDef auto="true" cell="p" sep="2px"/>
	<rows>
		<row>
			<cell type="panel" colspan="20">
				<rows>
					<row>
						<cell type="field" id="1" name="isPrimary" uitype="checkbox" default="false" readonly="true"/>
						<cell type="field" id="2" name="isSubstrate" uitype="checkbox" default="false" readonly="true"/>
						<cell type="field" id="5" name="parentCog" uitype="querycbx" colspan="3"/>
					</row>
				</rows>
			</cell>
		</row>
	</rows>
</viewdef>

```
Screenshot 2024-12-09 at 10 26 59 AM

Screenshot 2024-12-09 at 10 27 33 AM


acwhite211 (approved):

specifyweb/specify/models.py
```diff
@@ -1466,6 +1466,7 @@ class Collectionobject(models.Model):
     paleocontext = models.ForeignKey('PaleoContext', db_column='PaleoContextID', related_name='collectionobjects', null=True, on_delete=protect_with_blockers)
     visibilitysetby = models.ForeignKey('SpecifyUser', db_column='VisibilitySetByID', related_name='+', null=True, on_delete=protect_with_blockers)
     collectionobjecttype = models.ForeignKey('CollectionObjectType', db_column='CollectionObjectTypeID', related_name='collectionobjects', null=True, on_delete=models.SET_NULL)
+    coparentcog = models.ForeignKey('CollectionObjectGroup', db_column='CoParentCogID', related_name='+', null=True, on_delete=protect_with_blockers)
```
sharadsw, line 1469: (optional) Would it make more sense to name the field `parentCog` so the name is consistent across both COG and CO tables?

grantfitzsimmons:
Will https://github.com/specify/specify7/issues/5437 be fixed in this PR?

specifyweb/specify/datamodel.py
```diff
@@ -8298,6 +8299,7 @@
             Relationship(name='collection', type='many-to-one', required=False, relatedModelName='Collection', column='CollectionID'),
             Relationship(name='cogType', type='many-to-one', required=True, relatedModelName='CollectionObjectGroupType', column='COGTypeID'),
             Relationship(name='parentCog', type='many-to-one', required=False, relatedModelName='CollectionObjectGroup', column='ParentCogID'),
+            Relationship(name='coParentCog', type='many-to-one', required=False, relatedModelName='CollectionObject', column='CoParentCogID'),
```
sharadsw, line 8302: I don't think you need to add the field `coParentCog` to COG here as you already added it to CO and the field doesn't actually belong to COG

grantfitzsimmons (changes requested):
Issue is in https://github.com/specify/specify7/issues/5437, as CO/COG children do not display the parent in the `parentCog` or `coParentCog` QCBX if not set using that QCBX. This is not unique to this implementation as it already exists with `parentCog`, so let me know if this is out of scope.

This functions as you'd expect and makes the QCBX work properly otherwise! COs can be added to COGs, but a corresponding COJO is created automatically. 

image

If you create a new Consolidated COG, typically the first item entered would be "Primary". When using this approach, it is not set as primary automatically, enabling users to create a Consolidated COG without any primary COJO.

image

See I can't unset a subsequent addition to the COG:
https://github.com/user-attachments/assets/e1b437e5-0602-4fad-b90c-7ba184b1cf9f

Deleting a COJO created after setting this FK also doesn't remove the link, so it still preserves the COG FK in the QCBX while no corresponding COJO record exists:

https://github.com/user-attachments/assets/d247d420-f27e-4ac9-ae35-51e10c804812

specifyweb/businessrules/rules/cojo_rules.py
```diff
@@ -50,4 +58,15 @@ def cog_post_save(cog):
             childcog=cog,
             childco=None,
             parentcog=cog.parentcog
+        )
+
+@orm_signal_handler('post_save', 'Collectionobject')
+def cog_post_save(co):
```
sharadsw, line 64: Can you rename this function as we already have a `cog_post_save` function in this file? It shouldn't make any difference functionally but it would be less confusing

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -202,6 +202,20 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
...
+            reason: resourcesText.parentCogSameAsChild(),
+            saveBlockerKey: PARENTCOG_KEY,
+          };
+        }
+
+        return {
+          isValid: true,
+          saveBlockerKey: PARENTCOG_KEY,
+        };
+      },
```
sharadsw, line 218: Do we need this business rule at all? The `parentCog` business rule in COG helps avoid linking a COG to itself - but that will not be a problem in a CO

specifyweb/frontend/js_src/lib/components/DataModel/types.ts
```diff
@@ -6512,8 +6511,9 @@ export type CollectionObjectGroup = {
     readonly yesno2: boolean | null;
     readonly yesno3: boolean | null;
   };
-  readonly toOneDependent: RR<never, never>;
+  readonly toOneDependent: { readonly cojo: CollectionObjectGroupJoin | null };
   readonly toOneIndependent: {
+    readonly coParentCog: CollectionObject | null;
```
sharadsw, line 6516: COG -> coParentCog is not needed (see other comment)

sharadsw (changes requested):

specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
```diff
@@ -241,6 +241,23 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
...
+        CollectionObjectGroupJoin.get('childCog') ===
+          CollectionObjectGroupJoin.get('parentCog') &&
+        typeof CollectionObjectGroupJoin.get('childCog') === 'string' &&
+        typeof CollectionObjectGroupJoin.get('parentCog') === 'string'
+      ) {
+        setSaveBlockers(
+          CollectionObjectGroupJoin,
+          CollectionObjectGroupJoin.specifyTable.field.childCog,
+          ['Cog added to itself'],
+          'Cog to COG'
```
sharadsw, line 257: I think the blocker message can be added to a localization file and the blocker key can be added to `businessRulesUtils.ts` with other keys:
https://github.com/specify/specify7/blob/6b071d7ce274d5f85d17974e4c8e368b96727f3f/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleUtils.ts#L5-L7

sharadsw (approved):
Looks good!

emenslin (changes requested):
### Testing instructions

- open CO form 
- save 
- create a new COG 
- save 
- [x] verify you can see the new COG 
- [x] verify you can open the new COG with edit tool 
- [x] verify you can see the CO in the COG when you open the record

---

If you use the QCBX to add a parent COG to the CO, save, then remove the parent COG on the CO form, then go to the parent COG the CO isn't removed. Co child should be removed from the COG if the COG is removed from the CO. This also applies to adding a parent COG to a COG.

https://github.com/user-attachments/assets/18c9e4fe-0257-41a5-9105-2818b5cfc0df




https://github.com/user-attachments/assets/2f6892d8-323f-48fd-b61f-c8c557e51ed6



Also, #5450 was not fixed in this pr 

https://github.com/user-attachments/assets/b2beab6b-4080-4ca6-b5ec-084002d562da


[Specify 7 Crash Report - 2024-12-06T15_22_52.757Z.txt](https://github.com/user-attachments/files/18040156/Specify.7.Crash.Report.-.2024-12-06T15_22_52.757Z.txt)

melton-jason (commented):
Isn't this already theoretically possible in Specify? 😅 

Take a look at the following form definition and workflow, entirely on `production`!: 

https://github.com/user-attachments/assets/6f0ff0e7-ca1d-4a33-a631-ddc0c048a717

This would meet all of the requirements of #5428 except for maybe "Carry forward the COG a CO is a member of when entering data in a series from the CO form" (I think the COJO -> childCog would still point to the previous CO?) 

Of course there is a major problem here (the error in the end), but it's a separate bug!  
As seen in the Network request, Specify isn't setting the `CO -> cojo` relationship on the COJO. Which is probably due to being a `one-to-one` relationship; something we should fix anyways!

Adding a new relationship to CollectionObejct -> COG is _very_ tricky to get right. If I'm understanding correctly, it's fundamentally supposed to represent the exact same thing as `CO -> cojo -> parentCog`. 
Ensuring the two are always in sync is very complicated, and never completely enforceable (and querying, making form definitions, and workbench uploads can be very confusing for users). 

This is not saying it's a bad idea in it's totality, just that things will get very messy if we have two entirely different relationships which represent the same thing (same logic applies for `COG -> parentCog` with `COG -> cojo -> parentCog`) 

Things would be a lot cleaner (and probably easier to implement in the long run!) if we only have one relationship at the database level. 

I'd imagine there would be three alternative approaches here: 

- Remove `COG -> parentCog` and `CO -> parentCog` and instead make additional/change default view definitions and fix that `one-to-one` bug!
- Keep `COG -> parentCog` and `CO -> parentCog` in the database (adding their reverse relationships), but remove `COJO -> childCog`, and `COJO -> childCO`
  - COJO would then become something more of a `COG Attributes` table, and we could rename COJO and `COJO -> parentCOG` accordingly
- Remove `COG -> parentCog` and `CO -> parentCog` from the database and instead make them virtual fields from `COG -> cojo -> parentCog` and `CO -> cojo -> parentCog`

grantfitzsimmons (changes requested):
When I add a COJO record to the form via the subview and add a new CO to an exisiting COG, I get the following error:

```
BusinessRuleException at /api/specify/collectionobject/ Both childcog and childco cannot be null. 
```

[Specify 7 Crash Report - 2024-12-09T20_26_04.566Z.txt](https://github.com/user-attachments/files/18067700/Specify.7.Crash.Report.-.2024-12-09T20_26_04.566Z.txt)


https://github.com/user-attachments/assets/44df53fd-24db-44d7-aa58-c0e97c815bdc


I am also unable to remove the COJO record once it is added, forcing me to add a parent COG despite accidentally clicking on the plus. I am forced to create a new COG and delete it just to save the record.  This definitely has the capacity to be a problem for users.

image

grantfitzsimmons (changes requested):
- [X] verify you can see the new parentCOG 
- [X] verify you can open the new COG with edit tool 
- [X] verify you can see the CO in the COG when you open the record

There are cases where the user must refresh the page, but the CO is in the COG upon opening it in another tab or refreshing the page.


- [X] verify you can see the parent COG 
- [X] verify you can open the parent COG with edit tool 
- [X] verify you can see the CO in the COG when you open the record

Most of the time, but see below.


- [X] verify you can see the new parentCOG 
- [X] verify you can open the new COG with edit tool 
- [X] verify you can see the CO in the COG when you open the record

Most of the time, but see below.


- [X] verify you can see the parent COG 
- [X] verify you can open the parent COG with edit tool 
- [X] verify you can see the CO in the COG when you open the record


- [X] verify the parent cog has been removed from the child form 

I think there is a usability concern here with the requirement for users to reload the page each time they remove a COG. I spoke with @CarolineDenis about this, and we understand that this is the only place in the application that functions this way. Other "join" tables are similarly affected, but this one is likely to be noticed by users due to the fundamental workflow likely brushing up against it. 

https://github.com/user-attachments/assets/04fd5c35-118b-4b8d-9654-7b79a0c7cbef

This is not to say the test as written is failing (as it is passing per the written instructions), but it is to say we should try to resolve this here or open a new issue.

>PS: it is not possible to delete a parent on the child form

This leads to an error that will be seen to the user if they attempt this deletion:

```
IntegrityError at /api/specify/collectionobjectgroup/7/ (1451, 'Cannot delete or update a parent row: a foreign key constraint fails (`geo`.`collectionobject`, CONSTRAINT `collectionobject_CoParentCogID_952ffbf8_fk_collectio` FOREIGN KEY (`CoParentCogID`) REFERENCES `collectionobjectgroup` (`collectionobjectgroupid`))') 
```

Is this something that can be handled by save blockers? If this is outside the scope of this PR, we need to open another issue for this. I don't believe this is outside of a usual workflow if one were to extensively use COGs.


- [X] verify the parent cog has been removed from the child form 



- [X] verify you have a tool tip blocker saying that you cannot add a cog to itself 

Very happy about this change! No errors 🎉 


- [X] verify you have a tool tip blocker saying that you cannot add a cog to itself 

🎉 


- [X] verify you can see the name of the parent cog 

---

I tried to use carry forward with this PR as I feel strongly that users who utilize the COG functionality will want to enter constituent parts in a series.

Carry forward configuration:
image

After clicking "Carry Forward":

https://github.com/user-attachments/assets/9e3291b3-4b76-4266-91cc-af150c24a5ec

```
Error: unhandled relationship type
    at eval (webpack-internal:///./lib/components/DataModel/resourceApi.ts:244:19)
    at Array.map (<anonymous>)
```

[Specify 7 Crash Report - 2024-12-11T15_34_31.612Z.txt](https://github.com/user-attachments/files/18098478/Specify.7.Crash.Report.-.2024-12-11T15_34_31.612Z.txt)

Unchecking "COG Join" from this menu prevents the error.

If this is outside the scope of this PR, we must open another issue for this.

---

If you add a CO to a COG, it appears in the list of children _most of the time_. If you change the COG parent of that CO, it does not appear in the list of children when you open the record in a dialog until you refresh the page.

https://github.com/user-attachments/assets/46d60920-fe11-46fe-bbdd-7379a1b898aa

---

You can cause an issue quickly if you "unset" the COG parent (delete the parent COG from the QCBX– natural workflow for Specify users who are "removing" a link like Accession, Cataloger, etc). See that after you remove it, the delete button for the subview appears. Once clicked, you cannot save since the "Parent Cog" field is required, yet invisible.

https://github.com/user-attachments/assets/a0e9687d-bcf9-449c-ae6f-c37c30888716

You might hope to add it to a new COG, but unfortunately, that does not resolve the issue:

https://github.com/user-attachments/assets/fc5dafa3-e80a-4e9f-84c2-f4e034c79485

grantfitzsimmons (approved):
See #5463, #5462, #5461, #5460

PR #5442, opened 2024-12-04
Add cojo uniqueness rules

### LINKED ISSUE #5441: Save needs to be blocked when a COG child is already a child to another COG
Currently, when a user adds a child to a COG that is already a part of another COG, an error gets thrown and we see a crash report. The app needs to have a save blocker instead of the crash.

This can be accomplished by adding uniqueness rules to the `COJO -> childCog` and `COJO -> childCo` fields. It seems the uniqueness rules were already intended to be added by changing the default rules file:
https://github.com/specify/specify7/blob/6b071d7ce274d5f85d17974e4c8e368b96727f3f/specifyweb/businessrules/uniqueness_rules.json#L70-L75

However, the migration that is responsible for adding these rules `0002_default_unique_rules` was never re-run and so these rules never got added to most databases.

sharadsw:
Fixes #5441

### Checklist

- [ ] Add automated tests

### Testing instructions

- Create a COG
- Add a CO child that already is a child to another COG
- [ ] Verify save is blocked
- Add a COG child that already is a child to another COG
- [ ] Verify save is blocked

### DIFF
--- a/specifyweb/businessrules/migrations/0005_cojo.py
+++ b/specifyweb/businessrules/migrations/0005_cojo.py
@@ -1,0 +1,34 @@
+from django.db import migrations
+
+from specifyweb.businessrules.uniqueness_rules import create_uniqueness_rule, DEFAULT_UNIQUENESS_RULES
+
+"""
+    Applies the COJO uniqueness rule to the database.
+"""
+def apply_migration(apps, schema_editor):
+    cojo_rules = DEFAULT_UNIQUENESS_RULES["CollectionObjectGroupJoin"]
+
+    for rule in cojo_rules:
+        fields, scopes = rule["rule"]
+        isDatabaseConstraint = rule["isDatabaseConstraint"]
+        create_uniqueness_rule(
+            'Collectionobjectgroupjoin', None, isDatabaseConstraint, fields, scopes, apps
+        )
+
+"""
+    Revert is skipped as this rule should ideally have been applied through the 0002_default_unique_rules migration.
+    The rule will be reversed when 0002_default_unique_rules is reversed.
+"""
+def revert_migration(apps, schema_editor):
+    pass
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('specify', '0002_geo'),
+        ('businessrules', '0004_catnum_uniquerule')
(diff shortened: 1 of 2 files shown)


alesan99 (approved):
### Testing instructions



- Create a COG
- Add a CO child that already is a child to another COG
- [x] Verify save is blocked
- Add a COG child that already is a child to another COG
- [x] Verify save is blocked
---

No crashes! 👍👍

chrome_h51Asdr4EI


image

emenslin (changes requested):
### Testing instructions



- Create a COG
- Add a CO child that already is a child to another COG
- [x] Verify save is blocked
- Add a COG child that already is a child to another COG
- [x] Verify save is blocked

Looks good, save is now blocked, however I ran into an issue when trying to save a COG. It looks like the COGType picklist is using the "super types" picklist instead of the COGTypes picklist and I get an error when I try to save. 

Issue branch: 

https://github.com/user-attachments/assets/81671144-f9e3-4aaa-aafc-7a97159159fc


[Specify 7 Crash Report - 2024-12-05T15_59_36.588Z.txt](https://github.com/user-attachments/files/18026079/Specify.7.Crash.Report.-.2024-12-05T15_59_36.588Z.txt)


Production: 


https://github.com/user-attachments/assets/ceda69ce-10c5-444c-9fd9-a015a25bd9e6

sharadsw:
@emenslin Thanks! I'm not able to login to the issue branch of this db because it's stuck on 'Starting'. I think it might be a db/migration error since the PR doesn't touch any code related to the cogtype picklist. Can you try on a different db?

emenslin:
If you have a child cog and child co that are both blocking the save, should the save blocker only list one of the items blocking the save or should it list both? Right now it only says that one is blocking it even if there are multiple items blocking it so I wanted to check if this was the intended behavior first. 

https://github.com/user-attachments/assets/84228ac7-4d4d-433c-9ab4-4f0bcbf743dd

sharadsw:
@emenslin I think it's fine since general save blockers (as opposed to blockers on a specific field) can only display one blocker at a time

emenslin (approved):

Issue #5446, opened 2024-12-05
Default form for TectonicUnit uses Taxon QCB


alesan99:
**Describe the bug**
The default form for TectonicUnit data entry lets me select Taxon as the accepted Tectonic Unit

**To Reproduce**

1. Use the most up-to-date default geo forms
2. Go to data entry for tectonicUnit
3. Try to enter an Accepted Tectonic Unit into the QCB

**Expected behavior**
Should use a TectonicUnit QCB

**Screenshots**

https://github.com/user-attachments/assets/e4fbb97c-029e-4b67-ad53-7bfd025786fd

https://github.com/user-attachments/assets/2460ba77-894d-440b-a8f4-becefc24bcab

Please fill out the following information manually:
- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: 6b071d7ce274d5f85d17974e4c8e368b96727f3f
- Database Name: chadrontest_2024_08_15 (local copy)
- Collection name: Paleo
- User Name: testuser

CarolineDenis:
@grantfitzsimmons , 
Since we will change xml files in this pr could we add the changes for the parentCog as well? (On CO & COG forms)

pashiav:
Found duplicate options when searching in the text field of accepted Tectonic Unit.


<img width="1270" alt="Screenshot 2024-12-09 at 9 34 13 AM" src="https://github.com/user-attachments/assets/a54cdfae-1b6b-47ff-a7b6-251ee1dacb3e">
<img width="1152" alt="Screenshot 2024-12-09 at 9 34 26 AM" src="https://github.com/user-attachments/assets/b14d438b-e709-413c-898e-faa065e7eb69">

grantfitzsimmons:
1. It's crazy that Specify didn't handle this case and looked at the Taxon tree 😂 
2. @pashiav Can you try this again on this PR?

https://github.com/specify/specify7/pull/5448

pashiav:
Duplicates are fixed in #5448!

PR #5448, opened 2024-12-05
Update default forms

### LINKED ISSUE #5447: COG can be set as 'primary' or 'substrate' from the grid view
**Describe the bug**
COGs should not be able to be the primary or substrate object in a COG, but the subview allows you to set this when collapsed. 

This can be fixed in the XML if a code solution cannot dynamically make the field read-only based on the presence or absence of a `childCO` or `childCOG`.

grantfitzsimmons:
Fixes #5447 and #5446

There have been a number of changes since the first round of default XML changes. To respond to those, this new PR has been opened. This resolves the following:

## Changes

* Uses the `cojo` subview rather than the `parentCog` mechanism to link CO -> COG and COG -> COG
* This makes `isPrimary` and `isSubstrate` read-only from the table view
* Adds `TectonicUnit` to the `TypeSearches` so that they can be found when using the QCBX for `acceptedParent` and `parent`
* Corrects an issue with `TectonicUnit` looking at `Taxon` records in the `acceptedParent` QCBX

## Testing Instructions

You will need to make sure you are using the Specify 7 static files and not the 6 before proceeding. Make sure you delete any conflicting `TypeSearches` app resources or other form definitions.

- [ ] Verify that the `cojo` subview appears instead of the `parentCog` field in the default COG form and default CO form in the 'Geology' discipline (add a new 'Geology' form definition in app resources to confirm)
- [ ] Verify that you cannot set `isPrimary` or `isSubstrate` to true if the COG child is a COG
- [ ] Verify that you can search for Tectonic Units by name on the Tectonic Unit node forms


grantfitzsimmons:
Pause on testing– I am troubleshooting an issue 👍

alesan99 (approved):
## Testing Instructions

- [x] Verify that the `cojo` subview appears instead of the `parentCog` field in the default COG form and default CO form in the 'Geology' discipline (add a new 'Geology' form definition in app resources to confirm)
- [x] Verify that you cannot set `isPrimary` or `isSubstrate` to true if the COG child is a COG
- [x] Verify that you can search for Tectonic Units by name on the Tectonic Unit node forms

image

Working as expected! 👍👍👍👍

sharadsw (approved):

PR #5455, opened 2024-12-09
Allow for the override of SPECIFY_THICK_CLIENT from environment

acwhite211:
Allow for spcloud and test-panel to configure the Specify Django settings parameter SPECIFY_THICK_CLIENT inorder to allow for a simpler spcloud docker-compose file.  This will still allow self hosted instance to remain unchanged, so no change for self-hosted instances is required.

See here https://github.com/specify/specify7-test-panel/compare/main...sp7-only for use of this parameter.


config/test_sp7.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- This file is to test that the sp7_only config files are beind used in the static files api -->
```
grantfitzsimmons, line 2: We can remove this before merging, right?

grantfitzsimmons (commented):
https://github.com/specify/specify7-test-panel/commit/9d8924a727f0222527a812a793b06329da57c681 looks like it will solve the issue we're having. Need to find the time to test this on the panel.

CarolineDenis (approved):

sharadsw (approved):

specifyweb/settings/specify_settings.py
```diff
@@ -1,8 +1,9 @@
+import os
 
 # Specify 7 requires the files from a Specify 6 install.
 # This setting should point to a directory containing an installation
 # of Specify 6 of the same version as the Specify database.
-THICK_CLIENT_LOCATION = '/opt/Specify'
+THICK_CLIENT_LOCATION = os.environ.get('THICK_CLIENT_LOCATION', '/opt/Specify')
```
melton-jason, line 6: This allows local installations to set a THICK_CLIENT_LOCATION environment variable and use that instead of modifying this variable directly. 

Do we want to add note of this to the installation instructions? 
https://github.com/specify/specify7/blob/18f5c15a8e54d945b2c03a7877d453c8e1ea5669/README.md?plain=1#L491-L493

melton-jason: And for Dockerized instances, do we want to provide documentation for how this environment variable can be used?
Or maybe just how it's used with correlation to the `docker-compose.yml` (and `Dockerfile`)? 

For the users who wish to go in depth for customizing the setup of their instance, there's a LOT of moving parts. They might not even know what's possible/not-possible, and I could see how easy it might be to get stuck.

melton-jason (approved):

specifyweb/settings/specify_settings.py
grantfitzsimmons: 1. I don't think we should encourage anyone to configure new local installations
2. Documentation on supported environment variables are always desirable 😄

acwhite211:
Not really needed anymore with updates to spcloud and the test-panel.  We could still add this sometime if we find it easier.

PR #5464, opened 2024-12-11
Fix the worker in development

grantfitzsimmons:
When the variable is specified, I cannot connect to the worker. Requests go unanswered, and I have been manually specifying the database name. However, when I remove the `-Q` option that tells Celery which queue to listen to, I no longer encounter this problem.

> Is no one else having this issue?

### Testing instructions

After applying this change:
- [ ] Create a WorkBench data set
- [ ] Validate it
- [ ] Upload it
- [ ] Roll it back


grantfitzsimmons:
New lead– might it be related to running mariadb locally instead of in Docker?

PR #5431, opened 2024-12-03
Allow catalogNumber format to be determined by CollectionObjectType

melton-jason:
Fixes #5232, #5233 
See #5234

⚠️ **Note:** This PR affects database migrations. See [migration testing instructions](https://specify.github.io/testing/pull_requests.html#prs-tagged-with-label-migration).

Aside from implementing #5232 and #5233, there are some other improvements in this PR!

**Schema Config**

- When selecting a Field Format in the Schema Configuration tool, Specify now groups field formatters for the field they selected (Field Formats that explicitly reference the field in `fieldName=fieldname`) above the rest

Screenshot 2024-12-03 at 3 08 19 PM

**Migrations**
There were some bugs in other migrations which used the functions defined in [specifyweb/specify/update_schema_config.py](https://github.com/specify/specify7/blob/production/specifyweb/specify/update_schema_config.py). These bugs have been fixed in this PR, and a migration has been written to fix these bugs for databases already affected: [specify/0017_schemaconfig_fixes.py](https://github.com/specify/specify7/blob/issue-5232/specifyweb/specify/migrations/0017_schemaconfig_fixes.py)

**UIFormatter picklists for Collection -> catalogNumFormatName and CollectionObjectType -> catalogNumberFormatName**
- Everywhere in the application*, these fields will now appear as a picklist whose items are the names of `CO -> catalogNumber` Field Formatters

Screenshot 2024-12-03 at 3 21 02 PM

*on the DataEntry Form, they will only appear as a PickList if `uitype="combobox"` is used on the form definition

### Checklist

- [ ] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [ ] Add automated tests
- [ ] Add relevant issue to release milestone

### Testing instructions

- Define a Field Format for `Collection Object -> catalogNumber` which is different than the current `Collection -> Catalog Num Format Name`
  - An example field format called `Rocks` is provided below:  
```xml
<format system="false" name="Rocks" class="edu.ku.brc.specify.datamodel.CollectionObject" fieldname="catalogNumber" default="false">
	<autonumber>edu.ku.brc.af.core.db.AutoNumberGeneric</autonumber>
	<field type="constant" size="4" value="ROCK"/>
	<field type="separator" size="1" value="-"/>
	<field type="numeric" size="6" inc="true"/>
</format>
```
- Go to the Data Entry form to create a new Collection Object Type
[...]

Issue #5232, opened 2024-08-16
Implement Catalog Numbering Configurations Based on COT


grantfitzsimmons:
**Description:**  
We need to develop a system that allows for multiple catalog number formats based on either the collection or COT. This includes options for auto-numbering based on type and shared formats across multiple COTs.

**Acceptance Criteria:**
-  Users can set a single catalog number format for all COs within a collection (`CatalogFormatNumName` in the `collection` table, this is already supported)
-  Users can define multiple formats for COs within a specific COT (done in the XML app resource `UIFieldFormatters`)
-  Implement auto-numbering based on the selected type for each COT.
-  Allow multiple COTs to share the same catalog number format (they can use the same `CatalogFormatNumName` across multiple COTs, and auto-numbering should be shared)
-  Update the `CollectionObjectType` table to include a `CatalogFormatNumName`.

**Default numbering behavior:**
We need to also establish a default logic for assigning catalog numbers. If a COT does not have a specific series set, it should default to the collection’s catalog number format.

**Acceptance Criteria:**
-  If no COT-specific format is set, the system defaults to the collection’s format.
-  Ensure that the default settings are clearly documented and accessible to users.

grantfitzsimmons:
Related to https://github.com/specify/specify7/issues/5233

grantfitzsimmons:
In the query builder, users should be able to select from multiple COT names to determine the field format used in the query.

PR #4615, opened 2024-03-08
Add support for explicitly choose a formatter in QB

CarolineDenis:
Fixes #1959

Notes: 
Code was lost during a merge. Feature added in https://github.com/specify/specify7/commit/d52aee94a01d126b2ef83f0dea1f690bb59f6ee1

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good
      and self-explanatory (or properly documented)
- [x] Add relevant issue to release milestone

### Testing instructions

- Navigate to the Query Builder (QB).
- Select a table that you know has one or several formatters (To check, go to app resources, navigate to Record Formatter, choose the table, and verify the list of formatter names available).
- Choose "formatted."

- [ ] Verify that in the query line, after selecting the "formatted" field box, a drop-down appears with a list of formatter choices (corresponding to the list in app resources).
- [ ] Verify that changing the formatter changes the way the field is displayed in the results table.

PR #5485, opened 2024-12-18
Support COT catalog number configurations in the Query Builder

### LINKED ISSUE #5474: Support COT-based Catalog Numbering Configurations in the Query Builder
**Description:**  
https://github.com/specify/specify7/issues/5232 covered the new catalog numbering configurations based on the `CollectionObjectType` of the CO being created, but it did not explicitly mention how this should behave in the query builder.

**Acceptance Criteria:**
- When a user queries on 'Collection Object → Catalog Number' from any base table:
   - If there is **more than one** `CollectionObjectType` in the current collection **with** different `CatalogFormatNumName` values, then a button with a gear should appear to the right (similar to the gear and mechanism to choose a custom table format/aggregation).
      - Upon clicking the gear, the user should see a list of `CollectionObjectType` `Name` values, along with a list item that says "None"
      - The user should be able to pick from any of the COTs in the current collection
      - The field should now enforce the field format present in the selected COT's `CatalogFormatNumName` field
      - If the user selects "None" instead of a COT Name, no format should be enforced
- Other fields should not have this capability at this time.

melton-jason:
Fixes #5474
Related to #5232

Allows the selection of Collection Object Type (COT) Catalog Number formats in the Query Builder!: 

Screenshot 2024-12-17 at 11 34 05 PM

Known edge case: https://github.com/specify/specify7/issues/5487 and https://github.com/specify/specify7/issues/5488

### Checklist

- [ ] Add automated tests

### Testing instructions
- Create a CollectionObjectType with a Catalog Number Format Name which differs from that of the Collection's default Catalog Number format
  - You can run a query on the Collection table with the `Catalog Num Format Name` field to determine the  default format
-  Create a new Collection Object query, add Catalog Number to the query, and switch the type from `Any` to any of: 
   - Equal 
   - Greater than
   - Less than
   - Greater or Equal to
   - Less or Equal to
   - Between
   - In
   - (all other types-like Contains or Starts With- have validation disabled by default)
- [ ] Ensure that by default, the Collection's default Catalog Number format is enforced 
- Switch the Catalog Number format to the previously created Collection Object Type using the new gear icon and picklist 
- [ ] Ensure the input of the field is now validating based on the format defined on the chosen Collection Object Type
- Add one or more OR filters to the Catalog Number field (this can be done through the `+` button) with a type from the list above 
- [ ] Ensure each filter on the Catalog Number field with validation can independently have its own format
- [ ] Save the Query and refresh the page: the Query should load successfully
- [ ] Ensure the query returns correct results when using filters with multiple Catalog Number formats
  - You may need to create/modify Collection Object records with the desired Collection Object Type 
  - For reference see https://github.com/specify/specify7/pull/5431#issue-2716025555 
 
- If there is only one CollectionObjectType or all CollectionObjectTypes either don't specify a Catalog Number format or use the Collection's default, then the gear icon to select a formatter should not appear on Catalog Number
- `Collection Object -> Catalog Number` along with formatted/aggregated fields are the **_only_** filters which should have the gear icon to select a format
- [ ] Briefly generally test other functionality of the Query Builder, such as: running old queries, selecting a different record formatter (https://github.com/specify/specify7/pull/4615), Query From Tree Node, etc.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/reducer.ts
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/reducer.ts
@@ -156,11 +156,6 @@ export const reducer = generateReducer<MainState, Actions>({
       fields: replaceItem(state.fields, line, {
         ...state.fields[line],
         mappingPath: newMappingPath,
-        dataObjFormatter:
-          mappingPathIsComplete(newMappingPath) &&
-          action.currentTableName === action.newTableName
-            ? undefined
-            : state.fields[line].dataObjFormatter,
       }),
       autoMapperSuggestions: undefined,
     };
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Formatter.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Formatter.tsx
@@ -115,9 +115,13 @@ export function CatalogNumberFormatSelection({
       availableFormatters={availableFormatters}
       currentFormat={formatter}
       showSingular={
-        !Object.values(schema.collectionObjectTypeCatalogNumberFormats).some(
-          (format) => format === null || format === schema.catalogNumFormatName
-        )
+        Object.values(schema.collectionObjectTypeCatalogNumberFormats).some(
+          (format) => format !== null && format !== schema.catalogNumFormatName
+        ) ||
+        (formatter !== undefined &&
+          !Object.values(schema.collectionObjectTypeCatalogNumberFormats)
+            .filter((formatName) => formatName !== null)
+            .includes(formatter))
       }
       onChange={handleChange}
     />
@@ -178,6 +182,14 @@ function FormatSelect({
                 {`${title} ${isDefault ? resourcesText.defaultInline() : ''}`}
               </option>
             ))}
+            {currentFormat !== undefined &&
+            !availableFormatters
+              .map(({ name }) => name)
+              .includes(currentFormat) ? (
+              <option key="invalidCOT" value={currentFormat}>
+                {queryText.invalidPicklistValue({ value: currentFormat })}
+              </option>
+            ) : undefined}
           </Select>
         </div>
       )}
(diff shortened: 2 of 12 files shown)


specifyweb/frontend/js_src/lib/components/QueryBuilder/reducer.ts
```diff
@@ -156,11 +156,6 @@ export const reducer = generateReducer<MainState, Actions>({
       fields: replaceItem(state.fields, line, {
         ...state.fields[line],
         mappingPath: newMappingPath,
-        dataObjFormatter:
-          mappingPathIsComplete(newMappingPath) &&
-          action.currentTableName === action.newTableName
-            ? undefined
-            : state.fields[line].dataObjFormatter,
```
melton-jason, line 163: Take a close look at this change @specify/dev-testing!
In my local testing, everything seems to work okay with this change, but I've yet to go deeper through the code and fully assess the consequences of the change.

alesan99 (changes requested):
### Testing instructions
- [x] Ensure that by default, the Collection's default Catalog Number format is enforced 
- [c] Ensure the input of the field is now validating based on the format defined on the chosen Collection Object Type
- [x] Ensure each filter on the Catalog Number field with validation can independently have its own format
- [c] Save the Query and refresh the page: the Query should load successfully
- [x] Ensure the query returns correct results when using filters with multiple Catalog Number formats
- [x] If there is only one CollectionObjectType or all CollectionObjectTypes either don't specify a Catalog Number format or use the 
- [x] Briefly generally test other functionality of the Query Builder, such as: running old queries, selecting a different record formatter (https://github.com/specify/specify7/pull/4615), Query From Tree Node, etc.

---
The functionality looks good! 👍👍
I just have a minor note

If multiple COTs share the same formatter the selection dropdown always chooses the first one with that format. It's a little weird but the functionality works since it is the same formatter after all.
I imagine in practice most people will have a different formatter for each COT if they aren't sticking with the default, but if someone comes across this it may appear broken.

https://github.com/user-attachments/assets/b8ed934c-a1c7-4952-a316-9ddda6bb2a1b

The queries also not filtered very strictly and YEAR-###### includes ######## results, which is more noticeable when doing 'less than' queries.
Though this is in line with normal Catalog Number behavior and its easily solvable by adding COT to your query.

CarolineDenis (approved):

sharadsw (approved):

alesan99 (approved):
I've made a separate report for the issue I encountered -> https://github.com/specify/specify7/issues/5503

Looks good to go 👍👍👍

acwhite211 (approved):

PR #5489, opened 2024-12-18
Support COT catalog number format configuration in WorkBench

### LINKED ISSUE #5473: Support COT Catalog Numbering Configurations in the WorkBench
**Description:**
https://github.com/specify/specify7/issues/5232 covered the new catalog numbering configurations based on the `CollectionObjectType` of the CO being created, but it did not explicitly mention the WorkBench. This issue is here to ensure that support is added properly.

**Acceptance Criteria:**

- When a user creates a data set including Collection Objects
   - The column mapped to `CatalogNumber` will use the default COT for the collection to determine the format/auto-numbering behavior **unless** the COT is manually specified in the mapping by means of a column being mapped to `CollectionObjectType -> Name`
   - If the `CollectionObjectType -> Name` is mapped, the format/auto-numbering behavior should be dictated by the catalog number format specified in `CatalogFormatNumName` field within the `CollectionObjectType` table
- The WorkBench must be able to handle multiple COTs and catalog number formats within the **same** data set
- This solution should not work only where `CollectionObject` is the base table as COs can be created from other base tables
- WorkBench data sets created before `v7.9.6.2` should still work as expected and be uploadable

melton-jason:
Fixes #5473
Related to #5232 

https://github.com/user-attachments/assets/2e43fa43-8559-44ba-9b7a-a2707d28dbca

### Checklist

- [ ] Add automated tests

### Testing instructions
- Create or find a CollectionObjectType with a Catalog Number Format Name which differs from that of the Collection's default Catalog Number format
   - You can run a query on the Collection table with the `Catalog Num Format Name` field to determine the  default format
- Create a Collection Object WorkBench data set which includes `Collection Object -> catalogNumber` and `Collection Object -> collectionObjectType -> name`
-  Consider the following possibilities for `CollectionObjectType -> name` (these will be used in the next step): 
   - The prior Collection Object Type's name from the first type is selected 
      - This corresponds to the Collection's default format
   - Any other Collection Object Type's name with a different `Catalog Number Format Name` is provided
   - No `Collection Object Type -> name` is provided 
- For each of the above CollectionObjectType configurations, have a row in the data set where: 
   - The catalogNumber matches the expected format of the Collection Object Type
   - The catalogNumber does not match the expected format of the Collection Object Type
- [ ] Validate the data set and ensure that for each row the CollectionObjectType's Catalog Number format is being properly respected
- Remove the invalid rows and Upload the data set 
- [ ] Ensure the Collection Objects are properly uploaded and with the correct Catalog Number and Collection Object Type

- [ ] Perform the above steps with a data set which doesn't have CollectionObject as a base table. 
For example, some tables which map to CollectionObject are: 
   - Accession -> collectionObjects
   - Paleo Context -> collectionObjects 
   - Collecting Event -> collectionObjects
   - `Collection Relationship -> leftSide` and `Collection Relationship -> rightside`
   - Exsiccata Item -> collectionObject

### DIFF
--- a/specifyweb/specify/parse.py
+++ b/specifyweb/specify/parse.py
@@ -43,11 +43,11 @@ class ParseSucess(NamedTuple):
 ParseResult = Union[ParseSucess, ParseFailure]
 
 
-def parse_field(collection, table_name: str, field_name: str, raw_value: str) -> ParseResult:
+def parse_field(collection, table_name: str, field_name: str, raw_value: str, with_formatter = None) -> ParseResult:
     table = datamodel.get_table_strict(table_name)
     field = table.get_field_strict(field_name)
 
-    formatter = get_uiformatter(collection, table_name, field_name)
+    formatter = get_uiformatter(collection, table_name, field_name) if with_formatter is None else with_formatter
 
     if field.is_relationship:
         return parse_integer(field.name, raw_value)
--- a/specifyweb/specify/uiformatters.py
+++ b/specifyweb/specify/uiformatters.py
@@ -324,8 +324,9 @@ def get_uiformatters(collection, obj, user) -> List[UIFormatter]:
         if f is not None
     ]
     if tablename.lower() == 'collectionobject':
-        cat_num_format = get_catalognumber_format(collection, obj, user)
-        if cat_num_format: uiformatters.append(cat_num_format)
+        cat_num_format = getattr(getattr(obj, 'collectionobjecttype', None), 'catalognumberformatname', None)
+        cat_num_formatter = get_catalognumber_format(collection, cat_num_format, user)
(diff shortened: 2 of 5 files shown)


alesan99 (approved):
### Testing instructions
- [x] Validate the data set and ensure that for each row the CollectionObjectType's Catalog Number format is being properly respected
- [x] Ensure the Collection Objects are properly uploaded and with the correct Catalog Number and Collection Object Type

- [x] Perform the above steps with a data set which doesn't have CollectionObject as a base table. 

---

Working as intended 👌
chrome_RPCoC2hPBe

Even holds up in tricky conditions!
chrome_SVgRdVwCqV

CarolineDenis (approved):

sharadsw (approved):

PR #6031, opened 2025-01-02
Add borders to subviews

### LINKED ISSUE #6030: Add borders to subforms
It's difficult to determine which fields belong to which form, especially when a form contains many subviews, fields, and tables.

Conveying to users that "X field belongs to Collecting Event" or "Y field belongs to Agent" is challenging because there is no visual distinction between fields with data from a subview and those from the base table.

While some users may be comfortable with the positioning, it becomes problematic when mapping in the WorkBench and query builder is considered. Data entry is only one aspect; users struggle to identify that "Z field is on this table because it is grouped with relevant fields." This leads to extra clicking and user confusion.

Consider cases like this:

image

* The "Tissues" Object Relationship plugin belong to Collection Object, but the user would not be able to make that distinction. It seems like it is part of the preparation.
* The "Total Count" and "Actual Count" seem like part of the Col Obj Attribute table, but that's not the case.

We could make this much clearer to the user by adding borders:

image

**Before:**
image

**After:**
image

grantfitzsimmons:
Fixes #6030

This has long bothered me and others, but it only took a few minutes to add a fix. Please let me know your thoughts and if you feel this should be a preference, but I am leaning towards it being the default as to remove the common confusion described in the aforementioned issue.

**Before:**
image

**After:**
image
### Checklist

### Testing instructions

This PR adds borders to children of subviews. These borders should only display when:
1. The subview item has data ("No Data" should not have a border)
2. The subview item is not collapsed
3. The subform displays under a header (independent forms without an immediate subview heading should not have a border)

The border should appear whenever a subview has a child (either in a grid or form).

If you have suggestions about visibility or prominence, please let me know. I aimed to keep it subtle to avoid bothering users who are familiar with the system, while still making it visible enough for those learning or reviewing it to use as a guide.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
@@ -65,7 +65,7 @@ export const DataEntry = {
   >(
     'DataEntry.Grid',
     'div',
-    `items-center p-1 -ml-1 gap-2`,
+    `items-center p-1 gap-2`,
     ({
       viewDefinition,
       display,
--- a/specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
@@ -138,7 +138,7 @@ export const DataEntry = {
   SubFormHeader: wrap(
     'DataEntry.SubFormHeader',
     'legend',
-    'gap-2 flex font-bold border-b border-gray-500 pt-5 pb-1 items-center',
+    'gap-2 flex font-bold border-b border-gray-500 pt-3 pb-1 items-center',
     ({ children, ...props }) => ({
       // A hack for Safari. See https://github.com/specify/specify7/issues/1535
       children: <div {...props}>{children}</div>,
--- a/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
@@ -322,7 +322,7 @@ export function ResourceView<SCHEMA extends AnySchema>({
               ? 'hidden'
               : hasNoData
                 ? ''
-                : 'border border-gray-500 border-t-0 rounded-b p-2'
+                : 'border border-gray-500 border-t-0 rounded-b p-1'
           }
         >
           {formattedChildren}
--- a/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
@@ -315,7 +315,7 @@ export function ResourceView<SCHEMA extends AnySchema>({
           </DataEntry.SubFormTitle>
           {headerComponents}
         </DataEntry.SubFormHeader>
-        <div className={isCollapsed ? 'hidden' : (hasNoData ? '' : 'border rounded-bl rounded-br')}>
+        <div className={isCollapsed ? 'hidden' : (hasNoData ? '' : 'dark:border-gray-700 border border-t-0 rounded-bl rounded-br')}>
           {formattedChildren}
         </div>
       </DataEntry.SubForm>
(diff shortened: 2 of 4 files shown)


grantfitzsimmons:
Some real-world examples:

image
image
image
image

CarolineDenis:
I think it should be the default :+1:

CarolineDenis (approved):

specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
```diff
@@ -65,7 +65,7 @@ export const DataEntry = {
   >(
     'DataEntry.Grid',
     'div',
-    `items-center p-1 -ml-1 gap-2`,
+    `items-center p-1 gap-2`,
```
maxpatiiuk, line 68: -ml-1 seems to have been added here 3 years ago: https://github.com/specify/specify7/commit/691cc00ffd8f27693b2be76e5d16e733ea96cc0b#diff-2d8f82f3760b387dcbc745011e630a34606900c7031b02082811b5cf41748dd9L208

from what I remember, it was added to prevent focus outline from being cut-off when the button is placed in the first/last column:
Screenshot 2025-01-02 at 18 40 57

surprisingly though, it appears that the button was being cut off the way this code was, and is no longer cut off without -ml-1 🙆

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
maxpatiiuk, line 1: would it make sense to make the same single border cover the header too, not just the content of the subview?

image

and optionally with a bit more spacing:

Screenshot 2025-01-02 at 18 46 11

maxpatiiuk (approved):
don't think this needs to be a pref - having it on by default sounds good

specifyweb/frontend/js_src/lib/components/Atoms/DataEntry.tsx
grantfitzsimmons: image

😂 On `issue-6030`

specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx
grantfitzsimmons: I considered this approach, but it quickly became cluttered, and the spacing took up too much valuable screen space. Think of nested subviews (Collectors in Collecting Event) and their appearance when collapsed. While I believe this idea works well in many situations and looks more aesthetic in some cases, I'm unsure if we want to change the appearance so significantly and make standard separators visually distinct from a typical subview. I intended for this to be a minimal guide, but I can play with this some more soon.

grantfitzsimmons:
Before:
image

After:
image

alesan99 (approved):
### Testing instructions

This PR adds borders to children of subviews. These borders should only display when:
- [x] The subview item has data ("No Data" should not have a border)
- [x] The subview item is not collapsed
- [x] The subform displays under a header (independent forms without an immediate subview heading should not have a border)
- [x] The border should appear whenever a subview has a child (either in a grid or form).

---

I think the border looks great, its visible on my monitor and it doesn't look jarring.

I checked various forms with different window sizes, along with Browse in Forms and the picklist form in Schema Config. It looks like the extra padding doesn't cause any issues with scrolling as far as I can tell 👍

chrome_azegtOWuHn

CarolineDenis (approved):

emenslin (changes requested):
- [x] The subview item has data ("No Data" should not have a border)
- [x] The subview item is not collapsed
- [x] The subform displays under a header (independent forms without an immediate subview heading should not have a border)
- [x] The border should appear whenever a subview has a child (either in a grid or form).

Looks good, the only thing I noticed is if you have a subview rendered as a grid there is a slight gap above the names when you scroll.

Issue branch: 

https://github.com/user-attachments/assets/73c57315-4942-40cf-a659-95259066c2cc



Production: 

https://github.com/user-attachments/assets/e3e63526-c43e-43c6-a462-32ac8f4bf158

grantfitzsimmons:
@emenslin Good catch– thank you!

> Looks good, the only thing I noticed is if you have a subview rendered as a grid there is a slight gap above the names when you scroll.

Fixed here:

https://github.com/user-attachments/assets/bb11fee6-79ac-4735-af53-b6e62d484d34

emenslin (approved):
- [x] The subview item has data ("No Data" should not have a border)
- [x] The subview item is not collapsed
- [x] The subform displays under a header (independent forms without an immediate subview heading should not have a border)
- [x] The border should appear whenever a subview has a child (either in a grid or form).

Looks good, previous problem is fixed! 

https://github.com/user-attachments/assets/e0515bcf-e63d-44fe-8426-ac47d7d34e9b

PR #6037, opened 2025-01-03
Add "default" initialize property to QCBX

### LINKED ISSUE #3994: Add ability to set any agent query combo box to the current user's agent
>In my preparationproperty form, I need the agent qrycbox to have a default value (the person who logged in).

​The only query combo box which can be set to the currently logged-in user is the `cataloger` field. Here is the specific piece of code that sets the cataloger field to the current logged-in user's agent:
https://github.com/specify/specify7/blob/312139ad3ee9290dd00500e764dd0f0c36f8b026/specifyweb/frontend/js_src/lib/components/FormFields/QueryComboBox.tsx#L80-L86

Unfortunately, since it is hard-coded to only work for the `cataloger` field, it cannot be used for other fields.

**Requested By:** Willem at SAIAB

alesan99:
Fixes #3994

Adds ability for a QCBX to have a user-defined default selected record.

The following property on a QCBX can be set to a specify api uri:
`initialize='default=/api/specify/agent/1/'`
It can also be set to
`CURRENT_AGENT`
`CURRENT_USER`
`BLANK` (Can be used to override hardcoded defaults, like with cataloger)

Notes:
- I chose to use resource URIs rather than record IDs to avoid confusion about ids and to simplify the code. At a glance I think its more clear, but it is a bit unnecessarily detailed. I don't anticipate this functionality being used often, so I think its okay to prioritize clarity over ease of typing. I can switch back to record IDs if those seem better though.
- The Query Builder currently uses a similar keyword `"currentSpecifyUserName"` under the hood, but I chose to go with the sp6 convention of all caps.

### Checklist

- [ ] Add automated tests

### Testing instructions

- Edit any form definition that currently contains a QCBX.
- Add `default=/api/specify/TABLE/ID/` to the QCBX 'initialize' property. (Replace TABLE and ID)
- Load the form or view it in the visual form editor
- [ ] If you set it to a URI, make sure it resolves to the record with the ID you entered
- [ ] If you set it to `CURRENT_AGENT`, make sure it resolves to the agent belonging to the current logged-in user.
- [ ] If you set it to `CURRENT_USER`, make sure it resolves to the current logged-in user.
- Edit the cataloger QCBX in the CollectionObject form.
- [ ] Make sure it fills in automatically like normal before you make any changes
- [ ] If you set it to `BLANK`, make sure it no longer fills in the current agent.
- [ ] Make sure https://github.com/specify/specify7/pull/6215 is still fixed


specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
```diff
@@ -246,6 +247,7 @@ const processFieldType: {
             : getProperty('viewBtn')?.toLowerCase() === 'true',
         typeSearch: getProperty('name'),
         searchView: getProperty('searchView'),
+        defaultRecordId: getProperty('defaultRecordId'),
```
melton-jason, line 250: (just a note) 

If desired, you can do whatever additional parsing here that you need! 

For example, you can check if the value here is one of the defaults and if not, then try and parse it as an integer and otherwise return undefined. 

That way, when working with `defaultRecordId` later down the line, you already know if it's supposed to be converted to a resource_uri, is one of the defaults, etc. 

If you do change how the defaultRecordId is parsed, don't forget to modify the type definition accordingly and other places it's used (like the syncer).

specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
```diff
@@ -862,6 +862,7 @@ const queryComboBoxSpec = (
       syncers.maybe(syncers.toBoolean),
       syncers.default<boolean>(true)
     ),
+    defaultRecordId: syncers.xmlAttribute('initialize defaultRecordId', 'skip'),
```
melton-jason, line 865: (also a note, building on the above comment) 

Whatever parsing behavior you implement in the form parsing code should also be reflected here in the Syncer as well (the Syncer is in charge of converting xml to json and vice-versa; for ease of use for visual editor). 

One benefit of working with the Syncer is that it's _also_ in charge of emitting errors to the user while they are using the CodeMirror text editor. 

For example the following error when trying to render a `to-many` relationship with a QueryComboBox: 

https://github.com/user-attachments/assets/13d8c1c4-5c4a-4b88-a858-5b46dae10372

Is caused because of the following console.error statement on the QueryComboBox syncer!: 

https://github.com/specify/specify7/blob/c55f4dde6f9c8e18051794a32ec3561f77b2ec65/specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts#L820-L836

Maybe you want to check the value: is one of the defaults, the default is being used correctly[^1], a valid number and a record exists with the ID, etc.

[^1]: We know CURRENT_AGENT and CURRENT_USER should only be used for relationships to Agent and Specifyuser (respectively): currently a user can technically use the defaults in any QueryComboBox

melton-jason (approved):
Nice work! 🥳 

My comments are primarily about code style, improving the code that was already there, and ways you could maybe make this even better for both dev and user if you wanted to.
Nothing strikes me as too pressing, and certainly not required (though still worth consideration hopefully). 

Are you planning on adding automated tests to this PR, or will that be later? 👀 

I'm a fan of the defaults, and I love the potential of the feature.
 
I find it a little unfortunate that we have to expose the concept of IDs to the user here: that will certainly require some documentation on the feature and won't be initially intuitive. Hopefully this is a feature that'll save users lots of time and configured infrequently! 

@specify/ux-testing 
Does this solve the most commonly reported use case for #3994? 
Do most users have just a single record they want to set as default for a field?

specifyweb/frontend/js_src/lib/components/FormParse/fields.ts
```diff
@@ -61,6 +61,7 @@ export type FieldTypes = {
       readonly hasViewButton: boolean;
       readonly typeSearch: string | undefined;
       readonly searchView: string | undefined;
+      readonly defaultRecordId: string | undefined;
```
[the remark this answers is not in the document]
alesan99: The default values are stored in DefaultRecordPreset in QueryComboBox/helpers.ts now 👍

specifyweb/frontend/js_src/lib/components/FormEditor/viewSpec.ts
alesan99: I couldn't find an example of syncer being used to validate a property inside the "init=" property so this seems more complicated than I thought for this PR.

specifyweb/frontend/js_src/lib/components/QueryComboBox/helpers.ts
```diff
@@ -230,3 +230,58 @@ export function pendingValueToResource(
...
+  } else if (field.name === 'receivedBy') {
+    const record = toTable(resource, 'LoanReturnPreparation');
+    record?.set(
+      'receivedBy',
+      record?.get('receivedBy') ?? userInformation.agent.resource_uri,
+      {
+        silent: true,
+      }
+    );
+  }
```
sharadsw, line 286: This suggestion might be hyper-optimization so I'll leave it up to you to decide whether to include it or not:
These fallback else-if blocks could potentially be wrapped inside a function `setLegacyDefaults` or something similar just to make the hook more readable. The hook would essentially look like:

```
if (defaultRecord !== undefined) {
    // do something
} else {
   // setLegacyDefaults()
}
```

sharadsw (approved):
Looks good. I like the inclusion of automated tests!

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
```diff
@@ -87,38 +89,10 @@ export function QueryComboBox({
...
-      record?.set(
-        'receivedBy',
-        record?.get('receivedBy') ?? userInformation.agent.resource_uri,
-        {
-          silent: true,
-        }
-      );
-    }
-  }, [resource, field]);
+  useQueryComboBoxDefaults({ resource, field, defaultRecord });
```
sharadsw, line 95: Also, I noticed the hook doesn't use a `useEffect` anymore. Can you verify if that is needed or not?

lexiclevenger (approved):
## Testing instructions




- Edit any form definition that currently contains a QCBX.
- Add `default=/api/specify/TABLE/ID/` to the QCBX 'initialize' property. (Replace TABLE and ID)
- Load the form or view it in the visual form editor
- [x] If you set it to a URI, make sure it resolves to the record with the ID you entered
- [x] If you set it to `CURRENT_AGENT`, make sure it resolves to the agent belonging to the current logged-in user.
- [x] If you set it to `CURRENT_USER`, make sure it resolves to the current logged-in user.
- Edit the cataloger QCBX in the CollectionObject form.
- [x] Make sure it fills in automatically like normal before you make any changes
- [x] If you set it to `BLANK`, make sure it no longer fills in the current agent.
- [x] Make sure https://github.com/specify/specify7/pull/6215 is still fixed
---
Everything works, looks great!

Areyes42 (approved):
-   [x] If you set it to a URI, make sure it resolves to the record with the ID you entered
-   [x] If you set it to `CURRENT_AGENT`, make sure it resolves to the agent belonging to the current logged-in user.
-  [x]  If you set it to `CURRENT_USER`, make sure it resolves to the current logged-in user.
-  [x]  Make sure it fills in automatically like normal before you make any changes
-  [x]  If you set it to `BLANK`, make sure it no longer fills in the current agent.
-  [x]  Make sure [Remove silent option when setting specify user in QCBX #6215](https://github.com/specify/specify7/pull/6215) is still fixed
***
Looks good!

specifyweb/frontend/js_src/lib/components/QueryComboBox/__tests__/defaultRecord.test.ts
```diff
@@ -0,0 +1,54 @@
+import { requireContext } from '../../../tests/helpers';
+import { getResourceApiUrl } from '../../DataModel/resource';
+import { tables } from '../../DataModel/tables';
+import { userInformation } from '../../InitialContext/userInformation';
+import { useQueryComboBoxDefaults } from '../helpers';
+
+requireContext();
+
+test('Query Combo Box with default record selection', () => {
```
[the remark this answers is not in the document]
alesan99: Weird, fixed now 👍

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
alesan99: It didn't seem to cause issues but I restored it out of an abundance of caution to make sure it acts 100% the same.

grantfitzsimmons:
@alesan99 We need to make sure this is documented for users in the [Editing Forms](https://discourse.specifysoftware.org/t/editing-forms-in-specify-7/1557) documentation!

PR #6047, opened 2025-01-08
Add id in empty child COJO condition to avoid subview to disappear

### LINKED ISSUE #6040: Removing cat # from a childCO or name from childCOG causes all the fields to disappear
**Describe the bug**
Introduced in pr #6018, in a COG form with a childCO if you expand the childCO you can remove the cat # and if you do all the fields in the subview disappear. 

**To Reproduce**

1. Go to COG form (go to existing or create new) 
2. Expand childCO 
3. Remove cat #

**Expected behavior**
The cat # should be read only (similar to how it is in places like interactions preparations), or if we want users to be able to remove it for some reason then the fields should remain visible when the cat # is removed but still have the save blocker 

**Screenshots**

https://github.com/user-attachments/assets/9a6d60b7-ac1d-4567-809b-3d1a0cc6afbf

- OS: Windows 11
- Browser: Chrome
- Database Name: ojsmnh
- User Name: spadmin
- URL: https://ojsmnh20241003wb-production.test.specifysystems.org/specify/

CarolineDenis:
Fixes #6040

### Checklist

### Testing instructions

- Go to COG form (go to existing or create new)
- Expand childCO
- Remove cat #
- [ ] Verify the subview does not disappear
- Expand childCOG 
- Remove name 
- [ ] Verify the subview does not disappear

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,6 +23,9 @@ jobs:
           - linux/amd64
           - linux/arm64
     steps:
+      - name: Pre-pull Base Image
+        run: docker pull ubuntu:18.04
+    
       - name: Prepare
         id: prep
         run: |
@@ -87,6 +90,7 @@ jobs:
             GIT_SHA=${{ github.sha }}
           tags: ${{ steps.prep.outputs.tags }}
           push: true
+          retries: 3
 
       - name: Export digest
         run: |
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,9 +23,6 @@ jobs:
           - linux/amd64
           - linux/arm64
     steps:
-      - name: Pre-pull Base Image
-        run: docker pull ubuntu:18.04
-    
       - name: Prepare
         id: prep
         run: |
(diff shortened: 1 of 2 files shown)


grantfitzsimmons (approved):
Much safer solution, thank you!

melton-jason (approved):
Haven't looked much into that code, but didn't even realize you can use dot notation to traverse related records with conditional forms, nice! 

This works as a fix for the common case for now --- viewing a COJO from a related resource. 
Although, the problem still happens when directly viewing the COJO. 

Screenshot 2025-01-08 at 10 10 24 AM

It probably doesn't have to be in this PR, but it'd be nice to resolve this problem eventually. An easy fix would just be to update the form definition to at least include both the `childCo` and `childCog` when both are empty: when one is set, the other condition would take precedence and update the form. 
Business rules should still prevent the user from saving a COJO with either no childCO and childCOG, or both a childCO and childCOG. 

As far as the condition is concerned for conditional forms, what is the intended behavior when a relationship is provided? 
Currently, the behavior is to format the resource (which led to #6040). 
I would think the intended behavior when providing a to-one relationship is to assume the user means `related.id` and base the conditional from that (so the user doesn't have to have knowledge of database ids: or even if they do, a solution like the one in this PR might not be obvious).

pashiav (approved):
- [X] Verify the subview does not disappear
- [x] Verify the subview does not disappear
---
Looks good! Subviews do not disappear.

Screenshot 2025-01-08 at 10 41 05 AM

PR #6054, opened 2025-01-08
Pass is_strict in op_age_period to search_co_ids_in_time_period

### LINKED ISSUE #6017: "Strict" age queries are not working properly
**Reference Query:** https://calvertmarinemuseum20241114-production.test.specifysystems.org/specify/query/19/

In the `calvertmarinemuseum_2024_11_14` database, the "Cretaceous" period has a `startPeriod` of `145` and an `endPeriod` of  `66`. (`/specify/view/geologictimeperiod/35/`)

## Non-strict:
image

See that 119 results are returned, all of which belong specifically to the Paleocence series/epoch, which has a `startPeriod` of `66` and `endPeriod` of `56`:

image

OK– follows the logic described. Because `66` falls within the range `145-66`, though narrowly, it meets the requirements written for the feature.

## Strict:

Same query, but it's strict this time:

image

**Still** 119 results, which doesn't make sense if the intention is to exclude any results that do not fall **completely** within that range, which would exclude the Paleocene records from the results.

image

That's a 🐛 👀

_Originally posted by @grantfitzsimmons in https://github.com/specify/specify7/issues/5465#issuecomment-2561188987_

acwhite211:
Fixes #6017

Fix the bug involving the extend age period queries where the strict mode was not working correctly.  Simple fix by passing is_strict in op_age_period function to the search_co_ids_in_time_period function.  This was accidentally omitted during testing in the original PR.

### Checklist

### Testing instructions

- Use the calvertmarinemuseum_2024_11_14 on the test panel
- Go to the saved query named "Cretaceous Query", also should be at https://calvertmarinemuseum20241114-production.test.specifysystems.org/specify/query/19/
- See that the extend age query field is set to "non-strict mode", and the period name is "Cretaceous"
- Press the "Count" button to run the count query and mark down the resulting number.
- Set the age query field to "strict mode"
- Again, Press the "Count" button to run the count query and mark down the resulting number.
- [x] See that the count from this last query is either zero or any number less than the count from the previous query.

### DIFF
--- a/specifyweb/stored_queries/query_ops.py
+++ b/specifyweb/stored_queries/query_ops.py
@@ -138,4 +138,4 @@ class QueryOps(namedtuple("QueryOps", "uiformatter")):
 
     def op_age_period(self, field, value, query, is_strict=False):
         time_period_name = value
-        return field.in_(search_co_ids_in_time_period(time_period_name, require_full_overlap=False))
+        return field.in_(search_co_ids_in_time_period(time_period_name, require_full_overlap=is_strict))


CarolineDenis (approved):

emenslin (approved):
- [x] See that the count from this last query is either zero or any number less than the count from the previous query.

Looks good, results change between strict and nonstrict

Areyes42 (approved):
- [x] See that the count from this last query is either zero or any number less than the count from the previous query.
***
Looks good, count is lower on strict mode compared to non-strict 👍

grantfitzsimmons:
> See that the count from this last query is either zero or any number less than the count from the previous query.

I'm not sure if these testing instructions are comprehensive enough to determine whether the query is functioning as expected. Many factors could reduce the count of a query, but it seems to be working! I created some examples and found success:

image

Nice job!

PR #6055, opened 2025-01-08
Display age value instead of id in query result for age queries

### LINKED ISSUE #6016: "Age" field returns `CollectionObjectID`, should return useful information
## **Describe the bug**
Currently, adding the "Age" field to a query creates a column displaying the `CollectionObjectID`. This identifier is likely unfamiliar to users and may cause confusion in the interface. This is not a catalog number (identifiable) or any other useful information.

## **To Reproduce**

1. Build a "Collection Object" query in a paleo/invertpaleo/geology discipline's collection
2. Add the "Age" field to the query
3. Run the query
4. See that the Collection Object ID is all that is shown in the results

image

### Show the full range computed by the back end for each CO

- The backend currently search across 3 tables: relative age - absolute age and paleo context. 
- Each value coming from those tables, if any, is transform into a start date and en end date. 
- That new computed range is used to define which CO is returned or not. 
- The column for the virtual field 'Age' in the query result should return that computed range. 
- This could therefore be a different value for each line 

Let say the user adds the age field in his mapping and use the range filter.
He enters the values 120 - 10 in his filter. 
One CO in the db has those value for each age table once converted to range numbers: 
       Absolute age: 110 - 50
       Relative age: 100 - 50
       PaleoContext: 80 - 40 
The maximum range for this CO is then defined to be 110 - 40. 

==> This 110 - 40 should be the value displayed in the query result column

CarolineDenis:
Fixes #6016

### Checklist

### Testing instructions

- create a query with age on CO table 
- run the query 
- The age name should be displayed in the result column age if you selected a name and if you selected a range than the range. The column will be empty if you do not select a name or enter a range.
- [ ] verify the age value is displayed in the column age

Screenshot 2025-01-16 at 6 21 37 AM
Screenshot 2025-01-16 at 6 21 53 AM


emenslin (changes requested):
- [x] verify the age value is displayed in the column age

This does work but wouldn't it make more sense for the range to show what the individual CO range is instead of the range just showing up as whatever you type? 

https://github.com/user-attachments/assets/bca818bf-511c-4e35-b740-a901f3492b16


In this video 000097062 has a range of 3600-3200 and 000097063 has a range of 100000-0 but based on the query results it looks like they both have the same range that I typed. I can approve if the plan is to implement this later but I am a little confused by the current implementation.

specifyweb/stored_queries/execution.py
```diff
@@ -559,7 +559,30 @@ def execute(session, collection, user, tableid, distinct, count_only, field_spec
...
+        if age_field_position is not None:
+            results = [
+                tuple(
+                    age_value.replace(",", " - ") if i == age_field_position + 1 else value
+                    for i, value in enumerate(result)
+                )
+                for result in results
+            ]
+
+        return {'results': results}
```
sharadsw, line 585: I think we should look for a better place to do this if possible. Instead of 'fixing' the query after it has been constructed, it would be better to just construct the query correctly. Looks like the code iterates over each column and each row from the result, which will be super slow for a sufficiently large query. 

Maybe the fix should go wherever the age values are computed? I'd look into this file: https://github.com/specify/specify7/blob/production/specifyweb/specify/geo_time.py

CarolineDenis:
Change in the requirements

PR #6062, opened 2025-01-13
Allow for re-running of creating default cots with django command

### LINKED ISSUE #5466: Default COTypes are not always being set
**Describe the bug**
In some dbs the default COT is not set for all the collections. This is in ojsmnh and geodefaults where no migrations are failing and is potentially in other dbs. 

**To Reproduce**

1. Go to ojsmnh 
2. Use the 'COT query' 
3. See that not all collections have a COT

OR: 

1. Go to geodefaults 
2. Query on collection 
3. Add COT -> name 
4. See it is blank 

(can also check other dbs but it is not on every db) 

**Expected behavior**
Defaut COT should be automatically set for every collection in a db. 

**Screenshots**
Screenshot 2024-12-12 094102

- OS: Windows 11
- Browser: Chrome
- Database Name: ojsmnh, geodefaults
- User Name: spadmin
- URL: https://ojsmnh20241003wb-production.test.specifysystems.org/specify/query/244/

acwhite211:
Fixes #5466

It seems some databases that went through the migrations didn't get their default COTs set properly.  Instead of making a new migration that all databases would need to run, this PR creates a Django migration function that can run selectively on effected databases.  Run the Django command with the CLI command `ve/bin/python manage.py create_default_collection_types`

### Checklist

### Testing instructions

- Run `ve/bin/python manage.py create_default_cots` on a database that doesn't have all of it's default COTs set.
- [ ] See that the default COTs now exists in the database.


CarolineDenis:
Dp we have a list of the dbs that miss the COTs?

grantfitzsimmons:
`beaty` and `ojsmnh` are the only two databases we know are definitely affected by this issue, but I can do a deeper dive if needed. As a very simple first pass, we could just want to iterate a SQL query to find cases where `CollectionObjectTypeID` is null in the `collection` table after the migrations are run.

grantfitzsimmons:
>Instead of making a new migration that all databases would need to run, this PR creates a Django migration function that can run selectively on effected databases. Run the Django command with the CLI command `ve/bin/python manage.py create_default_collection_types`

**One thought:** introducing a new database migration that all databases must run does not cause any harm (outside of our time testing migrations). This migration can functionally make no changes if it is not needed on most databases. Asking users to manually run this migration if the first one fails—something we have encountered at least twice—seems risky. 

Do we understand the underlying cause that resulted in these migrations failing initially? If we know that this migration must be run twice or three times on some databases, can we not do this automatically as part of the initial migration?

I am happy, however, to see that there is a mechanism to do this with a Django command, and would like to see this expanded for other common things: re-creating uniqueness rules, re-running the permissions migrations, etc. Thank you for building this!

grantfitzsimmons (changes requested):
Using a copy of `ojsmnh`:

I ran the initial migration and see that

```sql
select CollectionName, CollectionObjectTypeID from collection;
```

| CollectionName | CollectionObjectTypeID |
| :--- | :--- |
| Ichthyology | null |
| Fossil Vertebrates | null |
| Botany | 3 |
| Invertebrates | 4 |
| Herpetology | null |
| Fossil Invertebrates | null |
| Fossil Plants | null |
| Mammalogy | null |
| Ornithology | null |
| Library | 10 |

> - Run `ve/bin/python manage.py create_default_collection_types` on a database that doesn't have all of it's default COTs set.
> - [ ] See that the default COTs now exists in the database.

I had to run `ve/bin/python manage.py create_default_cots` (seems the name changed in development), but it failed:

```
~/GitHub/specify7 on issue-5466 = *7 ❯ docker exec -it specify7-specify7-1 /bin/bash                                                   Py specify7 at 09:26:36 AM
specify@f3d6d1a9a216:/opt/specify7$ ve/bin/python manage.py create_default_collection_types
Unknown command: 'create_default_collection_types'. Did you mean create_default_cots?
Type 'manage.py help' for usage.
specify@f3d6d1a9a216:/opt/specify7$ ve/bin/python manage.py create_default_cots
  File "manage.py", line 25, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/specify7/ve/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/opt/specify7/ve/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/specify7/ve/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/specify7/ve/lib/python3.8/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/opt/specify7/ve/lib/python3.8/site-packages/django/core/management/base.py", line 511, in handle
    raise NotImplementedError('subclasses of BaseCommand must provide a handle() method')
NotImplementedError: subclasses of BaseCommand must provide a handle() method
specify@f3d6d1a9a216:/opt/specify7$ 
```

melton-jason:
> `beaty` and `ojsmnh` are the only two databases we know are definitely affected by this issue, but I can do a deeper dive if needed

> Do we understand the underlying cause that resulted in these migrations failing initially? If we know that this migration must be run twice or three times on some databases, can we not do this automatically as part of the initial migration?

We can theoretically download a backup of the databases for which this is a known Issue and revert + rerun the migrations to verify that the migrations are not creating default COTs. 
(If this _is_ the case, we already have the databases to find the root cause as to why the default COTs  aren't being created!) 

This seems like a problem with the migration which creates the default COTs, so I think we should at least fix the bug in the migration so this same problem does not happen to any future instances. 

I like the idea of having commands to handle applying defaults to resources created in Specify 6!  It's a very simple solution, albeit with the downside of the user having to know about the commands and use them.  

Other solutions which would apply the defaults automatically (maybe with [AppConfig.ready](https://docs.djangoproject.com/en/3.2/ref/applications/#django.apps.AppConfig.ready) or some other initialization hook) get complex quickly.

grantfitzsimmons:
I would argue this does _not_ fix https://github.com/specify/specify7/issues/5466, but it does something useful.

I tested [another database locally](https://docs.google.com/document/d/1JC37l2GTUfETuJESYTyy7Z-do1BEI5inpH3wEcyEZIY/edit?usp=sharing) and found that it also missed setting some COTs.

acwhite211:
@grantfitzsimmons Should work now.  Both the new django command and the old `create_default_collection_types` in the 0002_geo.py migration file.  I'll add a new migration that will run this function with the corrected logic to make sure all databases are correct.

acwhite211:
Good for re-review @CarolineDenis @grantfitzsimmons

sharadsw (approved):

CarolineDenis (approved):

grantfitzsimmons (approved):
Fixes it in the database I've been testing. 💪

Issue #846, opened 2021-06-13
Unable to disambiguate CoA records


maxxxxxdlp:
I suspect that this issue is database-specific, but even if it is, there is no way to deal with it on the user's end:

A column that belongs to the "Collectiion Attribute" table had ambiguous values. Disambiguation dialog does not format the records properly:

![Screen Shot 2021-06-13 at 1 18 04 PM](https://user-images.githubusercontent.com/40512816/121803383-f29d3500-cc49-11eb-89a2-cd4a6c230fd7.png)

I also can't see the data in the form view:

![Screen Shot 2021-06-13 at 1 18 02 PM](https://user-images.githubusercontent.com/40512816/121803400-1496b780-cc4a-11eb-8886-71c8ea15864d.png)

maxxxxxdlp:
Is adding more columns (to solve the ambiguousness) the only sensible solution in this case? If so, the disambiguation dialog should say that this is an option, instead of showing null values.

benanhalt:
COA shouldn't be shared by COs so this shouldn't be a problem. There is supposed to be a mechanism in the uploader that handles that but it seems like something isn't working. I'll take a look.

benanhalt:
I haven't been able to reproduce this. Do you still have the data set and mapping?

maxxxxxdlp:
I don't remember the mapping anymore (at a time I thought this was a generic issue and so did not attach it here).

Here is the dataset, though, it is not very useful without the mapping:
[7000_CO.csv](https://github.com/specify/specify7/files/6695335/7000_CO.csv)

I am not able to reproduce this either so would close the issue for now.

benanhalt:
This is a bit tricky. You are only going to see the MatchedMultiple COAs when there is error that stops the CO from being uploaded. It's actually correct that the COA matched multiple records, it just doesn't usually care because it is a one-to-one relationship and if it has to upload the CO it is going to upload a new COA.

I'm not sure the best way to fix this. Basically the MatchedMultiple should be ignored on one-to-one relationships but there isn't any easy way to know on the frontend which relationships those are. I could maybe add an "ignore" field to the MatchedMultiple result in those cases. Or I might be able to suppress them altogether so there would just be no UploadResult for the COA in this case.

maxxxxxdlp:
If you can't easily add the "ignore" field on the back end, I have a utility function to get a relationship type from the mapping path

maxxxxxdlp:
[...]

Issue #1835, opened 2022-07-14
WorkBench upload status error for success


grantfitzsimmons:
![image](https://user-images.githubusercontent.com/37256050/179069413-bc4602be-ffbd-48af-a09f-6884ff00db94.png)

I have encountered this many times over the past few months and it is very inconsistent. This message appears after a large successful update at random times.

grantfitzsimmons:
<img width="691" alt="image" src="https://user-images.githubusercontent.com/37256050/224121895-383a54e6-db94-45ba-abaa-dd40d643e2a9.png">

Still funny each time it happens

grantfitzsimmons:
Still happens quite often (unsurprisingly).

emenslin:
I ran into this earlier on a small data set but could not recreate again, not sure what's going on

PR #6073, opened 2025-01-14
Allow to upload COG in WB

### LINKED ISSUE #5418: Issues with uploading COGs through workbench
**Describe the bug**
Overall uploading COGs in the workbench does not always seem to work as expected except for the basic cases with just cog name and type. I'm going to list a few things I ran into in this issue, some may be issues and some may be the correct behavior but might just need further explanation, either way I wanted to list them all out just in case. 

1. When adding COG children to a COG you can mark isPrimary and isSubstrate as true
2. Along with that 'isPrimary' seems to be hidden in workbench but not anywhere else, if this is the expected behavior it should be explained
3. Sometimes the COGtype doesn't get set properly and shows the api, this normally fixes itself but still shouldn't be shown to the user in the first place
4. Making a data set with parent cog will throw an error when validating 
5. If a CO is already a child to a COG and you try to map it to a different COG the error message isn't very clear
 

**To Reproduce**

1. Go to workbench 
2. Select COG as base table 
3. Map all required fields for COG 
4. Go through 'children' and map to CO and COG 

Example mappings based on WB required fields: https://nbmmnb20240628-production.test.specifysystems.org/specify/workbench/plan/116/ 

[upload_plan.json](https://github.com/user-attachments/files/18275942/upload_plan.json)

Query of existing COGs: https://nbmmnb20240628-production.test.specifysystems.org/specify/query/47/

**Expected behavior**
Either these issues should be fixed or there should be an explanation for why the current behavior is correct 

**Screenshots**

Setting child COG 'isSubstrate' to true: 

https://github.com/user-attachments/assets/764478fd-e091-40c6-a26e-d0d96d2bac46

Shows api: 

https://github.com/user-attachments/assets/d27c20ba-1395-4bf5-b8aa-f79f2d917279

I uploaded this data set and made no changes and then I couldn't roll it back, also shows the api: 

https://github.com/user-attachments/assets/701b4f38-f114-4291-ba6f-34cf17576d84

Data set link: https://nbmmnb20240628-production.test.specifysystems.org/specify/workbench/114/ 

Can't validate data set if it includes cog -> parentcog (likely the new behavior instead of video above): 

https://github.com/user-attachments/assets/07190b73-afbc-4c58-b079-b4db3a517da8

CO is a child of another COG but the error message shows under 'isSubstrate' instead of the cat # and the error message has too much information in my opinion.

Screenshot 2024-12-26 105034

- OS: Windows 11
- Browser: Chrome
- Database Name: nbmnb_2024_06_28 (updates on ojsmnh20241003wb)
- Collection name: Amphibian and Reptile Observations
- User Name: spadmin
- URL: https://nbmmnb20240628-production.test.specifysystems.org/specify/

**Additional context**
I probably missed some things so I encourage whoever works on this to check other cases in the UI to see if the behavior is correct or not. (Updated with any changes/fixes on 12/26/2024)

CarolineDenis:
Fixes #5418 

### Checklist

- [ ] Add automated tests

### DIFF
--- a/specifyweb/businessrules/rules/cojo_rules.py
+++ b/specifyweb/businessrules/rules/cojo_rules.py
@@ -1,18 +1,17 @@
-import os
-import sys
 from enum import Enum
+
+from django.db.models import Max
 
 from specifyweb.businessrules.exceptions import BusinessRuleException
 from specifyweb.businessrules.orm_signal_handler import orm_signal_handler
 from specifyweb.specify.models import Collectionobjectgroupjoin
+
 
 class COGType(Enum):
     DISCRETE = "Discrete"
     CONSOLIDATED = "Consolidated"
     DRILL_CORE = "Drill Core"
 
-def is_running_tests():
-    return any(module in sys.modules for module in ('pytest', 'unittest'))
 
 @orm_signal_handler('pre_save', 'Collectionobjectgroupjoin')
 def cojo_pre_save(cojo):
@@ -41,24 +40,33 @@ def cojo_pre_save(cojo):
         cojo.childcog is not None
         and cojo.childcog.cojo is not None
         and cojo.childcog.cojo.id is not cojo.id
-        and not is_running_tests()
     ):
-        raise BusinessRuleException('ChildCog is already in use as a child in another COG.')
+        raise BusinessRuleException(
+            'ChildCog is already in use as a child in another COG.')
 
     if (
         cojo.childco is not None
         and cojo.childco.cojo is not None
         and cojo.childco.cojo.id is not cojo.id
-        and not is_running_tests()
     ):
-        raise BusinessRuleException('ChildCo is already in use as a child in another COG.')
-    
+        raise BusinessRuleException(
+            'ChildCo is already in use as a child in another COG.')
+
+    if cojo.precedence is None:
+        others = Collectionobjectgroupjoin.objects.filter(
+            parentcog=cojo.parentcog
+        )
+        top = others.aggregate(Max('precedence'))['precedence__max']
+        cojo.precedence = 0 if top is None else top + 1
+
+
 @orm_signal_handler('post_save', 'Collectionobjectgroupjoin')
 def cojo_post_save(cojo):
     """
         For Consolidated COGs, mark the first CO child as primary if none have been set by the user
     """
-    co_children = Collectionobjectgroupjoin.objects.filter(parentcog=cojo.parentcog, childco__isnull=False)
+    co_children = Collectionobjectgroupjoin.objects.filter(
+        parentcog=cojo.parentcog, childco__isnull=False)
     if len(co_children) > 0 and not co_children.filter(isprimary=True).exists() and cojo.parentcog.cogtype.type == COGType.CONSOLIDATED.value:
         first_child = co_children.first()
         first_child.isprimary = True
--- a/specifyweb/workbench/upload/upload_table.py
+++ b/specifyweb/workbench/upload/upload_table.py
@@ -367,7 +367,9 @@ class BoundUploadTable(NamedTuple):
 
         info = ReportInfo(tableName=self.name, columns=[pr.column for pr in self.parsedFields], treeInfo=None)
 
-        toOneResults_ = self._process_to_ones()
+        local_to_ones, remote_to_ones = separate_to_ones(model, self.toOne)
+
+        toOneResults_ = self._process_to_ones(local_to_ones)
 
         multi_one_to_one = lambda field, result: self.toOne[field].is_one_to_one() and isinstance(result.record_result, MatchedMultiple)
 
@@ -412,13 +414,13 @@ class BoundUploadTable(NamedTuple):
             if match:
                 return UploadResult(match, toOneResults, {})
 
-        return self._do_upload(model, toOneResults, info)
+        return self._do_upload(model, toOneResults, remote_to_ones, info)
 
-    def _process_to_ones(self) -> Dict[str, UploadResult]:
+    def _process_to_ones(self, toOnes: Dict[str, BoundUploadable]) -> Dict[str, UploadResult]:
         return {
             fieldname: to_one_def.process_row()
             for fieldname, to_one_def in
-            sorted(self.toOne.items(), key=lambda kv: kv[0]) # make the upload order deterministic
+            sorted(toOnes.items(), key=lambda kv: kv[0]) # make the upload order deterministic
         }
 
     def _match(self, model, toOneResults: Dict[str, UploadResult], toManyFilters: FilterPack, info: ReportInfo) -> Union[Matched, MatchedMultiple, None]:
@@ -463,7 +465,7 @@ class BoundUploadTable(NamedTuple):
         else:
             return None
 
-    def _do_upload(self, model, toOneResults: Dict[str, UploadResult], info: ReportInfo) -> UploadResult:
+    def _do_upload(self, model, toOneResults: Dict[str, UploadResult], remoteToOnes: Dict[str, BoundUploadable], info: ReportInfo) -> UploadResult:
         missing_requireds = [
             # TODO: there should probably be a different structure for
             # missing required fields than ParseFailure
@@ -516,12 +518,14 @@ class BoundUploadTable(NamedTuple):
 
         self.auditor.insert(uploaded, self.uploadingAgentId, None)
 
+        remoteToOneResults = _upload_remote_to_ones(model, uploaded.id, remoteToOnes)
+
         toManyResults = {
             fieldname: _upload_to_manys(model, uploaded.id, fieldname, self.uploadingAgentId, self.auditor, self.cache, records)
             for fieldname, records in
             sorted(self.toMany.items(), key=lambda kv: kv[0]) # make the upload order deterministic
         }
-        return UploadResult(Uploaded(uploaded.id, info, picklist_additions), toOneResults, toManyResults)
+        return UploadResult(Uploaded(uploaded.id, info, picklist_additions), {**toOneResults, **remoteToOneResults}, toManyResults)
 
     def _do_insert(self, model, **attrs) -> Any:
         return model.objects.create(**attrs)
@@ -547,13 +551,13 @@ class BoundMustMatchTable(BoundUploadTable):
     def force_upload_row(self) -> UploadResult:
         raise Exception('trying to force upload of must-match table')
 
-    def _process_to_ones(self) -> Dict[str, UploadResult]:
+    def _process_to_ones(self, toOnes: Dict[str, BoundUploadable]) -> Dict[str, UploadResult]:
         return {
             fieldname: to_one_def.match_row()
-            for fieldname, to_one_def in self.toOne.items()
+            for fieldname, to_one_def in toOnes.items()
         }
 
-    def _do_upload(self, model, toOneResults: Dict[str, UploadResult], info: ReportInfo) -> UploadResult:
+    def _do_upload(self, model, toOneResults: Dict[str, UploadResult], remoteToOnes: Dict[str, BoundUploadable], info: ReportInfo) -> UploadResult:
         return UploadResult(NoMatch(info), toOneResults, {})
 
 
@@ -588,3 +592,25 @@ def _upload_to_manys(parent_model, parent_id, parent_field, uploadingAgentId: Op
         ).force_upload_row()
         for record in records
     ]
+
+def _upload_remote_to_ones(parent_model, parent_id, remoteToOnes: Dict[str, BoundUploadTable]) -> Dict[str, UploadResult]: 
+    toOnes: Dict[str, UploadResult] = dict()
+    for field_name, upload_table in sorted(remoteToOnes.items(), key=lambda kv: kv[0]): 
+        related_field = parent_model._meta.get_field(field_name).field
+        related_column = related_field.attname
+        static = {**upload_table.static, related_column: parent_id}
+        toOnes[field_name] = upload_table._replace(static=static).process_row()
+    return toOnes
+
+
+def separate_to_ones(parent_model, toOnes: Dict[str, BoundUploadable]): 
+    remote_to_ones = dict()
+    local_to_ones = dict()
+
+    for field_name, uploadable in toOnes.items(): 
+        field = parent_model._meta.get_field(field_name)
+        if field.concrete: 
+            local_to_ones[field_name] = uploadable
+        else: 
+            remote_to_ones[field_name] = uploadable
+    return local_to_ones, remote_to_ones
--- a/specifyweb/frontend/js_src/lib/components/DataModel/schemaOverrides.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/schemaOverrides.ts
@@ -237,6 +237,13 @@ const fieldOverwrites: typeof globalFieldOverrides = {
   CollectionObject: {
     collectionObjectType: { visibility: 'optional' },
   },
+  CollectionObjectGroupType: {
+    type: { visibility: 'optional' },
+  },
+  CollectionObjectGroupJoin: {
+    precedence: { visibility: 'optional' },
+    isSubstrate: { visibility: 'optional' },
+  },
   LoanPreparation: {
     isResolved: { visibility: 'optional' },
   },
--- a/specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Preferences/UserDefinitions.tsx
@@ -1103,8 +1103,7 @@ export const userPreferenceDefinitions = {
         },
       },
       recordSet: {
-        title: () =>
-          tableLabel('RecordSet'),
+        title: () => tableLabel('RecordSet'),
         items: {
           recordToOpen: definePref<'first' | 'last'>({
             title: preferencesText.recordSetRecordToOpen(),
@@ -1348,7 +1347,9 @@ export const userPreferenceDefinitions = {
             defaultValue: false,
             type: 'java.lang.Boolean',
           }),
-          orderByField: definePref<'fullName' | 'name' | 'nodeNumber' | 'rankId'>({
+          orderByField: definePref<
+            'fullName' | 'name' | 'nodeNumber' | 'rankId'
+          >({
             title: preferencesText.sortByField(),
             requiresReload: false,
             visible: true,
@@ -1413,8 +1414,7 @@ export const userPreferenceDefinitions = {
          * This would be replaced with labels from schema once
          * schema is loaded
          */
-        title: () =>
-            tableLabel('Geography'),
+        title: () => tableLabel('Geography'),
         items: {
           treeAccentColor: definePref({
             title: preferencesText.treeAccentColor(),
@@ -1435,8 +1435,7 @@ export const userPreferenceDefinitions = {
         },
       },
       taxon: {
-        title: () =>
-          tableLabel('Taxon'),
+        title: () => tableLabel('Taxon'),
         items: {
           treeAccentColor: definePref({
             title: preferencesText.treeAccentColor(),
@@ -1457,8 +1456,7 @@ export const userPreferenceDefinitions = {
         },
       },
       storage: {
-        title: () =>
-          tableLabel('Storage'),
+        title: () => tableLabel('Storage'),
         items: {
           treeAccentColor: definePref({
             title: preferencesText.treeAccentColor(),
@@ -1479,8 +1477,7 @@ export const userPreferenceDefinitions = {
         },
       },
       geologicTimePeriod: {
-        title: () =>
-          tableLabel('GeologicTimePeriod'),
+        title: () => tableLabel('GeologicTimePeriod'),
         items: {
           treeAccentColor: definePref({
             title: preferencesText.treeAccentColor(),
@@ -1501,8 +1498,7 @@ export const userPreferenceDefinitions = {
         },
       },
       lithoStrat: {
-        title: () =>
-          tableLabel('LithoStrat'),
+        title: () => tableLabel('LithoStrat'),
         items: {
           treeAccentColor: definePref({
             title: preferencesText.treeAccentColor(),
@@ -1523,8 +1519,7 @@ export const userPreferenceDefinitions = {
         },
       },
       tectonicUnit: {
-        title: () =>
-          tableLabel('TectonicUnit'),
+        title: () => tableLabel('TectonicUnit'),
         items: {
           treeAccentColor: definePref({
             title: preferencesText.treeAccentColor(),
@@ -2052,10 +2047,7 @@ import('../DataModel/tables')
           ),
           'Unable to find tree full name value'
         );
-        overwriteReadOnly(
-          name, 
-          'title', 
-          getField(tables.Taxon, 'name').label);
+        overwriteReadOnly(name, 'title', getField(tables.Taxon, 'name').label);
         overwriteReadOnly(
           fullName,
           'title',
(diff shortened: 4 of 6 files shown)


specifyweb/workbench/upload/upload_table.py
```diff
@@ -588,3 +592,25 @@ def _upload_to_manys(parent_model, parent_id, parent_field, uploadingAgentId: Op
         ).force_upload_row()
         for record in records
     ]
+
+def _upload_remote_to_ones(parent_model, parent_id, remoteToOnes: Dict[str, BoundUploadTable]) -> Dict[str, UploadResult]:
```
realVinayak, line 596: Sorry for the intrusion. Curious if this could be fixed by secretly parsing these cases as to-many with just one item? Currently, I believe the remoteToOnes are also not used for matching (which would be fixed by to-many). Also, the below process_row should be a force_row? Need to think more if nested to many support from batch edit aids here or not.

With parsing as to-many would also make batch-edit simpler! Maybe just the parsing would be chaged (tomanyrecord is not needed anymore).

melton-jason: Yeah you're absolutely right! 🤦 
The `remoteToOnes` were not matching at all. 

Is something like https://github.com/specify/specify7/pull/6073/commits/04b576e206d14a416d8edca9fbe4f74bebedd669 what you had in mind? 

This might be getting a little too caught on semantics but this still isn't really _parsing_ the remote toOne relationships as toMany. 
Are there any other changes that could be made to make this easier to work with for batch-edit? 

Ping for @sharadsw as well. 
Feel free to look over this PR and provide any insights for how the changes in this branch interact with batch-edit.

realVinayak: lol you started late night coding too!

I think that's much better, since code is not repeated anymore. But they are still not used for matching, right? That's why I recommended handling it at the parsing stage (much higher level than this lower level).

By parsing, I was thinking that any such relationship (remote 1-to-1) will _appear_ as to-many (with 1 item), so, ideally, none of these lines would be needed (in this file).

1. Maybe here you can try parsing modification (I guess don't bother with deferred, it's removed in batch-edit anyways)? So, after this, remote-to-ones would, really, exist as to-manys with one item. https://github.com/specify/specify7/blob/5466aea12b1bdf4f6daad0d08908f5e3012fc9c4/specifyweb/workbench/upload/upload_plan_schema.py#L252

There are other different implementations you can do: 

2. overwrite to_one and to_many directly (not reassign - named tuple) in BoundUploadTable. That is, remove remote-to-one from to_one and add it to to_many. Not recommended, it'd make code in future harder (non-deterministic and impure function), although, it has its appeal (you only need to add 1 function to BoundUploadTable --- BoundUpdateTable will inherit it). With this change, we'd need no change to support batch-edit (another appeal).

3. Make a function that passes over bounded uploadables later on. But this would complicate things (you'd either need to type ignore or type cast or implement some toToManyRecord method for each BoundUploadable / ScopedUploadable etc).

That's why doing it at the bare parsing might be nicer. We'd need to have slight code duplication when determining the IDs. https://github.com/specify/specify7/blob/4d82298ac772122992afd7f767c3dc25e2e52faa/specifyweb/workbench/upload/upload_table.py#L161
If you go with parsing, could just add a comment to consider remoteToOnes? We can come to that later on. that, from what I remember, is the only change needed to make it batch-edit.

If you go by implementation 2, you'd not need that code duplication BUT you'd need to more careful in using toOnes or toManys. To be honest, it doesn't matter much, I don't know what's better, both have some drawbacks. Keeping it in parsing would play nicer later on (everything below it won't care about remoteToOnes,  I like that a lot from abstraction perspective). I'd have just did the parsing, and if it causes bugs / more complexity downstream (other than the apply_batch_edit_pack) , just try a different implementation later on. But, definitely, if you come up with a better implementation, do it!

It's, however, weird that this issue is just now being visible! But makes sense, previous COA to CO were just misinterpreted as to-manys. Also, if you think about it, this also fixes zero-to-one.....

In the current state of the PR, just that logic would need to be carried to batch-edit somewhere. Gosh its hard to look at months old code again. You'd need to separate them out again.

realVinayak: TEST: how does front-end handle seeing the tomany result (for to-one)?

specifyweb/businessrules/rules/cojo_rules.py
```diff
@@ -41,24 +40,33 @@ def cojo_pre_save(cojo):
...
     if (
         cojo.childco is not None
         and cojo.childco.cojo is not None
         and cojo.childco.cojo.id is not cojo.id
-        and not is_running_tests()
     ):
-        raise BusinessRuleException('ChildCo is already in use as a child in another COG.')
-    
+        raise BusinessRuleException(
+            'ChildCo is already in use as a child in another COG.')
```
grantfitzsimmons, line 53: Can this error be more friendly? 

> One of these COs is already a child in another COG

CarolineDenis: agree but should be defer to new issue

emenslin (changes requested):
- [ ] Ensure you can disambiguate related records when there is more than one potential match
    - [x] Ensure that matched records do not count as "New Cells"
    - [x] Create a record set with the uploaded results, and ensure that all data is uploaded correctly
    - [x] Ensure that all created records are deleted and all independent relationships (for example,  CO -> cojo if assigning an existing CO to a new/existing COG) are reverted to their state prior to the upload
  - [x] Ensure CollectionObjects can be added to existing COGs via the WorkBench
  - [x] Ensure that COGs can be created as part of the upload process for a CO and that COs can be associated with 
  - [ ] Ensure a COJO can not be created with both a childCo and childCog
  - [ ] Ensure a COJO can not be created without a childCo and childCog
  - [x] Ensure the `precedence` of the COJO records are automatically set based on the order they are added

--- 

Cannot disambiguate 


https://github.com/user-attachments/assets/1a71a3c8-b587-450a-a564-0cab2896ab6e


Can upload a COJO without any children and can upload a COJO with both a childCO and childCOG 

https://github.com/user-attachments/assets/bb780064-bbce-4a88-84b2-0cabe06d2e17

specifyweb/workbench/upload/upload_table.py
melton-jason: > By parsing, I was thinking that any such relationship (remote 1-to-1) will appear as to-many (with 1 item), so, ideally, none of these lines would be needed (in this file).
> 
> Maybe here you can try parsing modification (I guess don't bother with deferred, it's removed in batch-edit anyways)? So, after this, remote-to-ones would, really, exist as to-manys with one item.

I do like this approach more than the 'half-to-one', 'half-to-many' approach this PR takes currently. 
As you mentioned, the abstraction is a big plus, it uses the systems already in place for the WB, and it's simple to implement on the backend (the only work we'd need to do is to separate the remote/local to-ones when parsing in [upload_plan_schema.py](https://github.com/specify/specify7/blob/production/specifyweb/workbench/upload/upload_plan_schema.py))

However, won't the consequence of going with the full to-many approach for remote to-one relationships be that to-many relationships won't be supported until nested to-manys are supported (#4929, #5417)? 

This was fine for the two `zero-to-one` relationships (Locality -> geoCoordDetails and Locality -> localityDetails)  because those respective 'XDetail' tables don't have any to-many relationships, but this is not the case for the remote to-one relationships. 

For the context of COGs, this would mean that any to-many relationships from `COJO -> childCo` and `COJO -> childCog` (this includes all to-manys from `COG -> children -> childCo` and `COG -> children -> childCog`) won't be supported until nested to-many support (which could be totally fine if Batch Edit is being included in the same release, or the requirements allow it) 

We would need some work on the frontend as well to allow not mapping to-many relationships from remote-to-one relationships (at this point, the frontend can just be in charge of putting remote to-one relationships as to-many in the upload plan). 
--- 

> overwrite to_one and to_many directly (not reassign - named tuple) in BoundUploadTable. That is, remove remote-to-one from to_one and add it to to_many.

I've been considering a similar, more "pure" but far less abstraction-oriented approach for this, which would be to separate out the remote to-one relationships when parsing and have a `remoteToOne` attribute for UploadTables.
The `remoteToOne` on upload tables would be parsed and scoped like `toOne`, uploaded/rollbacked similarly to `toMany`,  and included in the `toOneResults` of the UploadResult. Of course an approach would require a lot of "repeating ourselves" (e.g., the operations on remoteToOne would be identical to toOne until the Bound stage, after which operations would mirror of that toMany), and require more work on the Batch Edit side. 

Or we could just separate out the remote-to-ones in the Bound stage when uploading/matching/rollback, which might be identical and easier to work with in other areas, but be more fragile if we're moving entires from `toOne` to `toMany` and vise versa.

realVinayak: 1. Currently, in production, we allow things in workbench (from frontend) that backend simply doesn't support. For example, `CO -> CE -> collectors -> agent -> agent address`. That's straight-up invalid (backend doesn't bother matching with agent address). This is _technically_ nested-to-many (depends on your definition) but backend doesn't expect it.  Although it is fixed in batch edit PRs, it is _fine_ to have some bugs slip through right now which will be fixed soon. I did a bad job of putting everything in a single PR.

INFO: (I'm confused) 

> For the context of COGs, this would mean that any to-many relationships from COJO -> childCo and COJO -> childCog

Is this a valid example? The current PR only cares if the foreign key is on the remote side, and I thought COJO contains those FKs (so, it really is a to-one).
https://github.com/specify/specify7/blob/9563d8854f9af014ae069f52b0b12382762467e5/specifyweb/specify/models.py#L7612-L7613
The current PR won't care for these (which is good). So, to-manys would be supported out of these relationships. I don't mean to nitpick - want to make sure I'm not just creating more confusion for you guys.

_Technically_ `COG -> children -> childCo` should be supported, already (without _any_ change from this PR), right? If they are not, there's likely an unrelated bug.

Consider below: 

COG -> (to-many to COJO) children -> (to-one to collection object and FK on COJO) CO 

^ the above, is equivalent to something like `collecting event -> collectors -> agent`, right? (Assuming not caring about dependent, pls). "Collectors is many-to-one to agent" is also not a valid argument since FK to CO / COG likes on COJO.

3. So, the only relationships that seem to be affected would be the "cojo" relationship from both CO and COG, right? This is because those are the only reverse of OneToOne fields (that means, FK from CO and COG _for_ relationship "cojo" does not lie in CO and COG, but instead on COJO table).  By affected, I mean the code in this PR only concerns about those. 

4. Now, assuming that you are doing something like CO -> COJO (but, if I'm wrong about the relationship, the argument still holds). I'd not recommend adding another "remoteToOne" prop - I think it's messy. To be honest, I don't even see whether matching would be possible _without_ considering it as a to-many (like how would you do it? You can't match the remote object as a to-one). You'd _need_ to consider it as a to-many (it's just not possible otherwise, the _entire_ CO along with COJO needs to be matched _together_). This argument is essentially identical to why we match to-manys by constructing a _single_ query along with the main object. That is, CO -> Dets get matched together, and not like Dets get matched and then CO gets matched.

However, one of your arguments was that nested-to-manys will be needed. After this, would it be? I don't see a to-many from COJO (no table seems to have relationships to COJO -- that's effectively enough). You _don't_ need nested-to-manys because we can't make a relationship going like 

`CO -> COJO (remote-to-one -- parsed as to-many) -> [any table] (to-many)`.

5. In conclusion, I don't think nested-to-manys are needed. There will be a bug (indicated in 1st point, for something like CO -> COJO -> COG / CO -> to-many [this may be fictious] ) but that's an already existing bug (fixed in either batch edit PR). So, IMO, making a front-end change to put them in to-many (as you proposed) would be a nice way for this case. Otherwise, parse it as to-many. In the current implementation (without parsing change), you could force a match, by using below (be aware that it's replaced in batch edit PR)

https://github.com/specify/specify7/blob/9563d8854f9af014ae069f52b0b12382762467e5/specifyweb/workbench/upload/upload_table.py#L560

Regarding rollbacks, ypu'd need to put them in a to-many, but I think that's what you were saying anyways.

TLDR: parsing as to-many should work just fine ([since we can't have a to-many from COJO](https://github.com/specify/specify7/blob/b68f9a37009c32ece2ea3e7db4e2f0189506fa88/specifyweb/frontend/js_src/lib/components/DataModel/types.ts#L6526-L6553)) -- so we don't need nested-to-many support. Making a front-end change (to put in to-many) is even better, if not too hard. We could just change the parse/unparse without changing the actual relationship type.

realVinayak: Forgot to explicitly address this point and tie it back.

> (this includes all to-manys from COG -> children -> childCo and COG -> children -> childCog) won't be supported until nested to-many support 

I don't think so, because none of these relationships in the path are "remote-to-ones", right? Any to-many that follows these should not be allowed by the frontend - but it does, a bug (see point 1 from previous comment) - fixed in batch edit.

melton-jason: Sorry for the confusion: yep, the remote to-ones would currently be `CO -> cojo` and `COG -> cojo`, so my points about `COJO -> childCO` and `COJO -> childCog` would actually apply to their reverse relationships. 
I think I was confused and switched around the relationships at the time of writing that: probably the result of one too many late-night coding sessions, heh. 

Not being able to map `CO -> cojo -> children` and `COG -> cojo -> children` currently is much more manageable (and makes more sense). 

I reverted the changes made on the backend for this PR and instead went with the same approach we take with `zero-to-one` and similar relationships on the frontend: by mapping them as a singular `toMany` in the upload plan: https://github.com/specify/specify7/pull/6073/commits/07dfcf2a227b45dc70d465e6d218929292d59812

The implementation in the above commit creates a new function, `relationshipIsRemoteToOne`, and uses it in conjunction with `relationshipIsToMany` in the workbench mapping code. 

Ideally, I think the logic should probably be included directly in the `relationshipIsToMany` helper function, but that helper function has been used extensively throughout the frontend: it is not just "scoped" to the WorkBench. In areas outside of the WorkBench, remote to-one relationships should currently be treated as to-one relationships (e.g., with the main resource API). 
Either way, when using the helper function we would need to have additional logic to either expand the concept of a to-many relationship in the workbench (the approach taken in this branch) or to restrict the concept of a to-many relationship for the areas outside of the Workbench. 

Alternatively, we could define a 'context' parameter for  `relationshipIsToMany`, define additional `relationshipIsToMany` functions: one for each context, remove usages of `relationshipIsToMany` outside of the WorkBench (potentially renaming the function), etc.

melton-jason:
Hey guys! 

> Cannot disambiguate 
> 
> 
> https://github.com/user-attachments/assets/1a71a3c8-b587-450a-a564-0cab2896ab6e
> 
> 
> Can upload a COJO without any children and can upload a COJO with both a childCO and childCOG 
> 
> https://github.com/user-attachments/assets/bb780064-bbce-4a88-84b2-0cabe06d2e17

From https://github.com/specify/specify7/pull/6073#pullrequestreview-2557103884

These Issues have been fixed. You should now be able to properly disambiguate. The businessrules to prevent saving when a COJO had had neither a childCO or childCOG or both a childCO and childCOG were disabled in https://github.com/specify/specify7/commit/fc850091ad3223211ff0a9f53dfc794496510390, but have been re-enabled! 

The actual content of the error message can likely be modified/improved in whichever PR tackles https://github.com/specify/specify7/pull/6073#discussion_r1918823483. 

--- 

Now, remote one-to-one relationships are instead parsed as a toMany relationship. What this means is that `CO -> cojo` and `COG -> cojo` will be treated the same way we treat `Locality -> localityDetails` and `Locality -> geoCoordDetails` in the Workbench. 
From the perspective of a user of the Web-Application, this means that an additional click is needed to map the "toOne" as a singular "toMany" : 

**Before**

https://github.com/user-attachments/assets/96a74419-4502-4d81-b9d1-70ffd0a318a6


**Now**

https://github.com/user-attachments/assets/7582740a-6727-416c-be91-9b13b6f3c9de

This decision is the result of the discussion in https://github.com/specify/specify7/pull/6073#discussion_r1915397675 (thank you for the insightful replies, `realVinayak`). This approach leverages functionality which already exists in Specify: no change to the internal logic of disambiguating, uploading, and rollback was needed. Because of this, the changes in this branch will hopefully have a minimal impact for batch edit (https://github.com/specify/specify7/pull/4929 and https://github.com/specify/specify7/pull/5417). 

Also, I've included a fix for #1835. You should now longer longer receive the error message for a successful validation/, upload, or rollback.

specifyweb/workbench/upload/upload_table.py
realVinayak: The current WB implementation seems fine -- for future reference, batch edit's construction of upload plan needs to be tweaked a little bit:

https://github.com/specify/specify7/blob/b68f9a37009c32ece2ea3e7db4e2f0189506fa88/specifyweb/stored_queries/batch_edit.py#L392-L396

TBH, either BE PR also improves another thing: matching of one-to-ones. Previously, one-to-ones get treated as to-ones (that leads to ugly logic here: https://github.com/specify/specify7/blob/36f7a7eddb9216cf634034105a36ad7672fe6571/specifyweb/workbench/upload/upload_table.py#L374-L389). The above was added by Ben to fix https://github.com/specify/specify7/issues/846.

However, implementation in production also causes a bug here:
https://github.com/specify/specify7/blob/36f7a7eddb9216cf634034105a36ad7672fe6571/specifyweb/workbench/upload/tests/testuploading.py#L441-L478

4929 reworks the matching, and now one-to-one is matched _together_ with the object (just like to-manys) -- which is what we wanted out of this PR. This matching of one-to-one together also allows multiple matches to be treated naturally (and simplifies upload logic below:) 

Production:
https://github.com/specify/specify7/blob/36f7a7eddb9216cf634034105a36ad7672fe6571/specifyweb/workbench/upload/upload_table.py#L484-L495

vs BE:
https://github.com/specify/specify7/blob/b68f9a37009c32ece2ea3e7db4e2f0189506fa88/specifyweb/workbench/upload/upload_table.py#L683-L695

Further, that skipped test is also reenabled in BE. 

So, it might make sense to revisit this later on too (since, backend will be able to handle it more naturally)

emenslin (changes requested):
- [x] Ensure you can disambiguate related records when there is more than one potential match
    - [x] Ensure that matched records do not count as "New Cells"
    - [x] Create a record set with the uploaded results, and ensure that all data is uploaded correctly
    - [x] Ensure that all created records are deleted and all independent relationships (for example,  CO -> cojo if assigning an existing CO to a new/existing COG) are reverted to their state prior to the upload
  - [x] Ensure CollectionObjects can be added to existing COGs via the WorkBench
  - [x] Ensure that COGs can be created as part of the upload process for a CO and that COs can be associated with 
  - [x] Ensure a COJO can not be created with both a childCo and childCog
  - [x] Ensure a COJO can not be created without a childCo and childCog
  - [ ] Ensure the `precedence` of the COJO records are automatically set based on the order they are added

--- 

Can successfully disambiguate now and errors are caught with COJO! 
Screenshot 2025-01-21 101108

I know this is a little different so maybe this won't be fixed here but sometimes I get a validation error (that you physically have to press close on) but then it says it successfully validates, this could be just a data set error but I'm not sure. 

https://github.com/user-attachments/assets/b8ccc85c-869c-4de6-b2ae-b9d65245c63f


Not sure if precedence is being set correctly, when querying it shows mutliple children under the same COG with a precedence of zero but I'm not sure if this has to do with it being an existing COG or not. 

image

melton-jason:
> I know this is a little different so maybe this won't be fixed here but sometimes I get a validation error (that you physically have to press close on) but then it says it successfully validates, this could be just a data set error but I'm not sure. 
> 
> https://github.com/user-attachments/assets/b8ccc85c-869c-4de6-b2ae-b9d65245c63f

Were the mappings for this data set from before the recent changes? You will get the error message shown in the video when the upload plan contains either `CO -> cojo` or `COG -> cojo` as a `toOne` relationship. 

Is this the [New Data Set Thu Jan 16 2025 (2)](https://ojsmnh20241003wb-issue-5418.test.specifysystems.org/specify/workbench/279/)  in the Fossil Invertebrates collection for user spadmin? 
I'm not able to recreate the issue, and the [upload_plan.json](https://github.com/user-attachments/files/18495320/upload_plan.json) is correct for that data set: you theoretically should never encounter the error with the data set given the current upload plan. 

For testing the changes after https://github.com/specify/specify7/pull/6073#issuecomment-2603804804, I would recommend creating a new Data Set, or if working on pre-existing data sets: unmap and remap any `COG -> cojo` and `CO -> cojo` mappings. 

I've updated the example `upload_plan.json` in the testing instructions. 

If you _do_ encounter the same problem again, can you include the upload plan of the Data Set in the message? (You can copy the contents shown in `Tools -> Upload Plan` from the Data Set view). 

---

> Not sure if precedence is being set correctly, when querying it shows mutliple children under the same COG with a precedence of zero but I'm not sure if this has to do with it being an existing COG or not. 
> 
> image

Do you know if these children were associated with the parent CollectionObjectGroup while on this branch? 

Precedence should be set for any newly created or saved CollectionObjectGroupJoin records on this branch. The change is not applied retroactively which means that any existing CollectionObjectGroupJoin records will not have their precedence set until they are saved. 

Additionally, the precedence is automatically _**only**_ set if a precedence is not explicitly set for the CollectionObjectGroupJoin record being saved: this means that if the user manually specifies each CollectionObjectGroupJoin's precedence (e.g., setting them all to 0), then it's possible for there to be multiple CollectionObjectGroupJoin records with the same precedence in the same parent. 
Do you think we should be strict about ensuring a unique precedence for a CollectionObjectGroupJoin record?

emenslin:
I cannot recreate the issue, but I know I had updated the mappings before testing that data set so I'm not sure what happened. I did notice though that the current workbench naming convention makes uploading COGs pretty confusing. This is probably out of the scope of this pr but is there anything we can do to fix this? I just feel like this isn't super clear 
image

Upload plan: 
[New Data Set Tue Jan 21 2025 (2).json](https://github.com/user-attachments/files/18496433/New.Data.Set.Tue.Jan.21.2025.2.json)

I also noticed that querying on the COG table then going to collection object group join -> parent -> name, the heading is 'Arent' instead of 'Parent'

image

It does seem like precedence is being correctly set for newly created COGs so maybe someone else edited some of the ones with multiple zeros that I was seeing. Based on how it's been described to me, I think it would make more sense to have unique precedence for each CollectionObjectGroupJoin record, however, I can see letting users decide what they want the precedence to be, including letting multiple have the same number. @grantfitzsimmons what do you think? Will users want to have the same precedence for multiple CollectionObjectGroupJoin records or will they want a unique one for each?

emenslin (changes requested):
See above comment

CarolineDenis:
@emenslin, 

> I also noticed that querying on the COG table then going to collection object group join -> parent -> name, the heading is 'Arent' instead of 'Parent'

this is fixed in prod

grantfitzsimmons:
>@grantfitzsimmons what do you think? Will users want to have the same precedence for multiple CollectionObjectGroupJoin records or will they want a unique one for each?

They will likely want something unique, but I don't think we should rule out the possibility of having the same precedence.

melton-jason:
Thanks for your contributions to the PR everyone! We will merge this PR for now, so here's a comment to wrap things up: 

> I also noticed that querying on the COG table then going to collection object group join -> parent -> name, the heading is 'Arent' instead of 'Parent'

From https://github.com/specify/specify7/pull/6073#issuecomment-2605764839

I think the Issue you found probably has the same cause as #6099, so I have re-opened the Issue and left your findings as a comment: https://github.com/specify/specify7/issues/6099#issuecomment-2607870516

> I did notice though that the current workbench naming convention makes uploading COGs pretty confusing. This is probably out of the scope of this pr but is there anything we can do to fix this? I just feel like this isn't super clear

Could you expand more on what you mean by this? Do you think the actual labels for the fields/relationships should be changed, the auto-generated column names for some mapping path should be changed, or something else entirely? Regardless, it will need to be a new Bug/Feature Request: anything to formalize/specify the requirements of the expected behavior is beneficial!

sharadsw (approved):

acwhite211 (approved):

PR #6093, opened 2025-01-16
Define 0022_ensure_default_cots to be revertible

### LINKED ISSUE #6092: The 0022_ensure_default_cots migration is not currently revertable
The 0022_ensure_default_cots needs to be made revertible.

The tricky thing about reverting this migration is deciding which default COTs to delete exactly.  Since it it creating default COTs for any collection that doesn't have one, delting all default COTs would be the straight forward thing to do.  The problem is that the collection tables that this migration is fixing already have some default COTs defined, so when reverting, there is no way to know which already existed before the migration and which existed after the migration.
The two options for the revert function is to delete all default COTs, or to do nothing.  The original 0002_geo migration didn't need to do anything because reverting just deleted the table. https://github.com/specify/specify7/blob/6ca833fe4f1af8925e3f73037c7c4d0e8e88a679/specifyweb/specify/migrations/0002_geo.py#L39-L41

If they were already fine before the migration, do we want to allow reverting back to a prior migration and having no default COTs? Either solution, the default COTs would be fixed after running all the migrations.

**To Reproduce**

I the specify 7 container, run:
```bash
python manage.py migrate; #migrate
python manage.py migrate specify 0021; #revert
python manage.py migrate; #migrate again
```

acwhite211:
Fixes #6092

Add a reverse migration function to 0022_ensure_default_cots.

The revert migration does the following:
1. Set all collection records to have a null collectionobjecttype.
2. Set all the collection records to have a null collectionobjecttype.
3. Delete all CollectionObjectType records.

### Checklist

- [x] Add automated tests
- [x] Add a reverse migration if a migration is present in the PR

### Testing instructions

- In the specify 7 container, run `python manage.py migrate; #migrate`, see that all migrations run correctly
- In the specify 7 container, run `python manage.py migrate specify 0021; #revert`
- [x] See that the 0022 migration has been reverted `python manage.py showmigrations`
- In the specify 7 container, run `python manage.py migrate; #migrate again`
- [x] See that all the default COTs exist by checking that each collection record has an associated COT in its CollectionObjectType foreign key field.

### DIFF
--- a/specifyweb/specify/migrations/0022_ensure_default_cots.py
+++ b/specifyweb/specify/migrations/0022_ensure_default_cots.py
@@ -7,7 +7,17 @@ class Migration(migrations.Migration):
         create_default_collection_types(apps)
 
     def revert_default_collection_types(apps, schema_editor):
-        pass
+        # Set all collection records to have a null collectionobjecttype
+        Collectionobject = apps.get_model('specify', 'Collectionobject')
+        Collectionobject.objects.all().update(collectionobjecttype=None)
+
+        # Set all the collectionobject records of the collection
+        # NOTE: Do we want to do this?
+        Collectionobject.objects.update(collectionobjecttype=None)
+
+        # Delete all CollectionObjectType records
+        Collectionobjecttype = apps.get_model('specify', 'Collectionobjecttype')
+        Collectionobjecttype.objects.all().delete()
 
     dependencies = [
         ('specify', '0021_update_hidden_geo_tables'),
--- a/specifyweb/specify/migrations/0022_ensure_default_cots.py
+++ b/specifyweb/specify/migrations/0022_ensure_default_cots.py
@@ -11,9 +11,9 @@ class Migration(migrations.Migration):
         Collectionobject = apps.get_model('specify', 'Collectionobject')
         Collectionobject.objects.all().update(collectionobjecttype=None)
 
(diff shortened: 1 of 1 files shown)


sharadsw (approved):

CarolineDenis (approved):

grantfitzsimmons (approved):
> - [x] Add automated tests

🤨 


> - [x] See that the 0022 migration has been reverted `python manage.py showmigrations`
> - In the specify 7 container, run `python manage.py migrate; #migrate again`
> - [x] See that all the default COTs exist by checking that each collection record has an associated COT in its CollectionObjectType foreign key field.

Looks good! :+1:

``` bash
specify@47db5c1fd56f:/opt/specify7$ ve/bin/python manage.py migrate specify 0020
Operations to perform:
  Target specific migration: 0020_add_tectonicunit_to_pc_in_schema_config, from specify
Running migrations:
  Unapplying specify.0022_ensure_default_cots... OK
  Unapplying specify.0021_update_hidden_geo_tables... OK
specify@47db5c1fd56f:/opt/specify7$ ve/bin/python manage.py migrate specify
Operations to perform:
  Apply all migrations: specify
Running migrations:
  Applying specify.0021_update_hidden_geo_tables... OK
  Applying specify.0022_ensure_default_cots... OK
```

All COTs were created correctly.

PR #6142, opened 2025-01-22
Add default query mapping path to filter by COT tree def in QB search

### LINKED ISSUE #6098: QCBX query builder tool can add a Determination from the wrong tree for the COT
**Describe the bug**
Using the QB from the determination QCBX, the user can add a determination from a tree that is not appropriate to the collection object's COT.

**To Reproduce**

1. Create a new CO and set a COT
2. Click the Search button next to the Determination QCBX
3. Query for a Taxon from a tree that does not match the COT and click Select
4. Save the CO record and see that the determination is still there

**Expected behavior**
Either the QB search results should be filtered or the user should be prevented from saving the determination (save blocker or clearing the determination subview).

**Screenshots**

https://github.com/user-attachments/assets/9edd7a2a-13c3-4dbc-8185-2b235c35eaa4

[Specify 7 System Information - 2025-01-16T20_41_19.089Z.txt](https://github.com/user-attachments/files/18445120/Specify.7.System.Information.-.2025-01-16T20_41_19.089Z.txt)

- OS: Mac
- Browser: Chrome
- Database Name: 
geo_2024_12_05_1_production
- Collection name: Geology
- User Name: spadmin
- URL: https://geo202412051production-production.test.specifysystems.org/specify/view/collectionobject/835/

sharadsw:
Fixes #6098

### Checklist

- [ ] Add automated tests

### Testing instructions

- Create a new CO and set a COT
- Click the Search button next to the Determination QCBX
- [ ] Verify there is already a mapping line for filtering by COT tree def

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
@@ -550,6 +550,10 @@ export function QueryComboBox({
                                         : fieldName === 'taxonTreeDefId'
                                           ? {
                                               field: 'definition',
+                                              queryBuilderFieldPath: [
+                                                'definition',
+                                                'id',
+                                              ],
                                               isRelationship: true,
                                               operation: 'in',
                                               isNot: false,
--- a/specifyweb/frontend/js_src/lib/components/SearchDialog/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/SearchDialog/index.tsx
@@ -35,7 +35,7 @@ import type { QueryFieldFilter } from '../QueryBuilder/FieldFilter';
 import { queryFieldFilters } from '../QueryBuilder/FieldFilter';
 import { QueryFieldSpec } from '../QueryBuilder/fieldSpec';
 import { QueryBuilder } from '../QueryBuilder/Wrapped';
-import { MappingPath } from '../WbPlanView/Mapper';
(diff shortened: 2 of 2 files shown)


CarolineDenis (approved):

melton-jason (approved):
Nice work! 
Since the `queryBuilderFieldPath` overrides the `field`, this can probably be simplified by removing the `queryBuilderFieldPath` from the QueryComboBoxFilter type and instead allowing the `field` to be an Array of strings (MappingPath). 

e.g., change 
https://github.com/specify/specify7/blob/0c9839d5c6172188f6d09bc4041aac78aabcbd3b/specifyweb/frontend/js_src/lib/components/SearchDialog/index.tsx#L44

to 

```ts
readonly field: MappingPath | (string & (keyof CommonFields | keyof SCHEMA['fields']));
```

(And handling the case when `field` is an Array in `toQueryFields`)

emenslin (commented):
- [x] Verify there is already a mapping line for filtering by COT tree def

Just want to check this is the correct behavior before approving but all ranks show up even if they are not part of the tree you are querying on. This isn't necessarily a problem but it could just potentially be confusing so I wanted to double check. 

https://github.com/user-attachments/assets/9e13b84e-21f8-418e-b0c2-92fb8e7adff6

sharadsw:
@emenslin Yeah that is expected. The mapping line should ensure Taxon that don't belong to the COT don't show up in the results

PR #6148, opened 2025-01-23
Clear cache when changing collection

### LINKED ISSUE #6116: Automatically clear the cache when the user switches collections or logs out
We have encountered several significant issues due to how Specify caches information from the stored `/context/system_info.json` endpoint.

A long-running consequence of us not doing this is #3160, and more recently this became a big deal with #6096 

Neither of these issues are resolved yet, but @melton-jason has a simple solution:

To fix these issues, the cache should be cleared when:

-  A user chooses a collection.
-  A user logs out.

> The Issue can come up WAY more frequently than you might think if the database has at least one geo discipline and one non-geo discipline. 
> 
> Due to the caching behavior, you can even easily recreate #6064! 
> 
> See my below videos recreating both this Issue and #6064, both times in a new incognito tab (nothing initially cached). 
> 
> **Logging into non-geo/paleo discipline prevents it from showing for paleo/geo collections**
> 
> https://github.com/user-attachments/assets/4551e1df-339f-45aa-b6c0-547ed49d59ae
> 
> **Logging into geo/paleo disciplines shows Age for non-paleo/geo collections**
> 
> https://github.com/user-attachments/assets/d8bac53b-5172-417d-ba1e-30f7289aadf7
> 
> What's worse is that you need to clear your cache and/or log out multiple times for the value to completely clear from the cache! 
> 
> --- 
> 
> The Issue here is that to determine the discipline type, Specify fetches the information from the stored `/context/system_info.json` endpoint
> 
> https://github.com/specify/specify7/blob/181c2de53d2d457ff13e5a1532527e1fc2f2422d/specifyweb/frontend/js_src/lib/components/WbPlanView/navigator.ts#L545-L546
> 
> https://github.com/specify/specify7/blob/181c2de53d2d457ff13e5a1532527e1fc2f2422d/specifyweb/frontend/js_src/lib/components/InitialContext/systemInfo.ts#L30-L60
> 
> The Client stores its own value of the endpoint value _along_ with caching the url: this is a problem with the application, not a generic browser cache Issue. 
> 
> I think a fine solution to the problem is to clear the application and browser cache whenever: 
> - the user chooses a collections
> - the user logs out
> 
> This would also fix a lot of other cache-related Issues as well! 
> 
> An implementation for clearing the cache when a user chooses/switches collection can theoretically be as simple as modifying the following from: 
> 
> https://github.com/specify/specify7/blob/9563d8854f9af014ae069f52b0b12382762467e5/specifyweb/frontend/js_src/lib/components/RouterCommands/SwitchCollection.tsx#L38-L43
> 
> to: 
> 
> ```js
>       async () =>
>         ping('/context/collection/', {
>           method: 'POST',
>           body: collectionId!.toString(),
>           errorMode: 'dismissible',
>         })
>           .then(clearAllCache)
>           .then(() => globalThis.location.replace(nextUrl)),
> ```
>  

 _Originally posted by @melton-jason in [#6096](https://github.com/specify/specify7/issues/6096#issuecomment-2599081823)_

CarolineDenis:
Fixes #6116

### Checklist

### Testing instructions

- Choose a db that has a geo/paleo collection and a non geo/paleo collection 
- Open SP with any collection
- Open the query builder with CO table
- [ ] Verify that virtual field age is present if geo Coll and not present if not geo 
- change the collection 
- [ ] Verify that virtual field age is present if geo Coll and not present if not geo 
- [ ] Verify that virtual field is present only when needed when you log out than login to new collection instead of using change collection tool

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ Django==3.2.15
 mysqlclient==2.1.1
 SQLAlchemy==1.2.11
 requests==2.32.2
-pycryptodome==3.15.0
+pycryptodome==3.21.0
 PyJWT==2.3.0
 django-auth-ldap==1.2.15
 jsonschema==3.2.0
--- a/specifyweb/frontend/js_src/lib/components/Core/Main.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Core/Main.tsx
@@ -25,6 +25,7 @@ export type MenuItem = {
   readonly enabled?: () => Promise<boolean> | boolean;
   readonly icon: JSX.Element;
   readonly name: string;
+  readonly onClick?: () => Promise<void>;
 };
 
 /*
--- a/specifyweb/frontend/js_src/lib/components/Header/UserTools.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Header/UserTools.tsx
@@ -18,6 +18,7 @@ import { userInformation } from '../InitialContext/userInformation';
 import { Dialog, LoadingScreen } from '../Molecules/Dialog';
 import { OverlayContext } from '../Router/Router';
 import { locationToState } from '../Router/RouterState';
+import { clearAllCache } from '../RouterCommands/CacheBuster';
 import { MenuButton } from './index';
 import { useUserTools } from './menuItemProcessing';
 
@@ -144,6 +145,18 @@ function UserToolsColumn({
               const isExternalLink = isExternalUrl(url);
               // Make links to another entrypoint trigger page reload
               const LinkComponent = isExternalLink ? Link.NewTab : Link.Default;
+
+              const handleClick = async (): Promise<void> => {
+                if (url === '/accounts/logout/') {
+                  await clearAllCache();
+                }
+              };
+      
+              const handleOnClick = ():void => {
+                handleClick().catch((error) => {
+                  console.error('Error occurred during cache clearing:', error);
+                });
+              };
               return (
                 <li key={url}>
                   <LinkComponent
@@ -154,6 +167,7 @@ function UserToolsColumn({
                         : undefined
                     }
                     href={url}
+                    onClick={handleOnClick}
                   >
                     {icon}
                     {title}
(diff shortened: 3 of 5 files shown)


sharadsw (approved):

pashiav (changes requested):
- [ ] Verify that virtual field age is present if geo Coll and not present if not geo 
- [ ] Verify that virtual field age is present if geo Coll and not present if not geo 
---
Age field is not present for initial opening of Specify on a paleo collection:
https://calvertmarinemuseum20241114-issue-6116.test.specifysystems.org/specify/
1. Log in as spadmin, collection: `Vertebrate` _(type: vertpaleo)_ -> open QB on CO table
	- See that there **is not** an Age filter.
2. Switch collection to `Invertebrate` _(type: invertpaleo)_-> open QB on CO table
 	- See that there **is** an Age filter.
3. Do steps 1 and 2 again but start with collection `Invertebrate`, then switch to `Vertebrate`.
	- See that there **is not** an Age filter for `Invertebrate` now.
	- See that there **is** an Age filter for `Vertebrate` now.

**Video:**

https://github.com/user-attachments/assets/2280dae8-3239-44a6-b765-92c8997c351b


<br><br>

Also, if you refresh, the Age field disappears from paleo dbs:

https://github.com/user-attachments/assets/f5c52bcc-9bbf-4c89-a5b0-aea1a2db3a67

acwhite211 (approved):

alesan99 (dismissed):
- [ ] Verify that virtual field age is present if geo Coll and not present if not geo
- change the collection
- [ ] Verify that virtual field age is present if geo Coll and not present if not geo
 ---
 I can confirm a cache clear is being triggered upon switching collections.
 Age is only present when appropriate and the about Specify dialog appears to keep up as well. :+1::+1:
But I can recreate @pashiav's issue. The cache isn't cleared when logging out and switching collections through the initial dialog.

grantfitzsimmons (approved):
- [X] Verify that virtual field age is present if geo Coll and not present if not geo 
- [X] Verify that virtual field age is present if geo Coll and not present if not geo 
- [X] Verify that virtual field is present only when needed when you log out than login to new collection instead of using change collection tool

Issue #5322, opened 2024-10-10
`Taxon does not belong to the same tree as this Object Type` despite taxon being in the tree corresponding to the object type


grantfitzsimmons:
**Describe the bug**

This was reported by Chris at the Gothenburg museum of natural history. Test panel database is `gnm_2024_10_10`

> Two of our users (kareb1, carho18) encounter the same issue while adding records in the collection GNM Mollusca_Ovriga.
> 
> After logging in, the first registration saves normally. When creating a new record, though, the taxon field locks with an error message, then continues to be locked even when we try searching for a new species. Nothing seems to be wrong in the taxon tree.
> 
> After a logout/login it works for one record again.
>
> | ID      | Name                              | CollectionObjectTypeID |
> |---------|-----------------------------------|-------|
> | 589824  | GNM Mollusca_Ovriga              | 18    |
> 
> - Those users can without errors adding records in other collections.
> - Other users can without errors adding records in GNM Mollusca_Ovriga
> - I cannot reproduce the error in my test environment (database copy from October 1) using one of these user’s credentials.
> - It may be that this error first occurred after October 1. We receive a db backup file to our Azure on the 1:st every month, so I have not got a fresh db to check with.

This is in the `gnm` database, where the `CollectionObjectTypeID` is present at the collection level in all cases:

| ID      | Name                              | CollectionObjectTypeID |
|---------|-----------------------------------|-------|
| 4       | GNM Entomology                    | 1     |
| 32769   | GNM Acanthocephala               | 2     |
| 65536   | GNM Cestoda                      | 3     |
| 98304   | GNM Hydrozoa                     | 4     |
| 131072  | GNM Porifera                     | 5     |
| 163840  | GNM Protozoa                     | 6     |
| 229377  | Generalkatalogen                 | 7     |
| 262145  | Värddjur                         | 8     |
| 294912  | GNM Trematoda                    | 9     |
| 327680  | GNM Nematoda                     | 10    |
| 360448  | GNM Crustacea                    | 11    |
| 393216  | GNM Turbellaria                  | 12    |
| 425984  | GNM Anthozoa                     | 13    |
| 458752  | GNM Gastropoda                   | 14    |
| 491520  | GNM Polychaeta                   | 15    |
| 524288  | GNM Gephyrea                     | 16    |
| 557056  | GNM Echinodermata                | 17    |
| 589824  | GNM Mollusca_Ovriga              | 18    |
| 655360  | GNM Cephalopoda                  | 19    |
[...]

Issue #5299, opened 2024-10-01
Save blocker prevents user from selecting valid taxon from querycombobox in Collection Object form


mpitblado:
**Describe the bug**
When searching for a taxon in a querycombobox from the Collection Object table in Data Entry, an error appears when a valid taxon is selected, preventing the user from being able to save the record.

**To Reproduce**
1. Click on Data Entry
2. Select Collection Object
3. Scroll down to the Determinations subview
4. Enter any taxonomic identification into the Taxon querycombobox
5. Notice that the field will turn red, an error message will appear, and the save button for the Collection Object will be disabled.

**Expected behavior**
The user is able to enter a taxonomic identification that is a node in the tree and click save.

**Screenshots**
![image](https://github.com/user-attachments/assets/0ead89fd-351a-4878-a9fa-ad91eaff07e8)

**Crash Report**

N/A

Please, also fill out the following information manually:

- OS: Windows 10
- Browser: Firefox 130.0.1
- Specify 7 Version: 7.9.7
- Database Name: beaty
- Collection name: Birds, mammals, vascular plants tested. Assumed to affect all collections

**Reported By**
Beaty Biodiversity Museum

**Additional context**
This appears to be a regression with the 7.9.7 release. May be resolved by 572c2493da5c4773ab375cd71ac81109cdf5aebd

grantfitzsimmons:
I can recreate:
<img width="1030" alt="image" src="https://github.com/user-attachments/assets/4a704ee4-a95e-438b-a565-b92a6b9ff136">

grantfitzsimmons:
![image](https://github.com/user-attachments/assets/8246efd1-4be7-47a9-8775-1d5cce2f9127)

This is a migration issue

sharadsw:
Save blocker has been removed in #5248

PR #6175, opened 2025-01-29
Block save on invalid determinations

### LINKED ISSUE #6097: Can have multiple determinations with different COTypes
**Describe the bug**
You can set the COT, add a determination, switch the COT, add another determination, and save and both determinations stay on the form even though they are from different COTs. Only the one that matches the COT will actually be uploaded and appear in the tree but both stay on the forms without error. 

**To Reproduce**

1. Go to any collection with multiple COT (or make one) 
2. Add a determination 
3. Fill out taxon
4. Add a blank determination 
5. Switch COT 
6. Fill out taxon 
7. Save 
8. See both determinations saved 

**Expected behavior**
Should not be able to have a CO with determinations from different COTs. 

**Screenshots**

https://github.com/user-attachments/assets/53a4491c-46f7-4117-90db-bb8276ea0c89

If you add a determination, then switch the COT, then try to add another determination, and save it removes both determinations: 

https://github.com/user-attachments/assets/e44dced7-c440-4709-abec-255a16173635

- OS: Windows 11
- Browser: Chrome
- Database Name: ojsmnh
- User Name: spadmin
- URL: https://ojsmnh20241003wb-production.test.specifysystems.org/specify/view/collectionobject/228317/

sharadsw:
Fixes #6097

There isn't a way right now to trigger an interactive dialog through business rules (which is where we track field changes) and so I've added a save blocker to Determination -> Taxon if the COT changes and determinations become invalid. Since we have a working Taxon QCBX filter and all migrations work fine (🤞) the save blocker should be a lot less strict and we shouldn't be seeing issues like #5299 and #5322 again. 

The implementation this time around is also different - we're checking validity of determinations only when COT changes instead of checking it when adding a Taxon, so ideally it should catch all possible cases of having an invalid determination

### Checklist

- [x] Add automated tests

### Testing instructions

- Create a CO
- Add a valid determination
- Change COType
- [ ] Verify there's a save blocker on invalid determination(s)
- Test different ways of resolving the save blocker
  - Enter a valid Taxon for the new COType
    - [ ] Verify save blocker is removed
  - Change back to original COType
    - [ ] Verify save blocker is removed

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
@@ -131,6 +131,41 @@ describe('Collection Object business rules', () => {
   };
   overrideAjax(otherCollectionObjectTypeUrl, otherCollectionObjectType);
 
+  test('CollectionObject -> determinations: Save blocked when a determination does not belong to COT tree', async () => {
+    const collectionObject = getBaseCollectionObject();
+    collectionObject.set(
+      'collectionObjectType',
+      getResourceApiUrl('CollectionObjectType', 1)
+    );
+
+    const determination =
+      collectionObject.getDependentResource('determinations')?.models[0];
+
+    const { result } = renderHook(() =>
+      useSaveBlockers(determination, tables.Determination.getField('Taxon'))
+    );
+
+    await act(async () => {
+      await collectionObject?.businessRuleManager?.checkField(
+        'collectionObjectType'
+      );
+    });
+    expect(result.current[0]).toStrictEqual([
+      resourcesText.invalidDeterminationTaxon(),
+    ]);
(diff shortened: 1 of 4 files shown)


CarolineDenis (approved):

alesan99 (approved):
- [x] Verify there's a save blocker on invalid determination(s)
  - Enter a valid Taxon for the new COType
    - [x] Verify save blocker is removed
  - Change back to original COType
    - [x] Verify save blocker is removed 
---

Working good 👍👍👍

I want to note a slight inconsistency I found though:
You can save it normally, and you can erase the Taxon on an existing record, but upon switching the type it will no longer let you make it empty.

https://github.com/user-attachments/assets/2f6d2034-de12-4399-9589-7bd2c1d5d7cd

acwhite211 (approved):

PR #6194, opened 2025-02-04
Add workbench validation for cotype

### LINKED ISSUE #6187: WB does not always match COT to determination
**Describe the bug**
Uploading a CO data set to workbench with multiple trees can sometimes led to CO with a COT and determination that do not match. No error is thrown and there is seemingly nothing wrong with the uploaded records (no unload protect, save blocker, etc.) but the determination is part of another tree. I didn't get a chance to do a whole ton of testing so there might be other weird behavior with uploading CO/COT/determinations in workbench that I didn't notice.  

**To Reproduce**

1. Go to workbench 
2. Select CO as base table 
3. Add mappings for cat #, COT, and to at least one rank from two different trees (example upload plan: [COT test.json](https://github.com/user-attachments/files/18625464/COT.test.json))
4. Add an existing taxon from a tree that is not the same as the COT 
5. Validate 
6. See no errors are thrown 
7. Upload 
8. See record looks normal 
9. Query/go to tree and see taxon is from a different tree 

**Expected behavior**
Some sort of error should be thrown in the cell in workbench so you cannot upload without fixing it. There might be a better solution because I know errors aren't always ideal but this should definitely be caught before upload. 

**Screenshots**

Tree for each COT: 
Image

Taxon trees: 

https://github.com/user-attachments/assets/c246cea3-adb0-4433-8e24-367248ea976b

Actual upload: 

https://github.com/user-attachments/assets/db6e7191-4187-4ed9-9e73-7f7b65e6a452

- OS: Windows 11
- Browser: Chrome
- Database Name: ojsmnh
- User Name: spadmin
- URL: https://ojsmnh110242024production-production.test.specifysystems.org/specify/workbench/101/

**Additional context**
I was not able to check everything but some other COT behavior in workbench is that; if there is no COT, all taxa are uploaded to the default tree (expected behavior as far as I know), if there’s only one tree but multiple COTs, it will upload taxon to whatever tree is mapped no matter what the type is, this is also a problem but if it needs to be a separate issue let me know. 

https://github.com/user-attachments/assets/712e3e82-1f5c-49fd-9fd9-5c4c9227b4b3

sharadsw:
Fixes #6187

### Checklist

- [ ] Add automated tests

### Testing instructions

- Create a WB dataset that maps COType and Taxons
- Add a row where COType tree definition does not belong to the Taxon in that row
- Validate
- [ ] Verify there is a validation error for entered COType

### DIFF
--- a/specifyweb/workbench/upload/treerecord.py
+++ b/specifyweb/workbench/upload/treerecord.py
@@ -335,6 +335,31 @@ class ScopedTreeRecord(NamedTuple):
 
     def rescope_tree_from_row(self, row: Row) -> Tuple["ScopedTreeRecord", Optional["WorkBenchParseFailure"]]:
         """Rescope tree from row data."""
+        
+        def validate_trees_with_cotype(row: Row, treedefs_in_row: Set[int]):
+            # TODO: Need a better way to do this
+            # Find a way to send cotype column when ScopedTreeRecord instance is created?
+            COL_NAME = "Type"
+            def find_cotype_in_row(row: Row):
+                return row.get(COL_NAME, None)
+            
+            def get_cotype_tree_def(cotype_name: str):
+                # TODO: Need to scope by collection?
+                cotypes = models.Collectionobjecttype.objects.filter(name=cotype_name)
+                return cotypes[0].taxontreedef.id if len(cotypes) > 0 else None
+        
+            cotype_value = find_cotype_in_row(row)
+            if not cotype_value: return None
+
+            cotype_treedef = get_cotype_tree_def(cotype_value)
+            if not cotype_treedef: return None
+
+            # Check only the first treedef assuming all ranks belong to same tree
+            # Validation for multiple ranks is done elsewhere
+            if len(treedefs_in_row) > 0 and cotype_treedef == list(treedefs_in_row)[0]:
+                return None
+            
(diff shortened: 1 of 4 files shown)


specifyweb/workbench/upload/treerecord.py
```diff
@@ -420,16 +445,10 @@ def check_root(root):
...
-
-        # Determine the target treedef based on the columns that are not null
-        targeted_treedefids = {rank_column.treedef_id for rank_column in ranks_columns_in_row_not_null}
-        if not targeted_treedefids:
-            # return self, WorkBenchParseFailure('noRanksInRow', {}, None)
-            return self, None
-        elif len(targeted_treedefids) > 1 and len(unique_treedef_ids) > 1:
-            logger.warning(f"Multiple treedefs found in row: {targeted_treedefids}")
-            error_col_name = list(ranks_columns_in_row_not_null)[0].column_fullname
-            return self, WorkBenchParseFailure("multipleRanksInRow", {}, error_col_name)
```
sharadsw, line 432: Looks like this is a duplicate of what `handle_multiple_or_no_treedefs` does

alesan99 (approved):
### Testing instructions
- [x] Verify there is a validation error for entered COType
---

image

The validation error is being shown when appropriate 👍👍👍

CarolineDenis (approved):

Issue #6126, opened 2025-01-21
Batch Edit: Batch edit for relationships


sharadsw:
Component of https://github.com/specify/specify7/pull/4929

There are 4 different types of relationships, in general. 
    - To-one dependent (for ex. collectionobjectattribute),
    - To-one independent (for ex. Cataloger, CollectingEvent [when not embedded])
    - To-many dependent (for ex. determinations, preparations),
    - To-many independent (for ex. None for CO as the base table)

- ### Fields
    Modifying any field is possible, other than nodenumber, fullname, and highestchildnodenumber. If some fields get updated, only those fields are highlighted
- ### To-one dependent (for ex. collectionobjectattribute)
    These relationships get directly updated, and are not matched. If the to-one is not in the db, it'll create one. 
    This also includes collectingevent when embedded. <br/>
    <u>Test cases to consider</u>:
    - When mapped, the record is directly updated.
    - When mapped, if the record is not present, it'll be created, if not null values are present.
    - If the record previously had values, and the values are removed (making the cells completely empty), the to-one dependent record will be deleted. Since it is possible that there may be other fields in the database (but not in the query), we may accidentally delete the record. Eample: user selected collectionobject -> collectionObjectAttribute -> remarks. And say they set remarks to empty in the spreadsheet, it is is possible that integer1 field in collectionObjectAttribute may have some value. To prevent accidental deletion, by default, we look at all the fields in the database for that record (other than system fields), to determine whether we can delete the record or not. This behaviour is controlled by a remote preference. (described in a later section):

- ### To-many dependent (for ex. determinations)

    Same as to-one dependent. These relationships get directly updated. If the corresponding record is not present, a new one gets created. 

    <u>Test cases to consider</u>:
    - When mapped, the record is directly updated.
    - When mapped, if the record is not present, it'll be created, if not null values are present.
    - If the cell data is removed, and if every other field is empty in the database (can be disabled via a preference), the record will be deleted.

- ### To-one independent (for ex. cataloger)

    These relationships get matched, and uploaded (if match is not found). During upload, it performs a clone of the record (cloning all the non-unique fields, and dependents). The clone takes into account relationships also mapped. That is, if agent needs to be cloned, and you have mapped agentspecialty, it'll take the agentspecialty mapped (rather than cloning previous's agentspecialty).

    <u>Test cases to consider:</u>

[...]

PR #6196, opened 2025-02-05
Batch edit for multiple trees

### LINKED ISSUE #6011: Add support for querying on any tree table field from a chosen rank
# Refer to https://github.com/specify/specify7/discussions/5935 for more details

<sup>Originally posted by **grantfitzsimmons** March 28, 2024</sup>
Currently in Specify 6 and 7 there is no ability in the query builder to search on any arbitrary field in the tree table once you select a specific rank.

#### At (any rank)

See that all fields and relationships are exposed in the list of fields to choose from

image

The `stringid`s are constructed as you'd expect:

```
1,9-determinations,4-preferredTaxon.taxon.fullName
```

#### At {Rank} level

Only `author` and `fullName` can be chosen from the list of fields once you select a level in the tree.

image

The `stringid`s associated with these rank-specific query items are unusual. They are formatted as follows:

```
1,9-determinations,4-preferredTaxon.taxon.Kingdom
1,9-determinations,4-preferredTaxon.taxon.Kingdom Author
```

It looks like there is special logic handling these `treeNodeCols` when constructing a string ID for these items.

Currently, users can only choose two fields at any given tree rank:
* Author
* Full Name

Users should be able to select any rank in a tree (e.g. Family, Genus, Species, etc.) and select any field in the tree table at that level.

Instead of seeing only these two fields:
image

You should see this:
image

The list of fields should be comprehensive, including `Author`, `Common Name`, `Full Name`, `GUID`, `Is Preferred`, `Name`, `Protected Status`, `Remarks`, `Source`.

**Reported By**
Louisiana State University Museum of Natural Science, University of Washington,  University of Florida, CSIRO, University of Kansas, many others

---

Initially reported in #2364 

@grantfitzsimmons:
> Currently in Specify 7 and 6 you are restricted to only seeing the taxon full name and author for each rank. This restriction causes issues when it comes to running a query from the collection object or creating reports and labels where you want to reference a specific field for a specific rank.
> 
> In this case, I am reworking the University of Washington's labels to reference the family number, stored in a text field at the family level. 
> 
> There is no way to reference this number without using relationships through the "Parent Taxon" several times and hoping that the parent is the family. For their database, all specimen records are determined to species at the moment, but this is going to change in the future and the flexibility is necessary. I have to stack parent relationships to ensure that it will not be missed.

@maxpatiiuk said the following in [this comment](https://github.com/specify/specify7/issues/2364#issuecomment-1474650584):

> From what I remember Ben told me, this restriction was put in place artificially just to match the limitations of specify 6 (yeah, this is from the dark dark days where sp7 was limited just to not outcompete specify 6)
> 
> Here is the code for that:
> https://github.com/specify/specify7/blob/f4cd803a5a159785cb080f90aede319186e4dd02/specifyweb/stored_queries/queryfieldspec.py#L27-L30 (it currently allows only Author, ID and Group Number fields - and full name implicitly if field name was not provided). Modifying that code and lifting the restriction on the front-end should fix this

Last march I asked for that [code to be removed](https://github.com/specify/specify7/issues/2364#issuecomment-1477986529) but no forward movement has occurred since that discussion.

**Issues to Resolve:**
- [x] #4723
- [ ] Add support for querying on any tree table field from a chosen rank</div>

### LINKED ISSUE #6127: Batch Edit: Batch edit for (multiple) trees
Component of https://github.com/specify/specify7/pull/4929

- ### Trees

    - Workbench method:<br/>

    - Update method: <br/>

Notes from #5417: 
- Batch edit currently doesn't work for multiple trees. Relevant code to consider: https://github.com/specify/specify7/pull/5417#discussion_r1905952721
- Adding tree ranks to a query for batch edit does not work as expected. See: https://github.com/specify/specify7/pull/5417#pullrequestreview-2543815320

### LINKED ISSUE #6315: Batch Edit: Some tree rank relationship fields throw errors in the QB
**Describe the bug**
[Batch Edit for Multiple Trees](https://github.com/specify/specify7/pull/6196) adds the ability to query on any field from a tree rank, fixing #6011. Some relationship fields added here throw errors after running the query.

**To Reproduce**

1. Go to the Query Builder
2. Choose a base table that has a tree (Taxon, Storage, Geography, etc.)
3. Add a field from one of the following related tables (NOT `(formatted)` or `(aggregated)`): Definition, CreatedByAgent, ModifiedByAgent, Taxon Citations, Taxon Attributes. Most of the relationships that are hidden by default will throw an error.
4. Run the Query
5. See error

**Expected behavior**
Query results should be returned without errors.

**Screenshots**

https://github.com/user-attachments/assets/3a46517e-50b3-40cc-8023-ef2d285679ab

**Crash Report**

[Specify 7 Crash Report - 2025-03-11T18_01_37.536Z.txt](https://github.com/user-attachments/files/19195783/Specify.7.Crash.Report.-.2025-03-11T18_01_37.536Z.txt)

- OS: Mac
- Browser: Chrome
- Specify 7 Version: issue-6127
- Database Name: ojsmnh_1_10_24_2024_migrations5417
- Collection name: Fossil Invertebrates
- User Name: spadmin
- URL: https://ojsmnh110242024migrations5417-issue-6127.test.specifysystems.org/specify/query/new/taxon/

sharadsw:
Fixes #6127
Fixes #6011 
Fixes #6315

> [!WARNING]  
> This PR does not directly contain a migration but since it's based on #5417, make sure to use a dataset where #5417 was tested.

This PR enables batch editing on Tree tables and enables support for querying on all tree fields when choosing a specific rank (no longer limited to Author and Full Name). Functionally, this PR will work the same as the workbench when uploading to a tree rank. 

## Batch editing with tree as base table
- The `name` field is required when querying on a tree rank
- When choosing a specific rank (and not `any rank`), Batch Edit expects to have all ranks lower than the chosen rank in the query
  - If the user does not add the "missing" fields, they will be prompted to pick the trees they wish to batch edit
  image
    - Users can pick which trees they want in case of conflicting ranks
    - Batch edit will only add ranks from selected trees
    - If no trees are selected, batch edit will use ALL trees
  - When selected ranks do not exist in multiple trees but there are missing ranks:
  image
  - Example: Say you have the following ranks for trees `Tree 1` and `Tree 2`: 
    - `Tree 1 -> Species` and `Tree 1 -> Subspecies`
    - `Tree 2 -> Species` (lowest rank)
  - And you make a query with:
    - `Tree 1 -> Species -> name`
  - Batch edit will implicitly add missing ranks (based on which trees were selected) and augment them to the dataset. The dataset will then look like
    - `Tree 1 -> Species -> name`
    - `Tree 1 -> Subspecies -> name`
    - `Tree 2 -> Species -> name`

### Batch editing a Tree path
- When editing a tree 'path' using a Tree as a base table with Batch Edit, the records are not edited directly. Rather the record is cloned and uploaded to the tree, which means functionally it will work the same as the workbench. The implementation in this PR will be more apparent in #6283 when trying to edit tree records through a relationship. eg: Determinations (more info here: https://github.com/specify/specify7/pull/6196#issuecomment-2677584539)
image

### Editing a single node
- A single node's simple fields can be edited using a query containing a `(any rank)` query field
image

### Checklist

- [ ] Add automated tests

#### Test #6011 
- Create a query with a Tree base table (eg: Taxon, Storage, Geography, Lithostrat, Chronostrat, Tectonic Unit)
- [ ] Verify all tree fields can be queried with 

#### Test batch edit with a single tree in db
1. Edit a single node
  - Make a tree query with `(any rank)` and some other simple fields
  - Click Batch Edit
  - Edit fields
  - [ ] Test validate/commit
  - [ ] Verify edited cells were highlighted as Updated Cell
  - [ ] Test rollback

2. Edit a tree path
  - Make a tree query with multiple rank names
  - Batch edit
  - Edit fields
  - [ ] Test validate/commit/rollback
  - [ ] Verify changes are highlighted as New Cell and the tree path is a new upload to the Tree

3. Test missing ranks
  - Make a tree query with a rank that isn't the lowest rank
  - Click batch edit
  - [ ] Verify there is a missing rank dialog indicating the lower ranks will be added
  - Click continue
  - [ ] Verify missing ranks were added correctly

#### Test batch edit with multiple trees in the db
- [ ] Repeat steps for editing single node
- [ ] Repeat steps for editing a tree path
- [ ] Repeat steps for missing ranks
  - [ ] Verify missing ranks dialog allows you to select trees to edit with
  - [ ] Verify ranks of selected trees are added correctly
  - [ ] Verify ALL trees are used when no trees are selected in the dialog


specifyweb/stored_queries/query_construct.py
```diff
@@ -71,48 +72,40 @@ def handle_tree_field(self, node, table, tree_rank, next_join_path, current_fiel
...
+        cases = []
+        field = None # just to stop mypy from complaining.
+        for ancestor in ancestors:
+            field_spec = make_tree_field_spec(ancestor)
+            query, orm_field, field, table = field_spec.add_spec_to_query(query)
+            # Field and table won't matter. Rank acts as fork, and these two will be same across siblings
+            for treedefitem_param in treedefitem_params:
+                cases.append((getattr(ancestor, treedefitem_column) == treedefitem_param, orm_field))
+
+        column = sql.case(cases)
```
sharadsw, line 101: Tried to adapt the code from #4929 for multiple trees. Queries seem to work fine but would appreciate your feedback here @realVinayak

realVinayak: Sure, I'll look this over the weekend

specifyweb/stored_queries/query_construct.py
```diff
@@ -71,48 +72,40 @@ def handle_tree_field(self, node, table, tree_rank, next_join_path, current_fiel
...
-        column = sql.case([case for per_ancestor in cases_per_ancestor for case in per_ancestor])
+        treedefitem_params = []
+        for _, rank_id in treedefs_with_ranks:
+            treedefitem_param = sql.bindparam(
+                'tdi_%s' % query.param_count,
+                value=rank_id
+            )
+            treedefitem_params.append(treedefitem_param)
+            param_count = self.param_count + 1
+            query = query._replace(param_count=param_count)
```
realVinayak, line 83: Are you sure the param is needed? You could do something like

```suggestion

treedefitem_params = [treedefitem_id for (_, treedefitems) in treedefs_with_ranks for treedefitem_id in treedefitems]
```

Probably need to adjust the intention 😅.

realVinayak: Nothing should be changed below though (in the for loop).

realVinayak (commented):
RE https://github.com/specify/specify7/pull/6196#discussion_r1947258143. 
Yeah this is the correct idea! But, shouldn't be needing the params at all. I'm not sure why it was included in the initial query builder by Ben.

realVinayak:
RE: https://github.com/specify/specify7/pull/6196#issuecomment-2644232019.

Yeah, that's what I was stuck on before and was a hard problem. Actually, there are atleast two ways of approaching it:

## First way:
Allow query builder to support specific trees (related: https://github.com/specify/specify7/pull/5251). This would involve a bit of frontend change, although, I've no memory / idea of how hard that will be, best person might actually be Jason. On the backend (assuming treedef is available in queryfieldspec)

a. We still need to make the worst case assumption that the user selects a rank from any tree (so we still need to take the maximum of treedefitems). This part is already done. See https://github.com/specify/specify7/blob/7e478b967a4b2be2e5cfbc868535a574f8535c88/specifyweb/stored_queries/query_construct.py#L48-L52

b. Need to restrict this to just one tree. https://github.com/specify/specify7/blob/7e478b967a4b2be2e5cfbc868535a574f8535c88/specifyweb/stored_queries/query_construct.py#L68-L71
Simplest way will be to add a `treeid` member field in `TreeRankQuery` when parsing. Somewhere here: https://github.com/specify/specify7/blob/7e478b967a4b2be2e5cfbc868535a574f8535c88/specifyweb/stored_queries/queryfieldspec.py#L189-L194

c. Now, when say a new queryfieldspec with a treeid is encountered, the case-when block will select the correct case where the treedefitem_id is encountered. In the current implementation if say ranks from treedefs 5 and 6 are searched, then _both_ are searched. This is because of the following two places:
https://github.com/specify/specify7/blob/befd6a625df3ef0999d5df8bc14b4aea4302acd8/specifyweb/stored_queries/query_construct.py#L92-L94
https://github.com/specify/specify7/blob/befd6a625df3ef0999d5df8bc14b4aea4302acd8/specifyweb/stored_queries/query_construct.py#L139-L142
(The _or_ clause makes the magic. Although, the comment in second is outdated --- I removed model param because it was buggy).

d. By the time you get here, the tree query will work for a specific tree. Next is to make it work with batch-edit. _All_ we need to do is make a format like this (taken from https://github.com/specify/specify7/pull/5091#discussion_r1752667819).

```
"mustMatchTreeRecord": { 
   "ranks": { 
     "SomeTreeName~>Class": { 
       "treeNodeCols": { 
         "name": { 
           "matchBehavior": "ignoreAlways", 
           "nullAllowed": true, 
           "default": null, 
           "column": "Class" 
         } 
       }, 
       "treeId": 1 
     }, 
 } 
``` 
Note: I'm pretty pretty sure treeId was not necessary anymore (double check this, it makes things more simpler. We can still make things work though.). 

e. Add a function like `get_wb_name` to `TreeRankQuery`. Assuming that there is a `treeId` member field, it'll be something like (in pseudocode). **Yes, getting self.treeId (so the ID) instead of the treeName -- not a mistake! -- it makes things simpler later**
```
def get_wb_name(self): return self.treeId + "~>"  + self.name # Please use the constant for "~>" though.
```

f. Replace https://github.com/specify/specify7/blob/7e478b967a4b2be2e5cfbc868535a574f8535c88/specifyweb/stored_queries/batch_edit.py#L399
by  `node.name.lower() if not isinstance(node, TreeRankQuery) else node.get_wb_name()`.

And that should be all. This is because the relationship name (and hence the rank) will contain treedefid. If the "treeId" prop is required (like in here: https://github.com/specify/specify7/blob/eb3c8200bad92af9f43c063655ba47f8a02badf2/specifyweb/frontend/js_src/lib/tests/fixtures/uploadplan.1.json#L319). Trace the code out in TreeRecord.py (as introduced in https://github.com/specify/specify7/pull/5091) and see if it actually is needed. I can spend time for a more in-depth review laer. That file (after 5091) is waaay too bloated (and I intended on refactoring as part of the merge of production and 4929). If it looks like it is actually needed, construct TreeRankRecord itself. To do that, extract the id from the key here (which was added by get_wb_name, and then later get the name of the treedef). The key, in question, is here:
https://github.com/specify/specify7/blob/7e478b967a4b2be2e5cfbc868535a574f8535c88/specifyweb/stored_queries/batch_edit.py#L934. Yeah, quite hacky, but things are unnecessarily complicated from bloating in TreeRecord.py.

g. You'd not need to make adjustments to parsing of batch-edit pack. 

h. Things should just work out of the box by this time now.


## Second way:
a. Don't make any changes to the frontend.
b. Before getting row plan map in batch-edit query below, augment the query such that if a rank appears, then you add all trees that contains that rank (in that discipline). So, essentially, you are making the choice for the user here (that's why no frontend changes).
https://github.com/specify/specify7/blob/7e478b967a4b2be2e5cfbc868535a574f8535c88/specifyweb/stored_queries/batch_edit.py#L1034

That is, suppose you have two trees Fossil and Rocks and both contain rank species. And user's query was (Determination -> Taxon -> Species -> Name), then rewrite this to contain two query fields:

(Determination -> Taxon -> Species (Fossil) -> Name)
(Determination -> Taxon -> Species (Rocks) -> Name)

So, each such field will get diverged into however many trees contain it. 
If you have some kind of filter, things become more complicated (in the worst case, you'd need to generate four fields instead of two -- the extra two will be for `OR null`). 
This approach is quite complicated to implement correctly. However, the all the backend changes from method 1 will still work here (so those are common). Maybe ask UX to see what is acceptable? 

I can look over the changes in TreeRecord.py more deeply later (on Tuesday/Thursday morning/evening). Querying a specific tree support on the backend is needed regardless. Definitely see if you have a better idea. On the frontend btw, it'll be probably be beneficial to add the treedef name in the list in BatchEdit.tsx that asks you to select extra fields. Also, if you feel things can be simplified in batch-edit in general (even for simple fields PR), definitely do that.

specifyweb/frontend/js_src/lib/components/WbPlanView/navigatorSpecs.ts
realVinayak, line 1: Also, this might be out-of-scope for a while, but can nested-to-many be enabled on frontend on either of the two PRs? On 4929, before the merge from production, nested-to-many were supported (and enabled on the frontend) so worked. Looks like that [small](https://github.com/specify/specify7/issues/2331) feature was lost in the merges somewhere.

Anyways, you can probably use these commits as reference (it is all I did to support nested-to-many on frontend in 4929 IIRC): The implementation is backwards compatible (so previous upload plans are not rejected).
(this enabled nested-to-many) https://github.com/specify/specify7/pull/4929/commits/6a6a234bb28a121621cf490768828acc93981867
(frontend unit tests needed to be adjusted for nested-to-manys) https://github.com/specify/specify7/pull/4929/commits/b36aa1739fc0f6d7d2d42394525f1fa3c9b96131

beach53:
Vinnie

You are keeping all this stuff in your head?  I'm worried about you.


James H. Beach
Specify Collections Consortium
Biodiversity Institute
University of Kansas
1345 Jayhawk Boulevard
Lawrence, KS 66045, USA

***@***.***
www.specifysoftware.org
Office: +1 785.864.4645
Cell: +1 785.331.8508

realVinayak:
Eh, just wanted to give Sharad more context about this - I recall how hard it was to fix bugs in xml-editor, and that's after we had videos from Max! If it makes you feel any better, I don't remember much beyond changes I did in batch-edit and nested-to-manys.

specifyweb/frontend/js_src/lib/components/WbPlanView/navigatorSpecs.ts
sharadsw: Added these to #6216

sharadsw:
NOTE: Querying on Taxon names that involve different trees don't work correctly in this PR. 
In prod:
Screenshot 2025-02-18 at 5 01 11 PM
`Deltodus` and `Test Genus` belong to different trees and the results work correctly. In this branch, a query like this returns no result at all.

realVinayak:
huh that is weird. By the way, is the Deltodus duplicated? What does the query look like?

sharadsw:
Weird, turns out one of the Deltodus actually belongs to Species instead: https://calvertmarinemuseum20250206-production.test.specifysystems.org/specify/view/taxon/38726/

realVinayak:
Ah that's fine. Curious what the sql query itself looks like on this branch for that query not working. It should be log level debug

sharadsw:
Prod: https://pastebin.com/imefgzi0
Batch edit: https://pastebin.com/ADt4HrGx

Prod looks at the treedefid as well in the case whens

realVinayak:
that's perfect thanks. I can look at that in an hour or so. I did see the treedefid discrepancy in the code, but that would not matter (so yes, the code from production is redundant).

realVinayak:
This is a wacky bug, and it also happens in issue-5413 (so batch edit too) but I think you already knew that. The diffs helped. Specifically, I noticed this:
image
That is, bug _had_ to be here:
https://github.com/specify/specify7/blob/4814f1161de9989e71ab2025866de71a100dc487/specifyweb/stored_queries/execution.py#L882-L889
The reason is because TreeRankQuery instances are dynamically constructed. The proper way to fix this would be use NamedTuples for fields and relationships. See https://github.com/specify/specify7/issues/628.

I guess a decent way could be to 
```
class TreeRankQuery(Relationship):
    # FUTURE: used to remember what the previous value was. Useless after 6 retires
    original_field: str

    def __hash__(self):
        return hash((TreeRankQuery, self.relatedModelName, self.name))

    def __eq__(self, value):
        return (
            isinstance(value, TreeRankQuery)
            and value.name == self.name
            and value.relatedModelName == self.relatedModelName
        )
```

sharadsw:
NOTE:
After internal discussion, the conclusion that was reached was:
- No change to frontend queries as we need the ability to search on ranks across all trees
- The error dialog (https://github.com/specify/specify7/pull/6196#issuecomment-2644232019) will instead be a prompt that notifies the users what fields are missing
- The missing fields will then be added on the backend when creating the batch edit dataset

sharadsw:
@realVinayak Since the frontend queries are not to be changed and we want to add missing fields to the dataset directly, I think the second way you mentioned sounds appropriate. Can you clarify what you mean by adding two extra fields for `OR null`?

Also, I tried hardcoding some things around to get a feel for what changes need to be made. When I make a change to a Taxon and commit, it gets highlighted as a new cell and a new node gets uploaded to the tree. The behavior is the same in #4929 before the prod merge (commit 2d86ba45d611a2cea2fff03f2babc013bddf13b5). Is this expected?

realVinayak:
Sorry for the delay.

**Question 1**
Yeah, the second one seems like the way to go. Actually, I tried implementing it myself, and realized I misspoke when I said adding two extra fields for the null. My initial thought process was to make it so the user makes a query like `(Species->name == "whatever")`, then this needs to be broken into (assuming species occurs in treedefs 1 and 2)

```
(Species->name == "whatever" and treedef_id == 1) OR (Species->name == "whatever" and treedef_id == 2)
```

Note here that the OR is needed. Otherwise, you'd be constructing 

```
(Species->name == "whatever" and treedef_id == 1) AND (Species->name == "whatever" and treedef_id == 2)
```

which can never be true. So, I thought this could be fixed by making the below (which is what the message above meant)

```
(Species->name == "whatever" and treedef_id == 1 OR Species->name is null)  AND (Species->name == "whatever" and treedef_id == 2 OR Species->name is null)
```

But the last one is actually wrong because then the cases where both are null will also be added.

In conclusion, the first variant is correct. Since only one of the OR conditions can be true at a time anyways, it's good.

**Question 2**
Yes, that's "expected". Actually, we can't sanely do batch-edit on tree ranks (notice the specificity), it's all a lie. There's just isn't a clear definition for it, so I just made it _exactly_ like workbench, but that record data is cloned. That is, say you have something like

image

And you bring it in batch-edit, like below and change the name (the yellow), it'll try cloning all the attributes of the previous "TestSpeciesOther", and try inserting like workbench. So, you can never actually do batch-edit _on_ tree ranks.

image

This might seem like there's then no need of supporting tree ranks, but, bulk changing determinations can be done this way (the changed node will be matched and created if not found). Does this make sense? I'd recommend asking grant if he has sometime to discuss this, I did demo this quite a bit before. 

BTW, here is the corresponding query (for the above screenshots) 
image

You can access the query at http://localhost:5050/specify/query/261/. Lmk if that link doesn't work /j.

**Misc**

**Why do we care about the "highest" rank?**
Just to be on the same page about trees.

Highest means from the bottom. Lower rankid, more higher in the tree.

There is no harm in always showing all the ranks to the user. That is, every time a relationship to tree (with ranks) appears, we can show all the ranks for all the trees. This simplifies implementation _quite_ a bit (all that findMissingRank stuff won't be needed). However, this is bad for usability because, you know, some have like 27 ranks + multiple trees can't make it better. So, we show the _least_ ranks in _least_ trees needed to get stuff done. For a single tree, it'll be all the ranks below the highest. 

But, wait -- do we always need that? If the user selects "Species" and "Genus", are there cases where we can show _just_ those two ranks? Yes, in fact, if taxon is the base table, you can do that. But, if you are something like CO->det->taxon, we need to show all the ranks below the highest (genus). 

(however, what about below the lowest rank, so instead of everything below genus, what about everything below species? why would that not work? -- think about this! there are cases where this won't work)

In the case of multiple tree, we can do the exact same thing, but for all the trees the selected ranks can be part of. In fact, we can only show _just_ those trees that the user selected via ranks (rather than showing _all_ trees).

Need to be careful here, there can be more than one path to the exact same tree. Think about "CO->CollectingEvent->CEA->taxon" and "CO->det->taxon".

**misc again**

Batch-edit is two steps as you know: 
1. Query to dataset
2. Batch-edit within workbench framework

I actually got the first part done for multiple trees yesterday (sunday) - I was curious about it. I don't expect you to use it (that's why I have not made a PR on sp7 repo). However, if you're curious about it, here is the PR equivalent on my fork of sp7 repo (comparing my changes to this branch): https://github.com/realVinayak/specify7/pull/1. The main function you'd be tracing is [this](https://github.com/realVinayak/specify7/blob/687849bfbe170268a4e59ee15f3f20161402764b/specifyweb/stored_queries/batch_edit_query_rewrites.py#L90). The above screenshots are actually from that branch. You might recognize the different tree names. Here are some brief feature notes:

_Good things_
1. Fields are added on the backend. **Currently, they are not localized**. However, the treedef name are added to the column name (to better disambiguate) -- you can see that in the above workbench screenshot (that's from the above PR), where "botany" and "paleo taxonomy tree" is added.
2. Only the most required fields, for the _least_ ranks, in the _least_ trees are shown. That is, if you selected Genus, and there happens to be ["Species"] and ["Species", "Subspecies"] below in two different trees, it'll just show those three.
3. If you select "Species", and there is another tree with ranks ["rank1", "rank2"], that later tree won't be included (because ranks aren't there)
4. Filters are automatically, and correctly handled. The OR trick did it.
5. Currently, it acts on the RowPlanMap, so if there are more than 1 path to the same tree, they are still correctly handled.
6. It's generic for all trees

_Bad things:_
1. Doesn't show missing fields on the frontend. Currently, if just directly constructs the dataset from the query (without a popup). However, this can still be done -- it's more simpler than backend logic.
2. Doesn't do any localizing -- needs hacks.
3. This is **just** the first part of batch-edit. haven't spent more than 5 minutes on the workbench/upload/tree_record.py file when coding this, I don't know if the unit tests run correctly (probably need changes) and if the workbench part (component 2) of batch-edit works or not (probably doesn't). However (this might help you in your implementation) -- batch edit for treerecord is significantly more simpler than for uploadtables (because we can't have in-place edits, so all we're doing is just clone). I'd recommend going back to 4929 before the prod merge and checking it out for yourself how less complicated it is, compared to uploadtables (non-tree relationship) edits.
4. It doesn't necessarily order the columns correctly -- that's complicated.
5. It's quite complicated in general, will require unit tests
6. Might make future changes (if sp7 supports multiple treedef query) harder, although, it's quite considerate towards it.

_Reflections_
I did think about different variations -- doing insertions of fields on the frontend, vs on the backend. Since all the query construction happens on the backend, it does make things a bit easier -- at the cost of localizing, but that can be fixed with hacks... BUT, any of this is not supposed to mean there isn't a better and simpler implementation than what I did in that PR.

sharadsw:
@realVinayak Thank you! Your approach is a lot cleaner than what I was trying to do. I'm gonna use your code if that's okay.

I'm not sure TreeRankRecord is needed for the upload plan. It seems upload works the same as long as the key follows the format `<treename>~><rankname>` but might as well keep TreeRankRecord just to keep it consistent between WB and Batch Edit. Upload will use the same code regardless

realVinayak:
> I'm gonna use your code if that's okay.

Yeah that's fine - it's open source and public. Still, if you feel you can simplify logic more, please do so!

> It seems upload works the same as long as the key follows the format......

Yeah, I ended up making a conservative assumption that I need to use TreeRankRecord. I think the apply_batch_edit_pack for TreeRecord would need some adjustement (since, internally, the batch_edit.py uses the key of format `<treename>~><rankname>~><treedef_id>` but that won't be a hard change (just would require change to apply_batch_edit_pack).

specifyweb/workbench/upload/treerecord.py
```diff
@@ -293,7 +295,7 @@ def to_json(self) -> Dict:
             rank_key = rank.rank_name if isinstance(rank, TreeRankRecord) else rank
             treeNodeCols = {k: v.to_json() if hasattr(v, "to_json") else v for k, v in cols.items()}
             
-            if len(cols) == 1:
+            if len(cols) == 1 and not isinstance(rank, TreeRankRecord):
                 result["ranks"][rank_key] = treeNodeCols["name"]
```
sharadsw, line 299: Looks like this change breaks the test for unparsing an upload plan
https://github.com/specify/specify7/blob/7ff9c53f839046c4d136891258da94bbccd2b723/specifyweb/workbench/upload/tests/testschema.py#L37-L39

The expected value for the unparsed plan expects a TreeRankRecord for every rank. Could simply fix this by changing the original upload plan to include TreeRankRecords, which shouldn't break anything functionally as upload should work fine either way. Need to investigate if there are side effects

sharadsw: Here's a diffcheck: https://www.diffchecker.com/7wYAWTtk/

sharadsw: Removed this to fix the test rather than changing the test. Functionally, nothing should change since upload plan still works when the key name is in the format `<treename>~><rankname>`

sharadsw:
BUG: Frontend missing field calculation is not always accurate. Backend calculation seems fine

CarolineDenis (approved):

lexiclevenger (commented):
I noticed that many of the hidden relationship fields (i.e. Created By Agent > First name) that were added for #6011 throw errors if you query on them. Should that be addressed in this PR, or should I write it up separately?

[Specify 7 Crash Report - 2025-03-10T18_51_50.087Z.txt](https://github.com/user-attachments/files/19169029/Specify.7.Crash.Report.-.2025-03-10T18_51_50.087Z.txt)

Screenshot 2025-03-10 at 1 48 07 PM

emenslin:
I don't agree with ranks from other trees being carried over because it feels inconsistent and confusing. I could kind of understand lower ranks from the same tree being necessary but adding ranks from other trees feels like it would complicate things. This is just an example so not necessarily realistic, but this db has 3 different trees with overlapping ranks and even though I just wanted to map family, genus, species, and subspecies, it ends up adding 17 columns. 


https://github.com/user-attachments/assets/c0d0b461-d51d-49d3-8071-ba571e30a1cc



Aside from narrowing down the query by looking for specific taxa (which could still overlap between trees), could we add another way so you could select what trees you do or don't want before batch editing? I know some work was done on #5251 about being able to select the specific tree and then select the rank for each tree (like how workbench is), is there any plan to implement that or can we make a plan for it?

sharadsw:
@lexiclevenger I think a separate issue for that would be great

sharadsw:
@emenslin That's a very fair concern. We actually discussed this issue in a dev meeting before and came to this conclusion (https://github.com/specify/specify7/pull/6196#issuecomment-2672787436) because we need the ability to query over any rank across all trees in the db - which means we cannot really restrict queries to a specific tree and need to add extra fields directly from the backend instead when batch editing.

One potential alternative we discussed was to have tree names in the query (like in the WB) along with an option `(any tree)`  similar to `(any rank)` which would allow us to still have the ability to query across any rank+tree. However, we decided against it because it would be big in scope as it has implications in Queries as well as the Workbench.

I'm happy to have this in a separate issue as an improvement though. We will have to implement it if we decide to go ahead with #5251

emenslin:
This sounds like the ideal solution in my opinion, I understand that it would likely take a while, but this seems a lot better usability wise than the current implementation. I know that not all users will have multiple trees, but the current implementation would be bad for users who do. 

> One potential alternative we discussed was to have tree names in the query (like in the WB) along with an option (any tree) similar to (any rank) which would allow us to still have the ability to query across any rank+tree

Some other solutions until we could implement that could be allowing users to collapse columns they don’t want to use, a dialog after pressing batch edit that could let users select which trees they want included, having the data set ordered by tree then rank instead of rank than tree. Using my original video for an example: show ranks for paleo taxonomy tree from family -> subspecies, then family -> subspecies for mammalogy, etc., instead of the current implementation where it shows each trees family first, then genus, then species, etc. 

There's probably some better solutions out there but these were just some that I could think of. Any thoughts @specify/ux-testing?

specifyweb/workbench/upload/treerecord.py
sharadsw: Looks like we need this to make the data mapper work correctly in #6126

sharadsw:
@emenslin I like the idea of selecting trees from a dialog. It would add more development time but it would simplify batch edit datasets in #6126. Currently in #6126 when editing CO -> determinations you can end up with a ridiculously large number of tree rank columns because each missing rank from each tree will get added for each determination that exists in a CO.

Reordering the extra columns doesn't seem straightforward. I'll look it into it but we may have to push that to a later issue

Any thoughts on this? @grantfitzsimmons

grantfitzsimmons:
To me, it seems that this is a necessary extension if we want to maximize the usefulness of batch edit in any collection with multiple trees / types. Since I believe that's what we want to do, I think it justifies the complexity.

> @emenslin I like the idea of selecting trees from a dialog. It would add more development time but it would simplify batch edit datasets in #6126. Currently in #6126 when editing CO -> determinations you can end up with a ridiculously large number of tree rank columns because each missing rank from each tree will get added for each determination that exists in a CO.

sharadsw:
Refactored a bunch of stuff to group missing ranks by tree name:
image

The plan is to add checkboxes next to tree name if the user has multiple trees. If there is only a single tree in the db, the dialog will only notify of missing ranks being added for that tree (and no checkbox)

Let me know if you have any suggestions @specify/ux-testing

sharadsw:
This is what I have now
image

- Users can pick which trees they want in case of conflicting ranks
- Batch edit will only add ranks from selected trees
- If no trees are selected, batch edit will use ALL trees like before

When selected ranks do not conflict but there are missing ranks:
image

sharadsw:
@specify/ux-testing This can be tested again
- Missing ranks dialog will now allow users to select trees they wish to edit with
- Ordering of tree columns needs more investigation and will probably be added in #6126 or another issue

emenslin (changes requested):
Cannot batch edit taxon or storage. Chronostratigraphy, lithostratigraphy, and geography all do seem to be working though, in case this is a db problem here's a link to the db: https://ojsmnhbatchedit20250404-issue-6127.test.specifysystems.org/specify/query/214/

https://github.com/user-attachments/assets/45b93017-5579-4de3-8195-4bcc2f85de01


[Specify 7 Crash Report - 2025-04-10T15_14_45.752Z.txt](https://github.com/user-attachments/files/19687418/Specify.7.Crash.Report.-.2025-04-10T15_14_45.752Z.txt)

sharadsw:
@emenslin Should be fixed now. To-many queries will be disabled with a dialog 
image

emenslin (changes requested):
#### Test batch edit with a single tree in db
1. Edit a single node
  - [x] Test validate/commit
  - [x] Verify edited cells were highlighted as Updated Cell
  - [x] Test rollback

2. Edit a tree path
  - [x] Test validate/commit/rollback
  - [x] Verify changes are highlighted as New Cell and the tree path is a new upload to the Tree

3. Test missing ranks
  - [x] Verify there is a missing rank dialog indicating the lower ranks will be added
  - [x] Verify missing ranks were added correctly

#### Test batch edit with multiple trees in the db
- [x] Repeat steps for editing single node
- [x] Repeat steps for editing a tree path
- [x] Repeat steps for missing ranks
  - [x] Verify missing ranks dialog allows you to select trees to edit with
  - [x] Verify ranks of selected trees are added correctly
  - [x] Verify ALL trees are used when no trees are selected in the dialog

--- 
Everything looks good except you cannot batch edit chronostrat 

https://github.com/user-attachments/assets/1fe68c13-3f97-47bc-a0de-02c657466b0f


Link to the db: https://ojsmnh20250404batcheditlatest-issue-6127.test.specifysystems.org/specify/

sharadsw:
@emenslin Should be fixed now

pashiav (approved):
#### Test #6011 
- [X] Verify all tree fields can be queried with 

#### Test batch edit with a single tree in db
1. Edit a single node
  - [X] Test validate/commit
  - [X] Verify edited cells were highlighted as Updated Cell
  - [X] Test rollback

2. Edit a tree path
  - [X] Test validate/commit/rollback
  - [X] Verify changes are highlighted as New Cell and the tree path is a new upload to the Tree

3. Test missing ranks
  - [X] Verify there is a missing rank dialog indicating the lower ranks will be added
  - [X] Verify missing ranks were added correctly

#### Test batch edit with multiple trees in the db
- [x] Repeat steps for editing single node
- [x] Repeat steps for editing a tree path
- [x] Repeat steps for missing ranks
  - [x] Verify missing ranks dialog allows you to select trees to edit with
  - [x] Verify ranks of selected trees are added correctly
  - [x] Verify ALL trees are used when no trees are selected in the dialog
---
Looks good! 

---
**Notes**
After completing testing on the Test Panel, I was doing a quick gen test and BE validation stopped working on [calvertmarinemuseum_batch_edit_2025_02_10](https://calvertmarinemuseumbatchedit20250210-issue-6127.test.specifysystems.org/specify/). Looks to be the same as https://github.com/specify/specify7/issues/6431. 

Working locally so I will approve this PR.

lexiclevenger (commented):
#### Test #6011 
- Create a query with a Tree base table (eg: Taxon, Storage, Geography, Lithostrat, Chronostrat, Tectonic Unit)
- [x] Verify all tree fields can be queried with 

#### Test batch edit with a single tree in db
1. Edit a single node
  - Make a tree query with `(any rank)` and some other simple fields
  - Click Batch Edit
  - Edit fields
  - [x] Test validate/commit
  - [x] Verify edited cells were highlighted as Updated Cell
  - [ ] Test rollback

2. Edit a tree path
  - Make a tree query with multiple rank names
  - Batch edit
  - Edit fields
  - [ ] Test validate/commit/rollback
  - [x] Verify changes are highlighted as New Cell and the tree path is a new upload to the Tree

3. Test missing ranks
  - Make a tree query with a rank that isn't the lowest rank
  - Click batch edit
  - [x] Verify there is a missing rank dialog indicating the lower ranks will be added
  - Click continue
  - [x] Verify missing ranks were added correctly

#### Test batch edit with multiple trees in the db
- [x] Repeat steps for editing single node
- [x] Repeat steps for editing a tree path
- [x] Repeat steps for missing ranks
  - [x] Verify missing ranks dialog allows you to select trees to edit with
  - [x] Verify ranks of selected trees are added correctly
  - [x] Verify ALL trees are used when no trees are selected in the dialog
  ---
  Looks good, all trees can be batch edited. However, I couldn't roll back any of the data sets. Maybe due to #6427? Here is the error mesage:
Screenshot 2025-04-22 at 2 31 34 PM

lexiclevenger:
Here is a data set that won't roll back: https://ciscollectionsbatchedit220250210-issue-6127.test.specifysystems.org/specify/workbench/28

sharadsw:
@lexiclevenger Can you try again? Should be fixed now

sharadsw:
Created a different issue for cases where rollback fails due to a disambiguation error: https://github.com/specify/specify7/issues/6443

Datasets without a disambiguation error should have a working rollback

emenslin (dismissed):
#### Test #6011 
- [x] Verify all tree fields can be queried with 

#### Test batch edit with a single tree in db
1. Edit a single node
  - [x] Test validate/commit
  - [x] Verify edited cells were highlighted as Updated Cell
  - [x] Test rollback

2. Edit a tree path
  - [x] Test validate/commit/rollback
  - [x] Verify changes are highlighted as New Cell and the tree path is a new upload to the Tree

3. Test missing ranks
  - [x] Verify there is a missing rank dialog indicating the lower ranks will be added
  - [ ] Verify missing ranks were added correctly

#### Test batch edit with multiple trees in the db
- [x] Repeat steps for editing single node
- [x] Repeat steps for editing a tree path
- [x] Repeat steps for missing ranks
  - [x] Verify missing ranks dialog allows you to select trees to edit with
  - [ ] Verify ranks of selected trees are added correctly
  - [x] Verify ALL trees are used when no trees are selected in the dialog

--- 

Missing ranks are added in a random order, is this supposed to happen? This is an example with a single taxon tree but it also happens with multiple trees. 

https://github.com/user-attachments/assets/0d96a2f2-f0dc-4150-ae26-2765c3b19906


Also I did run into this api rank problem causing unload protect a couple times, it seems to only happen when batch editing on 'any rank' but I'm not sure if it happens anywhere else. 


https://github.com/user-attachments/assets/5b044a2a-feec-4646-8d54-3a9803f53a63

sharadsw:
@emenslin Looks like ranks were being sorted alphabetically. I have changed it so that the user's order of ranks in the query is kept as is. 

Created a separate issue for the unload protect error: https://github.com/specify/specify7/issues/6448

sharadsw:
@emenslin Actually, I had to revert the change. Trying to fix column ordering breaks tree datasets in relationships. I think we'll have to push it for a different milestone. Can you create a issue for it?

lexiclevenger (approved):
#### Test batch edit with a single tree in db
1. Edit a single node
  - [x] Test validate/commit
  - [x] Verify edited cells were highlighted as Updated Cell
  - [x] Test rollback

2. Edit a tree path
  - [x] Test validate/commit/rollback
  - [x] Verify changes are highlighted as New Cell and the tree path is a new upload to the Tree

3. Test missing ranks
  - [x] Verify there is a missing rank dialog indicating the lower ranks will be added
  - [x] Verify missing ranks were added correctly

#### Test batch edit with multiple trees in the db
- [x] Repeat steps for editing single node
- [x] Repeat steps for editing a tree path
- [x] Repeat steps for missing ranks
  - [x] Verify missing ranks dialog allows you to select trees to edit with
  - [x] Verify ranks of selected trees are added correctly
  - [x] Verify ALL trees are used when no trees are selected in the dialog
---
Looks good! I didn't run into any issues and was able to roll back all of the data sets as long as they didn't have disambiguation errors.

acwhite211 (approved):

PR #6205, opened 2025-02-06
Pass SSO users through choose_collection to assign needed collection cookie 

### LINKED ISSUE #5323: NoMatchingRuleException: SSO redirects user to collection id that they do not have access to
**Describe the bug**
A user is successfully able to setup and link a Specify account with an identity broker, and *complete a successful sign in through SSO once*. The next time they go to sign in (on a different machine), they receive the following error.

`{"NoMatchingRuleException": [{"collectionid": 4, "userid": 8, "resource": "/system/sp7/collection", "action": "access"}]}`

Interpreting the error seems straightforward, `userid 8` is trying to access `collectionid 4`, which they do not have access to. The question is why SSO is directing them there in the first place. SSO should return a user to the same collection they left off at (because the session is still active application side), but they never would have been able to have access to this collection, and thus could not previously have been in it.

```
Division 1:
[Collection]
[Collection]
[Collection] <- collection id 4
...

Division 2:
[Collection] <- user id 8 has access here, and only here
```

When the user *is* given read-only access to `collectionid 4` SSO works just fine, confirming the above interpretation. Note that through this process #2405 comes into play, but I will keep those details out of this report and see if they have been documented there.

**To Reproduce**
My hunch is that this error is related to #2405, as all other users have worked fine except for this user, which is the only one in a separate division. At this point, I am not certain on how to reproduce this, and am opening this issue because we have the opportunity to narrow down what can be causing this (as we have several users left who haven't been setup yet in that division). The first thing that I will pay attention to for the next user, is whether the collection/division that the admin is in when creating the invite link matters. Perhaps if the admin is in division 1, while creating a link for a user in division 2, then the wires get crossed.

I do not have a hypothesis for why their first SSO sign-in worked the first time for them, before they were given read-only access to `collectionid 4`.

I do not know why SSO routed to `collectionid 4` specifically out of the other available options.

**Expected behavior**
The user is able to sign in through SSO and will always be placed into the one collection that they have access to (if that is the case), regardless of division.

**Screenshots**
N/A. Error message appears as black text on blank web browser window.

**Crash Report**

N/A

- Specify 7 Version: 7.9.6.2
- Database Name: beaty

**Reported By**
Beaty Biodiversity Museum

**Additional context**

mpitblado:
Fixes #5323

> As an external contributor, please feel free to let me know if outside PRs are not preferred, and happy to close. Appreciate any and all feedback

Currently, it appears that users who go through the `oic_callback` flow can experience an error in which they do not have the necessary permissions to access collection id 4. I believe this is due to a collection cookie not being set. Fixing this would involve either implementing something similar to what `choose_collections` does in `oic_callback`, but to avoid duplication, it seems easier to just pass the user through `choose_collection`. Passing the user through `choose_collection` is done elsewhere, such as the invite_token flow.

With the switch, I have thought about potential duplication of spuserexternalid entries. In `oic_callback`, `request.session['external_user']` should only be created if the id does not already exist. Thus, the if condition in `choose_collection` should not trigger, and another id should not be created.

Condition for adding external_user to the session
https://github.com/specify/specify7/blob/4be07f55ff35e45e0b7ea606de8738c345de1169/specifyweb/accounts/views.py#L211-L221

Condition in `choose_collection`, checks for `external_user` in session before creating new entry in spuserexternalid
https://github.com/specify/specify7/blob/4be07f55ff35e45e0b7ea606de8738c345de1169/specifyweb/accounts/views.py#L350-L361

The rest of  `choose_collection` handles the cookie, and collection form, and should cause no issue.

As a result of this PR, users of SSO would now no longer enter straight into their last used collection unless they only have access to one collection, instead they would have to pick the collection, with the last used collection being selected as default (similar to those using local login).

### Checklist

- [ ] Add automated tests (N/A)
- [ ] Add relevant issue to release milestone (N/A)
- [ ] Add relevant documentation (Tester - Dev) (N/A)
- [ ] Add a reverse migration if a migration is present in the PR (N/A)

### Testing instructions

I have tested the below on 7.9.6.2 with a local keycloak instance with the changes in this PR applied.

- Confirm that user with access to all collections can login through SSO
- Confirm that user without access to `collection id=4` can log in through SSO. 
- Confirm that user without access to `collection id=4` can log in through SSO after not logging out through Specify's logout button and clearing cookies (such as through a private window or by manually deleting cookies).
- Confirm that a new user can link their specify account with an external provider (user presses SSO button after having a local account created). Note, this is not the invite_link flow.


specifysoftware:

https://discourse.specifysoftware.org/t/problems-with-login-over-entraid-regarding-rights/2146/15

melton-jason (approved):
Nice sleuthing and solution! 

I tested this with a local KeyCloak instance as well. 
I also ran the frontend and backend tests to make sure none failed as a result of this change: all passed! 
(If you're unsure of how to run the tests locally, check out the [Automatic Testing](https://github.com/specify/specify7/wiki/Automatic-Testing) page of the developer documentation!)

Thanks for being mindful of not duplicating any `spuserexternal` records. 
I believe you're right in your assessment that a `SpUserExteranlID` can only be created through the SSO process when one doesn't exist already (specifically, if Specify does not already have the User's external id for the given `provider` already recorded).  

I apologize for this review taking so long from when you first opened the PR. It honestly completely slipped under my radar until you mentioned it on the Speciforum two weeks ago. 
For PRs in the future, feel free to directly mention me (via @melton-jason) or request my review, so I get notified and it sticks around in my Notifications here in GitHub! 

---

I did some further investigation into why the User was being routed specifically to the Collection with an ID of 4, and here's what I found! 

The problem was being caused in the context middleware which injects the `specify_collection`, `specify_user_agent`, and `specify_user` into the request: 

https://github.com/specify/specify7/blob/499dafb7b7c4482f8df2893d43c12a9714f6bdc3/specifyweb/context/middleware.py#L56-L61

These attributes are used throughout the backend in other endpoints so operations can have easy access to some scoping context. Whenever the attributes are first accessed, Specify evaluates the attributes; the value is then "cached" on the request object, and can be accessed without further evaluation. 

Specifically, the problem occurred when determining the `specify_collection` when the `collection` cookie was not set: 

https://github.com/specify/specify7/blob/499dafb7b7c4482f8df2893d43c12a9714f6bdc3/specifyweb/context/middleware.py#L26-L36

(At least in my local testing, the problem was occurring because the `collection` cookie was not set before/during the `/accounts/login/` and `/accounts/oic_callback/` endpoints are hit: which makes sense because there shouldn't be a collection to "scope" the requests to). 

Specify is arbitrarily choosing the first collection in the list of all collections: which is usually always the oldest Collection because MySQL/MariaDB _usually_ sorts by Primary Key if no other sorting behavior is defined. 

For completeness, the first evaluation of `request.specify_collection` most often occurs in another middleware--the permission middleware: 

https://github.com/specify/specify7/blob/499dafb7b7c4482f8df2893d43c12a9714f6bdc3/specifyweb/permissions/middleware.py#L16-L24

The reason why routing to `/accounts/choose_collection/` does not result in an error and works in the first place over the frontend `/specify/` url is because the `/accounts/choose_collection/` url explicitly bypasses the permission middleware

https://github.com/specify/specify7/blob/499dafb7b7c4482f8df2893d43c12a9714f6bdc3/specifyweb/accounts/urls.py#L27-L30

The `/accounts/choose_collection/` endpoint then, as you saw, properly sets the `collection` cookie. 

Either in this PR or another (because the change could arguably be out of scope) I think we can additionally modify the context middleware to be a little more robust. Specifically, if the `request.user` is defined (the user is logged in), then we can at least select a Collection that the `request.user` has access to, instead of arbitrarily selecting one.

funkyferdy:
@melton-jason & @mpitblado : Glad to hear that we have a solution and such a detailed explanation to this issue! Looking forward to test this once is released.

CarolineDenis (approved):

grantfitzsimmons (approved):

PR #6235, opened 2025-02-12
Fix setting loan prep quantities for consolidated prep siblings

### LINKED ISSUE #6223: When using COs from consolidated COG in interactions, the quantity is not set to the max
- Create a COG with consolidated type 
- Add minimum 2 COs children 
- Create preps for each CO with a min count of 2 
- Create a new loan 
- Enter one of the COs cat num 
- Select quantity 1 only for one CO in the prep dialog before clicking apply and leave it to 0 for the rest
- Verify that the siblings have been added to the loan 
- See that the quantity for each as been set to 1 instead of their maximum 

Issue with this: 
You could now create another loan with the rest of the preps even if they are a part of the consolidate

acwhite211:
Fixes #6223

Fix setting loan prep quantities for consolidated prep siblings.  When one sibling's quantity is set to its max amount, then all of the consolidated siblings should be set to there max.  The max amount is set when the loan prep's quantity is equal to the count of the associated prep.

### Checklist

### Testing instructions

- Create a new consolidated collection object group.  Create three collection objects in the group.  For each collection object, create a preparation.  Set the count of the first prep to >= 2, the second to >= 2, and the third to >= 2.

image

- Navigate to the "Interactions" tab on the side menu and then select "Loan".
- Select "By CIDA/Catalog Numbers", and add the catalog number of the first CO in your new COG, and press "Next"
- Select only the first CO in the list, and set the "Selected" amount to the "Available" amount.  Press "Apply"

image

- Enter a loan number and press "Save".
- [x] See that all the loan preps are set to their max quantity, where the quantity equals the count.

image

NOTE: 
- the quantity will be set to its max only once you have saved your loan

### DIFF
--- a/specifyweb/interactions/cog_preps.py
+++ b/specifyweb/interactions/cog_preps.py
@@ -576,7 +576,7 @@ def enforce_interaction_sibling_prep_max_count(interaction_obj):
         return interaction_obj
 
     interaction_preps = InteractionPrepModel.objects.filter(**{filter_fld: interaction_obj})
-    interaction_prep_ids = set(interaction_preps.values_list("id", flat=True))
+    interaction_prep_ids = set(interaction_preps.values_list("preparation_id", flat=True))
     for interaction_prep in interaction_preps:
         prep = interaction_prep.preparation
         sibling_preps = get_all_sibling_preps_within_consolidated_cog(prep)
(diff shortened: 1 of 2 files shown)


CarolineDenis (approved):

sharadsw (approved):

emenslin (approved):
About my review above (https://github.com/specify/specify7/pull/6235#pullrequestreview-2615289320) from @CarolineDenis 

> basically we don't have access to the quantity and the cog type until the backend is called and we call the backend only once the save is pressed

This is therefore the intended behavior. Although I would still like it if this can be fixed in the future, I do understand now that it is not currently possible.

PR #6236, opened 2025-02-13
Use default typesearch when no custom field

### LINKED ISSUE #6233: Support Default `TypeSearches` Alongside Custom Configurations
Currently, users who implement a custom `TypeSearches` file within App Resources face a challenge when new type search tables or fields are introduced. This manual update requirement not only adds extra workload but also creates inconsistencies in user experience, particularly when new schema expansions occur.

A major problem here is that the `TypeSearches` app resource (see https://github.com/specify/specify7/commits/production/config/backstop/typesearch_def.xml) is almost always custom defined in databases that have used Specify 6 since customizing the fields searched in a query combo box update this file. Once customized, Specify looks at this custom resource rather than the system one.

This leaves users with custom `TypeSearches` unable to use the query combo box for things like Collection Object Groups, Tectonic Units, or any new tables/fields that we add in the future. The only workaround is for users to click on the 🔍 (search) icon next to the query combo box to search manually for records.

## Problem Statement

When new fields are added to the system, such as the recent addition of geo-related fields, users with a custom `TypeSearches` file must manually update their configurations to ensure that new fields, including `cojo` and `taxonTreeDef`, are searchable. This is particularly problematic for users who expect seamless integration and functionality with the system's default search capabilities.

### Current Limitations:
-  Users with a custom `TypeSearches` file do not automatically receive updates or expansions related to default type searches.
-  **Specifically affected tables include:**
	  - `collection`
	  - `specifyuser`
	  - `institution`
	  - `collectionobjectgroup`
	  - `collectionobjectgrouptype`
	  - `tectonicunit`

## Proposed Solution

Implement a logic that automatically fills in the gaps of custom `TypeSearches` files with the system's default settings. This would ensure that any new fields or tables added to the system are automatically integrated into the user's search capabilities without requiring manual intervention.

This proposal does **not** involve updating the custom XML, rather just allowing Specify to "fall back" on a system default rather than looking only at the custom definition for every table.

### Current Workaround

As a stop gap measure for 7.10, you can update the XML for your `TypeSearches` app resource with the following lines between the `<typesearches>` element in the resource:

```xml
    <typesearch tableid="1016"   name="CollectionObjectGroup" searchfield="name" displaycols="name"    format="%s"     dataobjformatter=""/>
    <typesearch tableid="1018"   name="CollectionObjectGroupType" searchfield="name" displaycols="name"    format="%s"     dataobjformatter=""/>
    <typesearch tableid="23"  name="Collection"             searchfield="collectionName" displaycols="collectionName" format="%s"  dataobjformatter=""/>
    <typesearch tableid="94"  name="Institution"            searchfield="name"            displaycols="name"               format="%s"     dataobjformatter=""/>
    <typesearch tableid="72"  name="SpecifyUser"            searchfield="name"            displaycols="name"           format="%s"     dataobjformatter=""/>
    <typesearch tableid="1027"  name="TectonicUnit"         searchfield="fullName"        displaycols="fullName"           format="%s"     dataobjformatter="TectonicUnit"/>
    <typesearch tableid="1026"  name="TectonicUnitTreeDefItem"       searchfield="name"   displaycols="name"               format="%s"     dataobjformatter=""/>
```

### Benefits:
-  **Enhanced User Experience:** Users will benefit from seamless access to new fields without needing to update their custom configurations.
-  **Reduced Maintenance:** This would significantly decrease the maintenance burden on admin users who regularly need to update their `TypeSearches` files to be able to use basic functions of the app.
-  **Consistency Across Systems:** Ensures that all users, regardless of their custom configurations, have access to the same search capabilities and functionalities.

### Discussed in https://github.com/specify/specify7/discussions/6232

<sup>Originally posted by **bronwyncombs** February 12, 2025</sup>
Currently, a custom TypeSearches file in App Resources overrides system defaults. This poses an inconvenience when new typeSearch tables/fields are added since the App Resource must be updated to make new qcbx fields searchable. 

It would be much better for the users if there was logic to fill in the gaps of custom TypeSearches files with the system defaults.  This is especially beneficial when the schema is expanded and users expect new forms/fields to behave with the system's default search abilities. 

The addition of geo, for example, means databases with a TypeSearches file will need to manually configure this for the cojo and taxonTreeDef fields qcbx search to work. </div>

CarolineDenis:
Fixes #6233

### Checklist

### Testing instructions

TODO


sharadsw (approved):

acwhite211 (approved):

alesan99 (approved):
image


I had come a cross this issue before, but now its using the default 👍

I think the same fix should be applied to record formatters though, since now I can see my custom dataobjformatters also overrides all the new defaults

grantfitzsimmons:
>I think the same fix should be applied to record formatters though, since now I can see my custom dataobjformatters also overrides all the new defaults

I strongly agree with @alesan99, should we open a new issue or extend this PR?  @CarolineDenis

CarolineDenis:
@grantfitzsimmons, we can open an issue

grantfitzsimmons:
You mentioned one argument to do it? I'll open an issue.

lexiclevenger (approved):
I tested the following tables from #6233 :

- [x] collection
- [x] specifyuser
- [x] institution
- [x] collectionobjectgroup
- [x] collectionobjectgrouptype
- [x] tectonicunit

---
Looks good. The institution querycbx and type search shows `altname` instead of `name`, but it also does that in 7.9.6.2, so I will write up a new issue for it.

specifysoftware:

PR #6245, opened 2025-02-14
Remove cast to default catalog number formatter

### LINKED ISSUE #6239: Catalog number format not displaying correctly in search results
1. Go to kufish production database on test panel - KUFish_2025_01_02_prod
2. Enter a collection object with the tissue COT which has the catalog number format 2025-******
3. After saving the record, perform a search for that record
4. Note that in the search results, the catalog number does not display in the correct format - instead of 2025-00001, I see 000002025 and the tooltip says 2025

*When reporting an issue on the `production` branch or any development version of Specify, please try to recreate the issue on the latest tagged release (`v7`) and indicate if it does not occur there, indicating a regression.*  This is a new feature so not possible

**Expected behavior**
The catalog number format and tooltip should be correct for each COT type

**Screenshots**

Image
Image

**Crash Report**

If the bug resulted in an error message, please click on "Download Error Message" and attach it here. If there is no error, please follow [these instructions](https://discourse.specifysoftware.org/t/download-specify-7-system-information/1614) to download your System Information.

- OS: [e.g. Windows 10] Windows 11
- Browser: [e.g. Chrome, Safari] Chrome
- Specify 7 Version: [e.g. 7.6.1] production
- Database Name: [e.g. kufish] (you can see this in the "About Specify 7" dialog) KUFish_2025_01_02_prod
- Collection name: [e.g. KU Fish Tissue] KU fish
- User Name: [e.g. SpAdmin] abentley

**Reported By**
KU

**Additional context**

sharadsw:
Fixes #6239

Removes an ancient cast to the default numeric catalog number format in queries. Casting to a format shouldn't really be necessary since we now support different formatters. The issue was caused because queries would convert all catalog number to match the default format whenever a COT with the default formatter was chosen with the gear icon.

### Checklist

- [ ] Add automated tests

### Testing instructions

- Create a CO query
- [ ] Verify all COs have their respective type's format
- [ ] Re-test https://github.com/specify/specify7/pull/5485


CarolineDenis (approved):

melton-jason (approved):
Good work! I'm surprised -and glad- the change is this minimal 😄! 

My primary concern when initially viewing this change was that removing the numeric cast would lead to non-sensical searching for collections using the CatalogNumberNumeric formatter (e.g., for strings `'2' > '100'`, the same Issue as https://github.com/specify/specify7/pull/3501#pullrequestreview-1431053184). 

I think it's important to note that it's definitely still possible to recreate some "non-sensical" searching and sorting results. 

_Most_ importantly in my opinion however, is that the data can be misrepresented in Query Results, which can lead to the confusing results (thankfully not in exports). 

The following video demonstrates both of these concerns: 

https://github.com/user-attachments/assets/96fda759-5ef2-4e68-a8f5-b5b8304cde20

It might be confusing to a user why `123` is greater than `2100`, `2` is somehow greater than `123`, and `3` is the largest catalog number. 
While it still might be nonsensical, I think the underlying problem here is primarily that catalog numbers which can be coerced to the default collection's format _are_, regardless of their actual representation in the database. 
This also holds true with the API; if you query on catalognumbers greater than `"2"` when the collection's format is CatalogNumberNumeric for example, the value is assumed to be `"000000002"` rather than `"2"` regardless of the value of the `formatname` on the queryfield; which might be beneficial in some cases, but harmful in others. 

Regardless, this is still a good fix and step in the right direction!
The concerns might not be blockers of the PR or release, I just thought I'd leave a comment describing the behavior so it's known and documented!

(On a side note, if the query results should be ordered in a particular way, should the exported results - like through CSV - have the same sorting? Here is the export of the Query shown above:  
[Catalog Number Query - Mon Feb 17 2025.csv](https://github.com/user-attachments/files/18829616/Catalog.Number.Query.-.Mon.Feb.17.2025.csv))

combs-a (changes requested):
- [X] Verify all COs have their respective type's format
- [ ] Re-test https://github.com/specify/specify7/pull/5485

The formatting is fixed, which is great!

Relating to the inconsistency with strings that Jason brought up, I don't think it's caused by this PR--rather, it seems like something that should've been mentioned in the previous PR since this is present in production too.

Here is an example query and its result (on this PR):

image

image

And another, where we've changed the query to search for any record *greater* than the lowest result in the previous...

image

image

Now 2024-000001 is greater than 230, when 230 was greater than it before, which is definitely a bit of strange behavior.

If it would be possible, it may be a good idea to still have some kind of filter to automatically *only* include results that are of the COT selected, perhaps. This is possible by just selecting the name of the COT, but that does not support cases in which you want to query for COs with different COTs.

I think for an example query and expected result, it'd be something like:

```
Catalog Number -> Greater Than -> 2025-000001 [COT: FossilTest]
OR -> Equal -> 1 [COT: Vertebrate Paleontology]

Results:
Any CO that has the COT FossilTest with a catalog number greater than 2025-000001 as well as the CO that has the COT Vertebrate Paleontology
```

If this is out of the scope of this PR, let me know and I'll approve it--but I do agree with Jason that this behavior should be addressed since it might be strange to users.

pashiav (dismissed):
- Create a CO query
- [ ] Verify all COs have their respective type's format
- [ ] Re-test https://github.com/specify/specify7/pull/5485
---
When hovering over a line from a CO query, the tool-tip no longer shows up. Works in production (shown in video as well). Tested on DBs `calvertmarinmuseum` and `kufish`.

issue branch: https://kufish20250102prod-issue-6239.test.specifysystems.org/specify/query/new/collectionobject/

production: https://kufish20250102prod-production-2.test.specifysystems.org/specify/query/new/collectionobject/


https://github.com/user-attachments/assets/7d5106d9-7459-45bb-a6c2-85605b3c5dc9


^not finished testing but leaving bug info here.

emenslin (changes requested):
- [x] Verify all COs have their respective type's format
- [ ] Re-test https://github.com/specify/specify7/pull/5485

Results are not being filtered properly. Not sure the exact logic here but for some cat # they show up in the results when they shouldn't or they don't show up in results when they should. In this video I query on cat # greater than 2025-001 and you can see that '2025-001' shows up when it should not and then when I query on equal to 2025-001 it doesn't show up when it should. This does not seem to be consistent with every cat # in a db but I was able to recreate on ojsmnh and chadron. 

https://github.com/user-attachments/assets/4202d9ca-ce48-4664-b0b7-ea626f202b03


dbs: https://ojsmnh110242024production-issue-6239.test.specifysystems.org/specify/query/191/ 
https://chadrontest2-issue-6239.test.specifysystems.org/specify/query/new/collectionobject/

sharadsw:
@combs-a Thanks! I'm gonna open a different PR for filtering COs by a selected COT formatter since the issue was not introduced by this PR and the scope of that seems bigger than what this PR was intended for. Created an issue for it: https://github.com/specify/specify7/issues/6253 

@pashiav I initially thought the tooltip disappearing was a bug but it turns out Specify only displays a tooltip in Queries when the value at a cell does not match its formatted value. Since this PR does not convert catalog numbers to a specific format anymore, the values always match the formatted value and hence there's no tooltip.

@emenslin It's strange, I'm not able to recreate the same results as in the video. I get correct results when querying with `2025-001` 

Equal
image

Greater than
image

However, I did see one weird result where `2025-001` gets included even when querying on the type `Rocks`. I think the issue here ultimately boils down to the same as what Anya mentioned, will be fixed in #6253

sharadsw:
Created #6254 for filtering COs by COT formatter

combs-a (approved):
- [X] Verify all COs have their respective type's format
- [X] Re-test https://github.com/specify/specify7/pull/5485

<hr>

Looks good, all tests passed!

image

emenslin (approved):
- [x] Verify all COs have their respective type's format
- [x] Re-test https://github.com/specify/specify7/pull/5485

I ran into these issues but they do seem to be similar to my last issue so I assume they will be addressed in #6253 but I wanted to comment them just so they're known. 

https://github.com/user-attachments/assets/e0839f6a-0f96-4a62-841a-bcf7194ed93d



https://github.com/user-attachments/assets/3c410588-eb47-4fce-9f50-05fce2441127

PR #6249, opened 2025-02-17
Add uniqueIdentifier to storage table

### LINKED ISSUE #6247: Add `uniqueIdentifier` to Storage
Storage locations are identified inconsistently, leading to human error when assigning a storage location for preparations and when handling movements from one location to another.

We need to add a `uniqueIdentifier` field to the `storage` table. This `uniqueIdentifier` should be **manually assigned** rather than auto-generated, allowing us to uniquely identify storage locations. This will help ensure that all locations are globally uniquely identified, facilitating accurate and error-free movement of preparations.

There should be a uniqueness rule to make sure this is unique.

One alternative could be to continue using a combination of existing identifiers or repurposing a text field, but this may lead to inconsistencies and increased potential for human error. Another option could be to auto-generate `uniqueIdentifiers`, but we prefer manual assignment for better control and tracking. These `uniqueIdentifiers` are likely to be barcodes assigned using an external system, but it should be customizable so that field formats can be used in the application if necessary.

**Reported By**  
KU Mammals, Birds, and Herpetology, requested formally by @zzeppozz 

**Additional context**  
We are currently conducting regular meetings with KU Mammals, Birds, and Herpetology to discuss this enhancement and its implications for our workflow, and this will likely be used as an important part of a later mechanism to move preparations from one location to another simply by entering the starting `uniqueIdentifier` (using a barcode scanner) and ending `uniqueIdentifier` in the interface.

CarolineDenis:
Fixes #6247

> [!WARNING]  

### Checklist

### Testing instructions

- Open schema config 
- Select storage table 
- [ ] Verify the new field GUID is there 
- Create or edit your storage view def 
- Add guid field as not readonly field 
- Save 
- Open new storage data entry 
- Enter required info but not guid 
- [ ] verify you can save without a GUID 
- Enter a GUID (sequence of numbers or letters or mix ) 
- Copy the GUID (you will need it later)
- Save 
- [ ] verify you can save with guid 
- Create a new storage entry 
- Enter the GUID you copied 
- Save 
- [ ] Verify you cannot save because the GUID is not unique

### DIFF
--- a/specifyweb/specify/migrations/0024_add_guid_storage.py
+++ b/specifyweb/specify/migrations/0024_add_guid_storage.py
@@ -1,0 +1,41 @@
+from django.db import migrations, models
+import specifyweb.specify.models
+
+from specifyweb.specify.migration_utils.update_schema_config import revert_table_field_schema_config, update_table_field_schema_config_with_defaults
+from specifyweb.specify.migration_utils.sp7_schemaconfig import MIGRATION_0024_FIELDS as SCHEMA_CONFIG_MOD_TABLE_FIELDS
+
+def update_fields(apps):
+    Discipline = apps.get_model('specify', 'Discipline')
+
+    # Add guid -> storage
+    for discipline in Discipline.objects.all():
+        for table, fields in SCHEMA_CONFIG_MOD_TABLE_FIELDS.items(): 
+            for field in fields: 
+                update_table_field_schema_config_with_defaults(table, discipline.id, field, apps)
+
+def revert_update_fields(apps):
+    # Remove guid -> storage
+        for table, fields in SCHEMA_CONFIG_MOD_TABLE_FIELDS.items(): 
+            for field in fields: 
+                revert_table_field_schema_config(table, field, apps)
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('specify', '0023_update_schema_config_text'),
+    ]
+
+    def apply_migration(apps, schema_editor):
+        update_fields(apps)
+
+    def revert_migration(apps, schema_editor):
+        revert_update_fields(apps)
+
+    operations = [
+        migrations.AddField(
+            model_name='storage',
+            name='guid',
+            field=models.IntegerField(blank=True, db_column='Guid', default=0, null=True),
+        ),
+        migrations.RunPython(apply_migration, revert_migration, atomic=True)
+    ]
--- a/specifyweb/specify/migrations/0024_add_guid_storage.py
+++ b/specifyweb/specify/migrations/0024_add_guid_storage.py
@@ -1,4 +1,5 @@
 from django.db import migrations, models
+from specifyweb.businessrules.uniqueness_rules import DEFAULT_UNIQUENESS_RULES, create_uniqueness_rule
 import specifyweb.specify.models
 
 from specifyweb.specify.migration_utils.update_schema_config import revert_table_field_schema_config, update_table_field_schema_config_with_defaults
--- a/specifyweb/specify/migrations/0024_add_guid_storage.py
+++ b/specifyweb/specify/migrations/0024_add_guid_storage.py
@@ -15,9 +15,9 @@ def update_fields(apps):
 
 def revert_update_fields(apps):
     # Remove guid -> storage
-        for table, fields in SCHEMA_CONFIG_MOD_TABLE_FIELDS.items(): 
-            for field in fields: 
-                revert_table_field_schema_config(table, field, apps)
+    for table, fields in SCHEMA_CONFIG_MOD_TABLE_FIELDS.items(): 
+        for field in fields: 
+            revert_table_field_schema_config(table, field, apps)
 
 class Migration(migrations.Migration):
 
--- a/specifyweb/specify/migrations/0024_add_guid_storage.py
+++ b/specifyweb/specify/migrations/0024_add_guid_storage.py
@@ -35,7 +35,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='storage',
             name='guid',
-            field=models.IntegerField(blank=True, db_column='GUID', default=0, null=True),
+            field=models.CharField(blank=True, db_column='GUID', max_length=255, null=True),
         ),
         migrations.RunPython(apply_migration, revert_migration, atomic=True)
     ]
--- a/specifyweb/specify/migrations/0024_add_guid_storage.py
+++ b/specifyweb/specify/migrations/0024_add_guid_storage.py
@@ -35,7 +35,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='storage',
             name='guid',
-            field=models.IntegerField(blank=True, db_column='Guid', default=0, null=True),
+            field=models.IntegerField(blank=True, db_column='GUID', default=0, null=True),
         ),
         migrations.RunPython(apply_migration, revert_migration, atomic=True)
     ]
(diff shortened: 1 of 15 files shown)


acbentley:
Did we determine for this whether the users were asking for a user-defined GUID or an auto-filled "real" GUID? like in other tables?

grantfitzsimmons (changes requested):
>NOTES:
>To force uniqueness on the guid, add uniqueness rule in the schema config

GUID by its very definition must be globally unique, therefore we need to have a non-configurable system uniqueness rule for this

specifyweb/specify/migrations/0024_add_guid_storage.py
```diff
@@ -0,0 +1,41 @@
...
+    for discipline in Discipline.objects.all():
+        for table, fields in SCHEMA_CONFIG_MOD_TABLE_FIELDS.items(): 
+            for field in fields: 
+                update_table_field_schema_config_with_defaults(table, discipline.id, field, apps)
+
+def revert_update_fields(apps):
+    # Remove guid -> storage
+        for table, fields in SCHEMA_CONFIG_MOD_TABLE_FIELDS.items(): 
+            for field in fields: 
+                revert_table_field_schema_config(table, field, apps)
```
sharadsw, line 20: Indentation error?

acbentley:
> > NOTES:
> > To force uniqueness on the guid, add uniqueness rule in the schema config
> 
> GUID by its very definition must be globally unique, therefore we need to have a non-configurable system uniqueness rule for this

By that same definition, a GUID field should not be self-created as that too breaks (potentially) the GLOBALLY unique characteristic surely.

melton-jason:
Currently, only `CollectionObject -> guid` has a defined uniqueness rule and that rule is removable/modifiable by the user. 

https://github.com/specify/specify7/blob/0fa890a73279c8bb61d388c0210cb7f4397c490a/specifyweb/businessrules/uniqueness_rules.json#L65-L68

If it's a requirement, we can modify the existing `guid_rules` business rule (or add a new business rule) to enforce something like an 'implicit' uniqueness rule for guid (which would be non-configurable and not present in the schema config, unless we define the uniqueness rules for the other tables with the guid field). 

https://github.com/specify/specify7/blob/0fa890a73279c8bb61d388c0210cb7f4397c490a/specifyweb/businessrules/rules/guid_rules.py#L7-L16

It's not too difficult to search for duplicates within a single table: 
```py
# Assuming we're in some 'global' signal handler, like set_guids(sender, obj)
duplicates = sender.objects.only('id').filter(guid=obj.guid).exclude(id=obj.id)
if len(duplicates) > 0: 
    # raise a business rule exception following the expected uniqueness rule signature
```

If GUIDs should always be unique across all tables (e.g., no record should have the same GUID from any other record regardless of table) then things get a little more complicated, but conceptually I think we'd probably want to perform a join across all tables with the guid field (rather than applying the single case across applicable tables, to prevent hitting the database multiple times quickly).

acbentley:
> Currently, only `CollectionObject -> guid` has a defined uniqueness rule and that rule is removable/modifiable by the user.
> 
> https://github.com/specify/specify7/blob/0fa890a73279c8bb61d388c0210cb7f4397c490a/specifyweb/businessrules/uniqueness_rules.json#L65-L68
> 
> If it's a requirement, we can modify the existing `guid_rules` business rule (or add a new business rule) to enforce something like an 'implicit' uniqueness rule for guid (which would be non-configurable and not present in the schema config, unless we define the uniqueness rules for the other tables with the guid field).
> 
> https://github.com/specify/specify7/blob/0fa890a73279c8bb61d388c0210cb7f4397c490a/specifyweb/businessrules/rules/guid_rules.py#L7-L16
> 
> It's not too difficult to search for duplicates within a single table:
> 
> ```python
> # Assuming we're in some 'global' signal handler, like set_guids(sender, obj)
> duplicates = sender.objects.only('id').filter(guid=obj.guid).exclude(id=obj.id)
> if len(duplicates) > 0: 
>     # raise a business rule exception following the expected uniqueness rule signature
> ```
> 
> If GUIDs should always be unique across all tables (e.g., no record should have the same GUID from any other record regardless of table) then things get a little more complicated, but conceptually I think we'd probably want to perform a join across all tables with the guid field (rather than applying the single case across applicable tables, to prevent hitting the database multiple times quickly).

Is there no uniqueness rule inherently built into the way these are generated?

sharadsw (approved):

lexiclevenger (approved):
### Testing instructions

- [x] Verify the new field GUID is there 
- [x] verify you can save without a GUID 
- [x] verify you can save with guid 
- [x] Verify you cannot save because the GUID is not unique 
---
Looks great! GUID must be unique, and the uniqueness rule is non-configurable.
Screenshot 2025-02-19 123535

grantfitzsimmons:
@melton-jason @acbentley 

If you feel strongly about this, we could rename this something like "Unique Identifier" instead. That way, it’s clear that we’re aiming for uniqueness within just the storage tree, but we don’t expect these to be unique globally. 

These identifiers need to be unique within the institution, and really only unique for the storage tree. I'm not sure that these GUIDs are beneficial in any context other than managing the location of physical specimens and for object tracking.

Not sure we necessarily need to add logic to ensure that GUIDs are unique outside of specific contexts, such as Collection Object GUID, Taxon GUID, or Storage GUID. The chances of a repeat value appearing among these are very low, especially if they are auto-generated. If duplicates are manually entered between tables for any reason, it’s likely not a major issue, but I understand the concern. From an academic perspective, we would want to maintain the "globally unique" aspect.

For now, the bottom line is that we need to ensure that a field intended to be unique is enforced as such. I believe @CarolineDenis's solution is sufficient at this stage, though I welcome a name change if you both feel it is necessary.

grantfitzsimmons (approved):
### Testing instructions

- [x] Verify the new field GUID is there 
- [x] verify you can save without a GUID 
- [x] verify you can save with guid 
- [x] Verify you cannot save because the GUID is not unique 

Migrations all good:

```bash
~/GitHub/specify7 on issue-6247 = !8 ❯ docker exec -it specify7-specify7-1 /bin/bash                                                        at 08:48:11 PM
specify@2cf30bc205b3:/opt/specify7$ ve/bin/python manage.py migrate businessrules 0005
[20/Feb/2025 02:48:54] [ERROR] [specifyweb.attachment_gw.views:237] Bad attachment key.
Operations to perform:
  Target specific migration: 0005_cojo, from businessrules
Running migrations:
  Unapplying businessrules.0006_storage_guid... OK
specify@2cf30bc205b3:/opt/specify7$  ve/bin/python manage.py migrate specify 0023
[20/Feb/2025 02:49:26] [ERROR] [specifyweb.attachment_gw.views:237] Bad attachment key.
Operations to perform:
  Target specific migration: 0023_update_schema_config_text, from specify
Running migrations:
  Unapplying specify.0024_add_guid_storage... OK
specify@2cf30bc205b3:/opt/specify7$ 
```

👍

grantfitzsimmons (dismissed):
Requirement were updated during the usability meeting, will re-review once the field name changes.

lexiclevenger (approved):
- [x] Verify the new field GUID is there 
- [x] verify you can save without a GUID 
- [x] verify you can save with guid 
- [x] Verify you cannot save because the GUID is not unique 
---
Looks good! Everything is functional, and GUID is now named uniqueIdentifier.

Areyes42 (approved):
- [x]  Verify the new field GUID is there
-  [x]  verify you can save without a GUID
-  [x]  verify you can save with guid
-   [x] Verify you cannot save because the GUID is not unique
***
uniqueIdentifier looks good!
Screenshot 2025-02-21 at 12 20 26 PM

acwhite211 (approved):

specifysoftware:

PR #6252, opened 2025-02-17
Ensure Correct COT Selection in QB When Sharing Formatters

### LINKED ISSUE #5503: Catalog Number
**Describe the bug**
Related to https://github.com/specify/specify7/pull/5485

If multiple COTs share the same formatter the selection dropdown always chooses the first one with that format. It's a little weird but the functionality works since it is the same formatter after all.
I imagine in practice most people will have a different formatter for each COT if they aren't sticking with the default, but if someone comes across this it may appear broken.

**To Reproduce**

1. Create two CollectionObjectTypes with the same Catalog Number Format Name which differs from that of the Collection's default Catalog Number format
2. Create a new Collection Object query, add Catalog Number to the query, and switch the type from `Any` to `Equal`
3. Try to select both COTs, see that it selects the first COT regardless of which one you click.

**Expected behavior**
The COT you click on should be selected to avoid confusion. Or some sort of indicator that the behavior happens because they have the same formatter (not sure of an elegant way to do this)

**Screenshots**

https://github.com/user-attachments/assets/b8ed934c-a1c7-4952-a316-9ddda6bb2a1b

**Crash Report**

If the bug resulted in an error message, please click on "Download Error Message" and attach it here. If there is no error, please follow [these instructions](https://discourse.specifysoftware.org/t/download-specify-7-system-information/1614) to download your System Information.

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: production (SHA once https://github.com/specify/specify7/pull/5485 is merged)
- Database Name: KU_Birds
- User Name: spbirdadmin

Areyes42:
Fixes #5503

### Checklist

- [ ] Add automated tests

### Testing instructions

1. Create two CollectionObjectTypes with the same Catalog Number Format Name which differs from that of the Collection's default Catalog Number format
2. Create a new Collection Object query, add Catalog Number to the query, and switch the type from `Any` to `Equal`
3. Try to select both COTs
- [ ] Verify that it selects the proper COT regardless of the Catalog Number format


specifyweb/frontend/js_src/lib/components/QueryBuilder/Formatter.tsx
```diff
@@ -140,9 +140,24 @@ function FormatSelect({
...
 }): JSX.Element | null {
   const [formatterSelectIsOpen, setFormatterSelect] = React.useState(false);
-
+  const [selectedIndex, setSelectedIndex] = React.useState<number | null>(null);
   const id = useId('formatters-selection');
 
+  const compositeValue = React.useMemo(() => {
+    if (availableFormatters && availableFormatters.length > 0) {
+      if (selectedIndex !== null && availableFormatters[selectedIndex]) {
+        return `${availableFormatters[selectedIndex].name}|||${selectedIndex}`;
```
sharadsw, line 149: Since this string format is used in multiple places, can you make a function that accepts two params `name` and `index` and returns this generated string? It would help not only with reusability but also to document what the string represents

specifyweb/frontend/js_src/lib/components/QueryBuilder/Formatter.tsx
```diff
@@ -173,13 +188,28 @@ function FormatSelect({
             className={customSelectElementBackground}
             disabled={handleChange === undefined}
             id={id('list')}
-            value={currentFormat}
-            onValueChange={handleChange}
+            value={compositeValue}
+            onValueChange={(value) => {
+              const parts = value.split('|||');
```
sharadsw, line 193: In a similar vein, can you make `|||` a constant and reuse that here and in the function you create? It would help document the purpose of the string and keep any future changes more isolated.

sharadsw (changes requested):
I like your approach. Functionally your code looks good, just pointing out some minor code smells.

Areyes42:
> I like your approach. Functionally your code looks good, just pointing out some minor code smells.

Good point, I've added some helpers to separate the logic and hopefully make it a little more intuitive. Let me know if you think this looks better!

specifyweb/frontend/js_src/lib/components/QueryBuilder/Formatter.tsx
```diff
@@ -173,13 +203,26 @@ function FormatSelect({
...
+                setSelectedIndex(parsed.index);
+                if (handleChange) {
+                  handleChange(parsed.name);
+                }
+              } else {
+                setSelectedIndex(null);
+                if (handleChange) {
+                  handleChange(value);
+                }
+              }
```
sharadsw, line 219: (optional) some shorthands if you want to make it cleaner
```suggestion
              const index = parsed?.index ?? null;
              const name = parsed?.name ?? value;
              setSelectedIndex(index);
              handleChange?.(name);
```

sharadsw (approved):
Looks good, nice work!

lexiclevenger (approved):
- [x] Verify that it selects the proper COT regardless of the Catalog Number format

---

Looks good!


https://github.com/user-attachments/assets/ed8a1aae-88b1-4029-a8b4-6d12a72e2531

pashiav (approved):
- [X] Verify that it selects the proper COT regardless of the Catalog Number format
---
Works as expected!

melton-jason (commented):
Nice work on this! I like the implementation; pretty clever idea! 🚀 

I will mention that while the CollectionObjectType the user selects while the page is loaded will always be respected, it will always revert to the first CollectionObjectType in the list with the same format when re-rendered!

https://github.com/user-attachments/assets/3825793e-ca5c-45b4-89b6-675841d2215a

This is because only the format name (nothing of the CollectionObjectType) is saved to the `SpQueryField` in the database; when the page is rendered initially, it isn't formatted with the `formatterSeparator` and `index` and is just the name of the formatter itself. 
So Specify has no clue which CollectionObjectType it can be referring to. 

There's also a somewhat smaller Issue were changing the CollectionObjectType to another which shares the same format doesn't trigger a change on the field (because nothing has actually changed on the SpQueryField!) 

https://github.com/user-attachments/assets/9a72cc65-2823-4dcf-9dbe-7f65ec0f1a94

---

We're relatively having a lot of Issues regarding this feature (no thanks to my initial implementation in the first place!), and I think that's usually an indicator to take a step back and re-evaluate the core problem and potential solutions. 

@specify/ux-testing, @specify/dev-testing 

Instead of having a separate option for each CollectionObjectType, could we instead have an option for each Format and have the title of the option be a combination of all the COT names with that format? 

It could look something like this: 

Screenshot 2025-02-26 at 9 30 38 PM


Or we can go even more descriptive to try and be as clear as possible: the user _isn't_ selecting/filtering on COT, they're only applying a different formatter for the field: 

Screenshot 2025-02-26 at 9 54 29 PM

Alternatively, we could move away from the `SpQueryField -> FormatName` approach entirely and come up with something different which would allow the user to select an individual COT name.

emenslin:
I agree with Jasons comment and I think that grouping them together might make the most sense since it does not necessarily matter which specific one you use. I also think since we don't really have the gear other places aside from when you have something mapped as 'formatter' or 'aggregator' that adding 'Format As: ...' in this instance makes sense but we can go with something else if people have other opinions @specify/ux-testing

Areyes42:
@melton-jason @emenslin 

Great suggestions! I've updated the logic to group COTs with identical formatters together. This should mitigate issues related to re-rendering and state changes in the form, since it no longer compares against the same formatter. Also, I've added the "Format As:" identifier to improve clarity, but we can adjust it if it seems too confusing.

emenslin (approved):
- [x] Verify that it selects the proper COT regardless of the Catalog Number format

Looks good! I ran into an issue where the query results were incorrect (I set the filter to equal and instead it showed all results), I couldn't recreate it but I would love if we could get one more testing review to verify that it is not an issue

pashiav (approved):
- [x] Verify that it selects the proper COT regardless of the Catalog Number format=
--- 

Looks good! 

@emenslin I didn't run into what you mentioned, it's working on my end!

melton-jason (commented):
Nice work!
Coming along nicely 👌

CarolineDenis (approved):

PR #6261, opened 2025-02-20
Refactor useResourceValue hook 

### LINKED ISSUE #6183: CO form does not update immediately when adding COT with a different catalog number format
**Describe the bug**
After assigning a collectionObjectType that uses a different catalog number format from the collection default to a CO record, the Save button is disabled and must be clicked twice to save the record.

**To Reproduce**

1. Create a collection object type
2. Change the catalog number format to something different to the collection default and save
3. Create a new collection object
4. Assign the newly created collection object type
5. Try to save the collection object record
6. Click 'Save' twice and see that the record is saved

**Expected behavior**
The CO form should update immediately to prevent the save blocker from appearing.

**Screenshots**

https://github.com/user-attachments/assets/6ecd7c3a-c648-4cd4-b615-af96037690c7

**Crash Report**

[Specify 7 System Information - 2025-01-30T20_19_56.588Z.txt](https://github.com/user-attachments/files/18608275/Specify.7.System.Information.-.2025-01-30T20_19_56.588Z.txt)

- OS: Mac
- Browser: Chrome
- Database Name: geo_2024_12_05_1_production
- Collection name: Geology
- User Name: spadmin
- URL: https://geo202412051production-production.test.specifysystems.org/specify/view/collectionobject/9715/

CarolineDenis:
Fixes #6183 

### Checklist

- [ ] Add automated tests

### Testing instructions

A) Issue test 
- Create a collection object type
- Change the catalog number format to something different to the collection default and save
- Create a new collection object
- Assign the newly created collection object type
- Try to save the collection object record
- [ ] Verify the save button is enabled and you can save 

B) General test 

- [ ] Verify non-relationship fields of different types behave properly
- [ ] Verify to-one relationships rendered as a QueryComboBox behave properly
- [ ] Ensure the 'Field is required' save blocker is set when applicable
- [ ] Ensure the 'Please match requested format' save blocker is displayed when a value is invalid for a field according to its formatter
- [ ] Ensure fields have their defaults properly set in accordance with their formatter


lexiclevenger (approved):
- [x] Verify the save button is enabled and you can save 
- [x] Verify non-relationship fields rendered as a QueryComboBox behave properly
- [x] Verify to-one relationships rendered as a QueryComboBox behave properly
- [x] Ensure the 'Field is required' save blocker is set when applicable
- [x] Ensure the 'Please match requested format' save blocker is displayed when a value is invalid for a field according to its formatter
- [x] Ensure fields have their defaults properly set in accordance with their formatter
----
Looks good! However, changing the COType of an existing CO causes the new catalog number to save with the temporary value # instead of auto-incrementing.
 
https://github.com/user-attachments/assets/515ec8eb-d274-4ac4-a105-618b72a0de59

lexiclevenger (dismissed):
Meant to request changes

combs-a:
Question for testing; could you provide examples of non-relationship fields that can be rendered as a QCBX? Not sure of fields in the schema that matches that description. 😅

melton-jason:
> Looks good! However, changing the COType of an existing CO causes the new catalog number to save with the temporary value # instead of auto-incrementing.
>  
> https://github.com/user-attachments/assets/515ec8eb-d274-4ac4-a105-618b72a0de59

From https://github.com/specify/specify7/pull/6261#pullrequestreview-2631066591 by @lexiclevenger 

Nice catch! Though the explanation here is a little awkward, as there's a couple distinct parts of autonumbering which need explanation 😅

Firstly -as far as I can tell- autonumbering has ever only worked when creating new records, and not when updating old records. This would mean that if you were to use the default CatalogNumberNumeric formatter and save an existing record with a `#########` catalogNumber, then Specify wouldn't even try to autonumber the CollectionObject. 
I'm unsure if this was an intentional design decision (that existing records would not be autonumbered) or not. 
Of course, it is much simpler to only apply autonumbering to new records, so that might be the case (or part of it)? 

Secondly, technically the default pattern of the incrementing formatter is a valid value for the formatter! With CatalogNumberNumeric for example, `#########` _is_ a valid value for the format, so Specify sees nothing wrong with validating and saving a CollectionObject with the format. 
I think this goes hand-in-hand with the aforementioned assumption that only new records will have catalogNumbers which match the incrementing default pattern (`###`...). 

Here's a video recreating this exact Issue in `v7.9.6.2`!: 

https://github.com/user-attachments/assets/077f6fef-4283-4ef0-9ed5-67063bc0e239

Of course, with the changes relating to changing the catalogNumber when the CollectionObjectType changes this is probably more apparent. 

What would you suggest as a solution? 
Here are some potential solutions I came up with, any number of which can be implemented: 

- Allow autonumbering to apply to existing records
- Once a record exists in the database, treat `##`.. as an invalid value for the formatter
- For existing CollectionObjects, don't change the CatalogNumber when the COT changes

---

> Question for testing; could you provide examples of non-relationship fields that can be rendered as a QCBX? Not sure of fields in the schema that matches that description. 😅 

From https://github.com/specify/specify7/pull/6261#issuecomment-2675755258 by @combs-a 

Hmmm, yes there should be no non-relationships that can be rendered as a QCBX. 
I think the testing requirement should probably be "Verify non-relationship fields of different types behave properly" or "Ensure relationships rendered as a `field` cell are readonly and properly formatted/aggregated"? 
I've updated the testing instructions!

pashiav (approved):
- [X] Verify the save button is enabled and you can save 
- [X] Verify non-relationship fields of different types behave properly
- [X] Verify to-one relationships rendered as a QueryComboBox behave properly
- [X] Ensure the 'Field is required' save blocker is set when applicable
- [X] Ensure the 'Please match requested format' save blocker is displayed when a value is invalid for a field according to its formatter
- [X] Ensure fields have their defaults properly set in accordance with their formatter
---

Looks good! I tested on `calvertmarinemuseum20250206`.

I agree with @combs-a's comment. If not within the scope of this PR, should we create an issue and add it to a milestone?

sharadsw (approved):

melton-jason:
Haven't made any functional changes to this PR in recent commits (https://github.com/specify/specify7/pull/6261/files/034605aa29ac8a9f17a00bde10f6715b7dce05ad..e5977ae17e182790e4ba495d186614f21d7a3e14), just working on adding automated tests!

PR #6262, opened 2025-02-20
Detect if first login

### LINKED ISSUE #6211: Add a new way to detect if the user connects for the first time
For the new setup implementation in Sp7, we need to introduce a mechanism to detect whether a user is connecting for the first time. This will determine whether to display the traditional dashboard or the setup page accordingly.

CarolineDenis:
Fixes #6211

### Checklist

### Testing instructions
- Open a blank db (with no institutions) 
- [ ] Verify that you land on an empty screen displaying the message "Welcome! No institutions are available at the moment."


grantfitzsimmons (commented):
This code suggests that there may be multiple `institution` records in a single database, both in what is displayed to the user and in the API endpoint. Currently, I'm not aware of any plans to support more than one institution per database, but perhaps this decision was made with the understanding that this may not always be the case.

Is this intentional, or should it return only one institution at any given time?

specifyweb/specify/urls.py
```diff
@@ -13,6 +13,9 @@
     url(r'^specify/merge/status/(?P<merge_id>[0-9a-fA-F-]+)/$', views.merging_status),
     url(r'^specify/merge/abort/(?P<merge_id>[0-9a-fA-F-]+)/$', views.abort_merge_task),
 
+    #check for institution at login
+    url(r'^specify/institutions/$', views.get_all_institutions),
```
sharadsw: Can be removed after switching the api to call to `api/specify/institution`

specifyweb/frontend/js_src/lib/components/Login/index.tsx
```diff
@@ -22,9 +24,40 @@ import type { OicProvider } from './OicLogin';
...
+  );
+
   return React.useMemo(() => {
     const nextUrl = parseDjangoDump<string>('next-url') ?? '/specify/';
     const providers = parseDjangoDump<RA<OicProvider>>('providers') ?? [];
+
+    if (institutions?.length === 0 || institutions === undefined) {
+      // Display here the new setup pages 
+      return <p>Welcome! No institutions are available at the moment.</p>;
+    }
```
sharadsw: I would suggest having this block outside the memoized component that is returned. Looks like the memo is for identifying which login provider needs to be returned so I think this can go outside and would also eliminate the need for institution as a dependency

specifyweb/frontend/js_src/lib/components/Login/index.tsx
```diff
@@ -22,9 +24,40 @@ import type { OicProvider } from './OicLogin';
 import { OicLogin } from './OicLogin';
 
 export function Login(): JSX.Element {
+  const [institutions] = useAsyncState(
+    React.useCallback(
+      async () =>
+        ajax<RA<string>>(`/api/specify/institutions/`, {
```
sharadsw: Can be switched to `api/specify/institution`

grantfitzsimmons (approved):
### Testing instructions
- Open a blank db (with no institutions) 
- [X] Verify that you land on an empty screen displaying the message "Welcome! No institutions are available at the moment." 

Screenshot 2025-02-25 at 11 05 56 AM

CarolineDenis:
@grantfitzsimmons, 
updated to single institution

specifyweb/frontend/js_src/lib/components/Login/index.tsx
```diff
@@ -22,9 +24,40 @@ import type { OicProvider } from './OicLogin';
...
+  );
+
   return React.useMemo(() => {
     const nextUrl = parseDjangoDump<string>('next-url') ?? '/specify/';
     const providers = parseDjangoDump<RA<OicProvider>>('providers') ?? [];
+
+    if (institution === null) {
+      // Display here the new setup pages 
+      return <p>Welcome! No institutions are available at the moment.</p>;
+    }
```
sharadsw, line 59: I would suggest having this block outside the memoized component that is returned. Looks like the memo is for identifying which login provider needs to be returned so I think this can go outside to reduce future bloat if we add setup pages here and would also eliminate the need for institution as a dependency

specifyweb/specify/urls.py
```diff
@@ -13,6 +13,9 @@
     url(r'^specify/merge/status/(?P<merge_id>[0-9a-fA-F-]+)/$', views.merging_status),
     url(r'^specify/merge/abort/(?P<merge_id>[0-9a-fA-F-]+)/$', views.abort_merge_task),
 
+    #check for institution at login
+    url(r'^specify/institution/$', views.get_institution),
+
```
sharadsw, line 18: Creating a new route `api/specify/institution` or `api/specify/institutions` might not be a good idea since we already have a route `api/specify/institution` which returns institution data for an authorized user. Especially the route name without the s could potentially override the existing route and create side effects

How about a route called `is_new_user` instead which directly checks the number of institutions and returns a boolean?

specifyweb/specify/views.py
```diff
@@ -1367,3 +1367,9 @@ def parse_locality_set_foreground(collection, column_headers: List[str], data: L
         return 422, errors
 
     return 200, parsed
+
+
+#check for institution at login
+def get_institution(request):
+    institution = spmodels.Institution.objects.first()
+    return http.JsonResponse(institution if institution else None, safe=False)
```
sharadsw, line 1375: If we go with a route `is_new_user` instead, this function could simply calculate a boolean with something like
```
is_new_user = len(spmodels.Institution.objects.all()) == 0
```

sharadsw (changes requested):
Code looks fine functionally. However, I think we should reconsider the route we're creating, especially after [0f25eee](https://github.com/specify/specify7/pull/6262/commits/0f25eee241378cf2dfe851089430a35bf49af602) (sorry!)

specifyweb/frontend/js_src/lib/components/Login/index.tsx
CarolineDenis: I don't think we can as it would make the useMemo being called conditionally

sharadsw (approved):

acwhite211 (approved):

combs-a:
Could you add on more instructions or provide a completely blank database when you have time? I don't believe any of the UX testers have been told how to make a database like that, since typically we go through Sp6.

Issue #6248, opened 2025-02-17
Batch Edit: Catalog Number format validation does not follow COType's formatter


sharadsw:
Found in #5417 

From @combs-a:

> Also, I tried out COs that use COTs with the AlphaNumByYear type, and batch edit flagged the correct format as invalid, and incorrect formats as valid. Is fixing this within the scope of the PR? Wanted to check since the same thing works fine in the Workbench.
> 
> The incorrect format is AAAAAA, which was the default catno format before I changed it for FunnyTurtles.

![Image](https://github.com/user-attachments/assets/a4b8e1bf-c7a6-43c9-9dc1-6d2ac0977e26)

This happens because when relationships are disabled, validation for the formatter falls back to using the collection's default numeric formatter. This can be fixed by enabling relationships so that validation uses the correct formatter

Issue #6128, opened 2025-01-21
Batch Edit: Ensure rollback works as expected


sharadsw:
Component of #4929 

Deferring this to a new issue so we can test rollback on all parts of Batch Edit and not just for simple fields.

From #4929:
> ## Rollbacks
> 
> Rollbacks are complicated to perform. In the current design, whenever user creates a batch-edit dataset, via the query builder, it makes two datasets. User can only see one of them. The second is a "backer" of the first, and contains a FK to the first (so we can find backer of a dataset later). When rollback is requested, for every row in the main one, we find the original row in the backer, and perform the regular batch-edit update on it. Essentially, it applies original snapshot.
> 
> This is highly experimental, so it is recommended to always take a backup of the db, but this should work in a good amount of cases.

Notes from #5417:
- Rollback fails sometimes. In my testing so far, it fails when a dataset has picklists (might just be related to the picklist error mentioned in #5417) see: https://github.com/specify/specify7/pull/5417#pullrequestreview-2543815320

sharadsw:
Added #6416 as a subissue here

PR #6283, opened 2025-02-26
Batch edit for relationships

### LINKED ISSUE #6126: Batch Edit: Batch edit for relationships

sharadsw:
Fixes #6126, #6248

> [!WARNING]  
> Use a db used for testing other batch edit PRs or create a new one

Adding docs from #4929:

# Batch-editing

## Implementation and design

1. The current implementation uses workbench, and query builder. 
2. The workbench and batch-edit dataset are differntiated at the user level using a new "isupdate" field in spdataset table. DEV note: There is no difference at the code level -- everything is as general as possible. In fact, the isupdate field is only used at the code level to follow a special rollback procedure
3. The batch-edit datasets can seen using the batch-edit overlay accessible via the batch-edit menu item. Currently, the only way to create a new dataset is via the query builder interface.
4. To make a batch-edit dataset, go to the query builder, and add the relevant fields to the query. Some fields and relationships are not supported. Nested-to-many, for instance, are supported in workbench, but aren't in the batch-edit. Special fields like nodenumber and highestchildnodenumber, fullname field in tree are also not supported.
5. Other than relationships mentioned above, every field is supported. If an unsupported field like nodenumber is added, it is rendered as readonly. However, you cannot make nested-to-manys visible (this is different than being able to map it) --- you can map nested-to-manys, and even filter on them. As long as they are hidden, it'll not block. You can also arbitrarily add formatted and aggregated fields (they are unmapped, and are ignored)

## Batch edit behaviors

- Make a query with columns in the base table, and select relationships to edit. There are 4 different types of relationships, in general. Some example relationships for Collection Object as base table:
    - To-one dependent (for ex. collectionobjectattribute),
    - To-one independent (for ex. Cataloger, CollectingEvent [when not embedded])
    - To-many dependent (for ex. determinations, preparations),
    - To-many independent (for ex. None for CO as the base table)

- ### Fields
    The following fields are readonly. All other simple fields when changed will be updated
```
[
    "timestampcreated",
    "timestampmodified",
    "version",
    "nodenumber",
    "highestchildnodenumber",
    "rankid",
    "fullname",
    "age",
]
```
NOTE: Columns that map to `(formatted)` or `(aggregated)` values will also be readonly

- ### To-one dependent (for ex. collectionobjectattribute)
    These relationships get directly updated. If the to-one is not in the db, it'll create one. 
    This also includes collectingevent when embedded. <br/>
    <u>Test cases to consider</u>:
    - When mapped, the record is directly updated.
    - When mapped, if the record is not present, it'll be created, if non-empty values are present.
    - If the record previously had values, and the values are removed (making the cells completely empty), the to-one dependent record will be deleted. Since it is possible that there may be other fields in the database (but not in the query), we may accidentally delete the record. 
    - Example: user selected collectionobject -> collectionObjectAttribute -> remarks. And say they set remarks to empty in the spreadsheet, it is is possible that integer1 field in collectionObjectAttribute may have some value. To prevent accidental deletion, by default, we look at all the fields in the database for that record (other than system fields), to determine whether we can delete the record or not. This behavior is controlled by a preference. (described in a later section):

- ### To-many dependent (for ex. determinations)

    Same as to-one dependent. These relationships get directly updated. If the corresponding record is not present, a new one gets created. Does not consider matching records.

    <u>Test cases to consider</u>:
    - When mapped, the record is directly updated.
    - When mapped, if the record is not present, it'll be created, if non-empty values are present.
    - If the cell data is removed, and if every other field is empty in the database (can be disabled via a preference), the record will be deleted.

- ### To-one independent (for ex. cataloger)

    These relationships get matched, and uploaded (if a match is not found). During upload, it performs a clone of the record (cloning all the non-unique fields, and dependents). The clone takes into account relationships also mapped. That is, if agent needs to be cloned, and you have mapped `agentspecialty`, it'll use the mapped agentspecialty (rather than cloning previous's agentspecialty).

    <u>Test cases to consider:</u>

    - Start from a collectionobject with a cataloger, and map some fields. Change some of the values (like, say, lastname and firstname) to of agents that are present in the db. verify that agent gets matched. Note that the match can be performed with just the visible fields, or can also include fields in the database, not included in the query. This is controlled via a preference. By default, to be cautious in matching, it uses just the fields visible in the query.
    - If it is unable to match, it'll clone the existing agent, with data from the sheet. Make an agent with addresses/specialties/variants. Make sure the workbench is able to clone the agent correctly, and if you've provided some dependents in the mapping, it takes it.
    - Similar to workbench, you could customize the match behaviour by changing the matching options (like "never ignore", "ignore when blank", and "always ignore")

- ### To-many independent

    Same as to-many dependent. The only difference is that we always perform an update (we don't delete these). If a mapped record is not present, it'll create one, without any matching. 

    <u>Test cases to consider: </u>
    - Make collection objects, and assign them collectingevent. Do a query using collectingevent as the base table, and add fields of the CO table. Verify that resetting all fields does not delete the collection object (you'll also need to disable the preference that says to look at all fields for null checks)
    - If a record is not present, it'll create one, if there is a non-empty field.    

- ### Trees

    - Workbench method:<br/>

    - Update method: <br/>

    In both of the above methods, fullname, nodenumber, highestchildnodenumber is completely readonly. 
    
Additionally, tree relationship columns that use `(any rank)` are disabled in Batch Edit. Since `(any rank)` is only meant to be used for editing exist tree nodes, it is only editable when the base table of your query is a Tree and not when it is part of a relationship.

Example:
`Collection Object -> Determination -> Taxon -> (any rank) -> name` : **READONLY**
`Taxon -> (any rank) -> name` : **EDITABLE**

## Results

There are 4 new different type of results; 

image

- ### NoChange
Reported when the record was meant to updated, but no change occurred. That is, all the values from the db were the same. This is not visible to the user.

- ### Updated
Reported when the record's simple fields were changed. This does not consider relationships (they are reported with different result). 
NOTE: Sometimes an update could just come from an internal change. For example: For Locality records, the Workbench updates `lat1text` and `long1text` internally which Batch edit will record as an update to a simple field.

- ### Deleted
Reported when a record is deleted. Happens when a dependent relationship's cells are all empty (depending on the Batch Edit preference chosen for that dataset).

- ### MatchedAndChanged
Reported when a to-one independent was matched to an existing record, different than the current one.

- The side panel also shows the results per table, for different categories.

## Preferences

There are four different preference options.

- ### Batch Edit Preferences (2)

- <b> Defer For Match </b>
This preference controls whether database fields are included for matching or not. Defaults to true.

- <b> Defer For Null </b>
This preference controls whether database fields are included for determining if the record is null or not. For dependents, null records are deleted, so this preference is used to control the caution batch-edit follows. Defaults to false.

The preferences can be accessed from going to Data Mapper > Batch Edit Preferences
image

- ### User Preferences (1)

- <b> Number of query rows </b>

Determines how many number of query results are used for batch-edit. Defaults to 5000.

- <b> Show rollback button </b>
Can enable/disable rollback

image

## Rollbacks

Rollbacks are complicated to perform. In the current design, whenever user creates a batch-edit dataset, it makes two datasets internally. User can only see one of them. The second is a "backer" of the first, and contains a FK to the first (so we can find backer of a dataset later). When rollback is requested, for every row in the main one, we find the original row in the backer, and perform the regular batch-edit update on it. Essentially, it applies original snapshot.

This is highly experimental, so it is recommended to always take a backup of the db, but this should work in a good amount of cases.

### Known Limitations for Rollback
- Since rollback is a batch edit upload of the original dataset, rollback can only re-add data for the columns that were mapped in the original dataset

## Record Set

- Queries from record set are supported for batch editing

## Disallowed Tables
The batch edit button will be disabled for the following tables

### System tables
```
spuserexternalid
spattachmentdataset
uniquenessrule
uniquenessrulefield
message
spmerging
userpolicy
userrole
role
rolepolicy
libraryrole
libraryrolepolicy
spdataset
```

### Hierarchy tables
```
collection
discipline
division
institution
```

### Misc tables
```
spauditlog
```

### Scope Change Error
If a change in scoping attribute (collection, division, discipline etc.) is detected when validating/committing, a validation error will be raised for that row. Users must delete that row from the dataset before continuing. This can happen if the query that was used contained records from tables that are scoped at different levels and the workbench internally tries to change scope of a record.

For example:
- If you make a Collecting Event query with a mapping to Collecting Event -> CollectionObjects, you could run into this error
- This is because Collecting Event is scoped higher than collection level and so the query results contain COs that do not belong in the current collection
- When validating/committing the dataset, the workbench internally tries to scope records by the current collection leading to a scope change error

image

#### Exception
Scope change is allowed only for Loan records. This is because Loan records created from Interactions are scoped only at discipline level while Loan records created from the Workbench are scoped at discipline as well as division level. If you happen to batch edit Loan records that were created from Interactions, batch edit will internally add a division level scoping to such Loan records (since BE uses the Workbench internally). This exception allows Loan records to actually be batch edited without running into a scope change error

## Behavior with multiple trees
- The `name` field is required when querying on a tree rank
- When choosing a specific rank (and not `any rank`), Batch Edit expects to have all ranks lower than the chosen rank in the query
  - If the user does not add the "missing" fields, they will be prompted to pick the trees they wish to batch edit
  image
    - Users can pick which trees they want in case of conflicting ranks
    - Batch edit will only add ranks from selected trees
    - If no trees are selected, batch edit will use ALL trees
  - When selected ranks do not exist in multiple trees but there are missing ranks:
  image
  - Example: Say you have the following ranks for trees `Tree 1` and `Tree 2`: 
    - `Tree 1 -> Species` and `Tree 1 -> Subspecies`
    - `Tree 2 -> Species` (lowest rank)
  - And you make a query with:
    - `Tree 1 -> Species -> name`
  - Batch edit will implicitly add missing ranks (based on which trees were selected) and augment them to the dataset. The dataset will then look like
    - `Tree 1 -> Species -> name`
    - `Tree 1 -> Subspecies -> name`
    - `Tree 2 -> Species -> name`

### Batch editing a Tree path
- When editing a tree 'path' using a Tree as a base table with Batch Edit, the records are not edited directly. Rather the record is cloned and uploaded to the tree, which means functionally it will work the same as the workbench. 
- image
- When editing a path as a relationship (for eg: CO -> Determination -> Taxon -> (multiple ranks)), the complete edited path will either be matched and changed or there will be a new upload if there isn't a match

### Editing a single node
- A single node's simple fields can be edited using a query containing a `(any rank)` query field
image
- This can only be done when the base table itself is a tree. When the base table is not a tree and (any rank) query is part of a relationship, that column will be readonly.
image

### Checklist

- [x] Add automated tests

### Testing instructions

NOTE: Refer to the docs above for more info on tests

I would suggest testing on very small datasets so that it is easy to track what changes were made. For the general test section in the instructions below, feel free to use a larger dataset

#### Permissions
- [ ] Verify user with batch edit permissions can batch edit
- Create a user with no batch edit permissions
- [ ] Verify there is no batch edit button in Query results for that user

#### Readonly fields
```
[
    "timestampcreated",
    "timestampmodified",
    "version",
    "nodenumber",
    "highestchildnodenumber",
    "rankid",
    "fullname",
    "age",
]
```
- Create a batch edit dataset with any/all of the above fields
- [ ] Verify the fields are readonly

#### To-one dependent fields
- Map a column to a to-one dependent field (eg: CO -> CollectionObjectAttribute)
- Enter/edit data
- Validate/Commit
- [ ] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [ ] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)
  - You may need to choose the Batch Edit preference for `Use only visible fields for empty record check`

#### To-many dependent fields
- Map a column to a to-many dependent (eg: CO -> Determinations)
- Enter/edit data
- Validate/Commit
- [ ] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [ ] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)
 - You may need to choose the Batch Edit preference for `Use only visible fields for empty record check`. Refer the video below:
https://github.com/user-attachments/assets/313c1bc7-9213-479e-978f-820d2b99358e

#### To-one independent fields
- Map a column to a to-one independent (eg: CO -> Cataloger)
- Enter/edit data
- Validate/Commit
- [ ] Verify a new record (New Cell) is created if entered data does not match
- [ ] If entered data matches an existing record, verify cell is `Matched And Changed`
- Change some fields of the relationship in a row that already had a related record (eg: change last name of a cataloger of a CO)
  - [ ] Verify a new record was created and other fields of the new record are cloned from the original
  - Example: If you change the last name of a cataloger, a new Cataloger will be created with all fields of the previous one but with the new last name
- [ ] Clear all values of the relationship when it already exists and verify No Change is highlighted but the relationship is updated (you may have to change the null check preference)
  - NOTE: For independent relationships, the related record will not be deleted but the base table field will not longer be associated related record
  - Example:
    - Map a column to `CO -> Cataloger -> First Name` and `CO -> Cataloger -> Last Name`
    - Batch edit
    - For rows where a Cataloger already exists, clear `First Name` and `Last Name` cells
      - NOTE: In this example, you will need to enable the Batch Edit preference for empty record check.
    - Validate/Commit
    - After commit, the CO will no longer have a cataloger
    - The cell will not be highlighted with any kind of change
    - The original cataloger record will not be deleted from the db

#### To-many independent
- Map a column to a to-one independent (eg: Collecting Event -> Collection Objects)
- Enter/edit data
- Validate/Commit
- [ ] Verify a new record is created for all entered data
  - Example: If you enter catalog number for CE -> collectionObjects, they will all be highlighted as new cell
- [ ] Verify clearing values does not delete the record
- [ ] Verify matching does not occur

#### Tree queries
- Map a tree relationship (eg: CO -> Determination -> Taxon)
- Map a column to `(any rank)` and to a simple field of the tree
- Batch Edit
- [ ] Verify columns that used `(any rank)` are readonly

- Create a new dataset with columns that rank to specific ranks
- [ ] Test matching an entire Tree path
- [ ] If values are new a new node is created
- NOTE: Same behaviors as #6196 apply: If some values are new, nodes are cloned and updated instead of being updated directly

- [ ] Test missing ranks dialog with one tree relationship in the query
- [ ] Test missing ranks dialog with different trees in the relationships (eg: `CO -> Determination -> Taxon` and `CO -> CollectingEvent -> Locality -> Geography` in the same dataset)
- [ ] General test Batch Edit with different trees in the dataset

#### Cell Types
- [ ] Verify `Updated Cell` is highlighted only for simple fields of a record
- [ ] Verify `Matched And Changed` is highlighted when a data for a relationship matches an existing record (will depend on Batch Edit preferences)
- [ ] Verify `Deleted Cell` is highlighted when data is cleared for a relationship (will depend on Batch Edit preferences)
- [ ] Verify `New Cell` is highlighted when new data is entered for a relationship (like in the Workbench)

#### Preferences
- Go to Data Mapper > Batch Edit Preferences
- Use only visible fields for match:
  - [ ] When selected, verify a match happens with visible fields
  - [ ] When not selected, verify a match no longer happens with visible fields
  - [ ] (optional) When not selected, verify a match happens only if you match data for ALL fields (including hidden fields that contain data) of a relationship

- Use only visible fields for empty record check
  - [ ] When selected, verify cleared cells are deleted with visible fields
  - [ ] When not selected, verify cleared cells are no longer deleted with visible fields
  - [ ] (optional) When not selected, cleared cells are only deleted if you clear data for ALL fields (including hidden fields that contain data) of a relationship

#### User Preferences
- Change number of rows to use
- [ ] Verify the preference works
- Uncheck show rollback button
- [ ] Verify rollback is not visible in batch edit

#### General test
- [ ] General test batch edit with a combination of different field types in the dataset together

#### Rollback
- [ ] Test rollback with different base tables and types of relationships in the dataset
  - NOTE: If you run into rollback errors, make sure to note exactly what changes were made in the dataset before commit to make it easier to debug

#### BE from a Record Set query
- Query from a Record Set
- [ ] Test Batch Edit from the query

#### Disallowed tables
- [ ] Verify batch edit button is disabled for disallowed tables
- [ ] Verify there is a tooltip when you hover on the button

#### Test #6248
- Create a BE dataset with CO -> Catalog Number and CO -> CoType
- Change catalog number / cotypes
- [ ] Verify catalog numbers that do not follow the COType's formatter are marked as invalid


sharadsw:
TODO: Change upload plan construction for remote to ones

specifyweb/stored_queries/batch_edit.py
```diff
@@ -406,7 +406,7 @@ def _recur_row_plan(
 
         rel_type = (
             "to_many"
-            if node.type.endswith("to-many") or node.type == "zero-to-one"
+            if node.type.endswith("to-many") or node.type == "zero-to-one" or node.is_remote_to_one()
```
sharadsw, line 409: This took care of the upload plan changes needed for remote to-ones. However, changes made to COJO records don't get updated when committing. Changes made to a COJO record's `childCO` or `childCog` work fine too.

specifyweb/workbench/upload/clone.py
```diff
@@ -97,7 +97,7 @@ def _cloned(value, field, is_dependent):
             ).all()  # Clone all records separatetly
         ]
         for (field, is_dependent) in marked
-        if is_dependent and not field.concrete
+        if is_dependent and not field.concrete and not (field.one_to_one or field.many_to_one)
```
sharadsw, line 100: Had to add this because it tried to clone `childCO` and `childCog` records which are one-to-ones and would throw an error

sharadsw (commented):
Upload plan construction for COGs work but COJO records do not get cloned + updated. Any pointers on where I can look? @realVinayak

sharadsw:
TODO: Consider parsing 'fake' decimals to integer for lat/long values

sharadsw:
NOTE:
Test missing ranks error dialog when using different tree tables together in the query. 
Example:
- CO -> Determination -> Taxon
- CO -> CollectingEvent -> Locality -> Geography

pashiav (changes requested):
#### Permissions
- [X] Verify user with batch edit permissions can batch edit
- [X] Verify there is no batch edit button in Query results for that user

#### Readonly fields
- Create a batch edit dataset with any/all of the above fields
- [X] Verify the fields are readonly

#### To-one dependent fields
- [X] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [X] For a many-to-one (eg: CO -> CollectionObjectAttribute), if entered values matched an existing record, verify cells were `Matched And Changed`
- [X] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)

#### To-many dependent fields
- [X] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [ ] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)

#### To-one independent fields
- [X] Verify a new record (New Cell) is created if entered data does not match
- [X] If entered data matches an existing record, verify cell is `Matched And Changed`
- [X] Clear all values of the relationship when it already exists and verify No Change is highlighted but the relationship is updated

#### User Preferences
- Change number of rows to use
- [X] Verify the preference works
- Uncheck show rollback button
- [X] Verify rollback is not visible in batch edit

#### General test
- [X] General test batch edit with a combination of different field types in the dataset together

#### Rollback
- [X] Test rollback with different base tables and types of relationships in the dataset
  - NOTE: If you run into rollback errors, make sure to note exactly what changes were made in the dataset before commit to make it easier to debug

#### BE from a Record Set query
- Query from a Record Set
- [X] Test Batch Edit from the query



**Could not complete following tests due to errors (#1 in my comment):**
Tree queries
To-many independent
Cell Types
Disallowed tables
Test #6248

---

Great documentation btw!


---
**Issues**
**1**. Error occurs when querying any table (may take a couple tries but it is reproducible):
[Specify 7 Crash Report - 2025-04-07T14_02_44.596Z.txt](https://github.com/user-attachments/files/19633817/Specify.7.Crash.Report.-.2025-04-07T14_02_44.596Z.txt)

Query any table (e.g. CO) -> add any mapping (e.g. [catalogNumber](https://calvertmarinemuseumbatchedit20250210-issue-6126.test.specifysystems.org/specify/query/new/collectionobject/), Timestamp Created) -> Query 

`PR` | `Production`:

https://github.com/user-attachments/assets/efc289e5-7cfe-46e7-ae03-416f5b157e9b

**Note:** tested on on `cavertmarinemuseum` & `ciscollections`. 

- Immediately occurred on `calvertmarinemuseum`. 

- Initially did not occur on `ciscollections`, but appeared after some time of testing - no longer allowing querying at all.
**Link:** https://ciscollectionsbatchedit20250210-issue-6126.test.specifysystems.org/specify/query/new/collectionobject/
**Crash report:**
[Specify 7 Crash Report - 2025-04-07T18_39_13.950Z.txt](https://github.com/user-attachments/files/19637509/Specify.7.Crash.Report.-.2025-04-07T18_39_13.950Z.txt)



- Also tested `ciscollections` locally and got this error when querying (not consistent on local, but reproducible):
[Specify 7 Crash Report - 2025-04-07T17_47_02.769Z.txt](https://github.com/user-attachments/files/19636833/Specify.7.Crash.Report.-.2025-04-07T17_47_02.769Z.txt)
Screenshot 2025-04-07 at 1 39 49 PM



<br>

<br>

**2.** 
> To-many dependent fields
> - [ ] Clear all values of the relationship and verify the record was deleted (highlighted 
> as Deleted Cell)

Determinations is read-only. Is this expected?
https://ciscollectionsbatchedit20250210-issue-6126.test.specifysystems.org/specify/workbench/34


Screenshot 2025-04-07 at 11 43 14 AM

<br>

---

pashiav:
@sharadsw 

> - Can you try clearing cache and retrying for the crash error? 

Cleared cache on Specify, logged out, and cleared on browser - but still have the same result.

> - Which column was the readonly column mapped to? If it's `Determination -> (formatted)` then it is expected to be readonly as we cannot edit the format. 

Ah I must have selected that - thank you for the clarification!

sharadsw:
@pashiav Pushed a fix for the crash

sharadsw:
@emenslin Thank you for testing!
> When trying to delete a row for CO -> determination, you cannot delete the entire cell because ‘is Current’ is required in the schema config.
- For deleting a record from a cell, you may have to change your batch edit preferences (the PR notes have info on it). With the default preferences, a related record will only be deleted if ALL fields of that relationship exist in the dataset and have empty values. You can change this behavior by selecting 'Use visible fields for empty record check` in batch edit preferences

https://github.com/user-attachments/assets/abc7e053-4fb8-44c7-b3db-fdf9ac83dee4

> I got this error when testing. I’m not 100% sure how I got it and I cannot recreate it but I think it was when I was doing something in batch edit and then clicked on workbench.
- That's a strange bug. Does it exist after refreshing the page? I haven't been able to recreate it so far but I'll look into it if I run into it

> When I try to batch edit this query the rows will shift around and sometimes it looks like it only has 15 rows when there should be 300+ rows.
- The rows briefly not being visible seems like a general workbench bug. I think we have encountered it before as well. It will have to be fixed in a different PR since it's not caused by batch edit

> When I removed all the information for CO -> Cataloger it does not remove the cataloger from the CO.
- For me clearing a cataloger works fine but only on really small datasets. For larger datasets, it doesn't always seem to work. Working on a fix!

> When trying to test a CE -> CO data set, I got this error when trying to validate. Link to the data set:
- The issue here is that because Collecting Event is scoped higher than collection (scoped at discipline level), the query results will contain COs that do not belong to your current collection but do belong to your current collection's discipline. The CO with id `18721` actually belongs to Fossil Vertebrates rather than Fossil Invertebrates. When you commit the dataset, it detects a scope change for the CO's collection and blocks the update - which is expected behavior because scope change is not allowed in Batch Edit.

> I’m not sure if this error would be fixed by the batch edit trees pr but when testing CO -> det -> taxon, it seems like it might be checking for duplicate nodes in any tree, not just the taxon trees in the collection you are in. I’m not sure if it is throughout the entire db or what but there was a disambiguation error for a tree node that was not in the collection I was using.
- Looking into it. There's no disambiguation error when the mapping line has a specific rank. When the mapping line has `any rank` there is a disambiguation error possibly due to the mapping line not being associated with a specific tree/rank.

pashiav (commented):
#### Permissions
- [X] Verify user with batch edit permissions can batch edit
- [X] Verify there is no batch edit button in Query results for that user

#### Readonly fields
- Create a batch edit dataset with any/all of the above fields
- [X] Verify the fields are readonly

#### To-one dependent fields
- [X] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [X] For a many-to-one (eg: CO -> CollectionObjectAttribute), if entered values matched an existing record, verify cells were `Matched And Changed`
- [X] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)

#### To-many dependent fields
- [X] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [ ] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)

#### To-one independent fields
- [X] Verify a new record (New Cell) is created if entered data does not match
- [X] If entered data matches an existing record, verify cell is `Matched And Changed`
- [X] Clear all values of the relationship when it already exists and verify No Change is highlighted but the relationship is updated

#### User Preferences
- Change number of rows to use
- [X] Verify the preference works
- Uncheck show rollback button
- [X] Verify rollback is not visible in batch edit

#### General test
- [X] General test batch edit with a combination of different field types in the dataset together

#### Rollback
- [X] Test rollback with different base tables and types of relationships in the dataset
  - NOTE: If you run into rollback errors, make sure to note exactly what changes were made in the dataset before commit to make it easier to debug

#### BE from a Record Set query
- Query from a Record Set
- [X] Test Batch Edit from the query

**Tree queries
To-many independent
Cell Types
Disallowed tables
Test #6248**

---

✅ Crash fixed!

---
**Issues**

> #### To-many dependent fields
> - [ ] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)

I may be understanding it wrong, but cell is not highlighted as a deleted cell when values are cleared:
https://ciscollections20250210-issue-6126.test.specifysystems.org/specify/workbench/20 
Screenshot 2025-04-08 at 11 43 17 AM

---
Incomplete review but wanted to comment - will fully review later.

emenslin:
@sharadsw 

> For deleting a record from a cell, you may have to change your batch edit preferences (the PR notes have info on it). With the default preferences, a related record will only be deleted if ALL fields of that relationship exist in the dataset and have empty values. You can change this behavior by selecting 'Use visible fields for empty record check` in batch edit preferences

I get what you're saying here but wouldn't the error message make it unclear? In the video you sent the error says `Field required by schema config` and when you change the preference it fixes it, but with that error I don't think it's clear that the user would need to change the preference to solve the problem. Additionally this could still be unclear because you could have all the information for a determination in the data set but you would still get the error.

https://github.com/user-attachments/assets/8e0c0d43-a26c-4073-b665-bc8d155ac7c6


> That's a strange bug. Does it exist after refreshing the page? I haven't been able to recreate it so far but I'll look into it if I run into it 

I refreshed and it did go away. I brought it up to Grant when I ran into it and he seemed to somewhat know what it was so you might have to ask him for more information when he gets back. 

> The issue here is that because Collecting Event is scoped higher than collection (scoped at discipline level), the query results will contain COs that do not belong to your current collection but do belong to your current collection's discipline.

Is this something that can be fixed? I understand that you shouldn't be able to validate/commit the data set since it includes CO from other collections, but is there a way to catch it before the data set is made? My only thought is that this could happen to anyone who has shared CEs since shared CEs are scoped to discipline and someone might not realize they won't be able to batch edit their data set.

pashiav (commented):
#### Permissions
- [X] Verify user with batch edit permissions can batch edit
- Create a user with no batch edit permissions
- [X] Verify there is no batch edit button in Query results for that user

#### Readonly fields
- [X] Verify the fields are readonly

#### To-one dependent fields
- [X] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [X] For a many-to-one (eg: CO -> CollectionObjectAttribute), if entered values matched an existing record, verify cells were `Matched And Changed`
- [X] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)

#### To-many dependent fields
- [X] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [X] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)

#### To-one independent fields
- [X] Verify a new record (New Cell) is created if entered data does not match
- [X] If entered data matches an existing record, verify cell is `Matched And Changed`
  - [X] Verify a new record was created and other fields of the new record are cloned from the original
- [X] Clear all values of the relationship when it already exists and verify No Change is highlighted but the relationship is updated


#### Tree queries
- [X] If matched, cell is `Matched And Changed`
- [X] If not matched, cell is a `New Cell` with cloned values

- Create a new dataset with columns that rank to specific ranks
- [X] Test matching an entire Tree path
- [x] If values are new a new node is created

- [X] Test missing ranks dialog with one tree relationship in the query
- [X] Test missing ranks dialog with different trees in the relationships
- [X] General test Batch Edit with different trees in the dataset

#### Cell Types
- [X] Verify `Updated Cell` is highlighted only for simple fields of a record
- [X] Verify `Matched And Changed` is highlighted when a data for a relationship matches an existing record (will depend on Batch Edit preferences)
- [X] Verify `Deleted Cell` is highlighted when data is cleared for a relationship (will depend on Batch Edit preferences)
- [X] Verify `New Cell` is highlighted when new data is entered for a relationship (like in the Workbench)

#### Preferences
- Use only visible fields for match:
  - [X] When selected, verify a match happens with visible fields
  - [X] When not selected, verify a match no longer happens with visible fields

- Use only visible fields for empty record check
  - [X] When selected, verify cleared cells are deleted with visible fields
  - [X] When not selected, verify cleared cells are no longer deleted with visible fields
 
#### User Preferences
- Change number of rows to use
- [X] Verify the preference works
- Uncheck show rollback button
- [X] Verify rollback is not visible in batch edit

#### General test
- [X] General test batch edit with a combination of different field types in the dataset together

#### Rollback
- [X] Test rollback with different base tables and types of relationships in the dataset

#### BE from a Record Set query
- [X] Test Batch Edit from the query

#### Disallowed tables
- [X] Verify batch edit button is disabled for disallowed tables
- [x] Verify there is a tooltip when you hover on the button

#### Test #6248
- [X] Verify catalog numbers that do not follow the COType's formatter are marked as invalid

Didn't complete testing: **To-many independent**



---

**Comments**
---

While testing **Tree queries**:

I changed CO catnum **5565**'s determination taxon `name` to "testing" from "vesuvianite", but after commit, the record still has `fullname` as "vesuvianite". I understand that new records should have all the same values, but is this expected for `fullname`?
Committed BE: https://ciscollectionsbatchedit20250210-issue-6126.test.specifysystems.org/specify/workbench/40
Screenshot 2025-04-09 at 11 29 54 AM

https://ciscollectionsbatchedit20250210-issue-6126.test.specifysystems.org/specify/view/collectionobject/30/?recordsetid=105 
Screenshot 2025-04-09 at 11 40 33 AM


<br>

<br>

--
For:

> #### Cell Types
> - [ ] Verify `Updated Cell` is highlighted only for simple fields of a record

I checked it off, but thought I'd ask: are there specific fields we should be testing?

<br>

<br>

--
The **Batch Edit Preferences** button isn't disabled/read-only for uploaded BE Data Sets, so it causes an error on Save if modified.
1. Go to an uploaded BE Data Set (https://ciscollectionsbatchedit20250210-issue-6126.test.specifysystems.org/specify/workbench/plan/44/) -> Click Data Mapper
2. Click Batch Edit Preferences -> change preferences
3. Save -> See error
[Specify 7 Crash Report - 2025-04-09T17_05_49.900Z.txt](https://github.com/user-attachments/files/19671022/Specify.7.Crash.Report.-.2025-04-09T17_05_49.900Z.txt)

https://github.com/user-attachments/assets/f039e2c5-47e3-4d58-a88d-20a19a978ccf


Will this be addressed in this PR?

sharadsw:
@pashiav The checkboxes for the preferences are now disabled after upload. Thanks!

For your other questions:
- If `fullname` wasn't mapped in the dataset, then it will not be updated. Batch edit will only update the mapped columns
- There aren't any specific fields to test, feel free to general test any simple fields of any table/relationship

sharadsw:
@emenslin I did some more testing for the CO > Cataloger behavior. For clearing an independent relationship, you would need to enable the batch edit preference for empty record check to clear the relationship. Or map all fields of the relationship and clear all cells. However, since the relationship is independent it will not delete the related record and instead just clear it from the base record.

https://github.com/user-attachments/assets/4e790f96-94ff-401a-be95-18132cec373b

sharadsw:
TODO:
- [ ] Check if empty columns can be kept in the dataset
- [ ] Check if rollback prefs can be changed to delete on rollback
- [ ] Add localization for workbench failure errors (`resultsParser.ts`)
- [ ] Make scope change error a validation error?

sharadsw:
@emenslin This can be retested
Fixed:
- Scope change error is now a validation error. Rows with scope change can be deleted in a dataset
- Changed the placeholder string to a blank string
- Created a different issue for the disambiguation error for (any rank) queries: https://github.com/specify/specify7/issues/6413. I will continue working on that issue but it looks like it needs significant refactoring to fix. We will probably have to push it to a different milestone

sharadsw:
@specify/ux-testing There's a slight change in behavior for trees. All relationships that use a `(any rank)` tree query will be readonly in batch edit datasets. However, when the base table itself is a tree, (any rank) columns will still be editable and work like it does in #6196. This is because `(any rank)` queries should only be used for editing existing tree nodes which is not always possible in a relationship (eg: a CO -> determination may not always have a taxon value)

sharadsw:
@pashiav Thank you! This is a limitation in the current implementation of rollback. Since rollback is a batch edit upload of the original dataset, rollback can only re-add data for the columns that were mapped in the original dataset using the same upload plan. Since the taxon columns are readonly, they do not exist in the upload plan and rollback cannot add it back.

Fixing this could be quite big in scope since we would have to rework rollback and so I have created a new issue for it (https://github.com/specify/specify7/issues/6427) and also added it to the list of all rollback issues (https://github.com/specify/specify7/issues/6128)

pashiav (approved):
#### Permissions
- [X] Verify user with batch edit permissions can batch edit
- [X] Verify there is no batch edit button in Query results for that user

#### Readonly fields
- [X] Verify the fields are readonly

#### To-one dependent fields
- [X] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [X] For a many-to-one (eg: CO -> CollectionObjectAttribute), if entered values matched an existing record, verify cells were `Matched And Changed`
- [X] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)

#### To-many dependent fields
- [X] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [X] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)

#### To-one independent fields
- [X] Verify a new record (New Cell) is created if entered data does not match
- [x] If entered data matches an existing record, verify cell is `Matched And Changed`
  - [X] Verify a new record was created and other fields of the new record are cloned from the original
- [X] Clear all values of the relationship when it already exists and verify No Change is highlighted but the relationship is updated (you may have to change the null check 

#### To-many independent
- [X] Verify a new record is created for all entered data
- [X] Verify clearing values does not delete the record
- [X] Verify matching does not occur

#### Tree queries
- [x] Verify columns that used `(any rank)` are readonly
- [X] Test matching an entire Tree path
- [X] If values are new a new node is created
- [x] Test missing ranks dialog with one tree relationship in the query
- [x] Test missing ranks dialog with different trees in the relationships (eg: `CO -> Determination -> Taxon` and `CO -> CollectingEvent -> Locality -> Geography` in the same dataset)
- [x] General test Batch Edit with different trees in the dataset

#### Cell Types
- [x] Verify `Updated Cell` is highlighted only for simple fields of a record
- [x] Verify `Matched And Changed` is highlighted when a data for a relationship matches an existing record (will depend on Batch Edit preferences)
- [x] Verify `Deleted Cell` is highlighted when data is cleared for a relationship (will depend on Batch Edit preferences)
- [x] Verify `New Cell` is highlighted when new data is entered for a relationship (like in the Workbench)

#### Preferences
  - [x] When selected, verify a match happens with visible fields
  - [x] When not selected, verify a match no longer happens with visible fields
  - [x] When selected, verify cleared cells are deleted with visible fields
  - [x] When not selected, verify cleared cells are no longer deleted with visible fields

#### User Preferences
- [X] Verify the preference works
- [X] Verify rollback is not visible in batch edit

#### General test
- [X] General test batch edit with a combination of different field types in the dataset together

#### Rollback
- [x] Test rollback with different base tables and types of relationships in the dataset

#### BE from a Record Set query
- [X] Test Batch Edit from the query

#### Disallowed tables
- [X] Verify batch edit button is disabled for disallowed tables
- [X] Verify there is a tooltip when you hover on the button

#### Test #6248
- [x] Verify catalog numbers that do not follow the COType's formatter are marked as invalid
---

Great work - looks good! 

---

**Note**
Only on the test panel, clicking Queries in the nav bar causes there to be a `Failed to Fetch error` on `calvertmarinemuseum_2025_02_10` (not the batch edit db `calvertmarinemuseum_batch_edit_2025_02_10`). All other nav bar items work as expected.

**Link to db:** https://calvertmarinemuseum20250210-issue-6126.test.specifysystems.org/specify/overlay/queries/

[Specify 7 Crash Report - 2025-04-21T16_12_49.553Z.txt](https://github.com/user-attachments/files/19836608/Specify.7.Crash.Report.-.2025-04-21T16_12_49.553Z.txt)

https://github.com/user-attachments/assets/f8ed0d42-9d6f-4004-9458-d1c2dc9a1146



Locally, it's working as expected, so I'm approving this PR.

emenslin (changes requested):
#### Permissions
- [x] Verify user with batch edit permissions can batch edit
- [x] Verify there is no batch edit button in Query results for that user

#### Readonly fields
- [x] Verify the fields are readonly

#### To-one dependent fields
- [x] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [x] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)

#### To-many dependent fields
- [x] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [x] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)


#### To-one independent fields
- [x] Verify a new record (New Cell) is created if entered data does not match
- [x] If entered data matches an existing record, verify cell is `Matched And Changed`
- Change some fields of the relationship in a row that already had a related record (eg: change last name of a cataloger of a CO)
  - [x] Verify a new record was created and other fields of the new record are cloned from the original
- [x] Clear all values of the relationship when it already exists and verify No Change is highlighted but the relationship is updated (you may have to change the null check preference)


#### To-many independent
- [x] Verify a new record is created for all entered data
- [x] Verify clearing values does not delete the record
- [x] Verify matching does not occur

#### Tree queries
- [x] Verify columns that used `(any rank)` are readonly


- [x] Test matching an entire Tree path
- [x] If values are new a new node is created


- [x] Test missing ranks dialog with one tree relationship in the query
- [x] Test missing ranks dialog with different trees in the relationships (eg: `CO -> Determination -> Taxon` and `CO -> CollectingEvent -> Locality -> Geography` in the same dataset)
- [x] General test Batch Edit with different trees in the dataset

#### Cell Types
- [x] Verify `Updated Cell` is highlighted only for simple fields of a record
- [x] Verify `Matched And Changed` is highlighted when a data for a relationship matches an existing record (will depend on Batch Edit preferences)
- [x] Verify `Deleted Cell` is highlighted when data is cleared for a relationship (will depend on Batch Edit preferences)
- [x] Verify `New Cell` is highlighted when new data is entered for a relationship (like in the Workbench)

#### Preferences
Use only visible fields for match:
  - [x] When selected, verify a match happens with visible fields
  - [x] When not selected, verify a match no longer happens with visible fields
  - [x] (optional) When not selected, verify a match happens only if you match data for ALL fields (including hidden fields that contain data) of a relationship

- Use only visible fields for empty record check
  - [x] When selected, verify cleared cells are deleted with visible fields
  - [x] When not selected, verify cleared cells are no longer deleted with visible fields
  - [x] (optional) When not selected, cleared cells are only deleted if you clear data for ALL fields (including hidden fields that contain data) of a relationship

#### User Preferences
- [x] Verify the preference works
- [x] Verify rollback is not visible in batch edit

#### General test
- [x] General test batch edit with a combination of different field types in the dataset together

#### Rollback
- [x] Test rollback with different base tables and types of relationships in the dataset

#### BE from a Record Set query
- [x] Test Batch Edit from the query

#### Disallowed tables
- [x] Verify batch edit button is disabled for disallowed tables
- [x] Verify there is a tooltip when you hover on the button

#### Test #6248
- [x] Verify catalog numbers that do not follow the COType's formatter are marked as invalid

--- 

Rollback doesn't delete new taxon that were created when batch editing. I didn't get a chance to fully test it but it at least happens through CO -> determinations -> taxon, I'm not sure if it happens through other routes though. Other than that it looks good! 

https://github.com/user-attachments/assets/52155c92-d830-45b1-8ec1-e1cc1a9e77d9

sharadsw:
@emenslin Thanks for the review! The current implementation of rollback cannot delete a newly created independent relationship (Taxon in this case). Rollback is a batch edit upload of the original dataset. Batch edit can only delete dependent relationships, which means rollback can also only delete dependent relationships. I think we need to document this as a limitation rather than a bug.

With that being said, there is a bug right now with rolling back a dependent to-many but that issue is already written up here:  https://github.com/specify/specify7/issues/6416

sharadsw:
Created https://github.com/specify/specify7/issues/6451 and added it to the list of rollback limitations/bugs: https://github.com/specify/specify7/issues/6128

acwhite211 (approved):

emenslin (changes requested):
#### Permissions
- [x] Verify user with batch edit permissions can batch edit
- [x] Verify there is no batch edit button in Query results for that user

#### Readonly fields
- [x] Verify the fields are readonly

#### To-one dependent fields
- [x] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [x] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)


#### To-many dependent fields
- [x] Verify a new record (New Cell) is created for rows that had a empty record before edit
- [x] Clear all values of the relationship and verify the record was deleted (highlighted as Deleted Cell)


#### To-one independent fields
- [x] Verify a new record (New Cell) is created if entered data does not match
- [x] If entered data matches an existing record, verify cell is `Matched And Changed`
  - [x] Verify a new record was created and other fields of the new record are cloned from the original
- [x] Clear all values of the relationship when it already exists and verify No Change is highlighted but the relationship is updated (you may have to change the null check preference)

#### To-many independent
- [x] Verify a new record is created for all entered data
- [x] Verify clearing values does not delete the record
- [x] Verify matching does not occur

#### Tree queries
- Map a tree relationship (eg: CO -> Determination -> Taxon)
- [x] Verify columns that used `(any rank)` are readonly

- Create a new dataset with columns that rank to specific ranks
- [x] Test matching an entire Tree path
- [x] If values are new a new node is created


- [x] Test missing ranks dialog with one tree relationship in the query
- [x] Test missing ranks dialog with different trees in the relationships (eg: `CO -> Determination -> Taxon` and `CO -> CollectingEvent -> Locality -> Geography` in the same dataset)
- [x] General test Batch Edit with different trees in the dataset

#### Cell Types
- [x] Verify `Updated Cell` is highlighted only for simple fields of a record
- [x] Verify `Matched And Changed` is highlighted when a data for a relationship matches an existing record (will depend on Batch Edit preferences)
- [x] Verify `Deleted Cell` is highlighted when data is cleared for a relationship (will depend on Batch Edit preferences)
- [x] Verify `New Cell` is highlighted when new data is entered for a relationship (like in the Workbench)

#### Preferences
  - [x] When selected, verify a match happens with visible fields
  - [x] When not selected, verify a match no longer happens with visible fields
  - [x] (optional) When not selected, verify a match happens only if you match data for ALL fields (including hidden fields that contain data) of a relationship

- Use only visible fields for empty record check
  - [x] When selected, verify cleared cells are deleted with visible fields
  - [x] When not selected, verify cleared cells are no longer deleted with visible fields
  - [x] (optional) When not selected, cleared cells are only deleted if you clear data for ALL fields (including hidden fields that contain data) of a relationship

#### User Preferences
- [x] Verify the preference works
- [x] Verify rollback is not visible in batch edit

#### General test
- [x] General test batch edit with a combination of different field types in the dataset together

#### Rollback
- [ ] Test rollback with different base tables and types of relationships in the dataset


#### BE from a Record Set query
- [x] Test Batch Edit from the query

#### Disallowed tables
- [x] Verify batch edit button is disabled for disallowed tables
- [x] Verify there is a tooltip when you hover on the button

#### Test #6248
- [x] Verify catalog numbers that do not follow the COType's formatter are marked as invalid

--- 
I couldn't find an issue for it so I don't think this has been reported but, changing locality name in a CO batch edit prevents roll back.


https://github.com/user-attachments/assets/46ab730f-1e9c-46bd-b8f5-99785467eda3


Error: 
```
{
  "uploaderstatus": {
    "operation": "unuploading",
    "taskid": "df0e1c89-6bca-4e84-9dc1-a08350a5b60b"
  },
  "taskstatus": "FAILURE",
  "taskinfo": "RollbackFailure('Unable to roll back Collectingevent object (36060) because it is now referenced by another record.')"
```

sharadsw:
@emenslin Created a new issue for it: https://github.com/specify/specify7/issues/6468

sharadsw:
@emenslin I created an issue for it: https://github.com/specify/specify7/issues/6470 
It needs more investigation but looks like it happens when multiple ranks overlap between two trees. The frontend calculation of missing ranks is wrong in such cases.

PR #6290, opened 2025-02-28
Create ExpressSearchConfig static file and register with app_resources

### LINKED ISSUE #5470: TypeError upon logging in to a user without an `ExpressSearchConfig` app resource
**Describe the bug**
Users who do not have an `ExpressSearchConfig` app resource will experience an error toast immediately upon logging into `production` as of https://github.com/specify/specify7/commit/c75cf54c3e1c8ed39036950c278a52610a3616a7.

```
TypeError at /context/available_related_searches.json cannot unpack non-iterable NoneType object Request Method: GET Request URL: http://localhost/context/available_related_searches.json 
```

**To Reproduce**

1. Go to App Resources
2. Filter all resources until you only see "Express Search Config"s
3. Delete any `ExpressSearchConfig` app resource associated with your current user
4. Log out or clear your cache
5. See the error toast

**Expected behavior**

It should use the default express search config resource.

**Crash Report**

[Specify 7 Crash Report - 2024-12-12T21_53_32.665Z.txt](https://github.com/user-attachments/files/18117883/Specify.7.Crash.Report.-.2024-12-12T21_53_32.665Z.txt)

Encountered on the `geo_swiss` instance as well as on a local copy of `eurl`
macOS, Vivaldi

This does not happen on `v7.9.6.2`

acwhite211:
Fixes #5470

Create ExpressSearchConfig static file and register with app_resources.  This allows a default express search config file to be used if a user hasn't set one up.

### Checklist

### Testing instructions

- Login as a non-admin user that doesn't have a custom express search config setup.
- Open up Specify's main page
- [x] See that no error toasts appear in the wild

### DIFF
--- a/config/common/app_resources.xml
+++ b/config/common/app_resources.xml
@@ -2,4 +2,5 @@
 <files>
       <file level="3" mimetype="text/xml" name="WebLinks" description="Web Link Formats" file="weblinks.xml"/>
       <file level="3" mimetype="text/xml" name="InteractionsTaskInit" description="Interactions Sidebar Setup" file="interactionstask.xml"/>
+      <file level="3" mimetype="text/xml" name="ExpressSearchConfig" description="Express Search Config" file="expresssearchconfig.xml"/>
 </files>
--- a/config/common/expresssearchconfig.xml
+++ b/config/common/expresssearchconfig.xml
@@ -1,0 +1,671 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<search>
+  <tables>
+    <searchtable>
+      <tableName>Agent</tableName>
+      <displayOrder>0</displayOrder>
+      <searchFields>
+        <searchfield>
+          <fieldName>lastName</fieldName>
+          <isSortable>true</isSortable>
+          <isAscending>true</isAscending>
+          <order>3</order>
+        </searchfield>
+      </searchFields>
+      <displayFields>
+        <displayfield>
+          <fieldName>firstName</fieldName>
+          <formatter />
+          <order>1</order>
+          <isWebServiceField>false</isWebServiceField>
+          <isInitialInUse>false</isInitialInUse>
+        </displayfield>
+        <displayfield>
+          <fieldName>lastName</fieldName>
+          <formatter />
+          <order>0</order>
+          <isWebServiceField>false</isWebServiceField>
+          <isInitialInUse>false</isInitialInUse>
+        </displayfield>
+      </displayFields>
+      <wsFields />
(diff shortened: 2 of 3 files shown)


CarolineDenis (approved):

grantfitzsimmons (approved):
Approving this, but with three important notes:

1. The testing checklist here does not verify that behavior has not been modified between 7.9.6.2 and 7.10, which may leave us open to issues if we inadvertently break this behavior.

1. Someone in @specify/developers  should open a new issue to fix the fact that this (`available_related_searches.json`) file is not being automatically generated. 

	https://github.com/specify/specify7/blob/production/specifyweb/frontend/js_src/lib/tests/ajax/static/context/available_related_searches.json
	
	My understanding is that before 7.10, Specify would automatically create this file based on the list of available express search options. See below:
	
	https://github.com/specify/specify7/blob/56bebce77488a13783e9d43782330d348a7f5896/specifyweb/context/views.py#L412-L425
	
	These 'related searches' are defined in this file:
	
	https://github.com/specify/specify7/blob/56bebce77488a13783e9d43782330d348a7f5896/specifyweb/express_search/related_searches.py#L25-L37
	
	This means it would be theoretically quite simple to extend the default simple search behavior using the old approach, but with this patch (does not fix the underlying issue, simply provides a default static file), it adds another level to the complication.

2. Before merging, can you explain how was this default `expresssearchconfig.xml` sourced? I'm concerned it may change the default searching behavior in the app.

	Upon removing all `ExpressSearchConfig` app resources in a database and logging in through Specify 6, I get this which seems to be as "default" as it comes:
	
	[DefaultExpressSearchConfig.txt](https://github.com/user-attachments/files/19054962/DefaultExpressSearchConfig.txt)
	
	The file in this PR is nearly [twice the size and is structured differently](https://raw.githubusercontent.com/specify/specify7/issue-5470/config/common/expresssearchconfig.xml).

sharadsw (approved):

acwhite211:
@grantfitzsimmons Let me know if there is a different file to use for the default ExpressSearchConfig.  I'll into the automatic generation of the file.

pashiav (approved):
- [x] See that no error toasts appear in the wild

---
Working on `eurl31424` now!

Noting here that I couldn't reproduce the original issue #5470 on `calvertmarinemuseum` nor `ciscollections` on `production`.

bronwyncombs (approved):
- [x] See that no error toasts appear in the wild

tested logged in with specify user 'test' on `nmb_rinvert_2024_08_01`

PR #6324, opened 2025-03-14
Improve `DataObjFormatters` Handling for New Tables

### LINKED ISSUE #6243: Support Default `DataObjFormatters` Alongside Custom Configurations
image

Currently, users who implement a custom `DataObjFormatters` file within App Resources face a challenge when new tables are introduced. 
It requires a manual update which not only adds extra workload but also creates inconsistencies in user experience, particularly when new schema expansions occur, but also leaves our team unable to update defaults when they need to change.

A major problem here is that the `DataObjFormatters` app resource (see https://github.com/specify/specify7/blob/production/config/backstop/dataobj_formatters.xml) is **nearly always** custom defined in databases since it is nearly always necessary to edit the table formats and aggregation for how data fields appear on labels, queries, and exports.
Once customized, Specify looks at this custom resource rather than the system one.

This leaves users with custom `DataObjFormatters` unable to use the query combo box, query builder, and other query mechanisms to display things like the Collection Object Group (formatted), COG children (aggregated), Tectonic Unit (formatted), Collection Object Type (formatted), or any new tables/fields that we add in the future. The only workaround is for users to click on the 🔍 (search) icon next to the query combo box to search manually for records and search specific fields from there.

## Problem Statement

When new fields are added to the system, such as the recent addition of geo-related fields, users with a custom `TypeSearches` file must manually update their configurations to ensure that new fields, including `cojo` and `taxonTreeDef`, are searchable. This is particularly problematic for users who expect seamless integration and functionality with the system's default search capabilities.

### Current Limitations:
-  Users with a custom `DataObjFormatters` file (almost every collection) will not automatically receive updates or expansions related to default table formats and aggregation.
-  **Specifically affected tables include:**
	  - `specifyuser`
	  - `collectionobjectgroup`
	 - `collectionobjecttype`
	  - `collectionobjectgrouptype`
	  - `tectonicunit`
	  - `tectonicunittreedef`
	  - `tectonicunittreedefitem`

## Proposed Solution

Implement a logic that automatically fills in the gaps of custom `DataObjFormatters` files with the system's default settings. This would ensure that any new fields or tables added to the system are automatically.

This proposal does **not** involve updating the custom XML, rather just allowing Specify to "fall back" on a system default rather than looking only at the custom definition for every table.

### Current Workaround

Users can define custom `DataObjFormatters` table formats and table aggregations for new tables manually, but they can't use our defaults.

### Benefits:
-  **Enhanced User Experience:** Users will benefit from seamless access to new tables in queries, QCBXs, and on the forms without needing to update their custom configurations.
-  **Reduced Maintenance:** This would significantly decrease the maintenance burden on admin users who regularly need to update their `DataObjFormatters` files to be able to use **basic** functions of the app.
-  **Consistency Across Systems:** Ensures that all users, regardless of their custom configurations, have access to the same preview capabilities and functionalities.

Areyes42:
Fixes #6243

Refer to the original issue for more detailed context, but essentially this PR implements a fallback mechanism for DataObjFormatters that allows users with custom configurations to merge with default presets. This change should let users with custom DataObjFormatters now automatically access new table formatters added to the system without needing to manually update their configs.

**Specifically affected tables include:**
-   `specifyuser`
-   `collectionobjectgroup`
-   `collectionobjecttype`
-   `collectionobjectgrouptype`
-   `tectonicunit`
-   `tectonicunittreedef`
-   `tectonicunittreedefitem`

### Checklist

- [ ] Add automated tests

### Testing instructions

1. Navigate to App Resources → Record Formatters
2. Make a backup of the current DataObjFormatters
3. Remove a table formatter from the custom DataObjFormatters
4. Open a query on one of the table formatters that you deleted
5. Run the query with '(formatted)' selected
    - [ ] Verify the records are displayed with proper formatting
    - [ ] Repeat these steps with all the affected tables
6. Navigate to a form that has a query combo box for the affected tables
    - [ ] Verify records are properly formatted when using the search button in the query combo box
7. Go back to the record formatters and define a new custom table formatter
    - [ ] Verify the new custom formatter overrides the system formatter


emenslin (changes requested):
- [x] Verify the records are displayed with proper formatting
   - [ ] Repeat these steps with all the affected tables
   - [x] Verify records are properly formatted when using the search button in the query combo box
   - [x] Verify the new custom formatter overrides the system formatter 

--- 

`TectonicUnitTreeDef` and `TectonicUnitTreeDefItem` don't have a default format, everything else looks good! 

image

Areyes42:
> * [ ]  Repeat these steps with all the affected tables
> 
> `TectonicUnitTreeDef` and `TectonicUnitTreeDefItem` don't have a default format, everything else looks good!

I added those two tables to the default dataobj formatters, so they should be good now.

specifyweb/context/app_resource.py
```diff
@@ -58,6 +58,42 @@ def get_app_resource(collection, user, resource_name, additional_default=False):
...
+                custom_formatter = from_fs
+                break
+        
+        # If custom formatter not found, try to load the default formatter from the backstop level
+        default_formatter = load_resource_at_level(collection, user, 'Backstop', resource_name)
+        
+        # If both custom and default formatters are found, merge them
+        if custom_formatter is not None and default_formatter is not None:
+            merged_formatter = merge_formatters(custom_formatter[0], default_formatter[0])
+            return merged_formatter, custom_formatter[1], custom_formatter[2]
```
sharadsw, line 86: Just something to note here: 
One possible edge case here is that we merge formatters even if the `custom_formatter` was the resource found at the Backstop level (if `from_fs` was not None)

It seems `merge_formatters` does ensure we do not duplicate existing formatters so that's good! Although I do not think we would ever run into a scenario like that since most users will have a DataObjFormatter resource higher than Backstop anyway

specifyweb/context/app_resource.py
```diff
@@ -58,6 +58,42 @@ def get_app_resource(collection, user, resource_name, additional_default=False):
     logger.info('looking for app resource %r for user %s in %s',
                 resource_name, user and user.name, collection and collection.collectionname)
 
+    # Handling for DataObjFormatters to support fallback to defaults
+    if resource_name == 'DataObjFormatters':
```
sharadsw, line 62: I wonder if this can be generalized to account for all app resources? It doesn't have to be done in this PR since it would extend the scope quite a bit but it's just something to think about since this function already has a different behavior for TypeSearches

specifyweb/context/app_resource.py
```diff
@@ -58,6 +58,42 @@ def get_app_resource(collection, user, resource_name, additional_default=False):
...
+        # Return the custom formatter if it exists, otherwise return the default
+        if custom_formatter is not None:
+            return custom_formatter
+        
+        if default_formatter is not None:
+            return default_formatter
+        
+        return None
+    
     if not additional_default:
```
sharadsw, line 97: Some context on what I mentioned about TypeSearches:
The `additional_default` prop was added so that the frontend can directly query the resource at the Backstop level. For TypeSearches the frontend makes 2 queries and has logic for figuring out the fallback TypeSearch.

I do like the merge approach a lot more so this is something we will want to change if we ever look to generalizing the fallback logic

sharadsw (approved):
I think the changes here look good! My comments are just for discussion on trying to generalize fallback for all app resources instead of having different logic for different ones

emenslin (approved):
- [x] Verify the records are displayed with proper formatting
   - [x] Repeat these steps with all the affected tables
   - [x] Verify records are properly formatted when using the search button in the query combo box
   - [x] Verify the new custom formatter overrides the system formatter 


Looks good, tectonic unit issue is fixed!
image

CarolineDenis (approved):

pashiav (approved):
- [X] Verify the records are displayed with proper formatting
- [X] Repeat these steps with all the affected tables
- [X] Verify records are properly formatted when using the search button in the query combo box
- [X] Verify the new custom formatter overrides the system formatter 

---
Looks good! Working as expected.

PR #6337, opened 2025-03-19
Disable tree actions in view mode

### LINKED ISSUE #5007: Tree actions remain accessible in read-only mode
**Describe the bug**
In read-only mode, users with tree permissions can still access the tree actions they have roles for (e.g. move, merge, synonymize).

**To Reproduce**

1. Login to or create a user that has tree permissions
2. Go to a tree
3. Enable read-only mode
4. See that you can still access tree actions

**Expected behavior**
You should only be able to search or view tree nodes in read-only mode.

**Screenshots**

https://github.com/specify/specify7/assets/130881183/0dda02ef-1624-4810-8baf-21abd722213e

What it should be:

- OS: macOS 14.5
- Browser: Chrome
- Specify 7 Version: Tested in edge & PR #5001
- Database Name: KUFish_5_16_23
- User Name: spfishadmin & testread

CarolineDenis:
Fixes #5007

### Checklist

### Testing instructions

- Login to or create a user that has tree permissions
- Go to a tree
- Enable read-only mode
- [ ] Verify tree actions are disabled

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
@@ -76,6 +76,8 @@ export function TreeViewActions<SCHEMA extends AnyTree>({
   const disableButtons =
     focusedRow === undefined || typeof currentAction === 'string';
 
+  const isViewMode = !hasTablePermission(tableName, 'update');
+
   return currentAction === undefined ||
     actionRow === undefined ||
     focusedRow === undefined ||
@@ -117,7 +119,7 @@ export function TreeViewActions<SCHEMA extends AnyTree>({
           disabled={focusedRow === undefined}
           isRoot={isRoot}
           label={
-            hasTablePermission(tableName, 'update')
+            isViewMode
               ? commonText.edit()
               : commonText.view()
           }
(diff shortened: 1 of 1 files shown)


specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
```diff
@@ -117,7 +119,7 @@ export function TreeViewActions<SCHEMA extends AnyTree>({
           disabled={focusedRow === undefined}
           isRoot={isRoot}
           label={
-            hasTablePermission(tableName, 'update')
+            isViewMode
               ? commonText.edit()
               : commonText.view()
```
sharadsw, line 124: Currently, the button says Edit instead of View when read only mode is turned on

specifyweb/frontend/js_src/lib/components/TreeView/Actions.tsx
```diff
@@ -76,6 +76,8 @@ export function TreeViewActions<SCHEMA extends AnyTree>({
   const disableButtons =
     focusedRow === undefined || typeof currentAction === 'string';
 
+  const isViewMode = !hasTablePermission(tableName, 'update');
+
```
sharadsw, line 80: Just to make things consistent with other parts of Specify, I wonder if we could use `ReadonlyContext` here?
We can do something like
```
```
and then use `isReadOnly` instead of `isViewMode`.

Functionally nothing will change but it will just ensure we handle readonly mode in the same way everywhere.

emenslin (approved):

sharadsw (approved):

acwhite211 (approved):

grantfitzsimmons (approved):
- [X] Verify tree actions are disabled

image

PR #6344, opened 2025-03-19
Batch Edit: Use scoped picklist for formatting in batch edit

### LINKED ISSUE #6339: Batch edit picklist using values
**Describe the bug**
If you create a batch edit data set with a picklist it uses the value instead of the title. This doesn't seem to be consistent with every picklist but I'm not sure what makes it use the values or not. For example, on ojsmnh20250314 with an accession data set the 'type' picklist will show the values but the 'status' picklist shows the titles.

**To Reproduce**

1. Go to queries
2. Select any base table that has picklists (e.g. accession) 
3. Add picklist(s) to query
4. Batch edit 
5. See some picklist titles changed to the value name

**Expected behavior**
Picklists should always use the title and not the value

**Screenshots**

https://github.com/user-attachments/assets/79577752-ba13-4033-a28e-578cc80fe172

- OS: Windows 11
- Browser: Chrome
- Database Name: ojsmnh20250314
- Collection name: fossil invert
- User Name: spadmin
- URL: https://ojsmnh20250314-production.test.specifysystems.org/specify

sharadsw:
Fixes #6339

Notes in https://github.com/specify/specify7/issues/6339#issuecomment-2738060575

### Checklist

- [ ] Add automated tests

### Testing instructions

- On `ojsmnh` use collection Fossil Invertebrates
- Make an Accession Query with fields Type and Status
- Click Batch Edit
- [ ] Verify picklist titles are being used
   - NOTE: Other collections may not always use picklist titles due to problems mentioned here: https://github.com/specify/specify7/issues/6339#issuecomment-2738060575
- [ ] General test picklists in Batch Edit with other tables 
- [ ] General test picklists in Workbench

### DIFF
--- a/specifyweb/specify/utils.py
+++ b/specifyweb/specify/utils.py
@@ -94,5 +94,8 @@ def get_picklists(collection: spmodels.Collection, tablename: str, fieldname: st
     picklists = None
     if len(schema_items) > 0 and schema_items[0].picklistname:
         picklists = spmodels.Picklist.objects.filter(name=schema_items[0].picklistname)
+        collection_picklists = picklists.filter(collection=collection)
+        if len(collection_picklists) > 0:
+            picklists = collection_picklists
 
     return picklists, schemaitem
--- a/specifyweb/workbench/upload/scoping.py
+++ b/specifyweb/workbench/upload/scoping.py
@@ -116,15 +116,8 @@ def extend_columnoptions(
     toOne = toOne or {}
     picklists, schemaitem = get_picklists(collection, tablename, fieldname)
 
-    if not picklists:
-        picklist = None
-    else:
-        collection_picklists = picklists.filter(collection=collection)
-
-        picklist = (
-            picklists[0] if len(collection_picklists) == 0 else collection_picklists[0]
-        )
-
+    # Picklists are already scoped by collection in get_picklists if possible
+    picklist = None if picklists is None else picklists[0]
 
     ui_formatter = get_or_defer_formatter(collection, tablename, fieldname, row, toOne, context)
     scoped_formatter = (


CarolineDenis (approved):

acwhite211 (approved):

bronwyncombs (approved):
- [x] Verify picklist titles are being used
Screenshot 2025-03-20 at 11 37 41 AM
Screenshot 2025-03-20 at 11 38 03 AM
Screenshot 2025-03-20 at 11 39 53 AM

- [x] General test picklists in Batch Edit with other tables 

Screenshot 2025-03-20 at 11 49 10 AM

- [x] General test picklists in Workbench
Screenshot 2025-03-20 at 11 54 18 AM

I also want to comment on an existing behavior that came up during testing but already exists. If if you save a mapping, proceed to the data set interface, then later go back to data mapper and change fields and/or the base table, the column titles reflect the initial fields mapped to the columns in the data mapper and data set. However, validation still catches "illegal" pick list values and reflects the updated options so this seems like a front end only discrepancy.

melton-jason:
Nice work! 😎 
No new problems here, just further documenting/explaining the caveats and drawbacks of the current solution, some potential fixes, and how the current solution might still be the best without fundamentally changing how PickLists work and are scoped. 

> - Accession is scoped at a higher level than collection - which means Queries will display the same Accession records regardless of which collection you are in.
>   - If we scope picklist formatting by collection, there will be some collections where the formatter doesn't look at the 'correct' picklist for Accession because each collection has a different one
> - In other tables, scoping picklist formatter by collection will not cover cases where a user somehow has an invalid entry in a picklist field for that collection

From https://github.com/specify/specify7/issues/6339#issuecomment-2738060575 by @sharadsw 

Yeah, not sure the best way to handle this would be? If we want to keep PickLists scoped at the Collection level, our options are pretty limited. If this is something we want to address, I imagine we could maybe try: 

- (For tables scoped above Collection?) Keeping the initial PickList query filtering as unscoped as possible on the BackEnd, and try to aggregate/merge the PickLists. In other words, titles/values not present in one PickList could be "merged" into the other and the aggregated PickList items be used for validation purposes
  - Although the "merge" process itself could be very complicated, especially when considering case sensitivity (i.e., is `collection=Collection` in most cases?) , possible separators between words (space vs. underscore vs. hyphen), etc. 
  - Essentially, this would be expanding a "valid" value for a field with a PickList to be all values for PickLists across Collections (within the same Discipline? Or among all PickLists with the same name for PickLists assigned to the field?) 

To demonstrate/document the current behavior (in the QB as well, which was fixes as well with this change), I have two Collections in the same Discipline: `Fish` and `Fish2`, and will be considering an `AccessionType` picklist set on the field `Accession -> type`, which is defined with different titles + values in each Collection (both are largely based from the default AccessionType picklists with small modifications) 

The `AccessionType` PickList in the `Fish` Collection: 

| Title      | Value      |
|------------|------------|
| Field Work | field_work |
| Collection | cln        |

`AccessionType` in the `Fish2` Collection: 

| Title        | Value        |
|--------------|--------------|
| Field Work   | field_work   |
| Collection   | collection   |
| Finding Nemo | I_shall_call_him_squishy |

_A Query from the Fish Collection_

fish_collection_query

_A Query from the Fish2 Collection_

fish2_collection_query

_Initiating a Batch Edit from the Fish Collection_


https://github.com/user-attachments/assets/108c7298-4c6a-4e35-990f-0f63d523756d

This example is relatively tame - and probably the most common- compared to what is possible (and technically valid)  in Specify. Imagine a case where one or more PickLists across Collections have different types! For Example, one of the Collection's AccessionType PickList could instead be of type `Value From Field`, or even of type `Entire Table` with a formatter.

The current solution is simple (albeit while being very restrictive in some cases), so definitely worth talking about changing (the design + implementation of PickLists from Specify 6) in the future!

melton-jason:
> If if you save a mapping, proceed to the data set interface, then later go back to data mapper and change fields and/or the base table, the column titles reflect the initial fields mapped to the columns in the data mapper and data set

From https://github.com/specify/specify7/pull/6344#pullrequestreview-2703519352 by @bronwyncombs 

Yeah, I'm pretty sure the intention of the feature is actually to be helpful and not override any user-specified column names, but it oftentimes falls short in some cases --- especially in a testing environment when the data within Data Sets (and particularly the "type" of data within each column) is variable. 

Imagine you are given or have created a spreadsheet representing some Specify-external data of specimens. You would theoretically be very familiar and comfortable with exactly what each column is is supposed to represent by name alone. 
When you import the spreadsheet using the WorkBench, you have to figure out (or are somehow otherwise given) how the columns from the spreadsheet map to fields in Specify. 
Even after mapping the fields, if you were only given the Specify mapping of a column and not the spreadsheet name for that column, you may need to look at the data, consult some external/given source, and/or perform slightly more mental gymnastics to make the connection that: Specify mapping = specific column in the spreadsheet. 

The core assumption is that the user-provided spreadsheet column names would be more comfortable and familiar than the Specify mapping. 

When you explicitly add a new column to a Data Set and map it to a field in Specify, Specify _always_ knows that specific column is brand-new and so it autogenerates a name for the column based on the mapping. 
When you edit the mapping of an existing column, Specify doesn't know if that column (name) came from something autogenerated, from a spreadsheet, or some other user-provided external source so it errs on the safe side and doesn't overwrite it: assuming the "type" of data within that column hasn't changed, and the user is more comfortable with the existing column name to describe the data. 

You're right in that it's a visual-only discrepancy/effect (the column names), and doesn't actually affect how the data is validated and processed. 

@specify/ux-testing @bronwyncombs 
What would you propose as a solution? 

From the problem, I can think of two potential solutions, both of which could potentially be implemented: 
- Allow the columns to be renamed in the "Data Set" view
  - Probably ideally via an option from right-clicking the column in the Data Set View (i.e., added to the column context menu)? 
  - In a side note, from an accessibility standpoint, is the column context menu (easily and intuitively) accessible using only mouse and keyboard? 
- Keep track of where each column came from. If a column name was autogenerated from the mapper previously, rename it whenever the column is re-mapped in the Field Mapping view.

PR #6348, opened 2025-03-20
Set isPrimary to the last added collector

### LINKED ISSUE #2300: Collector IsPrimary should be checked by default for the first collector
Collector `IsPrimary` should be checked by default for the first collector. Currently the behavior is to not make `IsPrimary` true unless it is explicitly checked.

1. _In contrast, Specify 6 makes `IsPrimary` true for all records. This should be fixed as well, see issue https://github.com/specify/specify6/issues/1153_

This should be established as a business rule that allows checking another collector as primary in addition to the first if desired.

CarolineDenis:
Fixes #2300

### Checklist

### Testing instructions

- Open collecting event form 
- Verify that you have a subview for collector and that the field isPrimary is visible 
- Add a collector 
- Save 
- [ ] Verify the collector isPrimary field is set to true when first added 
- Add a second collector 
- Save 
- [ ] Verify the last collector added is not set to isPrimary true and that the first one added is still true
- Add a third collector 
- Save 
- Delete one collector that is not the primary 
- [ ] Verify the primary is still set to the correct collector 
- Delete the collector that is the primary 
- [ ] Verify that the other one has been set to isPrimary 

Notes: 
No save blockers were added to this new business rule because we still want to allow users to save the CE form without a primary collector if they want to.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
@@ -32,6 +32,7 @@ import type {
   CollectionObject,
   CollectionObjectGroup,
   CollectionObjectGroupJoin,
+  Collector,
   Determination,
   DNASequence,
   LoanPreparation,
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (changes requested):
From https://github.com/specify/specify7/issues/2300

>Collector IsPrimary should be checked by default for the first collector. Currently the behavior is to not make IsPrimary true unless it is explicitly checked.

The _first_ collector needs to be primary. When entering data based on a label, the first agent in the list is often considered the "primary". If this takes the last added collector and makes it primary by default, it would misrepresent the data

grantfitzsimmons (approved):
- [X] Verify the collector isPrimary field is set to true when first added 
- [X] Verify the last collector added is not set to isPrimary true and that the first one added is still true
- [X] Verify the primary is still set to the correct collector 
- Delete the collector that is the primary 
- [X] Verify that the other one has been set to isPrimary 

I did some testing and can confirm that this business rule adjustment makes it so that the first collector is primary, even if the `isPrimary` checkbox is not visible on the form.
If there are two or more collectors and the primary is removed, the next oldest (second entered) will become primary.

acwhite211 (approved):

Areyes42 (approved):
-   [x] Verify the collector isPrimary field is set to true when first added
-   [x] Verify the last collector added is not set to isPrimary true and that the first one added is still true
-   [x] Verify the primary is still set to the correct collector
-   [x] Verify that the other one has been set to isPrimary
***
Looks good!
Screenshot 2025-04-02 at 3 24 27 PM

PR #6358, opened 2025-03-24
Prevent from adding empty value from prep dialog

### LINKED ISSUE #6050: Can add a loan preparation without a quantity
**Describe the bug**
You can empty out the Selected field when selecting preparations for a new loan and save it without a quantity. Furthermore, you can empty the quantity field on a existing loan and still save.

Unsure if the second behavior is intended (because you don't get a save blocker when having the field empty or 0 on an existing loan), but the first one definitely isn't. When creating a new loan, setting the value to 0 means that it shouldn't be selected. This is effectively the same case.

**To Reproduce**

1. Go to Interactions -> Loan
2. Choose either record set or catalog number
3. At the list of preparations, empty one of the Selected fields
4. See that even though it's empty, it's now checked
5. Click apply and see that it goes through
6. Add a loan number
7. Try to save, see that a save blocker pops up for a second
8. Click somewhere else on the form and see that the save blocker is cleared and you can now save

**Expected behavior**
Similar to negative numbers and numbers larger than the amount of preparations, this should not be allowed; preferably clicking Apply should block this.

**Screenshots**

https://github.com/user-attachments/assets/52cbdd1b-febb-4648-8c2f-041a401266cc

- OS: Windows 10
- Browser: Firefox and Chrome
- Specify 7 Version: production (v7.9.11) and v7.9.6.2 
- Collection name: Herp-PC
- User Name: testuser
- URL: N/A

CarolineDenis:
Fixes #6050

### Checklist

### Testing instructions

todo

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Interactions/PrepDialog.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Interactions/PrepDialog.tsx
@@ -186,6 +186,9 @@ export function PrepDialog({
               const loanPreparation = toTable(result, 'LoanPreparation');
               loanPreparation?.set('quantityReturned', 0);
               loanPreparation?.set('quantityResolved', 0);
+              if (Number.isNaN(result.get('quantity'))) {
+                result.set('quantity', 0)
+              }
               return result;
             })
           );


specifyweb/frontend/js_src/lib/components/Interactions/PrepDialog.tsx
```diff
@@ -177,6 +177,9 @@ export function PrepDialog({
           const items = filterArray(
             preparations.map((preparation, index) => {
               if (selected[index] === 0) return undefined;
+              if (Number.isNaN(selected[index])) {
+                return undefined
+              }
```
sharadsw, line 182: can merge both conditions

sharadsw (approved):

acwhite211 (approved):

emenslin (approved):
- [x] Verify that the prep has not been added to the loan
- [x] Verify that only prep with the amount >= 1 has been added to the loan and that the empty has not been added

Looks good, preps without any number are not added. 

https://github.com/user-attachments/assets/e6455295-6572-4ede-bcbe-c1368ca775d5

PR #6363, opened 2025-03-25
Allow to empty count for bulk carry

### LINKED ISSUE #5228: Cannot clear the number field for Bulk Carry Forward, making typing into it unnecessarily complicated
Unsure if this should be a bug or request--it is intended behavior, but the intended behavior is strange and should be changed.

**Describe the bug**
Currently, the behavior of Bulk Carry Forward means that you have to have at least 1 digit in the field. This is understandable behavior, but it makes typing into the field somewhat clunky.

For example--if you want to type in 9, you'd either have to click the arrow several times, or because the default value is 1, type in 9 after it (19) and then delete the 1. This doesn't align much with other count fields in Specify, and using the arrow keys for larger amounts is a bit annoying.

**To Reproduce**

1. Go to a record
2. Make sure Carry Forward and Bulk Carry Forward are checked in the Form Meta menu
3. Attempt to clear the count field next to the Carry Forward button
4. See that you can't clear it, but can type into it regardless

**Expected behavior**
You should be able to clear the count field and type in a new value easily. If the value is 0 or null, then Carry Forward should be disabled. (Grant's evaluation)

**Screenshots**
Hard to show with just a screen recording, but I'm trying to backspace in this video. Backspace also reset the count back to 1 when it was a single digit number besides 1 (did not just decrement it). 

https://github.com/user-attachments/assets/7bf0614d-577f-49e3-8199-a192282f033b

**Crash Report**

- OS: Windows 10
- Browser: Firefox
- Specify 7 Version: `edge`
- Database Name: `sdnhmherps_2_8_24`
- Collection name: `Herp-PC`
- User Name: `testuser`

CarolineDenis:
Fixes #5228

### Checklist

### Testing instructions

- Go to a record
- Make sure Carry Forward and Bulk Carry Forward are checked in the Form Meta menu
- [ ] Verify you can clear the count field next to the Carry Forward button and set it to whichever number you want

Notes: 
If the user enters 0, clears the count box, or leaves it empty, the standard carry-forward behavior will be triggered when the "Carry Forward" button is clicked. This will create a new record based on the carry-forward configuration.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
@@ -256,6 +256,7 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
             <Input.Integer
               aria-label={formsText.bulkCarryForwardCount()}
               className="!w-fit"
+              max={5000}
               min={1}
               placeholder="1"
               value={carryForwardAmount}
--- a/specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
@@ -260,9 +260,7 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
               placeholder="1"
               value={carryForwardAmount}
               onValueChange={(value): void =>
-                Number.isNaN(value)
-                  ? setCarryForwardAmount(1)
-                  : setCarryForwardAmount(Number(value))
+                setCarryForwardAmount(Number(value))
               }
             />
           ) : null}


sharadsw (approved):

acwhite211 (approved):

emenslin (changes requested):
- [x] Verify you can clear the count field next to the Carry Forward button and set it to whichever number you want

Right now you can type in any number, we should set a max to this field like we do for other numeric fields to help prevent any potential problems. 

https://github.com/user-attachments/assets/8b0b2e77-8d2c-4466-a337-866932c08d0c

emenslin (approved):
- [x] Verify you can clear the count field next to the Carry Forward button and set it to whichever number you want

If you try to carry forward over 1000 it takes a long time and sometimes it fails and throws an error. Realistically I don't think users would do this, I just wanted to make note of it.

Areyes42 (approved):
- [x] Verify you can clear the count field next to the Carry Forward button and set it to whichever number you want
***
Looks good, and the 5000 carry forward limit works properly

PR #6369, opened 2025-03-26
Use automatic typeSearch if none defined in xml

### LINKED ISSUE #6278: Automatic `TypeSearches`
Currently, configuring default values for the `TypeSearches` app resource requires manual input, which can be time-consuming and prone to error. This manual setup can lead to inconsistencies and inefficiencies, especially when new tables are added or system tables need to be searched for any reason.

I suggest implementing an automated solution that utilizes a utility like `getMainTableFields` to dynamically retrieve the `searchfield` for requests to the backend. This would streamline the setup process by automatically populating default values based on the existing table configurations.

From @melton-jason:
> We could use a utility like `getMainTableFields` to get the `searchfield` for the request to the backend 
> https://github.com/specify/specify7/blob/8ee5be2f3c16efcd17ba6eea1ff206d8fb403ede/specifyweb/frontend/js_src/lib/components/Formatters/formatters.ts#L275-L301

As a much less elegant and short term solution, we could at least update the `typesearch_def.xml` file with a type search for _all_ tables that users may need to search.

In addition, we can consider implementing a feature to infer a `TypeSearch` configuration for a table when one does not already exist. This approach ensures that we provide a basic level of functionality, which is certainly better than having no configuration at all.

CarolineDenis:
Fixes #6278

### Checklist

### Testing instructions

- Open a form with a QCBX that does not have a defined typeSearch in xml 
- type inside the box and verify that you can search using the main field 

The code will check as follow: 
- Check the user defined custom type search
- Check type search in the default xml typeSearch
- If none of above use getMainTableFields and use the first field for the search

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryComboBox/useTypeSearch.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryComboBox/useTypeSearch.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 
 import { useAsyncState } from '../../hooks/useAsyncState';
 import { getAppResourceUrl } from '../../utils/ajax/helpers';
-import type { RA } from '../../utils/types';
 import { filterArray, localized } from '../../utils/types';
 import type { LiteralField, Relationship } from '../DataModel/specifyField';
 import type { SpecifyTable } from '../DataModel/specifyTable';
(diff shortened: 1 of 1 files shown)


pashiav (approved):
- [X] - type inside the box and verify that you can search using the main field 


---
Looks good! Can search using the main field.

sharadsw (approved):

PR #6380, opened 2025-04-01
Inclusion of null taxontreedef in query results

### LINKED ISSUE #5479: Adding tree ranks to queries will incorrectly filter valid records
Related to code I pushed a while ago. Wanted to make this issue before but just got time. Introduced in https://github.com/specify/specify7/pull/5036.

There's a bug below
https://github.com/specify/specify7/blob/b69dac305cd93466a93b2dea773174df5f14524f/specifyweb/stored_queries/query_construct.py#L93
It'll _always_ apply the `in` predicate. That means if you make a query like Determination -> Taxon (species) -> Fullname, with "any" filter, determinations without a taxon will be absent. This is not correct, and a regression

I haven't tested this anywhere yet (since sp7demofish still seems to run 7.9.6.2). If someone on testing could verify this, it'll be good. 

Instructions
1. Make a collectionobject. Assign taxon (species) to it. You could technically assign anything, but for simplicity just assign species OR remember what rank was assigned.
2. Make another collectionobject, Do not assign taxon to it.

Run the following query. Basically Determination -> Taxon (species) -> Fullname
image

You'll notice that only collectionobject from first step is returned, and from second step is not returned (even though it should, just blank for species field).

This bug is on production, and not on 7.9.6.2, so couldn't test it myself.

The correct fix would be to change the predicate to be
https://github.com/specify/specify7/blob/b69dac305cd93466a93b2dea773174df5f14524f/specifyweb/stored_queries/query_construct.py#L93

```
import sqlalchemy
sqlalchemy.or_(getattr(node, treedef_column).in_(defs_to_filter_on), getattr(node, treedef_column) == None)
```

Technically you could push None outside, but query optimizers should optimize that out anyways

Not related to https://github.com/specify/specify7/issues/5402. 🤯

acwhite211:
Fixes #5479

Include Collection Objects with null taxon determinations to be returned when searching for `Any` taxon.

Changes the `WHERE` clause from:
```
WHERE
	collectionobject.`CollectionID` = 4
	AND taxon_1.`TaxonTreeDefID` IN (1)
```
to
```
WHERE
	collectionobject.`CollectionID` = 4
	AND (taxon_1.`TaxonTreeDefID` IN (1)
		OR taxon_1.`TaxonTreeDefID` IS NULL)
```

### Checklist

### Testing instructions

- Create a new collection object with a species taxon determination.  Note the catalog number for later.
- Create a new collection object without any taxon determination.  Note the catalog number for later.
- Create a collection object query with a filter for the two catalog numbers and 'Determinations -> Taxon -> Species -> Full Name -> Any'.  See the screen shot for the full example query.
- [x] See that in the query results, both of the newly created collection objects appear.

Screenshot of first CO with Taxon:
image

Screenshot of second CO without Taxon:
image

Screenshot of query with results:
image

### DIFF
--- a/specifyweb/stored_queries/query_construct.py
+++ b/specifyweb/stored_queries/query_construct.py
@@ -2,7 +2,7 @@ import logging
 from collections import namedtuple, deque
 from typing import Tuple, List
 
-from sqlalchemy import orm, sql
+from sqlalchemy import orm, sql, or_
 
 import specifyweb.specify.models as spmodels
 from specifyweb.specify.tree_utils import get_treedefs
@@ -109,7 +109,9 @@ class QueryConstruct(namedtuple('QueryConstruct', 'collection objectformatter qu
 
         defs_to_filter_on = [def_id for (def_id, _) in treedefs_with_ranks]
         # We don't want to include treedef if the rank is not present.
-        new_filters = [*query.internal_filters, getattr(node, treedef_column).in_(defs_to_filter_on)]
+        new_filters = [
+            *query.internal_filters,
+            or_(getattr(node, treedef_column).in_(defs_to_filter_on), getattr(node, treedef_column) == None)]
         query = query._replace(internal_filters=new_filters)
         
         return query, column, current_field_spec.get_field(), table


CarolineDenis (approved):

sharadsw (approved):

pashiav (approved):
- [x] See that in the query results, both of the newly created collection objects appear.
---

Looks good! 
Screenshot 2025-04-02 at 12 05 19 PM

emenslin (approved):
- [x] See that in the query results, both of the newly created collection objects appear.

Looks good!

image

PR #6405, opened 2025-04-09
Prevent adding a component to itself

### LINKED ISSUE #6392: Components: Prevent Component from being its own parent
Part of #6388 

In the `component` table, there is a field that allows parent-child relationships. We need business logic to prevent a component from being its own parent. This should not be possible in most cases since each component will have the hierarchy managed as dependent children, but it needs to be prevented.

CarolineDenis:
Fixes #6392 

> [!WARNING]  

### Checklist

- [ ] Add automated tests


CarolineDenis:
NOTES: 
Should component exist by themselves or only in a CO context? 
If yes, should we leave the magnifying glass next to the children subview in the component?

Issue #6394, opened 2025-04-07
CO Children: Create default forms


grantfitzsimmons:
We need to create the default forms and app resources for the collection object parent-child module.

CarolineDenis:
NOTES: 
Change for child - parent CO?

PR #6408, opened 2025-04-11
Create a parent child relationship on CO table

### LINKED ISSUE #6407: Parent CO: Add a new relationship from CO to CO
Add Parent-Children Relationship to Collection Object (CO) Table

**Description:**

We need to introduce a **parent-children relationship within the CO table**, allowing a Collection Object (CO) to directly have one or multiple children COs. Each child CO can itself have its own children, enabling a nested structure. This is similar to the current COG (Collection Object Group) concept, but with a simpler and more intuitive structure — *like COGs, without needing to create a COG*.

### Background & Motivation:

This change aims to streamline workflows, especially in disciplines like geology. Currently, expressing relationships between related physical objects (e.g., a large rock and its mineral components) requires the creation of abstract COGs that are either:
- Populated with all COs as siblings, or
- Used to group child objects under a parent, then grouped again for related sub-elements.

This process is often cumbersome and introduces unnecessary abstraction.

### Example Use Case:
In geology:
- **Parent CO**: Rock1 (Taxon: Rock1)
- **Child COs**: Mineral1 and Mineral2 (Taxa: Mineral1, Mineral2)

Previously:
- A COG would be created, and either all 3 items would be added as siblings, or a hierarchy would be created through COG + nested COGs.

With this update:
- **Rock1** is created as a parent CO.
- **Mineral1** and **Mineral2** are created directly as its children, each with their own catalog number (or none), taxon, properties, and types.
- No need to create a separate COG unless desired.

### Implementation Notes:
- The CO table will require a `parent_id` field to link a CO to its parent CO (nullable for top-level COs).
- The relationship is **recursive**, allowing COs to form trees.
- This feature does **not** replace COGs — both systems can coexist and be used in the same collection.
- Parent-child links are intended to represent real-world containment or belonging (e.g., "this object is part of that one") in a faster, more intuitive way than COGs.

### Benefits:
- Simplifies UI/UX for linking related objects.
- Reduces complexity when expressing object relationships.
- Keeps geological and similar data structured, clear, and closer to physical reality.

CarolineDenis:
Fixes #6407

> [!WARNING]  

### Checklist

- [ ] Add automated tests

### Testing instructions

- Go to app resource > collection object 
- Add this snippet to your CO form 
```
<row>
	<cell type="subview" viewname="CollectionObjectChild" id="6" name="children" colspan="14"/>
</row>
```
- Create a new CO view def called CollectionObjectChild
```
<viewdef name="CollectionObjectSub (2)" class="edu.ku.brc.specify.datamodel.CollectionObject" type="form" gettable="edu.ku.brc.af.ui.forms.DataGetterForObj" settable="edu.ku.brc.af.ui.forms.DataSetterForObj">
	<desc>The Collection Object Subform for appraisal and project forms.</desc>
	<enableRules/>
	<columnDef>100px,2px,120px,5px,100px,2px,98px,5px,95px,2px,98px,5px,95px,2px,98px,p:g</columnDef>
	<columnDef os="lnx">115px,2px,152px,5px,107px,2px,100px,5px,110px,2px,100px,5px,110px,2px,100px,p:g</columnDef>
	<columnDef os="mac">130px,2px,150px,5px,127px,2px,121px,5px,120px,2px,121px,5px,119px,2px,121px,p:g</columnDef>
	<columnDef os="exp">p,2px,min(p;150px),5px:g,p,2px,p:g(2),5px:g,p,2px,p:g(2),5px:g,p,2px,p:g(2)</columnDef>
	<rowDef auto="true" cell="p" sep="2px"/>
	<rows>
		<row>
			<cell type="field" id="1" name="catalogNumber" uitype="formattedtext"/>
			<cell type="field" id="2" name="collectionobjecttype" uitype="combobox"/>
		</row>
		<row>
			<cell type="subview" id="5" name="children" colspan="7" viewname="CollectionObjectChild"/>
		</row>
	</rows>
</viewdef> 
```
- Go to data entry collection object 
- Add a new child 
- Save 
- [ ] Verify you can add a child to the parent CO 
- Add a second child (sibling to the first you added) 
- Save 
- [ ] Verify you can add a sibling to a child CO, that you can add several children to a CO 
- Add a child to the first child 
- Save 
- [ ] Verify you can add a child to a child 
- [ ] Verify you can delete any children on a CO parent 

>>> See in comments bellow for pictures


specifyweb/frontend/js_src/lib/components/Forms/SubView.tsx
```diff
@@ -160,11 +160,7 @@ export function SubView({
 
   return (
     <SubViewContext.Provider value={contextValue}>
-      {(!RECURSIVE_RENDERING_EXCEPTIONS.has(parentResource.specifyTable) &&
```
CarolineDenis, line 163: to add back and improve to display children on CO

CarolineDenis:
NOTES: 
Add cascade deletion, if delete a child delete all its children

specifyweb/frontend/js_src/lib/components/Forms/SubView.tsx
sharadsw: Can you add this code back and add CO to the exception list? Without this we will have infinite re-rendering on certain subviews again (see second video in: https://github.com/specify/specify7/pull/5253#pullrequestreview-2269558472)

specifyweb/specify/migrations/0027_CO_children.py
```diff
@@ -0,0 +1,96 @@
...
+                    localized_items_desc.text = new_desc
+                    localized_items_desc.save() 
+
+                    localized_items_name.text = new_name
+                    localized_items_name.save() 
+
+def revert_update_fields(apps):
+        for table, fields in SCHEMA_CONFIG_TABLE_FIELDS.items(): 
+            for field in fields: 
+                revert_table_field_schema_config(table, field, apps)
```
sharadsw, line 53: Indentation error?

CarolineDenis:
Screenshot 2025-04-18 at 1 27 20 PM
Screenshot 2025-04-18 at 1 28 14 PM

combs-a (changes requested):
- [X] Verify you can add a child to the parent CO 
- [X] Verify you can add a sibling to a child CO
- [X] Verify that you can add several children to a CO 
- [X] Verify you can add a child to a child 
- [X] Verify you can delete any children on a CO parent 
<hr>

All of the tests were passed but I have some concerns regarding this feature.

Deleting a child CO from this form means that you permanently delete it, instead of removing the relationship connection between parent and child. Are children guaranteed to be dependent on the parent CO?

This also causes a cascade where children of children get deleted, and there's no warning like how we would typically handle deleting a CO from the button on the form (this record is referenced by... type of dialog), nor warning of permanent deletion.

I feel like this lack of warning is a bit dangerous, as there's no way to access the child CO from the parent form easily so that you can check that you're deleting the correct child. If they're not meant to be treated as COs and are entirely dependent on the parent CO, and therefore deleteable, I feel like having them display in queries might be a bit misleading? There's some issues with querying right now, and I'm unsure if you're able to exclude just children right now.

Additionally, while not apart of the testing of this PR, I tried some simple queries, and none of the COs I had that had children appeared when I attempted to display them with aggregated children.

image

But they do appear with Children -> Catalog Number -> Any.

image

Other people should give their opinions though! I might just be getting caught up on things that hopefully won't matter.

CarolineDenis:
@combs-a 
After discussing with the other developers, here are responses to the concerns raised:

1) **Child-Parent Dependency in Collection Objects (COs):**
Deleting a child CO will remove it from the database without a warning, similar to how preps are deleted.

2) **Cascade Deletion of Descendants:**
If a child CO has its own children (or even grandchildren), all descendants will be deleted in one action. No warning will be displayed—this mirrors the behavior when deleting a rank with nodes that have no CO references.

3) **Aggregator Handling:**
Aggregator support will be addressed as part of https://github.com/specify/specify7/issues/6394.

The current implementation meets the requirements for version 1.

combs-a (approved):
- [X] Verify you can add a child to the parent CO 
- [X] Verify you can add a sibling to a child CO, that you can add several children to a CO 
- [X] Verify you can add a child to a child 
- [X] Verify you can delete any children on a CO parent 
<hr>

All the tests passed, looking good! 😄 

Had a brief bug that I couldn't recreate but noting it here; for some reason I could add a child that did not have a unique Collection Object number. Tried to recreate it multiple times but didn't run into it again.

emenslin (approved):
- [x] Verify you can add a child to the parent CO 
- [x] Verify you can add a sibling to a child CO, that you can add several children to a CO 
- [x] Verify you can add a child to a child 
- [x] Verify you can delete any children on a CO parent 

Works as expected

PR #6429, opened 2025-04-18
Add a new treeDefault import feature

### LINKED ISSUE #6294: [Guided Setup] - Add a tool to import a default tree
**Proposed Enhancements:**

**1. Fetching Tree List from AWS Bucket**

- The default trees will be stored in an AWS bucket, and the backend will pull the data from there.

- A list of default trees, including their descriptions and sources, will be dynamically generated on the frontend by fetching a remote JSON file from https://files.specifysoftware.org/taxonfiles/taxonfiles.xml.

- The existing XML file needs to be converted to JSON format for compatibility.

**2. Displaying Tree List in UI**

- When the user clicks the "+" button in the tree viewer, a dialog will appear displaying the dynamically fetched list of trees.

- Upon clicking a tree, the tree data will be uploaded to the database.

**3. Rank Management Enhancements**

- Users can continue to edit and remove ranks using the existing rank editor.

- Users can remove sections of the tree by deleting the oldest parent node, which will remove all its child nodes as well.

Dependent issue to manage the ranks: 
https://github.com/specify/specify7/issues/6295

CarolineDenis:
Fixes #6294

### Checklist

- [ ] Add automated tests

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/TreeView/CreateTree.tsx
+++ b/specifyweb/frontend/js_src/lib/components/TreeView/CreateTree.tsx
@@ -2,10 +2,10 @@ import React from 'react';
 
 import { commonText } from '../../localization/common';
 import { treeText } from '../../localization/tree';
-import type { DeepPartial } from '../../utils/types';
+import type { DeepPartial, RA } from '../../utils/types';
 import { localized } from '../../utils/types';
 import { getUniqueName } from '../../utils/uniquifyName';
-import { Ul } from '../Atoms';
+import { H2, H3, Ul } from '../Atoms';
 import { Button } from '../Atoms/Button';
 import { className } from '../Atoms/className';
 import type {
@@ -21,6 +21,17 @@ import type { TreeInformation } from '../InitialContext/treeRanks';
 import { userInformation } from '../InitialContext/userInformation';
 import { Dialog } from '../Molecules/Dialog';
 import { defaultTreeDefs } from './defaults';
+
+type TaxonFileDefaultDefinition = {
+  readonly discipline: string;
+  readonly title: string;
+  readonly coverage: string;
+  readonly file: string;
+  readonly src: string;
+  readonly size: number;
+  readonly description: string;
+};
+type TaxonFileDefaultList = RA<TaxonFileDefaultDefinition>;
 
 export function CreateTree<
   SCHEMA extends AnyTree,
@@ -40,7 +51,27 @@ export function CreateTree<
     SpecifyResource<AnySchema> | undefined
   >(undefined);
 
-  const handleClick = (
+  const [treeOptions, setTreeOptions] = React.useState<
+    TaxonFileDefaultList | undefined
+  >(undefined);
+
+  React.useEffect(() => {
+    fetch('https://files.specifysoftware.org/taxonfiles/taxonfiles.json')
+      .then(async (response) => response.json())
+      .then((data: TaxonFileDefaultList) => {
+        setTreeOptions(data);
+      })
+      .catch((error) => {
+        console.error('Failed to fetch tree options:', error);
+      });
+  }, []);
+
+  const handleClick = (resourceFile: string): void => {
+    console.log(resourceFile);
+    // Pass resource file name to backend
+  };
+
+  const handleClickEmptyTree = (
     resource: DeepPartial<SerializedResource<TaxonTreeDef>>
   ) => {
     const uniqueName = getUniqueName(
@@ -78,9 +109,30 @@ export function CreateTree<
           onClose={() => setIsActive(0)}
         >
           <Ul className="flex flex-col gap-2">
+            <H2>{treeText.populatedTrees()}</H2>
+            {treeOptions === undefined
+              ? undefined
+              : treeOptions.map((resource, index) => (
+                  <li key={index}>
+                    <Button.LikeLink
+                      onClick={(): void => handleClick(resource.file)}
+                    >
+                      {resource.title}
+                    </Button.LikeLink>
+                    <div className="text-xs text-gray-500">
+                      {resource.description}
+                    </div>
+                    <div className="text-xs text-gray-400 italic">
+                      {`Source: ${resource.src}`}
+                    </div>
+                  </li>
+                ))}
+            <H2>{treeText.emptyTrees()}</H2>
             {defaultTreeDefs.map((resource, index) => (
               <li key={index}>
-                <Button.LikeLink onClick={(): void => handleClick(resource)}>
+                <Button.LikeLink
+                  onClick={(): void => handleClickEmptyTree(resource)}
+                >
                   {localized(resource.name)}
                 </Button.LikeLink>
               </li>
--- a/specifyweb/frontend/js_src/lib/localization/tree.ts
+++ b/specifyweb/frontend/js_src/lib/localization/tree.ts
@@ -594,4 +594,10 @@ export const treeText = createDictionary({
   meteorites: {
     'en-us': 'Meteorites',
   },
+  populatedTrees: {
+    'en-us': 'Populated trees',
+  },
+  emptyTrees: {
+    'en-us': 'Empty Trees',
+  },
 } as const);


specifyweb/specify/tree_views.py
```diff
@@ -536,6 +538,83 @@ def has_tree_read_permission(tree: TREE_TABLE) -> bool:
...
+        'ictheology': 'https://specify-software-public.s3.us-east-1.amazonaws.com/default_trees/col2008_fishes.csv',
+        'herpetology': 'https://specify-software-public.s3.us-east-1.amazonaws.com/default_trees/col2008_herps.csv',
+        'ornothology': 'https://specify-software-public.s3.us-east-1.amazonaws.com/default_trees/col2008_aves.csv',
+        'mammology': 'https://specify-software-public.s3.us-east-1.amazonaws.com/default_trees/col2008_mammalia.csv',
+        'entomology': 'https://specify-software-public.s3.us-east-1.amazonaws.com/default_trees/col2008_orthoptera.csv',
+        'botany': 'https://specify-software-public.s3.us-east-1.amazonaws.com/default_trees/col2008_poales.csv',
+        'invertebrate': 'https://specify-software-public.s3.us-east-1.amazonaws.com/default_trees/col2008_inverts.csv'
+    }
+    title_to_discipline = {
+        'fishes': 'ictheology',
```
CarolineDenis, line 558: Ichthyology

CarolineDenis:
NOTES: 
- There are no default for geology trees

CarolineDenis:
TODO: 
- When testing the feature, it fails when running `def create_default_trees_view(request):` because: 
`logged_in_collection_name = request.user.logincollectionname `
returns None for me

alesan99:
@emenslin All ranks should now be correctly downloaded 👍
And all ranks should have isInFullName, isEnforced, and fullNameSeparator configured now too.

Updated testing instructions with the new chrono strat tree too

emenslin (changes requested):
- [x] See that a list of default Taxon trees is displayed
- [x] Click on a default tree and see you get a notification that it is being created.
- [x] Wait and make sure the default tree is successfully created
- [x] Create a new empty tree and make sure it is created successfully.
- [x] Make sure you can start creating a tree and cancel it with the Cancel button in the popup.
- [x] Make sure you can cancel the tree from the "in progress" notification as well.
  - [x] The new secondary Geography tree should be created after a few minutes.
  - [x] The new Chronostrat tree should be created after a few minutes.

---

- I tried uploading all taxon trees, everything uploads successfully except the ornithology tree. 
- The root node for geography is just called 'Root' but should be called 'Earth'
- The root node for Chronostrat is called 'Root' but should be called 'Time' 
- The chronostratigraphy tree doesn't have an Eon rank. I'm not sure if this is necessary but I wanted to mention it in case it is. 
- The `full name separator` for the root ranks for both chronostrat and geology are set to ` ` but should be set to `, `.

alesan99:
> * I tried uploading all taxon trees, everything uploads successfully except the ornithology tree.

Hm not sure why its failing, I increased the re-connection delay, so it might work now?

Iwantexpresso (approved):
- [x] See that a list of default Taxon trees is displayed
- [x] Click on a default tree and see you get a notification that it is being created.
- [x] Wait and make sure the default tree is successfully created
- [x] Create a new empty tree and make sure it is created successfully.
- [x] Make sure you can start creating a tree and cancel it with the Cancel button in the popup.
- [x] Make sure you can cancel the tree from the "in progress" notification as well.
- [x] The a new secondary Geography tree should be created after a few minutes.
- [x] The new Chronostrat tree should be created after a few minutes.

just tested this on Calvert and it looks great! nice job y'all !

grantfitzsimmons (changes requested):
> - [X] See that a list of default Taxon trees is displayed

There is almost no space between the "Populated Trees" and "Empty Trees" sections, making it difficult to differentiate between the list of trees that are pre-loaded with data and the list of empty trees that can be chosen.

Image

More importantly, the source data we are providing says it is from **2008**, which is _incredibly_ out of date (for most disciplines). The defaults we provide in Specify 6 today are from Catalog of Life 2021 in most cases, see https://files.specifysoftware.org/taxonfiles/

https://files.specifysoftware.org/taxonfiles/taxonfiles_2.xml

Image

Collections curators and managers want to use the most up-to-date taxonomic data possible when using default data, so we need to make sure these updated spreadsheets are being used.

> - [X] Click on a default tree and see you get a notification that it is being created.

Image

Some suggestions:
* The dialog header "The Default Tree creation process has started." should not have a period in it (to be stylistically consistent with other dialogs in the app) and should either be entirely proper case or 'default tree' should be lower case (as it is not a proper noun).
    * Better yet, it should tell the user _what_ tree it is making, as while the 20,000+ records are added, there is no indication in this dialog confirming they made the correct selection. The notifications do include the name of the tree, but this is not visible to the user while this process happens
* There is no "Close" button so the user can resume their work, only a red **Cancel** button. Since users instinctively click the primary action button, they may accidentally cancel this and not understand it will happen in the background. We could add a **Close** button after the cancel button to prevent this from happening instinctively.`[Cancel] [Primary Action]`. 
    * In fairness, we might _not even need to show the user progress at all_, as long as the worker is functional, everything should _just work_.

> - [X] Wait and make sure the default tree is successfully created

This is quite slow... I chose the 'Ichthyology' tree and began populating it at 1:18 (according to the network requests). It crawled along, finishing populating entirely at 1:47 (29 minutes total) on the test panel.

Thinking of the (expectedly common) case of someone creating a database with several disciplines with a large number of trees, they would be forced to queue and wait for these jobs to finish over a long period of time.

It would make a lot of sense to batch these and make this action more efficient to prevent frustration.

The notification showing "Cancel" persists regardless of the state (success|failure).

Image

> - [X] Create a new empty tree and make sure it is created successfully.

That was fast 😄 

> - [X] Make sure you can start creating a tree and cancel it with the Cancel button in the popup.
> - [X] Make sure you can cancel the tree from the "in progress" notification as well.

This technically works, but–

Cancelling the tree still sends 3 notifications:
1. The Default Tree creation process has started.
2. Default Tree creation in progress.
3. Default Tree creation failure. 
     Aborted

**My suggestion:**
* Remove the 'Task ID' collapsible section from all notifications– the users should not see this.
* Remove the giant red "Cancel" button from the notification. If the user wants to remove this tree, they should **simply be able to do so** after it is created (https://github.com/specify/specify7/issues/7594)
* Send only two notifications, one saying the process has started and one saying the tree creation was cancelled. Saying "Default Tree creation failure. Aborted" is overwhelmingly negative and says:
    * The software failed unintentionally ("creation failure")
    * We use programmer lingo ("aborted" instead of "cancelled")

    Image

> - Create a Geography tree:
>   - Click execute and make note of the response you got below.

Image

>   - [X] The new secondary Geography tree should be created after a few minutes.
>   - You can track the progress by going to `yourdb.test.specifysystems.org/api/create_default_tree/status/YOUR_TASK_ID/`. You should see your task id in the response you got.

> - Create a Chronostratigraphy tree:
>   - Click execute and make note of the response you got below.

Image

>   - [X] The new Chronostrat tree should be created after a few minutes.

---

### Other Thoughts

The core issue, as I see it, is that the user needs to create a taxon tree populated with our default data.

From my conversation with @alesan99, I understand that a big challenge we face is the time it takes to build these trees. This PR so far is attempting to develop a robust system that will inform users about the current status of the background process, including notifications, status updates, a progress bar, and options to cancel the action, to address this slowdown.

My (perhaps naïve) perspective is that we should re-evaluate how we provide the default trees to make this process nearly instantaneous and eliminate previous complications. If we can download the full tree CSV file and import it all at once, the extra work may be unnecessary. The faster this process completes, the better the user experience. Fewer messages for the user also would improve the experience.

In Specify 6, this took only a few seconds (the file downloaded locally, the tree was populated, the user moved on). Maybe there are lessons there we can use to make this more elegant.

This PR is loading the data row-by-row while Specify 6 used batch database operations, resulting in many more database round-trips.

* We could use [bulk_create()](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#bulk-create) instead of individual saves
* Seems pertinent to pre-fetch all existing nodes and ranks
* Batch inserts in groups?
* Maybe we need to consider raw SQL for the import portion if that's feasible...

grantfitzsimmons (commented):
One usability concern I have is that if you dismiss the creation dialog, you are still shown the 'Add Tree' dialog, which may lead users to click the tree _again_ instinctively, creating two trees: 

https://github.com/user-attachments/assets/73ccd804-2cd3-4e07-ab9d-4afd92f4a877

**Solution:** Hide the 'Add Tree' dialog once creation begins

Since the tree creation is not fast yet (see https://github.com/specify/specify7/issues/7641), the user is left wondering if the tree they started creating is being made again. I was not sent any notification for either of the trees I made. In the logs for the `worker` or `specify7` container, I don't see an indication either of whether or not the tree is being created.

Until it becomes fast enough that the time between the tree creation being initiated and the time the tree exists, we need some mechanism to show the user this is happening. 

I imagine this is just a bug? It looks like the intention was for this notification to be sent @alesan99 

>"Everything should be as simple as it can be, but not simpler"

alesan99:
> **Solution:** Hide the 'Add Tree' dialog once creation begins

Done 👍
>
> Until it becomes fast enough that the time between the tree creation being initiated and the time the tree exists, we need some mechanism to show the user this is happening.
> 
> I imagine this is just a bug? It looks like the intention was for this notification to be sent @alesan99

It is indeed a bug, a notification should be sent out as soon as the worker receives the task.
Since nothing is being shown in the logs either I assume there was an issue with the worker receiving the task.

I could move the notification trigger to the main process, rather than the worker, but that risks lying in cases where the task didn't actually start (which looks like what's happening in this case maybe?). Perhaps I can at least add a failure notification if there is some way to detect if the worker didn't start the task.

PR #6434, opened 2025-04-22
Update start and end period in GeologicTimePeriod when reversed

### LINKED ISSUE #5412: Default `GeologicTimePeriod` records are not linear between `startPeriod` and `endPeriod`
**Describe the bug**
The IUGS ICC chart is linear in moving towards the modern day, but our Chronostrat data switches if the larger number is the start period or the end period.

In short– the `startPeriod` value should be **larger** than the `endPeriod` value (million year increments), but that is not always the case currently.

**To Reproduce**

1. Create a database/discipline with a default Chronostratigraphy tree (takes place automatically if you create a discipline that is either `invertpaleo` or `vertpaleo`)
2. Query Chronostrat Name, Start Period, and End Period
3. See that `startPeriod` is not always larger than `endPeriod`, but sometimes is!

**Expected behavior**
- Our new default tree needs to be consistent in the order between `startPeriod` and `endPeriod`.
- Users with existing trees must not experience a crash when using new tools dependent on these ranges like the EAQ (https://github.com/specify/specify7/issues/5410)
- We should consider migrating these values to be in the appropriate order if the user has not modified the defaults in any way.

**Screenshots**
image

[Chronostrat Comparison.xlsx](https://github.com/user-attachments/files/17904320/Chronostrat.Comparison.xlsx)

CarolineDenis:
Fixes #5412

> [!WARNING]  

### Checklist

### Testing instructions

- Open a db with production that has geologic time period tree 
- Open a node 
- Find a node that has a start period greater than the start or create one or update one => start 100 - end 200
- Save 
- Change the branch to use this pr and not prod 
- Open the node from the previous step 
- Verify the start period is now greater than the end. start 200 - end 100

### DIFF
--- a/specifyweb/patches/migrations/0004_chrono_start_end.py
+++ b/specifyweb/patches/migrations/0004_chrono_start_end.py
@@ -1,0 +1,37 @@
+from django.db import migrations, connection
+
+def reverse_faulty_end_start_period(apps, schema_editor):
+    from django.db import connection
+
+    with connection.cursor() as cursor:
+        # Check if the table geologictimeperiod exists
+        cursor.execute("""
+            SELECT COUNT(*)
+            FROM information_schema.tables
+            WHERE table_name = 'geologictimeperiod' AND table_schema = DATABASE();
+        """)
+        if cursor.fetchone()[0] == 1:
+            # If the table exists, perform the update
+            cursor.execute("""
+                UPDATE geologictimeperiod AS gtp
+                JOIN (
+                    SELECT GeologicTimePeriodID, StartPeriod AS orig_start
+                    FROM geologictimeperiod
+                    WHERE StartPeriod < EndPeriod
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (commented):
To clarify, this does not follow this request in the original issue:
> if the user has not modified the defaults in any way.

That means all chronostratigraphic records will now switch the values, so the `startPeriod` value should be larger than the `endPeriod` value. This does not consider if the defaults have or have not been modified.

Do we understand why the chronostrat tree was initially set up as it was? Are we comfortable updating user-created data without notifying anyone?

alesan99 (approved):
- [x] Verify the start period is now greater than the end. start 200 - end 100 
---
I agree with Grant since it did change some of the chronostrat records I added myself, but functionality wise it worked without issue.

Left: New
Right: Old

chrome_B943qmM3I3

Values have been updated correctly 👍

grantfitzsimmons (approved):
**TODO:** Add to release notes 👍

PR #6438, opened 2025-04-23
Prevent from decreasing prep count to lower num than loaned

### LINKED ISSUE #6385: Can edit preparation count when all preparations loaned
**Describe the Bug**
If all counts on a preparation are loaned, you can go to the CO and edit the count to be less than the amount that are on loan. This means the count on a CO form could be 1 but that same prep could be a part of multiple loans. 

**To Reproduce**

1. Create a new CO or edit existing one 
2. Add a preparation with a count of 2 
3. Create a loan and add 1 prep
4. Create a second loan and add the other prep 
5. Go back to the CO and change the count to 1 
6. See the count changed but both loans still exists

**Expected behavior**
You should not be able to edit the count to be lower than the preparations that are on loan

**Screenshots**

https://github.com/user-attachments/assets/07af45d3-891c-4510-bfd7-0fa6f168700d

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: 7.10.2 and 7.9.6.2
- Database Name: kufish20250214
- Collection name: fish voucher
- User Name: spfishadmin
- URL: https://kufish20250214-production.test.specifysystems.org/specify/

**Additional context**

> I don’t think that I do anything differently from what you describe, except that I would have multiple loans connected to a single prep type from a single prep record.
> 
> The example that has been problematic through multiple iterations of Specify 6 and 7 is a fossil cat skull.
> 
> We have one original, and multiple casts.
> 
> Original is prep# 0, the count is 1.
> 
> Resin cast replicas are prep# 1, the count varies over time as we loan, gift, use and break them and make new ones.
> 
> After recording a loan history, eventually the number of preps (prep# 1) available for loan and on loan would not match the number in the prep count field. (loan preps + available preps ≠ prep count).
> 
> I’ve gone as far as deleting all of the loan records to try and reconcile the issue. My recollection was that I ended up with negative preps. My recollection is that I deleted the specimen record as well and started from scratch, but in re-entering the loan records, etc., the prep count was again not equal to the number of preps on loan plus those available.
> 
> I’ve run into this problem when there are multiple loans of multiple preps tied to one object record.
> 
> It is not something that I do a lot of, but I have seen this on multiple occasions now, so I don’t think that it’s a problematic record, I think it has to be the loan preps not correctly totaling when there are multiple loans.
> 

**Reported By**
James C. Sagebiel @ The University of Texas on [discourse](https://discourse.specifysoftware.org/t/prep-count-vs-loan-prep-count-problems/2427)

CarolineDenis:
Fixes #6385

### Checklist

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
@@ -659,6 +659,35 @@ export const businessRuleDefs: MappedBusinessRuleDefs = {
     },
   },
   Preparation: {
+    fieldChecks: {
+      countAmt: async (prep): Promise<BusinessRuleResult | undefined> => {
+        const loanPrep = await prep.rgetCollection('loanPreparations');
+        const totalPrep = prep.get('countAmt') ?? 0;
+        let totalPrepLoaned = 0;
+
+        loanPrep.models.forEach((loan) => {
+          const quantity = loan.get('quantity') ?? 0;
+          totalPrepLoaned += quantity;
+        });
+
+        if (totalPrep < totalPrepLoaned) {
+          setSaveBlockers(
+            prep,
+            prep.specifyTable.field.countAmt,
+            [resourcesText.preparationUsedInLoan()],
+            PREPARATION_LOANED_KEY
+          );
+        } else {
+          setSaveBlockers(
+            prep,
+            prep.specifyTable.field.countAmt,
+            [],
+            PREPARATION_LOANED_KEY
+          );
+        }
+        return undefined;
+      },
+    },
     onRemoved: (preparation, collection): void => {
       if (preparation.get('isOnLoan') === true) {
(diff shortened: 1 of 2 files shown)


CarolineDenis:
Notes: 
This new rule triggers an issue in File "/opt/specify7/specifyweb/interactions/cog_preps.py", line 538, in `modify_update_of_loan_return_sibling_preps`
`in updated_interaction_data["loanpreparations"][loan_prep_idx].keys()`

emenslin (approved):
- [x] Verify save blocker 
- [x] Verify you can save 

Looks good
Screenshot 2025-04-25 101341

Areyes42 (approved):
-  [x] Verify save blocker
- [x]   Verify you can save
***
Looks good, working as expected!
Screenshot 2025-04-02 at 1 27 13 PM

sharadsw (approved):

acwhite211 (approved):

PR #6449, opened 2025-04-28
Create viewdef for COChild

### LINKED ISSUE #6394: CO Children: Create default forms

CarolineDenis:
Fixes #6394

### Checklist


config/geology/geology.views.xml
```diff
@@ -140,6 +147,9 @@
 					<cell type="label" labelfor="5"/>
 					<cell type="field" id="5" uitype="plugin" name="this" initialize="name=PartialDateUI;df=catalogedDate;tp=catalogedDatePrecision" default="today" uifieldformatter="Date"/>
 				</row>
+				<row>
+					<cell type="subview" viewname="CollectionObjectChild" id="6" name="children" colspan="14" defaulttype="table"/>
```
CarolineDenis, line 151: We should probably choose between COChild and cojo here and comment our the other one

config/geology/geology.views.xml
```diff
@@ -1427,5 +1437,28 @@
...
+			<columnDef os="exp">p,2px,min(p;150px),5px:g,p,2px,p:g(2),5px:g,p,2px,p:g(2),5px:g,p,2px,p:g(2)</columnDef>
+			<rowDef auto="true" cell="p" sep="2px"/>
+			<rows>
+				<row>
+					<cell type="label" labelfor="1"/>
+					<cell type="field" id="1" name="catalogNumber" uitype="formattedtext"/>
+					<cell type="label" labelfor="2"/>
+					<cell type="field" id="2" name="collectionobjecttype" uitype="combobox"/>
+				</row>
+				<row>
```
CarolineDenis, line 1455: we could have a new determination form for this one with only taxon

grantfitzsimmons: Added in [`8dac743` (#6449)](https://github.com/specify/specify7/pull/6449/commits/8dac743d4e66f5273141607e7d5bc5d6b7b98c02)

config/geology/geology.views.xml
grantfitzsimmons: Added in https://github.com/specify/specify7/commit/c2a95bc11948823b0e39d9bba9a3d1a551bc514e

grantfitzsimmons (approved):
I made some changes here:
- I moved the view definition for the "Component" to the `common.views.xml` file to make it easy to add to any discipline interested in using the feature
- I created a `ComponentDetermination` view definition that is independent of the one used for regular Collection Objects
- I created a separate table view definition for "Component" so that the `currentDetermination` virtual field would be shown instead of the `determinations` box when viewed in a grid, but these can be customized independently
- I added `name`, `reservedText`, and `reservedText2` to the "Component" views– we may want to consider renaming these fields in the schema **only if** they have the default label. They _should_ be reserved for us to repurpose.
- Added notes in the XML for COG vs. Component, added URLs to docs
- I hid `catalogNumber` in component since it may not be desirable to number these in many cases as they are not fully elevated objects, but this can be changed

I have some questions, but one important thought came to mind–
- How does auto-numbering work with multiple catalog number fields? During testing, using `#########` in multiple `catalogNumber` fields increments by 1 for each visible field. It appears to order the numbers based on the sequence in which I created the records.
- Is this intuitive enough for queries/WB imports/exports? At the least we need to make it possible to filter between CO components and COs as they are going to be treated much differently. Does this mean we need to update the stats page, default queries, etc.?

acwhite211 (approved):

sharadsw (approved):

grantfitzsimmons (approved):

PR #6462, opened 2025-04-30
Prevent from adding negative count for preparations

### LINKED ISSUE #4843: Preparations table allows countamt to be negative
**Describe the bug**
In the preparations subview for the collection object form, countamt can be negative.

**To Reproduce**

1. Go to a Collection Object form
2. Scroll down to Preparations
3. Decrease countamt to be negative and see that the form can be saved

**Expected behavior**
The minimum for the field should be 0 and the user should not be able to set a value below that.

[Specify 7 System Information - 2024-04-25T19_26_12.430Z.txt](https://github.com/specify/specify7/files/15122356/Specify.7.System.Information.-.2024-04-25T19_26_12.430Z.txt)

- OS: Mac
- Browser: Chrome
- Specify 7 Version: edge, 7.9.3
- Database Name: KUPaleoBot_8_31_2023
- Collection name: Paleobotany
- User Name: sppaleobotadmin

CarolineDenis:
Fixes #4843

### Checklist

### Testing instructions

- Go to a Collection Object form
- Scroll down to Preparations
- Decrease count amount to be negative
- [ ] Verify that you cannot save

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/businessRuleDefs.ts
@@ -16,6 +16,7 @@ import {
   PREPARATION_EXCHANGED_OUT_KEY,
   PREPARATION_GIFTED_KEY,
   PREPARATION_LOANED_KEY,
+  PREPARATION_NEGATIVE_KEY,
 } from './businessRuleUtils';
 import { cogTypes } from './helpers';
 import type { AnySchema, CommonFields, TableFields } from './helperTypes';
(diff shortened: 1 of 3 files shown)


sharadsw (commented):
While this works fine, it might be easier just to update the form definition to enforce a minimum count of 0 from XML.

It can be done with the `initialize` property. Something like this would achieve it
```
<cell type="field" id="3" initialize="min=0" name="countAmt" uitype="text"/>
```

image

emenslin (changes requested):
- [x] Verify that you cannot save 

Error message is fixed, however, since there are 2 different error messgaes if you click out of it the error keeps on growing and adding the line "Preparation count cannot be negative." I feel like this could be a big deal for users who might already have preparation counts set to negative but I can write this up separately if its out of scope. 


https://github.com/user-attachments/assets/7069f831-dc0e-4d18-8040-0516319ad720

sharadsw (approved):

emenslin (changes requested):
- [x] Verify that you cannot save 

Error still grows when it is negative but other than that it works as expected. 

https://github.com/user-attachments/assets/a29ee20d-e2ab-4d92-87d6-5f518af85332

emenslin (approved):
Misclick

combs-a (approved):
- [X] Verify that you cannot save (Default form)
- [X] Verify that you cannot save (Custom form)
<hr>

Looks good! Worked with both default and custom forms.

Error weirdness is also present on Firefox, just to note. Though it is slightly different (only clones once, it might be overflowing outside of the tooltip height?)

image

PR #6465, opened 2025-04-30
Hotfixes: Fullname query hotfixes for `7.10.2`

sharadsw:
Fixes #6463 
Fixes #6459

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good and
      self-explanatory (or properly documented)
- [ ] Add relevant issue to release milestone
- [ ] Add relevant documentation (Tester - Dev)
- [ ] Add automated tests
- [ ] Add a reverse migration if a migration is present in the PR

PR #6473, opened 2025-05-05
Fix ID queries with spaces in rank name

sharadsw:
The hotfix in #6465 broke querying on the ID field. Added a unit test to make sure it doesn't break again

### Checklist

- [x] Add automated tests

### Testing instructions

- [ ] General test queries with rank names (ie: do not use `(any rank)`)
- [ ] General test queries with rank names that have spaces
- Use the db `os_2025_04_30_09_33_44`
  - Go to Trees > Geography
  - Select a node in a rank that has spaces
  - Query from tree
  - [ ] Verify results are same as in version `7.10.0`
  - Select a node in a rank that DOES NOT have spaces
  - Query from tree
  - [ ] Verify results are same as in version `7.10.0`

### DIFF
--- a/specifyweb/stored_queries/queryfieldspec.py
+++ b/specifyweb/stored_queries/queryfieldspec.py
@@ -87,15 +87,15 @@ def find_tree_and_field(table: Table, fieldname: str):
         return tree_rank_and_field[0], mapping[""]
     
     # Handles case where rank name contains spaces
-    if len(tree_rank_and_field) >= 2:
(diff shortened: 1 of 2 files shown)


acwhite211 (approved):

emenslin (approved):
- [x] General test queries with rank names (ie: do not use `(any rank)`)
- [x] General test queries with rank names that have spaces
  - [x] Verify results are same as in version `7.10.0`
  - [x] Verify results are same as in version `7.10.0`

Looks good!

PR #6475, opened 2025-05-05
Handle date precision for partial dates in 'Day' and 'Month' query results

### LINKED ISSUE #4503: 'Day' and 'Month' date fields return `1` when the date is not complete in queries
**Describe the bug**

In the database, Specify stores the dates as 1/1/YYYY when only the year is known and indicated to Specify 6 that the 1s should be ignored based on the DateFieldPrecision value (for precise dates). Specify 7 has not had this logic implemented, therefore you are seeing the issue you described:

> Another thing I noticed just this morning is that the Specify 7 query does not take the date precision in account when calculating `Month` and `Day` values. So, for a record with an `eventDate` of just '1876', Specify 6 gives:
> 
> image
> 
> and Specify 7 gives:
> 
> image
> 
> I only use the Day and Month fields because I thought I could not access the DatePrecision in the query and it turns out that I was wrong, so I can probably fix this myself. However, I thought I'd note the discrepancy between Specify 6 and 7. People might rely on the Month and/or Day fields to be blank to know that they are dealing with an incomplete date.

**To Reproduce**

1. Build a query on a date field
2. Make sure the date is set to just a year
3. Query on 'Day', 'Month', and 'Year'
4. See that Day and Month return `1` instead of nothing

**Expected behavior**
It should show no number, not `1`.

**Reported By**
@nielsklazenga on the [Speciforum](https://discourse.specifysoftware.org/t/issues-with-dates-and-timestamps-in-specify-7-query/1538)

Areyes42:
Fixes #4503

This fixes an issue where queries on partial dates (like year-only) were returning 1 for 'Day' and 'Month'. Now, the code checks the precision of the date before extracting parts of it. 'Day' is only returned if the date is fully known, and 'Month' is only returned if the precision includes the month. 'Year' is always returned. If the date doesn’t have a precision field, the old behavior is used.

### Checklist

- [ ] Add automated tests

### Testing instructions

1. Make/edit a record and set the date value to a year-only format (e.g., CO Cat Date: 2025)
2. Build a query on a date field and include  'Day', 'Month', and 'Year'
3. Run the query
    - [ ] Verify that Day and Month return nothing, instead of 1
    - [ ] Verify other date fields did not get changed (e.g., Date edited, Timestamp Created)
4. Return to the record and modify the date format to full date and populate it
    - [ ] Re-run the query and verify it shows the correct dates


emenslin (approved):
- [x] Verify that Day and Month return nothing, instead of 1
    - [x] Verify other date fields did not get changed (e.g., Date edited, Timestamp Created)
    - [x] Re-run the query and verify it shows the correct dates

Looks good! 
Screenshot 2025-05-06 084725

pashiav (approved):
- [x] Verify that Day and Month return nothing, instead of 1
- [x] Verify other date fields did not get changed (e.g., Date edited, Timestamp Created)
- [x] Re-run the query and verify it shows the correct dates
---

Looks good!

**Before:**
Screenshot 2025-05-06 at 10 33 41 AM
**After:**
Screenshot 2025-05-06 at 10 33 56 AM

sharadsw (approved):

acwhite211 (approved):

PR #6476, opened 2025-05-06
Change new root to Life

### LINKED ISSUE #4371: Can't Edit Tree Roots in Detail / Record View
**Describe the bug**
If you open root record of a Tree. You can't make any changes to it as Specify vocally complains and cites it as an invalid record due to the lack of a parent.

**To Reproduce**
1. Open the Root Record of any tree
2. Make changes. I.E Rename it or something
3. Try Saving
4. See Error (Requires Parent)

**Expected behavior**
It would be pretty cool if we make roots an exception to the Parent Requirement for the Root. 

**Screen Recording**

https://github.com/specify/specify7/assets/53784701/be470b19-a5d4-4bf2-a7bd-977e17f94c7c

https://github.com/specify/specify7/assets/53784701/4824035b-0bf2-4ea2-aef8-342446939f0a

**Desktop:**
 - OS: Mac OS
 - Browser: Chrome
 - Specify 7 Version: 7.9 and XML Editor

**Database Name:** NHMD

CarolineDenis:
Fixes #4371

### Checklist

### Testing instructions

- Open taxon tree view 
- Create a new tree 
- Reload page 
- Click on the + in the node section to add the root 
- [ ] Verify it is now 'Life' instead of 'Root'

### DIFF
--- a/specifyweb/specify/tree_views.py
+++ b/specifyweb/specify/tree_views.py
@@ -437,14 +437,14 @@ def add_root(request, tree, treeid):
         raise Exception("Root node already exists.") 
 
     root = item.objects.create(
-        name="Root",
+        name="Life",
         isaccepted=1,
         nodenumber=1,
         rankid=0,
         parent=None,
         definition=tree_target,
         definitionitem=tree_def_item,
-        fullname="Root"
+        fullname="Life"
     )
 
     # Override node number with raw sql execution


acwhite211 (approved):

sharadsw (approved):

combs-a (commented):
- [X] Verify it is now 'Life' instead of 'Root' 
<hr>
The change is there, but I'm not sure that this should close the issue, since renaming it doesn't address the 'unable to edit' issue. Plus, Root is a more encapsulating name than Life generally (since we're introducing geology, and Life isn't applicable with them).

However, this can be done using batch edit. It's a bit roundabout, but possible.

CarolineDenis:
@combs-a please see https://github.com/specify/specify7/issues/4371#issuecomment-2845500324

combs-a:
> @combs-a please see [#4371 (comment)](https://github.com/specify/specify7/issues/4371#issuecomment-2845500324)

I do still have concerns. The comment Theresa left was more about being unable to change it if a collection wanted to, not that the default was not Life.

Since we are supporting Geology disciplines, Root is more encompassing overall. I feel that it is more likely that users with a geo collection will want to change it from Life -> Root versus other disciplines wanting to go from Root -> Life, especially since multiple trees were mostly intended for them.

Currently, the root can be edited from either the back-end or by a batch edit workaround. I personally think that these solutions, while not ideal, are fine for now, especially since the likelihood of someone asking for changing the root name from Root is relatively low in my opinion.

grantfitzsimmons (commented):
>The only way to achieve this would be to remove the data model restriction requiring each node to have a parent. However, doing so would introduce complexity, as we’d need to enforce a business rule to catch missing parents for all nodes that do require one. Since this issue only affects the root, the benefits don’t justify the risk. Instead, I can rename the root node to "life."

I think the best solution (considering the taxon trees for geology and other disciplines) is to use a unique view definition or ignore this business rule when the node is the root. 

Specify 6 enabled users to rename the 'Life' node in the same manner as renaming any node, except the parent query combo box was rendered as read-only.

image

We can already define a conditional view definition to hide the parent query combo box for tree records:

https://github.com/user-attachments/assets/1db6dcdc-098f-4c77-a7e0-2f4b937c43fb

```xml
      <cell type="panel" colspan="6">
        <rows>
          <row>
              <cell type="label" labelfor="parent"/>
        			<cell type="field" id="parent" uitype="querycbx" isrequired="true" initialize="name=Taxon;title=Taxon;editbtn=false;newbtn=false;editoncreate=true" name="parent" colspan="3"/>
          </row>
        </rows>
        <rows condition="definitionItem.rankId=0">
          <row>
            <cell type="label" label="This is a root node!" />
          </row>
        </rows>
      </cell>
```

We would need to not enforce the `Taxon - Parent: Field is required.` business rule in this case alone, where the `definitionItem.rankId=0`.

config/backstop/global.views.xml
```diff
@@ -1272,10 +1272,25 @@
...
+			<cell type="panel" colspan="4" initialize="align=left">
+				<rows>
+					<row>
+						<cell type="label" labelfor="parent" initialize="align=left"/>
+						<cell type="field" id="parent" uitype="querycbx" isrequired="true" initialize="name=Taxon;title=Taxon;editbtn=false;newbtn=false;editoncreate=true;align=left" name="parent" colspan="3"/>
+					</row>
+				</rows>
+				<rows condition="definitionItem.rankId=0">
+					<row>
+						<cell type="label" label="This is root node"/>
```
CarolineDenis, line 1289: need to chnage this

specifyweb/specify/tree_views.py
```diff
@@ -437,14 +437,14 @@ def add_root(request, tree, treeid):
         raise Exception("Root node already exists.") 
 
     root = item.objects.create(
-        name="Root",
+        name="Life",
```
CarolineDenis, line 440: do we want it back to root?

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
```diff
@@ -379,6 +379,20 @@ export function QueryComboBox({
       onClick={(): void => handleOpenRelated(true)}
     />
   );
+
+  const isTree =
```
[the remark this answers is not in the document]
grantfitzsimmons: Yes

CarolineDenis:
NOTES: 
- Change back to root 
- Change view def to conditional for all trees

specifyweb/specify/tree_views.py
grantfitzsimmons: Yes

PR #6483, opened 2025-05-08
Make resource open from formatter and agg readonly

### LINKED ISSUE #4795: View button does not open form in read-only in table formats and aggregation
**Describe the bug**
When clicking on the view button under preview in either table formats or aggregation the form does not open in read only and you can make changes and save them. 

**To Reproduce**

1. Go to record formatters
2. Click on any table (e.g. accesssion)
3. Scroll down to preview
4. Click on one of the view buttons next to a record 
5. See you can make a change to it and save

**Expected behavior**
The form should open in read-only and you should not be able to make any changes 

**Screenshots**

https://github.com/specify/specify7/assets/150940840/655db5cc-364a-4e3c-ba12-ea4a260a159b

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge
- Collection name: mammals
- User Name: Spadmin

[Specify 7 System Information - 2024-04-16T19_11_18.890Z.txt](https://github.com/specify/specify7/files/15001594/Specify.7.System.Information.-.2024-04-16T19_11_18.890Z.txt)

CarolineDenis:
Fixes #4795

### Checklist

### Testing instructions

- Go to record formatters
- Click on any table (e.g. accesssion)
- Scroll down to preview
- Click on one of the view buttons next to a record
- [ ] Verify the form resource in readonly

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Formatters/Preview.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Formatters/Preview.tsx
@@ -9,6 +9,7 @@ import type { GetOrSet, RA } from '../../utils/types';
 import { removeItem } from '../../utils/utils';
 import { Button } from '../Atoms/Button';
 import { Link } from '../Atoms/Link';
+import { ReadOnlyContext } from '../Core/Contexts';
 import { fetchCollection } from '../DataModel/collection';
 import type { AnySchema } from '../DataModel/helperTypes';
 import type { SpecifyResource } from '../DataModel/legacyTypes';
@@ -139,7 +140,16 @@ export function ResourcePreview({
     ),
     false
   );
-  return children((_, index) =>
-    formatted === undefined ? commonText.loading() : formatted[index]
+  /*
+   * Return children((_, index) =>
+   *   formatted === undefined ? commonText.loading() : formatted[index]
+   * );
+   */
+  return (
+    <ReadOnlyContext.Provider value>
+      {children((_, index) =>
+        formatted === undefined ? commonText.loading() : formatted[index]
+      )}
+    </ReadOnlyContext.Provider>
   );
 }


emenslin (changes requested):
- [ ] Verify the form resource in readonly 

You can't edit any of the fields but you can still delete records. 


https://github.com/user-attachments/assets/261ed07f-6fa2-4a68-ac51-cc0d6438a7e8

CarolineDenis:
@emenslin , there is no easy way to remove that delete button. 
Since there is a warning saying that it will permanently delete the resource, this is acceptable.

sharadsw (approved):

acwhite211 (approved):

melton-jason:
I think there might be an easy way to hide the delete button from the form! At least, by easy I mean the code change is _very_ minimal: only a single line (although the impacts of such change might not be "easy": another two-line solution is also proposed which currently has far less of an impact)

For some context, the reason that the Delete Button is hidden when using `Ready Only Mode` compared to the current solution on this branch is with the `hasTablePermission` line in the following snippet which controls whether or not to show the the delete button:

https://github.com/specify/specify7/blob/b0e4c8baf53e701274e26b9e5e47e86eaf7c0195/specifyweb/frontend/js_src/lib/components/Forms/ResourceView.tsx#L251-L257

The relevant line within `hasTablePermission` is that it checks the value of `readOnlyMode` in the cache: which is set to true whenever readonly mode is enabled

https://github.com/specify/specify7/blob/b0e4c8baf53e701274e26b9e5e47e86eaf7c0195/specifyweb/frontend/js_src/lib/components/Permissions/helpers.ts#L31

I think we could probably not show the Delete Button if the value of `ReadOnlyContext` is true. 
Of course, the impact of this change could be wide-spread, and there could be places which are readonly and should show the Delete Button: we would need to go through the uses of ReadOnlyContext.Provider to make sure. 
There is at least one potentially beneficial change: when rendering something like `CO -> accession -> collectionObjects` as a button, the base resource will no longer have a delete button in the independent relationship's subview:

https://github.com/user-attachments/assets/bb7c4906-c5eb-4a91-847b-22078be2b898

If we want a more isolated (less wide-spread) approach to this, or if there might be cases where we want to show the delete button when the form is readonly, I imagine we could add `InFormEditorContext` to the Record Formatter preview. Conceptually, this would be treating `InFormEditorContext` more as a "preview" context. 

https://github.com/specify/specify7/blob/dbb2f9a80eae1a49d026c1224dc592d0f63cfffc/specifyweb/frontend/js_src/lib/components/Formatters/Preview.tsx#L144-L148

The only (current) downside of doing so would be that the Form Meta menu would not be visible.

pashiav (changes requested):
- [ ] Verify the form resource in readonly 
---
**Bugs**

**1. Recieved `OperationalError at /api/specify/collectionobject/ (1054, "Unknown column 'collectionobject.ParentCOID' in 'SELECT'")` upon clicking View.** 
Clicking on the eyeball icon to see the resource causes an OperationalError. 

_https://calvertmarinemuseum202502102-issue-4795.test.specifysystems.org/specify/resources/app-resource/13/formatter/Accession/_
_Collection: Invertebrates_

https://github.com/user-attachments/assets/4818e08c-fbb2-4978-af9b-8ff3ea426e07


[Specify 7 System Information - 2025-06-09T14_58_01.692Z.txt](https://github.com/user-attachments/files/20656821/Specify.7.System.Information.-.2025-06-09T14_58_01.692Z.txt)

pashiav (approved):
- [x] Verify the form resource in readonly 
---
Looks good - form is read-only!
Screenshot 2025-06-09 at 12 52 18 PM

PR #6501, opened 2025-05-12
Fix build reliablity with re-try loops in Dockerfile

### LINKED ISSUE #6480: Specify not building for `amd64` systems when pushed to Docker Hub
Docker Hub images

The GitHub workflow we use for building & pushing Specify updates to Dockerhub is failing. This action takes place automatically after an update is pushed on GitHub:

https://github.com/specify/specify7/blob/production/.github/workflows/docker.yml

We added support for `arm64` in this PR: https://github.com/specify/specify7/pull/4068
We made it faster in this PR: https://github.com/specify/specify7/pull/4201
Finally, we fixed the dependencies preventing it from building on `arm64` in January: https://github.com/specify/specify7/pull/6152

One of us has had to build & push to Dockerhub each time manually, using a system like this:

```
 git fetch --all
git switch -c v7.10.2
docker buildx build \
  --platform linux/amd64,linux/arm64 \
  --tag specifyconsortium/specify7-service:v7 \
  --push .
```

acwhite211:
Fixes #6480

Fix the GitHub Action Workflow arm64 build reliability by wrapping problematic docker run commands, involving installs, in re-try loops.

### Checklist

### Testing instructions

Dev testing only
- [x] See that the arm64 build GitHub Action Workflow action completes successfully, and that image appears in docker hub.


acwhite211:
Had to add a re-try loop for the apt-get command as well 🤦

acwhite211:
Built successfully again.

Every apt-get and python package install seems to be a point where the GitHub workflow action for the arm64 build can fail.  In the future, I think I'll clean up our Dockerfile, and then create a separate Dockerfile that will just be used for the GitHub workflow actions arm64 build.

acwhite211:
I can edit the GitHub action so that it uses `Dockerfile_GitHub` instead of the `Dockerfile`, that way we could keep our Dockerfile cleaner, without these special re-try loops for the GitHub workflow actions arm64 build.  The downside would be that we would need to update both files when making Dockerfile changes.  Thoughts and opinions? @specify/developers

acwhite211:
Created a branch for having two Dockerfiles https://github.com/specify/specify7/compare/main...issue-6480-2

CarolineDenis (approved):

acwhite211:
The council has spoken, one Dockerfile it shall be 👑

PR #6523, opened 2025-05-16
Add a dialog to force saving the query before Batch Edit

### LINKED ISSUE #6516: Batch Edit: Clicking Cancel when asked to leave the query builder still creates a batch edit dataset in the background
**Describe the bug**
If the user selects `Cancel` when asked "Are you sure you want to leave this page?" by the query builder a Batch Edit dataset will still be created in the background. This leads to duplicate datasets being created if the user realizes they have not saved the query and would like to, as they then save the query under a name, and hit batch edit again.

**To Reproduce**

1. Create a new query on any table that is able to be batch edited
2. Select any fields that are able to be batch edited
3. Click `Batch Edit`
4. When asked if you would like to leave the page without saving, press `Cancel`.
5. Click on `Batch Edit` in the navigation menu
6. See that a dataset has been created anyways

**Expected behavior**
A dataset would not be created if the user cancels leaving the query page to go back and save the query.

- Specify 7 Version: v7.10.2.3
- Database Name: sp7demofish

**Reported By**
Beaty Biodiversity Museum

sharadsw:
Fixes #6516

- Adds a dialog to force users to save the query before running Batch Edit
- The previous behavior was the standard unload protect used in Specify which is triggered when the application tries to navigate elsewhere while having unsaved changes
- However, unload protect behavior cannot be changed without causing side effects everywhere and so this is a workaround to force users to save their query

### Checklist

- [ ] Add automated tests

### Testing instructions

- Create a query but do not save it
- Run Batch Edit
- [ ] Verify there is a dialog forcing users to save their query when the query has unsaved changes

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/BatchEdit/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/BatchEdit/index.tsx
@@ -2,7 +2,10 @@ import React from 'react';
 import { useNavigate } from 'react-router-dom';
 import type { LocalizedString } from 'typesafe-i18n';
 
+import { useBooleanState } from '../../hooks/useBooleanState';
 import { batchEditText } from '../../localization/batchEdit';
+import { commonText } from '../../localization/common';
+import { queryText } from '../../localization/query';
 import { ajax } from '../../utils/ajax';
 import type { RA } from '../../utils/types';
 import { filterArray } from '../../utils/types';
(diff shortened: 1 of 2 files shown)


emenslin (approved):
- [x] Verify there is a dialog forcing users to save their query when the query has unsaved changes

Looks good!

alesan99 (approved):
- [X] Verify there is a dialog forcing users to save their query when the query has unsaved changes
---

Works! 👍
image

PR #6527, opened 2025-05-20
Batch Edit: Fix uploaded taxon using incorrect fullname

### LINKED ISSUE #6526: Tree definition gets carried over, when changing trees using batch edit. Fullname is not correctly set in these cases.
**To Reproduce**
1. Make sure the discipline has two taxon trees. The taxon trees must also share at least one rank. In this example, two taxon trees are used, and one of the common ranks is "Subspecies". This rank is also the last rank in the respective trees.
2. Create a collection object with 1 determination and set the taxon to a "Subspecies". In this example, the selected taxon is "bugReportSubSpecies1". **Note that the taxontreedefID  for this tree is 1**. 

Image

3. Construct a query, for batch edit, like below. Select the common rank (Subspecies). The batch edit should add this rank for both the trees (since Subspecies rank exists in the other tree too).

Image

4. Run the query and click on "Batch Edit". Verify that the dataset construction added the "Subspecies" rank from the other tree too, similar to below:

Image

5. The other "Subspecies" column, in this example, is "Ichthyology - Subspecies - Taxon name". Remove the value from "Taxon - Subspecies - Taxon name", and added some value in "Ichthyology - Subspecies - Taxon name". This is a legal move, and, effectively, can be used to change trees, so should work. **Moreover, the other tree is taxontreedefid of 2. So, if a new taxon node is created, the new taxon should have the taxontreedefid of 2, and not 1 -- this is the bug that is shown here**.

Image

6. Click on Data Mapper > Batch Edit Preferences. Uncheck "Use only visible fields for match". Click Apply, and then Save.

Image

7. Go back to Data Editor and click on Validate and then click on Commit. The following result is expected:

Image

8. Go to the newly created taxon, in the form viewer. The form is **incorrect**, since the fullname and name do not match. That is, the fullname was incorrectly not reset for this taxon, as the fullname still contains the phrase "bugReportSubSpecies1", which was the name of the previous taxon name. 

Image

9. The id of this taxon is `14513`. Go to `/api/specify/taxon/14513`, and carefully notice the definition.

Image

**Expected behavior**
The bug is that the definition is incorrect. The taxon should be created in the Ichthyology tree definition, which is `/api/specify/taxontreedef/2`. At a database level, it can be verified that the definition was copied over from the previous taxon (from the original tree). Morever, all the fields are copied from the previous taxon node (this is almost correct, the fields should be copied over, but not across treedefinitions).

sharadsw:
Fixes #6526

Check issue for notes on the bug

### Checklist

- [ ] Add automated tests

### Testing instructions

1. Make sure the discipline has two taxon trees. The taxon trees must also share at least one rank. In this example, two taxon trees are used, and one of the common ranks is "Subspecies". This rank is also the last rank in the respective trees.
2. Create a collection object with 1 determination and set the taxon to a "Subspecies". In this example, the selected taxon is "bugReportSubSpecies1". **Note that the taxontreedefID  for this tree is 1**. 

Image

3. Construct a query, for batch edit, like below. Select the common rank (Subspecies). The batch edit should add this rank for both the trees (since Subspecies rank exists in the other tree too).

Image

4. Run the query and click on "Batch Edit". Verify that the dataset construction added the "Subspecies" rank from the other tree too, similar to below:

Image

5. The other "Subspecies" column, in this example, is "Ichthyology - Subspecies - Taxon name". Remove the value from "Taxon - Subspecies - Taxon name", and added some value in "Ichthyology - Subspecies - Taxon name". This is a legal move, and, effectively, can be used to change trees, so should work. **Moreover, the other tree is taxontreedefid of 2. So, if a new taxon node is created, the new taxon should have the taxontreedefid of 2, and not 1 -- this is the bug that is shown here**.

Image

6. Click on Data Mapper > Batch Edit Preferences. Uncheck "Use only visible fields for match". Click Apply, and then Save.

Image

7. Go back to Data Editor and click on Validate and then click on Commit. The following result is expected:

Image

8. Go to the newly created taxon, in the form viewer. 
9. [ ] Verify the name and full name are correct


CarolineDenis (approved):

emenslin (approved):
9. [x] Verify the name and full name are correct 

The only potential problem I noticed is that nothing seems to be wrong with the CO after upload even though the COT and determination don't match. Since this is already technically possible in production I won't request changes, but I wanted to make note of it.

combs-a (approved):
- [X] Verify the name and full name are correct 
<hr>

Didn't run into any issues, new full name was constructed correctly! Good work!

PR #6579, opened 2025-06-05
Replace selectseries with smushed

### LINKED ISSUE #6575: `SelectSeries` replicates `Smushed` purpose
In the `spquery` table:

- `Smushed` indicates if a query has the "Series" checkbox checked in Specify 6
- `SelectSeries` indicates if a query has the "Series" checkbox checked in Specify 7

We have a couple of options:
1. *Best option:* We create a new migration that removes the `SelectSeries` field and uses instead the `Smushed` field, meaning series queries from Specify 6 (including those for reports & labels) will work out of the box without any changes
2. *Second option:* We create a new migration that updates the `SelectSeries` field and sets it to `1` where `Smushed` = `1`, which will handle existing queries but not new ones.

acwhite211:
Fixes #6575

Create a new migration that removes the SelectSeries field and uses instead the Smushed field, meaning series queries from Specify 6 (including those for reports & labels) will work out of the box without any changes.  Also, needed to fix all the references to the SelectSeries field.

### Checklist

### Testing instructions

- [x] Run series queries and see that they runs successfully. See original series QB PR for more testing instructions https://github.com/specify/specify7/pull/4952


grantfitzsimmons:
>Do we want to change this to use v711-prerelease as the base branch or keep it as is with main?

I'd say `v711-prerelease`

CarolineDenis:
@grantfitzsimmons , agree with you

grantfitzsimmons (changes requested):
The "Series" checkbox seems to have no effect, both in the query results and after saving the record:

image


https://github.com/user-attachments/assets/703878c8-e407-43dd-a978-dbeedbc3a37a

grantfitzsimmons:
@acwhite211 This needs to be based on `v711-prerelease`, not `main`. We need to rebase without bringing over the other changes

PR #6587, opened 2025-06-05
Use user settings for interactions & tables list dialog + remove legacy option

### LINKED ISSUE #6387: Deprecate `dataentry_task.xml` and use 7-configured items instead
By default, Specify uses the tables defined in the data entry XML file for each discipline or the custom app resource (`DataEntryTaskInit ` app resource) under each user. We need to deprecate this and instead use the unique user preferences resource to determine which tables are used 

Currently, we use "Copy Specify 6 settings" as the default, but we should outright remove it.

Image

(Click on data entry in the side bar menu, than on the pencil) 

The tables shown in the dialog for 7 users are captured in the `UserPreferences` app resource:

```json
"general":{"shownTables":[1,10,2,4,5,3,121,69,7]}}
```
Expected behavior: 
Remove this dialog and always use custom settings

CarolineDenis:
Fixes #6387

### Checklist

### Testing instructions

- Click on data entry in the side bar 
- Click on the pencil 
- [ ] Verify it directly opens the "Configure data entry tables" dialog
- Add a table
- Close 
- Click on data entry in the side bar 
- [ ] Verify the new table you added as been added 
- Click on interaction in the side bar 
- [ ] Verify the table you added for data entry as not been added to interaction list 
- Click on the pencil 
- [ ] Verify it directly opens the "Configure interaction tables" dialog
- Add a table
- Close 
- Click on interactions in the side bar 
- [ ] Verify the new table you added as been added


pashiav (approved):
- [x] Verify it directly opens the "Configure data entry tables" dialog
- [x] Verify the new table you added as been added 
- [x] Verify the table you added for data entry as not been added to interaction list 
- [x] Verify it directly opens the "Configure interaction tables" dialog
- [x] Verify the new table you added as been added 
---
Looks good - removes source of tables dialog and works as expected!

melton-jason (changes requested):
Looking pretty clean! 👌 

Nothing critical that needs changed/considered: all of my concerns are really just for code maintainability and clarity. 

Apart from the below comments, the only other discussion left for this PR is whether we want to initially set the user's Specify 7 settings to whatever they had in Specify 6, rather than using a default. 
It might be a minor annoyance if people had configured their [DataEntryTaskInit](https://discourse.specifysoftware.org/t/app-resources-in-specify-7/582#data-entry-tables-documentationhttpsgithubcomspecifyspecify6blobmasterconfigfishdataentry_taskxml-12) and/or [InteractionsTaskInit](https://discourse.specifysoftware.org/t/app-resources-in-specify-7/582#interactions-tables-documentationhttpsgithubcomspecifyspecify6blobmasterconfigcommoninteractionstaskxml-13) and are expecting a consistent DataEntry/Interactions table list view between Specify 6 and Specify 7 and now have to set their Specify 7 configuration. 

Although again, that is arguably _very_ minor and I wouldn't mind saying (in the release notes) users would need to take the one-time step after updating to set their table list back to the way it is in Specify 6, and that their DataEntry/Interaction table list will no longer be synched from Specify 6 and 7.

CarolineDenis:
@melton-jason, 
We will let them know that they need to take the one-time step after updating to set their table list back to the way it is in Specify 6

specifyweb/frontend/js_src/lib/components/DataEntryTables/Edit.tsx
melton-jason, line 1: Ah, one more change in this file, whoops. 

When the user has a `"legacy"` option, no tables are shown

https://github.com/user-attachments/assets/56d587e8-998d-4aee-9f5b-cf313fd686c9

Instead of `tables.length === 0` when determining  the `visibleTables`: 

https://github.com/specify/specify7/blob/d992f7dafc59488164bf9abef08974e8038d776d/specifyweb/frontend/js_src/lib/components/DataEntryTables/Edit.tsx#L44-L47

I think we should be using `tableIds` instead (`tables` can still be `"legacy"`, so in that case the length of the string is not zero and `visibleTables` will try and iterate through all of `tableIds` to get the tables: which will be empty). 

If it's clearer, we could merge `tableIds` and `visibleTables`, and replace all of: 

https://github.com/specify/specify7/blob/d992f7dafc59488164bf9abef08974e8038d776d/specifyweb/frontend/js_src/lib/components/DataEntryTables/Edit.tsx#L41-L47

with something like: 

```ts
  const visibleTables =
    Array.isArray(tables) && tables.length > 0
      ? tables.map(getTableById)
      : filterArray(defaultVisibleForms[type].map(getTable));
```

PR #6599, opened 2025-06-09
Add a warning dialog before committing changes through batch Edit

### LINKED ISSUE #6596: Add warning dialog before committing a batch edit data set
We need something just like this:

Image

when committing changes from a Batch Edit data set.

>Before proceeding, please note that the following action may interrupt other users. This action may cause delays or temporary unavailability of certain features for Specify users. Please consider the impact on their experience. This **merge** cannot be undone

If we want to use the exact same dialog, we would need to change "Merge Records" to something like "Commit Data Set" in this case, then change "merge" to "action"

Users should also be able to check "Don't show this again" to not be prompted with this warning repeatedly.

CarolineDenis:
Fixes #6596

### Checklist

### Testing instructions

- Create a batch edit data set 
- Validate 
- Save 
- Click on commit 
- [ ] Verify there is a new warning dialog saying "'Before proceeding, please note that the following action may interrupt other users. This action may cause delays or temporary unavailability of certain features for Specify users. Please consider the impact on their experience. This action cannot be undone'" 
- [ ] Verify you can still commit your changes once you proceed

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/WbActions/WbUpload.tsx
+++ b/specifyweb/frontend/js_src/lib/components/WbActions/WbUpload.tsx
@@ -102,7 +102,7 @@ export function WbUpload({
           className={{
             container: dialogClassNames.narrowContainer,
           }}
-          dimensionsKey="merging-warning"
+          dimensionsKey="batchEdit-warning"
           header={batchEditText.commitDataSet()}
           onClose={undefined}
         >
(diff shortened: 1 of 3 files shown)


acwhite211 (approved):

alesan99 (approved):
The dialog no longer shows up when the checkbox is checked 👍👍

PR #6605, opened 2025-06-09
Toggle save query when selecting distinct

### LINKED ISSUE #6591: Checking `Distinct` or `Series` does not trigger save button / unload protect
**Describe the bug**
Checking `Distinct` or `Series` does not trigger save button / unload protect, requiring the user to make another change in the query to save the change.

**To Reproduce**
1. Go to a query
2. Check "Distinct" or "Series"
3. See that the save button is not enabled

**Expected behavior**
The user should be able to save immediately after checking either of these boxes. 

**Screenshots**

https://github.com/user-attachments/assets/9e1e3cff-72f9-4b65-ac6f-97d758de4347

**Crash Report**
Occurs in `v7.10.2.3` and `v711-prerelease`

CarolineDenis:
Fixes #6591

### Checklist

### Testing instructions

- Go to a query
- Check "Distinct"
- [ ] Verify save query gets enabled. 

Notes: Series is not remembered between sessions

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Attachments/__tests__/UploadAttachment.test.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Attachments/__tests__/UploadAttachment.test.tsx
@@ -1,24 +1,25 @@
-import React from 'react';
-import { mount } from '../../../tests/reactUtils';
-import { UploadAttachment } from '../Plugin';
-import { clearIdStore } from '../../../hooks/useId';
-import { LoadingContext } from '../../Core/Contexts';
-import { f } from '../../../utils/functools';
 import { fireEvent, waitFor } from '@testing-library/react';
-import { overrideAttachmentSettings } from '../attachments';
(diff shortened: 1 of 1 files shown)


emenslin (changes requested):
- [ ] Verify save query gets enabled. 

Save query does get enabled, but if you check series and save it doesn't actually save your selection but it does work for distinct.  

Series: 

https://github.com/user-attachments/assets/186ab123-abf2-4bee-a807-283984912afb


Distinct: 

https://github.com/user-attachments/assets/27ef7070-9adb-41e4-989f-3b86e37d5c80

emenslin:
If checking select series and saving does not save the preference than shouldn't the save just not be enabled when checking select series? As a user I would think that since save was enabled after checking it that when I pressed save that it would save my preference.

CarolineDenis:
@emenslin, yes you're right. Updated

PR #6606, opened 2025-06-09
Add Base Table selection for attachment uploads in WB

### LINKED ISSUE #6600: Add a Dialog to select tables that support Attachments (WB Attachments)
Uploading attachments through the workbench in https://github.com/specify/specify7/pull/6493 is only supported if the base table supports attachments, so there should be a special Base Table select dialog that only displays those tables.

The existing base table dialog should be reused (or at least as much of it as possible) only it should be filtered further to only show tables that have related attachment tables (don't show the attachment tables themselves)

alesan99:
Fixes #6600

**WIP**

### Checklist

- [ ] Add automated tests


emenslin (approved):
- [x] See that only base tables that support attachments are shown (You can check in http://localhost/specify/data-model/)
- [x] See that hidden tables are shown, but only if they support attachments
- [x] Check that the dialog behaves the same in other contexts (life if you make a data set without an upload plan then attempt to add an upload plan, or if you change the base table)
- [x] Make sure normal data sets still display all tables, even those without attachments

--- 

Looks good, I didn't run into any issues!

CarolineDenis (approved):

PR #6619, opened 2025-06-11
Fix downloading single attachment with "Download All" on Firefox

### LINKED ISSUE #6536: Download All button does nothing when there is only one attachment previewed
**Describe the bug**
If the user runs a query for `Collections Objects` with attachments, selects one, browses in forms, clicks the attachments icon, then presses `Download All` nothing will happen. This appears to be because a route to `/attachment_gw/proxy` is failing.

**To Reproduce**

1. Go to Queries
2. Search for Collection Objects with Attachments
3. Select one through using the checkbox
4. Click `Browse in Forms`
5. Select the attachment icon
6. Click `Download All`
7. Observe that nothing happens

**Expected behavior**
One image would be zipped and downloaded (same as if there were 2, or 3 etc)

**Screenshots**

Image

Image

**Crash Report**

N/A

- Browser: Firefox `138.0.4`
- Specify 7 Version: `7.10.2.3`
- Database Name: sp7demofish and beaty

**Reported By**
Beaty Biodiversity Museum

**Additional context**
Originally the button seems to have been disabled when there was a single attachment, however it passed QA at that time. https://github.com/specify/specify7/pull/6052#issuecomment-2615895253. This commit may also be relevant: 3f92de514d28916307fbb5095fb90da751f2cb0f

alesan99:
Fixes #6536 

It appears Firefox is stricter about downloading files from iframes than Chrome.
No other file downloads in Specify should be affected by this bug, since this was a unique case of a file being downloaded from a URL rather than being created by the front end (aka exports like CSVs).

### Checklist

- [ ] Add automated tests
- [x] Add a reverse migration if a migration is present in the PR

### Testing instructions

- Use Firefox
- Go to a record with a single attachment
- Click on the gallery button
image
- Click "Download All"
- [ ] Make sure the attachment was downloaded (it shouldn't be in a zip file)
- [ ] Make sure it works on Chrome too

NOTE: This should be tested on the test panel. I don't believe the issue can be recreated locally, but you can make sure you are testing it correctly by recreating the original issue.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Attachments/__tests__/UploadAttachment.test.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Attachments/__tests__/UploadAttachment.test.tsx
@@ -19,7 +19,7 @@ import { testAttachment } from './utils';
 requireContext();
 
 async function uploadFileMock(): Promise<SpecifyResource<Attachment>> {
-  return deserializeResource(testAttachment) ;
+  return deserializeResource(testAttachment);
 }
 
 beforeEach(() => {


CarolineDenis (approved):

pashiav (approved):
- [x] Make sure the attachment was downloaded (it shouldn't be in a zip file)
- [x] Make sure it works on Chrome too
---
Looks good - working on both Firefox and Chrome!

https://github.com/user-attachments/assets/d11770be-bcba-4966-bb87-c328cb438ab9

combs-a:
Couldn't recreate the issue on production personally so I can't test this PR; I'm on Firefox version 139.0.4 which is one ahead of the version reported in the issue. Noting this in case others have issues recreating the issue.

bhumikaguptaa (approved):
- [X] Make sure the attachment was downloaded (it shouldn't be in a zip file)
- [X] Make sure it works on Chrome too

Looks good!

PR #6625, opened 2025-06-11
Fetch attachment urls on the backend when download from record sets

alesan99:
Fixes #

### Checklist

- [ ] Add automated tests

### DIFF
--- a/specifyweb/attachment_gw/views.py
+++ b/specifyweb/attachment_gw/views.py
@@ -10,6 +10,7 @@ from os.path import splitext
 from uuid import uuid4
 from xml.etree import ElementTree
 from datetime import datetime
+from django.apps import apps
 
 import requests
 from django.conf import settings
@@ -22,6 +23,9 @@ from django.views.decorators.http import require_POST
 
 from specifyweb.middleware.general import require_http_methods
 from specifyweb.specify.views import login_maybe_required, openapi
+from specifyweb.specify import models
+from specifyweb.specify.models import Recordsetitem
+from specifyweb.specify.models_by_table_id import get_model_by_table_id
 
 from .dataset_views import dataset_view, datasets_view
 logger = logging.getLogger(__name__)
@@ -319,6 +323,27 @@ def download_all(request):
 
     attachmentLocations = r['attachmentlocations']
     origFileNames = r['origfilenames']
+
+    # Optional record set parameter
+    # Fetches all the attachment locations instead of using the ones provided by the frontend
+    recordSetId = r.get('recordsetid', None)
+    if recordSetId is not None:
+        attachmentLocations = []
+        recordset = models.Recordset.objects.get(id=recordSetId)
+        tableid = recordset.dbtableid
+        table = get_model_by_table_id(tableid)
+
+        join_table = apps.get_model(table._meta.app_label, table.__name__ + 'attachment')
+
+        # Reach all attachments
+        recordsetitem_ids = models.Recordsetitem.objects.filter(recordset__id=recordSetId)
+        for rsi in recordsetitem_ids:
+            record_id = rsi.recordid
+            join_records = join_table.objects.filter(**{table.__name__: record_id})
+            for join_record in join_records:
+                attachment = join_record.attachment
+                if attachment.attachmentlocation is not None:
+                    attachmentLocations.append(attachment.attachmentlocation)
 
     filename = 'attachments_%s.zip' % datetime.now().isoformat()
     path = os.path.join(settings.DEPOSITORY_DIR, filename)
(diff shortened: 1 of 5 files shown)


alesan99:
@specify/dev-testing Here's something I want to draw attention to:
Large record sets were represented with an array of ids of loaded records and padded with undefined to show the unloaded records `[1, 2, 3, undef, undef, undef]`
Large queries were represented with an array that only had the loaded record ids `[1, 2, 3]`.

This meant that the true number of records was misrepresented on large queries. (Also, though expected, the undefined values weren't entirely handled correctly by the attachment gallery)

I fixed this by both padding the ids in queries and sending the true number of records to the attachment gallery (aka not relying on the length of the array).
Either solution works and aren't needed together, but I left them both to ensure record sets and queries are consistent.

Should I leave in the padding? Large record sets pad their arrays already and it seems to work fine, but it still feels weird to have potentially massive arrays.
From my testing, it works okay, but if you have strong feelings one way or the other, let me know.

pashiav (commented):
- [ ] Verify the "Download All" button actually downloaded all the record set's attachments.
- [x] You should be able to scroll down and load more record set attachments.
- [x] Verify the "Download All" button prompts you to create a record set from the query if not all the attachments were loaded
- [x] You should be able to scroll down and load more query attachments.
- [x] Make sure "Download All" downloads all of the query's attachments.
---

**Bugs**

**1. Even with the increase, there is still a 504 Gateway Time-out when trying to download [this recordset](https://sp7demofish20250210-issue-6521.test.specifysystems.org/specify/view/collectionobject/42239/?recordsetid=21) with 2907 attachments.**

I understand this is a large number of attachments and may not be a common scenario, but I think this error could be handled more gracefully. @specify/ux-testing thoughts?

emenslin (changes requested):
- [x] Verify the "Download All" button actually downloaded all the record set's attachments.
- [x] You should be able to scroll down and load more record set attachments.
- [x] Verify the "Download All" button prompts you to create a record set from the query if not all the attachments were loaded
- [x] You should be able to scroll down and load more query attachments.
- [x] Make sure "Download All" downloads all of the query's attachments.

---

So everything does work, however downloading attachments takes forever even if there are not that many. I was testing on KUfish and most downloads seemed to take about 2 minutes but when checking on KUbirds a record set with about the same amount of attachments would be done in 10 seconds. I'm not sure if this is just a kufish problem or if it happens on more dbs but either way I think it should be looked into more.
Link to KUfish record set: https://kufish20250214-issue-6521.test.specifysystems.org/specify/view/collectionobject/35816/?recordsetid=214

Link to KUBirds record set: https://kubirds20240606-issue-6521.test.specifysystems.org/specify/view/collectionobject/129670/?recordsetid=231


Also I found a problem on ojsmnh where when I query for attachments and I open the attachment gallery it shows the wrong number of attachments. Not sure if this is on other dbs or not, it might have something to do with dowloading attachments as well but I'm not sure. 

https://github.com/user-attachments/assets/7bb6f777-d0a9-41c5-80f9-9e1aba601a63

alesan99:
> So everything does work, however downloading attachments takes forever even if there are not that many.

@emenslin Hm I have no idea what could be causing this but I optimized the backend asset server requests and queries to hopefully improve this.
Additionally, the zip files now get downloaded as they're being created, which should reduce download times greatly AND should allow for long downloads without experiencing a timeout error @pashiav 

The PR is ready to test again with all issues resolved 👍👍 (except for the one below vv )

> Also I found a problem on ojsmnh where when I query for attachments and I open the attachment gallery it shows the wrong number of attachments. Not sure if this is on other dbs or not, it might have something to do with dowloading attachments as well but I'm not sure.

I recreated this on main so it seems to be an existing issue. If a record appears more than once in the query results it will appear multiple times in "browse in forms", so its attachments will be multiplied when viewed in the gallery. While I could remove duplicate records from the gallery, I think this should be written up as a separate issue to also address records showing up multiple times in browse in forms.

https://github.com/user-attachments/assets/ce6c642f-3fb6-4ea0-96d2-d28a01da7a11

https://kubirds20240606-main.test.specifysystems.org/specify/query/76/

emenslin (approved):
- [x] Verify the "Download All" button actually downloaded all the record set's attachments.
- [x] You should be able to scroll down and load more record set attachments.
- [x] Verify the "Download All" button prompts you to create a record set from the query if not all the attachments were loaded
- [x] You should be able to scroll down and load more query attachments.
- [x] Make sure "Download All" downloads all of the query's attachments.

---

Looks good, the only issue I ran into was when trying to download a large amount of attachments (<1000) I would either get a failed to fetch error or if it successfully downloaded when I tried to open the zip file it would be invalid. I'm going to approve because it seemed to work consistently with up to 500 attachments (not sure what the limit is for when it stops working consistently) and I don't know if users are ever going to want to download more than that, I just wanted to make note of this behavior. 

Downloading attachments in kufish is still slow but faster than before (same record set is now done in about a minute and a half instead of two minutes)

CarolineDenis (approved):

alesan99:
I pushed a fix to correctly download files with paths in the name like "C:/path/to/file.jpg". I believe this could result in corrupted zip files sometimes.

All attachment names will now be sanitized to be filename-safe.

acwhite211 (approved):

Issue #6251, opened 2025-02-17
Batch Edit: Locality's `lat1text` and `lat2text` fields get updated internally and marks the record as updated


sharadsw:
From @combs-a:
> LatLon has been kind of fixed, but there are also potentially integers; ints are being converted to floats and that marks them as updated cells when they shouldn't be.

![Image](https://github.com/user-attachments/assets/311edd5c-12af-4789-bec7-0845fdb416eb)

The fields `lat1text` and `lat2text` are updated internally through the workbench which means Locality records can get marked as updated even when there are no changes visible to the user. This could potentially cause issues with matching records in batch edit when relationships are enabled in #6126 

More info here: https://github.com/specify/specify7/pull/5417#issuecomment-2646052721

CarolineDenis:
NOTES: 
- Workbench has the same behavior 
- Keep it as known behavior for now 
Relevant code is here:

[specify7/specifyweb/specify/parse.py](https://github.com/specify/specify7/blob/1b2ab816c6b8229e66ebb8469f3d9bfea3003e7f/specifyweb/specify/parse.py#L201-L216)

This was added for CoGe tool, we might break it if we fix this.

realVinayak:
^In the above solution, you could theoretically make an argument that doing match time, we'll not consider `lat1text` for to-one localities that already exist. I think that is fine - it is quite unlikely that `lat1text` field will be the only unique identifier. Same for `longitude`. In the worst case, we get disambiguation. Super easy for the user to fix (if lat1text does uniquely identify) - just map those during query process.

PR #6657, opened 2025-06-16
Disallow updating coordinate information in BatchEdit

### LINKED ISSUE #6655: Disallow mapping locality coordinate fields until officially supported
Currently, coordinate fields in BatchEdit behave exactly like Workbench: that is, BatchEdit expects fields mapped to the strictly decimal fields of `latitude1`, `longitude1`, `latitude2`, and `longitude2` from the Locality table to instead be representative of their corresponding text field counterpart (`lat1text`. `long1text`, `lat2text`, and `long2text` respectively). 
That is, both BatchEdit and WorkBench currently only match values of cells for the coordinate decimal fields to their text counterpart (for example, entering a value like `35.3457° N` for `latitude1` will actually only match Localities which have a `35.3457° N` `lat1text` ), and Specify will parse the value into a decimal and update the decimal field, `originalLatLongUnit`, and corresponding text field behind-the-scenes. 

This logic poses a problem with BatchEdit: when any of the decimal coordinate fields are mapped in the corresponding Query, their literal decimal values are returned. 
For example, consider the following example Data Set:

| Latitude1 | Longitude1 |
|-----------|------------|
| 39.04     | -95.28     |

With the first row, Specify will check for other Localities which have a `lat1text` of 39.04 and `long1text` of `-95.28` (which will result in new records being created as the actual `lat1text` and `long1text` of the record are `39.0400° N` and `95.2800° W` respectively). 
Specify will follow the same behavior as in the WorkBench and attempt to parse the `latitude1` and `longitude1` as `lat1text` and `long1text`, match, convert the values into decimals (which they already are), and update the `originalLatLongUnit` if needed. 
This behavior actually causes #6251.

A first-approach solution might be to treat the decimal coordinate fields as standalone -- that is, for BatchEdit change the matching behavior of the decimal fields to only match to the decimal fields, only allow decimal values with these fields, and do no further business logic (like updating the `originalLatLongUnit`). This was actually the first simple approach pushed to the repo, see https://github.com/specify/specify7/commit/30a3382a09f9b593d4d3f56d8149bc55385686c8, https://github.com/specify/specify7/commit/fbba7a7789247dc9666f102704df8433dc44ec33, and https://github.com/specify/specify7/commit/20dc725036dcb36ccf73a1a9ec231ee740d35a4f. 
There are some problems with this approach however. Primarily, there is no easy way to synchronize the decimal field with its corresponding text field (i.e., just given a decimal representation of a coordinate, synchronize the textual representation of the coordinate). 

For the upcoming `v7.11` release, it may be best to disable updating coordinate fields until 

I think an ideal solution from a user perspective might be to replace any decimal coordinate fields in the Query Results with their textual counterpart (e.g., replace just the result of `latitude1` with `lat1text` and maintain the `latitude1` mapping). 

To clarify, with the previous example of a Locality with `latitude1` of `39.04`, `longitude1` of `-95.28`, `lat1text` of `39.0400° N`, and `long1text` of `95.2800° W`, I think it would be beneficial to replace just the values of `latitude1` with the `lat1text` and `longitude1` with `long1text` in the Query Results (regardless of whether or not the text fields were explicitly mapped) and hide the text fields from the resulting BatchEdit Data Set if the text fields were mapped. Then the normal WorkBench behavior of coordinate information can take place.  

This way:  
- Stays consistent with user's expectations of working with coordinate information in a Data Set
  - Also, stays consistent with how Specify 6 handled BatchEdit  
- Coordinate information formatted in Decimal Degrees, Degree Minutes Seconds, Decimal Degrees, and/or with a degree symbol will be supported 
- All of the built-in coordinate tools (primarily coordinate conversion because GeoLocate and GeoMap support decimal coordinates) would work out-of-the-box with BatchEdit, and no changes to the WorkBench code will be needed

melton-jason:
Fixes #6655, #6664
Additionally, see #6251

This PR disallows BatchEdit of Queries containing any of the following fields:
- latitude1
- longitude1
- lat1text
- long1text
- latitude2
- longitude2
- lat2text
- long2text 

This change is a temporary one until better/official support for coordinate logic with BatchEdit is supported (slated for the next release: `v7.11.1`)

If any of these fields are mapped in the Query, a dialog similar to the one below will be shown to the user explaining the fields are not supported in BatchEdit currently: 

Screenshot 2025-06-16 at 11 05 18 AM

If the user creates a BatchEdit Data Set using the API via the `/stored_queries/batch_edit/` endpoint, then the fields will be read-only: 

Screenshot 2025-06-16 at 11 08 37 AM

⚠️ Note that this does not retroactively affect already existing BatchEdit Data Sets. If any of the now-disabled fields were mapped in Batch Edit Data Set in a prior version (like `v7.10.2.3`), then the user can still Validate, Commit, and/or RollBack the Data Set. 
@specify/testing, 
If required, we can raise an error to prevent the application (commit) of already existing Data Sets with these fields (to prevent creating duplicate Locality records-if Locality is not the base table- and/or overwriting the coordinate text fields. See #6251)
What do you guys think?

> 💡 @specify/dev-testing  currently, `oringinalLatLongUnit` and `srcLatLongUnit` are not included in the list of disallowed fields. These fields are mostly used by Specify 6 to determine what format the coordinate text fields are in (Decimal Degrees, DMS, etc.). From what I understand, these are not used by Specify 7 as Specify 7 infers the format of the text fields based on the structure of the value, so changing these via BatchEdit should not make a difference. 
>
> However, I am unsure about the impact regarding Specify 6 that changing either of the fields would have on the coordinate behavior and logic: I have not looked too in-depth regarding that aspect of the software
>
> Should we include these fields just to be safe? 

### Checklist

- [x] Add pr to documentation list
- [ ] Add automated tests

### Testing instructions
- In a Query with Locality as a the base table
- Include one or more of the disallowed fields in the Query and make sure the field(s) are marked as shown/included in the results
- [ ] Run the Query and attempt to BatchEdit the QueryResults and ensure a dialog appears preventing the BatchEdit explaining the fields are not allowed 
- Mark the disallowed fields as hidden in the results 
- [ ] Run the Query again and attempt to BatchEdit the QueryResults. Ensure that the dialog does not appear and the results can be edited

- In a  Query with Locality not as the base table
  - I would recommend using either CollectingEvent or CollectionObject via the paths CollectingEvent -> Locality, or CollectionObject -> CollectingEvent ->  Locality respectively, but any path to Locality to would do
- Include one or more of the disallowed fields in the Query and make sure the field(s) are marked as shown/included in the results
- [ ] Run the Query and attempt to BatchEdit the QueryResults and ensure a dialog appears preventing the BatchEdit explaining the fields are not allowed 
- Mark the disallowed fields as hidden in the results 
- [ ] Run the Query again and attempt to BatchEdit the QueryResults. Ensure that the dialog does not appear and the results can be edited

_Added after https://github.com/specify/specify7/pull/6657#pullrequestreview-2933093453_

- [ ] Ensure that any changes to the fields and filters in the Query (like checking/unchecking the `Show in results` button) show the `Query has unsaved changes dialog`

For convenience, here is the list of disallowed fields again: 
- latitude1
- longitude1
- lat1text
- long1text
- latitude2
- longitude2
- lat2text
- long2text

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/BatchEdit/index.tsx
+++ b/specifyweb/frontend/js_src/lib/components/BatchEdit/index.tsx
@@ -238,9 +238,11 @@ const containsSystemTables = (queryFieldSpec: QueryFieldSpec) =>
 const DISALLOWED_FIELDS: DeepPartial<{
   readonly [TABLE in keyof Tables]: RA<keyof Tables[TABLE]['fields']>;
 }> = {
-  // FEATURE: Remove these when lat/long is officially supported
-  // See https://github.com/specify/specify7/issues/6251 and
-  // https://github.com/specify/specify7/issues/6655
+  /*
+   * FEATURE: Remove these when lat/long is officially supported
+   * See https://github.com/specify/specify7/issues/6251 and
+   * https://github.com/specify/specify7/issues/6655
+   */
   Locality: [
     'latitude1',
     'longitude1',


alesan99 (approved):
- [x] Run the Query and attempt to BatchEdit the QueryResults and ensure a dialog appears preventing the BatchEdit explaining the fields are not allowed 
- [x] Run the Query again and attempt to BatchEdit the QueryResults. Ensure that the dialog does not appear and the results can be edited
- [x] Run the Query and attempt to BatchEdit the QueryResults and ensure a dialog appears preventing the BatchEdit explaining the fields are not allowed 
- [x] Run the Query again and attempt to BatchEdit the QueryResults. Ensure that the dialog does not appear and the results can be edited

---

Behavior looks correct! 👍

I did notice the user isn't forced to save the query once you make the dialog go away by hiding the locality fields:

https://github.com/user-attachments/assets/3c5f8811-c035-44f9-acd9-7c2d12b36587

This also happens on any saved query, so while its not caused by this PR it may be easier to encounter now so I'm making note of it. But I'll write this up as a separate issue https://github.com/specify/specify7/issues/6664 (unless you think its worth fixing here)

acwhite211 (approved):

PR #6710, opened 2025-06-18
add series entry button

alesan99:
Fixes #

### Checklist

- [ ] Add automated tests

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Forms/Save.tsx
@@ -242,10 +242,91 @@ export function SaveButton<SCHEMA extends AnySchema = AnySchema>({
     !tableValidForBulkClone(resource.specifyTable, resource) ||
     formatter === undefined;
 
+  const showSeriesEntry = true;
+  const [seriesEntryStart, setSeriesEntryStart] = React.useState<string | undefined>(undefined);
+  const [seriesEntryEnd, setSeriesEntryEnd] = React.useState<string | undefined>(undefined);
+
   return (
     <>
       {typeof handleAdd === 'function' && canCreate ? (
         <>
+            {resource.specifyTable.name === 'CollectionObject' &&
+            (isInRecordSet === false || isInRecordSet === undefined) &&
+            showSeriesEntry && canSave && !isCOGorCOJO
+            ? 
+              <>
+                <Input.Text
+                  aria-label={formsText.seriesEntryStart()}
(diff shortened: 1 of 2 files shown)


alesan99:
TODO:
- [ ] Add setting to enable series entry
- [ ] Customize series entry (Perhaps replace bulk carry?)
- [ ] Dialog for showing existing COs with catalog numbers
- [ ] Check if catalog nums are formatted correctly on frontend
- [ ] Check if catalog num range is correct on front end
- [ ] Option to disable record set creation

bhumikaguptaa (approved):
- [X] The Carry Forward range input should appear
- [X] Make sure you can only enter a valid start catalog number (It should use the catalog num of the CO you just saved)
- [X] Make sure you can only enter a valid end catalog number
- [X] Make sure you can't enter invalid ranges (where the start is bigger than the end for example)
- [X] Click Carry Forward and make sure all the COs in the range you specified were created
- [X] Click Carry Forward and see a dialog with a list of all existing catalog numbers.
- [X] Try Bulk Carry Forward and make sure no record set is created. But make sure all newly created records appear on the form page thing.

emenslin (changes requested):
- [x] The Carry Forward range input should appear
- [x] Make sure you can only enter a valid start catalog number (It should use the catalog num of the CO you just saved)
- [x] Make sure you can only enter a valid end catalog number
- [x] Make sure you can't enter invalid ranges (where the start is bigger than the end for example)
- [x] Click Carry Forward and make sure all the COs in the range you specified were created
- [x] Click Carry Forward and see a dialog with a list of all existing catalog numbers.
- [x] Try Bulk Carry Forward and make sure no record set is created. But make sure all newly created records appear on the form page thing.
- [x] Try creating a CO with a Collection Object Type with that uses a different formatter for the catalog number. Bulk Carry should still work, except if the formatter does not support autoincrementing (the button will be disabled).

--- 

I think if someone tries to carry forward over the range limit then it should just be blocked instead of throwing an error (something similar to the dialog when you try to carry forward over a number that already exists). With the current behavior the error also forces the user to go to the home screen which is not ideal. 

https://github.com/user-attachments/assets/215e097f-612d-48c8-8cfe-d5ad2b1d4eab

alesan99:
@emenslin I added a better dialog 👍 And if an error occurs for whatever reason you should be able to dismiss it now!
image


I did have to refactor parts of the code, so I could've broken some things. From my testing everything still seems to work, but you'll probably want to double check.

PR #6790, opened 2025-06-25
(tests): Add tests for fixMimeType

### LINKED ISSUE #6789: Unit Test: fixMimeType
https://github.com/specify/specify7/blob/1b11302c524d63cdb0d17d26c094640437fd98c3/specifyweb/frontend/js_src/lib/components/Attachments/attachments.ts#L282-L300

realVinayak:
Fixes #6789

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/Attachments/attachments.ts
+++ b/specifyweb/frontend/js_src/lib/components/Attachments/attachments.ts
@@ -297,7 +297,7 @@ export async function uploadFile(
  * See: https://github.com/specify/specify7/issues/1141
  * REFACTOR: remove this once that issue is fixed
  */
-function fixMimeType(originalMimeType: string): string {
+export function fixMimeType(originalMimeType: string): string {
   const maxLength = getField(tables.Attachment, 'mimeType').length;
   if (maxLength === undefined || originalMimeType.length < maxLength)
     return originalMimeType;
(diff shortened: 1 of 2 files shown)


specifyweb/frontend/js_src/lib/components/Attachments/attachments.ts
```diff
@@ -297,7 +297,7 @@ export async function uploadFile(
  * See: https://github.com/specify/specify7/issues/1141
  * REFACTOR: remove this once that issue is fixed
  */
-function fixMimeType(originalMimeType: string): string {
+export function fixMimeType(originalMimeType: string): string {
```
realVinayak, line 300: @CarolineDenis 

This function is no longer needed. The length of the mimeType field has been increased from 64 to 1024. The original purpose was to truncate long mimeTypes because there were cases where length was 71.

See https://github.com/specify/specify6/commit/2452bb29bf9e5ba20ef0d89b96f08504246f6291
where the length was increased in 6.

Below is from models.py, so the change is reflected in Specify7 too.

https://github.com/specify/specify7/blob/1b11302c524d63cdb0d17d26c094640437fd98c3/specifyweb/specify/models.py#L538

CarolineDenis: https://github.com/specify/specify7/issues/6795

PR #6831, opened 2025-06-30
CSRF Fix

### LINKED ISSUE #6802: Update `docker-compositions` to support `v7.11.0`+
Using the current `docker-compositions` deployment structure, we are receiving CSRF failed verification errors. This has been reported by users during the workshop at SPNHC and has been encountered by users at KU (test instance for herpetology initially had this issue)

https://github.com/specify/docker-compositions/pull/37

Before release, we need to make sure this is settled

grantfitzsimmons:
Fixes #6802 by solving it in this repository instead

>[!note]
>I am not the author of this PR, @acwhite211 pushed these changes. Credit goes to him!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -151,7 +151,7 @@ RUN cp -r specifyweb/settings .
 
 RUN echo \
         "import os" \
-        "\nfrom .specify_settings import *" \
+        "\nfrom . import specify_settings as specify_defaults" \
         "\nDATABASE_NAME = os.environ['DATABASE_NAME']" \
         "\nDATABASE_HOST = os.environ['DATABASE_HOST']" \
         "\nDATABASE_PORT = os.environ.get('DATABASE_PORT', '')" \
@@ -175,7 +175,7 @@ RUN echo \
         # - If still not defined, use the hard-coded default ['*']
         # See https://github.com/specify/specify7/pull/6831
         "\n_env_allowed_hosts = os.getenv('ALLOWED_HOSTS', None)" \
-        "\n_default_allowed_hosts = locals().get('ALLOWED_HOSTS', ['*'])" \
+        "\n_default_allowed_hosts = getattr(specify_defaults,'ALLOWED_HOSTS', ['*'])" \
         "\nALLOWED_HOSTS = _default_allowed_hosts if _env_allowed_hosts is None else _env_allowed_hosts.split(',')" \
         # Resolve CSRF_TRUSTED_ORIGINS in the following precedence:
         # - Use the CSRF_TRUSTED_ORIGINS environment variable (if present)
@@ -183,7 +183,7 @@ RUN echo \
         # - If still not defined, use the hard-coded default ['https://*', 'http://*']
         # See https://github.com/specify/specify7/pull/6831
         "\n_env_trusted_origins = os.getenv('CSRF_TRUSTED_ORIGINS', None)" \ 
-        "\n_default_trusted_origins = locals().get('CSRF_TRUSTED_ORIGINS', ['https://*', 'http://*'])" \
+        "\n_default_trusted_origins = getattr(specify_defaults,'CSRF_TRUSTED_ORIGINS', ['https://*', 'http://*'])" \
         "\nCSRF_TRUSTED_ORIGINS = _default_trusted_origins if _env_trusted_origins is None else _env_trusted_origins.split(',')" \
         > settings/local_specify_settings.py
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -169,7 +169,7 @@ RUN echo \
         "\nANONYMOUS_USER = os.getenv('ANONYMOUS_USER', None)" \
         "\nSPECIFY_CONFIG_DIR = os.environ.get('SPECIFY_CONFIG_DIR', '/opt/Specify/config')" \
         "\nhost = os.getenv('CSRF_TRUSTED_ORIGINS', None)" \
-        "\nCSRF_TRUSTED_ORIGINS = [origin.strip() for origin in host.split(',')] if host else []" \
+        "\nCSRF_TRUSTED_ORIGINS = [origin.strip() for origin in host.split(',')] if host else ['https://*', 'http://*']" \
         > settings/local_specify_settings.py
 
 RUN echo "import os \nDEBUG = os.getenv('SP7_DEBUG', '').lower() == 'true'\n" \
--- a/Dockerfile
+++ b/Dockerfile
@@ -169,8 +169,8 @@ RUN echo \
         "\nANONYMOUS_USER = os.getenv('ANONYMOUS_USER', None)" \
         "\nSPECIFY_CONFIG_DIR = os.environ.get('SPECIFY_CONFIG_DIR', '/opt/Specify/config')" \
         "\nhost = os.getenv('CSRF_TRUSTED_ORIGINS', None)" \
-        "\nALLOWED_HOSTS=os.getenv('ALLOWED_HOSTS', '').split(',') if os.getenv('ALLOWED_HOSTS') else ['*']" \
-        "\nCSRF_TRUSTED_ORIGINS = [origin.strip() for origin in host.split(',')] if host else ['https://*', 'http://*']" \
+        "\nALLOWED_HOSTS = globals()["ALLOWED_HOSTS"] if "ALLOWED_HOSTS" in globals() else (os.getenv("ALLOWED_HOSTS", "").split(",") if os.getenv("ALLOWED_HOSTS") else ["*"])" \
+        "\nCSRF_TRUSTED_ORIGINS = globals()["CSRF_TRUSTED_ORIGINS"] if "CSRF_TRUSTED_ORIGINS" in globals() else ([o.strip() for o in host.split(",")] if host else [])" \
         > settings/local_specify_settings.py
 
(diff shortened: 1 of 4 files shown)


acwhite211:
So there are a few different places to assign these variables:
- the settings/specify_settings.py file
- the settings/local_specify_settings.py file, which is generated and defined in the Dockerfile, and overrides settings made in the specify_settings.py file
- the docker-compose.yml file, which can use the .env file for local builds

I set the variable setting conditions in the local_specify_settings.py file to not override changes made to the default settings in the specify_settings.py file.  It also will not override any env defintions made in the docker-compose.yml file.  This way, we can support the setup of both local deployments and docker-composition deployments.

https://github.com/specify/specify7/blob/ea7b8d3776a007eecce56a9d5491b81327abfd49/Dockerfile#L172-L173

acwhite211 (approved):

Dockerfile
```diff
@@ -169,7 +169,8 @@ RUN echo \
         "\nANONYMOUS_USER = os.getenv('ANONYMOUS_USER', None)" \
         "\nSPECIFY_CONFIG_DIR = os.environ.get('SPECIFY_CONFIG_DIR', '/opt/Specify/config')" \
         "\nhost = os.getenv('CSRF_TRUSTED_ORIGINS', None)" \
-        "\nCSRF_TRUSTED_ORIGINS = [origin.strip() for origin in host.split(',')] if host else []" \
+        "\nALLOWED_HOSTS = globals()["ALLOWED_HOSTS"] if "ALLOWED_HOSTS" in globals() and globals()["ALLOWED_HOSTS"] != ["*"] else (os.getenv("ALLOWED_HOSTS", "").split(",") if os.getenv("ALLOWED_HOSTS") else ["*"])" \
```
melton-jason, line 172: If we ever update our defaults in [settings/specify_settings.py](https://github.com/specify/specify7/blob/main/specifyweb/settings/specify_settings.py), this change will make it so that the new default will be used over any user-provided `ALLOWED_HOSTS` (via environment variables) unless we also change this Dockerfile. 

For future-proofing and peace-of-mind, I would generally expect the order of precedence to always be: 
- Use user-defined variable if present
- Otherwise use default specified in [settings/specify_settings.py](https://github.com/specify/specify7/blob/main/specifyweb/settings/specify_settings.py)
- Fallback to the initial default for the variable (in this case for `ALLOWED_HOSTS`: `["*"]`)

So maybe something like the following? (written in Python, not ready for bash/echo): 

```py
env_allowed_hosts = os.getenv("ALLOWED_HOSTS", None)
default_allowed_hosts = globals().get("ALLOWED_HOSTS", ["*"])
ALLOWED_HOSTS = default_allowed_hosts if env_allowed_hosts is None else env_allowed_hosts.split(",") 
```

Or maybe something like: 

```py
_allowed_hosts = os.getenv("ALLOWED_HOSTS", globals().get("ALLOWED_HOSTS", ["*"]))
ALLOWED_HOSTS = _allowed_hosts.split(",") if isinstance(_allowed_hosts, str) else _allowed_hosts
```

Dockerfile
```diff
@@ -169,7 +169,8 @@ RUN echo \
...
+        "\nCSRF_TRUSTED_ORIGINS = globals()["CSRF_TRUSTED_ORIGINS"] if "CSRF_TRUSTED_ORIGINS" in globals() and globals()["CSRF_TRUSTED_ORIGINS"] != ["https://*", "http://*"] else ([o.strip() for o in host.split(",")] if host else [])" \
```
melton-jason, line 173: See the above comment (snippet provided below) regarding default behavior. The same applies for `CSRF_TRUSTED_ORIGINS`: 

> If we ever update our defaults in [settings/specify_settings.py](https://github.com/specify/specify7/blob/main/specifyweb/settings/specify_settings.py), this change will make it so that the new default will be used over any user-provided `ALLOWED_HOSTS` (via environment variables) unless we also change this Dockerfile. 

There is another issue specifically with `CSRF_TRUSTED_ORIGINS`: if the default is unchanged and the environment variable is not set, the value of `CSRF_TRUSTED_ORIGINS` will be `[]` rather than the intended default of `["https://*", "http://*"]`. 

```diff
- [o.strip() for o in host.split(",")] if host else []
+ [o.strip() for o in host.split(",")] if host else ["https://*", "http://*"]
```

A hypothetical implementation (in Python, not translated for echo in bash) might be similar to the one provided in the above comment for `ALLOWED_HOSTS`: 

```py
env_trusted_origins = os.getenv("CSRF_TRUSTED_ORIGINS", None)
default_trusted_origins = globals().get("CSRF_TRUSTED_ORIGINS", ["https://*", "http://*"])
ALLOWED_HOSTS = default_trusted_origins if env_trusted_origins is None else env_trusted_origins.split(",") 
```

melton-jason, line 173: I do get a python SyntaxError when building the containers: 

```
specify7-1         | SyntaxError: invalid syntax
specify7-1         | Traceback (most recent call last):
specify7-1         |   File "/opt/specify7/manage.py", line 25, in <module>
specify7-1         |     execute_from_command_line(sys.argv)
specify7-1         |   File "/opt/specify7/ve/lib/python3.9/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
specify7-1         |     utility.execute()
specify7-1         |   File "/opt/specify7/ve/lib/python3.9/site-packages/django/core/management/__init__.py", line 382, in execute
specify7-1         |     settings.INSTALLED_APPS
specify7-1         |   File "/opt/specify7/ve/lib/python3.9/site-packages/django/conf/__init__.py", line 102, in __getattr__
specify7-1         |     self._setup(name)
specify7-1         |   File "/opt/specify7/ve/lib/python3.9/site-packages/django/conf/__init__.py", line 89, in _setup
specify7-1         |     self._wrapped = Settings(settings_module)
specify7-1         |   File "/opt/specify7/ve/lib/python3.9/site-packages/django/conf/__init__.py", line 217, in __init__
specify7-1         |     mod = importlib.import_module(self.SETTINGS_MODULE)
specify7-1         |   File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module
specify7-1         |     return _bootstrap._gcd_import(name[level:], package, level)
specify7-1         |   File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
specify7-1         |   File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
specify7-1         |   File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
specify7-1         |   File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
specify7-1         |   File "<frozen importlib._bootstrap_external>", line 855, in exec_module
specify7-1         |   File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
specify7-1         |   File "/opt/specify7/settings/__init__.py", line 12, in <module>
specify7-1         |     from .local_specify_settings import *
specify7-1         |   File "/opt/specify7/settings/local_specify_settings.py", line 20
specify7-1         |     ALLOWED_HOSTS = globals()[ALLOWED_HOSTS] if ALLOWED_HOSTS in globals() and globals()[ALLOWED_HOSTS] != [*] else (os.getenv(ALLOWED_HOSTS, ).split(,) if os.getenv(ALLOWED_HOSTS) else [*]) 
```

Looks like the shell environment isn't of fan of the `"` characters when we're using `"` as the base string character: these would need to be properly escaped, so something like: 

```bash
        '\nALLOWED_HOSTS = globals()["ALLOWED_HOSTS"] if "ALLOWED_HOSTS" in globals() and globals()["ALLOWED_HOSTS"] != ["*"] else (os.getenv("ALLOWED_HOSTS", "").split(",") if os.getenv("ALLOWED_HOSTS") else ["*"])' \
        '\nCSRF_TRUSTED_ORIGINS = globals()["CSRF_TRUSTED_ORIGINS"] if "CSRF_TRUSTED_ORIGINS" in globals() and globals()["CSRF_TRUSTED_ORIGINS"] != ["https://*", "http://*"] else ([o.strip() for o in host.split(",")] if host else [])' \
```

or

```bash
        "\nALLOWED_HOSTS = globals()['ALLOWED_HOSTS'] if 'ALLOWED_HOSTS' in globals() and globals()['ALLOWED_HOSTS'] != ['*'] else (os.getenv('ALLOWED_HOSTS', '').split(',') if os.getenv('ALLOWED_HOSTS') else ['*'])" \
        "\nCSRF_TRUSTED_ORIGINS = globals()['CSRF_TRUSTED_ORIGINS'] if 'CSRF_TRUSTED_ORIGINS' in globals() and globals()['CSRF_TRUSTED_ORIGINS'] != ['https://*', 'http://*'] else ([o.strip() for o in host.split(',')] if host else [])" \
```

melton-jason (changes requested):
I love the idea behind using the defaults specified in [settings/specify_settings.py](https://github.com/specify/specify7/blob/main/specifyweb/settings/specify_settings.py) in the order of precedence when determining the Dockerized variables! 

https://github.com/specify/specify7/blob/2dbd08c3145923d46c3c93d171a741e8bd9d33c4/specifyweb/settings/__init__.py#L9-L14

Maybe we can incorporate the same idea for some of the other variables later.

And maybe to be more clear and less confusing for those not familiar with the Specify 7 build process, instead of `globals()` we can be explicit and import `.specify_settings` and use the defaults directly.

acwhite211 (approved):

PR #6842, opened 2025-07-01
Migrate `DisplayAuthor` to User Preferences

### LINKED ISSUE #6841: Move `TaxonTreeEditor.DisplayAuthor` to User Preferences

Currently, setting `TaxonTreeEditor.DisplayAuthor` is set in `RemotePreferences` for all users in all collections. It makes more sense to move the tree preferences to user preferences where they can be specific to a single user _or_ set in the [default preferences for the collection](https://discourse.specifysoftware.org/t/setting-default-user-preferences/1447).

grantfitzsimmons:
Fixes #6841 

Note that this **does not** migrate the preference from the `preferences` app resource, but it does move it to a more maintainable place and adds a UI for configuring it. It can be set as a default using the [default user preferences](https://discourse.specifysoftware.org/t/setting-default-user-preferences/1447) approach.

### Testing instructions

https://github.com/user-attachments/assets/d7029fdd-e2bb-4319-8d6a-f2da124a6469

- [ ] Check the preference "Show author in the tree" under "Tree Editor → Taxon" 

	image

- [ ] See that the author is shown after the taxon node in the tree
- [ ] Uncheck the preference
- [ ] See that the author is no longer shown after the taxon node in the tree

**Question:** Should this be enabled by default?

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Aside.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Aside.tsx
@@ -401,3 +401,7 @@ function ResourceItem({
     </ActiveLink>
   );
 }
+
+export const exportsForTests = {
+    mutateConformation
+}
--- a/specifyweb/frontend/js_src/lib/components/AppResources/__tests__/aside.test.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/__tests__/aside.test.tsx
@@ -1,0 +1,85 @@
+import type { RA } from '../../../utils/types';
+import type { AppResourcesConformation } from '../Aside';
+import { exportsForTests } from '../Aside';
+
+const { mutateConformation } = exportsForTests;
+
+describe('mutateConfirmation', () => {
+  test('child replaced', () => {
+    const conformations: RA<AppResourcesConformation> = [
+      {
+        children: [],
+        key: 'TestKey1',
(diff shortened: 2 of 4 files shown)


emenslin (approved):
- [x] Check the preference "Show author in the tree" under "Tree Editor → Taxon" 
- [x] See that the author is shown after the taxon node in the tree
- [x] Uncheck the preference
- [x] See that the author is no longer shown after the taxon node in the tree

--- 

I don't think it should be enabled by default since we don't have it enabled by default in remote preferences. My only suggestion is that we make it clear in documentation and the release announcement that users will have to enable it themselves.

bhumikaguptaa (approved):
- [X] Check the preference "Show author in the tree" under "Tree Editor → Taxon" 
- [X] See that the author is shown after the taxon node in the tree
- [X] Uncheck the preference
- [X] See that the author is no longer shown after the taxon node in the tree

Looks good!

lexiclevenger (approved):
- [x] Check the preference "Show author in the tree" under "Tree Editor → Taxon" 
- [x] See that the author is shown after the taxon node in the tree
- [x] Uncheck the preference
- [x] See that the author is no longer shown after the taxon node in the tree
---
Looks good! I also don't think it should be enabled by default considering many databases don't currently have the displayauthor preference enabled

PR #6888, opened 2025-07-02
Remove required for version in datamodel

### LINKED ISSUE #6835: Version is required in WorkBench uploads
**Describe the bug**
Introduced in https://github.com/specify/specify7/pull/6384, Version is a required field for all WorkBench uploads. 

Image
Image

**To Reproduce**

1. Create a Workbench Data Set, and select any table as the Base Table
2. See that Version is a required field
3. Click on any relationship
4. See that Version is required for all tables in the WorkBench

**Expected behavior**
A user should not have to map Version when uploading records through the WorkBench. 

- OS: macOS Sonoma 14.3 (M1 chip)
- Browser: Google Chrome 137.0.7151.120 (Official Build) (arm64)
- Specify 7 Version: `main` (https://github.com/specify/specify7/commit/2eb8f90cf0a3a7c1de7e26902c5c2ef360221092)

**Developer Notes**
We should either make Version not required by default in the [datamodel](https://github.com/specify/specify7/blob/main/specifyweb/specify/datamodel.py) (this would only have an effect on the application: it would affect the database behavior), or set a frontend [schemaOverride](https://github.com/specify/specify7/blob/main/specifyweb/frontend/js_src/lib/components/DataModel/schemaOverrides.ts)

CarolineDenis:
Fixes #6835

### Checklist

### Testing instructions

- Create a Workbench Data Set, and select any table as the Base Table
- [ ] Verify that Version is a not a required field

### DIFF
--- a/specifyweb/specify/datamodel.py
+++ b/specifyweb/specify/datamodel.py
@@ -8741,7 +8741,7 @@ datamodel = Datamodel(tables=[
             Field(name='remarks', column='Remarks', indexed=False, unique=False, required=False, type='text'),
             Field(name='timestampCreated', column='TimestampCreated', indexed=False, unique=False, required=True, type='java.sql.Timestamp'),
             Field(name='timestampModified', column='TimestampModified', indexed=False, unique=False, required=False, type='java.sql.Timestamp'),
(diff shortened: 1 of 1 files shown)


CarolineDenis:
@melton-jason https://github.com/specify/specify7/issues/6898

PR #6384, opened 2025-04-02
Update version=null to version=0 and make version required

CarolineDenis:
Fixes #2937

### Checklist

- [ ] Self-review the PR after opening it to make sure the changes look good and
      self-explanatory (or properly documented)
- [ ] Add relevant issue to release milestone

PR #6941, opened 2025-07-07
Add dependency to 0006 to avoid order issue

### LINKED ISSUE #6898: Fix migration issue
There is a side effect of #6384 which needs to also be resolved in `v7.11.1` (which can be done in a separate PR). 

There is a race condition between the two migrations added in #6384: [patches/0006_version_fix](https://github.com/specify/specify7/blob/main/specifyweb/patches/migrations/0006_version_fix.py) and [specify/0035_version_required](https://github.com/specify/specify7/blob/main/specifyweb/specify/migrations/0035_version_required.py). 

Both of the migrations will execute a `ALTER TABLE table MODIFY version ...` statement on their respective batch of tables, so the last migration which is run will be the resulting Version column. 

This is a problem when [patches/0006_version_fix](https://github.com/specify/specify7/blob/main/specifyweb/patches/migrations/0006_version_fix.py) is executed before [specify/0035_version_required](https://github.com/specify/specify7/blob/main/specifyweb/specify/migrations/0035_version_required.py). 
In the [specify/0035_version_required](https://github.com/specify/specify7/blob/main/specifyweb/specify/migrations/0035_version_required.py) migration, each version change will be equivalent to `ALTER TABLE <table> MODIFY Version INT NOT NULL`: 

https://github.com/specify/specify7/blob/6cf3a9b56a34580dcee94e60caa5ab210e818592/specifyweb/specify/migrations/0035_version_required.py#L53-L57

> 💡  Recall that setting the `default` attribute on fields does not modify the database behavior: https://stackoverflow.com/a/25713782. Django emulates the `default` behavior internally (similarly to [ForeignKey.on_delete](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.ForeignKey.on_delete)). 
> Django 5.0+ does have a way to update the database behavior for setting field defaults: [Field.db_default](https://docs.djangoproject.com/en/5.2/ref/models/fields/#db-default). Although for now, we need to manually use SQL to set the default. 

The result when the patches migration is applied before the specify migration is that `Version` is a required field for most tables with no default at the database level. 

Specifically, the Version field becomes: 
```
`Version` int(11) NOT NULL
```

CarolineDenis:
Fixes #6898 

### Checklist

- [ ] Add automated tests


melton-jason (approved):
This generally works, assuming we accept the risk users who have already ran the `patches/0006` migration would need to resolve the migration manually. 
The impact might be very high: this might affect Specify 6, Specify 7, as well as any tools that use the Specify 7 API (https://github.com/specify/specify7/pull/6888#pullrequestreview-2980298291). 
The subset of users this would affect would hopefully be small: internal and any who have pulled from `main` (or a branch derived of `main`) since https://github.com/specify/specify7/pull/6384 was merged (June 24, 2025). 

Ideally, I think this should be a separate migration altogether (which has dependencies on both `patches/0006` and `specify/0035`) so that even if someone did pull from `main` since https://github.com/specify/specify7/pull/6384, they would encounter no problems and don't have to manually resolve the version issue in the database. 

Given that the risk that a user has pulled `main`  may be sufficiently small (not included in a release, and within `main` slightly over 2 weeks), this approach is practically probably fine 🤞

PR #7032, opened 2025-07-11
uncomment apply_absolute_date

### LINKED ISSUE #7021: Querying (relative) timestampCreated for Collection Object returns all records
**Describe the bug**
When querying for collection objects based on the `timestampCreated` field in relative mode all records are returned, despite the timestamp created not falling within the designated range.

**To Reproduce**

1. Go to Query
2. Create a query with `catalogNumber` and `timestampCreated` in the CO table
3. Set the `timestampCreated` parameter to relative mode, then "Greater than or equal 1 day in the past"
4. See that all records are returned, observe that the `timeStampCreated` for records is outside of the window.
5. Set the query to some other value, such as 10 years in the past. Observe that the same number (the total number of CO in the collection) is returned.
6. As one final check, switch the `timestampCreated` filter to "any" and observe that an identical count is returned.

**Expected behavior**
Only the records that meet the criteria would be returned.

**Screenshots**
Query (sp7demofish, `v711-prerelease`)

Image

Results (sp7demofish, `v711-prerelease`)

Image

**Crash Report**

N/A

- OS: Windows 10
- Browser: Firefox 140.0.4 (64-bit)
- Specify 7 Version: `7.11-prerelease` and `7.11.0`
- Database Name: beaty and sp7demofish

**Reported By**
Beaty Biodiversity Museum

**Additional context**
I can replicate this on both sp7demofish (`v711-prerelease`) and beaty (`v7.11.0`). I initially noticed this bug because the statistics page showed an inaccurate count, shown below. **Possible 7.11.0 regression** as these statistics were behaving normally before.

Image

acwhite211:
Fixes #7021

Uncomment apply_absolute_date line in the execution.py file.  Looks like this line was commented out during development and needed to be added back it.  This is to fix the issue of querying for collection objects based on the timestampCreated field in relative mode all records are returned, despite the timestamp created not falling within the designated range.

### Checklist

- [ ] Add automated tests

### Testing instructions

- [ ] Create a specify query in the CO QB viewer.  Set the query to filter by the "Timestamp created" and set some condition on that filter.  See that the CO query results are limited and not returning all COs. 
- [ ] Go through testing instructions from: https://github.com/specify/specify7/pull/4952

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
@@ -234,19 +234,6 @@ function Wrapped({
       countOnly: mode === 'count',
     });
 
-    // Handle checking if catalogNumber is present for series query
-    const queryFields = getQueryFieldRecords?.(fields) ?? query.fields;
-    const canEnableSmushed =
-      queryFields.length === 0 ||
-      queryFields.some(
-        (field) => field.fieldName === 'catalogNumber' && field.isDisplay
-      );
-    queryResource.set('smushed', canEnableSmushed && query.smushed);
-    setQuery({
-      ...query,
-      smushed: canEnableSmushed && query.smushed,
-    });
-
     /*
      * Wait for new query to propagate before re running it
      * TEST: check if this still works after updating to React 18
@@ -323,6 +310,14 @@ function Wrapped({
       ),
     [state, table.name]
   );
+
+  React.useEffect(() => {
+    if (!showSeries)
+      setQuery({
+        ...query,
+        smushed: false,
+      });
+  }, [showSeries]);
 
   return treeRanksLoaded ? (
     <ReadOnlyContext.Provider value={isReadOnly}>
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
@@ -575,11 +575,17 @@ function Wrapped({
                 showSeries={showSeries}
                 tableName={table.name}
                 onRunCountOnly={(): void => runQuery('count')}
-                onSubmitClick={(): void =>
+                onSubmitClick={(): void => {
+                  const canEnableSmushed = query.fields.some(
+                    (field) => field.fieldName === "catalogNumber" && field.isDisplay
+                  );
+                
+                  queryResource.attributes.smushed = canEnableSmushed ? !(query.smushed ?? false) : false;
+
                   form?.checkValidity() === false
                     ? runQuery('regular')
                     : undefined
-                }
+                }}
                 onToggleDistinct={(): void =>
                   setQuery({
                     ...query,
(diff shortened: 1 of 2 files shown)


CarolineDenis (approved):

emenslin (changes requested):
- [x] Create a specify query in the CO QB viewer.  Set the query to filter by the "Timestamp created" and set some condition on that filter.  See that the CO query results are limited and not returning all COs. 
- [ ] Go through testing instructions from: https://github.com/specify/specify7/pull/4952

--- 

Running into a similar issue as https://github.com/specify/specify7/pull/4952#pullrequestreview-2651780490, if you check series, delete catalog number, then query series is still checked and you cannot check distinct and the count is incorrect.

https://github.com/user-attachments/assets/adea71c5-c382-4c12-8c76-02e8d2384fba



Also I was able to replicate this issue https://github.com/specify/specify7/pull/4952#pullrequestreview-2808169900 where if you check series then disable catalog number it throws an error. 

https://github.com/user-attachments/assets/f6c2cd79-5404-4415-aae9-44264cded5a2


[Specify 7 Crash Report - 2025-07-11T18_46_43.554Z.txt](https://github.com/user-attachments/files/21188757/Specify.7.Crash.Report.-.2025-07-11T18_46_43.554Z.txt)

Other than those two issues everything else looks good

acwhite211:
The first issue of removing catalog number and rerunning the query seems to be a front-end issue where the front-end is still putting `smushed=True` in the request.  I remember that we solved this issue, so it's odd that it is happening again, I'm investigating the issue.  The second issue of unselecting the catalog number is also causing the same issue.

acwhite211:
Thanks @emenslin, made some changes to the logic to fix that issue.  Looks like it's behaving correctly now.  Should be good now to retest, thanks.

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -575,28 +575,30 @@ function Wrapped({
                 showSeries={showSeries}
                 tableName={table.name}
                 onRunCountOnly={(): void => runQuery('count')}
-                onSubmitClick={(): void =>
+                onSubmitClick={(): void => {
+                  const canEnableSmushed = query.fields.some(
+                    (field) => field.fieldName === "catalogNumber" && field.isDisplay
```
CarolineDenis: this wasn't needed in https://github.com/specify/specify7/pull/4952

specifyweb/frontend/js_src/lib/components/QueryBuilder/Wrapped.tsx
```diff
@@ -233,6 +233,18 @@ function Wrapped({
       fields: getQueryFieldRecords?.(fields) ?? query.fields,
       countOnly: mode === 'count',
     });
+
+    // Handle checking if catalogNumber is present for series query
+    const queryFields = getQueryFieldRecords?.(fields) ?? query.fields;
+    const canEnableSmushed = queryFields.length === 0 || queryFields.some(
```
CarolineDenis, line 239: we have a show series variable below

acwhite211:
Thanks for noticing that @lexiclevenger! Looks like this is an issue in main and other older version of Specify.  I created a new issue https://github.com/specify/specify7/issues/7088, so we'll ignore that issue for this PR.  Thanks!

PR #7087, opened 2025-07-15
Batch Edit: Create new record for independent relationship if all visible fields are null.

### LINKED ISSUE #6569: New record is not created when independent relationship fields are cleared in batch edit
**Describe the bug**
When an independent relationship field is cleared in batch edit, the original value stays the same and no new record is created when changes are committed.

**To Reproduce**

1. Create a batch edit data set with independent relationship fields
2. Clear the values for those fields
3. Validate and commit changes
4. View the edited records and see that the original values have not been changed and no new record has been created

**Expected behavior**
A new record should be created with the new null values.

**Screenshots**
Original data: 

Image

Changes committed:

Image

Data after changes:

Image

- OS: Mac
- Browser: Chrome
- Specify 7 Version: 7.11
- Database Name: KUFish_2024_07_11_2024_08_15_1_
- Collection name: KU Fish Voucher
- User Name: spfishadmin
- URL: https://kufish20240711202408151-main.test.specifysystems.org/specify/workbench/66

alesan99:
Fixes #6569

New behavior:
- If "use only visible fields for null check" is disabled:
  - A new record should be created, with the new null fields.
- If "use only visible fields for null check" is enabled:
  - The related record should be removed from the base record.

### Checklist

- [ ] Add automated tests

### Testing instructions

- Create a query that contains an independent relationship. For example, Collection object -> Collecting event -> Locality.
- Click batch edit.
- Erase all visible fields for the independent relationship

image

- Disable "use only visible fields for null check"
- [ ] Upload and make sure a new record was created, or at least it attempted to create a new record (aka validation error).
- Enable "use only visible fields for null check"
- [ ] Upload and make sure the independent relationship was removed from the base record. Verify with a query.
- [ ] Do some general batch edit testing.
  - **Very important**, this PR potentially changes existing behavior when you leave fields blank. Make sure it still behaves as expected.
  - I recommend going through the testing checklist for Batch Edit https://docs.google.com/spreadsheets/d/1MR18fpxh-_q1yuc5xbsrPAo7veAkZGMSRsW1E4AqoWI/edit?gid=1133393897#gid=1133393897&range=A624:Q646

### DIFF
--- a/specifyweb/workbench/upload/upload_table.py
+++ b/specifyweb/workbench/upload/upload_table.py
@@ -603,6 +603,23 @@ class BoundUploadTable(NamedTuple):
             # nothing to upload
             return UploadResult(NullRecord(info), to_one_results, {})
         if not skip_match:
+            assert not is_edit_table, "Trying to match update table!"
+
+            if not filter_predicate.filters and self.current_id is not None:
+                # Technically, we'll get always the empty predicate back if self.current_id is None
+                # So, we can skip the check for "self.current_id is not None:". But, it 
+                # is an optimization (a micro-one)
+                filter_predicate = self.get_django_predicates(
+                    should_defer_match=self._should_defer_match,
+                    # to_one_results should be completely empty (or all nulls)
+                    # Having it here is an optimization.
+                    to_one_override=to_one_results,
+                    consider_dependents=False,
+                    # Don't necessarily reduce the empty fields now.
+                    is_origin=False,
+                    origin_is_editable=False
+                )
+
             match = self._match(filter_predicate, info)
             if match:
                 return UploadResult(match, to_one_results, {})


alesan99:
> So one thing I noticed is that the record is deleted, however, when trying to delete a locality from a CO record it also removes the CE from the CO, is that expected? It seems like the CE still exists in the DB it is just removed from the CO. Just commenting this as I am not sure if this is an actual issue or if I am misunderstanding the preferences.

@emenslin Good catch, that's not expected here. It looks like my fix was unlinking unrelated records somehow.

@realVinayak provided a fix that should comprehensively and correctly match and create records.
However, certain relationships still may not be getting cleared, I'm looking into it.

CarolineDenis (approved):

pashiav (changes requested):
- [ ] Upload and make sure a new record was matched or created.
- [ ] Upload and make sure the independent relationship was removed from the base record. Verify with a query.
 - [x] Try with CO -> CE -> Locality like in the issue. If it is working correctly, you should get a validation error if you set localityName to blank (with visible fields setting enabled) or the Locality should be successfully removed (with visible fields setting disabled). 
 - [ ] Try setting all fields blank on any to-many or one-to-one relationship.  Make sure it still behaves as expected.
---

<ins>**Bugs**</ins>

**1. Deleting accession from CO in batch edit causes error upon validation.**
Removing the accession number in BE causes a `Field is required by schema config` error. This does not occur on `main`. 

_**Note:** Commiting on `main` doesn't actually change anything in the record ([accession is not actually deleted](https://calvertmarinemuseumgen20250721-main.test.specifysystems.org/specify/view/collectionobject/3061/?recordsetid=167)), even though there is no error._

image


**Issue branch**: https://calvertmarinemuseumgen20250721-issue-6569.test.specifysystems.org/specify/workbench/36/
**Main**: https://calvertmarinemuseumgen20250721-main.test.specifysystems.org/specify/workbench/38

I tried a few combinations of preferences to see if I could get it to validate without errors, but still recieved the same result.

alesan99:
> **1. Deleting accession from CO in batch edit causes error upon validation.** Removing the accession number in BE causes a `Field is required by schema config` error. This does not occur on `main`.

@pashiav 
I tried this process and the behavior seems correct.
Enabling "Use visible fields for empty" and erasing accession number successfully removed the record. It doesn't appear BE acknowledge the change, and the record set created at the end uses the wrong records, but I think those were existing issues.
With the default settings the validation error seems correct.
https://kubirds20240606-issue-6569.test.specifysystems.org/specify/query/93/

I'll re-request your review but lmk if anything else pops up.

acwhite211 (approved):

PR #7093, opened 2025-07-15
Show field caption in schema config field list

mark-pitblado:
Implements the feature request in [this forum post](https://discourse.specifysoftware.org/t/have-the-caption-display-beside-the-database-name-in-the-schema-config/2701). I can create a corresponding feature issue on GitHub if this is something the team wants (and if this feature is accepted overall).

Shows the field caption beside the database name in the schema config field list. To help the user distinguish between the database field name, the caption, and the hidden indicator, the list uses the following formatting.

**databaseFieldName** *caption* (hidden)

The caption in the field name list will change in real-time as it is entered in the caption input box, so the user learns the link between the two. Below is an example of how this feature is helpful when using an attribute or properties table.

image

### Checklist

- [ ] Add automated tests

### Testing instructions

- [ ] Open the schema config tool and see that there are now captions displayed beside the field name.
- [ ] Change the caption for a field, and verify that the caption updates in real-time for the corresponding field in the field list.

All frontend tests have passed. I am not super familiar with TypeScript, so while functional, perhaps the typing could be further improved. Tested with Firefox 128.12.0 ESR


grantfitzsimmons (changes requested):
I love the idea behind this change– it would really make customizing the schema _much_ easier!

---

it looks good in Firefox, but in Chrome, the HTML styling inside `<option>` elements is not consistently supported.

**Firefox:** ✅ 
Image

**Chrome:** ❌ 
Image

---

I'll let the @specify/developers  offer guidance on how to implement this in a way that ensures consistent support across all browsers. Thanks for openign a PR to address this!

PR #7097, opened 2025-07-16
Revert get_collection to first collection name

### LINKED ISSUE #7061: Improve default collection detection for web asset server URLs
### Describe the bug

A recent change to Specify 7 (likely [#6176](https://github.com/specify/specify7/pull/6176)) altered the default behavior for how the collection is determined when generating web asset server URLs. The system now appears to default to the database name (`DATABASE_NAME`) instead of the collection's logical name.

For on-premise, single-collection instances where the database name does not match the collection name expected by the asset server, this change breaks all attachment URLs after an update. A user at the California Academy of Sciences experienced this when their URLs began using `coll=casbotany` (the database name) instead of the expected `coll=Botany`.

The current workaround is to explicitly set the `WEB_ATTACHMENT_COLLECTION` environment variable in the `docker-compose.yml` file. However, this default behavior is not intuitive for single-collection sites and can be difficult to diagnose.

### To Reproduce

1.  Configure a single-collection Specify 7 instance using Docker.
2.  In `docker-compose.yml`, set `DATABASE_NAME` to a value different from the collection's logical name (e.g., `DATABASE_NAME=casbotany`).
3.  Configure the `ASSET_SERVER_URL` to point to a valid web asset server that expects a different collection name (e.g., `Botany`).
4.  Ensure the `WEB_ATTACHMENT_COLLECTION` environment variable is **not** set.
5.  Log in to the Specify 7 instance.
6.  Navigate to a record with an attachment and attempt to view it.
7.  Observe that the generated URL for the attachment incorrectly uses the database name in the `coll=` parameter, causing the request to the asset server to fail.

### Expected behavior

In a single-collection instance, the system should intelligently default to using the collection's logical name (from `collection.CollectionName`) for the `coll=` parameter in asset server URLs. This would provide a more seamless update experience for the majority of on-premise institutions and would not require manual configuration of an extra environment variable.

### Reported By

Joe Russack at California Academy of Sciences

### Additional context

This issue represents a frustrating regression for single-collection users. The proposed solution is to refine the default logic: if an instance contains only one collection, that collection's name should be used as the default for `WEB_ATTACHMENT_COLLECTION`, overriding the current default behavior of using the database name.

acwhite211:
Fixes #7061

Revert get_collection to first collection name

### Checklist

- [ ] Add automated tests

### Testing instructions

- [ ] Check that the assets load when using each collection in the database.

### DIFF
--- a/specifyweb/attachment_gw/views.py
+++ b/specifyweb/attachment_gw/views.py
@@ -72,11 +72,8 @@ def get_collection(request=None):
     # do any better than using the first collection
     # and hoping that all the assets are in the same
     # folder.
-    # from specifyweb.specify.models import Collection
-    # return Collection.objects.all()[0].collectionname
-
-    # The default coll parameter to assets is the database name
-    return settings.DATABASE_NAME
+    from specifyweb.specify.models import Collection
+    return Collection.objects.all()[0].collectionname
 
 @openapi(schema={
     "get": {


grantfitzsimmons:
I let Joe know our plan to revert it, and he said the following: 
>I think it’s a good change, FWIW - I’d retain it, just revert to the old behavior if the variable remains unset. 
>And update your docker image example; I don’t >think< this change was reflected there? 

I agree on both points, @acwhite211 does it already use the environment variable if it is set in place of this logic?

grantfitzsimmons (approved):
- [x] Check that the assets load when using each collection in the database.

CarolineDenis (approved):

PR #7121, opened 2025-07-16
Switch to thumbnail for tiff images in attachment previews

### LINKED ISSUE #5356: TIF/TIFF files do not display a preview
**Describe the bug**
After uploading a TIF or TIFF file to Specify, a preview cannot be seen when viewed in the attachment window. Thumbnails are generated as expected, but these are not visible on the form.

**To Reproduce**

1. Open Specify
2. Upload a `TIF` or `TIFF` file (e.g. [leaves.tif.zip](https://github.com/user-attachments/files/17589547/leaves.tif.zip))
3. See that there is no preview 

**Expected behavior**
This should show a preview of the TIF file or at least use the thumbnail that was generated for it.

image

image

**System Report**
macOS Sequoia 15.0
Arc Browser, Google Chrome, Firefox
Tested on `sp7demofish` instance running `v7.9.6.2`

[Specify 7 System Information - 2024-10-31T15_18_33.231Z.txt](https://github.com/user-attachments/files/17589564/Specify.7.System.Information.-.2024-10-31T15_18_33.231Z.txt)

mark-pitblado:
Fixes #5356

Chrome, edge, and firefox [do not render tiffs](https://en.wikipedia.org/wiki/Comparison_of_web_browsers). Therefore, the simplest (only?) solution is to fall back to the thumbnail if there is a tiff image. The downside of this is that safari users that set the preference for original images will now see the thumbnail displayed, and users may be confused as to why they selected "show original" in the preferences, but an original is not being shown. Some library or additional dependency could probably be added to render the image, but that seems like an increase in complexity that is not justified by the increase in functionality.

This PR will also handle tiff images displaying as thumbnails in the attachments page and in the query results viewer.

image

image

image

### Checklist

- [ ] Add automated tests

### Testing instructions

- [ ] Ensure that `Attachment Preview Mode` is set to `Full Resolution` in preferences.
- [ ] Open a collection object with a tiff attachment. Verify that a preview (the thumbnail) of the image is now shown.

Note that https://github.com/specify/web-asset-server/issues/33 may still prevent some tiff images from showing properly, that error is separate and out of scope for this issue.


grantfitzsimmons (approved):
> Chrome, edge, and firefox [do not render tiffs](https://en.wikipedia.org/wiki/Comparison_of_web_browsers). Therefore, the simplest (only?) solution is to fall back to the thumbnail if there is a tiff image. 

I guess we should all just migrate to Safari... or GNOME Web!

>The downside of this is that safari users that set the preference for original images will now see the thumbnail displayed, and users may be confused as to why they selected "show original" in the preferences, but an original is not being shown. 

We do not officially support Safari due to a number of known inconsistencies, so I feel this is a reasonable behavior to enact for now.

>Some library or additional dependency could probably be added to render the image, but that seems like an increase in complexity that is not justified by the increase in functionality.

If we determine that implementing a better image viewing system is necessary, as @alesan99 is currently considering for WorkBench image previews, we can incorporate this into that project.

### Testing instructions

- [X] Ensure that `Attachment Preview Mode` is set to `Full Resolution` in preferences.
- [X] Open a collection object with a tiff attachment. Verify that a preview (the thumbnail) of the image is now shown.

😄 

image
image

Previous behavior (`v7.11.0.2`):
image

pashiav (approved):
- [x] Ensure that `Attachment Preview Mode` is set to `Full Resolution` in preferences.
- [x] Open a collection object with a tiff attachment
---
Looks good - TIFF/TIF files display properly when viewed!

Screenshot 2025-08-05 at 9 38 19 AM

PR #7122, opened 2025-07-16
Fetch notifications using the current server time

### LINKED ISSUE #4566: Exported CSV files from queries do not always notify the user
**Describe the bug**

If you click `Create CSV` for all results in a query from a tree node in production the notification never comes and you cannot download the file. If you select some rows and `Create CSV`, this is downloaded through the browser and the behavior is working as expected. 

CSVs in notifications do work normally for other saved and unsaved queries. 

**To Reproduce**

1. Go to tree
2. Select node
3. Click search icon in header to query
4. Click `Create CSV` for all results
5. Never get notification

**Screenshots**

https://github.com/specify/specify7/assets/135047322/52a719c7-bedd-4ece-96fb-23c1a0435bea

- OS: macOS
- Browser: Chrome
- Specify 7 Version: `v7.9.3.1`
- Database Name: [e.g. kufish] (you can see this in the "About Specify 7" dialog)

**Reported By**

CSIRO

**Additional context**
This related issue was also reported by CSIRO, but I was not able to recreate it. 

> When performing a query and clicking the "Create CSV"  on the results page (without selecting any items). The Notifications menu is not updating to advise that the download is ready until the browser page is refreshed or you logout and log back into the instance. 
Steps:
Perform a query and on the results page click the "Create CSV" button without selecting any items from the results page
For example my CSIRO colleague 
ran the test in demofish, (in KUFishvoucher collection) "I run the saved query 'type specimen'. I get 215 results. I hit 'Create CSV' (without selecting any items) wait a while for a notification, doesn't appear. Refresh the page, notification appears"
Expected:
System displays notification for download is ready
Actual:
System is not displaying notification for downloads is ready until page is refresh manually or logout/login occurs.

grantfitzsimmons:
Fixes #4566

This change fixes an issue where users in timezones with an offset larger than one set for Django (by default, `America/Chicago`) would not receive new notifications due to the fetching logic using the date of the user's browser locale and not the server time (which is used when creating new `notification_message` records), resulting in them not being fetched.

This PR adds a new API endpoint to fetch the current server time and integrates it into the notifications polling mechanism to ensure consistent time-based queries.

* Adds Django view and URL pattern to expose `/context/server_time.json`
* Introduces `fetchServerTime` utility to retrieve and parse server time
* Updates `useNotificationsFetch` to use server time in its polling URL

---

Since I see you requested a self-review, @melton-jason: please feel free to edit this PR however you see fit!

---

### Checklist

- [X] Add automated tests

### Testing instructions

Test the behavior in this PR against the latest tagged release:

- [ ] Use a browser extension to set your timezone to be something in the future relative to where you are (using a browser extension like https://change-timezone.pdfwork.com/ (for testing, I used the `Pacific/Apia` locale which has an offset of `780` , which is the time zone difference from UTC (in minutes) (with daylight consideration))
- [ ] 	Open the query builder and run a query. After it has been run, click "Export to CSV"
    - [ ] In `v7` with a time zone set far in the future, you **should not** receive any notification even after Specify fetches new notifications. When you reload the page, the new notification would be visible then.
    - [ ] In `issue-4566-time`, you **should** receive a notification once the browser sends a request to fetch notifications. You should be able to either wait for a notification to appear or click on the "Notifications" navigation menu item.


lexiclevenger (approved):
- [x] Use a browser extension to set your timezone to be something in the future relative to where you are (using a browser extension like https://change-timezone.pdfwork.com/ (for testing, I used the `Pacific/Apia` locale which has an offset of `780` , which is the time zone difference from UTC (in minutes) (with daylight consideration))
- [x] 	Open the query builder and run a query. After it has been run, click "Export to CSV"
    - [x] In `v7` with a time zone set far in the future, you **should not** receive any notification even after Specify fetches new notifications. When you reload the page, the new notification would be visible then.
    - [x] In `issue-4566-time`, you **should** receive a notification once the browser sends a request to fetch notifications. You should be able to either wait for a notification to appear or click on the "Notifications" navigation menu item.
---
Looks good! I received the notification without needing to refresh.

PR #7155, opened 2025-07-21
(tests): Add tests for getCoordinateColumns

### LINKED ISSUE #7154: Unit Test: getCoordinateColumns
https://github.com/specify/specify7/blob/b951e91b3bb2d9fd022298e8211d79c18a2fd7e6/specifyweb/stored_queries/execution.py#L434-L460

realVinayak:
Fixes #7154

### DIFF
--- a/specifyweb/stored_queries/execution.py
+++ b/specifyweb/stored_queries/execution.py
@@ -443,6 +443,10 @@ def getCoordinateColumns(field_specs, hasId):
     for fld in field_specs:
         if fld.fieldspec.table.name == "Locality":
             jp = fld.fieldspec.join_path
+            # BUG: In the cases where the basetable's formatted is added to the query,
+            # then the below branch will incorrectly skip incrementing the index.
+            # To fix this, simply do all the below logic in `if jp` (rather than continuing)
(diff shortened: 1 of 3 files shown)


specifyweb/stored_queries/execution.py
```diff
@@ -443,6 +443,10 @@ def getCoordinateColumns(field_specs, hasId):
     for fld in field_specs:
         if fld.fieldspec.table.name == "Locality":
             jp = fld.fieldspec.join_path
+            # BUG: In the cases where the basetable's formatted is added to the query,
+            # then the below branch will incorrectly skip incrementing the index.
+            # To fix this, simply do all the below logic in `if jp` (rather than continuing)
+            # See test: specifyweb.stored_queries.tests.test_execution.TestGetCoordinateColumns.test_formatted_fields)
```
realVinayak, line 449: @CarolineDenis this is a bug in the current code. Here, joinpath can be empty in cases where you have something like "locality (formatted)". In this case, the offsets will be incorrectly computed.

CarolineDenis: https://github.com/specify/specify7/issues/7169

PR #7201, opened 2025-07-24
(tests): getFormatterDef

### LINKED ISSUE #7198: Unit Test: getFormatterDef
https://github.com/specify/specify7/blob/82e5538fd7c91e558493aa813237f435a1db5cf1/specifyweb/stored_queries/format.py#L55-L110

realVinayak:
Fixes #7198


specifyweb/stored_queries/format.py
```diff
@@ -53,6 +53,9 @@ def __init__(self, collection, user, replace_nulls, format_agent_type=False, for
         self.format_picklist = format_picklist
 
     def getFormatterDef(self, specify_model: Table, formatter_name) -> Element | None:
+
+        # These functions don't belong here.
+        # REFACTOR: define them outside. This code design is very bad for test writing (and code quality)
```
realVinayak, line 58: For future reference, such design should be avoided (nesting functions, which can be easily defined outside). It makes adding tests harder, reduce readability.

For example, without deeper analysis, I cannot guarantee that the functions are independent (they may be doing some mutation). In this case, they are completely independent. Plus, adding tests are harder, because proper arguments needed to be reverse engineered.

CarolineDenis (approved):

PR #7337, opened 2025-08-20
[reorganization]: Move geologic time logic to its own app

### LINKED ISSUE #7336: Backend Reorg: Extract geology time logic from specify/ to backend/geology_time/

CarolineDenis:
Fixes #7336

### Checklist

### Testing instructions

- test extended age queries


acwhite211:
Usually, if you make a new Django app, that means you will have a set of distinct API calls, so you would have at least a `urls.py` file and a `views.py` file.  So, having just a `utils.py` file in a Django app is an odd choice.  If you want a separate directory, it would make the most sense to put it under the `stored_queries` app.  It seems you didn't add the `specifyweb.backend.geology_time` in the django settings, so it's not technically a django app, but then if we do that it would add some confusion as to which directory in the `backend` path is/isn't a django app without needing to look at the settings

emenslin (approved):
- [x] test extended age queries 

Looks good, everything seems to be the same as main

PR #7379, opened 2025-09-02
Support Locality Coordinates in Batch Edit

### LINKED ISSUE #6251: Batch Edit: Locality's `lat1text` and `lat2text` fields get updated internally and marks the record as updated

melton-jason:
Fixes #6251

> Currently, coordinate fields in BatchEdit behave exactly like Workbench: that is, BatchEdit expects fields mapped to the strictly decimal fields of `latitude1`, `longitude1`, `latitude2`, and `longitude2` from the Locality table to instead be representative of their corresponding text field counterpart (`lat1text`. `long1text`, `lat2text`, and `long2text` respectively). 
> That is, both BatchEdit and WorkBench currently only match values of cells for the coordinate decimal fields to their text counterpart (for example, entering a value like `35.3457° N` for `latitude1` will actually only match Localities which have a `35.3457° N` `lat1text` ), and Specify will parse the value into a decimal and update the decimal field, `originalLatLongUnit`, and corresponding text field behind-the-scenes

See https://github.com/specify/specify7/issues/6655#issue-3150411473

This PR re-enables support for editing Locality coordinates (Locality -> latitude1, Locality -> longitude1, Locality -> latitude2, Locality -> longitude1) within Batch Edit. 

Specifically, this implements the proposed implementation in https://github.com/specify/specify7/issues/6655#issue-3150411473: 

> I think an ideal solution from a user perspective might be to replace any decimal coordinate fields in the Query Results with their textual counterpart (e.g., replace just the result of `latitude1` with `lat1text` and maintain the `latitude1` mapping). 
> 
> To clarify, with the previous example of a Locality with `latitude1` of `39.04`, `longitude1` of `-95.28`, `lat1text` of `39.0400° N`, and `long1text` of `95.2800° W`, I think it would be beneficial to replace just the values of `latitude1` with the `lat1text` and `longitude1` with `long1text` in the Query Results (regardless of whether or not the text fields were explicitly mapped) and hide the text fields from the resulting BatchEdit Data Set if the text fields were mapped. Then the normal WorkBench behavior of coordinate information can take place.  
> 
> This way:  
> - Stays consistent with user's expectations of working with coordinate information in a Data Set
>   - Also, stays consistent with how Specify 6 handled BatchEdit  
> - Coordinate information formatted in Decimal Degrees, Degree Minutes Seconds, Decimal Degrees, and/or with a degree symbol will be supported 
> - All of the built-in coordinate tools (primarily coordinate conversion because GeoLocate and GeoMap support decimal coordinates) would work out-of-the-box with BatchEdit, and no changes to the WorkBench code will be needed

### Checklist

- [ ] Add automated tests


CarolineDenis (approved):

pashiav (approved):
- [x] Ensure that (save for the coordinate decimal fields) the resulting Batch Edit Data Set matches the Query Results (sorting and filters are respected)
  - [x] Ensure that the coordinate decimals fields are replaced with the record's corresponding text field in the Data Set (e.g., `Latitude 1 -> Lat1text`, `Longitude 1 -> Long1text1`, etc.) 
  - [x] Ensure the usual WorkBench spatial tools operate as expected (available if the Data Set includes any of the decimal coordinate fields and after clicking the `Tools` button. Current Spatial Tools are: Convert Coordinates, GeoLocate, GeoMap, etc.)
  - [x] Ensure matching behavior operates as expected with the coordinate decimal fields: Locality records can be matched  to existing Locality records and Locality records can be properly Disambiguated when there are more than one Locality records that match
  - [x] Commit the Data Set and ensure the record's corresponding coordinate decimal field(s) are updated to their expected decimal values
  - [x] Batch Edit the query and ensure the coordinate decimal field is present with the value of the coordinate text field
---

Looks good - working as expected! Didn't run into any unusual behavior.
Screenshot 2025-09-03 at 2 30 18 PM

emenslin (approved):
- [x] Ensure that (save for the coordinate decimal fields) the resulting Batch Edit Data Set matches the Query Results (sorting and filters are respected)
  - [x] Ensure that the coordinate decimals fields are replaced with the record's corresponding text field in the Data Set (e.g., `Latitude 1 -> Lat1text`, `Longitude 1 -> Long1text1`, etc.) 
  - [x] Ensure the usual WorkBench spatial tools operate as expected (available if the Data Set includes any of the decimal coordinate fields and after clicking the `Tools` button. Current Spatial Tools are: Convert Coordinates, GeoLocate, GeoMap, etc.)
  - [x] Ensure matching behavior operates as expected with the coordinate decimal fields: Locality records can be matched  to existing Locality records and Locality records can be properly Disambiguated when there are more than one Locality records that match
  - [x] Commit the Data Set and ensure the record's corresponding coordinate decimal field(s) are updated to their expected decimal values
  - [x] Batch Edit the query and ensure the coordinate decimal field is present with the value of the coordinate text field

---

Looks good! I ran into a potential problem with matching behavior where I can disambiguate but I was not able to get the matched and changed cell to appear but I'm not sure if this is the correct behavior or not

Gitesh307 (approved):
Looks Good!

image

PR #7397, opened 2025-09-08
Remove limit on number of matches in disambiguation results

### LINKED ISSUE #7238: Disambiguation dialog only lists up to 10 records
**Describe the bug**
The 'Disambiguate Multiple Record Matches' dialog in the WorkBench and Batch Edit only shows 10 records even if there are more possible records to match.

**To Reproduce**

1. Create more than 10 Geography records with the same name and rank but different parents
2. Go to the WorkBench
3. Create a data set with Geography as the base table
4. Map the rank of the Geography records from Step 1
5. Map the rank below that and save
6. In the first column, enter the name of the Geography records from step 1
7. Enter a value in the second column
8. Save and Validate
9. Right click on the resulting error cell and click Disambiguate
10. See that only 10 Geography records from Step 1 appear

**Expected behavior**
The dialog should list all possible matching records.

**Screenshots**

Possible matches (14):
Image

Disambiguation dialog (10):

Image

[Specify 7 System Information - 2025-07-25T18_37_30.959Z.txt](https://github.com/user-attachments/files/21436014/Specify.7.System.Information.-.2025-07-25T18_37_30.959Z.txt)

- OS: Mac
- Browser: Chrome
- Specify 7 Version: 7.9.6.2 through [main](https://github.com/specify/specify7/commit/964ac656a5150dd5694a6455ce001aa8f3652d89)
- Database Name: calvertmarinemuseum_2024_08_15
- Collection name: Invertebrates
- User Name: spadmin
- URL: https://calvertmarinemuseum20240815-v7962.test.specifysystems.org/specify/workbench/14/

**Reported By**
Chris Whitt from Bailey Matthews National Shell Museum & Aquarium

mark-pitblado:
Fixes #7238

I am not sure why there was a limit of the first 10 matches put in place (if anyone can think of it this PR will need to be reworked to solve the problem from a different angle!). This PR removes that limit entirely to solve the problem of only the first 10 matches being shown in disambiguation. I also tested with batch editing, which was the context where the limit was originally applied, and didn't run into any problems without it.

Formatted with [Black](https://github.com/psf/black) this time, since that is the formatter used in `pre-commit-config.yml`.

### Testing instructions

Adapted from the steps to reproduce the issue

- [ ] Create more than 10 Geography records with the same name and rank but different parents
- [ ] Go to the WorkBench. Create a data set with Geography as the base table
- [ ] Map the rank of the Geography records from Step 1
- [ ] Map the rank below that and save
- [ ] In the first column, enter the name of the Geography records from step 1
- [ ] Enter a value in the second column
- [ ] Save and Validate
- [ ] Right click on the resulting error cell and click Disambiguate
- [ ] See that **all** records from Step 1 appear. (Previously only ten would show)

Optional:
Any other testing that you think show why the limit was necessary.

### DIFF
--- a/specifyweb/backend/workbench/upload/treerecord.py
+++ b/specifyweb/backend/workbench/upload/treerecord.py
@@ -106,9 +106,9 @@ class TreeRank(NamedTuple):
         """
 
         assert self.rank_name is not None, "Rank name is required"
-        assert self.tree is not None and self.tree.lower() in SPECIFY_TREES, (
-            "Tree is required"
-        )
+        assert (
+            self.tree is not None and self.tree.lower() in SPECIFY_TREES
+        ), "Tree is required"
 
         if self.treedef_id is None:
             logger.info(
@@ -321,9 +321,12 @@ class ScopedTreeRecord(NamedTuple):
             definition_id=target_rank_treedef_id, parent=None
         ).first()
 
-        return self._replace(
-            treedef=target_rank_treedef, treedefitems=treedefitems, root=root
-        ), None
+        return (
+            self._replace(
+                treedef=target_rank_treedef, treedefitems=treedefitems, root=root
+            ),
+            None,
+        )
 
     """
         Adjusts tree scope for TreeRankRecords with NULL treedef_ids.
@@ -689,9 +692,9 @@ class BoundTreeRecord(NamedTuple):
             key = repr(sorted(tdiwpr.match_key() for tdiwpr in tried_to_match))
             return tdiwprs, MatchedMultiple(ids, key, info)
         else:
-            assert n_matches == 0, (
-                f"More than one match found when matching '{tdiwprs}' in '{model}'"
-            )
+            assert (
+                n_matches == 0
+            ), f"More than one match found when matching '{tdiwprs}' in '{model}'"
             if parent is not None:
                 info = ReportInfo(
                     tableName=self.name,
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (approved):
- [X] Create more than 10 Geography records with the same name and rank but different parents

>[!note]
>If any testers would like to recreate this, you can build a query on a tree table at a specific rank, export the CSV, and then add a column with the same value (e.g., `KansasTest`). After that, map it to the rank immediately below in a new WorkBench import and upload it.

I created 167 new 'KansasTest' states under a subset of countries:
Image

- [X] Go to the WorkBench. Create a data set with Geography as the base table
- [X] Map the rank of the Geography records from Step 1
- [X] Map the rank below that and save
- [X] In the first column, enter the name of the Geography records from step 1
- [X] Enter a value in the second column
- [X] Save and Validate
- [X] Right click on the resulting error cell and click Disambiguate
- [X] See that **all** records from Step 1 appear. (Previously only ten would show)

https://github.com/user-attachments/assets/68e12a99-8ba8-4c0f-ab9f-7a43240b41d0

Nice work as always! ⭐

emenslin (approved):
- [x] Create more than 10 Geography records with the same name and rank but different parents
- [x] Go to the WorkBench. Create a data set with Geography as the base table
- [x] Map the rank of the Geography records from Step 1
- [x] Map the rank below that and save
- [x] In the first column, enter the name of the Geography records from step 1
- [x] Enter a value in the second column
- [x] Save and Validate
- [x] Right click on the resulting error cell and click Disambiguate
- [x] See that **all** records from Step 1 appear. (Previously only ten would show)

---

Looks good, I checked in both workbench and batch edit and both are showing all results!

PR #7399, opened 2025-09-09
feat: add `ALLOW_SUPPORT_LOGIN` to Dockerfile

### LINKED ISSUE #7398: Support `ALLOW_SUPPORT_LOGIN` in Docker
IT support and admins sometimes need to impersonate users to troubleshoot, provide support, or conduct training.  Enabling it in managed deployments and development environments would allow administrators to use it securely within the Specify 7 docker container.

Allow enabling the `ALLOW_SUPPORT_LOGIN` environment variable in the Dockerfile and configuration so admins can generate support login tokens. This should be set to `True` in `specify_settings.py` or in the `docker-compose.yml` file for managed deployments. 

https://discourse.specifysoftware.org/t/allow-support-login-documentation/2838/1

Example configuration location:
https://github.com/specify/specify7/blob/214866318d0d0ae390fc06b66662979a50d60674/specifyweb/settings/specify_settings.py#L99-L100

Usage steps:
1. Enter the running Specify 7 container:
   ```bash
   docker exec -it specify7 /bin/bash
   ```
2. Run the support login command:
   ```bash
   ve/bin/python manage.py support_login --username ${username}
   ```
   Replace `${username}` with the desired account.
3. The generated token is valid for a time set by the `SUPPORT_LOGIN_TTL` environment variable (e.g., 300 seconds).
   Example output:
   ```bash
   The following token is valid for 300 seconds:
   /accounts/support_login/?token=1-1757106120-7394fa7d5ffc6f87fe8306d25c5b2c71b3f98942d9f9c46aea97f3eda725434b
   ```
4. Navigate to `base_url` + `/accounts/support_login/?token=...` to access the user's account.

- Temporarily resetting passwords and restoring the previous hash, which is time consuming and risky.
- Keeping support login disabled for all but development environments, requiring direct user intervention for troubleshooting.
- Relying on less secure or more manual methods for impersonation.

**Additional context**
> [!caution]
> This feature should be considered a security risk if exposed outside the container or to non-admin users. Only administrators with shell access to the container can use it, and it should be enabled only when necessary for support.
> The token is time-limited and usage is logged.

grantfitzsimmons:
Fixes #7398

### Checklist

### Testing instructions

https://discourse.specifysoftware.org/t/allow-support-login-documentation/2838/1

In your `.env` file, add the following:
```yml
ALLOW_SUPPORT_LOGIN=true
SUPPORT_LOGIN_TTL = 300
```

Usage steps:
1. Enter the running Specify 7 container:
   ```bash
   docker exec -it specify7 /bin/bash
   ```
2. Run the support login command:
   ```bash
   ve/bin/python manage.py support_login --username ${username}
   ```
   Replace `${username}` with the desired account.
3. The generated token is valid for a time set by the `SUPPORT_LOGIN_TTL` environment variable (e.g., 300 seconds).
   Example output:
   ```bash
   The following token is valid for 300 seconds:
   /accounts/support_login/?token=1-1757106120-7394fa7d5ffc6f87fe8306d25c5b2c71b3f98942d9f9c46aea97f3eda725434b
   ```
4. Navigate to `base_url` + `/accounts/support_login/?token=...` to access the user's account.


melton-jason (changes requested):
Currently[^1], the security of this approach _heavily_ relies on the SECRET_KEY to be kept a secret. 
If the SECRET_KEY is leaked, a malicious user could very easily **compromise _any_ user account on the system at _any_ time** without having to use the `python manage.py support_login` command. 

In the internal Specify Vault, I've put a video demonstrating exploiting this vulnerability:  https://drive.google.com/file/d/1P7zpPiUAxYQnSvhzfXn4UyJ3UkJPpi3h/view?usp=sharing

I know the SECRET_KEY should be kept secret, but this presents a **major** risk to institutions. 

There might be other ways to accomplish this, but I think we can make this more secure by adding a psuedo one-time salt/pad argument to the `manage.py support_login` command (or we can generate a key for the user at the time they run the command).  
We can then hash the token with this generated or provided key. 

The user-provided or generated key would then need to be provided when logging in via the `/accounts/support_login` endpoint. 

That way, a malicious agent would require _both_ the `SECRET_KEY` and the key generated (or provided) when running the `support_login` command to accomplish the exploit. 

I can make the changes to this PR  if needed. 

[^1]: This is not related to the changes in this PR, but with how the feature was initially implemented

---

On a couple _**much**_ smaller notes: 

Can we remove the following line? 😅 
Or add a comment linking to documentation. 

https://github.com/specify/specify7/blob/b45cef28f3c8a5109ee132f8c26a18b78d020f52/specifyweb/specify/auth/support_login.py#L22

Also, there's another option that's currently accepted by the `manage.py support_login` command: `--list`! 

https://github.com/specify/specify7/blob/b45cef28f3c8a5109ee132f8c26a18b78d020f52/specifyweb/specify/management/commands/support_login.py#L20-L26

It lists out all of the users, their user types, and whether they are a Specify 6 admin or not: 

Image

The output isn't very clean imo, and there's no way to view Specify 7 admins, so I'd be down to remove the option or clean it up and extend it!

Gitesh307 (approved):

PR #7404, opened 2025-09-11
Fix: replace legacy table icons with SvgIcon

### LINKED ISSUE #6504: Old table icons are used in result cell links
**Describe the bug**
The WorkBench/Batch Edit are using the old table icon for result cell links.

**To Reproduce**

1. Upload a data set using the WorkBench
2. Right click on a cell in the uploaded data set
3. See that the table icon for the record is using the Specify 6 style table icon

Image

**Expected behavior**
This should use the `SvgIcon` style that is used everywhere else in the app.

Gitesh307:
Fixes #6504

Replaced legacy Specify 6-style icons in WorkBench/Batch Edit result-cell links with standardized svgIcons assets. This ensures consistent icon usage across the app while avoiding reliance on outdated LegacyNonJSXIcons. The bug was caused by injecting raw SVG (Sp-6 styled) icons; the fix uses renderToStaticMarkup with our current icon set so visuals stay aligned without breaking existing DOM logic.

### Checklist

### Testing instructions

- Upload dataset using workbench.
- [ ] Right-click a cell in the uploaded dataset to check the link uses the SvgIcon table icon.
- [ ] Open Batch Edit on the same dataset and check same SvgIcon appears.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/WorkBench/WbSpreadsheet.tsx
+++ b/specifyweb/frontend/js_src/lib/components/WorkBench/WbSpreadsheet.tsx
@@ -7,6 +7,7 @@ import type Handsontable from 'handsontable';
 import type { DetailedSettings } from 'handsontable/plugins/contextMenu';
 import { registerAllModules } from 'handsontable/registry';
 import React from 'react';
+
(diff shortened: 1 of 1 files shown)


grantfitzsimmons:
I can test #7429 ?

PR #7413, opened 2025-09-12
Remove filtering uniqueness rule creation by discipline and avoid duplicates

### LINKED ISSUE #7412: Cases where not all uniqueness rules are created for new discipline in key migrations command
**Describe the bug**
Edit the logic for determining if a uniqueness rule record should be edited in the key migrations command.  Particularly, in the determination of whether a uniqueness rules should be applied to a discipline.  When a new discipline goes through all the migrations at startup, and then goes through the uniqueness rule commands in the key migrations, and then goes through other migrations in a newer version of Specify.  An issue can arise where the default uniqueness rules are not all added because it thinks the discipline isn't new due to uniqueness rules already existing for the discipline.  So more robust checking of discipline uniqueness rules is needed in the key migration command to avoid an issue like this.  See the testing instruction for a case where uniqueness rule table updates are not exactly what are wanted.

Here is the original comment on the key migrations PR.
https://github.com/specify/specify7/pull/6308#discussion_r2340917728

**To Reproduce**
1. Run migrations on a database on v7.10.  Make sure the database hasn't already gone through migrations that in are v7.11
2. Create a new Discipline in Specify 6.
3. Upgrade the database to use 7.11, so that the 7.11 migrations run.
4. Run the key migrations command.
5. See that not all the default uniqueness rules exist for the new discipline.

acwhite211:
Fixes #7412

Avoid the issue of missing uniqueness rules by not filtering disciplines, and instead making sure not to create any duplicates.  This way we can be sure not to be missing any default uniqueness rules while also making sure that we are not creating any duplicate uniqueness rules in the process.

### Checklist

### Testing instructions

- Use a database on v7.10 or v7.9.3, or prior.  Make sure the database hasn't already gone through migrations that in are v7.11
- Check the applied migrations with `ve/bin/python manage.py showmigrations`
- [x] Ensure that the business rules migrations doesn't have any migrations applied to it that are newer that `0005_cojo`
- [x] Ensure that the specify migrations doesn't have any migrations applied to it that are newer that `0023_update_schema_config_text`
- [x] Check the v7.10 or v7.9.3 branch tags, and rebuild and restart container so that the migrations run.
- Create a new Discipline in Specify 6.
- Upgrade the database to use this branch, startup the instance and see the migrations run.
- Run the key migrations command.
- [x] See that no duplicate default uniqueness rules were created for the new discipline.
- [x] The uniqueness rule count for each discipline should be 31.

```sql
select count(*), d.Name 
from uniquenessrule u
join discipline d on d.disciplineId = u.DisciplineID 
group by u.DisciplineID;
```

image

- [x] The uniqueness rule field count for each discipline should be 68.

```sql
select count(*), d.Name 
from uniquenessrule u
join discipline d on d.disciplineId = u.DisciplineID
join uniquenessrule_fields uf on uf.uniquenessruleid = u.uniquenessruleid
group by u.DisciplineID;
```

image

- [x] See that all the default uniqueness rules exist for the new discipline.  Do this by checking the database or by going through the uniqueness rules viewer in schema config for each table.  See query and snapshot.  Compare the results of the query to the file with all the defined default uniqueness rules https://github.com/specify/specify7/blob/main/specifyweb/backend/businessrules/uniqueness_rules.json.  For all the non-global default uniqueness rule fields in the uniqueness_rules.json file, there should be a matching record in the uniquenessrule_fields table for the new discipline.  Non-global uniqueness rules are rules that either have a dont' have a null scope, or if the field does not contain 'division' or 'institution'.

```sql
select u.modelName, u.isDatabaseConstraint, uf.fieldPath, uf.isScope 
from uniquenessrule u
join discipline d on d.disciplineId = u.DisciplineID
join uniquenessrule_fields uf on uf.uniquenessruleid = u.uniquenessruleid
where d.Name = '<New Discipline Name>'
order by u.modelName, uf.fieldPath;
```
Results should look like this:

|modelName|isDatabaseConstraint|fieldPath|isScope|
|---------|--------------------|---------|-------|
|Accessionagent|1|accession|1|
|Accessionagent|1|agent|0|
|Accessionagent|1|agent|0|
|Accessionagent|1|repositoryAgreement|1|
|Accessionagent|1|role|0|
|Accessionagent|1|role|0|
|Accessionauthorization|0|accession|1|
|Accessionauthorization|0|permit|0|
|Accessionauthorization|0|permit|0|
|Accessionauthorization|0|repositoryAgreement|1|
|Appraisal|1|accession|1|
|Appraisal|1|appraisalNumber|0|
|Author|1|agent|0|
|Author|0|orderNumber|0|
|Author|1|referenceWork|1|
|Author|0|referenceWork|1|
|Borrowagent|1|agent|0|
|Borrowagent|1|borrow|1|
|Borrowagent|1|role|0|
|Collectingeventauthorization|0|collectingEvent|1|
|Collectingeventauthorization|0|permit|0|
|Collectingtripauthorization|0|collectingTrip|1|
|Collectingtripauthorization|0|permit|0|
|Collection|0|code|0|
|Collection|0|collectionName|0|
|Collection|0|discipline|1|
|Collection|0|discipline|1|
|Collectionobject|0|catalogNumber|0|
|Collectionobject|0|collection|1|
|Collector|1|agent|0|
|Collector|1|collectingEvent|1|
|Determiner|1|agent|0|
|Determiner|1|determination|1|
|Disposalagent|1|agent|0|
|Disposalagent|1|disposal|1|
|Disposalagent|1|role|0|
|Extractor|1|agent|0|
|Extractor|1|dnaSequence|1|
|Fundingagent|1|agent|0|
|Fundingagent|1|collectingTrip|1|
|Gift|0|discipline|1|
|Gift|0|giftNumber|0|
|Giftagent|1|agent|0|
|Giftagent|1|gift|1|
|Giftagent|1|role|0|
|Groupperson|1|group|1|
|Groupperson|1|member|0|
|Loan|0|discipline|1|
|Loan|0|loanNumber|0|
|Loanagent|1|agent|0|
|Loanagent|1|loan|1|
|Loanagent|1|role|0|
|Localitycitation|1|locality|1|
|Localitycitation|1|referenceWork|0|
|Pcrperson|1|agent|0|
|Pcrperson|1|dnaSequence|1|
|Picklist|0|collection|1|
|Picklist|0|name|0|
|Preparation|1|barCode|0|
|Preparation|1|collectionobject__collection|1|
|Preptype|0|collection|1|
|Preptype|0|name|0|
|Taxontreedef|0|discipline|1|
|Taxontreedef|0|name|0|
|Taxontreedefitem|0|name|0|
|Taxontreedefitem|0|title|0|
|Taxontreedefitem|0|treeDef|1|
|Taxontreedefitem|0|treeDef|1|


acwhite211:
I did another test starting with v7.9.3 instead of v7.10.  A migration error occurred before getting to run the key migrations command.  I'm investigating.

```
  File "/opt/specify7/ve/lib/python3.12/site-packages/django/db/models/fields/__init__.py", line 2053, in get_prep_value
  return int(value)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'UniquenessRule'
```

melton-jason (approved):
Nice work, very simple! 👍 

Unfortunately, ensuring only the correct defaults are created is a very difficult problem. (e.g., how would we prevent creating a default rule that was intentionally removed?) 
Ideally we come up with a better solution here, but the problem is very complicated, so this solution might be fine for now.
  

When using this command as-is currently, we still have to watch out for the following case: 

- The database has one or more Uniqueness Rule migrations applied, up to some migration `j`
- One or more disciplines are created in Specify 6
- One or more migrations are run which create a rule with `isDatabaseConstraint=True`, up to some migration `k`
- When running this command, all uniqueness rules present at or before migration `j` will not be applied
  - i.e., the database will only have Uniqueness Rules from migrations `> j`  and `<= k`
  
  Currently, I don't think this is possible with the current set of migrations, though it may become possible in the future. 

@specify/developers @grantfitzsimmons 

To address the specific use case of a Discipline being created in Specify 6 and just wanting the defaults applied (i.e., there is little/no interaction of the Discipline in Specify 7 and it's reasonable that one or more default uniqueness rules have not been modified or deleted), could we add a flag/option to this command and/or define a new command which specifies the specific Disciplines for which to apply defaults? 

Operating under the assumption that the defaults (if they exist) have not been modified or deleted eliminates a _lot_ of complexity with this problem. 

If we add an option, it could look something like:  

```
python manage.py run_key_migration_functions --apply_uniqueness_rules <discipline 1 ID> --apply_uniqueness_rules <discipline 2 ID> ...
```

The default behavior (i.e., the option omitted or the DisciplineID not present in the list) could be to not touch Uniqueness Rules otherwise, or default to the behavior in this PR. 

If it were a different command, the usage may look something like: 
```
python manage.py run_key_migration_functions
python manage.py apply_uniqueness_rules <discipline 1 ID> <discipline 2 ID> ...
``` 

(or we could do both and the first merely calls the second: whether literally spawning a new process and calling the command in a shell, or just using the same Python logic/functions)

What do you guys think?

PR #7447, opened 2025-09-23
Decouple linting, formatting, and pre-commit from external dotfiles

### LINKED ISSUE #7402: Bring configuration files from maxpatiiuk's dotfiles into Specify respository
I think it would make sense to bring the current configuration settings for tools such as prettier, pre-commit, jscpd into the Specify orgnaization/repo. I have a huge amount of respect for Max, I hope that this would also make his life easier by reducing things he has to worry about if he changes his dotfiles.

For example, the current `prettier.config.js` imports the following lines via 

```javascript
import base from '@maxpatiiuk/prettier-config';
```

```javascript
/**
 * @see https://prettier.io/docs/en/configuration.html
 * @type {import("prettier").Config}
 */
const config = {
  // Single quote is more popular among JS libraries
  singleQuote: true,
  // GitHub renders Markdown in whitespace-insensitive way, so we should wrap prose
  proseWrap: 'always',
  plugins: [
    '@prettier/plugin-xml',
    'prettier-plugin-package',
    'prettier-plugin-sh',
    'prettier-plugin-tailwindcss',
  ],
};

export default config;
```

To see that the `singleQuotes: true` are part of the configuration, a developer has to do some digging. Instead, we could just move these lines into the config file already in the repository to reduce the number of places the code is. I don't see Specify needing to change anything about the files themselves, however, if a change is made, that only highlights why it would be beneficial here so the diffs can be better tracked.

Open to objections if there is a reason these should stay the way they are!

mark-pitblado:
Fixes #7402

This PR removes the link between [Max Platiiuk's dotfiles](https://github.com/maxpatiiuk/dotfiles/tree/main) and the Specify repository for formatting, linting and the pre-commit config. I have tried to stay as close to a copy and paste operation as possible, however the following two changes were made:

1. Previously, flake was pulling from a repository on GitLab, which required a login when I was testing with pre-commit. Since developers and testers for Specify are developing on GitHub, it seemed to make sense to switch to a GitHub based repo so that no additional setup is required.
2. The markdownlint used in pre-commit required ruby during my testing. Since ruby isn't one of the languages used by the Specify project, it seemed prudent to swap this out for a markdownlint based on nodejs. Markdownlinters *should* be pretty similar, there are very few rules for markdown.

Eslint is definitely where things get the most complicated, particularly because I don't think eslint has been a part of the ci pipeline for a while due to the high number of errors it produces?

https://github.com/specify/specify7/blob/cbff8981a9cbb2529eb90209adc801bcf5eda1f9/.github/workflows/test.yml#L270-L274

I think it may be prudent to think about *where* in the pipeline each type of error is expected to be caught, and which tools are mandatory to setup. For example, if a developer has pre-commit setup, are they good to go? Or should they also run `npm run test` on their machine? Which jobs are handled by the ci runners, compared to local testing? A good candidate for developer documentation.

> [!NOTE]  
> Gitguardian has a pre-commit config that can catch secrets *before* they are commited as a pre-commit check (or pushed as a pre-push check), whereas I believe the PR integration will only catch the secret *after* it has been committed (it can be then rotated/revoked, but that's a pain). An example of a case where it matters where a check is run.

### Dependency changes

The following dependencies were added to the Specify frontend. These were always dependencies for the project, they were just further upstream because they were dependencies for the dotfiles.

- [prettier-plugin-sh](https://www.npmjs.com/package/prettier-plugin-sh)
- [prettier-plugin-package](https://www.npmjs.com/package/prettier-plugin-package)
- [prettier-xml](https://github.com/prettier/plugin-xml)

### Checklist

### Testing instructions

- [ ] These changes are meta-changes, and shouldn't change any functionality of application code. The only test to be done is to run the ci and pre-commit pipelines and observe that there are no big changes suggested for files that have already been through the pipeline. Note that running the existing pre-commit configuration over the entire codebase will result in *a lot* of changes, it appears like many directories have never been touched by pre-commit (perhaps because they haven't seen changes in a while, or pre-commit is not consistently used).


grantfitzsimmons (approved):
Thank you so much for doing this! @CarolineDenis, we should discuss what changes need to be made to address
>Note that running the existing pre-commit configuration over the entire codebase will result in a lot of changes, it appears like many directories have never been touched by pre-commit (perhaps because they haven't seen changes in a while, or pre-commit is not consistently used).

CarolineDenis (approved):

mark-pitblado:
Looks like @g1rly-c0d3r has picked this task up in https://github.com/specify/specify7/pull/8491, so will close and let them proceed!

PR #7455, opened 2025-09-26
Fix autonumbering locking with GET_LOCK and RELEASE_LOCK

### LINKED ISSUE #6490: The Specify Worker unexpectedly stops responding
## **Describe the bug**

Quite often, we have a user report an issue to us where one of the following tools is unresponsive due to the **[Specify 7 Worker](https://github.com/specify/specify7/wiki#specify-worker)** process being offline:
- WorkBench
- Record Merging
- Batch Edit

This causes internal issues as well. @lexiclevenger has asked me to restart several instances (Donana, UC Davis, etc.) for conversion work. The worker often becomes unresponsive or fails unexpectedly. The logs for the worker in most, if not all, of these cases provide little information about the failure, and they seem accessible. The only way to reestablish the connection and allow the validation, uploading, and merging processes to resume is to completely restart the worker.

This is _not_ a new issue, and tickets related to this problem date back to 2021 (the introduction of this component).

In the WorkBench, users are told they should wait for another data set to be finished uploading:

https://github.com/specify/specify7/blob/fdc2301f29114a98a7a81a1e62812a7439ad93c1/specifyweb/frontend/js_src/lib/localization/workbench.ts#L1453-L1458

Of course, if the worker is unresponsive, this will never disappear, and they can be left waiting for hours/days until they message our team.
This issue is meant to capture such incidents and to call for a proper resolution to prevent these issues going forward.

As Max said in June 2021:

>As to the "Beginning the Validating..." message, currently, it can mean at least 3 things and there is no way for the front end to differentiate between them:
> 
> - The upload worker is preparing to validate the DS and validation should begin soon
> - The upload worker is currently busy with another DS and validation would not begin until that DS is validated
> - The upload worker is down or not responding. The validation would not begin and the "Beginning the Validating" would stay up forever
> - All of this also applies to Upload and Rollback.

## Reported By Institutions

Each bullet point has the subject line of the email/post where an issue was reported **due to the Specify Worker being offline**.

This is **not** a complete list, as often it was discused in meetings or under terms that made it difficult to find in the support tickets. I only went back to late 2023.

This is only what was directly reported to us, which further reduces the sample size.

### Natural History Museums of Denmark

- Web server timeouts related to WB validation (2024-06-24)

### University of Michigan

>We are currently experiencing an issue with multiple Specify 7 databases where Workbenches fail to complete validation. After selecting Validate, the process pauses at the Data Set Validation Status message for several minutes before eventually failing with a 'Failed aborting validation' message. 

- Specify 7 Workbench failures at 'Data Set Validation Status' message (2024-03-05)

### University of Kansas

From @acbentley (2024-02-21 on [Slack](https://specifydev.slack.com/archives/CC6V12D3J/p1708537285294759)):
> KU Ornithology just came to indicating that agent merging was not working for them.  They showed me an example of two agents with the last name Barbour that they wanted to merge.  When starting the merge process the merge dialog appears but then just sits there and doesn't complete as in the screenshot.  No error message is ever thrown.  I tried the same process with my admin account and got the same result.  I even tried a different agent duplicate (name A. A. Alcorn) and got the same result.  As I am not getting any error message I am unclear as to how to diagnose the issue.  Could someone take a look and see what is going on please?

From @acwhite211 (2024-02-21) on [Slack](https://specifydev.slack.com/archives/CC6V12D3J/p1708540122895369?thread_ts=1708537285.294759&cid=CC6V12D3J)):
> it was something to do with the connection between the redis and worker containers.  I restarted redis, and then restarted all the worker containers.  It's working now.
> The error on the worker side was
> ```
> [2024-02-20 22:54:24,034: ERROR/MainProcess] consumer: Cannot connect to redis://redis:6379/0: Error -2 connecting to redis:6379. Name or service not known..
> Trying again in 32.00 seconds... (16/100)
> ```
> But there were no errors on the redis side
> 

### College of Idaho
- Validate troubles in Batch Edit (2025-05-09)
- Data Set Validation Woes (2024-10-22)
- More workbench issues (2024-09-15)

From Theresa (2024-10-24):
> Grant is out of the office currently and he definitely has a bit more information on what was going on.  We did discuss a bit about options to prevent it from happening in the future.  While we are attempting to come up with a resolution, please let us know if it happens again.

 
### Donana

- I'm trying to validate this data set, but it keeps getting stuck on the 'data set validation status' dialog (2025-05-07)

### University of British Columbia
- Specify worker appears to be stuck (2025-02-12)

### Royal Botanic Gardens Edinburgh
- Unable to validate workbench import (2025-02-12)

### Museum fur Naturkunde, Berlin

- Specify 7 problems + Workshop (2025-04-03)

### The Hebrew University of Jerusalem

- Error during validation process (2023-12-07)

### Florida Fish and Wildlife

- Saving bug (?) in workbench (2024-07-23)

### Canadian Forestry Service

- Stuck on workbench validation (2024-12-04)

### The Ohio State University Mollusks

- CO: GUID field bugging out, and 1 other error (2024-09-10)

### University of Guam

- Attachment Upload Tool Issue (2025-01-23)
- issue opening jrxml files in Jaspersoft Studio (2024-10-02)

### theNAT (San Diego Natural History Museum)

- Worker was not running [[Speciforum](https://discourse.specifysoftware.org/t/specify7-login-error-for-one-user/1490/15?u=grant)] (2024-01-29)

### University of Minnesota Entomology

- Validation Not Working (2024-09-10)

### Montana State University

- Validation not working (2024-06-07)
- Workbench data set validation status (2024-01-12)

### Earlham College

- Merge Records Problem (2024-03-11)

## **To Reproduce**

The biggest issue is that we are not sure exactly what causes the worker to become unresponsive. 

1. Kill the `specify7-worker` container
2. Try to validate a data set, perform a merge, or upload data
3. See the user is given no message and is asked to simply wait forever

## **Expected behavior**

We need to make the worker processes more resilient to address unexpected and unresolvable outages.

acwhite211:
Fixes #6490

Look at Django docs, it seems that explicitly locking mysql tables from within a Django atomic transaction can lead to issues of tables not getting unlocked.  The `post_resource` function has the `transaction.atomic` decorator, which calls create_obj -> autonumber_and_save -> do_autonumbering -> lock_tables.  This is likely an issue.

LOCK TABLES implicitly commits any active transaction, and while tables are locked the connection is restricted to those tables only.  When Django later tries to manage the savepoint/transaction stack, the connection state can no longer matches expectations. This can leave the connection stuck.

This solution tires using `GET_LOCK` and `RELEASE_LOCK`, acting like a mutex.  Using these instead of `lock tables` should avoid problems of query transactions getting stuck.

This solution is still a work in progress.

### Checklist

### Testing instructions

TBD

### DIFF
--- a/specifyweb/specify/models_utils/lock_tables.py
+++ b/specifyweb/specify/models_utils/lock_tables.py
@@ -4,7 +4,6 @@ import logging
 
 logger = logging.getLogger(__name__)
 
-
 @contextmanager
 def lock_tables(*tables):
     cursor = connection.cursor()
@@ -13,8 +12,35 @@ def lock_tables(*tables):
         yield
     else:
         try:
+            # NOTE: Should not do within a transaction.atomic() block
             cursor.execute('lock tables %s' %
                            ' write, '.join(tables) + ' write')
             yield
         finally:
             cursor.execute('unlock tables')
+
+@contextmanager
+def mysql_named_lock(name: str, timeout: int = 10):
+    """
+    Connection-scoped mutex using MySQL GET_LOCK/RELEASE_LOCK.
+    Safe to use inside transaction.atomic().
+    """
+    if connection.vendor != "mysql":
+        yield
+        return
+
+    with connection.cursor() as cur:
+        cur.execute("SELECT GET_LOCK(%s, %s)", [name, timeout])
+        got = cur.fetchone()[0]
+
+    if not got:
+        raise TimeoutError(f"Could not acquire MySQL named lock {name!r}")
+
+    try:
+        yield
+    finally:
+        try:
+            with connection.cursor() as cur:
+                cur.execute("SELECT RELEASE_LOCK(%s)", [name])
+        except Exception:
+            logger.info("Failed to release MySQL named lock %r", name, exc_info=True)
--- a/specifyweb/specify/utils/autonumbering.py
+++ b/specifyweb/specify/utils/autonumbering.py
@@ -4,7 +4,7 @@ Autonumbering logic
 
 
 from .uiformatters import UIFormatter, get_uiformatters
-from ..models_utils.lock_tables import lock_tables
+from ..models_utils.lock_tables import mysql_named_lock
 import logging
 from typing import List, Tuple, Set
 from collections.abc import Sequence
@@ -43,12 +43,16 @@ def do_autonumbering(collection, obj, fields: list[tuple[UIFormatter, Sequence[s
         for formatter, vals in fields
     ]
 
-    with lock_tables(*get_tables_to_lock(collection, obj, [formatter.field_name for formatter, _ in fields])):
+    # Build a stable mutex key that scopes contention correctly.
+    # Example: per-discipline + model + the specific fields being autonumbered.
+    key_fields = ",".join(sorted(f.field_name for f, _ in fields))
+    lock_name = f"autonumber:{collection.discipline_id}:{obj._meta.db_table}:{key_fields}"
+
+    # Serialize the autonumbering critical section without table locks.
+    with mysql_named_lock(lock_name, timeout=10):
         for apply_autonumbering_to in thunks:
             apply_autonumbering_to(obj)
-
         obj.save()
-
 
 def get_tables_to_lock(collection, obj, field_names) -> set[str]:
     # TODO: Include the fix for https://github.com/specify/specify7/issues/4148


acwhite211:
I created some other solutions that try to avoid race conditions when doing autonumbering, but none are fully satisfactory:
- https://github.com/specify/specify7/compare/issue-6490-3 - Using a combination of named locks and table locks
- https://github.com/specify/specify7/compare/issue-6490-4 - Tries using Django's select_for_update to implement the necessary locking for autonumbering
- https://github.com/specify/specify7/compare/issue-6490-5 - Tries row-level locking with Django for autonumbering, but I don't think it fully guarantees that no race conditions could occur during autonumbering

acwhite211:
TODO: add indexes to all the common used autonumbering fields.

specifyweb/specify/migrations/0042_add_indexes_to_help_autonumbering.py
melton-jason: (just a note)
Very nice, glad we're getting some more indices here! 🏆 
I would love to see other indices (like those proposed in [Add missing field indexes (#7482)](https://github.com/specify/specify7/issues/7482)), although that could definitely be a future PR.

specifyweb/specify/models_utils/lock_tables.py
```diff
@@ -13,8 +12,59 @@ def lock_tables(*tables):
         yield
     else:
         try:
+            # NOTE: Should not do within a transaction.atomic() block
```
melton-jason: (optional)

Can you provide a reference here to #6490, https://github.com/specify/specify7/issues/6490#issuecomment-3020675840, https://github.com/specify/specify7/issues/6490#issuecomment-3340619060, and/or this PR?

specifyweb/specify/migrations/0001_initial.py
```diff
@@ -5762,6 +5762,36 @@ class Migration(migrations.Migration):
...
+        migrations.CreateModel(
+            name='AutonumschDsp',
+            fields=[
+                ('disciplineid', models.ForeignKey(db_column='DisciplineID', on_delete=django.db.models.deletion.DO_NOTHING, primary_key=True, serialize=False, to='specify.discipline')),
+            ],
+            options={
+                'db_table': 'autonumsch_dsp',
+                'managed': False,
+            },
+        ),
```
melton-jason: Are we keeping these in the initial migration? 
If so, my previous comments still apply: https://github.com/specify/specify7/pull/7455#discussion_r2543046732, https://github.com/specify/specify7/pull/7455#discussion_r2543051371

specifyweb/specify/models_utils/lock_tables.py
```diff
@@ -13,8 +12,63 @@ def lock_tables(*tables):
...
+        raise TimeoutError(f"Could not acquire MySQL named lock {lock_name!r}")
+
+    try:
+        yield
+    finally:
+        try:
+            with connection.cursor() as cur:
+                cur.execute("SELECT RELEASE_LOCK(%s)", [lock_name])
+        except Exception:
+            logger.info("Failed to release MySQL named lock %r", lock_name, exc_info=True)
```
melton-jason: Do we have to catch this Exception here?
I'm not sure if the `connection.cursor` can ever raise an exception, but this code might be ran inside of a transaction.

melton-jason (changes requested):
I had never heard of [GET_LOCK](https://mariadb.com/docs/server/reference/sql-functions/secondary-functions/miscellaneous-functions/get_lock) before this PR! 
Looks like it has potential and this implementation handles the implicit transaction commit problem with LOCK_TABLES mentioned in https://github.com/specify/specify7/issues/6490#issuecomment-3020675840 and https://github.com/specify/specify7/issues/6490#issuecomment-3340619060 👏. 

I don't think it's mentioned in this PR or on the Issue yet, but this PR would subsume/replace #5404. 

--- 

I will provide a note here that this implementation utilizing the advisory GET_LOCKS and [row level locks from FOR UPDATE](https://mariadb.com/docs/server/server-usage/storage-engines/innodb/innodb-lock-modes)  is more vulnerable  then the prior exclusive full table lock. 
Notably, if there are any database triggers or people are interacting directly with the database, outside from the row level update locks, there are little safe guards in the way to prevent "messing up" or interfering with Specify's autoincrementing code (e.g., Specify is vulnerable to race conditions related to record inserts).
For example, it is now possible for some agent with a direct connection to the database to insert a record while Specify is autoincrementing (after it has read the largest record but before committing the new record), resulting in Specify and the agent inserting "duplicate" or "out-of-sync with the numbering scheme" records

PR #7469, opened 2025-10-02
feat(queries): cast `catalogNumber` only if it is numeric

### LINKED ISSUE #7468: Catalog numbers in other collections are converted to integers, regardless of the configured format
**Describe the bug**
String catalog numbers shown in other collections via the Collection Relationship mechanism are still cast to integers, resulting in the catalog number displaying as `0` instead of the original value.

**To Reproduce**

1. Use a database where there is a collection relationship from one collection with a `CatalogNumberNumeric` format to one with a string
2. Build a query that includes both the catalog number for the object in the main collection and the catalog number from the related collection (via the Collection Relationship mechanism)
3. See that the string catalog numbers are incorrectly cast (shown as `000000000`)

In my example, the related catalog number is `F-235694`, but when cast, this becomes `0`. This is shown in the query results as `000000000`.

Image

[Cat # Issue.json](https://github.com/user-attachments/files/22651706/Cat.Issue.json)

*When reporting an issue on the `production` branch or any development version of Specify, please try to recreate the issue on the latest tagged release (`v7`) and indicate if it does not occur there, indicating a regression.*

**Expected behavior**
These non-numeric catalog numbers should not be cast to ensure the correct value is shown. If it is an integer, it should still be cast to preserve existing behavior.

**Screenshots**

**Current behavior:**
Image

**Desired behavior:**
Image

**Reported By**
Willem at SAIAB via [Specify Help Desk](https://specifysoftware.bolddesk.com/agent/tickets/250)

grantfitzsimmons:
Fixes #7468 

There is an issue where catalog numbers shown in other collections via the Collection Relationship mechanism are still cast to integers, resulting in the catalog number displaying as `0` instead of the original value. These non-numeric catalog numbers should not be cast to ensure the correct value is shown. If it is an integer, it should still be cast to preserve existing behavior.

**Screenshots**

**Current behavior:**
Image

**Desired behavior:**
Image

### Checklist

- [ ] Add automated tests

### Testing instructions
1. Use a database where there is a collection relationship from one collection with a `CatalogNumberNumeric` format to one with a string
2. Build a query that includes both the catalog number for the object in the main collection and the catalog number from the related collection (via the Collection Relationship mechanism)
3. See that the string catalog numbers are incorrectly cast (shown as `000000000`)

In my example, the related catalog number is `F-235694`, but when cast, this becomes `0`. This is shown in the query results as `000000000`.

Image

[Cat # Issue.json](https://github.com/user-attachments/files/22651706/Cat.Issue.json)

- [ ] Verify that this casting issue is no longer occuring
- [ ] Test reports and labels to verify there are no crashes when displaying the catalog number

### DIFF
--- a/specifyweb/backend/stored_queries/format.py
+++ b/specifyweb/backend/stored_queries/format.py
@@ -415,11 +415,15 @@ class ObjectFormatter:
         
         if self.numeric_catalog_number and specify_field is CollectionObject_model.get_field('catalogNumber') \
                 and all_numeric_catnum_formats(self.collection):
+            # Cast to Numeric only if the value is numeric to avoid casting strings like 'F-235694' to 0
             # While the frontend can format the catalogNumber if needed,
             # processes like reports, labels, and query exports generally
             # expect the catalogNumber to be numeric if possible.
             # See https://github.com/specify/specify7/issues/6464
-            return cast(field, types.Numeric(65))
+            return case(
+                [(field.op('REGEXP')('^-?[0-9]+(\.[0-9]+)?$'), cast(field, types.Numeric(65)))],
+                else_=field
+            )
 
         if self.format_types and specify_field.type == 'json' and isinstance(field.comparator.type, types.JSON):
             return cast(field, types.Text)


emenslin (approved):
- [x] Verify that this casting issue is no longer occuring
- [x] Test reports and labels to verify there are no crashes when displaying the catalog number

Looks good! 
image

acwhite211 (approved):

PR #7470, opened 2025-10-02
Fix: Apply case-sensitive filter in uniqueness validation

### LINKED ISSUE #4058: Case sensitivity discrepancy between front/backend uniqueness rules
> I am closing this for now. The uniqueness rule has been implemented in `v7.9`.
> 
> However, when verifying this behavior, I did stumble across a bug. The frontend uniqueness rule does its search in a case-sensitive manner while the backend searches and verifies in a case-insensitive way.
> 
> For example, if there is a specifyuser called `spadmin`, the frontend will allow creating a user with name `sPadmin`, but this will thow a backend exception. This behavior is most likely not specific to this rule, so I will open a new issue (hence the reasoning behind closing this one)

- https://github.com/specify/specify7/issues/3920#issuecomment-1741376567

Gitesh307:
Fixes #4058 

This PR addresses the case sensitivity discrepancy between frontend and backend uniqueness rules. Previously, the frontend enforced case-sensitive uniqueness, but the backend validation raised exceptions due to case-insensitive checks. The implementation has been updated so that backend uniqueness validation now uses case-sensitive matching (BINARY comparison) on string fields, while preserving the original behavior for related lookups and non-string fields. This brings the backend into alignment with the frontend, ensuring consistent validation across the application.

### Checklist

### Testing instructions

- [ ] Define a uniqueness rule on a text field (e.g., remarks) in Uniqueness Rules.
- [ ] Create a record (e.g., Accession) with a text value, e.g., abc.
- [ ] Create another record with the same text in different case, e.g., aBc.
- [ ] Verify: the record is created successfully without any uniqueness error.
- [ ] Verify that unique fields (e.g., username) correctly block case-sensitive duplicates on the frontend by showing a validation warning instead of allowing submission and throwing a backend BusinessRuleException.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
+++ b/specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
@@ -25,45 +25,6 @@ import type {
 mockTime();
 requireContext();
 
-const overrideCollectionObjectIexact = (
-  encodedCatalogNumber: string,
-  hasDuplicate: boolean
-): void => {
-  overrideAjax(
-    `/api/specify/collectionobject/?domainfilter=false&catalognumber__iexact=${encodedCatalogNumber}&collection=4&offset=0`,
-    {
-      objects: hasDuplicate
-        ? [
-            {
-              id: 1,
-              catalogNumber: decodeURIComponent(encodedCatalogNumber),
-              collection: '/api/specify/collection/4/',
-            },
-          ]
-        : [],
-      meta: {
-        limit: 20,
-        offset: 0,
-        total_count: hasDuplicate ? 1 : 0,
-      },
-    },
-    {},
-    true
-  );
-};
-
-(
-  [
-    ['000000001', true],
-    ['%23%23%23%23%23%23%23%23%23', false],
-    ['', false],
-    ['7', false],
-    ['2022-%23%23%23%23%23%23', false],
-  ] as const
-).forEach(([value, hasDuplicate]) =>
-  overrideCollectionObjectIexact(value, hasDuplicate)
-);
-
 describe('Borrow Material business rules', () => {
   const borrowMaterialId = 1;
   const borrowMaterialUrl = getResourceApiUrl(
--- a/specifyweb/frontend/js_src/lib/tests/ajax/static/businessrules/uniqueness_rules/3.json
+++ b/specifyweb/frontend/js_src/lib/tests/ajax/static/businessrules/uniqueness_rules/3.json
@@ -101,7 +101,7 @@
         "fields": ["catalogNumber"],
         "scopes": ["collection"],
         "modelName": "Collectionobject",
-        "isDatabaseConstraint": true
+        "isDatabaseConstraint": false
       }
     },
     {


emenslin (approved):
- [x] Define a uniqueness rule on a text field (e.g., remarks) in Uniqueness Rules.
- [x] Create a record (e.g., Accession) with a text value, e.g., abc.
- [x] Create another record with the same text in different case, e.g., aBc.
- [x] Verify: the record is created successfully without any uniqueness error.

Looks good, I tried with a few different text fields and I didn't run into any issues.

specifyweb/backend/businessrules/uniqueness_rules.py
```diff
@@ -136,7 +152,46 @@ def get_exception(conflicts, matchable, field_map):
...
+                        F(field_path),
+                        function='BINARY',
+                        output_field=dj_models.TextField(),
+                    )
+                    transformed_filters[alias] = value
+                    filter_kwargs.pop(field_path, None)
+
+            if annotations:
+                conflicts_query = conflicts_query.annotate(**annotations)
+                filter_kwargs.update(transformed_filters)
```
melton-jason, line 192: (optional) 

Similar to https://github.com/specify/specify7/pull/7450#discussion_r2382962118: 

Can you extract this case-sensitivity logic into one or more separate functions?
I imagine it could look something like: 

```py
def validate_unique(model, instance): 
    .... # Previous validate unique code

    if apply_case_sensitive: 
        conflicts_query, applied_case_sensitive_filters = case_sensitive_filters(conflict_query, matchable)
        filter_kwargs.update(applied_case_sensitive_filters)

    .... # Any other query transformations
    conflicts = conflicts_query.filter(**filter_kwargs)

def case_sensitive_filters(queryset, matchable): 
    # Code from above here
    ...
```

specifyweb/frontend/js_src/lib/components/DataModel/businessRules.ts
```diff
@@ -385,10 +393,20 @@ export class BusinessRuleManager<SCHEMA extends AnySchema> {
...
+          this.resource.specifyTable,
+          fieldName
+        ).at(-1);
+          if (
+            rule.isDatabaseConstraint &&
+            typeof value === 'string' &&
+            leafField !== undefined &&
+            leafField.isRelationship === false
+          )
+            return [`${fieldName}${djangoLookupSeparator}iexact`, value];
```
melton-jason, line 407: (optional) 

To make this a little more readable, we could add `iexact` to the `backendFilter` function: 

https://github.com/specify/specify7/blob/87e1107cb16365abcc910eb98d6f275a49f31f3d/specifyweb/frontend/js_src/lib/components/DataModel/helpers.ts#L122-L142

This could then be modified to something like the following (this could probably be cleaned up even further):

```suggestion
            return Object.entries(backendFilter(fieldName).caseInsensitiveEquals(value)).at(0)!
```

specifyweb/frontend/js_src/lib/components/DataModel/__tests__/businessRules.test.ts
```diff
@@ -25,6 +25,45 @@ import type {
...
+  [
+    ['000000001', true],
+    ['%23%23%23%23%23%23%23%23%23', false],
+    ['', false],
+    ['7', false],
+    ['2022-%23%23%23%23%23%23', false],
+  ] as const
+).forEach(([value, hasDuplicate]) =>
+  overrideCollectionObjectIexact(value, hasDuplicate)
+);
```
melton-jason, line 65: (can be handled in a different PR) 
Whoops! This change was needed because the frontend testing environment has `isDatabaseConstraint = true` for the Catalognumber must be unique to Collection rule: 

https://github.com/specify/specify7/blob/87e1107cb16365abcc910eb98d6f275a49f31f3d/specifyweb/frontend/js_src/lib/tests/ajax/static/businessrules/uniqueness_rules/3.json#L97-L106

This was true before #5400, but the default was changed such that `isDatabaseConstraint = false` after #5400. 
I never updated the frontend testing environment as a result of the change 🙈 

We just need to update the static uniqueness rules response and these lines can be removed. 

Otherwise, if we _do_ want to keep these lines, I would suggest making this solution more scalable for future tests: perhaps we could define some global in the test file to store all CollectionObjects that will be defined and use the global here instead of defining all catalognumbers separately (both in this override and when they are defined in the actual test).

melton-jason (approved):
Nice work! 👏 🚀 

Nothing major in my comments, primarily just refactoring/code maintainability suggestions.

PR #7457, opened 2025-09-29
Fix type declaration in group_concat and blank_nulls

acwhite211:
Fixes #7423

Resolve sqlalchemy compiling issues related to the `group_concat` and `blank_nulls` by making their definitions more friendly to sqlachlemy 1.4.  The main way to do this is my setting `type = String()`, which gives the expression a real TypeEngine, so the compiler/comparator path won’t hit None/NullType.

### Checklist

- [x] Self-review the PR after opening it to make sure the changes look good and
      self-explanatory (or properly documented)
- [x] Add relevant issue to release milestone
- [ ] Add pr to documentation list

### UX Testing instructions

- Use the vpl database.
- [x] Through the Specify Taxon Tree UI, merge two taxon nodes (e.g., 'Palaeoniscidae (in Palaeonisciformes)' into 'Palaeoniscidae (in Palaeoniscoidea)').  See that the taxon merge occurred without errors.

Running the merge:
Image

Taxon Tree after merge:
Image

- [x] Try merging two other random taxon nodes in three to make sure no errors occur.

### Development Testing instructions

- Query the database with:
    ```sql
    select count(*)
    from taxon t
    join taxon p on t.parentid = p.taxonid
    where t.nodenumber not between p.nodenumber and p.highestchildnodenumber;
    ```
- [x] See that it returns a count of `0`
- Run the `renumber_tree('taxon')` function directly via the Django Python shell.
    ```python
    from specifyweb.backend.trees import extras
    from django.db import connection
    extras.renumber_tree('taxon')
    ```

    ```
    ❯ docker exec -it specify7 /bin/bash
    specify@4f497e2137b1:/opt/specify7$ ve/bin/python manage.py shell
    Python 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    (InteractiveConsole)
    >>> from specifyweb.backend.trees import extras
    >>> from django.db import connection
    >>> extras.renumber_tree('taxon')
    [15/Sep/2025 12:14:47] [INFO] [specifyweb.specify.tree_extras:643] renumbering tree
    ```
- [x] The command `renumber_tree('taxon')` ran without errors.
- Query the database with:
    ```sql
    select count(*)
    from taxon t
    join taxon p on t.parentid = p.taxonid
    where t.nodenumber not between p.nodenumber and p.highestchildnodenumber;
    ```
- [x] The query should return 0.
[...]

PR #7528, opened 2025-10-30
New renumber_tree function to fix node numbering ordering issue in generated query

### LINKED ISSUE #7423: Tree renumbering function introduces invalid numbering, causing merge failures
**Description:**
There is an issue in Specify 7 where the `renumber_tree` function is not properly assigning `nodenumber` and `highestchildnodenumber` values, resulting in tree numbering errors that prevent merges from completing successfully. This issue was reported by Tomislav and has been consistently [reproducible in their database](https://drive.google.com/open?id=10rZphDlapJUUEQBamGqWqCnM78_LJV-q&usp=drive_fs).

**Steps to Reproduce:**

1. Query the database with:   
   ```sql
   select count(*) from taxon t join taxon p on t.parentid = p.taxonid
   where t.nodenumber not between p.nodenumber and p.highestchildnodenumber;
   ```

   See that it returns a count of `0`

1. Run the `renumber_tree('taxon')` function directly via Django shell.
    ```
    ❯ docker exec -it specify7 /bin/bash
    specify@4f497e2137b1:/opt/specify7$ ve/bin/python manage.py shell
    Python 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    (InteractiveConsole)
    >>> from specifyweb.specify import tree_extras
    >>> from django.db import connection
    >>> tree_extras.renumber_tree('taxon')
    [15/Sep/2025 12:14:47] [INFO] [specifyweb.specify.tree_extras:643] renumbering tree
    ```

3. Query the database with:   
   ```sql
   select count(*) from taxon t join taxon p on t.parentid = p.taxonid
   where t.nodenumber not between p.nodenumber and p.highestchildnodenumber;
   ```
   This returns a non-zero count (e.g., `2274`).

3. Attempt to merge two taxon nodes (e.g., 'Palaeoniscidae (in Palaeonisciformes)' into 'Palaeoniscidae (in Palaeoniscoidea)').

4. When `validate_tree_numbering('taxon')` is run, it throws:
   ```
   AssertionError: found 2274 nodenumbers not nested by parent
   ```

   ```bash
    >>> tree_extras.validate_tree_numbering('taxon')
    [15/Sep/2025 12:14:48] [INFO] [specifyweb.specify.tree_extras:598] validating tree
      File "<console>", line 1, in <module>
      File "/opt/specify7/specifyweb/specify/tree_extras.py", line 624, in validate_tree_numbering
        assert not_nested_count == 0, \
               ^^^^^^^^^^^^^^^^^^^^^
    AssertionError: found 2274 nodenumbers not nested by parent
   ```
**Specify 6 Behavior (Working as Expected):**
- Running "Update Taxon tree" and "Update Geography tree" in Specify 6 (which triggers the tree renumbering logic) successfully resets the bad count to 0. Merges in this version work without errors after that.
- In Specify 6, further merges do not cause the numbering issue to reappear.

**Specify 7 Behavior (Not Working):**
- In Specify 7, even after manually correcting the numbering (either by database fix or using external tools), every time a merge is attempted, the application automatically attempts to renumber the tree, re-introducing faulty numbering.
- This results in the error resurfacing with every merge action and the blocking of merges.
- This auto-renumber process makes it impossible to maintain consistent numbering or migrate data as needed.

**Impact:**
- Users are unable to merge taxonomy trees in Specify 7 databases due to this bug.
- The operations work in Specify 6.
- The bug blocks critical data curation operations.

**Environment:**
This is happening in `v7.11.1`. Testing with `6.8.03`.

**Relevant logs and output:**
After running renumber:
```
[INFO] renumbering tree
```
After validation:
```
AssertionError: found 2274 nodenumbers not nested by parent
```
After running "Update Taxon tree" or "Update Geography tree" in Specify 6:

> Count updated to 0. No errors present.

**Relevant code:**

https://github.com/specify/specify7/blob/214866318d0d0ae390fc06b66662979a50d60674/specifyweb/specify/tree_extras.py#L597-L625

https://github.com/specify/specify7/blob/214866318d0d0ae390fc06b66662979a50d60674/specifyweb/specify/tree_extras.py#L642-L734

**Reported by:**
Tomislav at University of Texas at Austin [Specify Help Desk ticket #159](https://specifysoftware.bolddesk.com/agent/tickets/150)

acwhite211:
Fixes #7423 

User-variable numbering doesn't seems reliable in MariaDB 11.8.  MariaDB 11.8 is more aggressive about optimizing/merging derived tables. With `(@rn := @rn + 1)` inside a subquery, the optimizer may reorder evaluation.  ORDER BY inside a derived table is not guaranteed.

The other issue with the existing renumber_tree function, used to taxon node renumbering, was due to the use of lexical numbering, instead of numerical numbering, used in the sorting of path expressions.  This issue is more likely to appear when there is a node with lots of children.

Numerically Sorted: `2 < 10 < 100`
Lexicographically Sorted: `"10" < "100" < "2"`

Other improvements like logging and readability improved as well in this new solution.

Attempts at making a CTE query solution for node renumbering were made, but no attempts were success in running within a reasonable time.

### Checklist

- [x] Add pr to documentation list

### Testing instructions

Same as testing instructions in #7423 and #7457

### UX Testing instructions

- Use the vpl database.
- [x] Through the Specify Taxon Tree UI, merge two taxon nodes (e.g., 'Palaeoniscidae (in Palaeonisciformes)' into 'Palaeoniscidae (in Palaeoniscoidea)').  See that the taxon merge occurred without errors.

Running the merge:
Image

Taxon Tree after merge:
Image

- [x] Try merging two other random taxon nodes in three to make sure no errors occur.

### Development Testing instructions

NOTE: When running locally, make sure to be using MariaDB version 11.8 to test this PR.

- Query the database with:
    ```sql
    select count(*)
    from taxon t
    join taxon p on t.parentid = p.taxonid
    where t.nodenumber not between p.nodenumber and p.highestchildnodenumber;
    ```
- [x] See that it returns a count of `0`
- Run the `renumber_tree('taxon')` function directly via the Django Python shell.
    ```python
    from specifyweb.backend.trees import extras
    from django.db import connection
    import logging
    extras.renumber_tree('taxon')
    extras.validate_tree_numbering('taxon')
    ```

    ```
    ❯ docker exec -it specify7 /bin/bash
    specify@4f497e2137b1:/opt/specify7$ ve/bin/python manage.py shell
    Python 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    (InteractiveConsole)
    >>> from specifyweb.backend.trees import extras
    >>> from django.db import connection
    >>> extras.renumber_tree('taxon')
    [15/Sep/2025 12:14:47] [INFO] [specifyweb.specify.tree_extras:643] renumbering tree
    ```
- [x] The command `renumber_tree('taxon')` ran without errors.
- Query the database with:
    ```sql
    select count(*)
    from taxon t
    join taxon p on t.parentid = p.taxonid
    where t.nodenumber not between p.nodenumber and p.highestchildnodenumber;
    ```
- [x] The query should return 0.
- [x] Through the Specify Taxon Tree UI, merge two taxon nodes (e.g., 'Palaeoniscidae (in Palaeonisciformes)' into 'Palaeoniscidae (in Palaeoniscoidea)').  See that the taxon merge occurred without errors.
- Go back to the python shell and run the `validate_tree_numbering('taxon')` command
```python
from specifyweb.backend.trees import extras
from django.db import connection
extras.validate_tree_numbering('taxon')
```
- [x] See that the `validate_tree_numbering('taxon')` command ran without errors

### DIFF
--- a/specifyweb/backend/trees/extras.py
+++ b/specifyweb/backend/trees/extras.py
@@ -747,7 +747,7 @@ def renumber_tree(table: str) -> None:
     cursor = connection.cursor()
     logger.debug(f"[renumber_tree] running for {table}")
 
-    cursor.execute("SET SESSION sql_safe_updates = 0") # might be needed for MariaDB 11
+    # cursor.execute("SET SESSION sql_safe_updates = 0") # might be needed for MariaDB 11, uncomment if updates don't occur
 
     # Sync rankid
     sql_sync = (


specifyweb/backend/trees/extras.py
```diff
@@ -742,3 +742,80 @@ def is_treedef(obj):
...
+            return ""  # only t0 exists
+        return "\n".join(
+            f"LEFT JOIN {tbl} t{j} ON t{j-1}.parentid = t{j}.{tbl}id"
+            for j in range(1, d)
+        )
+
+    def tree_path_expr(tbl: str, d: int) -> str:
+        # Highest ancestor first, leaf last; LPAD stabilizes lexical order
+        parts = ", ".join([f"LPAD(t{i}.{tbl}id, 12, '0')" for i in reversed(range(d))])
+        return f"CONCAT_WS(',', {parts})"
```
melton-jason: I don't think we have to care about the lexical vs numerical sorting differences here?

If we have a taxon tree like the following: 

```mermaid
flowchart TD
    1 --> 7
    1 --> 3
    3 --> 10
    3 --> 2
    10 --> 5
    5 --> 20
    10 --> 8
    1 --> 4
    4 -- Empty node = skipped rank --> A(" ")
    4 --> 50
    A -- Parent is 4 --> 6
```

The resulting paths (without padding 0s) would be as following: 

| TaxonID | Path       |
|---------|------------|
| 1       | 1          |
| 3       | 1,3        |
| 7       | 1,7        |
| 4      | 1,4        |
| 10      | 1,3,10     |
| 100     | 1,3,100    |
| 50      | 1,4,50      |
| 5       | 1,3,10,5   |
| 8       | 1,3,10,8   |
| 6        | 1,4,6      |
| 2       | 1,3,10,5,2 |

For each node, we still maintain the ancestry chain up to that node; I believe the differences between a lexicographic sort and numeric sort will **only** manifest at nodes with the same parent, but at that point we shouldn't care which child node gets traversed first when numbering in these cases (i.e., it's fine if `1,2` < `1,10`), the tree should be valid regardless of the traversal order of numbering for children nodes.

That being said, I think the only real concern in keeping or removing this `LPAD` approach is performance. 
I haven't done any analysis

melton-jason: In my own testing I removed the `tree_path_expr` and `tree_parent_joins` functions and replaced them (respectively) with the pre-existing `path_expr` and `parent_joins` functions. 
The tree re-renumbering still seemed to be successful when using the originals.

specifyweb/backend/trees/extras.py
```diff
@@ -742,3 +742,80 @@ def is_treedef(obj):
...
+        f"UPDATE {table} t "
+        f"JOIN {table}treedefitem d ON t.{table}treedefitemid = d.{table}treedefitemid "
+        f"SET t.rankid = d.rankid"
+    )
+    logger.debug(sql_sync)
+    cursor.execute(sql_sync)
+
+    # Compute max logical depth to size the LEFT JOIN ladder
+    cursor.execute(f"SELECT COUNT(DISTINCT rankid) FROM {table}")
+    depth = cursor.fetchone()[0] or 1
```
melton-jason: (If not in this PR, then perhaps for a different PR)
Separate trees

specifyweb/backend/trees/extras.py
```diff
@@ -742,3 +742,80 @@ def is_treedef(obj):
...
+            f"UPDATE {table} t\n"
+            f"JOIN (\n"
+            f"  SELECT parentid, MAX(highestchildnodenumber) AS hcnn\n"
+            f"  FROM {table}\n"
+            f"  WHERE rankid > %(rank)s\n"
+            f"  GROUP BY parentid\n"
+            f") sub ON sub.parentid = t.{table}id\n"
+            f"SET t.highestchildnodenumber = sub.hcnn\n"
+            f"WHERE t.rankid = %(rank)s"
+        )
```
melton-jason: (Just a note!)
I tested node numbering with different taxon trees in the same discipline

Image

melton-jason (approved):
Generally pretty solid! 👌 🚀 
I tested two different database backups (a copy of NHMD and KUFish) as well as many custom-built trees (including trees in the same discipline) using MariaDB 11.7.2 and MariaDB 10.11. 

I don't think there's anything inherently wrong or pressing with these changes[^1], although there does seem to be superfluous (unnecessary/extra) changes: using LPAD to enforce a numerical-style ordering, redefining the `parent_joins` and `path_expr` helper functions, disabling `sql_safe_updates`.

I don't think they have a functional impact on the code, although (especially with the LPAD changes) there may be other side-effects such as performance impacts or newly introduced breaking edge cases. 

I'd love to see some automated tests for this as well!

--- 
Here's some testing artifacts: 

Below is a graph of a simplified Taxon tree which had node numbering performed via `main`, and has 8 taxon records where the node number is not between the parent's node number and highest child node number. 

The notation of the graph for each node is of the form `nodeNumber <> highestChildNodeNumber`.

Image

Here is the same tree with node numbering performed on this branch (0e34b533c95556bc9ebb5fd082facfa01186d033): 

Image

And in Specify 6.8.03: 

Image

[^1]: besides perhaps conceptually not setting the `sql_safe_updates` back to the prior value, though I doubt Specify or Django has a use case where it expects it to enabled, so practically it's likely not going to result in an error 🤞

Iwantexpresso (approved):
- [x] Through the Specify Taxon Tree UI, merge two taxon nodes (e.g., 'Palaeoniscidae (in Palaeonisciformes)' into 'Palaeoniscidae (in Palaeoniscoidea)'). See that the taxon merge occurred without errors.
- [x] Try merging two other random taxon nodes in three to make sure no errors occur.

UX tests look great!

Development Testing below------------------------------


Query the database with:

select count(*)
from taxon t
join taxon p on t.parentid = p.taxonid
where t.nodenumber not between p.nodenumber and p.highestchildnodenumber;

- [x] See that it returns a count of 0

Run the renumber_tree('taxon') function directly via the Django Python shell.

- [x] The command renumber_tree('taxon') ran without errors.

Query the database with:
select count(*)
from taxon t
join taxon p on t.parentid = p.taxonid
where t.nodenumber not between p.nodenumber and p.highestchildnodenumber;

- [x] The query should return 0.

- [x] Through the Specify Taxon Tree UI, merge two taxon nodes (e.g., 'Palaeoniscidae (in Palaeonisciformes)' into 'Palaeoniscidae (in Palaeoniscoidea)'). See that the taxon merge occurred without errors.

Go back to the python shell and run the validate_tree_numbering('taxon') command

- [x] See that the validate_tree_numbering('taxon') command ran without errors

No issues found while following the Dev testing instructions either. Nice work!

Database: vpl_specify_2025_10_17
MariaDB: v11.8

specifyweb/backend/trees/extras.py
acwhite211: I changed it back to use the old path_expr and parent_joins functions.  Didn't get the old errors I used to get with new implementation, so I'm good with using the old ones.

PR #7552, opened 2025-11-19
feat(formatters): Add zero-to-one support

### LINKED ISSUE #2078: Adding display format from Locality Detail to Locality's table format causes an error
image

Screen Shot 2022-09-01 at 8 42 41 AM

image

I defined a display format for the Locality Details table and I added it to the end of my Locality table format.

image

image

db: `newmexico_august22`
version: `edge` and 6.8.01
browser: Chrome, macOS

grantfitzsimmons:
Fixes #2078

Matches the behavior present in Specify 6. This prevents the crash that used to take place when adding fields via the `geoCoordDetails` and `localityDetails` relationship fields in Locality.

Note that the preview in the dialog **will not currently** show these values, but they will be displayed in the query results.
At the moment, this is not consistent.

https://github.com/specify/specify7/issues/7570

### Checklist

### Testing instructions

image

Make sure to compare this behavior against `main`
- [ ] Create or edit the **Record Formatters** app resource for the current collection
- [ ] Add a **Locality** table format
- [ ] Add fields from `localityDetails` and `geoCoordDetails` to the format directly
- [ ] Verify that you can select all fields and relationships via these relationships
- [ ] Set this format as the default
- [ ] Edit the Locality record to include values in those fields
- [ ] Use the Query Builder and add `Locality (formatted)` to the query with the added fields from before
- [ ] Verify that the values are returned correctly and display in the query results _and_ in exported CSV files

### DIFF
--- a/.env
+++ b/.env
@@ -8,19 +8,25 @@ DATABASE_NAME=specify
 # See documenation https://discourse.specifysoftware.org/t/new-blank-database-creation-database-user-levels/3023
 
 # MASTER Database User
-# Full database administrator, used for initial setup and migrations requiring elevated privileges. 
+# Full database administrator, used for initial setup and migrations requiring elevated privileges.
+# This user should already be setup before running Specify.
 MASTER_NAME=root
 MASTER_PASSWORD=password
+MASTER_HOST=%
 
 # MIGRATOR Database User 
 # User with elevated privileges to perform migrations (create/drop/modify tables, etc.), for Django migration steps.
+# Make sure that the user is unique to just one database, otherwise use master.
 MIGRATOR_NAME=specify_migrator
 MIGRATOR_PASSWORD=specify_migrator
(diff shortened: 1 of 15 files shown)


bhumikaguptaa (approved):
Make sure to compare this behavior against `main`
- [X] Create or edit the **Record Formatters** app resource for the current collection
- [X] Add a **Locality** table format
- [X] Add fields from `localityDetails` and `geoCoordDetails` to the format directly
- [X] Verify that you can select all fields and relationships via these relationships
- [X] Set this format as the default
- [X] Edit the Locality record to include values in those fields
- [X] Use the Query Builder and add `Locality (formatted)` to the query with the added fields from before
- [X] Verify that the values are returned correctly and display in the query results _and_ in exported CSV files

I was able to compare the behavior between main and the issue branch, it works as expected in issue branch.

PR #7561, opened 2025-11-24
fix: user defined Form Definition persists across the app

### LINKED ISSUE #5168: "Copy existing view definition" does not show the Form Definition name defined by the user
**Describe the bug**
If a Form Definition is copied from default forms, the XML editor uses the name of the default forms instead of the name defined by the user. Then, when creating a new View Definition by copying an existing one, the dialog refers to the Form Definition name from the XML instead of the one defined by the user, making it difficult to identify the one you want to copy from.

**To Reproduce**

1. Go to App resources
3. Create a new Form Definition under Global Resources and copy one of the defaults (for example, "Fish"). Name it something other than the name of the one you copied
5. Switch to the XML editor and see that the "name=" value is the name of the default Form Definition you copied (for example, "Fish Views"

**Expected behavior**
When a Form Definition is created by copying a default forms, the name in the XML editor should be replaced by the one defined by the user.

**Screenshots**

https://github.com/user-attachments/assets/e3a4b988-bcd3-4af5-bcc8-4b362c51f293

Screenshot 2024-07-31 at 4 47 51 PM

[Specify 7 System Information - 2024-07-31T21_49_58.654Z.txt](https://github.com/user-attachments/files/16447488/Specify.7.System.Information.-.2024-07-31T21_49_58.654Z.txt)

- OS: Mac
- Browser: Chrome
- Specify 7 Version: 7.9.4 through edge
- Database Name: fwri_1_9_24
- Collection name: Ichthyology
- User Name: spadmin

kwhuber:
Fixes #5168 

`EditorWrapper.tsx` is where the template XML lives and eventually what the editor receives. By replacing the `<viewset name>` attribute, the editor will receive the user provided resource name before the XML reaches the editor. There are a few cases to look out for so the XML doesn’t break. For example, the following input could cause issues, emphasizing the need to replace them to XML entities: `&` resolved to `&amp;`, `<` resolved to `&lt;`, etc.  

The majority of the fix came from defining a new function, `replaceViewsetName`, inside the function, `useInitialData`. The fix makes the XML editor display the user inputted resource name (when it is provided) instead of the template name.

`replaceViewsetName` extracts the resource name from the resource parameter and replaces the first occurrence of the `<viewset name>` attribute using the following regex: https://github.com/specify/specify7/blob/74dd3d80906b927e9eae9cf8d5d69b859dfa155d/specifyweb/frontend/js_src/lib/components/AppResources/EditorWrapper.tsx#L227

`replaceViewsetName` is called everywhere that the template XML loaded. Specifically for this issue, it's called at the path where the template XML is stored: https://github.com/specify/specify7/blob/74dd3d80906b927e9eae9cf8d5d69b859dfa155d/specifyweb/frontend/js_src/lib/components/AppResources/EditorWrapper.tsx#L242-L247

Finally, the dependency array is updated to include the new resource object defined by the user: https://github.com/specify/specify7/blob/74dd3d80906b927e9eae9cf8d5d69b859dfa155d/specifyweb/frontend/js_src/lib/components/AppResources/EditorWrapper.tsx#L265

### Checklist

### Testing instructions

- [ ] Go to App Resources
- [ ] Create a new Form Definition under Global Resources and copy one of the defaults (for example, "Vertpaleo")
- [ ] Name it something other than the name of the one you copied
- [ ] Switch to the XML editor and see that the "name=" value is of the name that you defined (not of the default Form Definition you copied)

#### Example 
https://github.com/user-attachments/assets/1840de6a-5620-4420-9f87-62b275590bda

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/EditorWrapper.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/EditorWrapper.tsx
@@ -211,9 +211,27 @@ function useInitialData(
 ): string | false | undefined {
   return useAsyncState(
     React.useCallback(async () => {
+      const escapeXml = (s: string): string =>
+        s
+          .replace(/&/g, '&amp;')
+          .replace(/</g, '&lt;')
+          .replace(/>/g, '&gt;')
+          .replace(/"/g, '&quot;');
+
+      const replaceViewsetName = (data: string | null | undefined): string => {
+        const xml = data ?? '';
+        const resourceName = (resource as any)?.name ?? '';
+        if (typeof resourceName !== 'string' || resourceName.length === 0)
+          return xml;
+        return xml.replace(
+          /(<viewset\b[^>]*\bname=)(["])(.*?)\2/,
+          (_match, p1, p2) => `${p1}${p2}${escapeXml(resourceName)}${p2}`
+        );
+      };
+
       if (typeof initialDataFrom === 'number')
         return fetchResource('SpAppResourceData', initialDataFrom).then(
-          ({ data }) => data ?? ''
+          ({ data }) => replaceViewsetName(data)
         );
       else if (typeof templateFile === 'string') {
         if (templateFile.includes('..'))
@@ -224,7 +242,7 @@ function useInitialData(
           return ajax(`/static/config/${templateFile}`, {
             headers: {},
           })
-            .then(({ data }) => data ?? '')
+            .then(({ data }) => replaceViewsetName(data))
             .catch(() => '');
       }
       const subType = f.maybe(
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -33,6 +33,8 @@ services:
     build:
       context: ./.
       target: run-development
+    ports:
```
CarolineDenis: Same here. Should not be here

specifyweb/frontend/js_src/lib/components/Atoms/Form.tsx
```diff
@@ -353,9 +353,9 @@ export const Select = wrap<
        * the background in dark-mode. This is a fix:
        */
       if (props.required !== true && props.multiple === true) {
-        selected.map((option) => option.classList.add('dark:bg-neutral-100'));
+        selected.map((option) => option.classList.add('dark:bg-neutral-500')); // highlights selected object less bright
```
CarolineDenis: Why did you need these changes for this issue?

PR #7576, opened 2025-12-01
Prevent sending statistics from blocking Specify access

melton-jason:
Currently, the sending of statistics to the statistics server happens as a part of the frontend fetching the general System Information: 

https://github.com/specify/specify7/blob/9ebd35596664a05fab3f5859726ad66ae882a1de/specifyweb/frontend/js_src/lib/components/InitialContext/systemInfo.ts#L58-L112

Because there is data from the System Information that is required for the immediate functionality and behavior of some components, the endpoint was part of Specify's "Initial Context". 

All endpoints which are considered essential (i.e., must be called for normal frontend operation), are a part of this Initial Context, which blocks access to the Specify application until all endpoints have been resolved. 

https://github.com/specify/specify7/blob/9ebd35596664a05fab3f5859726ad66ae882a1de/specifyweb/frontend/js_src/lib/components/InitialContext/index.ts#L68-L91

https://github.com/specify/specify7/blob/9ebd35596664a05fab3f5859726ad66ae882a1de/specifyweb/frontend/js_src/lib/components/Core/ContextLoader.tsx#L20-L43

This caused problems when the Specify statistics server was not sending responses back to Specify clients, which resulted in the requests eventually timing out. 
The entire time the Specify frontend was waiting on the statistics server to send a response, the user would be blocked from the application. 

Because sending statistics is not vital to the application, I've implemented a "Secondary Context": a set of endpoints that Specify will call after fetching the initial context, and that will not block access to the frontend while being resolved. 
I removed the sending of statistics from the fetching of the System Information and defined a new Promise on the frontend to handle sending statistics. 

### Checklist

- [ ] Add automated tests

### Testing instructions

I tested this by artificially injecting some delay in the resolution of the new statistics endpoint.

https://github.com/user-attachments/assets/2df629f4-ac33-4857-a406-dba28f6249f0

You can use the code from the above video as a template in 1e6b59f41650b1b1f7a347347fb01bbba1b2a334. The stats endpoint should be now be called ~5 seconds after the stats promise is awaited. 
When testing interference with Initial Context, you can uncomment the stats endpoint in the list of Initial Context endpoints: 

https://github.com/specify/specify7/blob/1e6b59f41650b1b1f7a347347fb01bbba1b2a334/specifyweb/frontend/js_src/lib/components/InitialContext/index.ts#L118

- Use the [seconday-context-test](https://github.com/specify/specify7/compare/secondary-context...secondary-context-test) branch
	- Alternatively, otherwise artificially slow down the resolution of the stats endpoint
	  - This can be accomplished by modifying the code (e.g., the approach taken in my own testing), setting up Chrome Tools to throttle the stats endpoint (see something like https://www.debugbear.com/blog/chrome-devtools-throttle-individual-request), downloading an extension that supports throttling specific requests, etc. 
- Sign in to the Specify Instance
- (Optionally) Ensure request caching is disabled in your browser (this can be toggled by the `Disable cache` checkbox in Chrome Dev Tools. Request caching will be disabled as long as Dev Tools is open)
- [ ] Reload the page and ensure Specify is not affected by the slow statistics resolution

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Filters.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Filters.tsx
@@ -15,6 +15,7 @@ import { Input, Label } from '../Atoms/Form';
 import { icons } from '../Atoms/Icons';
 import { Link } from '../Atoms/Link';
 import { Dialog } from '../Molecules/Dialog';
+import { hasPermission } from '../Permissions/helpers';
 import {
   allAppResources,
   countAppResources,
@@ -23,7 +24,6 @@ import {
 } from './filtersHelpers';
 import type { AppResources } from './hooks';
 import { appResourceSubTypes, appResourceTypes } from './types';
-import { hasPermission } from '../Permissions/helpers';
 
 export function AppResourcesFilters({
   initialResources,
--- a/specifyweb/frontend/js_src/lib/components/Attachments/Plugin.tsx
+++ b/specifyweb/frontend/js_src/lib/components/Attachments/Plugin.tsx
@@ -23,11 +23,11 @@ import { loadingBar } from '../Molecules';
 import { Dialog } from '../Molecules/Dialog';
 import { FilePicker } from '../Molecules/FilePicker';
 import { ProtectedTable } from '../Permissions/PermissionDenied';
+import { collectionPreferences } from '../Preferences/collectionPreferences';
 import { userPreferences } from '../Preferences/userPreferences';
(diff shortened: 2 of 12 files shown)


acwhite211 (approved):

grantfitzsimmons:
@melton-jason Do you want me to test this? Are the instructions still up-to-date?

grantfitzsimmons (approved):
Specify is no longer impacted by the slow statistics resolution, even when severely throttled! 😄

PR #7578, opened 2025-12-02
Batch Workbench Uploads

### LINKED ISSUE #7577: WB upload speedup needed
A speedup to the workbench upload is needed for large datasets.  This is especially needed in order to speedup  the process of data conversions.  A possible solution might be to try batching the database operations for uploading each of the workbench rows into actual database records.  This would decrease the amount of database request are made by Django and also hopefully be more efficient on the database side as well.  This solution also applies to speeding up the workbench validation.

The goal is for the workbench validation and uploads to complete in a few minutes, as opposed to the hours that is can take on large datasets.

acwhite211:
Fixes #7577

Edit the workbench upload code to batch rows together used in workbench uploads and validation.  This is intended to speedup workbench uploads and validations.  Will need to continue tuning the batch size that seems to be optimal for workbench uploads.  Also, still working on adjusting the progress bar update for the batch upload code.

A good test case that has been causing problems with slow uploads is this one here: https://drive.google.com/file/d/1Mpr_KWMkCY74_yZv_knXiNGeG6TSKCYk/view?usp=drive_link
There are many fields in this file, here is a data mapping I made for testing purposes: https://drive.google.com/file/d/1eo56GKwGbMXV7luGD_SJ24b-ADxFb53X/view?usp=drive_link

You can generate large mineral data sets (which can be easily adapted to other disciplines) using this script: https://github.com/specify/data-management/blob/main/demo-data-generator/generate.py

You can set the range here:
https://github.com/specify/data-management/blob/57959833499a7f838a4ff28444c932ac4c966288/demo-data-generator/generate.py#L50

Generate it just by running:

```
 python3 meteorite-generate.py
```

### Checklist

### Testing instructions

Initial Testing:
- Run `docker compose down` and `docker compose up` to try to keep the memory allocation low before testing
- [ ] Run the validation process on workbench data that you know takes a decent amount of time to validate.  Run the validation in this branch and in the main branch for comparison.
- [ ] See if there was a speed up in the validation time compared to the main branch.
- [ ] See if the validation results looks the same as the validation results in the main branch

Further Testing:
- [ ] Run workbench validation on a large workbench dataset, see that it finishes within a few minutes.
- [ ] Run workbench upload on a large workbench dataset, see that it finishes within a few minutes.  Note only run upload on the workbench dataset that doesn't have any validation issues.
- [ ] Aim for at least 1000 rows per minute on average for the validation process.  Goal is for around 5000 rows per minute.


acwhite211:
Anyone who has been doing conversions or working with a large workbench validations that takes a while to run, go ahead and try running the validation on this branch and see if the results are the same and if there is a speed up.  If you can, put the workbench data on google drive so any issues can be recreated and debugged.  Thanks.

Don't plan on merges this branch for 7.12.  It doesn't seem to work with batch edit.  So, we'll likely just use this branch internally when needed.

combs-a (changes requested):
Initial Testing:
- [X] Run the validation process on workbench data that you know takes a decent amount of time to validate.  Run the validation in this branch and in the main branch for comparison.
- [X] See if there was a speed up in the validation time compared to the main branch.
- [ ] See if the validation results looks the same as the validation results in the main branch

<hr>
I haven't gotten a proper comparison or anything just yet, but did experience this error when uploading localities:

image

Maybe it's not running the business rule that sets SrcLatLongUnit to 0 when it isn't present anymore; issue is not present on main. Grant linked [the locality business rules](https://github.com/specify/specify7/blob/main/specifyweb/backend/businessrules/rules/locality_rules.py), so you can check the exact business rule.

Seems like both of them were around the same amount of time validating as well, since it was about 30 minutes for the test branch, and estimated the same for main.

It took around 20 minutes in main and the test branch.

Record set used was a set of about 52000 localities, can check with a collection object set as well. If the dataset needs to be bigger for testing, let me know--this seemed big enough!

acwhite211:
Made a change to the locality model that avoids the integrity error, on the Locality SrcLatLongUnit field, from occurring.  After running the locality upload again, the 52,193 rows were validated in 10 minutes.  So the validation ran about 5000 rows per minute.

image

acwhite211:
After updating this branch to main, and running the same validation again, the speed of the validation slowed down.  I'm investigating what exactly caused this slow down.

acwhite211:
This provided some speed up to the workbench batch upload, but broke some other workbench functionality, like batch edit.  This PR wasn't never meant to be merged, so it should stay in it's own branch.  I'm actually going to close this PR, since there are a lot of merge conflicts now from other updates to the workbench, like #8057.

PR #7620, opened 2026-01-07
Fix locked geo migration by specifying db alias

### LINKED ISSUE #7601: Lock wait timeout exceeded when running Geo migration
**Describe the bug**
Currently on `main` (`v7.12.0` development: https://github.com/specify/specify7/commit/e7123b6feab07e21d794bd1a2890c601de56b3e1), databases created from the Specify 6 wizard receive a `django.db.utils.OperationalError: (1205, 'Lock wait timeout exceeded; try restarting transaction')` error when running the [0002_geo](https://github.com/specify/specify7/blob/main/specifyweb/specify/migrations/0002_geo.py) migration. 

Below is a full traceback:
[traceback.txt](https://github.com/user-attachments/files/24396823/traceback.txt)

The traceback states the error happens at the following lines: 

https://github.com/specify/specify7/blob/e7123b6feab07e21d794bd1a2890c601de56b3e1/specifyweb/specify/migrations/0002_geo.py#L79

https://github.com/specify/specify7/blob/e7123b6feab07e21d794bd1a2890c601de56b3e1/specifyweb/specify/migration_utils/default_cots.py#L71-L79

This error does not occur on the same database when running migrations in earlier versions (e.g., `v7.11.3`).

**To Reproduce**

1. Have or create a database made using the Specify 6 Wizard which hasn't had the [0002_geo](https://github.com/specify/specify7/blob/main/specifyweb/specify/migrations/0002_geo.py) migration executed
  a. If needed, provided below is a database created from the Specify Wizard
3. Start up a dockerized Specify instance (or run migrations on a local Specify instance) on `main` or any version sufficiently past `v7.11.3`
4.  Inspect the containers logs and eventually observe the error

Here is a database created from the Specify Wizard with a single collection in a Botany discipline:

[from_wiz.sql](https://github.com/user-attachments/files/24396864/from_wiz.sql)

- OS: Apple M1 macOS Sonoma (14.3). 
- Specify 7 Version: `v7.12.0` development / versions after `v7.11.3`

acwhite211:
Fixes #7601

Specify the database alias in the geo migration functions to avoid the possible issue of tables locking and timing out.  Want to avoid a situation where a lock wait caused by the migration code using two different DB connections/aliases inside one atomic migration.  When running the setup tool branch on a new database created in Specify 6, migration errors in the geo migrations should no longer occur.

### Checklist

### Testing instructions

- [x] Run the geo migration on a new database created in Specify 6.  See that the migration runs and completes without errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Consortium web site is: https://specifysoftware.org
 
 Specify 7 Copyright © 2024 Specify Collections Consortium. Specify comes with
 ABSOLUTELY NO WARRANTY. This is free software licensed under GNU General Public
-License 2 (GPL2).
+License 3 (GPL3).
 
     Specify Collections Consortium
     Biodiversity Institute
--- a/specifyweb/backend/context/app_resource.py
+++ b/specifyweb/backend/context/app_resource.py
@@ -36,6 +36,19 @@ DISCIPLINE_DIRS = {
     "invertebrate": "invertebrate",
     "geology": "geology",
 }
+DISCIPLINE_NAMES = {
+    'fish': 'Ichthyology',
+    'herpetology': 'Herpetology',
+    'paleobotany': 'Paleobotany',
+    'invertpaleo': 'Invertebrate Paleontology',
+    'vertpaleo': 'Vertebrate Paleontology',
+    'bird': 'Ornithology',
+    'mammal': 'Mammalogy',
+    'insect': 'Entomology',
+    'botany': 'Botany',
+    'invertebrate': 'Invertebrate Zoology',
+    'geology': 'Geology',
+}
 
 FORM_RESOURCE_EXCLUDED_LST = [
     "fish/fishbase.views.xml",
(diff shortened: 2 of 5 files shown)


melton-jason (commented):
This did resolve my specific issue, although I wonder if this Issue could happen again in the future or on slower setups where we don't manually specify the correct database connection to use in a migration. 

Django actually supports automatic database routing using their [DATABASE_ROUTERS](https://docs.djangoproject.com/en/4.2/topics/db/multi-db/#automatic-database-routing) setting. 

Could we leverage their system for database routing to make this more robust? 
Django has built-in support for routing only for migrations. It would be very nice to automatically use the `migrations` connection for migrations: 

https://github.com/specify/specify7/blob/53d91efa64969a0d48804bb43622b6765fd0a837/specifyweb/settings/__init__.py#L63-L72

We just need to define a class that Django would use behind the scenes. See  https://docs.djangoproject.com/en/4.2/topics/db/multi-db/#allow_migrate.

Something like: 

```py
class SpecifyMigrationRouter:
    def allow_migrate(self, db, app_label, model_name=None, **hints):
        return db == "migrations"
```

Then add the following setting in [specifyweb/specify/settings/\_\_init\_\_.py](https://github.com/specify/specify7/blob/main/specifyweb/settings/__init__.py):

```py
DATABASE_ROUTERS = ["specifyweb.path.to.file.SpecifyMigrationRouter"]
```

With the above example, Django would automatically only use the `migrations` connection for migrations. 

---

Alternatively, when we run migrations on startup: 

https://github.com/specify/specify7/blob/53d91efa64969a0d48804bb43622b6765fd0a837/sp7_db_setup_check.sh#L425-L435

We can always specify the `migrations` connection when calling `manage.py migrate`

PR #7635, opened 2026-01-11
fix: deleting records in query builder has a minimum of zero results

### LINKED ISSUE #4988: Querying and deleting all results makes the results show as -1
**Describe the bug**
When making a query if you delete all the records that show up the number of results shows as -1 

**To Reproduce**

1. Create a query (or use existing one) 
2. Delete all the records that show up 
3. See results shows -1

**Expected behavior**
If there's no longer any results the count should show as 0 not -1

**Screenshots**

https://github.com/specify/specify7/assets/150940840/be8b8f8c-b54d-482c-a543-e45ea8cf8c0b

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge
- Collection name: mammals
- User Name: spadmin

**Additional context**
Possibly related to #4538 

https://github.com/specify/specify7/assets/150940840/b8994939-9b45-42f8-bd5d-d7e3240971af

[Specify 7 Crash Report - 2024-06-05T16_20_53.981Z.txt](https://github.com/user-attachments/files/15593545/Specify.7.Crash.Report.-.2024-06-05T16_20_53.981Z.txt)

kwhuber:
Fixes #4988

This PR introduces a change to fix a ui issue when a user deletes records through the query builder. Before the change, after deleting the last record in a specific query, the results header would incorrectly return as `Results: (-1)`. After the change, the results header will only show `Results: (0)` when the last record is deleted. 

This change is in the following line:  https://github.com/specify/specify7/blob/81f9974d7663271fe6c7fcd1c1295a52b7cbb5d6/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx#L172 

This prevents a negative account from showing up in the results header. 

### Checklist

- [ ] Add automated tests

### Testing instructions

- [ ] Create a query (or use existing one)
- [ ] Delete all the records that show up
- [ ] Verify that the results count is zero and non-negative

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/Results.tsx
@@ -169,7 +169,8 @@ export function QueryResults(props: QueryResultsProps): JSX.Element {
       setResults(newResults(results));
       if (removeCount === 0) return;
       setTotalCount((totalCount) =>
-        totalCount === undefined ? undefined : totalCount - removeCount
+        totalCount === undefined ? undefined : Math.max(0, totalCount - removeCount)
+      
       );
       const newSelectedRows = (selectedRows: ReadonlySet<number>) =>
         new Set(Array.from(selectedRows).filter((id) => id !== recordId));


emenslin (approved):
- [x] Create a query (or use existing one)
- [x] Delete all the records that show up
- [x] Verify that the results count is zero and non-negative

Looks good! 

https://github.com/user-attachments/assets/3f7ba525-1a28-4563-90e5-0f0d2629c496

grantfitzsimmons (commented):
In my testing, I had a query that returned two results (Collection Object). I deleted one. The count changed from `2` to `0`, and now when I select a result, I saw `1/0`.

https://github.com/user-attachments/assets/bdc360bf-9be2-4535-8640-eea2d7c4a431

Does this set a floor so the value shows `0`? Yes, but it does not address the underlying problem, which I have opened a new issue to represent here: https://github.com/specify/specify7/issues/7639

PR #7679, opened 2026-02-02
feat(queries): Add 'Select All' and 'Invert Selection' buttons to the Query Builder

### LINKED ISSUE #7485: Add "Select All" button for query results
From @danb213 at CSIRO in https://github.com/specify/specify7/discussions/5624#discussioncomment-11632743:
> Something that might help is to auto-tick the results that appear on first load of the query results, this would make it easier for users to either click "deselect all" to select a few results, or manually remove any outliers. In addition it would help clarify the process - after our UAT a common comment was on how to select all results, even though the system treats the list as being fully selected until at least one row is ticked. With the addition of a permanent "select all" and "deselect all" perhaps this behaviour needs to be changed?
>
>I think it would be ok if the "select all" and "invert" only appeared for up to x rows for performance reasons, that should satisfy our users.

Requested again by Corinna at CSIRO in https://github.com/specify/specify7/discussions/5624#discussioncomment-11632745, and Silvia at CSIC in https://github.com/specify/specify7/discussions/5624#discussioncomment-13411263:

>Specify users at RJB are missing the "select all" button for query results that was available in Specify6.
>They told me that, as use case, with a large number of results, selecting all but a few (15 or 25 out of hundreds or thousands) is difficult because they have to select all the records, scrolling down to the last one, and then deselect the few ones.

From @maxpatiiuk in https://github.com/specify/specify7/discussions/5624#discussioncomment-11632742:

> Those seemingly simple features are actually quite hard to implement, as not all query results are loaded in the browser.
> 
> I.e, if query returns 10k rows, the front-end can't fetch all 10k because that would take a long time.
> 
> Also, it is extremely impractical to find specific few rows you want to unselect among the 10k results.
> 
> See https://github.com/specify/specify7/issues/1406 for a longer explanation.
> 
> Potential workarrounds:
> 1. Add "Select All" and "Invert Selection", but have them only appear when query has fewer than x result rows.
> 2. Do "Select All" and "Invert Selection" on the back-end (quite complicated)
> 
> Instead, most use cases for these features can be satisfied like this:
> 1. You can quickly select several rows by selecting first row, and then shift-clicking on the other row, and all rows in between would be selected.
> 2. You can adjust the query filters to exclude the rows you don't want to select.

This was not implemented intentionally in the past, but I am confident we can find a solution that loads segments of the results gradually:
> We did that intentionally.
> There are several reasons:
> - If a query returns a million records, and you want to uncheck a few, it would have to load all million records first, which would be SLOW. The current solution allows you to select only records that are visible, thus ensuring that they are loaded already.
> - If you have many query results, it is unlikely you would want to uncheck a handful of them, because a better solution is to just adjust your filter conditions. Instead, you would want to uncheck a handful of results happen if you have only ~30 results, which is not hard to do under current design: check first checkbox, scroll to the end, and shift click on the last checbox

## Overview
This feature aims to enhance the query results UI by adding "Select All" and "Invert Selection" buttons that are always visible and functional, similar to functionality that was available in Specify6. Currently, the system treats the list as fully selected by default until at least one row is manually selected, which is causing confusion among users.

### UI Requirements
1. Add a "Select All" button that is always visible in the query results view
2. Keep the existing "Deselect All" button that appears when records are selected
3. Group these buttons in a logical controls area above the query results table

### Functional Requirements
1. "Select All" button:
   - When clicked, selects all visible records in the current query result
   - Should provide visual feedback (checkbox checked for all rows)
   - For performance reasons, may be limited to operate on a maximum number of records

2. Behavior changes:
   - Change the default behavior so query results are NOT treated as fully selected by default
   - Make selection state visually explicit (show checked checkboxes when items are selected)
   - Ensure record count indicators clearly show total vs. selected records

3. Performance considerations:
   - Make sure there is throttling for selection operations on large result sets
   - Possibly add an option to set maximum number of records that can be bulk-selected
   - Consider lazy-loading techniques for rendering large result sets

## Testing Requirements
1. Test with small, medium, and large result sets (10, 100, 1,000, 100,000+ records)
2. Verify performance with maximum supported result set size
3. Test all selection operations: select all, deselect all
4. Verify that downstream operations (export, create record set, etc.) work correctly with explicitly selected records

## Acceptance Criteria
1. Users can see and use "Select All" buttons at all times in query results
2. Selecting all records in a large result set completes in a reasonable time (<10 seconds)
3. Visual feedback clearly indicates which records are selected
4. All downstream operations work correctly with the new selection model
5. Performance remains acceptable for result sets of all sizes

## User Experience Flow
1. User runs a query that returns results
2. Initial state shows all records with unchecked checkboxes
3. User can click "Select All" to select all records at once
4. User can then manually deselect specific records they want to exclude
6. Record count indicator shows number of selected records out of total

**Requested By:** CSIRO, CSIC, and many others

### Discussed in https://github.com/specify/specify7/discussions/5624

<sup>Originally posted by **grantfitzsimmons** August 25, 2022</sup>
> When a user selects some records from the query list a "deselect all" button appears. However it would be very useful to also have options to "select all" and "invert selection" buttons that are always available

**Reported By:** **[CSIRO](https://app.asana.com/0/1201972579963460/1202844554923815/f)**</div>

Iwantexpresso:
Fixes #7485 
This Pr adds 2 new buttons to query builder results, select all and Invert Selection. These only appear after a query has been saved and have also been given a limit to only appear on queries with < 3 million results.

Sample demonstration 

https://github.com/user-attachments/assets/757d3377-b352-45f2-9e76-40d431664577

### Checklist

- [ ] Add automated tests

### Testing instructions
- open any database of your choice, go to query builder,create a query or open a saved one that displays results. If using a new query remember to save it as only then the button will be available.
- [ ] click the newly added select all button, see if all results are selected
- clear your selections, next we will be testing the invert all button
- [ ] Select some records, then click invert selection, make sure all records but the ones previously checked are selected
- [ ] Select no records,  then click invert selection, make sure the invert selection button is properly inverting the current selection. 
- [ ] Perform these check on at least one large database ( queries with more than 50,000 or 100,000+ results), also keep track of any latency longer than 10 seconds when using both invert selection and select all. 
- [ ] (optional) perform this test on a database with at least 500,000+ query results), this is an optional stress test so feel free to use a larger DB if you're able to


emenslin (changes requested):
- [ ] click the newly added select all button, see if all results are selected
- [ ] Select some records, then click invert selection, make sure all records but the ones previously checked are selected
- [ ] Select no records,  then click invert selection, make sure the invert selection button is properly inverting the current selection. 
- [ ] Perform these check on at least one large database ( queries with more than 50,000 or 100,000+ results), also keep track of any latency longer than 10 seconds when using both invert selection and select all. 
- [ ] (optional) perform this test on a database with at least 500,000+ query results), this is an optional stress test so feel free to use a larger DB if you're able to

---
Looks wise everything seems to be selected; however, I am running into this issue sometimes where it shows all results are selected but when I create a record set, not all records are added. I also checked the behavior against main (same query and selected no results) and all results are added. This does not seem to always happen but I can consistently recreate it with this query: https://ojsmnh20251211-issue-7485.test.specifysystems.org/specify/query/305/ 

Selecting all records individually:
 
https://github.com/user-attachments/assets/8090b7b4-abd4-4827-979c-961a33f69ab3


Using 'Select All' button: 

https://github.com/user-attachments/assets/d1a3b413-3979-4464-84a0-5f298e67c18d


I also ran into it when creating a CSV for that query and others. It seems to consistently cut off at either 80 or 120 records depending on the query, I'm not sure why. Here's another example query from a different DB: https://herbrbge20260105-issue-7485.test.specifysystems.org/specify/query/554/

PR #7684, opened 2026-02-04
fix: removed add button from `collectionPreferences` of the `SpAppResource` table 

### LINKED ISSUE #4950: Add button in collection preferences gives a 404 error
**Describe the bug**
Pressing add in collection preferences gives a 404 error 

**To Reproduce**

1. Go to app resources
2. Click on collection preferences
3. Press add

**Expected behavior**
It should not give a 404 error 

**Screenshots**

https://github.com/specify/specify7/assets/150940840/15a1e3d3-c706-4fb0-8440-731b63a6188a

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: edge
- Collection name: mammals
- User Name: spadmin

kwhuber:
Fixes #4950

This fix removes the add button from the `collectionPreferences` App Resource. In the UI, this is noticeable when you navigate to User Tools --> App Resources --> Collection Preferences. At the bottom right of the screen, only a save button appears now as the add button did not serve a necessary purpose.

The change comes by hiding the add button for the specific Collection Preferences App Resource (while for other App Resources it will be unaffected) in `Save.tsx`:
https://github.com/specify/specify7/blob/0c26e26f9303649ba51e9a9bc01910296990985c/specifyweb/frontend/js_src/lib/components/Forms/Save.tsx#L410

### Checklist

- [ ] Add automated tests

### Testing instructions

- [ ] Go to app resources, click on collection preferences, and verify that the add button is absent
- [ ] Verify that other app resources still have add button functionality


kwhuber:
As an aside, I have a comment about a couple minor UI bugs regarding the save button within collection preferences depending on where it's accessed. 

For example, going through the route of User Tools --> App Resources --> Collection Preferences, the save button begins as grayed out. After making edits but then returning to default state, it reappears as its beginning grayed out state. It makes more sense to have this be a functional close button. 

https://github.com/user-attachments/assets/5e476f3f-270c-436f-b181-1bd3fd288d89

Going straight to User Tools --> Collection Preferences, there is a functional close button as default. This makes sense! However, if I alter the page by selecting and unselecting options, I can never return to the beginning state close button its manipulated back to its default state. 

https://github.com/user-attachments/assets/5a5ad650-8c7a-407f-9945-f5dfd8da711e

emenslin (approved):
- [x] Go to app resources, click on collection preferences, and verify that the add button is absent

Looks good! My only note is that this applies to all app resources not just collection preferences, this seems intentional based on what you said at the beginning of the PR; however, since the testing instructions only mention checking collection preferences I wanted to make note of it here. 


Also just so you know I did look at your videos and I was not able to recreate the issue when going through app resources, and I honestly don't recall ever running into it (although I could be wrong) so there might be something else effecting that behavior. 

https://github.com/user-attachments/assets/73166028-0f1f-43dc-8ca8-56218744ca30

combs-a (approved):
- [X] Go to app resources, click on collection preferences, and verify that the add button is absent
- [X] Verify that other app resources still have add button functionality 

<hr>

It is disabled on Collection Preferences. 👍 

Not within in the scope of this PR since this does not cause an error when pressed, but wanted to note: or other things like Default Preferences and User Preferences, the Add button is still present. User Preferences also does not have an Add button if you navigate to it directly through the User Tools menu, so I wanted to note that it was inconsistent. I'll split it off into a separate QoL issue.

image

bhumikaguptaa (approved):
- [X] Go to app resources, click on collection preferences, and verify that the add button is absent
- [X] Verify that other app resources still have the add button functionality 

Works as expected. The add button is disabled in Collection Preferences.

emenslin (approved):
- [x] Go to app resources, click on collection preferences, and verify that the add button is absent
- [x] Verify that other app resources still have add button functionality 

---

>Wouldn't it make more sense if this was a close button (like in my second video above) instead when nothing has yet been altered? Then, if you do unclick boxes and then return to the state in which you entered the form (default), it lets you resave (but there is nothing to resave).
> 
> I think there should be a change made here to recognize that nothing has actually changed and the button should return to how it was when you entered the form.
> 
> Does that make sense?

I understand what you're saying now, and although I agree that it is likely out of the scope for this PR, I think it could be a good feature request! Some things you might want to consider are, if there is a close button instead of the current disabled save button, where will it take you to when you press it? And as for the second part with how it lets you resave even though no change has been made, this is everywhere in Specify (at least to my knowledge), do you think it could be confusing to add the behavior just here, or would your goal be to change it everywhere? You don't necessarily have to know the answer to any of those but just some things to think about if you decide to go that route.

Additionally, I actually liked how the add button was initially removed from all the app resources instead of just being removed from collection preferences. Although it is not necessary to do so, since collection preferences was the main problem with the 404 error, it might be something to consider doing in the future.

PR #7696, opened 2026-02-12
Safe implementation of stats request

### LINKED ISSUE #7693: New stats client request functionality
Re-implement the Specify web client request to the stats lambda function such that a request is made no more than once a day, as to avoid sending too many requests.  Each instance doesn't need to send a stats request more than once a day.  The request parameters should support the needed aggregated stats fields of collection count, user count, and collection object count.

We need to make sure with this new implementation that the stats request is hold up any of the other Specify functionality.  Also, any errors that occur from stats request should not prevent anything else from functioning.

Here is the PR that implemented the new stats request https://github.com/specify/specify7/pull/7415
Here is the PR that avoided issues of slowing down the client with the stats request https://github.com/specify/specify7/pull/7576
This issue needs to implement the new stats request by avoiding the original issues.

We discussed moving the stats request to the back-end, but decided it is sufficient to use local storage in the browser to reference and respect a timestamp value of the last request.

acwhite211:
Fixes #7693

This PR hardens front-end stats reporting so lambda telemetry cannot slow down app behavior and is not sent more than once every 24 hours per collection.  The web client stats request to the lambda could run too often and could also hold up other front-end work if the request was slow.   For this stats request, each instance only needs to send once per day, with the aggregated stats fields:
- collection count
- user count
- collection object count

We also need to make sure stats reporting does not impact normal Specify behavior. If the stats request fails or runs long, it should not block anything else.  Here's the front-end stats flow:
1. The lambda stats request is sent no more than once every 24 hours, using a localStorage timestamp key scoped by collection.
2. The lambda request runs in the background, such that it does not block other front-end operations.
3. A client timeout is applied, so that the request is cut off if it runs too long.
4. Errors are handled silently and do not affect other app functionality.
5. Stats parameters include collection count, user count, and collection object count.

We discussed moving stats submission to the back-end, but decided using a browser localStorage timestamp for gating is sufficient for this use case and keeps the implementation straightforward.

Added the lambda function for the stats aggregator to the data management repo https://github.com/specify/data-management/blob/main/lambda_functions/sp7_stats_aggregator.py

The new stats server has been setup at https://sp7-stats.specifycloud.org/

I also figured out how to get CNAME records to work between dreamhost and aws lambda functions so that the stats lambda function can have a permanent specify cloud url. stats-2.specifycloud.org

### Checklist

### Testing instructions

- [ ] Open the Specify main page.  In the network tab, see that the api request to the sp7-stats.specifycloud.org/capture endpoint was successful.
- [ ] Try refreshing the page and then immediately try doing other tasks, like opening a tree.  Make sure that the app doesn't expereince any irregular slow down due to slow/hanging stats requests.

### DIFF
--- a/specifyweb/settings/specify_settings.py
+++ b/specifyweb/settings/specify_settings.py
@@ -122,7 +122,7 @@ SUPPORT_LOGIN_TTL = 180
 
 # Usage stats are transmitted to the following address.
 # Set to None to disable.
-STATS_URL = "https://stats.specifycloud.org/capture"
+STATS_URL = "https://sp7-stats.specifycloud.org/capture"
 STATS_2_URL = "https://stats-2.specifycloud.org"
 
 # Workbench uploader log directory.


grantfitzsimmons (approved):
UX testing only:

- [X] Open the Specify main page.  In the network tab, see that the api request to the sp7-stats.specifycloud.org/capture endpoint was successful.
- [X] Try refreshing the page and then immediately try doing other tasks, like opening a tree.  Make sure that the app doesn't experience any irregular slowdown due to slow/hanging stats requests.

👍 It's not stuck on a failed network request.

PR #7700, opened 2026-02-13
fix(workbench): Match Agent records without assuming 'Person'

### LINKED ISSUE #3723: Matching behavior in the WorkBench assumes agents are "Person"
It looks like the WorkBench cannot match an agent that has an agent type other than "Person" based on fields from that table unless you specify the Agent Type in the upload.

Here's the record for `et al.`:

image

Here's what happens when I try to match it during a WorkBench upload:

image

To get it to properly recognize that it already exists, I had to specify the Agent Type "Group" for an agent that has a type of "Group" instead of "Person".

image

It looks like it assumes that all agents are people and matches with that assumption in mind.

image

## Resolution
Stop assuming that agent type is "Person" and match based on fields alone. 

I spoke to @melton-jason about this issue.

grantfitzsimmons:
Fixes #3723

Makes it so that Agent matching can be done for Agents of **any** type. Previously, the WorkBench would only look for records where `AgentType` was 'Person' (`1`) for matching purposes. This remedies that.

**This PR:**
image
 
**Current Behavior:**
image

### Testing instructions

Compare the behavior against the latest tagged release and the branch this is based on.
- [ ] Create a WorkBench data set that includes 'Agent' links that match to existing records, where the agent is of any other type (not 'Person')
- [ ] Add rows that match to 'Person' agents
- [ ] Add rows that do not match to 'Person' agents
- [ ] Add rows that create new agents of **each** type
- [ ] Test agent matching via complicated relationships (e.g. collectors, preparators, or other one-to-many links)
- [ ] Test agent matching via direct links (e.g. cataloger, agent1, agent2, etc.)
- [ ] Verify that matching works as expected!
- [ ] Verify that newly created records, when `AgentType` is not specified, are added as 'Person' agents

### DIFF
--- a/.env
+++ b/.env
@@ -43,6 +43,9 @@ ASSET_SERVER_URL=http://host.docker.internal/web_asset_store.xml
 # Make sure to set the `ASSET_SERVER_KEY` to a unique value
 ASSET_SERVER_KEY=your_asset_server_access_key
 
+# Information to connect to a Redis database
+# Specify will use this database as a process broker and storage for temporary
+# values
 REDIS_HOST=redis
 REDIS_PORT=6379
 REDIS_DB_INDEX=0
--- a/specifyweb/backend/businessrules/rules/attachment_rules.py
+++ b/specifyweb/backend/businessrules/rules/attachment_rules.py
@@ -30,8 +30,8 @@ def attachment_jointable_save(sender, obj):
 
     attachee = get_attachee(obj)
     obj.attachment.tableid = attachee.specify_model.tableId
-    scopetype, scope = Scoping(attachee)()
-    obj.attachment.scopetype, obj.attachment.scopeid = scopetype, scope.id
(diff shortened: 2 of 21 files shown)


bhumikaguptaa (approved):
- [x] Create a WorkBench data set that includes 'Agent' links that match to existing records, where the agent is of any other type (not 'Person')
- [x] Add rows that match to 'Person' agents
- [x] Add rows that do not match to 'Person' agents
- [x] Add rows that create new agents of **each** type
- [x] Test agent matching via complicated relationships (e.g. collectors, preparators, or other one-to-many links)
- [x] Test agent matching via direct links (e.g. cataloger, agent1, agent2, etc.)
- [x] Verify that matching works as expected!
- [x] Verify that newly created records, when `AgentType` is not specified, are added as 'Person' agents

Works as expected.

PR #7798, opened 2026-03-09
Fix stuck WorkBench attachment import

### LINKED ISSUE #7794: Cannot import attachments in WB
**Describe the bug**
Stuck on uploading after pressing the import attachments button in workbench. Normal workbench uploads work fine and you can upload attachments through the forms.

Edit: Batch Attachment Upload also gets stuck on the uploading dialog. You can select attachments and validate, but when you press upload it gets stuck on the dialog. 

**To Reproduce**

1. Go to workbench
2. Click on import attachments
3. Select attachments to import
4. See it gets stuck on uploading 

**Expected behavior**
Should upload normally like in 7.11.4

**Screenshots**

https://github.com/user-attachments/assets/41018b60-1d39-44bf-aa78-91ca16a3d1f8

- OS: Windows 11
- Browser: Chrome
- Specify 7 Version: 7.12
- Database Name: blankdbsetup2elizabeth20260302
- Collection name: geo
- User Name: spadmin
- URL: https://blankdbsetup2elizabeth20260302-main.test.specifysystems.org/specify/

acwhite211:
Fixes #7794

Identified and fixed two issues with the WB upload attachments.

The main issue was that the Upload promise could hang forever.  `uploadFile()` used `XMLHttpRequest` but only listened to `readystatechange`. On network errors, the promise could remain unresolved and the loading UI never closed.  This was solved by adding error, abort, and timeout handlers, set an upload timeout, and clean up listeners so the promise always settles.

There was another issue with an invalid React hook usage in the legacy WorkBench import-attachments flow.  `collectionPreferences.use(...)` was called inside a plain helper function 'uploadFiles, not a component/hook. That can throw before upload starts and leave the modal stuck. There was a simple solution to move the hook call into 'FilesPicked' and pass the resolved value into `uploadFiles(...)`.

Looking at the network requests, the POST request to the asset server now occur and are successful.

The hook fix in 'WbImportAttachments/index.tsx' removes an invalid React hook call and keeps behavior the same, except it no longer can fail/hang from that mistake.

The XHR fix in 'Attachments/attachments.ts' ensures the upload should complete or fail, never hang forever.

The only new behavior change is the new upload timeout. A very slow upload that takes longer than the timeout will now fail instead of waiting indefinitely.

---

After some follow up, it seems that the batch attachment upload is also having a similar issue as the WB attachment upload.  The problem is that when uploading in batch attachments, the loading is hanging as well.

Found an issue where the `collectionPreferences.use(...)` function was being called within `uploadFileWrapped(...)`.  This was causing an issue

The solution was to move the hook call into the component scope and pass the value down as a plain argument.

### Checklist

### Testing instructions

- Go to workbench
- Click on import attachments
- Select attachments to import
- [x] See it doesn't get stuck on uploading, and that it continues to the 'Select a Base Table with Attachments' dialog.
- [x] See that POST network request were made to the asset server, and that they were successful.
- Go to attachments and click "Import" and create a new batch attachment import.
- Select attachments to import and select a field, like catalog number for example.
- Click on "Save" and then "Upload"
- [x] See that the upload loading bar completes and doesn't stall.

image

image

image

### DIFF
--- a/specifyweb/backend/setup_tool/api.py
+++ b/specifyweb/backend/setup_tool/api.py
@@ -37,6 +37,7 @@ from specifyweb.backend.setup_tool.task_tracking import (
 from specifyweb.specify.models import Institution, Discipline
 from specifyweb.backend.businessrules.uniqueness_rules import apply_default_uniqueness_rules
 from specifyweb.specify.management.commands.run_key_migration_functions import fix_cots, fix_schema_config
+from specifyweb.specify.models_utils.model_extras import PALEO_DISCIPLINES, GEOLOGY_DISCIPLINES
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


acwhite211:
I think this solution solves the main issue and is good enough.  I added some extra logging on the front-end for asset connection errors.  Let me know if you want to add some further abstraction to the code related to the asset server connect from the front-end.

PR #7908, opened 2026-04-09
feat(queries): improve query density and add column resizing

### LINKED ISSUE #7542: Improve query density and add column resizing support
We need to improve readability and scan speed in query results by defaulting to compact, single-line rows while allowing optional wrapping and user-resizable columns.

### Functional Requirements
1. Results grid must support manual column resizing.
2. Results grid must use compact density by default, showing one visual row per result.
3. Text wrapping must be controlled by a user preference.
4. Wrapping preference must default to Off.
5. When wrapping is Off:
	1. Each cell must render only the first line of text.
	2. Embedded line breaks must not increase row height.
6. Column auto-sizing must occur when a result set is loaded.
7. Auto-size width for each column must be based on the larger of:
	1. The rendered header label width.
	2. The rendered widest cell value in the currently loaded results.
8. Column widths must be clamped to configurable minimum and maximum values.
9. If total column width exceeds viewport width, horizontal scrolling must be available.
10. Manual resize must override auto-size for that column until:
	1. A new query result set is loaded, or
	2. The user chooses reset-to-auto-size.
11. Full cell text must remain accessible when truncated via tooltip

Image

**Requested by:** User Advisory Committee, many users throughout the years

grantfitzsimmons:
Fixes #7542

This PR improves the Query Builder results table. The goal is to make query results easier to scan and less disruptive when users interact with column widths (largely matching the behavior in Specify 6):
- Increase density of information returned in the query by default.
- No longer wrap query result cells by default (added new preference to re-enable if desired).
- Keep initial column sizing readable and space-filling.
- Allow manual resizing of columns.

## Current
image

## This PR
image

### Implementation

- Added clearer width constraints for auto-size vs manual resize:
  - Auto-size remains bounded for readability.
  - Manual resize uses its own minimum width and does not apply the auto-size cap.
- Updated grid column sizing behavior:
  - Before manual resize: columns can expand to fill available width.
  - After manual resize starts: columns use fixed pixel widths so drag behavior is stable and predictable.
- Prevented width resets on result refresh by limiting when auto-size reset runs.
- Removed the Reset widths button from the Results toolbar.
- Improved cell overflow handling so wrapped content remains constrained to its own column and non-wrap mode uses single-line clipping behavior.
- Kept sortable headers and resize handles working with the updated layout.

### Testing instructions

Test these steps with both the "Line wrap" preference enabled and disabled where not specified:
image

- [ ] Open Query Builder and run a query with multiple visible columns (including long text and numeric fields).
- [ ] Confirm initial render auto-sizes columns and fills available table width.
- [ ] Resize one or more columns narrower and wider using the header resize handles.
- [ ] Confirm manual resizing is stable.
- [ ] Confirm long values stay constrained to their own column (not clipping into adjacent columns).
- [ ] In non-wrap mode, confirm cells remain single-line and clipped when narrow.
- [ ] In wrap mode, confirm content wraps within the column boundaries.
- [ ] Sort by several columns and confirm sorting still toggles correctly (column sizes may reset).


emenslin (approved):
- [x] Open Query Builder and run a query with multiple visible columns (including long text and numeric fields).
- [x] Confirm initial render auto-sizes columns and fills available table width.
- [x] Resize one or more columns narrower and wider using the header resize handles.
- [x] Confirm manual resizing is stable.
- [x] Confirm long values stay constrained to their own column (not clipping into adjacent columns).
- [x] In non-wrap mode, confirm cells remain single-line and clipped when narrow.
- [x] In wrap mode, confirm content wraps within the column boundaries.
- [x] Sort by several columns and confirm sorting still toggles correctly (column sizes may reset).

Looks good, I didn't run into any issues! The only thing I noticed is that the headings have ellipses when they are cut off, but the results themselves don't. I don't think this is needed so I will approve, however, it is something we should consider. 

https://github.com/user-attachments/assets/088676d8-50ab-486b-9e80-90208059314c

Iwantexpresso (approved):
- [x] Open Query Builder and run a query with multiple visible columns (including long text and numeric fields).
- [x] Confirm initial render auto-sizes columns and fills available table width.
- [x] Resize one or more columns narrower and wider using the header resize handles.
- [x] Confirm manual resizing is stable.
- [x] Confirm long values stay constrained to their own column (not clipping into adjacent columns).
- [x] In non-wrap mode, confirm cells remain single-line and clipped when narrow.
- [x] In wrap mode, confirm content wraps within the column boundaries.
- [x] Sort by several columns and confirm sorting still toggles correctly (column sizes may reset).

nice I have ran into no issues so far!,
although I do have a comment, the separation lines on the header row are not visible, since you cannot resize from any other row  This confused me a bit. At first, I thought  I should be able to resize the columns by grabbing the line divisor on any subsequent row, not the header row because the division/resizing lines there are not visible. This isn't that big of a deal on its own, It just lacks a little clarity when communicating this functionality now exists, as well as where its meant to be interacted with to a user. 

Image

kwhuber:
- [x] Open Query Builder and run a query with multiple visible columns (including long text and numeric fields).
- [x] Confirm initial render auto-sizes columns and fills available table width.
- [x] Resize one or more columns narrower and wider using the header resize handles.
- [x] Confirm manual resizing is stable.
- [x] Confirm long values stay constrained to their own column (not clipping into adjacent columns).
- [x] In non-wrap mode, confirm cells remain single-line and clipped when narrow.
- [ ] In wrap mode, confirm content wraps within the column boundaries.
  - Do numbers apply to this? If so, this is not working quite right (please see attached video). 
- [x] Sort by several columns and confirm sorting still toggles correctly (column sizes may reset).

https://github.com/user-attachments/assets/92044a2c-d44d-47ba-9331-521357f219f3

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
```diff
@@ -266,26 +294,37 @@ function Cell({
...
+      ? renderedValue.split(/\r?\n/u)[0]
+      : renderedValue;
+
   return (
     <div
       className={`
         ${getCellClassName(condenseQueryResults)}
         ${value === null ? 'text-gray-700 dark:text-neutral-500' : ''}
-        ${fieldSpec?.parser.type === 'number' ? 'justify-end tabular-nums' : ''}
+        ${fieldSpec?.parser.type === 'number' ? 'tabular-nums' : ''}
```
alesan99, line 324: I do think numbers should be wrapped too, however. I think this is the line that's not allowing that. Maybe there's a way to preserve the number-specific formatting and also allow wrapping?

alesan99 (approved):
- [x] Open Query Builder and run a query with multiple visible columns (including long text and numeric fields).
- [x] Confirm initial render auto-sizes columns and fills available table width.
- [x] Resize one or more columns narrower and wider using the header resize handles.
- [x] Confirm manual resizing is stable.
- [ ] Confirm long values stay constrained to their own column (not clipping into adjacent columns).
- [x] In non-wrap mode, confirm cells remain single-line and clipped when narrow.
- [x] In wrap mode, confirm content wraps within the column boundaries.
- [x] Sort by several columns and confirm sorting still toggles correctly (column sizes may reset).
---
Everything works at least as well as main 👍I think the header should allow double clicking to fit to the column title, and there should be lines to show the handles, but functionality wise its okay.

bhumikaguptaa (approved):
- [X] Open Query Builder and run a query with multiple visible columns (including long text and numeric fields).
- [X] Confirm initial render auto-sizes columns and fills available table width.
- [X] Resize one or more columns narrower and wider using the header resize handles.
- [X] Confirm manual resizing is stable.
- [X] Confirm long values stay constrained to their own column (not clipping into adjacent columns).
- [X] In non-wrap mode, confirm cells remain single-line and clipped when narrow.
- [X] In wrap mode, confirm content wraps within the column boundaries.
- [X] Sort by several columns and confirm sorting still toggles correctly (column sizes may reset).
--
Everything works as expected! I also noticed that when line wrap is enabled, numbers were not getting wrapped as the column width got smaller.

specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsTable.tsx
grantfitzsimmons: This is sorted in the latest patch

PR #7926, opened 2026-04-09
Sync column headers with data on query field reorder

### LINKED ISSUE #7086: Column headers mismatch when reordering the Query fields
**Describe the bug**

When querying and then reordering the display columns in the results view, the column headers become mismatched, displaying the wrong header for the data in the column.

**To Reproduce**

1. Go to CO in Queries
2. Add timestamp created and catalog number as mappings and check the series box
3. Query and see the results, now change the display(column) order and run the query

**Expected behavior**

After reordering the columns, each column header should accurately reflect the data displayed beneath it.

**Screenshots**

https://github.com/user-attachments/assets/48f571a6-bc5e-4eba-b4fa-9e21ee8b9c39

- OS: Windows 10
- Browser: Chrome
- Specify 7 Version: 7.11.0.2
- Database Name: KUFish_2025_02_14
- User Name: spfishadmin
- URL: https://kufish20250214-v711.test.specifysystems.org/specify/query/new/collectionobject/

**Additional context**
Same bug is recreated with several columns headers as well and only when the series box is checked

foozleface:
Fixes #7086
Contributed by @foozleface

When reordering fields in the query builder, the column headers in the results table get out of sync with the actual data columns. This happens specifically in "series" (smushed) mode, where the backend moves the catalog number column to position 1 in the SQL SELECT but the frontend field specs remain in the user-defined order.

### Implementation
- Added `reorderFieldSpecsForSeries()` which, when series mode is active on a CollectionObject query, moves the catalog number entry to the front of the fieldSpecs/queryFields arrays to match the backend column order.
- Applied the reordering in `useQueryResultsWrapper()` after building field specs but before rendering results.
- Fixed the sort handler to use `queryFields` (the reordered array) instead of `displayedFields` (the original order) when resolving field indices.
- Added unit tests covering no-op cases (series off, wrong table, catNum already first, catNum absent) and the actual reorder case.

### Testing instructions
- [ ] Create a CollectionObject query with catalog number and other fields
- [ ] Reorder the fields so catalog number is not first
- [ ] Enable "series" (smushed) mode and run the query
- [ ] Verify column headers match their data columns
- [ ] Run `npx jest --testPathPattern=ResultsWrapper`

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsWrapper.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryBuilder/ResultsWrapper.tsx
@@ -22,11 +22,49 @@ import {
   queryFieldsToFieldSpecs,
   unParseQueryFields,
 } from './helpers';
+import type { QueryFieldSpec } from './fieldSpec';
 import type { QueryResultRow } from './Results';
 import { QueryResults } from './Results';
 
 // TODO: [FEATURE] allow customizing this and other constants as make sense
 const fetchSize = 40;
+
+/**
+ * When series mode (smushed) is active, the backend skips the catalog number
+ * column from its normal position in the SQL SELECT and instead inserts a
+ * catalog number series range at position 1 (right after the ID column).
+ *
+ * This function reorders the frontend fieldSpecs/queryFields arrays so that
+ * the catalog number entry is at index 0 (matching position 1 in the result
+ * data, since position 0 is the ID which is stripped before rendering).
+ *
+ * Without this reordering, column headers become mismatched with data columns.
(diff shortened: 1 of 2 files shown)


Iwantexpresso (approved):
- [x] Create a CollectionObject query with catalog number and other fields
- [x] Reorder the fields so catalog number is not first
- [x] Enable "series" (smushed) mode and run the query
- [x] Verify column headers match their data columns
- [x] Run `npx jest --testPathPattern=ResultsWrapper`

Reordering when selecting series seems to be working as expected

Image

took me a little bit to set up the proper jest dependancies but tests seem to be doing good!
Image

PR #7939, opened 2026-04-09
Paginate delete blockers to prevent OOM

### LINKED ISSUE #7515: Linked Records button causing massive server slowdown
**Describe the bug**
Opening a high-level record with a lot of linked records can cause Specify worker to crash.
Our (perhaps underpowered) test server will run out of memory (and hit high-cpu) trying to generate the Linked records query to run in the database. The database itself doesn't have issues running the query to get the linked records, but (our current running theory) generating the `WHERE IN ()` for the Linked Records button maxes out the Specify worker and crashes our entire server. 

**To Reproduce**
1. Go to any of the trees at a high level (i.e. Life, Earth)
2. Click to edit the record
3. The linked records button will begin to automatically generate the results for the Linked records button
4. Doing so causes the server to run out of memory until it SIGKILLs the process. 

**Expected behavior**
I can't think of an expected behavior (other than not taking down the server), but maybe there should be a warning or limit to the number of records fetched when running the linked records query, or an optimisation to the backend query generation so it can handle larger records. Alternatively not running the linked query every time a record is opened but only when chosen would avoid accidentally clicked open on a high-level record and causing the issue. 

**Screenshots**
Opening this record from the Geography tree broke our test server for ~30 minutes fetching the Linked Records. The actual query on the database ran quickly but creating it used all available memory. I am aware that it's probably not best to have >900k locality records but that's the way our instance is set up. We've noticed significant slowdowns on anything with >50k linked records, and crashes at >200k. 

Image

**Crash Report**
There was no specific error message from Specify for this (other than a timeout message at one stage. 
There logs for the container itself show that it runs out of memory and SIGKILLs:

Image
System Information:

[Specify 7 System Information - 2025-10-28T05_47_03.613Z.txt](https://github.com/user-attachments/files/23179745/Specify.7.System.Information.-.2025-10-28T05_47_03.613Z.txt)

- OS: Windows
- Browser: Chrome
- Specify 7 Version: 7.11.2.1
- Database Name: specify_staging 
- Collection name: QLD Herbarium
- User Name: running on my own account [leo]

**Reported By**
Queensland Herbarium 

**Additional context**
Our test server is running RHEL 9, with the specify docker composition set up in Podman. 

Let me know if there's any other information that would be helpful.

foozleface:
Fixes #7515
Contributed by @foozleface

The `delete_blockers` endpoint used Django's `Collector.collect()` which loads ALL related objects into memory before returning. For a Taxon like "Life" with 900K+ determinations, this causes out-of-memory kills. This PR replaces the Collector with targeted COUNT queries and limited ID lists per PROTECT relationship, keeping memory bounded regardless of how many blocking rows exist.

### Implementation
- Replace `Collector.collect()` in `_collect_delete_blockers` with iteration over `obj._meta.related_objects`, filtering for `protect_with_blockers` on_delete
- For each PROTECT relationship, run a COUNT query and fetch at most `id_limit` (default 100) primary keys
- Add `total_count` field to the response so the frontend knows the full count even when IDs are capped
- Update frontend test fixtures to match the new `isstrict: undefined` serialization
- Add comprehensive backend tests for the pagination behavior

### Testing instructions
- [ ] Navigate to a record with many dependent records (e.g., a high-level Taxon node) and attempt to delete it
- [ ] Verify the delete blocker dialog shows the correct count and a reasonable number of linked record IDs
- [ ] Confirm no OOM or timeout on records with very large numbers of blockers
- [ ] Run the test suite: `python manage.py test specifyweb.backend.delete_blockers.tests.test_pagination`


acwhite211:
Move to this PR with additional fixes https://github.com/specify/specify7/pull/8018

PR #7941, opened 2026-04-09
Improve tree search efficiency

### LINKED ISSUE #7752: [Large Databases]: Make tree searches by name more efficient
For collections with exceptionally large taxonomy trees (e.g., ~1 million rows), the taxon search box severely degrades server performance. Every keystroke sends a query using a leading wildcard (`LIKE '%pattern%'`), which prevents standard index usage and forces a full table scan. In environments hosting multiple collections, a single user typing a scientific name rapidly queues up multiple 5-second scans. 

```sql
SELECT taxon.TaxonID, IFNULL(taxon.FullName, '') FROM taxon WHERE taxon.TaxonTreeDefID IN (1) AND taxon.FullName LIKE '%hydrolea%' ORDER BY IFNULL(taxon.FullName, '') LIMIT 0, 1000
```

This consumes disk I/O and starves other collections on the same server, potentially resulting in `502 Gateway Timeouts` for other users.

**Proposal**
To improve performance and reduce server load, searches should fetch a limited number of results per query (for example, limiting the initial fetch to 25 or 50 records instead of 1000) and paginate the remaining results when scrolling. Fetching smaller chunks will reduce the I/O tax per keystroke while still allowing users to find the correct name.

Several other strategies could be employed either alongside or instead of pagination to mitigate this issue:
1. Implement a strict debounce on the input field, waiting 500ms after the user stops typing before firing the query to the backend.
2. Prevent the query from firing until the user has typed at least 3 or 4 characters.
3. Migrate to using full text indexes and `MATCH() AGAINST()` for substring searches, which can utilize indexes for substring matches and resolve queries in milliseconds instead of seconds.

**Reported By**
Tomislav, UT Austin on [BoldDesk in ticket #616](https://specifysoftware.bolddesk.com/agent/tickets/616)

**Additional context**
This was identified via slow query logs where the taxon search in a botany collection (PRC) accounted for 72% of all slow queries on the server, causing cascading 502 errors for a vertebrate paleo collection (VPL) sharing the same MariaDB instance.

foozleface:
Fixes #7752
Contributed by @foozleface

Tree search in QueryComboBox sends a LIKE query per keystroke. On tree tables with 200K+ rows, using "contains" mode generates `LIKE '%pattern%'` which cannot use B-tree indexes and causes full table scans. This PR changes the default search mode from "contains" to "startsWith", enabling `LIKE 'pattern%'` which uses B-tree indexes. It also reduces the search result limit from 1000 to 50, since a typeahead dropdown never needs that many results.

### Implementation
- Change the default value of `treeSearchAlgorithm` user preference from `contains` to `startsWith` in `UserDefinitions.tsx`
- Export `QUERY_COMBO_BOX_SEARCH_LIMIT` constant set to 50 (was 1000)
- Add tests verifying the default search operator and result limit

### Testing instructions
- [ ] Open a form with a tree-based QueryComboBox (e.g., Taxon field on a Determination)
- [ ] Type a few characters and verify the typeahead dropdown populates quickly
- [ ] Verify that typing "rosa" matches "Rosaceae" and "Rosales" but not "Pterosaurus" (startsWith behavior)
- [ ] Check user preferences -- the tree search algorithm should now default to "starts with"
- [ ] Run the frontend tests: `npx jest --testPathPattern treeSearchEfficiency`

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryComboBox/__tests__/treeSearchEfficiency.test.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryComboBox/__tests__/treeSearchEfficiency.test.tsx
@@ -1,0 +1,79 @@
+/**
+ * Tests for tree search efficiency improvements (#7752).
+ *
+ * Tree tables (Taxon, Geography, Storage, etc.) can have 200K+ rows.
+ * Using LIKE '%pattern%' causes full table scans; LIKE 'pattern%' can use
+ * B-tree indexes. The typeahead also doesn't need 1000 results — 50 is
+ * more than enough for a dropdown.
+ */
+import { requireContext } from '../../../tests/helpers';
+import { serializeResource } from '../../DataModel/serializers';
+import { tables } from '../../DataModel/tables';
+import { queryFieldFilterSpecs } from '../../QueryBuilder/FieldFilterSpec';
+import { makeComboBoxQuery } from '../helpers';
+import { QUERY_COMBO_BOX_PAGE_SIZE } from '../index';
+
+requireContext();
+
+describe('Tree search efficiency', () => {
+  test('makeComboBoxQuery for tree tables uses startsWith by default', () => {
+    const query = makeComboBoxQuery({
+      fieldName: 'fullName',
+      value: 'rosa',
+      isTreeTable: true,
+      typeSearch: {
+        table: tables.Taxon,
+        searchFields: [[tables.Taxon.strictGetLiteralField('fullName')]],
+        name: 'Taxon',
+        title: 'Taxon',
+        formatter: '',
+        displayFields: undefined,
+        format: '%s',
+      },
+      specialConditions: [],
+    });
+
+    const serialized = serializeResource(query);
+    const searchField = serialized.fields.find(
+      (field: { readonly isDisplay: boolean }) => !field.isDisplay
+    );
+
+    // startsWith operator (id 15), NOT like (id 0)
+    expect(searchField?.operStart).toBe(queryFieldFilterSpecs.startsWith.id);
+    // Value should NOT be wrapped in % wildcards
+    expect(searchField?.startValue).toBe('rosa');
+  });
+
+  test('makeComboBoxQuery for non-tree tables uses startsWith by default', () => {
+    const query = makeComboBoxQuery({
+      fieldName: 'lastName',
+      value: 'smith',
+      isTreeTable: false,
+      typeSearch: {
+        table: tables.Agent,
+        searchFields: [[tables.Agent.strictGetLiteralField('lastName')]],
+        name: 'Agent',
+        title: 'Agent',
+        formatter: '',
+        displayFields: undefined,
+        format: '%s',
+      },
+      specialConditions: [],
+    });
+
+    const serialized = serializeResource(query);
+    const searchField = serialized.fields.find(
+      (field: { readonly isDisplay: boolean }) => !field.isDisplay
+    );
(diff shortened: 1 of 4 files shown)


specifyweb/frontend/js_src/lib/components/QueryComboBox/__tests__/treeSearchEfficiency.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+ * Using LIKE '%pattern%' causes full table scans; LIKE 'pattern%' can use
+ * B-tree indexes. The typeahead also doesn't need 1000 results — 50 is
+ * more than enough for a dropdown.
+ */
+import { requireContext } from '../../../tests/helpers';
+import { serializeResource } from '../../DataModel/serializers';
+import { tables } from '../../DataModel/tables';
+import { queryFieldFilterSpecs } from '../../QueryBuilder/FieldFilterSpec';
+import { makeComboBoxQuery } from '../helpers';
+import { QUERY_COMBO_BOX_SEARCH_LIMIT } from '../index';
```
grantfitzsimmons: This isn't included in this PR

grantfitzsimmons (changes requested):
> ### Testing instructions
> - [X] Open a form with a tree-based QueryComboBox (e.g., Taxon field on a Determination)
> - [X] Type a few characters and verify the typeahead dropdown populates quickly
> - [X] Check user preferences -- the tree search algorithm should now default to "starts with"
> - [X] Run the frontend tests: `npx jest --testPathPattern treeSearchEfficiency`

This is pretty simple, just changing the default preference, but it does speed up searches. The automatic test is not passing, however.

foozleface:
Fixed — `index.tsx` was missing from the PR. Added:

- `export const QUERY_COMBO_BOX_SEARCH_LIMIT = 50` (the constant the test imports)
- `limit: QUERY_COMBO_BOX_SEARCH_LIMIT` replacing the hardcoded `limit: 1000`

All 3 tests pass locally:
```
Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
```

specifyweb/frontend/js_src/lib/components/QueryComboBox/index.tsx
```diff
@@ -55,6 +55,10 @@ import { useTreeData } from './useTreeData';
 import { TreeDefinitionContext } from './useTreeData';
 import { useTypeSearch } from './useTypeSearch';
 
+// Typeahead dropdown doesn't need 1000 results — 50 is more than enough.
+// Reducing this from 1000 also cuts the DB query cost significantly.
+export const QUERY_COMBO_BOX_SEARCH_LIMIT = 50;
```
grantfitzsimmons: I'd prefer that we fetch query combo box results in batches on scroll rather than limiting it to 50

foozleface:
Updated per feedback — replaced the hard limit of 50 with **batch-on-scroll pagination**:

- Initial fetch returns first 50 results (`QUERY_COMBO_BOX_PAGE_SIZE`)
- Scrolling near the bottom of the dropdown fetches the next 50
- "Loading..." indicator shown while fetching
- Continues until all results are loaded

The `AutoComplete` component gets three new optional props (`onScrollEnd`, `isLoadingMore`, `extraItems`) — backward-compatible, existing callers unaffected.

All 12 tests across AutoComplete + QueryComboBox suites pass.

acwhite211:
Moved to this PR with additional fixes https://github.com/specify/specify7/pull/8016

PR #7942, opened 2026-04-09
Warn before editing shared records in QueryComboBox

### LINKED ISSUE #597: Increase limit to number of reports that can be displayed
If a database has a lot of reports, Specify only displays some in a truncated list. Newer reports do not appear on this list.

foozleface:
Fixes #597
Contributed by @foozleface

When Carry Forward copies foreign keys, multiple Collection Objects end up sharing the same related record (e.g., the same Locality via their Collecting Event). Clicking the pencil (edit) icon on a QueryComboBox field silently mutates the shared record, changing data for all referencing records without any warning. This PR adds a warning dialog that appears before editing non-dependent related records, showing how many records share the reference and offering three choices.

### Implementation
- Add a shared record check that queries the API for the count of records referencing the same related record
- Show a warning dialog with the record count, a link to inspect referencing records, and three action buttons:
  - **Cancel** -- close the dialog, do nothing
  - **Edit Shared** -- proceed to edit the shared record (existing behavior)
  - **Clone and Edit** -- create a copy of the record and edit the clone instead (safe default)
- Implement session memory via sessionStorage so users can choose "remember for this session" to skip repeated warnings
- Add localization strings for all dialog text (sharedRecordWarning, sharedRecordWarningDescription, editShared, cloneAndEdit)
- Generic implementation that works for all table types, not just Locality
- Add tests verifying the localization strings exist and include dynamic table name/count values

### Testing instructions
- [ ] Enable Carry Forward on a Collection Object form
- [ ] Create several COs using Carry Forward (they will share the same Collecting Event/Locality)
- [ ] On one of the COs, click the edit (pencil) icon on the Locality QueryComboBox field
- [ ] Verify a warning dialog appears showing how many records share this Locality
- [ ] Test all three buttons: Cancel (dialog closes), Edit Shared (opens editor), Clone and Edit (creates new record)
- [ ] Check "remember for this session" and verify subsequent edits skip the dialog
- [ ] Close and reopen the browser tab -- verify the session preference is cleared
- [ ] Run the frontend tests: `npx jest --testPathPattern sharedRecordWarning`

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/QueryComboBox/__tests__/sharedRecordWarning.test.tsx
+++ b/specifyweb/frontend/js_src/lib/components/QueryComboBox/__tests__/sharedRecordWarning.test.tsx
@@ -1,0 +1,59 @@
+/**
+ * Tests for shared record edit warning (#597).
+ *
+ * When using Carry Forward, Specify copies foreign keys — so multiple
+ * Collection Objects share the same Locality via their Collecting Event.
+ * Clicking the pencil (edit) icon on a non-dependent QueryComboBox field
+ * mutates the shared record, silently changing data for all referencing
+ * records.
+ *
+ * The fix shows a warning dialog before editing non-dependent related
+ * records, offering Clone and Edit as the primary safe action.
+ */
+
+import { requireContext } from '../../../tests/helpers';
+import { formsText } from '../../../localization/forms';
+
+requireContext();
+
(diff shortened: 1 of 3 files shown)


foozleface:
Superseded by #8030, which is the same change rebased onto current `main` and pushed to a branch on this repo so CI and Docker workflows can run. Closing this one.

PR #7509, opened 2025-10-21
Handle Disambiguation Case for Host Taxon in QB Results

acwhite211:
Fixes #7436

On the front-end, allow for the disambiguation between taxon and host taxon columns in the QB results.  Implement this as a special case in the `generateMappingPathPreview` function on the front-end.  This will show the column `CollectingEvent->CollectingEventAttribute->HostTaxon->Genius->FullName` to be displayed as `Host - Genus - Full name`, rather than `Genus - Full name`.  This allows the user to differentiate between other paths from Taxon.

This resolves the specific case described in the issue.  If we want a more general solution that can resolve any disambiguation between fields in the QB results, then we'll need to design a more complex solution on the front-end.  That solution would need to be different from `generateMappingPathPreview`, as it would need to have the context of all the fields in the query, and then have some way of deciding on alternative names when disambiguation as needed.

The issue on the back-end is with the query that gets generated:
```sql
SELECT
	collectionobject.`CollectionObjectID`,
	CASE
		WHEN (taxon_1.`TaxonTreeDefItemID` = 64) THEN taxon_1.`Name`
		WHEN (taxon_2.`TaxonTreeDefItemID` = 64) THEN taxon_2.`Name`
		WHEN (taxon_3.`TaxonTreeDefItemID` = 64) THEN taxon_3.`Name`
		WHEN (taxon_4.`TaxonTreeDefItemID` = 64) THEN taxon_4.`Name`
		WHEN (taxon_5.`TaxonTreeDefItemID` = 64) THEN taxon_5.`Name`
		WHEN (taxon_6.`TaxonTreeDefItemID` = 64) THEN taxon_6.`Name`
		WHEN (taxon_7.`TaxonTreeDefItemID` = 64) THEN taxon_7.`Name`
		WHEN (taxon_8.`TaxonTreeDefItemID` = 64) THEN taxon_8.`Name`
		WHEN (taxon_9.`TaxonTreeDefItemID` = 64) THEN taxon_9.`Name`
		WHEN (taxon_10.`TaxonTreeDefItemID` = 64) THEN taxon_10.`Name`
		WHEN (taxon_11.`TaxonTreeDefItemID` = 64) THEN taxon_11.`Name`
		WHEN (taxon_12.`TaxonTreeDefItemID` = 64) THEN taxon_12.`Name`
		WHEN (taxon_13.`TaxonTreeDefItemID` = 64) THEN taxon_13.`Name`
[...]

Issue #7436, opened 2025-09-22
Query Results Show Incorrect Data and Ambiguous Headers for Fields with the Same Name


bhumikaguptaa:
### Discussed in https://github.com/specify/specify7/discussions/7435

<div type='discussions-op-text'>

<sup>Originally posted by **bhumikaguptaa** September 22, 2025</sup>
**Description of issue**

When a query is constructed to display output fields from different paths that share the same name (e.g., FullName), the query results grid displays incorrect data and uses ambiguous, non-unique column headers. The grid appears to populate all columns with the same name using the data from the first column encountered, rather than the data from their respective paths.

This causes two major issues:
- The data presented in the results is factually incorrect.
- The user cannot distinguish between the columns due to the identical headers.

**Example Use Case**
A user queries a Collection Object record where:
The determination taxon is Mesocestoides lineatus.
The host taxon is Vulpes vulpes.

**Expected Behavior**

The results grid should display two distinct columns with unique, descriptive headers and the correct corresponding data. For instance:
Column 1 Header: Determinations: Species - Full name | Data: lineatus
Column 2 Header: Host taxon: Species - Full name | Data: vulpes

**Actual Behavior**

The results grid displays two columns with identical headers, and both show the data from the first path.
Column 1 Header: Species - Full name | Data: lineatus
Column 2 Header: Species - Full name | Data: lineatus (This is incorrect)

**Solution**

The system should generate unique column aliases for the results view, perhaps by prepending a parent table or a portion of the field path to the column header. This would resolve both the data mapping error and the user-facing ambiguity.

**Reported by** 

Alexis Beck, Le Muséum d’histoire naturelle de Genève
Link to the Bold desk ticket: https://specifysoftware.bolddesk.com/agent/tickets/197</div>

grantfitzsimmons:
From Alexis Beck at NHMG via email:
>I just wanted to report you that I was hoping two important fixes would be included in the release.
>
>Querying different taxon genus/species in the same query (eg. I want the current det + preferred taxon + host taxon). Only one of them is displayed in the query result

specifysoftware:
This issue has been mentioned on **Specify Community Forum**. There might be relevant details there:

https://discourse.specifysoftware.org/t/chrono-strat-and-chrono-strat-end-returning-the-same-value-in-queries/2923/5

PR #7978, opened 2026-04-15
Avoid MariaDB join explosion for disambiguated taxon QB results

### LINKED ISSUE #7977: New additional joins in tree queries overextending MariaDB
From @melton-jason 
> Pretty sure it was this PR that added the extra joins:
> https://github.com/specify/specify7/pull/7509
> 
> I've attached the query being run from `aumnh_inverts` both before the changes in the PR are applied and after.
> 
> Specifically, in the old query we were constructing something like the following:
> ```sql
> LEFT OUTER JOIN taxon AS taxon_18 ON taxon_18.`TaxonID` = determination_1.`PreferredTaxonID`
> ```
> 
> With the new query, it looks like we are duplicating most of the Taxon hierarchy chain to also include a PreferredTaxonID join against Taxon, resulting in something like the following:
> 
> ```sql
> LEFT OUTER JOIN taxon AS taxon_18 ON taxon_18.`TaxonID` = determination_1.`PreferredTaxonID`
> ...
> LEFT OUTER JOIN taxon AS taxon_34 ON taxon_33.`ParentID` = taxon_34.`TaxonID`
> ```
> 
> So in this case, we're essentially including a copy of the whole tree hierarchy to resolve a PreferredTaxon relationship.
> 
> In other words, when you now map a Determinations > PreferredTaxon (or any relationship to PreferredTaxon) in a query, the resulting SQL query is now including an additional copy of the entire tree hierarchy starting from the PreferredTaxon.
> i..e, we're building the parent chain for PreferredTaxon relationships and we were not in v7.11.4.
> 
> [new_query.sql](https://github.com/user-attachments/files/26752428/new_query.sql)
> [old_query.sql](https://github.com/user-attachments/files/26752431/old_query.sql)

acwhite211:
Fixes #7977

The change from #7509 fixed the QB disambiguation problem from #7436 by making taxon-related tree-rank fields resolve against the correct join path, but on some databases it also caused the generated SQL to exceed MariaDB's 61 table join limit.

The fix keeps the disambiguation behavior from #7436, but limits the SQL change to taxon tree rank fields. Instead of building a full parent-chain join in the main query for every taxon path, the main query now joins only to the starting taxon alias, such as Taxon, PreferredTaxon, or HostTaxon. Any requested taxon rank is then resolved in a correlated subquery rooted at that specific alias. This preserves correct disambiguation between taxon-related columns while avoiding the MariaDB “too many tables in a join” failure.

### Query Comparison

**old_query.sql:**
The main query had one full taxon ancestry chain off determination.TaxonID, and PreferredTaxonID was only joined directly.
This kept the join count lower, but it could resolve disambiguated taxon fields against the wrong taxon path.

**new_query.sql from the merged PR:**
The main query built separate full ancestry chains for both TaxonID and PreferredTaxonID.
This fixed the disambiguation problem, but it increased the outer join graph from 50 joins / 18 taxon aliases to 66 joins / 34 taxon aliases, which is what triggered the MariaDB limit on some databases.

**Initial changes in this PR:**
The main query still keeps separate starting aliases for the taxon paths, so the disambiguation fix remains.
The difference is that the parent-chain walk for taxon ranks is moved into correlated subqueries, so those extra taxon ancestry chains are no longer part of the main query’s outer join graph.
The query stays correct, but the expensive taxon tree expansion is split into separate SELECTs instead of one oversized join.

**Final changes in this PR:**:
The final version keeps the same disambiguated starting aliases, but resolves each requested taxon rank with a scalar correlated subquery that walks only the alias's own parent chain. So, instead of building a reusable taxon rank lookup for the whole tree, each taxon rank column now follows just the current Taxon or PreferredTaxon lineage and returns the matching rank value from that lineage.

This keeps the #7436 disambiguation fix, avoids the oversized outer-join graph that caused the MariaDB 61 table failure, and also avoids the slower full tree lookup plan from the initial draft of this PR. On the database that reproduced #7977, the problematic label query now completes successfully instead of timing out.

### Checklist

### Testing instructions

- Run the label generating query that caused a MariaDB 'Too many table joins' error.
- [x] See that the label now gets generated without any errors in the logs.

### DIFF
--- a/specifyweb/backend/stored_queries/query_construct.py
+++ b/specifyweb/backend/stored_queries/query_construct.py
@@ -45,36 +45,30 @@ class QueryConstruct(namedtuple('QueryConstruct', 'collection objectformatter qu
         table,
         next_join_path,
         treedefitem_params,
-        max_depth,
     ):
         field = next_join_path[0]
         treedefitem_column = table.name + 'TreeDefItemID'
+        treedef_column = table.name + 'TreeDefID'
         correlation_target = sa_inspect(start_alias).selectable
+        start_node_number = getattr(start_alias, "nodeNumber")
 
-        ancestors = [orm.aliased(mapped_cls)]
-        for _ in range(max_depth - 1):
-            ancestors.append(orm.aliased(mapped_cls))
-
-        cases = []
-        for ancestor in ancestors:
-            orm_field = getattr(ancestor, field.name)
-            for treedefitem_param in treedefitem_params:
-                cases.append(
-                    (
-                        getattr(ancestor, treedefitem_column) == treedefitem_param,
-                        orm_field,
-                    )
-                )
-
-        subquery = sql.select(sql.case(cases)).select_from(ancestors[0])
-        for parent, ancestor in zip(ancestors, ancestors[1:]):
-            subquery = subquery.outerjoin(ancestor, parent.ParentID == ancestor._id)
-
-        column = (
-            subquery.where(ancestors[0]._id == start_alias._id)
-            .correlate(correlation_target)
-            .scalar_subquery()
+        ancestor = orm.aliased(mapped_cls)
+        subquery = (
+            sql.select(getattr(ancestor, field.name))
+            .select_from(ancestor)
+            .where(
+                getattr(ancestor, treedefitem_column).in_(tuple(treedefitem_params)),
+                getattr(ancestor, treedef_column) == getattr(start_alias, treedef_column),
+                start_node_number.between(
+                    getattr(ancestor, "nodeNumber"),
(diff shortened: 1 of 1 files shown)


grantfitzsimmons (approved):
- [x]  See that the label now gets generated without any errors in the logs.

grantfitzsimmons:
@specify/ux-testing This can be tested on a copy of `pricollections` (like `pricollections_2026_04_15` on the test panel) using query ID `97` (Label Query) owned by the user `leslie`

Iwantexpresso (approved):
- [x]  See that the label now gets generated without any errors in the logs.

nice! I saw no errors on the logs nice work !

PR #7981, opened 2026-04-15
Revert Back-end changes from Handle Disambiguation Case for Host Taxo…

acwhite211:
Revert the back-end changes from the #7509 PR in order to avoid the error of SQL exceeding MariaDB's 61-table join limit.

### Checklist

### Testing instructions

- Run the label generating query that caused a MariaDB 'Too many table joins' error.
- [x] See that the label now gets generated without any errors in the logs.


melton-jason (approved):
Looks good! 👌 
Reverts the backend changes in #7509 and resolves the extra joins for Taxon relationships.

Issue #824, opened 2021-05-12
Modify search to find similar characters


tlammer:
RBGE requested the ability for the search to bring back not only exactly what is searched for, but also similar characters.  For example, searching on Dong Thap would bring back  Đồng Tháp.  Or, searching for Jose would bring back José as well as Jose.

maxxxxxdlp:
The fix for this may be as simple as adding `COLLATE "utf8mb4_general_ci"` to MySQL statements

Though, I am not sure if it would be best to add it during table creation (would affect all fields everywhere)
or for all text and string fields (would affect wb's disambiguation, among other things)
or add it query wide only when running the express search and query builder's query.

@benanhalt do you have any suggestions?

maxxxxxdlp:
Also, it seems like searching for `Đồng Tháp` when I have an agent with that last name does not return any matches. So besides collation, there may be an issue with encoding.

maxxxxxdlp:
Possible library for this - https://fusejs.io/ 

Though, it also depends on whether we want to solve this on the front-end or back-end

Issue #1596, opened 2022-05-14
"Create" permission for app resource tables is not useful


maxxxxxdlp:
Because of the way AppResource editor is working, when user creates new resource, it is saved before the editor opens.
This means that users with "Create" permission to the app resource tables, but no "Update" permission can only create empty app resources

maxxxxxdlp:
Fixed in 97b258751

To Test: a user with only `Create` and `View` permission to `Tools -> Resources` should be able to create a new resource

grantfitzsimmons:
<img width="1275" alt="image" src="https://user-images.githubusercontent.com/37256050/192839922-4ebc9afe-636e-4e82-af58-24321e72ffb1.png">

With just those two permissions, I annot access the App Resources option in the User Tools menu

_Directly navigating to https://freshfish-testability.test.specifysystems.org/specify/resources/ gives the following:_

<img width="1277" alt="image" src="https://user-images.githubusercontent.com/37256050/192840579-5d6f5c3e-4c8d-4662-bdba-8a6e9eeeac65.png">

maxxxxxdlp:
My comment was unclear. Create a user that has only `Create` and `View` permissions for `Tools -> Resources`, but does not have `Modify` or `Delete`.

Also give the user a `Table -> All Tables -> Read`

chanulee1:
User with only `Create` and `View` permissions can create a new app resource that is not empty.
Tested:
db: `freshfish`
user: `appTest`

PR #7985, opened 2026-04-16
fix(AppResources): default SpAppResource.description to name on creation

foozleface:
## Summary

The "New Resource" flow in `AppResources/Create.tsx` constructs an `SpAppResource` without a `description`, so the row is persisted with `Description = NULL`. The backend `report_runner` create path already defaults `description = name`; this PR makes the frontend match.

Specify 6 hard-fails when opening a label whose underlying `spappresource.Description` is NULL, so any site still running S6 alongside S7 hits this immediately. Reported as Specify support ticket **#824** — Grant confirmed the analysis and asked for a PR.

## Changes

- Extract `buildNewAppResource()` helper in `Create.tsx`; defaults `description: name.trim()`
- Export the helper via `exportsForTests` (same pattern as `getUrl` in this file)
- Add a unit test asserting the new resource has `description === name.trim()`

## Note for sites running both S6 and S7

If you already have rows with `Description IS NULL`, S6 will still hard-fail on them after this PR is merged — this fix only prevents *new* NULL rows. A one-time backfill is safe and cheap (description is a simple text label, free-form):

```sql
-- Run once per Specify database; backfills any historic NULL descriptions
-- so Specify 6 stops crashing on labels created by Specify 7.
UPDATE spappresource SET Description = Name WHERE Description IS NULL;
```

## Test plan

- [x] `npx jest --testPathPattern CreateAppResource` — 4/4 pass (3 existing + 1 new)
- [ ] Manually create a new App Resource via User Tools → App Resources → New Resource and confirm `spappresource.Description` is non-NULL in the DB
- [ ] Open the new resource as a label/report in Specify 6 and confirm it no longer crashes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where creating app resources with empty descriptions could cause errors. The description field now defaults to the trimmed resource name.

* **Tests**
  * Added tests to verify that app resource descriptions default correctly when not explicitly provided.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
+++ b/specifyweb/frontend/js_src/lib/components/AppResources/Create.tsx
@@ -259,17 +259,7 @@ function EditAppResource({
   readonly templateFile: string | undefined;
 }): JSX.Element {
   const resource = React.useMemo(
-    () =>
-      deserializeResource(
-        addMissingFields(type.tableName as 'SpAppResource', {
-          // I don't think this field is used anywhere
-          level: 0,
-          mimeType,
-          name: name.trim(),
-          specifyUser: userInformation.resource_uri,
-          spAppResourceDir: directory.resource_uri,
-        })
-      ),
+    () => buildNewAppResource(type, name, mimeType, directory),
     [directory, name, type, mimeType]
   );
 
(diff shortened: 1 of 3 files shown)


Iwantexpresso (changes requested):
- [x] `npx jest --testPathPattern CreateAppResource` — 4/4 pass (3 existing + 1 new)
- [ ] Manually create a new App Resource via User Tools → App Resources → New Resource and confirm `spappresource.Description` is non-NULL in the DB
- [ ] Open the new resource as a label/report in Specify 6 and confirm it no longer crashes

jest test results 

Image

Whenever the resource created the description field for the spappresource is currently staying as null. I have tried with creating 3 different app resources so far. I am a little unsure how request changes would whirl for this pr but I assume it wouldn't hurt just placing this in as one for now

foozleface:
Good catch — description was indeed still NULL. Root cause: the resource built in Create.tsx is never the one that gets saved. `EditAppResource`'s `onSaving` returns `false` (intentional, per #1596) and just navigates to the editor URL; `EditorWrapper` then builds a fresh skeleton from the URL params, and that skeleton — which had no description — is what `AppResourceEditor` actually persists.

5cfaccc moves the default onto that skeleton (`newAppResourceSkeleton`), so the saved record now carries `description = name` for both SpAppResource and SpViewSetObj, with tests covering the skeleton path. Could you re-test via User Tools → App Resources → New Resource? `spappresource.Description` should now match the resource name.

PR #7986, opened 2026-04-16
Fix crash for negated tree rank field query filters

### LINKED ISSUE #7982: Crash when using "not equal" operator on by rank tree fields
**Describe the bug**
Queries on tree rank fields using the "not equal" operator are causing a crash.

```
{"exception": "TypeError", "message": "Boolean value of this clause is not defined"
```

[Specify 7 Crash Report - 2026-04-16T13_40_54.429Z.txt](https://github.com/user-attachments/files/26787711/Specify.7.Crash.Report.-.2026-04-16T13_40_54.429Z.txt)

**To Reproduce**

1. Build a query on Collection Object
2. Add the line 'Determinations → Taxon → (Rank) → Full Name"
3. Run query
4. See the crash

**Example query:**
[Not Empty Contains.json](https://github.com/user-attachments/files/26787735/Not.Empty.Contains.json)

Image

**Reported By**
Jordi at MCNB

acwhite211:
Fixes #7982

Stored queries could crash when a rank tree field was used with a negated operator.  So using the QB with "not equal" on `Determinations -> Taxon -> Phylum` could throw an error.

Tree rank fields resolve to a SQLAlchemy `CASE` expression. The null-safe negation path used `mod_orm_field` or `orm_field`, which attempted to evaluate that SQL expression in Python boolean context and raised the error`TypeError: Boolean value of this clause is not defined`.

Fixed by replacing the boolean `or` fallback with an explicit `is not None` check before calling `null_safe_not()`.  Added a unit test to check this case for the future.

### Checklist

- [x] Add automated tests

### Testing instructions

- Open Query Builder for `CollectionObject`
- Add the field `Determinations -> Taxon -> Phylum`
- Choose a negated operator flow that maps to "not equal" / negated contains
- Enter a test value
- Run the query
- [x] Confirm the query completes without a crash or 500 error
- [x] Confirm rows with non-matching phylum values are still returned
- [x] Confirm rows with empty phylum values are still handled correctly

### DIFF
--- a/specifyweb/backend/stored_queries/queryfieldspec.py
+++ b/specifyweb/backend/stored_queries/queryfieldspec.py
@@ -421,7 +421,8 @@ class QueryFieldSpec(
 
             if negate:
                 if op_num in NULL_SAFE_NEGATE_OPS:
-                    predicate = null_safe_not(mod_orm_field or orm_field, f)
+                    field_expr = mod_orm_field if mod_orm_field is not None else orm_field
+                    predicate = null_safe_not(field_expr, f)
                 else:
                     predicate = sql.not_(f)
             else:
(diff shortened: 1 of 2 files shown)


emenslin (approved):
- [x] Confirm the query completes without a crash or 500 error
- [x] Confirm rows with non-matching phylum values are still returned
- [x] Confirm rows with empty phylum values are still handled correctly

Looks good, I checked with a couple different ranks and multiple operators and didn't run into any errors

grantfitzsimmons (approved):
- [x] Confirm the query completes without a crash or 500 error
- [x] Confirm rows with non-matching phylum values are still returned
- [x] Confirm rows with empty phylum values are still handled correctly

PR #7992, opened 2026-04-17
Prevent startup permission initialization from re-granting legacy collection access

acwhite211:
## Summary

This change stops the startup `run_key_migration_functions` path from re-importing legacy SP6 collection access into SP7 permissions.  Because the startup path re-ran the full initializer, users with legacy SP6 principal rows could regain `/system/sp7/collection` access on startup, even if that SP7 access had been intentionally removed.

Made running the key migrations function optional at startup with RUN_KEY_MIGRATION_ON_STARTUP=0 in the .env file.  Also fixed the warning of null tasks from happening in the key migration logs.

### Checklist

### DIFF
--- a/specifyweb/backend/setup_tool/schema_defaults.py
+++ b/specifyweb/backend/setup_tool/schema_defaults.py
@@ -84,6 +84,8 @@ def queue_apply_schema_defaults_background(discipline_id: int) -> str:
 @app.task(bind=True, max_retries=SCHEMA_DEFAULTS_MISSING_DISCIPLINE_MAX_RETRIES)
 def apply_schema_defaults_task(self, discipline_id: int):
     """Run schema localization defaults for one discipline in a background worker."""
+    task_id = getattr(self.request, 'id', None)
+
     try:
         discipline = Discipline.objects.get(id=discipline_id)
     except Discipline.DoesNotExist as exc:
@@ -98,9 +100,11 @@ def apply_schema_defaults_task(self, discipline_id: int):
                 discipline_id,
                 SCHEMA_DEFAULTS_MISSING_DISCIPLINE_MAX_RETRIES,
             )
-            finish_discipline_background_task(discipline_id, self.request.id)
+            if task_id is not None:
+                finish_discipline_background_task(discipline_id, task_id)
             return
     try:
         apply_schema_defaults(discipline)
     finally:
-        finish_discipline_background_task(discipline_id, self.request.id)
+        if task_id is not None:
+            finish_discipline_background_task(discipline_id, task_id)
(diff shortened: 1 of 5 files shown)


specifyweb/backend/setup_tool/schema_defaults.py
```diff
@@ -84,6 +84,8 @@ def enqueue() -> None:
 @app.task(bind=True, max_retries=SCHEMA_DEFAULTS_MISSING_DISCIPLINE_MAX_RETRIES)
 def apply_schema_defaults_task(self, discipline_id: int):
     """Run schema localization defaults for one discipline in a background worker."""
+    task_id = getattr(self.request, 'id', None)
```
melton-jason, line 87: (pretty optional)

Right now I see we're calling this function using the `run` method?

https://github.com/specify/specify7/blob/816121c99a86c5a7a37ef2851df9a8ea477c3ec1/specifyweb/specify/management/commands/run_key_migration_functions.py#L59

I see the Celery docs reccommend using `apply` to run tasks synchronously/locally:
https://docs.celeryq.dev/en/latest/reference/celery.app.task.html#celery.app.task.Task.apply

I'm not entirely sure what `run` does, but I'm assuming `apply` boils down to `run`. `run` generally states: 

> When called tasks apply the [run()](https://docs.celeryq.dev/en/latest/reference/celery.app.task.html#celery.app.task.Task.run) method. This method must be defined by all tasks (that is unless the __call__() method is overridden).
> **Run**
> The body of the task executed by workers.

https://docs.celeryq.dev/en/latest/reference/celery.app.task.html#celery.app.task.Task.run

Perhaps we can be more in-line with what Celery expects by instead using `apply` here. I'm not sure what  the tradeoffs are for going with the current approach over `apply`, but it seems to run the function anyway, so not too big of a deal.

melton-jason (approved):
Generally looks good! Nice work 💯

melton-jason (changes requested):
Actually, I don't think we can/should disable the `run_key_migration_functions` by default in the `v7.12.0.4` release, unfortunately. 

There's a lot of deduplication functionality in the `run_key_migration_functions` in the release that resolve regressions. 

We _could_ make an argument for keeping `run_key_migration_functions` disabled by default, but that would require everyone who had updated to `v7.12.0.0`-`v7.12.0.3` to intentionally re-enable the function and set the environment variable as part of their update process...

@grantfitzsimmons @acwhite211 
What are your thoughts on this?

grantfitzsimmons:
@melton-jason I say we keep it enabled for now and discuss for 7.12.1.

melton-jason:
I brought these changes except for the disabling of run_key_migration_functions into another branch/PR that I brought to `v7.12.0.4`: #7995

acwhite211:
Closing this PR, since changes were added to #7995

Issue #7797, opened 2026-03-09
System Config unknown table error: Spexportschema_exportmapping


bronwyncombs:
**Describe the bug**

While exploring the new system config tool, I got some toasts with the same error multiple times. 

[Specify 7 Crash Report - 2026-03-09T19_45_06.472Z.txt](https://github.com/user-attachments/files/25852257/Specify.7.Crash.Report.-.2026-03-09T19_45_06.472Z.txt)

Error: Trying to get unknown table: Spexportschema_exportmapping

**To Reproduce**

1. Go to demofish on main 
2. Go to system config 
3. Click around and wait 
4. See errors appear 

*When reporting an issue on the `production` branch or any development version of Specify, please try to recreate the issue on the latest tagged release (`v7`) and indicate if it does not occur there, indicating a regression.*

**Expected behavior**

No errors. 

**Screenshots**

<img width="1412" height="656" alt="Image" src="https://github.com/user-attachments/assets/414883b1-b9f4-4de7-b87f-8103a10b77f3" />

**Crash Report**

[Specify 7 Crash Report - 2026-03-09T19_45_06.472Z.txt](https://github.com/user-attachments/files/25852365/Specify.7.Crash.Report.-.2026-03-09T19_45_06.472Z.txt)

Please fill out the following information manually:
- OS:  mac Sequoia 15.0
- Browser: chrome
- Specify 7 Version: main 
Git SHA: | 6c4781d7e033b8e89ebad36a4be0ec66940ae1d6
-- | --
Git SHA:
[6c4781d7e033b8e89ebad36a4be0ec66940ae1d6](https://github.com/specify/specify7/commit/6c4781d7e033b8e89ebad36a4be0ec66940ae1d6)
- Database Name: sp7demofish
- Collection name: KUFishTissue
- User Name: sp7demofish
- URL: [e.g. https://ku_fish-v7.test.specifysystems.org/specify/workbench/215/]](https://sp7demofish20250909-main.test.specifysystems.org/specify/system-configuration/)

PR #7999, opened 2026-04-20
Fix cascaded Discipline delete blockers for Division

### LINKED ISSUE #7998: System Config delete blockers fail for Division when Discipline is reached through CASCADE
## Describe the bug

This is essentially the same underlying problem as #7797, but through a different entry point.

The hot fix in #7833 only handles the case where `Discipline` is the base table passed to the delete blockers endpoint.

The same issue still occurs when a non-`Discipline` table cascades into `Discipline` during blocker collection. A notable example is `Division` in the System Config tool.

Because `Discipline -> Division` is `on_delete=CASCADE`, collecting delete blockers for a `Division` also collects delete blockers for its related `Discipline`. The endpoint does not apply the same discipline-specific cleanup/handling in that cascaded case, so discipline-owned setup rows such as `Spexportschema_exportmapping` can still be returned.

That causes the frontend to hit the same kind of failure seen in #7797, because it attempts to treat `Spexportschema_exportmapping` as a regular datamodel table.

## To reproduce

1. Create or use a `Division` that has an associated `Discipline`.
2. Ensure that `Discipline` has discipline-scoped setup data that produces the same blocker chain as #7797, such as export schema mapping rows.
3. Open the `Division` record in the System Config tool in Specify.
4. Trigger delete blocker collection for that `Division`.

## Actual behavior

The delete blockers response can include setup/join-table rows such as:

- `Spexportschema_exportmapping`

This leads to the front-end error when it tries to resolve that table name.

## Expected behavior

Delete blocker collection for a parent record such as `Division` should apply the same discipline-specific handling used when `Discipline` is the base table.

The endpoint should not surface discipline-owned setup/join-table rows that are supposed to be cleaned up as part of the discipline delete flow.

acwhite211:
Fixes #7998

Fixes the remaining cascaded delete blockers cases related to #7797 that were not covered by #7833.

The previous fix only applied discipline-specific cleanup when `Discipline` was the base table being checked. That left a gap for parent records like `Division`, where Django cascades into related `Discipline` rows during delete blocker collection. In that path, discipline-owned setup rows such as `Spexportschema_exportmapping` could still surface and break the System Config UI.

I also found an issue with registry validation compared sorted policy keys.  It treated action array order as significant. That caused false `"operationPolicies is out of date"` errors when the front-end and back-end exposed the same actions in a different order.

## Changes

By normalize policy resource keys before comparison and normalize each resource's action array before comparison, the issue is solved.  Equivalent policy definitions no longer fail registry validation just because of action ordering differences.

The `delete_blockers` request it still taking a while to run when there are a lot of delete blockers.  In the future, we could try looking more into trying to speed it up.

### Checklist

- [ ] Add automated tests

### Testing instructions

- Open System Config and navigate to a `Division` that has a related `Discipline`. (ex. sp7demofish)
- Ensure that the related `Discipline` has discipline-scoped setup data, such as export schema mappings.
- Trigger delete blocker collection for the `Division`.
- [x] Confirm the delete blockers UI loads without error.
- [x] Confirm the 'link records' appear after loading.

### DIFF
--- a/specifyweb/specify/tests/test_delete_blockers.py
+++ b/specifyweb/specify/tests/test_delete_blockers.py
@@ -1,10 +1,11 @@
-import json
 from django.test import Client
+import json
 
-from specifyweb.backend.businessrules.exceptions import BusinessRuleException
-from specifyweb.backend.trees.tests.test_trees import GeographyTree
 from specifyweb.specify import models
-from specifyweb.specify.api.crud import delete_resource
+from django.db import router
+from django.test import TestCase
+from specifyweb.backend.delete_blockers.views import _collect_delete_blockers
+from specifyweb.backend.trees.tests.test_trees import GeographyTree
 
 def _url(obj):
     return f"/delete_blockers/delete_blockers/{obj._meta.model_name}/{obj.id}/"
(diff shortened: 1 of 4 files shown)


CarolineDenis (approved):

emenslin (approved):
- [x] Confirm the delete blockers UI loads without error.
- [x] Confirm the 'link records' appear after loading.

Looks good, delete blockers and linked records show up correctly. 

Image

kwhuber (approved):
- [x] Confirm the delete blockers UI loads without error.
- [x] Confirm the 'link records' appear after loading.

Tested on KU_Fish 👍.  

Image

specifyweb/specify/tests/test_delete_blockers.py
```diff
@@ -63,3 +79,250 @@ def test_children_dont_block_deletion(self):
...
+from specifyweb.specify import models
+from django.db import router
+from django.test import TestCase
+from specifyweb.backend.delete_blockers.views import _collect_delete_blockers
+from specifyweb.backend.trees.tests.test_trees import GeographyTree
+
+def _url(obj):
+    return f"/delete_blockers/delete_blockers/{obj._meta.model_name}/{obj.id}/"
+
+class TestDeleteBlockers(GeographyTree):
```
melton-jason, line 177: I think these test cases might have been incorrectly copied into the file? 
These imports, the `_url` function, and the `TestDeleteBlockers` class shadows/obscures objects of the same name defined previously in the file. 

Thankfully there didn't appear to be tests or functionality defined in the earlier version of the objects, so the behavior, and thus the test suite, remained the same within the latter defined objects. 

I've merged the duplicated imports, classes, and functions such that they are singly defined in [`c10446e` (#8043)](https://github.com/specify/specify7/pull/8043/commits/c10446efc197d5262a30e0194b9af49960fc04ae)

https://github.com/specify/specify7/blob/f4690f74a4c628cac6b821998fdf10e97f1042be/specifyweb/specify/tests/test_delete_blockers.py#L1-L166

PR #8016, opened 2026-04-23
Improve tree search efficiency#7941

### LINKED ISSUE #7752: [Large Databases]: Make tree searches by name more efficient

acwhite211:
Fixes #7752
Contributed by @foozleface
Based on https://github.com/specify/specify7/pull/7941 with fixes added.

Tree search in QueryComboBox sends a LIKE query per keystroke. On tree tables with 200K+ rows, using "contains" mode generates `LIKE '%pattern%'` which cannot use B-tree indexes and causes full table scans. This PR changes the default search mode from "contains" to "startsWith", enabling `LIKE 'pattern%'` which uses B-tree indexes. It also reduces the search result limit from 1000 to 50, since a typeahead dropdown never needs that many results.

### Implementation
- Change the default value of `treeSearchAlgorithm` user preference from `contains` to `startsWith` in `UserDefinitions.tsx`
- Export `QUERY_COMBO_BOX_SEARCH_LIMIT` constant set to 50 (was 1000)
- Add tests verifying the default search operator and result limit

### Testing instructions
- [ ] Open a form with a tree-based QueryComboBox (e.g., Taxon field on a Determination)
- [ ] Type a few characters and verify the typeahead dropdown populates quickly
- [ ] Verify that typing "rosa" matches "Rosaceae" and "Rosales" but not "Pterosaurus" (startsWith behavior)
- [ ] Check user preferences -- the tree search algorithm should now default to "starts with"
- [ ] Run the frontend tests: `npx jest --testPathPattern treeSearchEfficiency`


emenslin (changes requested):
- [x] Open a form with a tree-based QueryComboBox (e.g., Taxon field on a Determination)
- [x] Type a few characters and verify the typeahead dropdown populates quickly
- [x] Verify that typing "rosa" matches "Rosaceae" and "Rosales" but not "Pterosaurus" (startsWith behavior)
- [x] Check user preferences -- the tree search algorithm should now default to "starts with"

---

Switching the tree search algorithm to "contains" doesn't seem to work. 

Current search behavior in main: 

https://github.com/user-attachments/assets/9fa67b69-198f-474b-b25a-c6b09fae687e


Issue branch: 

https://github.com/user-attachments/assets/5ff2ecee-d11d-479c-aec8-0172aa45a61f

JDAM2k4 (changes requested):
- [x] Open a form with a tree-based QueryComboBox (e.g., Taxon field on a Determination)
- [x] Type a few characters and verify the typeahead dropdown populates quickly
- [x] Verify that typing "rosa" matches "Rosaceae" and "Rosales" but not "Pterosaurus" (startsWith behavior)
- [x] Check user preferences -- the tree search algorithm should now default to "starts with"

---

I am having the same issue as other reviewers. It seems that no matter the selection in User Preferences, the autofill search only returns searches in the form of "starts with - case insensitive."

https://github.com/user-attachments/assets/793a313d-32cb-4c22-91b0-7b2218faf6fd

rijulpoudel (changes requested):
- [x] Open a form with a tree-based QueryComboBox (e.g., Taxon field on a Determination)
- [x] Type a few characters and verify the typeahead dropdown populates quickly
- [x] Verify that typing "rosa" matches "Rosaceae" and "Rosales" but not "Pterosaurus" (startsWith behavior)
- [x] Check user preferences -- the tree search algorithm should now default to "starts with"

---

The testing instructions pass but both case sensitivity and the contains options look to be broken.
Image

Image

PR #8020, opened 2026-04-23
Fix N+1 queries in inheritance API#7936

### LINKED ISSUE #7875: Use batched iterators to evaluate large SQLAlchemy queries
Similar to #7864, by default SQLAlchemy loads all results of a query into a single buffer. This can lead to the python process allocating a very large chunk of memory that can be held for a long time.

SQLAlchemy supports iterating over very large result sets with a [yield_per](https://docs.sqlalchemy.org/en/14/orm/queryguide.html#fetching-large-result-sets-with-yield-per) method for queries.
Instead of allocating memory for the entire result set, SQLAlchemy will batch memory allocation into chunks to minimize the overall memory compared to allocating space for the entire result set up-front. 

From the SQLAlchemy docs: 

> Normally, the ORM will construct ORM objects for all rows up front, assembling them into a single buffer, before passing this buffer to the [Result](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result) object as a source of rows to be returned.

> The purpose of the yield_per option is to change this behavior so that the ORM result set is optimized for iteration through very large result sets (> 10K rows) ... When yield_per is used, the ORM will instead batch ORM results into sub-collections and yield rows from each sub-collection individually as the [Result](https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result) object is iterated, so that the Python interpreter doesn’t need to declare very large areas of memory which is both time consuming and leads to excessive memory use.

https://docs.sqlalchemy.org/en/14/orm/queryguide.html#fetching-large-result-sets-with-yield-per

acwhite211:
Fixes #7875
Contributed by @foozleface
Based on the https://github.com/specify/specify7/pull/7936 PR with additional fixes

The inheritance post-query processing in `api.py` ran per-row database queries to look up catalog numbers for parent and COG inheritance. On a query returning 500 rows, this produced 500+ individual SELECT statements. This rewrite collects IDs needing lookup, then bulk-prefetches all catalog numbers in 1-2 queries total. Also increases CSV `yield_per` from 1 to 2000, which was causing row-at-a-time fetching from the database.

### Implementation
- Rewrite `parent_inheritance_post_query_processing` to collect IDs needing catalog number lookup, then bulk-fetch via a single query using `select_related` and `values_list`
- Rewrite `cog_inheritance_post_query_processing` with two bulk-prefetch steps: (1) get parentcog_id for each child CO, (2) get primary member catalog number for each COG
- Change `yield_per(1)` to `yield_per(2000)` in CSV and KML export paths in `execution.py`
- Add N+1 regression tests that verify bulk queries are used

### Testing instructions
- [ ] Run a query that includes inherited catalog numbers (e.g., query on Components with catalogNumber column) and verify results are correct
- [ ] Export the same query to CSV and verify the file is complete and correct
- [ ] Run the test suite: `python manage.py test specifyweb.backend.inheritance.tests.test_n_plus_one`
- [ ] Optionally, enable Django query logging and confirm the inheritance processing uses 1-3 queries instead of N+1


acwhite211:
Still working on getting this PR working...

PR #8050, opened 2026-04-30
Implement idempotency test for `run_key_migration_functions`

### LINKED ISSUE #8042: Add record count idempotency tests for `run_key_migration_functions`
Add Django unit tests that verify the `run_key_migration_functions` management command only creates missing migration records on the first run, and does not create duplicate records when run again.

The test should capture relevant database record counts before the command runs, after the first run, and after a second run. The first run is expected to create or backfill records. The second run must leave all tracked record counts unchanged.

Prior to running the test, we'll need to construct database records similar to what Specify 6 does, in order to better catch actual edge cases. We should also add record changes that could occur with normal Specify 7 usage both before the first run, as well as between the first and second run.

## Background

`run_key_migration_functions` re-runs several important data migration helpers, including collection object type setup, schema config setup, permissions initialization, business rule initialization, app resource directory setup, tectonic rank setup, and misc backfills.  This was done to handle bringing in new records from Specify 6, which required all the key migration functions to handle being run multiple times, rather than just once.

Because this command may be run more than once in deployed environments, we need test coverage proving that it is idempotent with respect to record creation. Running the command a second time should not create duplicate records.

## Implementation Plan

1. Add a new Django test module, likely:
   `specifyweb/specify/tests/test_run_key_migration_functions.py`

2. Use the existing `ApiTests` fixture base from `specifyweb/specify/tests/test_api.py`.
   This already creates a realistic institution/division/discipline/collection/user setup.

3. Add helper functions:
   - `record_counts()` to snapshot counts for tracked models.
   - `count_diff(before, after)` to return only models whose counts changed.

4. Track models touched by the command, especially models where first-run creation is expected:
   - `Collectionobjecttype`
   - `Picklist`
   - `Picklistitem`
   - `Splocalecontainer`
   - `Splocalecontaineritem`
   - `Splocaleitemstr`
   - `UniquenessRule`
   - `UniquenessRuleField`
   - `LibraryRole`
   - `LibraryRolePolicy`
   - `Role`
   - `RolePolicy`
   - `UserRole`
   - `UserPolicy`
   - `Spappresourcedir`
   - `Tectonicunittreedef`
   - `Tectonicunittreedefitem`
   - `Tectonicunit`

5. In the test:
   - Snapshot counts before running the command.
   - Run `call_command("run_key_migration_functions", stdout=StringIO())`.
   - Snapshot counts after the first run.
   - Assert that at least one expected model count increased.
   - Run the command a second time.
   - Snapshot counts again.
   - Assert that the second-run diff is empty.

acwhite211:
Adds a Django regression test for the `run_key_migration_functions` management command to verify that it can be run repeatedly without creating duplicate migration-backed records.

The test snapshots relevant database record counts before the command runs, after the first run, and after a second run. The first run is expected to create or backfill records, while the second run must leave tracked record counts unchanged.

There is tracking on counts for models touched by the command, including collection object types, picklists, schema config records, permissions records, app resource dirs, uniqueness rules, and tectonic unit records. I also went ahead and added user-created picklist and permission records before the first run and between runs to simulate normal Specify 7 usage.  We could try doing more here to simulate other Specify 7 usage.

### Checklist

- [x] Add automated tests

### Testing instructions

- [ ] See the the back-end testing github action runs successfully.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for the key migration to ensure records and relationships are preserved during migration and that the process is idempotent.
  * Updated legacy stored-query test expectations to reflect adjusted error expectations for collection object query directions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/specify/specify7/pull/8050?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


acwhite211:
This one seems good to merge

PR #8059, opened 2026-05-05
Fix aggregated relation parsing for stored queries

### LINKED ISSUE #8040: (aggregated) not aggregating
**Describe the bug**
(aggregated) fields are not being treated as aggregated. When added to the query, multiple lines are returned.

**To Reproduce**
Use the `v7.12.0.4-paris` (discovered on `v7.12.0.4-berlin`) branch.

1. Use a copy of `naturkundemuseum` (e.g. `naturkundemuseum_2026_04_28`) and import or create this query:

    [Formatter Issue.json](https://github.com/user-attachments/files/27166016/Formatter.Issue.json) 

    Image

2. See that the results are not aggregated, but rather the collectors are represented on separate lines:

    Image

**This does not seem to be a problem on `naturkundemuseum` on `v7.12.0.4`:**

Image

**Expected behavior**
Only a single row should be present. The aggregated format should be respected.

**Reported By**
Marion at Museum für Naturkunde on [BoldDesk (#867)](https://specifysoftware.bolddesk.com/agent/tickets/867)

acwhite211:
Fixes #8040

Fixes saved stored queries where (aggregated) to-many relationship fields, like Collecting Event > Collectors (aggregated), were parsed incorrectly after reload. The parser was treating the repeated relationship name in the saved string id as a tree rank, which caused query execution to join through the related table directly and return duplicate rows instead of one aggregated value.  Solved by preserving already parsed to-many relation fields when reading saved stored query string ids.

### Checklist

### Testing instructions

https://github.com/specify/specify7/issues/8040#issue-4343416855

- Open Query Builder for Collection Object.
- Add a display field for Collecting Event > Collectors (aggregated).
- Use test data where one collection object’s collecting event has multiple collectors.
- Run the query and verify the collection object appears once, with all collectors shown in the aggregated collectors cell.
- Save the query, reopen it, and run it again.
- [ ] Verify the reopened query still returns one row per collection object and does not split collectors into duplicate rows.

### DIFF
--- a/specifyweb/backend/stored_queries/queryfieldspec.py
+++ b/specifyweb/backend/stored_queries/queryfieldspec.py
@@ -253,10 +253,21 @@ class QueryFieldSpec(
             node = table
 
         extracted_fieldname, date_part = extract_date_part(field_name)
-        field = node.get_field(extracted_fieldname, strict=False)
+        relation_already_in_path = (
+            is_relation
+            and len(join_path) > 0
+            and isinstance(join_path[-1], Relationship)
+            and join_path[-1].type not in {"many-to-one", "one-to-one"}
+            and join_path[-1].name.lower() == extracted_fieldname.lower()
+        )
+        field = (
+            None
+            if relation_already_in_path
+            else node.get_field(extracted_fieldname, strict=False)
+        )
 
         tree_rank_name = None
-        if field is None:  # try finding tree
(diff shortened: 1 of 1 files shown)


CarolineDenis (approved):

emenslin (approved):
- [x] Verify the reopened query still returns one row per collection object and does not split collectors into duplicate rows.

Looks good, results return only one row! 

Image

PR #8141, opened 2026-05-29
Merge pull request #6093 from specify/issue-6092

### LINKED ISSUE #6092: The 0022_ensure_default_cots migration is not currently revertable

CarolineDenis:
Define 0022_ensure_default_cots to be revertible

Fixes #6092 

### Checklist

- [ ] Add automated tests

## Summary by CodeRabbit

* **Chores**
  * Migration updated to add safe rollback support: rollbacks now clear collection-type associations and remove default collection types, ensuring migration can be reverted cleanly.

[Review Change Stack](https://app.coderabbit.ai/change-stack/specify/specify7/pull/8141?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/specifyweb/specify/migrations/0022_ensure_default_cots.py
+++ b/specifyweb/specify/migrations/0022_ensure_default_cots.py
@@ -6,11 +6,24 @@ class Migration(migrations.Migration):
     def handle_default_collection_types(apps, schema_editor):
         create_default_collection_types(apps)
 
+    def revert_default_collection_types(apps, schema_editor):
+        # Set all CollectionObject records to have a null CollectionObjectType
+        Collectionobject = apps.get_model('specify', 'Collectionobject')
+        Collectionobject.objects.all().update(collectionobjecttype=None)
+
+        # Set all the Collection records to have a null CollectionObjectType
+        Collection = apps.get_model('specify', 'Collection')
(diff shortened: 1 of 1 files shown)


CarolineDenis:
@coderabbit full review

Issue #7426, opened 2025-09-16
Add "ID" Field to Schema Config


grantfitzsimmons:
Users (e.g., NHM Geneva) require the ability to display the internal `ID` field (e.g., `collectionobject.CollectionObjectID`) for tables in Specify queries. This field is currently hidden by default and not exposed in schema configuration, limiting its visibility and configurability in queries.

There are several recent tables that have the ID field as visible (e.g. `collectionobjectgroup`), which should be seen as a reference:
<img width="1272" height="262" alt="Image" src="https://github.com/user-attachments/assets/d53c4eec-b723-41b1-895d-9a8b01d40812" />

What we need to do is:

- Make the `ID` field (primary key) for all tables available in Schema Config.
- By default, the `ID` field should remain hidden in the schema, but users should be able to configure its visibility.
- Make sure it is still **not editable** on the forms, WorkBench, Batch Edit, etc. and should be rendered as read-only

## Requirements

1. **Expose ID Field in Schema Config**
   - For every table in Specify, add a Schema Config record for the primary key `ID` field (e.g., `CollectionObjectID` in CollectionObject).
       - The 'Caption' should be `ID`, the 'Description' should be `Unique identifier for each record in the table. Ensures row-level uniqueness and is used for indexing and referential integrity. Non-editable, automatically assigned by the database.`
   - The Schema Config record should allow configuration of the field’s visibility in queries.

2. **Default Hidden State**
   - The `ID` field should be set as **hidden by default** for all tables in the schema.
   - Existing queries should not display the `ID` field unless users have explicitly configured it to be visible.

3. **Configurable Visibility**
   - Users must be able to toggle the default visibility of the `ID` field via Schema Config.
   - If set to visible, the `ID` field should appear as a selectable field in the query builder and be included in new queries.

4. **UI/UX**
   - Update Schema Config UI to display the `ID` field for all tables.
   - Clearly indicate the current visibility state (hidden or visible) and provide a toggle for users to change it.
   - If the `ID` field is hidden, it should not appear as an option in the query builder unless the user changes its visibility.

## User Story Example

> For our mammal collection, tissues coming from other institutions often don’t have a catalog number. They do have sample IDs, but these are not unique by design. For such objects, we prefer to use an internal ID as the identifier. The ID field is very convenient for this purpose, as it is simple to read (whereas a GUID is too complex).

## Acceptance Criteria

- [ ] Schema Config record exists for the `ID` field on all tables.
[...]

PR #8154, opened 2026-06-01
Issue 8124 8126 fix 7660 7682 2 6266 after review

### LINKED ISSUE #8065: Prevent `sync_schema_config_fields` from creating schema records for `{table}Id` fields
This was handled in https://github.com/specify/specify7/commit/b95e4211df9b222f4a019f8d2f08afdfdb281e50#diff-8c096fc6189e8fbf5f9c608c661989ccfecee166ec4222127550478775f9d1e9 in `v7.12.0.4-berlin`, but needs to be integrated into `main` again after changes from `v7.12.0.5` and `v7.12.0.6`. It was decided that creating schema config values for ID fields is not currently in the plans.

See it described here: https://discourse.specifysoftware.org/t/fix-missing-resources-in-a-specify-database/3599

It should only create schema config records for fields that are expected to have schema records (non ID fields). 

## Current
```
specify@62c69a171f67:/opt/specify7$ ve/bin/python manage.py sync_schema_config_fields --discipline-id 3
[28/Apr/2026 03:25:34] [WARNING] [specifyweb.backend.setup_tool.apps:18] Failed to retrieve last active setup task.
[28/Apr/2026 03:25:34] [WARNING] [specifyweb.backend.trees.apps:19] Failed to retrieve default tree creation information.
Discipline: Ornithology (ID=3)
Missing table containers:
- Component
- DNASequencingRunAttachment
- LocalityUpdate
- LocalityUpdateRowResult
Missing fields:
- AbsoluteAge: component
- Accession: accessionId
- AccessionAgent: accessionAgentId
- AccessionAttachment: accessionAttachmentId
- AccessionAuthorization: accessionAuthorizationId
- AccessionCitation: accessionCitationId
- Address: addressId
- AddressOfRecord: addressOfRecordId
- Agent: agentId, components
- AgentAttachment: agentAttachmentId
- AgentGeography: agentGeographyId
- AgentIdentifier: agentIdentifierId
- AgentSpecialty: agentSpecialtyId
- AgentVariant: agentVariantId
- Appraisal: appraisalId
- Attachment: attachmentId, spDataSetAttachments
- AttachmentImageAttribute: attachmentImageAttributeId
- AttachmentMetadata: attachmentMetadataID
- AttachmentTag: attachmentTagID
- AttributeDef: attributeDefId
- Author: authorId
- AutoNumberingScheme: autoNumberingSchemeId
- Borrow: borrowId
- BorrowAgent: borrowAgentId
- BorrowAttachment: borrowAttachmentId
- BorrowMaterial: borrowMaterialId
- BorrowReturnMaterial: borrowReturnMaterialId
- CollectingEvent: collectingEventId
- CollectingEventAttachment: collectingEventAttachmentId
- CollectingEventAttr: attrId
- CollectingEventAttribute: collectingEventAttributeId
- CollectingEventAuthorization: collectingEventAuthorizationId
- CollectingTrip: collectingTripId
- CollectingTripAttachment: collectingTripAttachmentId
- CollectingTripAttribute: collectingTripAttributeId
- CollectingTripAuthorization: collectingTripAuthorizationId
- Collection: userGroupScopeId
- CollectionObject: collectionObjectId
- CollectionObjectAttachment: collectionObjectAttachmentId
- CollectionObjectAttr: attrId
- CollectionObjectAttribute: collectionObjectAttributeId
- CollectionObjectCitation: collectionObjectCitationId
- CollectionObjectProperty: collectionObjectPropertyId
- CollectionRelType: collectionRelTypeId
- CollectionRelationship: collectionRelationshipId
- Collector: collectorId
- CommonNameTx: commonNameTxId
- CommonNameTxCitation: commonNameTxCitationId
- Component: absoluteAges, catalogNumber, collectionObject, componentId, createdByAgent, identifiedBy, identifiedDate, integer1, integer2, integer3, integer4, integer5, integer6, modifiedByAgent, name, number1, number2, number3, number4, number5, number6, proportion, relativeAges, role, text1, text2, text3, text4, text5, text6, timestampCreated, timestampModified, type, uniqueIdentifier, verbatimName, version, yesNo1, yesNo2, yesNo3, yesNo4, yesNo5, yesNo6
- ConservDescription: conservDescriptionId
- ConservDescriptionAttachment: conservDescriptionAttachmentId
- ConservEvent: conservEventId
- ConservEventAttachment: conservEventAttachmentId
- Container: containerId
- DNAPrimer: dnaPrimerId
- DNASequence: dnaSequenceId
- DNASequenceAttachment: dnaSequenceAttachmentId
- DNASequencingRun: dnaSequencingRunId
- DNASequencingRunAttachment: attachment, createdByAgent, dnaSequencingRun, dnaSequencingRunAttachmentId, modifiedByAgent, ordinal, remarks, timestampCreated, timestampModified, version
- DNASequencingRunCitation: dnaSequencingRunCitationId
- DataType: dataTypeId
- Deaccession: deaccessionId
- DeaccessionAgent: deaccessionAgentId
- DeaccessionAttachment: deaccessionAttachmentId
- Determination: determinationId
- DeterminationCitation: determinationCitationId
- Determiner: determinerId
- Discipline: tectonicUnitTreeDef, userGroupScopeId
- Disposal: disposalId
- DisposalAgent: disposalAgentId
- DisposalAttachment: disposalAttachmentId
- DisposalPreparation: disposalPreparationId
- Division: userGroupScopeId
- ExchangeIn: exchangeInId
- ExchangeInAttachment: exchangeInAttachmentId
- ExchangeInPrep: exchangeInPrepId
- ExchangeOut: exchangeOutId
- ExchangeOutAttachment: exchangeOutAttachmentId
- ExchangeOutPrep: exchangeOutPrepId
- Exsiccata: exsiccataId
- ExsiccataItem: exsiccataItemId
- Extractor: extractorId
- FieldNotebook: fieldNotebookId
- FieldNotebookAttachment: fieldNotebookAttachmentId
- FieldNotebookPage: fieldNotebookPageId
- FieldNotebookPageAttachment: fieldNotebookPageAttachmentId
- FieldNotebookPageSet: fieldNotebookPageSetId
- FieldNotebookPageSetAttachment: fieldNotebookPageSetAttachmentId
- FundingAgent: fundingAgentId
- GeoCoordDetail: geoCoordDetailId
- Geography: geographyId
- GeographyTreeDef: geographyTreeDefId
- GeographyTreeDefItem: geographyTreeDefItemId
- GeologicTimePeriod: geologicTimePeriodId
- GeologicTimePeriodTreeDef: geologicTimePeriodTreeDefId
- GeologicTimePeriodTreeDefItem: geologicTimePeriodTreeDefItemId
- Gift: giftId
- GiftAgent: giftAgentId
- GiftAttachment: giftAttachmentId
- GiftPreparation: giftPreparationId
- GroupPerson: groupPersonId
- InfoRequest: infoRequestID
- Institution: userGroupScopeId
- InstitutionNetwork: institutionNetworkId
- Journal: journalId
- LatLonPolygon: latLonPolygonId
- LatLonPolygonPnt: latLonPolygonPntId
- LithoStrat: lithoStratId
- LithoStratTreeDef: lithoStratTreeDefId
- LithoStratTreeDefItem: lithoStratTreeDefItemId
- Loan: loanId
- LoanAgent: loanAgentId
- LoanAttachment: loanAttachmentId
- LoanPreparation: loanPreparationId
- LoanReturnPreparation: loanReturnPreparationId
- Locality: collectingEvents, localityId
- LocalityAttachment: localityAttachmentId
- LocalityCitation: localityCitationId
- LocalityDetail: localityDetailId
- LocalityNameAlias: localityNameAliasId
- LocalityUpdate: collection, createdbyagent, localityUpdateId, modifiedbyagent, recordset, specifyuser, status, taskid, timestampcreated, timestampmodified
- LocalityUpdateRowResult: localityUpdateRowResultId, localityupdate, result, rownumber
- MaterialSample: materialSampleId
- MorphBankView: morphBankViewId
- OtherIdentifier: otherIdentifierId
- PaleoContext: paleoContextId
- PcrPerson: pcrPersonId
- Permit: permitId
- PermitAttachment: permitAttachmentId
- PickList: pickListId
- PickListItem: pickListItemId
- PrepType: prepTypeId
- Preparation: preparationId
- PreparationAttachment: preparationAttachmentId
- PreparationAttr: attrId
- PreparationAttribute: preparationAttributeId
- PreparationProperty: preparationPropertyId
- Project: projectId
- RecordSet: recordSetId
- RecordSetItem: recordSetItemId
- ReferenceWork: referenceWorkId
- ReferenceWorkAttachment: referenceWorkAttachmentId
- RelativeAge: component
- RepositoryAgreement: repositoryAgreementId
- RepositoryAgreementAttachment: repositoryAgreementAttachmentId
- Shipment: shipmentId
- SpAppResource: spAppResourceId
- SpAppResourceData: spAppResourceDataId
- SpAppResourceDir: spAppResourceDirId
- SpAuditLog: spAuditLogId
- SpAuditLogField: spAuditLogFieldId
- SpExportSchema: spExportSchemaId
- SpExportSchemaItem: spExportSchemaItemId
- SpExportSchemaItemMapping: spExportSchemaItemMappingId
- SpExportSchemaMapping: spExportSchemaMappingId
- SpFieldValueDefault: spFieldValueDefaultId
- SpLocaleContainer: spLocaleContainerId
- SpLocaleContainerItem: spLocaleContainerItemId
- SpLocaleItemStr: spLocaleItemStrId
- SpPermission: permissionId
- SpPrincipal: userGroupId
- SpQuery: spQueryId
- SpQueryField: isStrict, spQueryFieldId
- SpReport: spReportId
- SpSymbiotaInstance: spSymbiotaInstanceId
- SpTaskSemaphore: spTaskSemaphoreId
- SpVersion: spVersionId
- SpViewSetObj: spViewSetObjId
- SpVisualQuery: spVisualQueryId
- SpecifyUser: specifyUserId
- Storage: storageId
- StorageAttachment: storageAttachmentId
- StorageTreeDef: storageTreeDefId
- StorageTreeDefItem: storageTreeDefItemId
- Taxon: components, taxonId
- TaxonAttachment: taxonAttachmentId
- TaxonAttribute: taxonAttributeId
- TaxonCitation: taxonCitationId
- TaxonTreeDef: taxonTreeDefId
- TaxonTreeDefItem: taxonTreeDefItemId
- TreatmentEvent: treatmentEventId
- TreatmentEventAttachment: treatmentEventAttachmentId
- VoucherRelationship: voucherRelationshipId
- Workbench: workbenchId
- WorkbenchDataItem: workbenchDataItemId
- WorkbenchRow: workbenchRowId
- WorkbenchRowExportedRelationship: workbenchRowExportedRelationshipId
- WorkbenchRowImage: workbenchRowImageId
- WorkbenchTemplate: workbenchTemplateId
- WorkbenchTemplateMappingItem: workbenchTemplateMappingItemId
Run again with --apply to create missing records.
```

## Desired

```
[28/Apr/2026 04:14:40] [WARNING] [specifyweb.backend.trees.apps:19] Failed to retrieve default tree creation information.
Discipline: Geology (ID=1638400)
Missing table containers:
- Component
- DNASequencingRunAttachment
- LocalityUpdate
- LocalityUpdateRowResult
- SpDataSetAttachment
Missing fields:
- AbsoluteAge: component
- Agent: components
- Attachment: spDataSetAttachments
- CollectionObject: components
- Component: absoluteAges, catalogNumber, collectionObject, createdByAgent, identifiedBy, identifiedDate, integer1, integer2, integer3, integer4, integer5, integer6, modifiedByAgent, name, number1, number2, number3, number4, number5, number6, proportion, relativeAges, role, text1, text2, text3, text4, text5, text6, timestampCreated, timestampModified, type, uniqueIdentifier, verbatimName, version, yesNo1, yesNo2, yesNo3, yesNo4, yesNo5, yesNo6
- DNASequencingRunAttachment: attachment, createdByAgent, dnaSequencingRun, modifiedByAgent, ordinal, remarks, timestampCreated, timestampModified, version
- Discipline: tectonicUnitTreeDef
- Gift: agent1, agent2, agent3, agent4, agent5
- Loan: agent1, agent2, agent3, agent4, agent5
- Locality: collectingEvents
- LocalityUpdate: collection, createdbyagent, modifiedbyagent, recordset, specifyuser, status, taskid, timestampcreated, timestampmodified
- LocalityUpdateRowResult: localityupdate, result, rownumber
- RelativeAge: component
- SpDataSetAttachment: attachment, collectionMemberId, createdByAgent, modifiedByAgent, ordinal, remarks, spdataset, timestampCreated, timestampModified, version
- SpQueryField: isStrict
- Taxon: components
Applied missing schema config records.
```

### LINKED ISSUE #8101: Address code improvement review for 6266
Address improvement suggestions from CodeRabbit review 
https://github.com/specify/specify7/pull/8094#discussion_r3267633009

### LINKED ISSUE #8124: Address code improvement review for 7660
Address improvement suggestions from CodeRabbit review https://github.com/specify/specify7/pull/8125

### LINKED ISSUE #8126: Address code improvement review for 7682-2
Address improvement suggestions from CodeRabbit review https://github.com/specify/specify7/pull/8127

melton-jason:
Taken from https://github.com/specify/specify7/pull/8142#issue-4547272152 by @CarolineDenis:

Fixes #8124 
Fixes #8126 
Fixes #8101 
Fixes #8065 

### Checklist

- [ ] Add automated tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected schema configuration references in data migrations
  * Fixed discipline and tree definition linking issues
  * Improved admin and permission initialization logic
  * Enhanced query logging controls based on debug settings

* **Refactor**
  * Streamlined migration utilities for improved efficiency
  * Simplified multi-database routing during migrations
  * Optimized permission seeding with per-user checks
  * Restructured role assignment logic using set-based operations

* **Chores**
  * Removed legacy debugging code and unused migration helpers
  * Updated function signatures to remove unnecessary parameters


grantfitzsimmons:
PR review by `deepseek-v4-pro`:

---

## Summary by Category

### 🛡️ Defensive / Safety Improvements

| File | Change | Notes |
|------|--------|-------|
| `backend/stored_queries/execution.py` (×2) | Guarded `log_sqlalchemy_query(query)` behind `if settings.DEBUG:` | Prevents unconditional SQL logging in production in two call sites (`execute` and `series_post_query`). Good catch. |
| `specify/api/utils.py` | Moved `settings.DEBUG` guard into `log_sqlalchemy_query` itself + removed `literal_binds=True` | Centralizes the guard so all callers are protected. Removing `literal_binds` avoids parameter-binding errors on queries that use `%` or other special characters. ✅ |
| `specify/management/commands/run_key_migration_functions.py` | Changed `logger.error(f"An error occurred: {e}")` → `logger.exception("An error occurred while running key migrations")` | `logger.exception` auto-includes the traceback; the old code threw away the stack trace. Also removed the unused f-string (broke static analysis). ✅ |

---

### 🧩 Migration Correctness / Bug Fixes

| File | Change | Notes |
|------|--------|-------|
| `backend/businessrules/migration_utils.py` | `catnum_rule_uneditable`: now tracks matching rule IDs and calls `.update(isDatabaseConstraint=True)` on them when one already exists | **Previously:** if a catalog-number uniqueness rule already existed for a discipline, it was silently ignored (no-op). Now it ensures the existing rule is a DB constraint. This is a meaningful behavioral fix — ensures pre-existing rules are enforced at the DB level. ✅ |
| `backend/businessrules/uniqueness_rules.py` | `fix_global_default_rules`: replaced the two-query `Exists`/`OuterRef` subquery approach with a Python-set-signature comparison | **Old code** deleted discipline-level `UniquenessRuleField` rows that matched any global field, then cleaned up orphaned rules. **New code** builds a frozen set of `(fieldPath, isScope)` signatures per global rule and per discipline rule and compares them in memory, only deleting the discipline rule + its fields when there's an exact signature match. This is more deterministic and avoids the complex chained subquery approach that could have subtle Django ORM bugs. **⚠️ Risk:** If two rules have the same fields but in a different order (which `frozenset` handles), consider edge cases where fields are not the only distinguishing factor (e.g., rule-level metadata like `title` on the `UniquenessRule`). Should still be fine since the goal is to deduplicate discipline-level duplicates of global rules. |
| `specify/migration_utils/update_schema_config.py` (multiple sites) | `get_or_create` / `except MultipleObjectsReturned` → explicit `.filter().order_by('id').first()` + `.create()` if None | **Issue:** `get_or_create` with `discipline` + `name` + `schematype` can return `MultipleObjectsReturned` when duplicates exist from prior migrations. The old `except MultipleObjectsReturned` path used `.filter(...).first()`, which returned an arbitrary row. **Fix:** Always use `.filter(...).order_by('id').first()` + create-if-None. Deterministic row selection. ✅ |
| `specify/migration_utils/update_schema_config.py` | `name=field_name.lower()` → `name__iexact=field_name` (×7 locations) | This changes exact lowercase matching to case-insensitive exact matching. The old code assumed `field_name.lower()` would match the DB value; the new code properly handles mixed-case names like `collectionObjectType`. **⚠️ Verify:** `__iexact` depends on database collation. With case-insensitive collation, `__iexact` is equivalent to `__exact`, so this should be safe. ✅ |
| `specify/migration_utils/update_schema_config.py` | `update_cog_type_fields`: `container__name="CollectionObject"` → `"collectionobject"` | Fixes a case-sensitivity mismatch for `Splocalecontainer` names. ✅ |
| `specify/migration_utils/update_schema_config.py` | `update_0034_schema_config_field_desc`: `if not desc_str or not name_str: continue` → separate `if desc_str is not None` + `if name_str is not None` blocks | **Old code:** skipped the entire iteration if *either* was falsy, meaning in a partial-data scenario (one text string missing), neither got updated. **New code:** updates each independently. This is a correctness fix for partially populated locale data. ✅ |
| `specify/migration_utils/update_schema_config.py` | `reverse_hide_component_fields`: `items.update(ishidden=True)` → `items.update(ishidden=False)` | **This is a behavioral reversal fix.** The "reverse" function was setting `ishidden=True` (hiding items) instead of `ishidden=False` (unhiding them). Clear logic error. ✅ |
| `specify/migration_utils/update_schema_config.py` | `update_schema_config_field_desc_for_components`: added `language="en"` filter to `.update()` calls | Prevents accidentally overwriting non-English locale strings with English defaults. Important for multilingual collections. ✅ |
| `specify/migration_utils/sp7_schemaconfig.py` | `RelativeAgeCitation`: `absoluteAgeCitationId` → `relativeAgeCitationId` | **Bug fix.** The schema config listed the wrong field ID for `RelativeAgeCitation` — it was using `AbsoluteAgeCitation`'s ID. This would cause locale/schema configuration issues for relative age citations. ✅ |
| `specify/migrations/0008_ageCitations_fix.py` | `revert_migration`: `usc.update_relative_age_fields` → `usc.revert_relative_age_fields` | The revert function was calling the **forward** migration function instead of the **reverse** one. Rollback would have been a no-op or incorrectly applied. ✅ |
| `specify/migrations/0009_tectonic_ranks.py` | Reversed rollback order: now `revert_create_root_tectonic_node` before `revert_default_tectonic_ranks` | **Critical ordering fix.** Tectonic unit tree nodes depend on the tree def items (which are ranks), so you must delete the child (tree entries) before the parent (ranks). The old order was reversed and would violate FK constraints on rollback. ✅ |

---

### 🔗 Tectonic Unit Linking / Determinism

| File | Change | Notes |
|------|--------|-------|
| `specify/migration_utils/default_cots.py` | `fix_tectonic_unit_treedef_discipline_links`: replaced ad-hoc pairwise linking with a deterministic `zip()` | **Old code:** iterated disciplines and tree-defs independently, using `.first()` and `.last()` calls that could produce non-deterministic pairings across database engines. **New code:** `list()` + `.order_by('id')` + `zip(empty_disciplines, empty_tectonic_unit_treedefs, strict=False)`. Pairing is now deterministic by ID ordering. Excess disciplines (more than tree-defs) get new tree-defs created. ✅ |
| `specify/migration_utils/tectonic_ranks.py` | `create_default_tectonic_ranks`: simplified query + always `get_or_create` | Removed the `TectonicTreeDef.objects.values_list('discipline_id', flat=True)` exclusion since `.filter(tectonicunittreedef__isnull=True)` already filters those out. Simpler and more correct. ✅ |
| `specify/migration_utils/tectonic_ranks.py` | `create_root_tectonic_node`: added upsert logic — if a tree-def item with `name="Root"` already exists, update its `rankid`, `parent`, and `isenforced` | Previously only created if `(treedef, rankid=0, parent=None)` didn't exist. Now also handles the case where a Root node exists but has stale field values. ✅ |
| `specify/migration_utils/tectonic_ranks.py` | `revert_create_root_tectonic_node`: scoped the `TectonicUnit` delete to `name="Root", definition=tectonic_tree_def, parent__isnull=True` + reversed order (delete Units before TreeDefItems) | Old code deleted **all** TreeDefItems for the treedef and **all** TectonicUnits named "Root" globally (no discipline/definition scope). New code is properly scoped. ✅ |

---

### 🔐 Permissions / User Initialization

| File | Change | Notes |
|------|--------|-------|
| `backend/permissions/initialize.py` | `create_admins`: removed early-return guard, replaced with per-user `UserPolicy` check | **Old code:** if *any* admin policy existed, all admin creation was skipped entirely (regardless of which user). **New code:** each user is individually checked for an existing admin `UserPolicy`. If a user already has one, they're skipped; others still get policies created. This ensures new users added after the first admin are not missed. **`is_sp6_user_permissions_migrated` removal:** that function's logic is now inlined as the `UserPolicy.objects.filter(...).exists()` check. Verify this is semantically equivalent — the old function may have checked collection-scoped policies. ✅ |
| `backend/permissions/initialize.py` | `assign_users_to_roles`: raw SQL changed from `NOT IN (subquery)` to `NOT EXISTS (correlated subquery)` | **Old code** had two separate `NOT IN` blocks — one for user/role and one for collection. The second was `c.UserGroupScopeId NOT IN (SELECT DISTINCT r.collection_id ...)` which excluded *any* collection that had *any* user-role assignment, not just the current user's. **New code** is a single correlated `NOT EXISTS` that correctly checks whether *this specific user* has a role in *this specific collection*. This is a significant correctness fix for collections with multiple users using different roles. ✅ |

---

### 🗄️ Database Alias / Multi-DB Support

| File | Change | Notes |
|------|--------|-------|
| `backend/patches/migration_utils.py` | `update_is_accepted` and `update_coordinates`: added `.using(db_alias)` routing | Uses `schema_editor.connection.alias` to target the correct database in multi-DB setups. Previously all operations defaulted to `"default"`. Also switched from `tree_model.objects` to `tree_model._base_manager` (standard in migration functions to avoid filtered managers). ✅ |
| `specify/migrations/0002_geo.py` | Removed `create_default_collection_types` import from `specify.api.utils`, removed `using=db_alias` kwarg, imported from local migration utils instead | The `using` parameter was being passed but `create_default_collection_types` may not accept it (or it wasn't needed). The import is now from the migration-level utils where it belongs. ✅ |

---

## 🔍 Items That Deserve Extra Attention

1. **`fix_global_default_rules` refactor** — The shift from ORM subqueries to in-memory set comparison is cleaner, but verify this against a real database with multiple disciplines and overlapping global rules. A migration integration test would provide confidence.

2. **`is_sp6_user_permissions_migrated` removal** — This function was completely removed in favor of a direct `UserPolicy` filter. Double-check that the new inline check covers all cases the old function did (especially legacy SP6 migrated users who might have collection-scoped admin policies rather than global ones).

3. **`__iexact` vs `__exact` + `.lower()`** — On MySQL with `utf8mb4_general_ci` (case-insensitive collation, which is standard for Specify), `__iexact` and `.lower()` behave identically. However, if any database uses a case-sensitive collation, `__iexact` is actually **more correct** than the old `.lower()` approach. This change is an improvement.

4. **`revert_create_root_tectonic_node` ordering** — The fix reverses the delete order (units before tree-def items) which is correct for FK constraints. However, verify the parent tree-def items also get cleaned up — the function deletes `TectonicUnitTreeDefItem.objects.filter(treedef=tectonic_tree_def).delete()` which would cascade, but explicit ordering is still preferred.

---

## ✅ Overall Assessment

This PR is a well-scoped collection of fixes addressing review feedback from #8142. The changes fall primarily into:

- **Bug fixes** (wrong field IDs, wrong revert functions, wrong rollback ordering)
- **Correctness improvements** (per-user admin seeding, deterministic tectonic linking, locale string handling)
- **Safety improvements** (debug logging guards, DB alias support, better error logging)
- **Code quality** (simpler ORM patterns, deterministic queries)

No new features are introduced. All changes are in migration utilities, schema config helpers, and permission initialization — code that runs during migrations and setup, not hot-path request handling.

The changes are well-contained, each targeting a specific problem from the linked issues. I'd recommend testing the tectonic unit rollback (`0009_tectonic_ranks` revert) and the admin creation flow with multiple users to verify the per-user behavior.

grantfitzsimmons (commented):
To make my contribution more worthwhile, I am looking closely at the things deepseek suggested need extra attention. It seems like most of the changes seem obvious and should have been done in the first place!

> 1. **`fix_global_default_rules` refactor** — The shift from ORM subqueries to in-memory set comparison is cleaner, but verify this against a real database with multiple disciplines and overlapping global rules. A migration integration test would provide confidence.

To me it seems the new code is actually a total improvement (the old code **over-deleted** partial `Exists` matches, but the new code only deletes on _exact_ full-signature match). Happy with how this looks.

> 2. **`is_sp6_user_permissions_migrated` removal** — This function was completely removed in favor of a direct `UserPolicy` filter. Double-check that the new inline check covers all cases the old function did (especially legacy SP6 migrated users who might have collection-scoped admin policies rather than global ones).

This is still here despite not being used by any runtime code, and the new solution is more straightforward. Should be removed IMO.

https://github.com/specify/specify7/blob/72414b6a049ff8b78c193797e3864cf1c8336098/specifyweb/backend/permissions/initialize.py#L29-L33

> 3. **`__iexact` vs `__exact` + `.lower()`** — On MySQL with `utf8mb4_general_ci` (case-insensitive collation, which is standard for Specify), `__iexact` and `.lower()` behave identically. However, if any database uses a case-sensitive collation, `__iexact` is actually **more correct** than the old `.lower()` approach. This change is an improvement.

Happy with this!

> 4. **`revert_create_root_tectonic_node` ordering** — The fix reverses the delete order (units before tree-def items) which is correct for FK constraints. However, verify the parent tree-def items also get cleaned up — the function deletes `TectonicUnitTreeDefItem.objects.filter(treedef=tectonic_tree_def).delete()` which would cascade, but explicit ordering is still preferred.

Cascade is desirable here IMO.

---

Should we do some integration testing to make sure all is good in a multi-db environment (maybe using the scheme defined in `docker-compositions/docker/specifycloud`)?

CarolineDenis:
@grantfitzsimmons, yes integration testing is a must I think.

specifyweb/backend/businessrules/uniqueness_rules.py
```diff
@@ -490,58 +490,48 @@ def rule_is_global(scopes: Iterable[str]) -> bool:
...
+                discipline__isnull=True
+            ).prefetch_related("uniquenessrulefield_set")
+        }
+
+    with transaction.atomic():
+        # REFACTOR: See if we can simplify this even further. We should be able
+        # to collapse this query -> iteration -> check workflow to a single
+        # query.
+        # That would eliminate the N + 1 problem with this current approach,
+        # where every scoped rule needs to be evaluated.
```
melton-jason, line 536: (note for discussion)

I left this comment in the code, but I'll leave this as a comment on the PR for visibility.
There was a noticeable change in the asymptotic growth (time complexity) with this change.

Noticeably, the previous implementation of `fix_global_default_rules`: 
- Iterated over each distinct non-null DisciplineID
- For each DisciplineID, performed a minimum of 2 queries to identify the "duplicate" scoped Uniqueness Rule and another 2 queries to delete them

Importantly, the number of Queries and iteration only grew on the number of Disciplines. So assuming the number of Disciplines remained the same, there could be any number of added Uniqueness Rules within those Disciplines without seeing a slowdown or additional database queries.
(This is only partly/theoretically true: there will be a slowdown for those queries if there is a sufficiently large amount of Uniqueness Rules and the queries are performed on unindexed fields or in a way that can't utilize the indexes, though I'm holding that slowdown as negligible or constant between the two approaches). 

I've cleaned up the new approach and eliminated extra queries in [`6e37e1e` (this PR)](https://github.com/specify/specify7/pull/8154/changes/6e37e1e9e43d141270637061c5f66c5fa27c03a0), but there still is a change in the approach:
- Now we make a DB query to fetch all Uniqueness Rules that are scoped (have a non-null DisciplineID)
- For each rule, we check whether there is an identical unscoped Uniqueness Rule
- If there is an identified duplicate, we delete the Uniqueness Rule

Our "bottleneck" is now largely on the amount of scoped Uniqueness Rules, rather than the number of Disciplines.
Practically, I'm not sure how much of a slowdown this would introduce.
If we have the time, we can try and resolve this slowdown (I imagine we can use and filter on annotations on the query for scoped rules for duplicate rules, or just revert back to something like the past approach if there are no bugs), and/or we can evaluate how much of a slowdown this introduces.

specifyweb/backend/permissions/initialize.py
```diff
@@ -53,13 +53,28 @@ def create_admins(apps=apps) -> None:
...
+        # - Are not already Institution Admins
+        # - Have not already seen activity in Sp 7 (don't have Sp7 permissions)
+        #   - (The Institution Admin permission could have been intentionally
+        #      removed)
+        # - Are admins in Sp 6
+
+        # The ordering here for checks here is intentional: it's more likely a
+        # user has Sp 7 permissions than being an admin, so we do the former
+        # check first
+        if is_sp6_user_permissions_migrated(user=user, apps=apps):
```
melton-jason, line 70: (note here on why I brought back `is_sp6_user_permissions_migrated`)

I brought back `is_sp6_user_permissions_migrated` here.
It was needed for the following workflow:
- A user is an admin in Specify 6, and should not be an admin in Specify 7
- In this scenario, the user's Specify 7 Institution Admin privileges have been explicitly revoked

Previously, because this is a part of Run Key Migration Functions, the user in question would be re-made into an Institutional Admin, even if they were explicitly removed as an admin.

With `is_sp6_user_permissions_migrated`, we assume that users that don't have Sp 7 permissions _can't_ have been intentionally removed as Specify 7 Institution Admins, so we can safely migrate their Admin preference. Otherwise, we can't tell if the user was intentionally removed as an admin

melton-jason:
@coderabbitai review

specifyweb/specify/migration_utils/tectonic_ranks.py
```diff
@@ -101,43 +93,40 @@ def create_root_tectonic_node(apps):
...
+            TectonicUnit.objects.filter(
                 parent=None,
-                definition=tectonic_tree_def,
-                definitionitem=tectonic_tree_def_item
+                definition=OuterRef("pk")
             )
+        )
+    ).filter(
+        root_node_exists=False
+    )
```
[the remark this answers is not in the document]
melton-jason: This was a largely intentional decision. The Parent=NULL and RankID=0 both carry the same meaning: that a particular node in the tree is the "root" node. 
While technically `rankid` and `parent` can become "out of sync" in this regard, we'd like to prioritize parent as the root-identifying field because it is constrained as a self-referential foreign key while rankid could have any integer vaule. 
The code currently errs on the assumption that the rankid _might_ be out-of-sync with the parent field, so we prioritize the parent.
We could do a more complicated matching behavior where we handle such out-of-sync cases more gracefully, but they were omitted here because the application does generally keep these values in sync, and a root node by definition _always_ has a NULL parent and it's more likely for rankid to be messed up over parent.

We could modify other places that currently use both rankid and parent to identify the root node to follow similar logic.

specifyweb/specify/migrations/0002_geo.py
```diff
@@ -73,7 +72,7 @@ class Migration(migrations.Migration):
     
     def consolidated_python_django_migration_operations(apps, schema_editor):
         db_alias = schema_editor.connection.alias or 'migrator'
-        create_default_collection_types(apps, using=db_alias)
+        create_default_collection_types(apps)
```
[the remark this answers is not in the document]
melton-jason: I've implemented a more general way of ensuring all current and future functions within `run_key_migration_functions` will be executed with the same database connection that Migrations are run in. 
The important changes are included in
[`9f9df79` (this PR)](https://github.com/specify/specify7/pull/8154/changes/9f9df797b631157353253296059e5cb7f2ec73fd), can you take a look and give me your thoughts?

The only difference between the migrations database and the default database is the user that is used to connect.
Ideally, the functions executed as part of `run_key_migration_functions` should never require the elevated level of permissions that is provided to the migrations user (the ability to perform DDL statements like dropping and creating tables, indices, etc.).
With this in mind, I am generally on the fence about whether we should even allow the migrator user to perform `run_key_migration_functions`: they should only be data movements, and never DDL statements.

melton-jason:
@coderabbitai review

specifyweb/specify/migrations/0009_tectonic_ranks.py
melton-jason, line 1: I've modified these reverse migrations to resolve errors and better handle Tectonic Unit tree structures. 

Additionally, I've currently changed the migrations to be as close to the exact forward migration as possible and focus on not modifying or overwriting other user changes.
Previously, we were deleting all Tectonic Unit nodes and ranks.

For a point of discussion: how do we want to handle these reverse migration? Is the intention to delete all Tectonic nodes and ranks, or just those created by the forward migration?
If the end-goal for the end-user is to reverse all migrations of the version that introduced this migration (`v7.10.0`), then they would also need to revert [specify/0004_stratigraphy_age](https://github.com/specify/specify7/blob/main/specifyweb/specify/migrations/0004_stratigraphy_age.py) anyways, which would drop the Tectonic Unit tables.

At least in my testing thus far it looks like existing, modified rows cause issues with earlier migrations.
We will need to modify these reverse migrations further assuming we want these reverse migrations to be an exact reverse of the forward migrations. 

For example, I get the following error when reverting `specify/0006_fix_tectonic_tree_fields` when there exists existing TectonicUnit trees: 

`django.db.utils.IntegrityError: (1452, 'Cannot add or update a child row: a foreign key constraint fails (`specify`.`#sql-alter-41c-6d8d-f27`, CONSTRAINT `tectonicunittreedefi_TectonicUnitTreeDefI_d5e99533_fk_tectonicu` FOREIGN KEY (`TectonicUnitTreeDefID`) REFERENCES `tectonicunittreedef` (`TectonicUnitTreeDefID`))')`

specifyweb/specify/migrations/0009_tectonic_ranks.py
```diff
@@ -1,12 +1,107 @@
...
+        has_child_rank=Exists(
+            TectonicUnitTreeDefItem.objects.filter(
+                parent=OuterRef("pk")
+            )
+        )
+    ).filter(
+        has_nodes=False,
+        has_child_rank=False,
+        name="Root"
+    ).delete()
```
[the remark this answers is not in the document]
melton-jason: The general behavior of tackling these reverse tectonic unit migrations will be discussed further with the team. I've outlined and introduce some of the issue in https://github.com/specify/specify7/pull/8154#discussion_r3374035067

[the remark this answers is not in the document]
melton-jason: A follow-up GitHub issue is not needed at the moment, but thank you!

specifyweb/specify/migration_utils/sp7_schemaconfig.py
```diff
@@ -148,7 +148,7 @@
     'RelativeAge': ['number2', 'yesno2', 'relativeAgeId', 'relativeAgePeriod', 'text1', 'agent1', 'collectionDate', 'text2', 'agent2', 'date1', 'date2', 'collectionObject', 'relativeAgeCitations', 'number1', 'yesno1'],
     'CollectionObject': ['collectionObjectType', 'relativeAges', 'absoluteAges', 'cojo'],
     'AbsoluteAgeCitation': ['collectionMember', 'absoluteAgeCitationId'],
-    'RelativeAgeCitation': ['absoluteAgeCitationId', 'collectionMember'],
+    'RelativeAgeCitation': ['relativeAgeCitationId', 'collectionMember'],
```
melton-jason: I haven't done testing yet to determine whether we need a migration to resolve any erroneous old cases of `RelativeAgeCitation -> absoluteAgeCitationId`, so reviewers feel free to test and/or comment on this or skip it for now

melton-jason:
This should be ready to review and discuss now!
The PR was a lot of work, largely from: 
- Some of the fixes and improvements were largely incompatible with our datamodel (e.g., assuming a to-many relationship is to-one and/or not considering user customizations, as in the case https://github.com/specify/specify7/pull/8154/changes/BASE..72414b6a049ff8b78c193797e3864cf1c8336098#diff-94ecaa42162a27adbe0a1063db5707262cf295c5d0217e8ade3865b5f29cc461 or assuming a pairing of TectonicUnitTreeDefs and Disciplines based on the order of creation [`8893687` (this PR)](https://github.com/specify/specify7/pull/8154/changes/8893687d31ea2c05acb03b702b65f69867657de4), and would not hold up with a lot of Specify
- Some of the fixes/suggestions would have bright back bugs that the previous code already addressed
  - Especially consider the removal of `is_sp6_user_permissions_migrated`, as disucssed in https://github.com/specify/specify7/pull/8154#discussion_r3350025208
  - As well as the matching for TectonicUnit nodes/ranks: see https://github.com/specify/specify7/pull/8154/changes/BASE..72414b6a049ff8b78c193797e3864cf1c8336098#diff-94ecaa42162a27adbe0a1063db5707262cf295c5d0217e8ade3865b5f29cc461
- Some of the comments/fixes were addressed to specific functions/migrations that should have been applied globally or generally to other places in the code (other places that would exhibit the exact same behavior as the fixes from the)
- Another point was that performance was largely not considered. I've gone through and have optimized what I can to avoid significant slowdowns and/or resource usage in the database, but there still could be further improvements!

More generally, looks like when we're considering CodeRabbit's comments in the future we should keep in mind:
1. Are there any customizations within Specify that would make the suggested change invalid or have unexpected behavior?
2. Does the suggested improvement affect other areas of Specify and need to be applied elsewhere?
3. Does the suggested improvement hold up under scale? That is, will it negatively impact databases of arbitrary size and is there any way to incorporate performance improvements?

This isn't a comprehensive list and may not be accurate some times, but these are the most common/biggest issues I've noticed so far with AI code review (it seems to typically miss the "bigger picture" areas of Specify/development). 

Outside of the suggested changes from CodeRabbit, there were some problems with migrations that were a part of `run_key_migration_functions`, as well as some reverse migrations. In both cases we weren't considering the breadth of customizations that were valid for users. So the forward migrations or `run_key_migration_functions` could create records where it shouldn't (e.g., a new TectonicUnit for Disciplines that already had one) and reverse migrations were too general and could delete records created or edited by the user that may be unexpected. 

I have left comments in the code to walk through some of the changes and edge cases, hopefully to prevent us or future developers from re-introducing bugs the code was addressing (this would be a good area for automatic tests!).

There are some points of discussion for the @specify/developers and @specify/ux-testing teams:
- https://github.com/specify/specify7/pull/8154#discussion_r3357510500
- https://github.com/specify/specify7/pull/8154#discussion_r3374035067
- https://github.com/specify/specify7/pull/8154#discussion_r3349593049
  - Specifically, whether/when we'd want to address performance
  - This can be somewhat related to the comment regarding batching UniquenessRule queries made in https://github.com/specify/specify7/commit/dc3025f617884cd814d7b3b6acc15294217e26f5
    - https://github.com/specify/specify7/blob/457a590fc59595dcd8ed2daa8c00c92bddc5f8e0/specifyweb/backend/businessrules/uniqueness_rules.py#L197-L209

CarolineDenis:
# Migration Validation Report

## Initial Changes Applied

### Ichthyology

* **Collection Object (CO)**

  * `Text3` → renamed to **NewText3**, unhidden, required

* **Gift**

  * `Agent1` → renamed to **Agent111**, unhidden

* **Gift Preparation**

  * `Quantity Resolved` → renamed to **Quantity Resolved 2**
  * Added "2" to description

* **Collection Object Group**

  * `Text1` → converted to picklist using **Age Type**

* **COGType**

  * Added **NewType** to the `Type` picklist

* **Collector**

  * `Text2` → renamed to **Text1**
  * Description updated to **Text1**

* **Absolute Age**

  * `AgeType` assigned to **Age Type** picklist
  * Added picklist value **NewAge**

* **Component**

  * `CO` field caption changed to **Parent**

* **COType**

  * Catalog Number Format Name assigned to:

    * Format: `CatalogNumber`
    * Pattern: `2026-#####`

* **Collection Object Group (COG)**

  * `COGType` field caption changed to **Type of the COG**

* **Relative Age**

  * `RelativeAgeCitationID` unhidden

* **Storage**

  * Added text to **Unique Identifier** description

* **PaleoContext**

  * Updated table description

* **Accession**

  * `DateAccessionPrecision` unhidden

* **Absolute Age**

  * `Component` field unhidden

* **Discipline**

  * Added one new value to the `Type` picklist

* **Catalog Number Uniqueness**

  * Enabled uniqueness across **Component** and **Collection Object**

---

### Geology

* Created Geo discipline in System Configuration
* Created new collection
* Discipline Type not assigned to picklist
* `MIGRATION_0040_UPDATE_FIELDS` not applied
* `MIGRATION_0033_TABLES` not applied
* Added **SecondSuperStructure** rank to Tectonic Tree
* Added:

  * `Node1` under SuperStructure
  * `Node2` under SecondSuperStructure

---

## Rebuild

Executed:

```bash
docker down
docker compose up --build
```

---

## Post-Rebuild Verification

### Geology

* New rank persisted
* New tectonic nodes persisted
* Missing schema configuration issue still present

### Ichthyology

* Catalog Number uniqueness across Component and Collection Object persisted
* All schema configuration changes persisted

---

## Reverting Migrations

Attempted to revert all migrations.

### Failure

Migration:

```text
0006_fix_tectonic_tree_fields
```

Error:

```text
Cannot add or update a child row:
a foreign key constraint fails
(
  `fwri_2025_09_09`.`#sql-alter-1-15c`,
  CONSTRAINT `tectonicunittreedefi_TectonicUnitTreeDefI_d5e99533_fk_tectonicu`
  FOREIGN KEY (`TectonicUnitTreeDefID`)
  REFERENCES `tectonicunittreedef` (`TectonicUnitTreeDefID`)
)
```

---

## Reapplying Migrations

Migration:

```text
0009_tectonic_ranks
```

Failed with:

```text
Unknown column
'tectonicunittreedefitem.TectonicUnitTreeDefID'
in 'ON'
```

---

## Recovery Procedure

Executed:

```bash
python manage.py migrate specify 0009 --fake
python manage.py migrate
```

Result:

* All migrations reported as applied.

---

## Post-Migration Verification

### Geology

#### Fixed

* Discipline Type is now assigned to a picklist
* `MIGRATION_0040_UPDATE_FIELDS` applied
* `MIGRATION_0033_TABLES` applied

#### Remaining Issue

* Trees cannot be opened due to:

```text
tectonicunittreedefitem.TectonicUnitTreeDefID
```

---

### Ichthyology

#### Changes That Persisted

* Catalog Number uniqueness across Component and Collection Object remains enabled
* CO → Text3 renamed to NewText3
* COGType → NewType picklist value persists
* Collector → Text2 renamed to Text1 and description persists
* Absolute Age → AgeType assigned to Age Type picklist
* Absolute Age → NewAge picklist value persists

#### Changes Lost

* Gift → Agent111 reverted to Agent1 and hidden again
* Gift Preparation → Quantity Resolved 2 reverted
* Collection Object Group → Text1 no longer uses Age Type picklist
* Component → CO field caption reverted from Parent
* COType → Catalog Number format reverted to default
* COG → COGType caption reverted from "Type of the COG"
* RelativeAgeCitationID hidden again
* Storage description reverted
* PaleoContext table description reverted
* DateAccessionPrecision hidden again
* Absolute Age → Component hidden again
* Discipline → newly added picklist value lost

grantfitzsimmons (commented):
I recreated the same behavior as Caroline, following her steps followed. I can confirm that when adding new ranks and items to the tectonic unit tree, it does indeed block reverting the migrations. It seems we should probably let it be destructive, as it will remove any customizations to fields and tables _created_ via migrations (see that Specify 7-only fields and tables lose their customizations due to the defaults being applied via migrations once they are recreated).

I used a copy of the KU Mammals database. I could revert and reapply migrations without issue as long as there were no blockers.

I downloaded the Mammals discipline schema as JSON before and after reverting and re-applying the migrations and compared them. One thing to note:
* The `absoluteageid` (among other `*id` fields) were no longer present in the schema after reverting to `specify 0001` and running migrations again:
    Image

See https://github.com/specify/specify7/issues/7426, but I suppose they will be included again in 7.12.1

specifyweb/specify/migrations/0009_tectonic_ranks.py
melton-jason: In [`919267c` (this PR)](https://github.com/specify/specify7/pull/8154/changes/919267c49696eddd46916a7f52c53cc480447465), I've modified the reverse migration (specify/0009_tectonic_ranks) to delete all TectonicUnit related records. 

This is not a strict reverse of the forward migrations: if the user has Tectonic Trees, ranks, or nodes that differ from the defaults then they will be removed.

The upside/PRO of this approach is that it allows earlier migrations ([specify/0006_fix_tectonic_tree_fields](https://github.com/specify/specify7/blob/9f0ef8c2cc9b8e7661e6477482cd1b8f58b75c2d/specifyweb/specify/migrations/0006_fix_tectonic_tree_fields.py#L1), and [specify/0004_stratigraphy_age](https://github.com/specify/specify7/blob/9f0ef8c2cc9b8e7661e6477482cd1b8f58b75c2d/specifyweb/specify/migrations/0004_stratigraphy_age.py#L1)) to execute regardless of any customizations and changes to the TectonicUnit. 

The downside/CON of this approach is that _all_ TectonicUnit records are removed when this migration (specify/0009) is reversed. If the user has any TectonicUnit records they'd like to preserve and/or are generally not expecting to be removed when `specify/0009` is ran, then the records will be deleted without warning. 

An alternative approach here is that we could instead raise an error if the database contains non-default TectonicUnit information, and explain that the user should backup their TectonicUnit tree(s), delete them, and then proceed with the reverse migration. 

If it is desirable, we might be able to have the best of both worlds and prompt the user for input during the reverse migration (see Django's [InteractiveMigrationQuestioner](https://github.com/django/django/blob/5948e6657f8cb6aeaab1a5a45640d089230f461a/django/db/migrations/questioner.py#L90), and https://stackoverflow.com/a/57572147), asking the user if they have non-default TectonicUnit records/structure whether they want to continue (and delete the TectonicUnit trees) or stop the reverse migration.

melton-jason:
I have addressed the issues regarding the TectonicUnit reverse migrations from https://github.com/specify/specify7/pull/8154#issuecomment-4669365355.
In short, I have tentatively gone ahead with restoring the old behavior of deleting all TectonicUnit related records when `specify/0009_tectonic_ranks` is reverted. This would allow earlier migrations to be reverted without issue, but any TectonicUnit customizations will be remove/deleted. 
We could alternatively raise an error if we detect the user's database contains any customized TectonicUnit tree(s) and let the user backup and delete their TectonicUnit tree(s) before allowing the migration to be reversed.

Full details are present in the following discussion: https://github.com/specify/specify7/pull/8154#discussion_r3423070099

Feel free to provide thoughts on this discussion!

--- 

DisciplineType PickLists should now be created in newly created Collections (from the System Configuration tool), and the PickList should be assigned in newly created Disciplines. 

-- 

Regarding the Schema Configuration changes that are not present in newly created Disciplines from the System Configuration tool (noted as `MIGRATION_0040_UPDATE_FIELDS` and `MIGRATION_0033_TABLES` not being applied in https://github.com/specify/specify7/pull/8154#issuecomment-4669365355), this is because those changes aren't being applied at Discipline creation: only during their respective migrations. 

More specifically, the schema configuration defaults are applied by the following function:
https://github.com/specify/specify7/blob/9f0ef8c2cc9b8e7661e6477482cd1b8f58b75c2d/specifyweb/backend/setup_tool/schema_defaults.py#L19-L65

The default schema configuration when a Discipline is created is largely defined by the schema_localization_en.json file: 
https://github.com/specify/specify7/blob/main/config/common/schema_localization_en.json

The changes from `specify/0033_update_paleo_desc` and `specify/0040_components` were not made to the schema_localization_en.json file, so they were not reflected when creating new Disciplines.
For a short term solution, we can update the schema_localization_en.json file to have the changes from recent migrations.
Applying that solution to the future and indefinitely will not be that scalable however (we have already forgotten to update the file with the changes from at least two migrations).
We should hopefully come up with an easier longer-term solution to handle making Schema Configuration changes via migrations (and having them also apply when creating new Disciplines).

grantfitzsimmons (approved):
This review was kind of a mess so I reformatted using Gemini.

Good news is that it works!

### **Databases tested:**
| Database | Discipline |
|----------|-----------|
| `aumnh_fish_2026_06_26` | Ichthyology |
| `alfmuseum_2026_06_26` | Multi |
| `aumnh_herbarium_2026_06_26` | Botany |
| `aumnh_inverts_2026_06_26` | Invertebrate |
| `bohart_entomology_2026_06_26` | Entomology |
| `calvertmarinemuseum_2026_06_26` | Multi |

### Forward Migrations PASS ✅

Every database tested shows **"No migrations to apply"**, the branch's migrations are already present.

### Reverse Migrations, Core Apps Pass, Permissions Fail 🔴

| App | Status | Notes |
|-----|--------|-------|
| `workbench` (8 migrations) | ✅ PASS | All reverse cleanly |
| `setup_tool` (1) | ✅ PASS | All reverse cleanly |
| `businessrules` (9) | ✅ PASS | All reverse cleanly |
| `specify` (44) | ✅ PASS | All reverse cleanly (0023→0044) |
| `patches` (8) | ✅ PASS | All reverse cleanly |
| `notifications` (6) | ✅ PASS | All reverse cleanly |
| `accounts` (3) | ✅ PASS | All reverse cleanly |
| `attachment_gw` (1) | ✅ PASS | All reverse cleanly |
| **`permissions` (8)** | **❌ FAIL** | **Not reversible** — 0008 crashes |

Permission reverse failure is expected and was pre-acknowledged by @melton-jason 

### Schema Config (`splocale*`)

From the `0023_update_schema_config_text` migration logs, the revert function properly processes Schema Config items per table:

```
CollectionObjectGroup:     17 items reverted
CollectionObjectGroupJoin: 10 items reverted
AbsoluteAge:               12 items reverted
RelativeAge:               14 items reverted
TectonicUnit:              7 items reverted
TectonicUnitTreeDefItem:   5 items reverted
CollectionObject:          4 items reverted
TectonicUnitTreeDef:       2 items reverted
CollectionObjectType:      4 items reverted
CollectionObjectGroupType: 1 item reverted
AbsoluteAgeCitation:       1 item reverted
RelativeAgeCitation:       1 item reverted
```

**All re-applied cleanly** on forward migration with matching item counts.

### TectonicUnit Trees ✅

Migration `0009_tectonic_ranks` creates:
- 6 ranks per tree (Root, Superstructure, Tectonic Domain, Tectonic Subdomain, Tectonic Unit, Tectonic Subunit)
- Root nodes where missing (with nodenumber=1, isaccepted=1)
- Trees for Disciplines that lack them

**Reverse deletes all trees** (nodes → ranks → defs → discipline associations). This is a **data-loss operation** but the trees are reconstructed on re-apply with identical structure (just new primary keys).

### Uniqueness Rules ✅

`businessrules.0006` and `0007` successfully create rules on apply:
- **Storage.uniqueIdentifier** → Global scope
- **Deaccession.deaccessionNumber** → Global
- **Disposal.disposalNumber** → Global
- **ExchangeOut.exchangeOutNumber** → Division scope (Global)
- **AccessionAuthorization.permit** → scoped to accession/repositoryAgreement (per discipline)
- **CollectingEventAuthorization.permit** → scoped to collectingEvent (per discipline)
- **CollectingTripAuthorization.permit** → scoped to collectingTrip (per discipline)

### Some notes

We might want to put a warning for the `0009` migration since it deletes TectonicUnit trees on reverse? I imagine if you are reversing you know the risks. There are also some model changes in the `specify` app that we should resolve.

CarolineDenis:
I second Grant's review and obtained the same results. During this round of testing, I confirmed the following:

- Migrations applied successfully: All migrations, including for the Geo discipline created in the system config tool, executed properly.

- Data loss observed: All changes made within schema config and trees were lost after reverting and reapplying migrations.

melton-jason:
@coderabbitai

specifysoftware:

https://speciforum.org/t/specify-7-12-0-7-release-announcement/3863/1

PR #8311, opened 2026-07-15
feat(attachments): add "match by mapping file" support

### LINKED ISSUE #8302: Mapping File Support in the Batch Attachment Uploader
The Batch Attachment Uploader currently matches attachment files to records by parsing the *filename itself* against a selected matching field (e.g. Catalog Number). This works well when the field has a UIFormatter with no regex parts, since Specify can reliably derive the field value from the filename (and normalize things like leading zeros).

It breaks down for institutions where the matching field is free text or contains non-numeric characters — e.g. UMMZI-90126 at UMich. In these cases:

- Single-attachment-per-record uploads mostly work, because the whole filename (minus extension) is compared to the field value.
- Multi-attachment-per-record uploads do not work, because the uploader's `_suffix` convention (`UMMZI-90126_dorsal.jpg`) requires a strict, regex-free format to reliably split "the part that identifies the record" from "the part that's just a disambiguating suffix." Free-text fields don't have this structure, so the tool can't tell where the catalog number ends and the suffix begins.

Specify 6 solved this with an **attachment mapping file**, which was an explicit CSV that pairs each filename with the record identifier it belongs to, so the record/filename association never has to be inferred from the filename's structure. This decouples matching from field formatting entirely, and naturally supports many-files-per-record.

## Goals
- Let users provide an explicit filename → matching-field-value mapping, so multi-attachment matching works regardless of the matching field's format (or lack of one).
- Build this as a second mode within the existing Batch Attachment Uploader, using the same validation, disambiguation, upload, and rollback logic.
- Support many attachments per record via the mapping file, with no constraint on filename structure.

## Non-Goals
- Not replacing filename-pattern matching. That remains the default/existing mode for institutions whose formatted fields already work well with it.
- Not auto-generating mapping files from existing data (out of scope for v1).
- Not supporting arbitrary non-tabular mapping formats, CSV only.

## Proposed Solution
Add a **Matching Mode** choice when creating a new attachment dataset:
1. **Match by filename** (current/default behavior) — unchanged.
2. **Match by mapping file** (new) — user uploads a CSV that explicitly lists which record each file belongs to, instead of relying on the filename's format.

### Workflow for mapping-file mode
1. User selects the matching field, same picklist as today (Catalog Number, GUID, etc.) so this is what determines *which record* a row matches to.
2. User selects "Match by mapping file" as the matching mode.
3. User uploads a CSV. Specify presents a column-mapping step (similar to WorkBench import) so the user indicates which CSV column is the **match value** and which is the **filename**. (See open question on a remarks column.)
4. User selects/drags in the actual image files, same as today.
5. Specify cross-references the CSV against the selected files by **exact filename match** without parsing or normalization of the filename needed, since the record association comes from the CSV row, not the filename's structure.
6. Record matching (CSV match value → database record) reuses the existing disambiguation logic: single match, no match, or multiple matches, exactly as filename mode does today.

### Example (from the UMich data)

| Catalog Number | Attachment Name |
|---|---|
| UMMZI-90126 | UMMZI-90126.jpg |
| UMMZI-90175 | UMMZI-90175.jpg |
| UMMZI-90175 | UMMZI-90175_1.jpg |
| UMMZI-90176 | UMMZI-90176.jpg |
| UMMZI-90176 | UMMZI-90176_1.jpg |

Two rows sharing a catalog number simply means two attachments go to the same record — no special suffix convention needed, since the pairing is explicit.

## UI / Data Model Changes

- **Dataset creation dialog:** add a "Matching Mode" selector (Filename Pattern / Mapping File).
- **CSV upload + column picker:** reuse WorkBench-style mapping UI for choosing which columns are "Match Value" and "File Name."
- **Dataset table:** in mapping-file mode, the field-value and filename columns are populated from the CSV rather than derived from the filename; "Incorrectly Formatted" no longer applies in this mode (there's no format to violate).
- **Dataset metadata:** store the matching mode and column mapping on the dataset record so it's preserved on reload/resume, consistent with how the existing filename-mode config is stored.

## Edge Cases & Validation Rules
- CSV row references a filename that wasn't uploaded → flag as `File Missing`.
- An uploaded file isn't referenced by any CSV row → flag as `Not in Mapping File`.
- Duplicate filename entries within the CSV → validation error before matching proceeds.
- CSV match value has no matching record, or matches multiple records → reuse existing `No match` / `Multiple Matches. Click to Disambiguate` behavior.
- Filename case sensitivity — should be exact match, but worth surfacing a warning if a case-only mismatch is detected (common when files move between OS's).
- Large CSVs — should be validated/matched at the same scale the tool already handles for filename mode.
- Rollback behaves identically in mapping-file mode.

### Discussed in https://github.com/specify/specify7/discussions/5941

<sup>Originally posted by **grantfitzsimmons** April  9, 2024</sup>

>Yes, the catalog number field is a text field in several of the UMich databases. This enables support for the non-number characters they store in the field, like the UMMZI- suffix in this case. 
>
> In S6, we have been using an attachment mapping file which describes the catalog numbers and attachment filenames explicitly. I'm attaching an example here for reference. If a similar system could be added to S7, that would be great.

In this example, this mapping file is necessary to work around an issue with the catalog number field being formatted.

See this video to see how it worked: https://discourse.specifysoftware.org/t/attachment-mapping-file-video/101?u=grant

From @realVinayak:
> Yes, unless the catalog number field itself is formatted (has a formatter other than text), then multiple records would not be matched. Specify 6 "got around" the issue via the mapping file

Here's an example file sent to us by Garth at UMich:
**Example mapping file:**
> |Catalog Number|Attachment Name|
> |---|---|
> |UMMZI-90126|UMMZI-90126.jpg|
> |UMMZI-90159|UMMZI-90159.jpg|
> |UMMZI-90171|UMMZI-90171.jpg|
> |UMMZI-90175|UMMZI-90175_1.jpg|
> |UMMZI-90175|UMMZI-90175.jpg|
> |UMMZI-90176|UMMZI-90176_1.jpg|
> |UMMZI-90176|UMMZI-90176.jpg|
> |UMMZI-90179|UMMZI-90179_1.jpg|
> |UMMZI-90179|UMMZI-90179.jpg|
> |UMMZI-90209|UMMZI-90209_1.jpg|
> |UMMZI-90209|UMMZI-90209.jpg|
> |UMMZI-90231|UMMZI-90231_1.jpg|
> |UMMZI-90231|UMMZI-90231.jpg|
> |UMMZI-90234|UMMZI-90234_1.jpg|
> |UMMZI-90234|UMMZI-90234.jpg|
> |UMMZI-90237|UMMZI-90237.jpg|
> |UMMZI-119907|UMMZI-119907_1.jpg|
> |UMMZI-119907|UMMZI-119907.jpg|
> |UMMZI-119908|UMMZI-119908_1.jpg|
> |UMMZI-119908|UMMZI-119908.jpg|

It should be implemented as it was in Specify 6.

They tried to import using the batch attachment uploader but could not indicate multiple images associated with a single specimen due to the formatting for that field. 

**Reported By**
Garth at University of Michigan</div>

grantfitzsimmons:
Fixes #8302

This PR adds a **"Match by mapping file"** mode to the Bulk Attachment Uploader. Users can upload a CSV mapping file that explicitly pairs filenames with record identifiers (catalog numbers, etc.), enabling multi-attachment-per-record uploads without requiring filename structure conventions.

| filename | catalogNumber |
|----------|---------------|
| specimen1.jpg | 12345 |
| specimen1_2.jpg | 12345 |
| specimen2.jpg | 67890 |

This lets you attach multiple files to the same record, use files with names that don't follow a pattern, and see exactly what's expected before you start uploading.

When you first open the uploader, a dialog asks which matching method you want. If you choose mapping file, you upload your CSV and pick which columns contain the filename and the record identifier. The table then shows all your expected rows — rows waiting for a file show "Awaiting file", and any files you add are paired with the correct row automatically.

image

image

image

image

> [!WARNING]  

#### Migration changes
- Added `matchingmode` field to `Spattachmentdataset` model with choices `[('mappingFile', 'Mapping File')]` (`null` = filename mode). New migration `0002_add_matchingmode.py` in `attachment_gw` Django app.

### Checklist

- [x] Add automated tests

## Test sets

You can use this set of [test-files.zip](https://github.com/user-attachments/files/30032533/test-files.zip), tested with the KU Entomology database. Please also test a variety of other cases!

### 1. `catalog-number/`: Collection Object → Catalog Number

Test both the standard mapping and multi-mapping.

**Steps:**
1. Create new dataset → **Match by mapping file**
2. Upload `catalog-number/mapping.csv`
3. Verify columns auto-detect: `fileName` + `catalogNumber`
4. Click Proceed → 6 rows show "Awaiting file"
5. Drop all 6 `.jpg` files → status should clear
6. **Validate → Upload → Rollback**
7. Repeat with `mapping-multi.csv`

### 2. `taxon/`: Taxon → Full Name

**Steps:**
1. Create new dataset → **Match by mapping file**  
   (must have a Taxon table with matching records in your collection)
2. Upload `taxon/mapping.csv`
3. Verify columns auto-detect: `File Name` + `Full Name` (based on schema captions)
4. Click Proceed → 6 rows
5. Drop all 6 `.jpg` files
6. Validate, Upload, Rollback

### 3. `reference-work/`: Reference Work → Title

**Steps:**
1. Create new dataset → **Match by mapping file**  
   (must have a Reference Work table with matching records)
2. Upload `reference-work/mapping.csv`
3. Verify columns auto-detect: `fileName` + `Title`
4. Click Proceed
5. Drop all 5 images
6. Validate, Upload, Rollback

## Special Cases

- [ ] **CSV row without a matching file** — shows "Awaiting file" forever; Validate warns
    image

- [ ] Make sure that files not in the CSV are handled (they will not be matched or uploaded)
    image

- [ ] Make sure that duplicate filenames in CSV are handled, showing only one row after import
- [ ] Verify that when re-opening a saved dataset the mapping CSV data and column selections are preserved
- [ ] Test rollback and random interruptions (cancel during actions), make sure behavior is consistent with our current bulk attachment uploader
- [ ] Make sure the matching mode dialog shows on first visit (and subsequent visits if the data set is empty), but stays hidden after choosing (even on refresh/re-open)

- [ ] Verify that uploading and matching to these new tables and fields is functional:

https://github.com/specify/specify7/blob/d22e16db54906da69121b705cddfe9f6d2e06d79/specifyweb/frontend/js_src/lib/components/AttachmentsBulkImport/importPaths.ts#L45-L48

### Cases that are not handled
- Currently, you cannot upload a single attachment to multiple objects. This means you can't say `8.jpg` and list it multiple times in the mapping for different linked objects.

---

## Summary by CodeRabbit
- **New Features**
  - Added bulk attachment “match by mapping file” mode with CSV/TSV support and a matching-mode dialog.
  - Enabled column selection to map CSV match values to target filenames.
- **Bug Fixes**
  - Added server-side validation for the dataset matching mode and preserves the existing mode when omitted.
- **UI Improvements**
  - Show mapping-mode notes on dataset links and display mapping-specific messaging for missing/unmatched/duplicate rows.
- **Tests**
  - Expanded Jest coverage for mapping selection, deduplication, and placeholder handling.
- **Style/Documentation**
  - Minor formatting cleanups for URLs and links.


specifyweb/backend/attachment_gw/migrations/0002_add_matchingmode.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('attachment_gw', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='spattachmentdataset',
```
grantfitzsimmons, line 34: Waiting a bit on the way we finalize creating schema captions for fields. I imagine we may want to have a schema config entry for this field? No real substantive benefit for it, but the sync fields command would create one, so maybe better to add here.

specifyweb/frontend/js_src/lib/components/AttachmentsBulkImport/MappingFileSetup.tsx
```diff
@@ -0,0 +1,320 @@
...
+                  value === '' ? undefined : Number.parseInt(value)
+                )
+              }
+            >
+              <option disabled value="">
+                {commonText.select()}
+              </option>
+              {headers.map((header, index) => (
+                <option key={index} value={index}>
+                  {header || `Column ${index + 1}`}
```
melton-jason: Make sure to localize the default header before merging 👀

specifyweb/frontend/js_src/lib/components/AttachmentsBulkImport/MappingFileSetup.tsx
```diff
@@ -0,0 +1,320 @@
...
```
melton-jason: Same thing as the above comment, make sure the default header is localized before merging

emenslin (changes requested):
### 1. `catalog-number/`: Collection Object → Catalog Number

**Steps:**
- [x] **Validate → Upload → Rollback**
- [x] Repeat with `mapping-multi.csv`

### 2. `taxon/`: Taxon → Full Name

**Steps:**
- [x] Validate, Upload, Rollback

### 3. `reference-work/`: Reference Work → Title

**Steps:**

- [x] Validate, Upload, Rollback

## Special Cases

- [x] **CSV row without a matching file** — shows "Awaiting file" forever; Validate warns

- [x] Make sure that files not in the CSV are handled (they will not be matched or uploaded)

- [x] Make sure that duplicate filenames in CSV are handled, showing only one row after import
- [x] Verify that when re-opening a saved dataset the mapping CSV data and column selections are preserved
- [x] Test rollback and random interruptions (cancel during actions), make sure behavior is consistent with our current bulk attachment uploader
- [ ] Make sure the matching mode dialog hows on first visit (and subsequent visits if the data set is empty), but stays hidden after choosing (even on refresh/re-open)

---

Overall it looks good! I did notice that it always auto selects match to filename even if you last uploaded a mapping file which doesn't match up with the last point on the checklist. 

https://github.com/user-attachments/assets/b50d786a-13e5-4c6a-bbfd-656a4a83a3c3


Additionally if you have a row without a matching file, upload is disabled. I don't think this is necessarily a problem since it works similarly in bulk attachment upload, but I wanted to make note of it because I didn't realize I missed a file and was briefly confused why I couldn't upload. 

Lastly, another one that's not necessarily an issue but when you upload a mapping file, it'll correctly match the file name and value columns in the dialog but then when you go to the data set itself you have to reselect the path. I imagine this is the intended behavior, but we'll want to make sure to document it since I could see it being assumed that the path would automatically be selected.

grantfitzsimmons:
@emenslin I think you can approve this PR if I understand you correctly! Thanks for taking the time to review. 😁

> Overall it looks good! I did notice that it always auto selects match to filename even if you last uploaded a mapping file which doesn't match up with the last point on the checklist.

That bullet point is meant to communicate that the "Matching Mode" dialog should appear when you create a new attachment data set in only two cases:
* The first time the attachment data set is created
* If an existing attachment data set is opened, but no rows were added (e.g., no attachments or mapping files were selected)

It is not intended to remember the last choice you made, as the choice is unique for each attachment data set. It defaults to matching by filename, as that is the standard workflow at the moment and requires no additional input from the user.

>Additionally if you have a row without a matching file, upload is disabled. I don't think this is necessarily a problem since it works similarly in bulk attachment upload, but I wanted to make note of it because I didn't realize I missed a file and was briefly confused why I couldn't upload.

This is meant to replicate the current handling, as it prevents unexpected behavior, but I might adjust this to ignore rows that aren't part of the mapping instead.

>Lastly, another one that's not necessarily an issue but when you upload a mapping file, it'll correctly match the file name and value columns in the dialog but then when you go to the data set itself you have to reselect the path. I imagine this is the intended behavior, but we'll want to make sure to document it since I could see it being assumed that the path would automatically be selected.

This is the intended behavior, but I think there are two things happening here:

1. The user needs to map their spreadsheet columns to the "Match Value Column" and the "File Name Column":
    file-c5ed2580d4ebeafd9deb09ed2b320399

2. The user then needs to choose which field the "Match Value Column" is matching to.
    file-416c6062ed070192fb96aa459bf66790

I think it would be more intuitive from a UX perspective, perhaps in this PR or in the future, to have the path choice during the "matching mode" dialog. In a sense, you are not currently reselecting anything, but you are having to set things at two different points.

emenslin (approved):
> That bullet point is meant to communicate that the "Matching Mode" dialog should appear when you create a new attachment data set in only two cases:
> 
> * The first time the attachment data set is created
> * If an existing attachment data set is opened, but no rows were added (e.g., no attachments or mapping files were selected)
> 
> It is not intended to remember the last choice you made, as the choice is unique for each attachment data set. It defaults to matching by filename, as that is the standard workflow at the moment and requires no additional input from the user.

This makes a lot more sense, in this case it works as expected! 


> This is meant to replicate the current handling, as it prevents unexpected behavior, but I might adjust this to ignore rows that aren't part of the mapping instead.

Yeah, I could see it going either way. I think this behavior works as long as we document it, but I think either a dialog or something at the top of the screen (e.g. the warning when you have to reselect files) that informs the user why they can't upload since if you have a larger data set the only way to find out would be scrolling through and seeing that a file was missed. Either way I don't think this is necessarily an issue for this PR

> I think it would be more intuitive from a UX perspective, perhaps in this PR or in the future, to have the path choice during the "matching mode" dialog. In a sense, you are not currently reselecting anything, but you are having to set things at two different points.

I agree, I think having the path choice on the matching mode dialog would be much more intuitive. I think since it automatically detects the identifier in the matching mode dialog a user might think that they wouldn't have to select the path on the next screen and then might be confused when they can't upload. I don't think it is necessarily a problem, just not the most intuitive and should be documented. 

Overall, I don't think any of my issues need to be fixed in this PR as everything I tested worked correctly, just things we should either fix in the future or make sure is explained in documentation!

CarolineDenis (approved):

bronwyncombs (approved):
- [x] 1. `catalog-number/`: Collection Object → Catalog Number

- [x] 2. `taxon/`: Taxon → Full Name

- [x] 3. `reference-work/`: Reference Work → Title

## Special Cases

- [X] **CSV row without a matching file** — shows "Awaiting file" forever; Validate warns
    image

Note: Upload is disabled if/when a row has `Awaiting File` status, even if other rows are otherwise ready for uploading

- [X] Make sure that files not in the CSV are handled (they will not be matched or uploaded)
 
Note: Upload is enabled for other rows even if/when a row has `Not In Mapping File` status

- [X] Make sure that duplicate filenames in CSV are handled, showing only one row after import
- [X] Verify that when re-opening a saved dataset the mapping CSV data and column selections are preserved
- [X] Test rollback and random interruptions (cancel during actions), make sure behavior is consistent with our current bulk attachment uploader
- [X] Make sure the matching mode dialog shows on first visit (and subsequent visits if the data set is empty), but stays hidden after choosing (even on refresh/re-open)

specifysoftware:

https://speciforum.org/t/specify-6-eol-questions-attachment-mapping-files-and-repeat-label-printing/3871/4

grantfitzsimmons:
Waiting for `v7.12.0.7` to be merged so I can improve the migration

nfshoobs:
@grantfitzsimmons @emenslin will this allow bulk upload into tables not supported by the current bulk uploader? Or will it simply enable the use of this template file to upload into the currently supported tables?

grantfitzsimmons:
> @grantfitzsimmons @emenslin will this allow bulk upload into tables not supported by the current bulk uploader? Or will it simply enable the use of this template file to upload into the currently supported tables?

This will only support all currently supported tables in Batch Attachment Uploader!

Do you have an idea of which other tables you would like to upload, along with which identifier you want to match to?

nfshoobs:
FieldNotebookPageAttachment is the main one I see missing that is a pain to not have available in the GUI. Matching could be via the id field for the records (or any other field, honestly, like the pagenumber field)

grantfitzsimmons:
>FieldNotebookPageAttachment is the main one I see missing that is a pain to not have available in the GUI. 

I just added support for matching and uploading attachments just by matching to the `description` or `pageNumber` fields. I also added support for uploading Disposal attachments via matching to the `disposalNumber` field.

>Matching could be via the id field for the records (or any other field, honestly, like the pagenumber field)

It's a good suggestion! It would be nice to open this up more broadly in the future for any field

g1rly-c0d3r (changes requested):
### 1. `catalog-number/`: Collection Object → Catalog Number
**Steps:**
1. Create new dataset → **Match by mapping file**
2. Upload `catalog-number/mapping.csv`
3. Verify columns auto-detect: `fileName` + `catalogNumber`
4. Click Proceed → 6 rows show "Awaiting file"
5. Drop all 6 `.jpg` files → status should clear
6. **Validate → Upload → Rollback**
7. Repeat with `mapping-multi.csv`

### 2. `taxon/`: Taxon → Full Name

**Steps:**
1. Create new dataset → **Match by mapping file**  
   (must have a Taxon table with matching records in your collection)
2. Upload `taxon/mapping.csv`
3. Verify columns auto-detect: `File Name` + `Full Name` (based on schema captions)
4. Click Proceed → 6 rows
5. Drop all 6 `.jpg` files
6. Validate, Upload, Rollback

### 3. `reference-work/`: Reference Work → Title

**Steps:**
1. Create new dataset → **Match by mapping file**  
   (must have a Reference Work table with matching records)
2. Upload `reference-work/mapping.csv`
3. Verify columns auto-detect: `fileName` + `Title`
4. Click Proceed
5. Drop all 5 images
6. Validate, Upload, Rollback

## Special Cases

- [ ] **CSV row without a matching file** — shows "Awaiting file" forever; Validate warns

- [x] Make sure that files not in the CSV are handled (they will not be matched or uploaded)

- [x] Make sure that duplicate filenames in CSV are handled, showing only one row after import
- [x] Verify that when re-opening a saved dataset the mapping CSV data and column selections are preserved
- [x] Test rollback and random interruptions (cancel during actions), make sure behavior is consistent with our current bulk attachment uploader
- [x] Make sure the matching mode dialog shows on first visit (and subsequent visits if the data set is empty), but stays hidden after choosing (even on refresh/re-open)
- [x] Verify that uploading and matching to these new tables and fields is functional:

---

I did not ever get a validation warning for when a line in the mapping file is missing its attachment file, but the upload button was disabled:

https://github.com/user-attachments/assets/b7b6404a-5a2a-480c-9a1a-4d198ba1220a

grantfitzsimmons:
@g1rly-c0d3r Thanks for your review! Added a new dialog to handle reporting this (and I put them as bullet point items):

image

g1rly-c0d3r (approved):
Looks good! The dialog shows up now, and nothing broke, upload/roll back work as expected, even when canceling the upload when some files have not been uploaded, rolling back, then re-uploading.

PR #8313, opened 2026-07-15
Revert: Grant migrator privileges even when the user already exists

### LINKED ISSUE #8305: Cannot setup a new Specify 7 database locally / cannot access Specify 7 databases not opened in v7.12 locally
**Describe the bug**
When attempting to create a new database locally, it fails because the user `specify_migrator` does not have migration privileges. This is because the privilege grant updates only occur if the migrator user is a new user in specifically the `v7_12_0_7` branches.

This additionally affects opening databases that were not opened in v7.12, so if a user was to upgrade from an earlier version straight to v7.12.0.7, they'd be met with an error.

This issue does not exist on `v7_12_0_6` and the part of the code that causes it is not present in `main` either. It is [this specific conditional statement](https://github.com/specify/specify7/blob/ab280959c9bc87568127546a145006053852fe21/sp7_db_setup_check.sh#L217) that is causing it from what I can tell.

It does not appear in test panel due to test panel's unique configuration.

**To Reproduce**

1. Adjust your .env file so that the database name is one that doesn't exist, or one that was not used locally in v7.12.0.6.
2. Restart the docker containers with the updated .env file.
3. See the error message in the logs of the `specify7` container.

You will not be able to access the application unless you've accessed the database in v7.12.0.6.

**Expected behavior**
There should not be any errors preventing database setup.

Currently, the code change reflects concerns regarding granting privilege to the Specify 7 user. I believe that the current solution would work in the case where the institution is just coming to Specify 7, meaning that the migrator user hasn't been created yet. This would create a user that only has permissions on their single Specify 7 databases.

However, while I agree with this approach for safety, we also need to consider usability. If you have another Specify 7 database, this prevents creating a new one or opening an older one once you've set up Specify 7 once.

We should be sure to inform the user of the SQL user being created in this process in documentation and ideally in the actual setup stages (as a shell Y/N confirmation of some kind).

**Screenshots**
Here's an image of the logs (with potentially sensitive information removed). The blue is the database that I'm attempting to run on v7.12.0.7, but it was created and used previously. It was not opened on v7.12.0.6. All of the red are databases that I accessed while on v7.12.0.6.

Image

While this isn't an example of attempting to create a new database, the message was much of the same when doing so, except for the 'database already exists' message.

- OS: Windows 10
- Browser: N/A
- Specify 7 Version: `v7_12_0_6_base`
- Database Name: In private channels.
- Collection name: N/A
- User Name: N/A
- URL: N/A

CarolineDenis:
Fixes #8305 

### Checklist

- [ ] Add automated tests

### Testing instructions

1. Check that you have a `specify_migrator`* and `specify_user`* already created in MySQL.
  - _*The users may not be of those names. They need to match the `MIGRATOR_NAME` and `APP_USER_NAME` in the _.env_ file._
 2. Edit the `.env` file so that the DATABASE_NAME is one that does not exist yet.
 3. Build/rebuild your Docker containers.
   - [ ] Check the `specify7` container logs and make sure there are no migration errors.
4. Open the application in your browser.
  - [ ] Make sure that it's opening the Specify 7 initial database setup screen.
5. Stop the containers.
6. Edit the `.env` file so that the DATABASE_NAME is one that does exist and was not previously opened in v7.12.
  - You may run into migration errors while testing an old database.
7. Build/rebuild your Docker containers.
  - [ ] Check the `specify7` container logs and make sure there are no migration errors.
8. Open the application in your browser.
  - [ ] Make sure that it's opening the Specify 7 login page.

Additional checks:
- [ ] See that changing the migrator name to master in `.env` prints the proper log message.
- [ ] See that changing the app user name to master in `.env` prints the proper log message.
- [ ] See that changing the app user name to the migrator in `.env` prints the proper log message.
- [ ] See that having an underscore or other characters that commonly need to be escaped does not affect database setup.

### DIFF
--- a/seed-database/README.md
+++ b/seed-database/README.md
@@ -2,6 +2,11 @@
 
 Used by development docker composition only.
 
+Files in this folder are mounted into MariaDB's `/docker-entrypoint-initdb.d`.
+They are only processed when `/var/lib/mysql` is empty, so reusing the
+`database` volume will skip these imports. If you want to re-run the seed,
+remove the volume first (for example with `docker compose down -v`).
+
 Put an `.sql` dump of your database in this directory. You can call the
 file anything you want as long as it ends with `.sql`.
 
--- a/sp7_db_setup_check.sh
+++ b/sp7_db_setup_check.sh
@@ -216,14 +216,16 @@ fi
 
 if [[ "$NEW_MIGRATOR_USER_CREATED" -eq 1 ]]; then
   echo "Granting privileges to new user..."
-  echo "Executing: mysql -h \"$DB_HOST\" -P \"$DB_PORT\" -u \"$MASTER_USER_NAME\" --password=\"<hidden>\" -e \"GRANT ALL PRIVILEGES ON \`${SQL_DB_IDENTIFIER_ESCAPED_FOR_LIKE}\`.* TO '${SQL_MIGRATOR_NAME}'@'${CLIENT_HOST}'; FLUSH PRIVILEGES;\""
-
-  if ! mysql -h "$DB_HOST" -P "$DB_PORT" -u "$MASTER_USER_NAME" --password="$MASTER_USER_PASSWORD" -e "GRANT ALL PRIVILEGES ON \`${SQL_DB_IDENTIFIER_ESCAPED_FOR_LIKE}\`.* TO '${SQL_MIGRATOR_NAME}'@'${CLIENT_HOST}'; FLUSH PRIVILEGES;"; then
-    echo "Error: Failed to grant privileges to new user."
-    exit 1
-  fi
 else
-  echo "Skipping privilege grant for migrator user: user already exists. Verifying privileges on '${DB_NAME}'..."
+  echo "Migrator user already exists. Refreshing privileges on '${DB_NAME}'..."
+fi
+
+echo "Executing: mysql -h \"$DB_HOST\" -P \"$DB_PORT\" -u \"$MASTER_USER_NAME\" --password=\"<hidden>\" -e \"GRANT ALL PRIVILEGES ON \`${SQL_DB_IDENTIFIER_ESCAPED_FOR_LIKE}\`.* TO '${SQL_MIGRATOR_NAME}'@'${CLIENT_HOST}'; FLUSH PRIVILEGES;\""
+
+if ! mysql -h "$DB_HOST" -P "$DB_PORT" -u "$MASTER_USER_NAME" --password="$MASTER_USER_PASSWORD" \
+  -e "GRANT ALL PRIVILEGES ON \`${SQL_DB_IDENTIFIER_ESCAPED_FOR_LIKE}\`.* TO '${SQL_MIGRATOR_NAME}'@'${CLIENT_HOST}'; FLUSH PRIVILEGES;"; then
+  echo "Error: Failed to grant privileges to migrator user."
+  exit 1
 fi
 
 GRANTS_OUTPUT="$(mysql -N -B --raw -h "$DB_HOST" -P "$DB_PORT" \
--- a/sp7_db_setup_check.sh
+++ b/sp7_db_setup_check.sh
@@ -197,6 +197,11 @@ else
 fi
 
 # Create migrator user if it doesn't exist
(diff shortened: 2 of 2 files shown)


combs-a (changes requested):
I wrote these testing instructions, but they may need to be added onto if anyone spots any additional coverage needed!
<hr>
New database:

- [X] Check the `specify7` container logs and make sure there are no migration errors.
- [X] Make sure that it's opening the Specify 7 initial database setup screen.

Previously existing database (used pre v7.12):
- [ ] Check the `specify7` container logs and make sure there are no migration errors.
- [ ] Make sure that it's opening the Specify 7 login page.
<hr>

New database worked as expected.
Image

Was able to create a new database with no issues, but ran into an issue with the Specify app user when attempting to open an old database. Used the same database as I did in the originally reported issue.

Image

combs-a:
Noting this here--the regex also needs to be checked; there is an issue with not forming the database names correctly. Had a DB with an underscore in it and so it attempted to grant permissions to `test\_db` rather than `test_db`.

Not quite in the scope of just this PR but should be fixed.

g1rly-c0d3r (commented):
New database:

- [X] Check the `specify7` container logs and make sure there are no migration errors.
- [X] Make sure that it's opening the Specify 7 initial database setup screen.

Previously existing database (used pre v7.12):
- [x] Check the `specify7` container logs and make sure there are no migration errors.
- [x] Make sure that it's opening the Specify 7 login page.
<hr>

New database works as expected, and as long as I destroy the database volume when rebuilding the containers (`docker compose down -v`), using an existing database works as well. 

If I don't destroy the volume, mariadb will not source the `seed-database/*.sql` files, and in this case I have to add 
```sql
CREATE DATABASE IF NOT EXISTS `...`
```
to the start of the .sql file, So I'm not sure if that is the intended behavior. If I omit the `-v`, then I do have migration errors, because the existing database is not loaded by the mariadb container. If this is intended behavior I can approve.

combs-a:
I haven't run into that issue, I'll look into it!

g1rly-c0d3r (approved):

g1rly-c0d3r:
I also tested the new regex formatting for database names with underscores in them, and that worked as well! There were no weird escape characters getting misinterpreted as part of the database name.

combs-a (approved):
New database:
- [X] Check the `specify7` container logs and make sure there are no migration errors.
- [X] Make sure that it's opening the Specify 7 initial database setup screen.

Pre-existing database:
- [X] Check the `specify7` container logs and make sure there are no migration errors.
- [X] Make sure that it's opening the Specify 7 login page.

Additional checks:
- [X] See that changing the migrator name to master in `.env` prints the proper log message.
- [X] See that changing the app user name to master in `.env` prints the proper log message.
- [X] See that changing the app user name to the migrator in `.env` prints the proper log message.
- [X] See that having an underscore or other characters that commonly need to be escaped does not affect database setup.

<hr>

I used `new_db` and `cash_prev712` (a pre-existing database) to test the regex, and it worked fine. Other special characters should maybe be tested for completeness, but I think generally other characters are not used in filenames.

Used a mix of pre-existing and new databases to test the name changes as well, and it seemed all good. 👍

rijulpoudel (approved):
Tested this locally and everything worked as expected.

### New database

- [x] Existing migrator and app users were detected correctly.
- [x] The users received the required privileges on the new database.
- [x] No migration or database permission errors occurred.
- [x] The initial Specify 7 database setup screen opened successfully.

### Pre-existing database

- [x] The existing database started without migration or permission errors.
- [x] Migrator and app-user privileges were refreshed and verified.
- [x] The Specify 7 login page opened successfully.

### Additional checks

- [x] Using the master credentials for the migrator printed the expected messages and worked correctly.
- [x] Using the master credentials for the app user printed the expected messages and worked correctly.
- [x] Using the migrator credentials for the app user printed the expected messages and worked correctly.
- [x] A database name containing underscores was handled correctly without unwanted escaping.

I saw some non-fatal Schema Config warnings for old tables, but they did not interrupt the migration pipeline or application startup.

Everything looks good to me.

specifysoftware:

https://speciforum.org/t/specify-7-12-0-7-release-announcement/3863/1

PR #8370, opened 2026-07-28
Flag unreadable direction letters instead of discarding them; stop a minus cancelling its direction

foozleface:
Stacked on #8369 — please review that one first; it is the change that actually stops the data loss, and this one is ineffective without it.

Two parsing defects that both produce a **confident wrong answer** where a parse failure would have been the correct outcome.

### 1. Unrecognised direction letters are deleted rather than flagged

`trimLatLong()` keeps only `[\s\d"'\-.:ensw°]`, so any other direction letter vanishes and the remainder parses cleanly — in the wrong hemisphere:

```
"90° 20'14\" O"  ->  "90° 20'14\""  ->  +90.337     (should be -90.337)
```

`O` is *Oeste* — west — on Spanish, Portuguese and French labels. This is not malformed input: natural history collections hold labels written over 150 years in the collector's own language, and OCR or transcription faithfully carries the direction letter across, as it should. It is correct source data in a language the parser does not speak.

The app already knows how to say "I cannot read this" — out-of-range values like `19°49'60"` raise a validation message. `parse()` now does the same when trimming would discard a letter **and** no recognised `n/s/e/w` survives.

#### Please note: deliberately NOT mapped to west

That is the tempting fix and it is wrong. *Oeste/Ouest* mean west; *Ost/Oost* mean **east** in German and Dutch. Both appear inside our single collection:

```
776 rows   "90° 20'14\" O"   ->  -90.337   Guatemala / Mexico   (Oeste = West)
  8 rows   "5° 1' O"         ->   +5.017   Germany, Poland      (Ost   = East)
```

A blanket mapping would repair 776 records and silently corrupt 8, and elsewhere the ratio could invert. The institution knows what language its labels are in; the parser cannot.

The "and no direction survives" condition is load-bearing rather than defensive coding: without it the rule also rejects **227** coordinates that parse correctly today, because U+00BA (`º`) — the masculine ordinal typed instead of a degree sign — is itself a Unicode letter.

*Related, not fixed here:* German "Ost" still mis-reads as **south**, since the O and t are discarded and the surviving `s` is taken as the direction. Catching that needs real localised direction words. It is pinned as a `KNOWN LIMITATION` test so it is visible rather than assumed handled.

### 2. A leading minus and a direction letter cancel

```js
const sign = (direction === 's' || direction === 'w' ? -1 : 1) * originalSign;

"-157.14015 W"  ->  +157.14015     (both mean west; result is east)
```

Now reconciled by agreement: the minus decides when there is no direction letter; minus with `s`/`w` agrees and stays negative; minus with `n`/`e` contradicts and is rejected as ambiguous. Both pre-existing invalid cases stay invalid — `-124:34:23 N` contradicts, and `-90:05 S` agrees to −90.083 which still exceeds the latitude limit.

We first tried rejecting *any* redundant sign, but that would have blocked 934 real coordinates whose meaning is unambiguous. Agreement corrects them instead.

### Measured impact

Across 853,520 coordinate strings from five collections, running the real parser before and after:

| | |
|---|---|
| identical | 851,429 |
| **corrected** | **934** |
| newly rejected — *all previously producing wrong values* | 1,156 |
| **correct values lost** | **0** |

### Two things we would like your opinion on

1. Those 1,156 become save-blockers, which is a real workflow cost. A warning may suit you better than a hard block — the goal is that someone is *told*, not necessarily stopped.
2. Unparseable coordinates plot at (0,0) via `Leaflet/helpers.ts` rather than failing visibly, so "flag it" is not equally visible everywhere.

### Tests

`latLong.hemisphere.test.ts` covers both defects plus regression guards for correct input, and pins the `º`/`deg` cases this change must *not* affect.

### DIFF
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/LatLongUi.tsx
@@ -105,10 +105,19 @@ function Coordinate({
     if (isLoading.current && value === undefined) return;
     else isLoading.current = false;
 
-    const trimmedValue = trimLatLong(value?.toString() ?? '');
+    const rawValue = value?.toString() ?? '';
+    const trimmedValue = trimLatLong(rawValue);
     const hasValue = trimmedValue.length > 0;
+    /*
+     * Parse the RAW value, not the trimmed one. parse() trims internally, but it
+     * also inspects what trimming would discard so that an unrecognised direction
+     * letter makes the value invalid instead of silently changing hemisphere.
+     * Handing it a pre-trimmed string throws that information away before the
+     * check can run — "96° 57' O" would already have become "96° 57'" and would
+     * parse happily as EAST.
+     */
     const parsed = hasValue
-      ? ((fieldType === 'Lat' ? Lat : Long).parse(trimmedValue) ?? undefined)
+      ? ((fieldType === 'Lat' ? Lat : Long).parse(rawValue) ?? undefined)
       : undefined;
 
     const isValid = !hasValue || parsed !== undefined;
--- a/specifyweb/frontend/js_src/lib/components/FormPlugins/__tests__/LatLongUi.test.tsx
+++ b/specifyweb/frontend/js_src/lib/components/FormPlugins/__tests__/LatLongUi.test.tsx
@@ -59,7 +59,12 @@ describe('LatLongUi does not mutate the resource on render', () => {
   test('sliding to another record does not rewrite it, even after an edit', async () => {
     const first = makeLocality();
     const { rerender } = render(
-      <LatLongUi id={undefined} latLongType="Point" resource={first} step={undefined} />
+      <LatLongUi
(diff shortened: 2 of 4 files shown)


bronwyncombs:
Comments and Requested Changes:

- [ ] Save block for unexpected characters
   - [x] save is prevented for unexpected values (e.g., `10 20' 13’ xyz"`, `100 30’ P`,  `90 20’14" A"`,  `friend`, `50 friend`)
   - [ ] Save block behavior in UI is matched:

The Clone and Add buttons should be disabled:

https://github.com/user-attachments/assets/74936563-5897-4e10-93da-5ff1a2a33aad

Note: Add clears the current record of all data; clone works

Proceeding with saved blocker should be disabled: 

https://github.com/user-attachments/assets/c83bb609-4e30-4772-87ff-7ebb46b00150


- [x] Coordinate direction parsing
   - [x] Double coordinates aren't cancelled out
   - [x] mismatched coordinates have save blocked

PR #8400, opened 2026-08-05
Fix: Update attachments in BE instead of cloning record

### LINKED ISSUE #8298: Batch editing an attachment record does not edit the record and instead duplicates it
**Describe the bug**
When batch editing attachments either through a related table's (i.e. Collection Object) or an Attachment table query, it instead creates a new record with the edited information in Batch Edit.

It does not appear to affect non-Attachment tables from a brief test with Collection Object, but we may want to do further testing to ensure that it is only the Attachment table.

There are no related errors, and would likely cause confusion due to the duplications of the attachment records in the Attachments tab. Additionally, Edit History does not contain information about fields on Create, so a lot of editing information would be quietly lost. The only real indicator is the New Cells being flagged by Batch Edit.

**To Reproduce**

1. Go to the record(s) you want to edit and note the Attachment ID for the attachment you want to edit.
2. Create a query for batch editing, either through the Attachment table or a table with a relationship to Attachments (i.e. Collection Object).
3. Save the query and then click Batch Edit.
4. Edit the record(s) in some way, and then save and commit.
    - _Note that when you save, it'll then appear as new cells._
6. Open the edited Attachment record. You can do so by right-clicking an Attachment field in Batch Edit.
7. Click on the Settings menu for the Attachment record and click Edit History.
8. See that there has been no record of the edit made.

You can see further proof of this in the connecting table (i.e. Collection Object Attachment), as it will show that the ID of the attachment has been updated.

**Expected behavior**
The edit history should be recorded.

**Screenshots**
Here is a video of the Batch Edit taking place and the Edit History not being updated.

https://github.com/user-attachments/assets/e626af0e-104e-4aef-8277-66b99f7ccf79

Here is the duplicated record shown in the Attachments tab:

Image

**Crash Report**

- OS: Windows 10
- Browser: Firefox
- Specify 7 Version: v7.12.0.7
- Database Name: sdnhmherps
- Collection name: Herp-PC
- User Name: testuser
- URL: https://sdnhmherps20260608-v71207base.test.specifysystems.org/specify/view/collectionobject/80735/

**Additional context**
Here's an additional video of editing the attachment through the forms and the edit history being updated, as expected. Editing attachments via forms does not duplicate the record.

https://github.com/user-attachments/assets/ff064169-f61e-46ce-83c2-6848176885b8

Notes: 
- First, follow the steps described above
- Go to attachment page 
- Click on newly created attachment (duplicate) 
- See that when clicking on the button "Collection Object" the original collection object shows up
- Click then on the original / old attachment 
- See that an error is thrown when clicking the button "CO" : Cannot read properties of undefined (reading '_tableName')

CarolineDenis:
Fixes #8298 

### Changes
- Added explicit identity-preservation support through the upload pipeline.
- Added schema support so plans containing preserveIdentity validate.

### Checklist

- [ ] Add automated tests

### Testing instructions

- Follow issue description and verify original attachment is being modified rather than having a new one being created 

**Notes**: 
Attachments were cloned because: 

In [upload_table.py:986](https://github.com/specify/specify7/blob/main/specifyweb/backend/workbench/upload/upload_table.py):

1. In update mode, non-one-to-one related records go through process_row unless preserveIdentity is set.
- See BoundUpdateTable._process_to_ones.
- Without preserveIdentity, attachment went down the process_row path, not save_row(force=True).
2. process_row does match-first, then upload if no match.
- See BoundUploadTable._handle_row and BoundUploadTable._match.
- In that upload path, if current_id exists, upload means clone, not update.
3. See BoundUploadTable._do_upload and BoundUploadTable._do_clone.

So when we changed attachment fields, the old record often no longer matched the match predicate, and the fallback was “create via clone” from current_id, which produced duplicate attachments.

For many non-one-to-one relationships, Specify treats records as potentially reusable/shared and avoids unconditional in-place mutation, that conservative default is useful broadly.

==> preserveIdentity  flips that specific to-one node from match/clone semantics to forced in-place save when there is an existing id

### Next 
A better long-term model is an explicit per-node update strategy in the upload plan.
Each table instance in the relationship graph explicitly declares how updates should behave, instead of relying on one global default.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed batch edits for existing attachment records, preserving their identities and relationships.
  * Attachment updates now keep record counts stable and maintain audit history.
  * Improved one-to-one upload handling to update existing records when configured.

* **New Features**
  * Upload plans now support optional identity preservation settings for upload tables and related records.


CarolineDenis:
@coderabbit full review

g1rly-c0d3r (approved):
- [x] Follow issue description and verify original attachment is being modified rather than having a new one being created

---

No duplicate attachment records, and I was able to see all of the edit history.

rijulpoudel (approved):
- [x] Follow issue description and verify original attachment is being modified rather than having a new one being created

---

Didn't find any duplicates and the history showed it being modified.

kwhuber (approved):
- [x] Follow issue description and verify original attachment is being modified rather than having a new one being created
---
No duplicates created.

Issue #3369, opened 2023-04-18
Query OperationalError (Too many tables; MariaDB can only use 61 tables in a join)


carlosmbe:
**Describe the bug**
Crash when query is run in Specify 7. Works fine in Specify 6.

**To Reproduce**
1. Go to `pripaleo` as LJensen
OR the updated pripaleo database
2. Click on Queries
3. Scroll down to Query - Collection Objects
4. Run Query and See error

**Expected behavior**
Same query results as Specify 6

**Crash Reports**

[Specify 7 Crash Report - 2023-04-18T21_26_37.218Z (1).txt](https://github.com/specify/specify7/files/11266493/Specify.7.Crash.Report.-.2023-04-18T21_26_37.218Z.1.txt)
[Specify 7 Crash Report - 2023-04-18T21_26_37.218Z.txt](https://github.com/specify/specify7/files/11266494/Specify.7.Crash.Report.-.2023-04-18T21_26_37.218Z.txt)
[Specify 7 Crash Report - 2023-04-18T21_26_34.638Z (1).txt](https://github.com/specify/specify7/files/11266495/Specify.7.Crash.Report.-.2023-04-18T21_26_34.638Z.1.txt)
[Specify 7 Crash Report - 2023-04-18T21_26_34.638Z.txt](https://github.com/specify/specify7/files/11266496/Specify.7.Crash.Report.-.2023-04-18T21_26_34.638Z.txt)

**Screenshots**

![Screenshot 2023-04-18 at 4 23 26 PM](https://user-images.githubusercontent.com/53784701/232908645-cfe9a4a6-9c94-46d0-ab06-0944e184e23f.png)
![Screenshot 2023-04-18 at 4 23 39 PM](https://user-images.githubusercontent.com/53784701/232908648-82541633-df8c-4d8d-8e83-33439791eada.png)
![Screen Shot 2023-04-18 at 4 20 39 PM](https://user-images.githubusercontent.com/53784701/232908654-0e4c3638-8fc9-4854-bc0e-9d77a4c7e598.png)
![Screen Shot 2023-04-18 at 4 20 25 PM](https://user-images.githubusercontent.com/53784701/232908656-c5652064-327b-4556-a24c-23df5870cb42.png)

Desktop:
OS: Mac OSx Monterey
Browser: Chrome
Specify 7 Version: 7.8.10-prerelease

Database Name: pripaleo

Reported By
Carlos @carlosmbe

emenslin:
[...]

PR #8510, opened 2026-09-10
fix(queries): skip missing tree ranks

### LINKED ISSUE #3351: Invalid Query Caused By Non-Existent Tree Ranks
**Describe the bug**
As found out through the investigation of #3340 that some queries created and saved in Specify 6 can have fields that 7 treats as unmapped. 
When one of these queries is imported into the stats page. It just produces a stat called "Error"

This normally wouldn't be an issue as 7 doesn't let you save a query with unmapped fields. 

**To Reproduce**
1. Go to `DAO_Test` Database and log in as `hacole`. Enter the `Herbarium Collection`
2. Click on Statistics 
3. Scroll down to '哈哈哈哈' Alternatively. Just import the `GPI` query to any stat 

**Expected behavior**
A more intuitive description of the error. Or refusal to allow queries with unmapped fields 

**Screenshots**
Screenshot 2023-04-14 at 12 06 05 PM

Desktop:
OS: Mac OSx Ventura
Browser: Chrome
Specify 7 Version: 7.8.10-prerelease

Database Name:  `DAO_Test`

Reported By
@carlosmbe

grantfitzsimmons:
Fixes #3351

If a query, data export, (or soon, data view query) is executed in a collection that doesn't have matching ranks, this prevents it from crashing. This is a common problem for databases that share queries between collections, especially when those collections exist in other disciplines where some tree ranks might not exist. It also allows us to ship default queries or export mappings that function even if the ranks included in the defaults are not present in the collection(s).

file-aadc6b629ebf5b71ba3717018319b345

file-908159827dfe4acddce4929f932febe9

### Checklist

- [X] Add automated tests

### Testing instructions

1. Open or create a stored query containing a tree-rank field that is not present in the active collection. (If you are having a hard time finding one, I guess you can use this: [FakeRankQuery.json](https://github.com/user-attachments/files/32052427/FakeRankQuery.json))
2. Execute the query in the table view and confirm that it does not crash and that the remaining result columns stay aligned with their headers.
3. Apply a filter or sort using the unavailable rank and confirm that the query still executes while ignoring that field.
4. Export the same query and confirm that the export completes and the unavailable field is omitted or empty without shifting other columns.

There is automated coverage should verify both a missing displayed rank (preserving the result-column position) and a missing filtered/sorted rank (ignoring the field without changing the result set).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed query results so unresolved fields retain the correct column positions.
  * Prevented invalid filtering and sorting when a query field cannot be resolved.
  * Improved handling of queries involving unavailable or unmatched tree fields.
  * Added warnings instead of failing when tree-based fields cannot be matched.

### DIFF
--- a/specifyweb/backend/stored_queries/execution.py
+++ b/specifyweb/backend/stored_queries/execution.py
@@ -1086,6 +1086,10 @@ def build_query(
         )
 
         if field is None:
+            # Keep the result column in place when a displayed tree rank is
+            # missing, while omitting its filter and sort behavior.
+            if fs.display:
+                query = query.add_columns(sql.null())
             continue
 
         formatted_field = None
--- a/specifyweb/backend/stored_queries/query_construct.py
+++ b/specifyweb/backend/stored_queries/query_construct.py
@@ -29,6 +29,7 @@ class QueryConstruct(namedtuple('QueryConstruct', 'collection objectformatter qu
 
     def handle_tree_field(self, node, table, tree_rank: TreeRankQuery, next_join_path, current_field_spec: QueryFieldSpec):
         query = self
+        query_before_tree_joins = query
         if query.collection is None:  # Not sure it makes sense to query across collections
             raise AssertionError(
                 f"No Collection found in Query for {table}",
@@ -70,7 +71,13 @@ class QueryConstruct(namedtuple('QueryConstruct', 'collection objectformatter qu
             if (tree_rank.treedef_id is None or tree_rank.treedef_id == treedef_id)
             ] if tup[1] is not None]
 
-        assert len(treedefs_with_ranks) >= 1, "Didn't find the tree rank across any tree"
+        if not treedefs_with_ranks:
+            logger.warning(
+                "Didn't find tree rank %r across any %s tree; skipping field",
+                tree_rank.name,
+                table.name,
+            )
+            return query_before_tree_joins, None, None, table
 
         treedefitem_params = [treedefitem_id for (_, treedefitem_id) in treedefs_with_ranks]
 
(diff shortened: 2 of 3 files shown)


g1rly-c0d3r (approved):
### Testing instructions

- [x] Open or create a stored query containing a tree-rank field that is not present in the active collection. (If you are having a hard time finding one, I guess you can use this: [FakeRankQuery.json](https://github.com/user-attachments/files/32052427/FakeRankQuery.json))
- [x] Execute the query in the table view and confirm that it does not crash and that the remaining result columns stay aligned with their headers.
- [x] Apply a filter or sort using the unavailable rank and confirm that the query still executes while ignoring that field.
- [x] Export the same query and confirm that the export completes and the unavailable field is omitted or empty without shifting other columns.

---

Everything looked good, none of the columns were shifted and no errors happened. I looked at the re-exported query and the fake rank was still in the export, but the query results were empty. I'm not sure if you meant that in the exported JSON the fake rank should be omitted.

JDAM2k4 (approved):
### Testing instructions

- [x] Open or create a stored query containing a tree-rank field that is not present in the active collection. (If you are having a hard time finding one, I guess you can use this: [FakeRankQuery.json](https://github.com/user-attachments/files/32052427/FakeRankQuery.json))
- [x] Execute the query in the table view and confirm that it does not crash and that the remaining result columns stay aligned with their headers.
- [x] Apply a filter or sort using the unavailable rank and confirm that the query still executes while ignoring that field.
- [x] Export the same query and confirm that the export completes and the unavailable field is omitted or empty without shifting other columns.

---

Seems good to me. This may also fix [#3369](https://github.com/specify/specify7/issues/3369) but I'm not sure.

grantfitzsimmons:
@g1rly-c0d3r 

>Everything looked good, none of the columns were shifted and no errors happened. I looked at the re-exported query and the fake rank was still in the export, but the query results were empty. I'm not sure if you meant that in the exported JSON the fake rank should be omitted.

It is working as intended! I meant to export only the query results, not the query itself. The rank should remain, but it needs to be empty. Thanks for message and for testing!

grantfitzsimmons:
>Seems good to me. This may also fix https://github.com/specify/specify7/issues/3369 but I'm not sure.

@JDAM2k4 Sadly I do not think it will fix that :-(  
Thanks for testing!

CarolineDenis (approved):

PR #8536, opened 2026-09-15
[test]: Unit Test that edits a record set in a previous version

### LINKED ISSUE #8535: [Test]: Edit a record set that was created in a previous version

gabek96:
Fixes [#8535](https://github.com/specify/specify7/issues/8535)

A Unit test that deletes a record set that was created in a previous version

## Summary by CodeRabbit

* **Tests**
  * Updated automated coverage for editing recordsets created in previous application versions.
  * Refined test setup to use the current recordset representation.
  * Simplified validation of recordset edits and refresh behavior.


rijulpoudel (changes requested):
Assertions are missing. It renames, saves and then ends so it would pass even if the update never got applied. It looks like you had the assertions in the first commits but later got dropped.

